PR #1, opened 2012-06-06
Split zipkin.thrift

franklinhu:
- Split zipkin.thrift into:
  - scribe.thrift - Contains the structs and scribe service interface
  - zipkinCore - Core Zipkin structs
  - zipkinCollector - structs and interface for the collector service
  - zipkinQuery - structs and interface for query service
- Fix relative import


johanoskarsson:
+1

PR #9, opened 2012-06-07
zipkin-web: query result sorting

franklinhu:
Allow sorting by:
- timestamp (asc|desc)
- duration  (asc|desc)


johanoskarsson:
+1

PR #11, opened 2012-06-07
Zipkin Gem

franklinhu:
Split out tracing instrumentation into individual gem rather than nested in zipkin-web.


franklinhu:
Woops merged the tracing and query logic. Will redo this

PR #13, opened 2012-06-08
zipkin-tracer gem

franklinhu:
Add tracing to a Rack application
- In config.ru: `use ZipkinTracer::RackHandler`
- Configurable in Rails configs


zipkin-gem/lib/zipkin-tracer.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  class RackHandler
+    def initialize(app)
+      @app = app
+      @lock = Mutex.new
+      ::Trace.tracer = ::Trace::ZipkinTracer.new(CarelessScribe.new(Scribe.new()), 10)
+    end
+
+    def call(env)
+      # TODO HERE BE HACK. Chad made me do it!
```
johanoskarsson: We should probably remove this. A start might be to see if we can serve the static content outside of Ruby.

zipkin-gem/lib/zipkin-tracer/zipkin.thrift
```diff
@@ -0,0 +1,287 @@
```
johanoskarsson: Let's make this file a symlink

zipkin-gems/zipkin-tracer/lib/zipkin-tracer.rb
```diff
@@ -0,0 +1,52 @@
...
+  class RackHandler
+    def initialize(app)
+      @app = app
+      @lock = Mutex.new
+      ::Trace.tracer = ::Trace::ZipkinTracer.new(CarelessScribe.new(Scribe.new()), 10)
+    end
+
+    def call(env)
+      id = ::Trace::TraceId.new(::Trace.generate_id, nil, ::Trace.generate_id, true)
+      ::Trace.default_endpoint = ::Trace.default_endpoint.with_service_name("zipkinui").with_port(0) #TODO any way to get the port?
```
johanoskarsson: The service name is hardcoded to zipkinui here, port is also set to 0 always.

zipkin-gems/zipkin-tracer/lib/zipkin-tracer.rb
```diff
@@ -0,0 +1,72 @@
...
+          10
+        end
+
+      ::Trace.tracer = ::Trace::ZipkinTracer.new(CarelessScribe.new(scribe), scribe_max_buffer)
+    end
+
+    def call(env)
+      id = ::Trace::TraceId.new(::Trace.generate_id, nil, ::Trace.generate_id, true)
+      ::Trace.default_endpoint = ::Trace.default_endpoint.with_service_name(@service_name).with_port(@service_port)
+      ::Trace.sample_rate=(1)
```
johanoskarsson: This should probably also be configurable and have a more sensible default

johanoskarsson:
+1

PR #14, opened 2012-06-08
Seems the EB repo has moved. Updated the url

johanoskarsson:


dvryaboy:
We will likely release 3.0 next week at which point everything should be in maven.twttr

franklinhu:
+1

PR #16, opened 2012-06-11
zipkin-web: sort order fix

franklinhu:
- Sort query results correctly when doing additional searches
- Fix typo


johanoskarsson:
+1

PR #19, opened 2012-06-12
zipkin-query gem

franklinhu:
Split access to Zipkin Query service out into separate gem
- Fix generated ruby namespacing in `zipkin-thrift`
- Thrift files are symlinked in `zipkin-gems/zipkin-query/thrift`, and the generated code is in `zipkin-gems/zipkin-query/vendor`


johanoskarsson:
+1

PR #20, opened 2012-06-12
Split common classes into zipkin-common

franklinhu:
- Moved:
  - `Annotation` (and spec)
  - `Endpoint` (and spec)
  - `IncompleteTraceDataException`
  - `Span` (and spec)
- Left `Trace`, `TraceSummary` in `zipkin-server` since they're specific to the query service
- Left `ZooKeeperClientService` since don't want to introduce a ZK dependency in common project
- `IncompleteTraceDataException` can be later removed when we pull out the Thrift dependencies


johanoskarsson:
+1

PR #21, opened 2012-06-12
zipkin-query gem thrift fix

franklinhu:
- Needed `Endpoint` thrift structs, so moved through over from `zipkin-web/lib`
- Bumped version `0.0.1` -> `0.0.2`


johanoskarsson:
+1

PR #24, opened 2012-06-13
zipkin-scrooge

franklinhu:
Pull scrooge/thrift dependency out of `zipkin-common` and `zipkin-server` into `zipkin-scrooge`. This will allow us to do further refactoring easily.
- Add `AnnotationType` and `BinaryAnnotation` classes for direct mirroring of thrift structs
- Unify `toThrift` and `fromThrift` methods in `ThriftAdapter` that handles the different structs defined in the thrift file. We can in the future make this configurable in the collector service.
- Use one set of `Constants` rather than the thrift generated ones
- Fix tests
  - Move thrift related tests to `ThriftAdapterSpec`


zipkin-common/src/main/scala/com/twitter/zipkin/common/Annotation.scala
```diff
@@ -25,24 +25,11 @@ import com.twitter.zipkin.gen
  * initiated the request.
  */
 object Annotation {
-  val CoreClient = Seq(gen.Constants.CLIENT_SEND, gen.Constants.CLIENT_RECV)
-  val CoreServer = Seq(gen.Constants.SERVER_SEND, gen.Constants.SERVER_RECV)
+  val CoreClient = Seq(Constants.CLIENT_SEND, Constants.CLIENT_RECV)
```
johanoskarsson: Could just use the CoreClient in constants instead?

zipkin-common/src/main/scala/com/twitter/zipkin/adapter/Adapter.scala
```diff
@@ -0,0 +1,26 @@
+package com.twitter.zipkin.adapter
+
+import com.twitter.zipkin.common._
+
+trait Adapter {
```
johanoskarsson, line 29: Would be nice with a comment here describing what the adapter does, what the annotaitonTypeType means and so on.

zipkin-scrooge/src/test/scala/com/twitter/zipkin/adapter/ThriftAdapterSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+        val expectedAnn: Annotation = Annotation(123, "value", Some(Endpoint(123, 123, "service")))
+        val thriftAnn: gen.Annotation = ThriftAdapter(expectedAnn)
+        val actualAnn: Annotation = ThriftAdapter(thriftAnn)
+        expectedAnn mustEqual actualAnn
+      }
+    }
+
+    "convert AnnotationType" in {
+      "to thrift and back" in {
+
```
johanoskarsson: Should this be empty?

johanoskarsson:
+1

PR #25, opened 2012-06-13
Reduce the garbage generated in the collector a bit

johanoskarsson:
Probably doesn't make much of a difference, but doesn't hurt either.


franklinhu:
+1, maybe add to the log message to say exactly what was changed

PR #27, opened 2012-06-14
Move Scribe collector/config to zipkin-scribe

franklinhu:
Move the Scribe specific code to a new module: `zipkin-scribe` including:
- ScribeCollectorService
- ScribeCollectorService related configs
- collector-dev config
- Main class
- related tests

The collector can no longer be run from `zipkin-server`.


johanoskarsson:
+1

PR #31, opened 2012-06-14
Modularize remaining collector Scribe dependencies

franklinhu:
This change makes it easier to extend the collector to accept different input types. Doing so would only require writing an `Adapter` and a `ProcessFilter`.

Changes:
- Add `ProcessorFilter`
  - Composable filter on top of `Processor`s that can transform the input data type
- Move all deserialization and extraction of data from `CollectorService` and `WriteQueueWorker` to Scribe related `ProcessFilter` and `Processor`. 
- `WriteQueueWorker` now takes only one `Processor` rather than a sequence of them.
- Add `FanoutProcessor` to blast work to multiple `Processor`s
- Update README with module diagram


zipkin-scribe/src/main/scala/com/twitter/zipkin/collector/ResilientZKNode.scala
```diff
@@ -1,12 +1,12 @@
 /*
  * Copyright 2012 Twitter Inc.
- *
```
johanoskarsson, line 3: No need for this file to be changed as discussed on IM

zipkin-scribe/src/main/scala/com/twitter/zipkin/collector/processor/ScribeProcessorFilter.scala
```diff
@@ -0,0 +1,56 @@
...
+package com.twitter.zipkin.collector.processor
+
+import com.twitter.logging.Logger
+import com.twitter.ostrich.stats.Stats
+import com.twitter.scrooge.BinaryThriftStructSerializer
+import com.twitter.zipkin.adapter.ThriftAdapter
+import com.twitter.zipkin.common.Span
+import com.twitter.zipkin.gen
+
+class ScribeProcessorFilter extends ProcessorFilter[Seq[gen.LogEntry], Seq[Span]] {
```
johanoskarsson, line 33: Would be nice with a class level comment explaining the purpose of this class, same for the other filters and new classes.

johanoskarsson:
+1

PR #32, opened 2012-06-14
Add two simple startup scripts for the collector and query daemons

johanoskarsson:


README.md
```diff
@@ -133,7 +133,8 @@ We've developed Zipkin with <a href="http://www.scala-lang.org/downloads">Scala
 1. `ssh [server]`
 1. `unzip zipkin*.zip`
 1. `mkdir -p /var/log/zipkin`
-1. Start the collector and query daemon.
+1. `zipkin-scribe/scripts/collector.sh -f zipkin-scribe/config/collector-prod.scala`
+1. `zipkin-server/scripts/query.sh -f zipkin-scribe/config/query-prod.scala`
```
franklinhu: query script should be in `zipkin-server/config/query-prod.scala` still

zipkin-server/src/scripts/query.sh
```diff
@@ -0,0 +1,3 @@
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+VERSION=@VERSION@
+java -cp "$DIR/../libs/*" -jar $DIR/../zipkin-scribe_2.9.1-$VERSION.jar $*
```
franklinhu: same as above

franklinhu:
+1

PR #34, opened 2012-06-15
Pull out msg from Thrift struct in ScribeCollectorService

franklinhu:
Move code that pulls out the base64 encoded Span from the Thrift struct back to the CollectorService since putting the entire Thrift struct in the write queue causes major GC regression.


johanoskarsson:
+1

PR #35, opened 2012-06-18
Misc changes

franklinhu:
- Add metric for batch size (how many entries are added to the write queue)
- Remove `OstrichProcessor` from default processors


johanoskarsson:
+1

PR #42, opened 2012-06-25
Fix build.properties output package

franklinhu:
Sets correct package `com.twitter.zipkin` for build.properties. This fixes the issue where curling `host:post/server_info.txt` returns unknown values.


johanoskarsson:
+1

PR #46, opened 2012-06-26
Top annotations

franklinhu:
Add collector and query support for adding "top annotation" data that may be computed offline
- New collector endpoints
  - `storeTopAnnotations`
  - `storeTopKeyValueAnnotations`
- New query endpoints
  - `getTopAnnotations`
  - `getTopKeyValueAnnotations`


johanoskarsson:
+1

PR #50, opened 2012-06-27
Add annotation suggestions support in UI

franklinhu:
- Upgrade to zipkin-query v0.0.3
- Add autocomplete dropdown with suggestions if they exist


johanoskarsson:
+1.

PR #51, opened 2012-06-27
Add static query linking

franklinhu:
- Show 'link' button for linking
- pushState the query params so going 'back' from a trace page loads
  the original search results


johanoskarsson:
+1.

PR #52, opened 2012-06-27
Handle case where kvAnnotation attached to null span

franklinhu:


franklinhu:
Fixed upstream in Ruby code in #53

PR #53, opened 2012-06-27
zipkin-web: Remove filtering of empty spans

franklinhu:
Remove filtering of empty spans to fix UI side bug for case where span only has a KV annotation


johanoskarsson:
+1

PR #54, opened 2012-06-29
Made preprocessing more complete

jerryli9876:
Preprocessed : Now only merges spans
FindNames : Finds client side and service names
FindIDtoNames : Finds (id, service name)
Other files modified to accommodate these


johanoskarsson:
+1

PR #56, opened 2012-06-29
Add some traces for our own query process

johanoskarsson:
Lazy val the rootmost span, no point in regenerating that


franklinhu:
+1

PR #58, opened 2012-07-02
Added ExpensiveEndpoints and test files

jerryli9876:
ExpensiveEndoints : per service call, finds average run time
Util.scala : Changed repeatSpan to allow it to model API level requests

### DIFF
--- a/zipkin-hadoop/src/main/scala/com/twitter/zipkin/hadoop/ExpensiveEndpoints.scala
+++ b/zipkin-hadoop/src/main/scala/com/twitter/zipkin/hadoop/ExpensiveEndpoints.scala
@@ -28,16 +28,15 @@ class ExpensiveEndpoints(args : Args) extends Job(args) with DefaultDateRangeJob
 
   val spanInfo = PreprocessedSpanSource()
     .read
+    .filter(0) { s : SpanServiceName => s.isSetParent_id() }
     .mapTo(0 -> ('id, 'parent_id, 'cService, 'service, 'annotations))
       { s: SpanServiceName => (s.id, s.parent_id, s.client_service, s.service_name, s.annotations.toList) }
-    .flatMap('annotations -> 'duration) {
-      al : List[Annotation] => {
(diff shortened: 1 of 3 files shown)


zipkin-hadoop/src/main/scala/com/twitter/zipkin/hadoop/ExpensiveEndpoints.scala
```diff
@@ -0,0 +1,64 @@
...
+    .mapTo(0 -> ('id, 'parent_id, 'cService, 'service, 'annotations))
+      { s: SpanServiceName => (s.id, s.parent_id, s.client_service, s.service_name, s.annotations.toList) }
+    .flatMap('annotations -> 'duration) {
+      al : List[Annotation] => {
+        var clientSend : Option[Annotation] = None
+        var clientReceive : Option[Annotation] = None
+        var serverReceive : Option[Annotation] = None
+        var serverSend : Option[Annotation] = None
+        al.foreach( {
+          a : Annotation => {
```
johanoskarsson: Our code style is to have this on the above line. I know that's slightly different from what IntelliJ reformats to but let's try to stick to our version.

zipkin-hadoop/src/main/scala/com/twitter/zipkin/hadoop/ExpensiveEndpoints.scala
```diff
@@ -0,0 +1,64 @@
...
+          // to deal with the case where there is no server duration
+        if (clientDuration == None) serverDuration else clientDuration
+      }
+    }
+
+  val idName = PrepTsvSource()
+    .read
+  /* Join with the original on parent ID to get the parent's service name */
+  val spanInfoWithParent = spanInfo
+    .filter('parent_id){ id : Long => id != 0 }
```
johanoskarsson: Is it possible to use the Thrift isSetParent_id (or whatever it is called) here instead? As mentioned the id !=0 is fine as a hack, but if possible we should figure out the correct solution.

johanoskarsson:
+1

PR #63, opened 2012-07-06
Added preliminary script stuff for hadoop jobs

jerryli9876:


project/Project.scala
```diff
@@ -26,8 +26,12 @@ object Zipkin extends Build {
       parallelExecution in Test := false,
       libraryDependencies ++= Seq(
         "com.twitter" % "scalding_2.9.1"       % "0.5.3",
-        "com.twitter.elephantbird" % "elephant-bird-cascading2"       % "3.0.0",
-
+        "com.twitter.elephantbird" % "elephant-bird-cascading2"       % "3.0.0" intransitive(),
```
johanoskarsson, line 33: Let's add a comment here describing that the deps below are from elephant bird and why we made that intransitive.

project/Project.scala
```diff
@@ -37,7 +41,13 @@ object Zipkin extends Build {
           "Concurrent Maven Repo" at "http://conjars.org/repo")
         case Some(pr) => Seq() // if proxy is set we assume that it has the artifacts we would get from the above repo
       }),
+
+      mainClass in (Compile, packageBin) := Some("com.twitter.scalding.Tool"),
```
johanoskarsson: Do we need all of these now?

project/Project.scala
```diff
@@ -54,6 +64,56 @@ object Zipkin extends Build {
...
+      <dependencies>
+          <exclude module="jms"/>
+          <exclude module="jmxri"/>
+          <exclude module="jmxtools"/>
+          <exclude org="com.sun.jdmk"/>
+          <exclude org="com.sun.jmx"/>
+          <exclude org="javax.jms"/>
+          <exclude org="org.mortbay.jetty"/>
+      </dependencies>,
+    mergeStrategy in assembly := {
```
johanoskarsson, line 84: This isn't needed since the job runner isn't using sbt-assembly.

project/Project.scala
```diff
@@ -54,6 +64,56 @@ object Zipkin extends Build {
...
+    StandardProject.newSettings ++
+    assemblySettings ++
+    TravisCiRepos.newSettings).settings(
+
+    name := "zipkin-hadoop_job_runner",
+    version := "0.2.0-SNAPSHOT",
+    parallelExecution in Test := false,
+    libraryDependencies ++= Seq(
+      "com.twitter" % "scalding_2.9.1"       % "0.5.3",
+      "com.twitter.elephantbird" % "elephant-bird-cascading2"       % "3.0.0" intransitive(),
```
johanoskarsson: Do you need elephantbird for the job runner too?

zipkin-hadoop_job_runner/src/main/scala/com/twitter/zipkin/hadoop/ProcessPopularKeys.scala
```diff
@@ -0,0 +1,91 @@
...
+object ProcessPopularKeys {
+  def main(args : Array[String]) {
+    val c = new PopularKeysClient()
+    val portNumber = augmentString(args(2)).toLong
+    c.start(args(0), args(1), portNumber)
+  }
+}
+
+/**
+ * Connects to the server, then processes data from PopularKeys and sends it there
```
johanoskarsson: Let's add a bit more meat here. Make sure it's clear that it's a collector server we're talking about and that the popular keys power the typeahead part of the lookup page on the ui (the annotations/kv annotations input boxes).

project/Project.scala
```diff
@@ -8,7 +8,7 @@ import java.io.File
 object Zipkin extends Build {
 
   lazy val zipkin = Project(id = "zipkin",
-                            base = file(".")) aggregate(hadoop, test, thrift, server, common, scrooge, scribe)
+                            base = file(".")) aggregate(hadoop, hadoop_job_runner, test, thrift, server, common, scrooge, scribe)
```
johanoskarsson: Let's call it hadoop-job-runner instead

PR #66, opened 2012-07-06
Move Trace to zipkin-common (part 1)

franklinhu:
Moved TraceTimeline and TimelineAnnotation to zipkin-common, pulled out
some thrift dependencies

### DIFF
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/query/TimelineAnnotation.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/query/TimelineAnnotation.scala
@@ -17,5 +17,9 @@ package com.twitter.zipkin.query
 
 import com.twitter.zipkin.common.Endpoint
 
+/**
+ * Extension of `Annotation` that also includes span/service information useful for query side
+ * responses
+ */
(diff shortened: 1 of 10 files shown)


zipkin-common/src/main/scala/com/twitter/zipkin/query/TimelineAnnotation.scala
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.twitter.zipkin.query
+
+import com.twitter.zipkin.common.Endpoint
+
+case class TimelineAnnotation(timestamp: Long, value: String, host: Endpoint, spanId: Long, parentId: Option[Long],
```
johanoskarsson, line 24: Let's add a comment here and on the TraceTimeline that explains what it is

johanoskarsson:
+1 except the above nit

PR #67, opened 2012-07-07
Move trace (part 2)

franklinhu:
- Move Trace to zipkin-common
- Pull out thrift dependencies into zipkin-scrooge


johanoskarsson:
+1

PR #68, opened 2012-07-09
fixed zipkin-web to actually return values from servers

mosesn:
Ruby 1.9.3 fixed a problem where do loops were sometimes not returning their last value.  However, for compatibility with 1.9.2, the with_transport do loops need their last value to be assigned to a variable with a greater scope than the do loop, ie to be initialized outside of the do loop and then assigned within.  Either the necessary ruby version should be specified in the documentation, or this fix needs to be pulled in.

### DIFF
--- a/zipkin-web/app/controllers/traces_controller.rb
+++ b/zipkin-web/app/controllers/traces_controller.rb
@@ -119,16 +119,18 @@ class TracesController < ApplicationController
 
   def top_annotations
     service_name = params[:service_name] || ""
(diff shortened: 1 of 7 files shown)


johanoskarsson:
Good point about the version. We actually use the old school Ruby 1.8.7. This week is kinda crazy but I'll try to make some time to try this pull request out on 1.8.7 too. 
As a more long term solution @franklinhu is working on a branch that will replace the ruby parts with scala. It turns out that we didn't need a lot of logic in the web layer besides what the query daemon can do and the javascript.

franklinhu:
Merged thanks!

PR #69, opened 2012-07-17
Added simple job running script

jerryli9876:
run.sh : runs all job
run_job.sh : runs a single job

### DIFF
--- a/zipkin-hadoop/src/main/scala/com/twitter/zipkin/hadoop/sources/Util.scala
+++ b/zipkin-hadoop/src/main/scala/com/twitter/zipkin/hadoop/sources/Util.scala
@@ -28,6 +28,7 @@ import com.twitter.zipkin.gen
 object Util {
 
   val UNKNOWN_SERVICE_NAME = "Unknown Service Name"
+  val ERROR_MESSAGE = "500 Internal Server Error"
 
   /**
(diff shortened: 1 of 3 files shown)


zipkin-hadoop/src/main/scala/com/twitter/zipkin/hadoop/sources/Util.scala
```diff
@@ -28,6 +28,7 @@ import com.twitter.zipkin.gen
 object Util {
 
   val UNKNOWN_SERVICE_NAME = "Unknown Service Name"
+  val ERROR_MESSAGE = "500 Internal Server Error"
```
johanoskarsson: Is this used anywhere? I'd prefer if we could put it somewhere other than util. Ideally wherever it is used, but if that's not doable let's create a new file called AnnotationConstants or something like that.

zipkin-hadoop/src/scripts/run.sh
```diff
@@ -0,0 +1,71 @@
...
+
+usage() {
+cat << EOF
+usage: $0 options
+
+This script runs the hadoop queries
+
+OPTIONS:
+   -h      Show this message
+   -d      Date, either as one date, or start date and end date, separated by commos
```
johanoskarsson: Typo

PR #71, opened 2012-07-18
Added WhaleReport and spec

jerryli9876:
WhaleReport finds all traces w/ 500 Internal Server Errors and finds all spans in those traces with retries and/or timeouts


zipkin-hadoop/src/main/scala/com/twitter/zipkin/hadoop/WhaleReport.scala
```diff
@@ -0,0 +1,81 @@
...
+  val spanInfo = PreprocessedSpanSource()
+    .read
+    .mapTo(0 ->('trace_id, 'id, 'service, 'annotations, 'binary_annotations)) {
+    s: SpanServiceName => (s.trace_id, s.id, s.service_name, s.annotations.toList, s.binary_annotations.toList)
+  }
+
+  val errorTraces = spanInfo
+    .project('trace_id, 'binary_annotations)
+    .filter('binary_annotations) {
+    bal: List[BinaryAnnotation] =>
```
johanoskarsson, line 42: Our convention is to have this on the line above

zipkin-hadoop/src/main/scala/com/twitter/zipkin/hadoop/WhaleReport.scala
```diff
@@ -0,0 +1,81 @@
...
+        }
+      })
+  }
+    .project('trace_id)
+    .rename('trace_id -> 'trace_id_1)
+
+  val filtered = spanInfo
+    .flatMap('annotations -> 'error) { al : List[Annotation] => {
+        var error : Option[String] = None
+        al.foreach { a : Annotation =>
```
johanoskarsson: We can use "find" here since it's only one entry we're looking for.

zipkin-hadoop/src/main/scala/com/twitter/zipkin/hadoop/WhaleReport.scala
```diff
@@ -0,0 +1,74 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.twitter.zipkin.hadoop
+
+import com.twitter.zipkin.gen.{BinaryAnnotation, Constants, SpanServiceName, Annotation}
+import com.twitter.scalding.{Tsv, DefaultDateRangeJob, Job, Args}
+import sources.{Util, PreprocessedSpanSource}
```
franklinhu: Nit: absolute imports are preferred to relative ones

johanoskarsson:
There's one unresolved nit, but other than that +1

PR #72, opened 2012-07-18
Add duration to annotations.

johanoskarsson:
Useful for timing blocks of code or other operations that don't warrant a full Span.


franklinhu:
+1

PR #73, opened 2012-07-18
Add static trace rendering page

franklinhu:
- Add export button on trace page
- Add /static to allow users to paste in the exported JSON to render the trace page

### DIFF
--- a/zipkin-web/app/controllers/traces_controller.rb
+++ b/zipkin-web/app/controllers/traces_controller.rb
@@ -39,9 +39,6 @@ class TracesController < ApplicationController
     @trace_id = params[:id]
   end
 
-  def static
-  end
-
   def get_trace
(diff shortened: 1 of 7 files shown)


zipkin-web/app/controllers/traces_controller.rb
```diff
@@ -39,6 +39,9 @@ def show
     @trace_id = params[:id]
   end
 
+  def static
```
johanoskarsson: Is this needed?

johanoskarsson:
+1

zipkin-web/app/controllers/traces_controller.rb
franklinhu: Apparently not. Handed by Rails black magic.

PR #75, opened 2012-07-18
Finatra

franklinhu:
Port `zipkin-web` from Rails to Scala (using Finatra) for a variety of reasons
- Simpler deployment (all JVM-based)
- Simpler web logic (no need to replicate all the common data classes)
- Operational transparency

Changes:
- JsonAdapter and JsonQueryAdapter
  - necessary since Javascript doesn't really support Longs, so we need to convert them to strings


franklinhu:
The vast majority of the js/css is copied over from zipkin-web and doesn't need to be reviewed. It's only there because sbt doesn't seem to package symlinked resources.

project/Project.scala
```diff
@@ -275,6 +280,31 @@ object Zipkin extends Build {
...
+      version := "0.2.0-SNAPSHOT",
+      resolvers += "finatra" at "http://repo.juliocapote.com",
+      resolvers += "codahale" at "http://repo.codahale.com",
+
+      libraryDependencies ++= Seq(
+        "com.posterous" % "finatra" % "4.3.2",
+
+        "com.twitter.common.zookeeper" % "server-set" % "1.0.7",
+
+        "com.twitter" % "finagle-serversets" % "5.0.0",
```
johanoskarsson: Have you tried pulling in the same version of finagle that we use for the rest of the project? Too different?

zipkin-finatra/src/main/scala/com/twitter/zipkin/adapter/JsonAdapter.scala
```diff
@@ -0,0 +1,78 @@
...
+object JsonAdapter extends Adapter {
+  type annotationType = Annotation
+  type annotationTypeType = AnnotationType
+  type binaryAnnotationType = JsonBinaryAnnotation
+  type endpointType = Endpoint
+  type spanType = JsonSpan
+  type traceSummaryType = JsonTraceSummary
+
+  def apply(a: annotationType): Annotation = a
+  //def apply(a: Annotation): annotationType = a
```
johanoskarsson: Can these be removed?

zipkin-finatra/src/main/scala/com/twitter/zipkin/web/App.scala
```diff
@@ -0,0 +1,217 @@
...
+
+  get("/static") { request =>
+    render(path = "static.mustache", exports = new StaticObject)
+  }
+
+  get("/api/query") { request =>
+    /* Get trace ids */
+    val traceIds = QueryRequest(request) match {
+      case r: SpanQueryRequest => {
+        log.debug(r.toString)
```
johanoskarsson: Move out the logging to before the match?

zipkin-finatra/src/main/scala/com/twitter/zipkin/web/App.scala
```diff
@@ -0,0 +1,217 @@
...
+import com.twitter.util.Future
+import com.twitter.zipkin.adapter.{JsonQueryAdapter, JsonAdapter, ThriftQueryAdapter, ThriftAdapter}
+import com.twitter.zipkin.gen
+import com.twitter.zipkin.config.ZipkinWebConfig
+import java.nio.ByteBuffer
+import java.text.SimpleDateFormat
+import java.util.Calendar
+import org.jboss.netty.handler.codec.http.HttpResponse
+
+class App(config: ZipkinWebConfig, client: gen.ZipkinQuery.FinagledClient) extends FinatraApp {
```
johanoskarsson, line 35: I'd suggest adding some comments to this class. Maybe even describe the use and purpose of each endpoint?

zipkin-finatra/src/main/scala/com/twitter/zipkin/web/ZipkinWeb.scala
```diff
@@ -0,0 +1,89 @@
...
+  var server: Option[Server] = None
+
+  def start() {
+    val clientBuilder = ClientBuilder()
+      .codec(ThriftClientFramedCodec())
+      .hostConnectionLimit(1)
+
+    val clientService = config.zkClient match {
+      case None => {
+        clientBuilder.hosts(new InetSocketAddress("localhost", 3002))
```
johanoskarsson: Shouldn't this be in a config and not hardcoded?

zipkin-finatra/src/main/scala/com/twitter/zipkin/web/QueryRequest.scala
```diff
@@ -0,0 +1,63 @@
+package com.twitter.zipkin.web
+
+import com.twitter.zipkin.gen
+import java.text.SimpleDateFormat
+import com.twitter.util.Time
+import com.posterous.finatra.Request
+
+object QueryRequest {
```
johanoskarsson, line 8: Would be nice with a few comments in here. Maybe describing the type of requests or just what happens in the apply method

franklinhu, line 8: Fixed in 1dd303552a6664a5867eba23ef13369bd9a0a266

zipkin-finatra/src/main/scala/com/twitter/zipkin/web/ZipkinWeb.scala
franklinhu: Fixed in    56a6f08

zipkin-finatra/src/main/scala/com/twitter/zipkin/web/App.scala
franklinhu, line 35: Fixed in    729df15

zipkin-finatra/src/main/scala/com/twitter/zipkin/web/App.scala
franklinhu: Fixed in 729df15

zipkin-finatra/src/main/scala/com/twitter/zipkin/adapter/JsonAdapter.scala
franklinhu: Fixed in    48e2efe

project/Project.scala
franklinhu: Fixed in    6535894

johanoskarsson:
+1

PR #76, opened 2012-07-18
Fix 404 in README (link to zipkin.thrift)

shawnps:
Under the "Zipkin query daemon" there is a link to: 

https://github.com/twitter/zipkin/blob/master/zipkin-thrift/src/main/thrift/zipkin.thrift

but I'm guessing this should now be:

https://github.com/twitter/zipkin/blob/master/zipkin-thrift/src/main/thrift/zipkinQuery.thrift


johanoskarsson:
Thanks for this, we'll merge it shortly. (turns out there's a bug in our merging script :)

franklinhu:
Merged thanks!

PR #77, opened 2012-07-19
Readding counts by category.

johanoskarsson:
Useful for finding out when to turn of b3 but also for figuring out the hdfs issue


franklinhu:
+1

PR #80, opened 2012-07-23
Misc finatra fixes

franklinhu:
- Fix span details modal on static trace page
- Remove unnecessary code
- Sort span annotations by timestamp when dumping to json


johanoskarsson:
+1

PR #81, opened 2012-07-23
Extract zipkin version string

franklinhu:


franklinhu:
Woops this is unnecessary

PR #85, opened 2012-07-25
Replace the word submit on the lookup page with something more relevant

johanoskarsson:
Submit doesn't really say what it does. I settled on "Find traces" but I'm open to suggestions. Look up traces? Just "Look up"? Don't want to call it Search since we don't really search as much as look things up in a very strict manner.


franklinhu:
+1, maybe modify the commit msg when you actually merge it though

PR #86, opened 2012-07-30
Debugflag

johanoskarsson:
If the debug flag is set on a span we ensure it is stored. This allows developers to force tracing on a request


zipkin-server/src/main/scala/com/twitter/zipkin/collector/processor/SamplerProcessorFilter.scala
```diff
@@ -29,7 +29,11 @@ class SamplerProcessorFilter(sampler: GlobalSampler) extends ProcessorFilter[Seq
     spans.flatMap { span =>
       span.serviceNames.foreach { name => Stats.incr("received_" + name) }
 
-      if (sampler(span.traceId)) {
+      /**
+       * If the span was created with debug mode on we guarantee that it will be
+       * stored no matter what our sampler tells us
+       */
+      if (span.debug || sampler(span.traceId)) {
```
franklinhu: add some stats here so we can see how much debug traffic we get

franklinhu:
+1

PR #89, opened 2012-07-31
Common refactor

franklinhu:
- Move `Trace`, `TraceSummary` from `common` package to `query` since they are only used in the query service.
- Cleanup from the move


johanoskarsson:
+1

PR #91, opened 2012-08-01
Trace: merge spans and sort by timestamp by default

franklinhu:


johanoskarsson:
Do we always want the spans to be merged and sorted? As long as there's no breakage and noticeable overhead I'm +1 on the change.

franklinhu:
I think any `Trace` that gets returned in a Query call will be merged and sorted anyway for easier consumption so there shouldn't be an extra performance hit.

johanoskarsson:
Ok, sounds good to me. +1

PR #97, opened 2012-08-02
Grep by annotations value. Useful for finding particular trace ids

johanoskarsson:


zipkin-hadoop/src/main/scala/com/twitter/zipkin/hadoop/GrepByAnnotation.scala
```diff
@@ -0,0 +1,20 @@
+package com.twitter.zipkin.hadoop
```
franklinhu, line 16: license header

franklinhu:
+1

PR #98, opened 2012-08-02
Remove Trace dependency from Storage

franklinhu:
Rather than returning a Trace from Storage, return a sequence of Spans and let the QueryService deal


johanoskarsson:
+1

PR #100, opened 2012-08-03
Added getTracesExist support

jerryli9876:


zipkin-server/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraStorage.scala
```diff
@@ -89,6 +89,24 @@ trait CassandraStorage extends Storage with Cassandra {
     }
   }
 
+  def getTracesExist(traceIds: Seq[Long]): Future[Seq[Boolean]] = {
+    CASSANDRA_GET_TRACE.incr
```
franklinhu: you should be incrementing a different counter

zipkin-server/src/main/scala/com/twitter/zipkin/query/QueryService.scala
```diff
@@ -144,6 +144,15 @@ class QueryService(storage: Storage, index: Index, aggregates: Aggregates, adjus
     }
   }
 
+  def getTracesExist(traceIds: Seq[Long]): Future[Seq[Boolean]] = {
```
franklinhu: this name is kinda unintuitive. something like traceExists or tracesExist would be better

zipkin-server/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraStorage.scala
```diff
@@ -89,6 +89,24 @@ trait CassandraStorage extends Storage with Cassandra {
...
+    Future.collect {
+      traceIds.grouped(storageConfig.traceFetchBatchSize).toSeq.map { ids =>
+        traces.multigetRows(ids.toSet.asJava, None, None, Order.Normal, 1).map { rowSet =>
+          ids.flatMap { id =>
+            val spans = rowSet.asScala(id).asScala.map {
+              case (colName, col) => ThriftAdapter(col.value)
+            }
+            Some(!spans.isEmpty)
```
franklinhu: Since all the values are going to be `Some(..)`, you're better of just using `.map` rather than `.flatmap` and returning the boolean in the closure rather than the `Option`. It reduces some overhead

zipkin-thrift/src/main/thrift/zipkinQuery.thrift
```diff
@@ -120,6 +120,8 @@ service ZipkinQuery {
 
     //************** Fetch traces from id **************
 
+    list<bool> getTracesExist(1: list<i64> trace_ids) throws (1: QueryException qe);
```
franklinhu: It may be a bit cleaner if we return a `map<i64, bool>` rather than just a list, although not necessary

zipkin-server/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraStorage.scala
```diff
@@ -90,6 +91,35 @@ trait CassandraStorage extends Storage with Cassandra {
...
+        traces.multigetRows(ids.toSet.asJava, None, None, Order.Normal, 1).map { rowSet =>
+          var traceSet = new HashSet[Long]()
+          ids.foreach { id =>
+            val spans = rowSet.asScala(id).asScala.map {
+              case (colName, col) => ThriftAdapter(col.value)
+            }
+            if (!spans.isEmpty) {
+              traceSet += spans.head.traceId
+            }
+          }
```
franklinhu: You can avoid using the HashSet with something like

```
  ids.flatMap {
    // return None if we don't want to include the id
    // return Some(id) if we do
  }.toSet
```

zipkin-server/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraStorage.scala
```diff
@@ -90,6 +91,35 @@ trait CassandraStorage extends Storage with Cassandra {
...
 
   /**
+   * Finds traces that have been stored from a list of trace IDs
+   *
+   * @param traceIds a List of trace IDs
+   * @return a Set of those trace IDs from the list which are stored
+   */
+
+  def tracesExist(traceIds: Seq[Long]): Future[Set[Long]] = {
+//    CASSANDRA_GET_TRACE.incr
```
franklinhu: Add a different counter for this, so we can see how many times it's called

zipkin-server/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraStorage.scala
```diff
@@ -25,6 +25,7 @@ import com.twitter.conversions.time._
 import scala.collection.JavaConverters._
 import com.twitter.zipkin.config.{CassandraConfig, CassandraStorageConfig}
 import com.twitter.zipkin.adapter.ThriftAdapter
+import collection.immutable.{HashSet, HashMap}
```
franklinhu: Shouldn't need this import

zipkin-server/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraStorage.scala
```diff
@@ -90,6 +94,31 @@ trait CassandraStorage extends Storage with Cassandra {
...
+  def tracesExist(traceIds: Seq[Long]): Future[Set[Long]] = {
+    CASSANDRA_TRACE_EXISTS.incr
+    Future.collect {
+      traceIds.grouped(storageConfig.traceFetchBatchSize).toSeq.map { ids =>
+        traces.multigetRows(ids.toSet.asJava, None, None, Order.Normal, 1).map { rowSet =>
+          ids.flatMap { id =>
+            val spans = rowSet.asScala(id).asScala.map {
+              case (colName, col) => ThriftAdapter(col.value)
+            }
+            if (spans.isEmpty) None else Some(spans.head.traceId)
```
franklinhu: Should put if/else on separate lines for better readability, with braces

franklinhu:
Just two small nits, be sure to update the pull request title and description. +1 other than that

PR #101, opened 2012-08-06
Trace spec refactor

franklinhu:
- Move Thrift conversion spec to ThriftQueryAdapter
- Replace thrift/scrooge dependencies with comparable common dependencies
- Move TraceSpec to zipkin-common


johanoskarsson:
+1

PR #102, opened 2012-08-07
Preliminary Version of formatting emails using Mustache

jerryli9876:
Made all the line processing simpler by using only lists

Removed standardization stuff for now in ServiceNameList

using mustachejava to format all email service reports


zipkin-hadoop-job-runner/src/main/scala/com/twitter/zipkin/hadoop/HadoopJobClient.scala
```diff
@@ -27,29 +28,12 @@ import java.util.Scanner
...
-   */
-  def getKeyValue(lineScanner: Scanner): String
-
   /**
    * Process a key and its value
    * @param s the key passed
    * @param value values associated with the key
    */
-  def processKey(s: String, value: List[String])
+  def processKey(s: String, value: List[List[String]])
```
franklinhu: It seems to make more sense to have `processKey` still take a `List[String]` and have the caller do the flattening

zipkin-hadoop-job-runner/src/main/scala/com/twitter/zipkin/hadoop/WriteToFileClient.scala
```diff
@@ -58,6 +44,8 @@ object WriteToFileClient {
 
   protected var pws : HashMap[String, PrintWriter] = new HashMap[String, PrintWriter]()
 
+  def toHtmlName(s: String) = s + ".html"
```
franklinhu: This is okay for now, but we may want to consider using a templating engine like mustache for this kind of stuff. Concatenating HTML strings is kinda janky

https://github.com/spullara/mustache.java/

zipkin-hadoop-job-runner/src/main/scala/com/twitter/zipkin/hadoop/HadoopJobClient.scala
```diff
@@ -58,33 +42,79 @@ abstract class HadoopJobClient(val combineSimilarNames: Boolean) {
...
-      println(service + ", " + values)
       processKey(service, values)
     }
   }
+}
+
+object HadoopJobClient {
+
+  val DELIMITER = ":"
+  var serviceNames = new HashMap[String, String]()
```
franklinhu, line 82: Any reason why you moved this into the singleton rather than in an the class instance? Keeping state in singletons can be scary

jerryli9876, line 82: The idea was that we'd populate this once at the beginning of postprocessing, then we'd never edit it again. That way we'd save some time (instead of populating it every time we postprocess some job) and we'd also make sure that the standardized names are consistent throughout all the jobs, once we start using standardized names.

zipkin-hadoop-job-runner/src/main/scala/com/twitter/zipkin/hadoop/HadoopJobClient.scala
jerryli9876: I made that change because I noticed that every time you want to process a line, you need to break it down into tokens so you can figure out what is the key and what is the value anyways

zipkin-hadoop-job-runner/src/main/scala/com/twitter/zipkin/hadoop/HadoopJobClient.scala
franklinhu, line 82: Ok that sounds alright. For clarity you should `import scala.collection.mutable` and construct the object with `mutable.HashMap`

zipkin-hadoop-job-runner/src/main/scala/com/twitter/zipkin/hadoop/email/EmailContent.scala
```diff
@@ -0,0 +1,165 @@
...
+  private var oneLineResults = List[OneLineResults]()
+  private var body = List[Body]()
+  private var header = List(new Header(serviceName))
+  private var document = List[Html]()
+
+  def html() = {
+    document.asJava
+  }
+
+  class Html(var header: java.util.List[Header], var body: java.util.List[Body] ) {}
```
franklinhu: you can just make them `case class`s since they don't have a body

zipkin-hadoop-job-runner/src/main/scala/com/twitter/zipkin/hadoop/email/EmailContent.scala
```diff
@@ -0,0 +1,165 @@
...
+    val mustache = mf.compile("email.mustache")
+    mustache.execute(pw, this).flush()
+  }
+
+}
+
+object EmailContent {
+
+  // Ensure that we never make a different EmailContent for the same service
+  private var templates = HashMap[String, EmailContent]()
```
franklinhu: `Map[String, EmailContent]()` should be sufficient

zipkin-hadoop-job-runner/src/main/scala/com/twitter/zipkin/hadoop/sources/Util.scala
```diff
@@ -0,0 +1,64 @@
...
+  def isDataDir(f: File) = {
+    f.getName.charAt(0) != '_'
+  }
+
+  /**
+   * Traverses a directory and applies a function to each file in the directory
+   * @param func a function which takes a File and returns a Unit
+   * @param f a File representing a directory and which we want to apply func on
+   */
+  def traverseFileTree(func: File => Unit, f: File): Unit = {
```
franklinhu: I think in general people like to put any functions that are passed in second in case people want to pass in lambda. You could do something like

`def traverseFileTree(f: File)(func: File => Unit)`

which makes lambdas a bit prettier

franklinhu:
+1

PR #104, opened 2012-08-09
zipkin-finatra fixes

franklinhu:
- Bump to Finatra `0.2.3`
- Fix /api/query response
- Fix timestamp text on query results


johanoskarsson:
+1

PR #105, opened 2012-08-09
Make traces exist gem

jerryli9876:
- Added has-sampled.rb, which uses tracesExist
- Bumped zipkin-query to 0.1.0
- Added tracesExist to thrift files and generated ruby code


franklinhu:
+1

PR #106, opened 2012-08-10
Fixup Scribe script, and add a localhost Finatra config

glynd:
Hi,
Just fixed an issue in the collector.sh (referencing the wrong JAR)
Added a config file for Finatra which uses the localhost zookeeper and doesn't override the queryhost
(This means they match the other config within the repo)

### DIFF
--- a/zipkin-finatra/config/web-localhost.scala
+++ b/zipkin-finatra/config/web-localhost.scala
@@ -1,0 +1,31 @@
+/*
+* Copyright 2012 Twitter Inc.
+*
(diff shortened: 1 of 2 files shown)


franklinhu:
This looks good, but I only see the changes to `collector.sh` in the diff. Did you forget to add the config for Finatra?

glynd:
Hopefully that solves it - still not used to these new fangled source control systems ;)

franklinhu:
+1

franklinhu:
Merged, thanks for your contribution @glynd!

PR #110, opened 2012-08-13
Fixed wrong jar filename in collector.sh

gavri:
Fixed the name of the zipkin-scribe jar file in collector.sh


gavri:
Closed because this https://github.com/twitter/zipkin/pull/106 covers it

PR #112, opened 2012-08-13
Improve local js,css development

franklinhu:
Making changes to js,css can be painful with a packaged jar since a compilation is needed to repackage any new changes. This change allows for the paths to the static resources to be configured at runtime. While developing locally, you can then set the prefix to bypass the jar and the browser will resolve it on your local filesystem.


johanoskarsson:
+1

PR #114, opened 2012-08-15
Update docs for local development

franklinhu:
- Fix previously incorrect documentation on local development for UI

### DIFF
--- a/zipkin-finatra/config/web-dev.scala
+++ b/zipkin-finatra/config/web-dev.scala
@@ -23,10 +23,14 @@ new ZipkinWebConfig {
 
   /**
    * Making changes to js/css can be painful with a packaged jar since a compilation is needed to
-   * repackage any new changes. `resourcePathPrefix` can be hacked to point to the directory
-   * on your local file system so the browser resolves it outside of the jar. Example:
+   * repackage any new changes.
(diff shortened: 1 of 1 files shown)


zipkin-finatra/config/web-dev.scala
```diff
@@ -23,10 +23,14 @@ new ZipkinWebConfig {
 
   /**
    * Making changes to js/css can be painful with a packaged jar since a compilation is needed to
-   * repackage any new changes. `resourcePathPrefix` can be hacked to point to the directory
-   * on your local file system so the browser resolves it outside of the jar. Example:
+   * repackage any new changes.
+   * A simple hack is to stand up a simple Python HTTP server and point `resourcePathPrefix` it.
```
johanoskarsson, line 27: Is there anything we can do to avoid this step? Seems annoying that we'd require a third party http server for just that.

franklinhu, line 27: Talked to Johan IRL about this. The python server is just nice to have if you're doing JS/CSS development since the resources get packaged in the jar and won't be refreshed without a full compile; it's not necessary to run the Scala UI.

johanoskarsson:
+1

PR #117, opened 2012-08-22
Ensure that we can receive Scribe messages with endline at the end

johanoskarsson:
Some codestyle changes and removed a few lines of unused code in my quest to find the right code to test :)


franklinhu:
+1

PR #118, opened 2012-08-23
Backbone (part 1)

franklinhu:
Initial changes to move the front end JS to Backbone
- Span, Service Name models
- Index page service name, span name populated through Backbone models/views
- Some Finatra side changes to accommodate Backbone


zipkin-finatra/src/main/scala/com/twitter/zipkin/web/App.scala
```diff
@@ -70,8 +87,23 @@ class App(config: ZipkinWebConfig, client: gen.ZipkinQuery.FinagledClient) exten
    * - adjust_clock_skew = (true|false), default true
    */
   get("/api/query") { request =>
+    query(request).map(render.json(_))
+  }
+
+  def query(request: Request): Future[Seq[JsonTraceSummary]] = {
+    QueryRequest(request) match {
+      case None => {
```
johanoskarsson: Not important, but you can tighten up a lot of these match clauses by doing them on one line

zipkin-finatra/src/main/scala/com/twitter/zipkin/web/App.scala
```diff
@@ -108,9 +140,17 @@ class App(config: ZipkinWebConfig, client: gen.ZipkinQuery.FinagledClient) exten
    * Returns the total list of services Zipkin is aware of
    */
   get("/api/services") { request =>
-    log.debug("/api/services")
+    log.debug("/api/service")
```
johanoskarsson: Should be services I assume

zipkin-finatra/src/main/scala/com/twitter/zipkin/web/QueryRequest.scala
```diff
@@ -40,32 +48,34 @@ object QueryRequest {
...
-          case None => {
-            request.params.get("annotation_key") match {
-              case Some(key) => {
-                request.params.get("annotation_value") match {
-                  case Some(value) => {
-                    KeyValueAnnotationQueryRequest(serviceName, key, value, endTimestamp, limit, order)
-                  }
-                  case None => {
-                    ServiceQueryRequest(serviceName, endTimestamp, limit, order)
+    Some {
```
johanoskarsson: This is getting kinda indentation heavy.  One thing would be to to val blah = request.params.... and then return the Some(blah) version. Perhaps break the time_annotation part into a private method? I dunno.

johanoskarsson:
Couple of nits that you may or may not wish to fix. +1 otherwise. Note that as usual I have no idea what is going on in the javascript, so I will assume it works well.

PR #121, opened 2012-08-23
Fix NPE when querying for time annotations

franklinhu:


johanoskarsson:
Is null unsupported there? Is that a regression or has it always been?

franklinhu:
I don't think I ever tested it in the Finatra app. Pretty sure Rails massaged the `nil` into a valid thrift argument.

johanoskarsson:
+1

PR #122, opened 2012-08-23
Backbone (part 2)

franklinhu:
- Remove some Rails-isms in JS files
- Backbone-ify index querying


johanoskarsson:
Rubberstamp +1 due to lack of JS knowledge :)

caniszczyk:
On just a random side note... have you look at AngularJS vs. Backbone?

Too late now but something for future reference: http://www.angularjs.org/

franklinhu:
Hadn't actually seen that. Hm will keep that in mind for the future

PR #125, opened 2012-08-24
Add dependency endpoints

jerryli9876:
Made the page for the dependency tree sans any AJAX or connecting to the server. Since I learned JavaScript last week I reckon a lot of this will be crap.


mosesn:
In the future, you might want to make many, smaller pull requests instead of just one massive one.  Also, you probably don't want to put all of that commented out stuff in, if you just want to be able to use it later, obliterate it, because it's still in git.  Source control is magic :sparkles:

jerryli9876:
Yeah I apologize for the massive pull request; I didn't even realize that I hadn't made so many commits.

franklinhu:
Yeah it might be better to split this into separate pull requests
- backend api changes (collector and query)
- hadoop changes
- web/UI changes

PR #126, opened 2012-08-27
Return IndexTraceId from storage.Index

franklinhu:
- Change `storage.Index` to return a `Seq[IndexTraceId]` rather than a vanilla list of trace IDs. Each `IndexTraceId` contains a `traceId` and its associated `timestamp`.


johanoskarsson:
+1

PR #128, opened 2012-08-28
Dependency graph endpoints 2

jerryli9876:
Part 1 redux of several pull requests that will make up the dependency graph.

Fixed some minor issues with the hadoop scripts, and with email.mustache.
Added dependency graph endpoints on the server side.

There are a few files such as QueryService and Index that changed that I didn't touch...not sure what's going on there.


zipkin-server/src/main/scala/com/twitter/zipkin/storage/Index.scala
```diff
@@ -48,7 +47,7 @@ trait Index {
    * Only return maximum of limit trace ids from before the endTs.
    */
   def getTraceIdsByAnnotation(serviceName: String, annotation: String, value: Option[ByteBuffer],
-                              endTs: Long, limit: Int): Future[Seq[IndexedTraceId]]
+                              endTs: Long, limit: Int): Future[Seq[Long]]
```
johanoskarsson: You need to pull in master

zipkin-finatra/src/main/resources/public/css/application.css
```diff
@@ -680,4 +680,4 @@ div#annotation-modal {
   width: 700px;
 }
 
-/* END show.html.erb */
+/* END show.html.erb */
```
johanoskarsson: Nop diff

zipkin-server/src/main/scala/com/twitter/zipkin/query/QueryService.scala
```diff
@@ -324,7 +324,7 @@ class QueryService(storage: Storage, index: Index, aggregates: Aggregates, adjus
    */
   private def getTraceIdDurations(
     traceIds: Future[Seq[Long]]
-  ): Future[Seq[TraceIdDuration]] = {
```
johanoskarsson, line 327: Nop diff

zipkin-server/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraAggregates.scala
```diff
@@ -32,12 +32,20 @@ import com.twitter.cassie.{Column, ColumnFamily}
...
   val topAnnotations: ColumnFamily[String, Long, String]
+  val dependencies: ColumnFamily[String, Long, String]
 
   val Delimiter: String = ":"
 
   /**
    * Get the top annotations for a service name
    */
+  def getDependencies(serviceName: String): Future[Seq[String]] = {
+    getDependencies(serviceName)
```
johanoskarsson: Isn't this calling itself?

zipkin-server/src/schema/cassandra-schema.txt
```diff
@@ -13,4 +13,5 @@ create column family ServiceNameIndex with comparator = LongType;
 create column family AnnotationsIndex with comparator = LongType;
 create column family DurationIndex with comparator = LongType;
 
-create column family TopAnnotations with comparator = LongType;
\ No newline at end of file
+create column family TopAnnotations with comparator = LongType;
```
johanoskarsson, line 21: Let's add a comment here describing each these two column families.

zipkin-finatra/src/main/resources/public/css/application.css
jerryli9876: I honestly can't figure out what the diff here is. The version in my repo is exactly the same as the one in master.

johanoskarsson: Do a git checkout master filename to get that version

zipkin-server/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraAggregates.scala
```diff
@@ -70,6 +80,17 @@ trait CassandraAggregates extends Aggregates with Cassandra {
   }
 
   /** Synchronize these so we don't do concurrent writes from the same box */
+  def storeDependencies(serviceName: String, endpoints: Seq[String]): Future[Unit] = synchronized {
+    val remove = dependencies.removeRow(serviceName)
+    val batch = dependencies.batch()
+    endpoints.zipWithIndex.foreach { case (endpoint: String, index: Int) =>
+      batch.insert(serviceName, new Column[Long, String](index, endpoint))
+    }
+    remove()
```
johanoskarsson, line 89: You shouldn't have to block here, but instead chain the futures so you can return one that executes this first and then the batch

johanoskarsson:
+1 except for the blocking part above.

PR #131, opened 2012-08-28
Dependency graph finatra clean

jerryli9876:
Part 2 of Dependency graph stuff.
- Front end for aggregates view with dependency tree without any AJAX. Nothing for service reports on the aggregates side yet.

Note: My Javascript is likely crap


zipkin-finatra/src/main/resources/public/css/aggregates.css
```diff
@@ -0,0 +1,54 @@
...
+}
+
+.popover-content {
+  max-height: 200px !important;
+  max-width: 1000px;
+  overflow: scroll;
+}
+
+@media (min-width: 1200px) {
+/*  .row {
```
franklinhu: Remove commented out

zipkin-finatra/src/main/resources/public/css/aggregates.css
```diff
@@ -0,0 +1,54 @@
+#global-dependency {
```
franklinhu, line 17: There should probably be a license header here

zipkin-finatra/src/main/resources/public/js/application-aggregates.js
```diff
@@ -0,0 +1,175 @@
...
+        date: self_val,
+        current: self_val,
+        starts: 0,
+        onBeforeShow: function(){
+          self.DatePickerSetDate(self_val, true);
+        },
+        onChange: function(formated, dates){
+          self.val(formated);
+        }
+      }).blur(function(){
```
franklinhu: Ddi you intend to attach a `blur` handler?

zipkin-finatra/src/main/resources/public/js/application-aggregates.js
```diff
@@ -0,0 +1,175 @@
...
+  var templatize = Zipkin.Util.templatize
+    , TEMPLATES = Zipkin.Util.TEMPLATES
+    ;
+
+  var initialize = function () {
+    Zipkin.Base.initialize();
+
+    var href = window.location.href;
+
+    $(".js-zipkin-navbar > li > a").each(function (index, elem) {
```
franklinhu, line 32: Probably doesn't need to be fixed, but I believe jQuery selectors are evaluated in ascending hierarchy, so for something like `.js-zipkin-navbar > li > a` will get all the `a` anchors on the page, select those that have `li` as parents, etc. This can get pretty expensive if you have lots of anchors on the page. In contrast, if you attach a class to the anchor selector `a.some-class` then there will be fewer objects initially selected.

zipkin-finatra/src/main/resources/public/js/application-aggregates.js
```diff
@@ -0,0 +1,175 @@
...
+        onChange: function(formated, dates){
+          self.val(formated);
+        }
+      }).blur(function(){
+      });
+    });
+
+    $(".nav-dropdowns > li > ul > li").each(function(index, elem) {
+      $(elem).on('mouseover', function (event) {
+        $(".nav-dropdowns > li > ul > li").each(function (i, e) {
```
franklinhu: `e` is usually used for events. Should avoid using it for elements. Can use `el` instead

zipkin-finatra/src/main/resources/public/js/application-aggregates.js
```diff
@@ -0,0 +1,175 @@
...
+  var DependencyTree = (function DependencyTree() {
+
+    return {
+      _init: function() {
+
+    var nodes = [{name: "foo"},
+                 {name: "bar"},
+                 {name: "baz"}];
+
+	var links = [{source: 0, target: 1, value: 0.1, count: 0.1},
```
franklinhu: spacing

zipkin-finatra/src/main/resources/public/js/zipkin-global-dependency.js
```diff
@@ -0,0 +1,212 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*global d3:false */
+
+//= require d3-2.9.1
```
franklinhu: You can remove this

zipkin-finatra/src/main/resources/public/js/zipkin-global-dependency.js
```diff
@@ -0,0 +1,212 @@
...
+    d.selected = false;
+    this.currentTarget = null;
+
+    nodeSelector(d.name).popover('hide');
+    this.redraw();
+  };
+
+
+  var getText = function(d) {
+    var s = "<table><thead><tr><th>Service Called</th><th># of calls</th></thead>";
```
franklinhu: This should probably happen in a template

zipkin-finatra/src/main/resources/public/js/zipkin-global-dependency.js
```diff
@@ -0,0 +1,212 @@
...
+      , height = this.height
+      , border = this.border
+      ;
+
+    var formatNumber = d3.format(",.0f"),
+        format = function(d) { return formatNumber(d) + " TWh"; },
+        color = d3.scale.category20();
+
+    var findCircularDependencies = function(links) {
+
```
franklinhu: empty?

zipkin-finatra/src/main/scala/com/twitter/zipkin/config/CssConfig.scala
```diff
@@ -28,4 +28,6 @@ class CssConfig extends StaticResourceConfig {
     "datepicker.css",
     "application.css"
   )
+
+  val localAggregatesResources = localResources ++ Seq("aggregates.css")
```
franklinhu: You can just add this to localResources. It's not a big deal if we serve the CSS and JS on all the pages

zipkin-finatra/src/main/scala/com/twitter/zipkin/web/App.scala
```diff
@@ -312,6 +316,15 @@ class App(config: ZipkinWebConfig, client: gen.ZipkinQuery.FinagledClient) exten
...
 
+  private def wrapAggregatesView(v: View) = new View {
+    val template = "templates/layouts/application.mustache"
+    val rootUrl = config.rootUrl
+    val innerView: View = v
+    val javascripts = config.jsConfig.aggregatesResources
+    val stylesheets = config.cssConfig.aggregatesResources
+    lazy val body = innerView.render
+  }
+
```
franklinhu: If you add the JS and CSS to `localResources` you won't need this function

zipkin-finatra/src/main/scala/com/twitter/zipkin/config/CssConfig.scala
jerryli9876: The problem is that I need to set width of the container to some larger number than usual (otherwise the dependency graph looks horrible) and if I include it in the localResources the width of the other pages is also set to something large.

zipkin-finatra/src/main/resources/public/js/application-aggregates.js
jerryli9876, line 32: That's good to know. For this page, both are equivalent though, at least, I believe.

zipkin-finatra/src/main/resources/public/css/aggregates.css
franklinhu, line 17: Uhh any reason this isn't the full license header?

zipkin-finatra/src/main/scala/com/twitter/zipkin/config/CssConfig.scala
franklinhu: Why not just add a CSS class to the container on the aggregates page?

zipkin-finatra/src/main/resources/templates/aggregates.mustache
```diff
@@ -0,0 +1,62 @@
...
+          </a>
+          <ul class="dropdown-menu" role="menu" aria-labelledby="drop1">
+            <li><a tabindex="-1" id="dropdown-by-count" class="active">By Number of Calls</a></li>
+            <li><a tabindex="-1" id="dropdown-by-duration">By Duration</a></li>
+          </ul>
+        </li>
+      </ul>
+    </div>
+    <div class="clearfix pull-right">
+      Aggregates For <input class="span2 date-input" id="date" name="date" type="text" value="08-24-2012">
```
franklinhu: Is this date value supposed to be hardcoded?

zipkin-finatra/src/main/resources/templates/aggregates.mustache
```diff
@@ -0,0 +1,62 @@
...
+    </div>
+  </div>
+</div>
+
+<div class="span5 offset3" style="vertical-align: 50%;" id="loading-data">
```
franklinhu: Ideally we want all style in the stylesheets

zipkin-finatra/src/main/resources/public/js/application-aggregates.js
```diff
@@ -0,0 +1,174 @@
...
+          var prevWidth = $(window).width();
+          $(window).resize(function () {
+            var newWidth = $(window).width();
+            if (Zipkin.Base.windowResized(prevWidth, newWidth)) {
+              var w = newWidth >= Zipkin.Config.MAX_AGG_WINDOW_SIZE ? Zipkin.Config.MAX_AGG_GRAPHIC_WIDTH : Zipkin.Config.MIN_AGG_GRAPHIC_WIDTH;
+              globalDependencies.resize(w + 200);
+              prevWidth = newWidth;
+            }
+           });
+        })();
```
franklinhu: This probably doesn't need to be wrapped in an anonymous function

zipkin-finatra/src/main/resources/public/js/application.js
```diff
@@ -120,6 +120,27 @@ Zipkin.Base = (function() {
...
+
+        $(".js-zipkin-navbar > li > a").each(function (i, e) {
+          if (e.hash != hash) {
+            $(e.hash).hide();
+          }
+        });
+        $(hash).show();
+        $("html body").animate({scrollTop: 0}, "slow");
+      });
+    });
```
franklinhu, line 129: Instead of iterating through the pills and binding a click handler to each you can just bind it via the selector

``` javascript
$(".js-zipkin-navbar > li").click(function(elem) {...});
```

zipkin-finatra/src/main/resources/public/js/application-aggregates.js
jerryli9876: What's the preferred format?

zipkin-finatra/src/main/resources/public/js/application-aggregates.js
```diff
@@ -0,0 +1,175 @@
...
+    $(".container").addClass("container-aggregates");
+
+    var href = window.location.href;
+
+    $(".js-zipkin-navbar > li > a").each(function (index, elem) {
+      var parent = $(elem).parent();
+      if ($(elem).attr("href") == href ) {
+        parent.addClass("active");
+      } else if (parent.hasClass("active")) {
+        parent.removeClass("active");
```
franklinhu, line 37: Can just do 

``` javascript
} else {
  parent.removeClass("active");
}
```

zipkin-finatra/src/main/resources/public/js/application-aggregates.js
franklinhu: It looks like all of this is nested inside `_init` but it's at a lower indentation level.

zipkin-finatra/src/main/resources/public/js/application.js
```diff
@@ -120,6 +120,16 @@ Zipkin.Base = (function() {
     // Bind click handler for brand button
     $(".brand").click(brandClick);
 
+    // Hook up trace nav buttons
+    $(".js-zipkin-navbar > li").click(function () {
+        if (!$(this).hasClass("active")) {
+          $(".js-zipkin-navbar > li").each(function (i, el) {
+            $(el).removeClass("active");
+          });
+          $(this).addClass("active");
```
franklinhu, line 127: It's probably simpler if you just use the selector...

``` javascript
$(".js-zipkin-navbar > li.active").removeClass("active");
```

zipkin-finatra/src/main/resources/public/js/zipkin-global-dependency.js
```diff
@@ -0,0 +1,224 @@
...
+      var node = svg.append("g").selectAll(".node")
+          .data(this.nodes)
+        .enter().append("g")
+          .attr("id", function(d) { return "node-id-" + d.name; })
+          .attr("class", "node")
+          .attr("transform", function(d) { return "translate(" + d.x + "," + d.y + ")"; })
+          .on("click", Zipkin.Util.bind(this, hoverEvent))
+          .attr("rel", "popover")
+          .attr("data-original-title", function(d) { return d.name; })
+//          .attr("data-content", function(d) { return getText(d); })
```
franklinhu: remove?

zipkin-finatra/src/main/resources/public/js/zipkin-global-dependency.js
```diff
@@ -0,0 +1,224 @@
...
+      function dragmove(d) {
+        d3.select(this).attr("transform", "translate(" + d.x + "," + (d.y = Math.max(0, Math.min(height - d.dy, d3.event.y))) + ")");
+        sankey.relayout();
+        link.attr("d", path);
+      }
+    return svg;
+  };
+
+  return GlobalDependencies;
+})(Zipkin);
```
franklinhu: You don't actually use this passed in variable so you can omit it. It's global anyway

zipkin-finatra/src/main/resources/templates/aggregates.mustache
```diff
@@ -0,0 +1,62 @@
...
+        <li class="active"><a id="dependencies-pill" href="#dependency">Dependencies</a></li>
+        <li><a id="service-report-pill" href="#service-report">Service Reports</a></li>
+      </ul>
+    </div>
+    <div class="trace-nav pull-right">
+      <ul class="nav nav-dropdowns" role="navigation">
+        <li class="dropdown">
+          <a id="drop1" href="#" role="button" class="dropdown-toggle" data-toggle="dropdown">Metric Type<b class="caret"></b>
+          </a>
+          <ul class="dropdown-menu" role="menu" aria-labelledby="drop1">
```
franklinhu: `aria-labelled-by`?

zipkin-finatra/src/main/resources/public/js/application-aggregates.js
```diff
@@ -0,0 +1,175 @@
...
+    });
+
+    $(".nav-dropdowns > li > ul > li").each(function(index, elem) {
+      $(elem).on('mouseover', function (event) {
+        $(".nav-dropdowns > li > ul > li").each(function (i, el) {
+          $(el).removeClass("active");
+        });
+      }).on('click', function(event) {
+        $(elem).addClass("active");
+        s = $("#global-dependency > div");
```
franklinhu, line 88: Is it defined somewhere else? If not, it should be `var s` so we don't pollute the global scope.

franklinhu:
One nit, +1 other than that.

PR #132, opened 2012-08-28
Add complex query support in QueryService 

franklinhu:
- Add endpoint `getTraceIds` that allows for queries that specify multiple annotations or binary annotations that must be satisfied. Expected result is the intersection of all traces that satisfy the conditions. This endpoint is intended to eventually replace the `getTraceIdBy*` endpoints.


franklinhu:
Ugh wrong head ref

PR #133, opened 2012-08-28
Add complex query support in QueryService

franklinhu:
- Add endpoint `getTraceIds` that allows for queries that specify multiple annotations or binary annotations that must be satisfied. Expected result is the intersection of all traces that satisfy the conditions. This endpoint is intended to eventually replace the `getTraceIdBy*` endpoints.

### DIFF
--- a/zipkin-server/src/main/scala/com/twitter/zipkin/query/QueryService.scala
+++ b/zipkin-server/src/main/scala/com/twitter/zipkin/query/QueryService.scala
@@ -118,8 +118,10 @@ class QueryService(storage: Storage, index: Index, aggregates: Aggregates, adjus
 
       sliceQueries match {
         case Nil => {
-          /* No queries: return a default empty response */
-          Future.exception(gen.QueryException("Invalid query"))
+          /* No queries: get service level traces */
+          index.getTraceIdsByName(serviceName, None, endTs, limit).map {
+            constructQueryResponse(_, limit, order)
+          }.flatten
         }
         case head :: Nil => {
           /* One query: just run it */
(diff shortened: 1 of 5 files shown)


franklinhu:
For complex queries (ones that require more than one index lookup), we do some fancy stuff:
- Get one column (timestamp) from each slice and find the minimum (adjusted timestamp). Because we're computing a set intersection , it's impossible to have valid trace IDs that occur after the latest value for a particular slice. 
- Run the actual queries that fetch the appropriate number of columns with the adjusted timestamp (padded by a minute to take care of write delays, time skew).

zipkin-thrift/src/main/thrift/zipkinQuery.thrift
```diff
@@ -81,10 +81,34 @@ enum Order { TIMESTAMP_DESC, TIMESTAMP_ASC, DURATION_ASC, DURATION_DESC, NONE }
  */
 enum Adjust { NOTHING, TIME_SKEW }
 
+struct QueryAnnotation {
```
johanoskarsson: Is this for some future use? Vs just using the string straight in the QueryRequest

zipkin-server/src/main/scala/com/twitter/zipkin/query/QueryService.scala
```diff
@@ -73,6 +73,124 @@ class QueryService(storage: Storage, index: Index, aggregates: Aggregates, adjus
...
+            queries.map {
+              _.execute(index)
+            }
+          }.map {
+            _.flatten.map {
+              _.timestamp
+            }.min
+          }.map { alignedTimestamp =>
+            /* Pad the aligned timestamp by a minute */
+            val ts = alignedTimestamp + 1.minute.inMicroseconds
```
johanoskarsson: Move this out into a constant (or perhaps config?) somewhere with a description as to why we do it

zipkin-thrift/src/main/thrift/zipkinQuery.thrift
franklinhu: Fixed in b1a5057

johanoskarsson:
+1

PR #137, opened 2012-08-31
Add UI complex query support

franklinhu:
Add feature to get traces based on an intersection of multiple annotations and/or key-value annotations. For example, traces that have both `url=/my/url.json` and `finagle.retry` annotations.

Changes:
- Add common classes QueryRequest, QueryResponse to mirror previously added thrift structs

Breaking changes:
- Some query form input field names have changed:
  - `service_name` to `serviceName`
  - `span_name` to `spanName`
  - `end_datetime` to `endDatetime`
  - `annotations` is now sent as an array of the form `annotations[0]`, `annotations[1]`, etc.
  - `annotation_key` and `annotation_value` are similarly sent as `keyValueAnnotations[0][key]`, `keyValueAnnotations[0][val]`, etc.


johanoskarsson:
Looks good, but should be possible to squeeze in a test or two

johanoskarsson:
+1

PR #140, opened 2012-09-05
Revert to /traces/:id route for backward compatibility

franklinhu:


johanoskarsson:
Do we want both? Either way, I'm +1

franklinhu:
I just changed it back to /traces/ since that's what's currently in place in the ruby version

PR #143, opened 2012-09-06
Chain remove/add aggregates data futures

### LINKED ISSUE #129: CassandraAggregates Futures
As Johan mentions, you shouldn't have to block on the remove in the storage methods, but instead chain the futures so you can return one that executes this first and then the batch.

franklinhu:
Fixes #129


johanoskarsson:
+1

PR #144, opened 2012-09-07
Hadoop cleanup (part 1)

franklinhu:
- Add `TimeGranularity.Day` and `TimeGranularity.Hour` types 
- Move time granularity to be specified at the job level as an argument rather than in the type, eg. `Source(TimeGranularity.Day)` rather than `DailySource`


johanoskarsson:
+1

PR #147, opened 2012-09-18
Collector service/filter refactor

franklinhu:
- Refactor the collector to use Finagle `Service` and `Filter` instead of `Processor` and `ProcessorFilter`


zipkin-server/src/main/scala/com/twitter/zipkin/collector/processor/IndexService.scala
```diff
@@ -39,5 +41,11 @@ class IndexProcessor(index: Index, indexFilter: IndexingFilter) extends Processo
     }
   }
 
-  def shutdown() = index.close()
+  protected def failureHandler(method: String): (Throwable) => Unit = {
+    case t: TooManyWaitersException =>
```
johanoskarsson: Is this reported as a metric elsewhere? Seems like something we'd like to know.

johanoskarsson:
+1. Make sure you canary this before full rollout since it touches a lot of core code

PR #148, opened 2012-09-18
Refactor index filter to be composable

franklinhu:
- Rename `DefaultClientIndexingFilter` to `ClientIndexFilter`
- Refactor `ClientIndexFilter` to a Finagle Filter so it's composable with Finagle Services


zipkin-server/src/main/scala/com/twitter/zipkin/collector/filter/ClientIndexFilter.scala
```diff
@@ -16,9 +16,18 @@
  */
 package com.twitter.zipkin.collector.filter
 
+import com.twitter.finagle.{Service, Filter}
+import com.twitter.util.Future
 import com.twitter.zipkin.common.Span
 
-class DefaultClientIndexingFilter extends IndexingFilter{
+class ClientIndexFilter extends Filter[Span, Unit, Span, Unit] {
```
johanoskarsson, line 28: Can you add a comment here describing what the class does?

johanoskarsson:
+1. See nit above.

Issue #152, opened 2012-09-19
Cleanup modules


franklinhu:
Currently we have a confusing hierarchy of modules for the collector and query services including:
- `zipkin-common`: common classes shared between the two
- `zipkin-thrift`: Thrift structs and interfaces for both services
- `zipkin-scrooge`: module to generate Scala from Thrift via Scrooge
- `zipkin-scribe`: scribe specific classes for collector, including the service
- `zipkin-server`: combination of code for collector and query, including the query service

I think we can clean this up by breaking up the latter two modules and make things more composable.
- `zipkin-cassie`: implementation of `Aggregates`, `Index`, and `Storage` interfaces that use Cassandra as the backing store
- `zipkin-collector-core`: core code for the collector service
- `zipkin-collector-scribe`: scribe specific classes
- `zipkin-collector-service`: module to compose everything together
- `zipkin-query-core`: core code for query service
- `zipkin-query-service`: module to compose everything together

mosesn:
Where will the Aggregates, Index, and Storage interfaces live?  Core?

johanoskarsson:
Overall a good idea. Some feedback:
- I'd change zipkin-cassie to zipkin-cassandra so as not to tie it to the client used.
- We might get away without the -core modules for now, we will get tons of modules with this change.
- Make sure you clearly write up which module does what in the readme (or maybe even a readme per module)

franklinhu:
@mosesn I've already moved them to zipkin-common so they can be shared across all the modules

PR #154, opened 2012-09-20
Collector refactor

franklinhu:
Move collector specific code to `zipkin-collector-core`

This is a part of #152


johanoskarsson:
+1

PR #155, opened 2012-09-20
Fix zipkin-cassandra testing

franklinhu:
- Move `zipkin-cassandra` dependency from `zipkin-collector-core` to `zipkin-scribe`
- Move cassandra test resources to `zipkin-cassandra`
- Make cassie dependency transitive


johanoskarsson:
+1

PR #159, opened 2012-09-21
zipkin-query refactor

franklinhu:
- Rename `zipkin-server` to `zipkin-query-core`
- Pull service specific code to `zipkin-query-service`


project/Project.scala
```diff
@@ -261,17 +261,30 @@ object Zipkin extends Build {
...
 
-      /* Add configs to resource path for ConfigSpec */
-      unmanagedResourceDirectories in Test <<= baseDirectory {
-        base =>
-          (base / "config" +++ base / "src" / "test" / "resources").get
-      }
-    ).dependsOn(common, scrooge, cassandra)
+    libraryDependencies ++= testDependencies,
+
+    PackageDist.packageDistZipName := "zipkin-server.zip",
```
johanoskarsson: This should change too I assume?

franklinhu: Good catch. Fixed in e515858

johanoskarsson:
+1

PR #161, opened 2012-09-22
Default sample rate to 1

franklinhu:
Setting the default to 100% makes it one less thing new users have to worry about when spinning up a collector.


johanoskarsson:
Hmm, I'm not sure this is the right place to change the default sample rate. Would it not be better to educate new users on how they can use the sampler that lets everything through? Or even better change that for them in the dev config?
With this change I'm worried something might go wrong and it'll fall back to the default, blowing up our cluster.

franklinhu:
That makes sense. Closing this PR and will figure out a cleaner way to do this.

PR #162, opened 2012-09-22
Slim down README.md

franklinhu:
Since we have all the content at twitter.github.com/zipkin, it makes sense to slim down the README.md for easier consumption


johanoskarsson:
In the readme we should provide links to the rest of the documentation so that the user doesn't get stuck after reading the getting started parts.

franklinhu:
I added a 'Full documentation' section in the readme, and also moved some of the sections from doc/overview.md since I feel like they belong on the main readme.

johanoskarsson:
+1

PR #163, opened 2012-09-24
Cleanup Future[Void] -> Future[Unit] conversion

franklinhu:
Use the `.unit` method that transforms a `Future[_]` to `Future[Unit]`


johanoskarsson:
+1

PR #166, opened 2012-09-25
hadoop: use UtcDateRangeJob

franklinhu:
- Use `UtcDateRangeJob` rather than `DefaultDateRangeJob` (TZ=Pacific)
- Fix `run_jobs.rb` file header


johanoskarsson:
+1

PR #167, opened 2012-09-26
hadoop: Replace TimeGranularity case objects with fields

franklinhu:
Hadoop can't seem to create case objects through reflection, so use
a val in TimeGranularity obj instead


johanoskarsson:
+1

PR #168, opened 2012-09-27
Change StatsService to ServiceStatsFilter

franklinhu:
It makes more sense for the stats to be a Filter rather than a Service


johanoskarsson:
+1

PR #169, opened 2012-09-28
Replace ThriftAdapter with implicit conversions

franklinhu:
- Cleanup conversion between common and Thrift with `.toClassName` calls rather than explicit `ThriftAdapter(_)`


johanoskarsson:
+1

PR #170, opened 2012-10-02
finatra: Fix JSON Span serialization

franklinhu:
JsonAdapter incorrectly assumed that a Span has at least one annotation. 
- Let start time of the span be -1 if there are no annotations
- Let the duration of the span be 0 if there are no annotations


johanoskarsson:
Wouldn't it be better to leave out timestamps completely out of the json if we don't have a value for it?

mosesn:
I think jerkson treats options that way automagically.

Edit: yeah, you probably want to leave them as options rather than getOrElse-ing them.

franklinhu:
Yeah that was kind of silly. It's now updated to just shove the Option into the JsonSpan, so jerkson should take care of it.

johanoskarsson:
+1

PR #171, opened 2012-10-02
Make conversions toThrift, toClass methods into lazy vals

franklinhu:
Since the objects passed in are immutable, we can make the
.toThrift and .toClassName methods into lazy vals to avoid
additional overhead from repeated calls


johanoskarsson:
+1

PR #174, opened 2012-10-08
collector: add kafka store

franklinhu:
Add module that allows users to write their spans from the collector to a Kafka queue

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,21 @@
+Zipkin is a distributed tracing system.
+Copyright 2012 Twitter, Inc.
+
+Zipkin has a number of optional third party dependencies:
+
+Cassandra
+Apache Public License 2.0
+http://cassandra.apache.org/
+
+Kafka 0.7.0
+Apache Public License 2.0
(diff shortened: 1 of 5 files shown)


caniszczyk:
Looks like Kafka is being moved to the ASF incubator which is good:
https://github.com/apache/kafka

What sucks is they have no releases yet there.

We should probably start adding a NOTICE (e.g., https://github.com/twitter/scalding/blob/develop/NOTICE) file to track these third party dependencies...

caniszczyk:
Apparently there is a 0.7.1 release... not available on any maven repo that I can find...

http://incubator.apache.org/kafka/downloads.html

franklinhu:
They didn't seem to have an official Scala 2.9.\* build, and the 2.8 jar is ABI incompatible with 2.9. I just used one that @jasonjckn published on clojars. I'll add the NOTICE file.

johanoskarsson:
+1. Perhaps add a readme describing the module but also mentioning that it is experimental as per your email.

PR #176, opened 2012-10-11
respects limit

mosesn:
## motivation

Can't look up traces by span.
## fix

respects the "limit" field.
## testing

changed the spec to expect proper "limit" field.  I don't know why it didn't in the first place.  Seems like the test was incorrect too.

### DIFF
--- a/zipkin-query-core/src/main/scala/com/twitter/zipkin/query/QueryService.scala
+++ b/zipkin-query-core/src/main/scala/com/twitter/zipkin/query/QueryService.scala
@@ -98,16 +98,16 @@ class QueryService(storage: Storage, index: Index, aggregates: Aggregates, adjus
 
       val sliceQueries = Seq(
         spanName.map { name =>
-          Seq(SpanSliceQuery(serviceName, name, endTs, 1))
+          Seq(SpanSliceQuery(serviceName, name, endTs, limit))
         },
         queryRequest.`annotations`.map {
           _.map { a =>
-            AnnotationSliceQuery(serviceName, a, None, endTs, 1)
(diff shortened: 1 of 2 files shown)


franklinhu:
This might be a Cassandra specific use case that should be refactored into the query service. The motivation behind only fetching one column first for each subquery is to address sparse return sets. 
For example, imagine you're joining two subqueries that fetch slices (`S1` and `S2`) from their respective indices. These slices are ordered by timestamp, so let's let the last entry in each slice have timestamp `T1` and `T2`. Since spans from the same trace _probably_ have a small difference in timestamps, we should set the `endTimestamp` to be `min(T1, T2)`. This will fix the problem where you have a very sparse index joined with a giant one so you only zero in on the time slice where the small index has entries.
You can see that do the actual query a bit father down in that method.

Are you seeing bad/unexpected behavior in Redis?

mosesn:
I am only getting one trace back.  I think what is happening is that, because I am only asking about one span, it goes to the "case head :: nil" case and then executes the query by itself.  Clearly, head.execute is executing with a limit of one trace id.  It seems like this might not be the correct solution to the problem, because I didn't understand what was going on, but I think this is definitely a problem.  This change fixed my case where I was only getting one trace back.  I invited you to chat on gchat, which might be a more efficient way of figuring this out.

mosesn:
This newer fix changes the "one slice requested" case to fit the "many slices requested" solution.

franklinhu:
This makes sense and looks good. Could you add a unit test so it won't regress in the future?

franklinhu:
Going to go ahead and merge this since we have some users who are blocked by this.

PR #182, opened 2012-10-17
ui: replace static link with url update

franklinhu:
Rather than displaying a `link` button to the static search page, update the URL so it contains the same info.


johanoskarsson:
+1

PR #185, opened 2012-10-22
Make server address configurable

franklinhu:
Users may not want to bind to the default `InetAddress.getLocalHost` if they have multiple network interfaces.


johanoskarsson:
+1

PR #188, opened 2012-10-23
Project file refactor

franklinhu:
- Pull default settings into a method
- Set the project organization
- Set crossPaths = false, removing the Scala version from the generated artifacts


johanoskarsson:
+1

PR #190, opened 2012-10-24
ui: change default query client port to 9411

franklinhu:
Default the query client to use the query service's default port


johanoskarsson:
+1

PR #192, opened 2012-10-24
ui: catch binary annotation conversion errors

franklinhu:
Converting a malformed binary annotation throws exceptions that break the UI. In this case just catch the exception and say it could not be read.


johanoskarsson:
+1

johanoskarsson:
Perhaps add a test?

PR #195, opened 2012-10-31
collector: Default to sample all spans in dev config

franklinhu:


zipkin-collector-service/config/collector-dev.scala
```diff
@@ -60,6 +62,9 @@ new ScribeZipkinCollectorConfig {
 
   override def adaptiveSamplerConfig = new NullAdaptiveSamplerConfig {}
 
+  /* Default to sample everything */
```
johanoskarsson: Perhaps describe what the collector sampling stage does in a bit more detail here? Mention that if the collector gets too much traffic the next step would be to use the adaptive sampler.

johanoskarsson:
+1

PR #200, opened 2012-11-08
fix a small bug in the UI javascript's timeUnit function

timjr:
timeUnit() needs to return a two-item list, and it shouldn't subscript numbers...


franklinhu:
Looks good to me. Thanks!

PR #201, opened 2012-11-08
redis backend to zipkin, CR by Dan Simon

mosesn:
## motivation

we didn't want to use the cassandra backend, so we used made an alternative storage backend in redis.
## implementation

redis data structures are generally pretty good, but we needed slightly different ways of thinking about them for zipkin.  I have slightly different abstractions around them, which comprise most of this code.  Index and Storage end up mostly being just about manipulating those data structures.


johanoskarsson:
Thanks for contributing this back. Great to have support for more storage systems.
I'll add a few comments inline, but one general request would be to add more comments. Class level describing what the class does and how + method level if it's not immediately clear what is going on (err on the side of adding too many comments).

zipkin-redis/src/main/scala/com/twitter/zipkin/config/RedisStorageConfig.scala
```diff
@@ -0,0 +1,34 @@
...
+
+package com.twitter.zipkin.config
+
+import com.twitter.zipkin.storage.redis.RedisStorage
+import com.twitter.finagle.redis.Client
+import com.twitter.util.Duration
+import com.twitter.conversions.time.intToTimeableNumber
+
+trait RedisStorageConfig extends StorageConfig {
+  lazy val _client: Client = Client("0.0.0.0:%d".format(port))
```
johanoskarsson: Seems restrictive to hardcode 0.0.0.0 here. Add another val for host?

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/ExpiringValue.scala
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package com.twitter.zipkin.storage.redis
+
+import com.twitter.util.Time
+
+case class ExpiringValue[A](expiresAt: Time, value: A)
+
+object ExpiringValue {
+  def apply[A](expiresAt: Long, value: A): ExpiringValue[A] =
```
johanoskarsson, line 33: Comment here describing what the expiresAt is seconds would be nice

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
```diff
@@ -0,0 +1,233 @@
...
+    for (serviceName <- span.serviceNames.toSeq
+      if serviceName != "")
+      yield serviceArray.add(serviceName))
+
+  /**
+   * Index the span name on the service name. This is so we
+   * can get a list of span names when given a service name.
+   * Mainly for UI purposes
+   */
+  override def indexSpanNameByService(span: Span): Future[Unit] = if (span.name != "") Future.join(
```
johanoskarsson: I'd generally prefer to not cram in too much in addition to the method declaration on one line

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/package.scala
```diff
@@ -0,0 +1,96 @@
...
+import com.twitter.util.Time
+import org.jboss.netty.buffer.ChannelBuffers
+import org.jboss.netty.util.CharsetUtil
+import java.nio.charset.Charset
+import org.jboss.netty.buffer.ChannelBuffer
+
+/**
+ * Useful conversions for encoding/decoding.
+ */
+package object redis {
```
johanoskarsson, line 30: Some of these don't seem to be tested, worth adding a few specs

zipkin-redis/src/test/scala/com/twitter/zipkin/storage/redis/RedisIndexSpec.scala
```diff
@@ -0,0 +1,227 @@
...
+      redisIndex.getSpanNames("service")() mustEqual Set(span1.name)
+    }
+
+    "index and get service names" in {
+      redisIndex.indexServiceName(span1)()
+      redisIndex.getServiceNames() mustEqual Set(span1.serviceNames.head)
+    }
+
+    /*
+    "index only on annotation in each span with the same value" in {
```
johanoskarsson: There's some old Cassandra references here that I assume can be safely removed

johanoskarsson:
For those of us that aren't familiar with redis a quick readme describing how to use this module with redis would be cool (including a quick starter on how to get redis running or a link to such a doc).

mosesn:
re comments:
do you want me to add comments to private methods too, or only non-private?  also, should methods which override methods in traits (which already have comments) also be commented?

johanoskarsson:
For the comments you don't have to add one if the trait you extend have comments already, unless your implementation does something worth bringing up.

For private methods that are complex or hard to understand I would add a comment, but not needed for the ones where the name and parameters or the code is self explanatory. For public methods I would do the same, but lower your sensitivity to complex methods, meaning: unless it's absolutely clear what is going on I'd add a comment. Methods like getTimeToLive: Duration you don't have to add a comment to for example.

zipkin-redis/src/main/scala/com/twitter/zipkin/config/RedisIndexConfig.scala
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.twitter.zipkin.config
+
+import com.twitter.zipkin.storage.redis.RedisIndex
+import com.twitter.finagle.redis.Client
+import com.twitter.util.Duration
+import com.twitter.conversions.time.intToTimeableNumber
```
franklinhu: should sort the imports on all files

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisHash.scala
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package com.twitter.zipkin.storage.redis
+
+import com.twitter.finagle.redis.Client
+import com.twitter.util.Future
+import java.nio.ByteBuffer
+import org.jboss.netty.buffer.ChannelBuffer
+
+class RedisHash(database: Client, name: String) {
```
franklinhu, line 28: docs about what this does?

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
```diff
@@ -0,0 +1,233 @@
...
+        case AnnotationType.I16 => bin.value.getShort.toString
+        case AnnotationType.I32 => bin.value.getInt.toString
+        case AnnotationType.I64 => bin.value.getLong.toString
+        case _ => ChannelBuffers.copiedBuffer(bin.value)
+      }
+
+      def binaryAnnoStringify(bin: BinaryAnnotation, service: String): String =
+        redisJoin(service, bin.key, encodeAnnotation(bin))
+
+      val time = span.lastAnnotation.get.timestamp
```
franklinhu, line 151: probably shouldn't be calling `.get` on the Option

zipkin-redis/src/test/scala/com/twitter/zipkin/storage/redis/RedisArrayMapSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+    val buf1 = ChannelBuffers.copiedBuffer("value1")
+    val buf2 = ChannelBuffers.copiedBuffer("value2")
+    val buf3 = ChannelBuffers.copiedBuffer("value3")
+
+    doBefore {
+      _client.flushDB()
+      arrayMap= new RedisArrayMap(_client, "prefix", None)
+    }
+
+    "be able to insert an element properly" in {
```
franklinhu: for brevity we can omit the "be able to"

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
```diff
@@ -0,0 +1,233 @@
...
+      val binaryAnnos: Seq[Future[Unit]] = span.serviceNames.toSeq flatMap { serviceName =>
+        span.binaryAnnotations map { binaryAnno =>
+          binaryAnnotationsListMap.add(
+            binaryAnnoStringify(binaryAnno, serviceName),
+            time,
+            span.traceId
+          )
+          Future.Unit
```
franklinhu: This breaks the linking of the futures. If you want to convert a `Future[A]` into `Future[Unit]` you can use `myFuture.unit`

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
```diff
@@ -0,0 +1,233 @@
...
+  )
+
+  /**
+   * Store the service name, so that we easily can
+   * find out which services have been called from now and back to the ttl
+   */
+  override def indexServiceName(span: Span): Future[Unit] = Future.join(
+    for (serviceName <- span.serviceNames.toSeq
+      if serviceName != "")
+      yield serviceArray.add(serviceName))
```
franklinhu: Not a huge fan of the `for(..)` syntax. might be clearer as something like

```
span.serviceNames collect { case name if name != "" => serviceArray.add(name) }
```

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisSetMap.scala
```diff
@@ -0,0 +1,38 @@
...
+import com.twitter.util.Future
+import org.jboss.netty.buffer.ChannelBuffer
+import com.twitter.util.Duration
+
+class RedisSetMap(database: Client, prefix: String, defaultTTL: Option[Duration]) {
+  private[this] def preface(key: String) = "%s:%s".format(prefix, key)
+
+  def add(key: String, buf: ChannelBuffer): Future[Unit] = database.sAdd(preface(key), List(buf)) flatMap { _ =>
+    defaultTTL match {
+      case Some(ttl) => database.expire(preface(key), ttl.inSeconds) flatMap (_ => Future.Unit)
```
franklinhu: `.unit` instead?

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisSortedSetMap.scala
```diff
@@ -0,0 +1,41 @@
...
+import com.twitter.finagle.redis.protocol.ZRangeResults
+import com.twitter.util.Duration
+
+class RedisSortedSetMap(database: Client, prefix: String, defaultTTL: Option[Duration]) {
+  def preface(key: String) = "%s:%s".format(prefix, key)
+
+  def add(key: String, score: Double, buffer: ChannelBuffer): Future[Unit] =
+    database.zAdd(preface(key), score, buffer) flatMap { _ =>
+      defaultTTL match {
+        case Some(ttl) => database.expire(preface(key), ttl.inSeconds) flatMap { _ => Future.Unit }
```
franklinhu: `.unit`?

mosesn:
Thanks for all of your prompt feedback.  I've made changes that address all of the comments you've laid out for me.  Looking forward to more input, thanks!

zipkin-redis/src/main/scala/com/twitter/zipkin/config/RedisStorageConfig.scala
mosesn: Good point, done.

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/ExpiringValue.scala
mosesn, line 33: Done.

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
mosesn: Moved it all over to the next line

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/package.scala
mosesn, line 30: pretty much fully tested now

zipkin-redis/src/main/scala/com/twitter/zipkin/config/RedisIndexConfig.scala
mosesn: can't wait until organize imports for scala-ide can handle effective scala conventions . . . but done.

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisHash.scala
mosesn, line 28: added them.

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
mosesn, line 151: good point, made it a map instead.

zipkin-redis/src/test/scala/com/twitter/zipkin/storage/redis/RedisArrayMapSpec.scala
mosesn: done.

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
mosesn: neat, didn't know about that, thanks

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
mosesn: In general I like the for syntax, but I think you're right that collect makes it cleaner in this case.

zipkin-redis/src/test/scala/com/twitter/zipkin/storage/redis/RedisIndexSpec.scala
mosesn: Yep, completely forgot to clean up this class.  Much nicer now.

mosesn:
@johanoskarsson I added a redis doc, you can probably yum/apt-get install redis-server if you're on linux and try it out yourself.  Not sure how fully featured you want the redis doc to be.  I have basically no experience writing technical documents except for my own gratification, so I know that I am pretty nooby.  Also, is it in the right place?

re: comments, I added comments where they seemed useful.  I tended toward scaladoc-style comments rather than // comments because I feel like it's easier for // comments to get separated from the code they're referring to, unless you have something like

``` scala
code() //comment describing code.
```

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/ExpiringValue.scala
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * ExpiringValue represents a value with a time to live.
+ * expiresAt is the time when ExpiringValue will expire.
+ * value is the value that will expire.
+ */
+case class ExpiringValue[A](expiresAt: Time, value: A)
+
+object ExpiringValue {
+
+  // expiresAt is a long value in seconds from the epoch.
```
johanoskarsson: I'd actually prefer javadoc style here with a @param expiresAt <desc> so that it shows up properly in generated docs.

caniszczyk:
Also not sure if you're aware guys, Redis is available on Travis CI for testing purposes... you just have to ensure it's started: http://about.travis-ci.org/docs/user/database-setup/

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/ExpiringValue.scala
mosesn: Oops, I completely forgot how @param works, I'll go back and change all of those tonight.  Good catch.

franklinhu:
@caniszczyk We have Travis CI disabled right now due to maven.twttr.com's flakiness :(

caniszczyk:
What dependencies are on maven.twttr.com that we can't publish to Sonatype OSS and Maven Central?

Finagle?

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
```diff
@@ -0,0 +1,194 @@
...
+  private[redis] def redisJoin(items: String*) = items.mkString(":")
+
+  override def getTraceIdsByName(serviceName: String, span: Option[String],
+    endTs: Long, limit: Int): Future[Seq[IndexedTraceId]] =
+    serviceSpanMap.get(
+      serviceName,
+      span,
+      ttl map (dur => endTs - dur.inMicroseconds),
+      endTs,
+      limit) map zRangeResultsToSeqIds
```
franklinhu, line 91: would prefer to wrap each of these method blocks in braces

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
```diff
@@ -0,0 +1,194 @@
...
+  override def indexTraceIdByServiceAndName(span: Span) : Future[Unit] = Future.join(
+    (span.serviceNames toSeq) map { serviceName =>
+      (span.lastAnnotation map { last =>
+        serviceSpanMap.put(serviceName, Some(span.name), last.timestamp, span.traceId)
+      }).getOrElse(Future.Unit)
+    }
+  )
+
+  override def indexSpanByAnnotations(span: Span) : Future[Unit] = Future.join(
+    {
```
franklinhu, line 138: maybe a brace at the start of the function, and the `Future.join` inside?

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
```diff
@@ -0,0 +1,194 @@
...
+  override def indexServiceName(span: Span): Future[Unit] = Future.join(
+    span.serviceNames.toSeq collect {case name if name != "" => serviceArray.add(name)}
+  )
+
+  override def indexSpanNameByService(span: Span): Future[Unit] =
+    if (span.name != "") Future.join(
+      for (serviceName <- span.serviceNames.toSeq
+        if serviceName != "")
+        yield spanMap.add(serviceName, span.name)
+    ) else Future.Unit
```
franklinhu: would be clearer if styled like:

```
if (...) {
  ...
} else {
  ...
}
```

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
```diff
@@ -0,0 +1,194 @@
...
+
+  override def indexSpanDuration(span: Span): Future[Void] = {
+    traceHash.get(span.traceId) map {
+      case None => TimeRange.fromSpan(span) map { timeRange =>
+        traceHash.put(span.traceId, timeRange)
+      }
+      case Some(bytes) => indexNewStartAndEnd(span, bytes)
+    }
+    Future.Void
```
franklinhu: This breaks the future chaining. If converting to `Future[Void]`, prefer `myFuture.voided`

mosesn: Does Future[Void] even use future chaining?  Also, why is it Future[Void] instead of Future[Unit]?

edit: ahh, I thought Future[Void] was special, didn't realize it was just a Future wrapped around a null.  Still seems wrong to use Future[Void] instead of Future[Unit].

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisListMap.scala
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Inserts an item into a list
+   */
+  def put(key: String, value: ChannelBuffer): Future[Unit] = {
+    val string = preface(key)
+    Future.join(Seq(database.lPush(string, List(value)),
+    defaultTTL match {
+      case Some(ttl) => database.expire(string, ttl.inSeconds)
+      case None => Future.Unit
+    }))
```
franklinhu: A little confused about what this is supposed to do. It's pushing a key/value to Redis, then setting the expiration?
Is it okay for the expiration to be set before `lPush` does? It seems like this should be more like:

```
database.lPush(string, List(value)) flatMap { _ =>
  defaultTTL match {
    case Some(ttl) => database.expire(string, ttl.inSeconds).unit
    case None => Future.Unit
  }
}
```

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
franklinhu: Yeah I think we should make the interface return `Future[Unit]` :\ I'll file a ticket

franklinhu: we can fix it in a future review

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
mosesn, line 91: The scala style guide seems to suggest that I should not use braces in this case.  I'm not certain that this will become more clear if I wrap it in curly braces.

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
mosesn: The Scala style guide suggests omitting braces for if/else expressions, but I can see the argument for

``` scala
if (span.name != "")
  Future.join(
    for (serviceName <- span.serviceNames.toSeq
      if serviceName != "")
      yield spanMap.add(serviceName, span.name)
  )
else
  Future.Unit
```

Is that more like what you would like to see?

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisListMap.scala
mosesn: Yeah, you're right.  I didn't really understand Futures when I wrote that.

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
franklinhu: Yeah that's a bit clearer

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
mosesn: fixed this to use .voided for now, I'll switch it when the trait changes.

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
mosesn: changed this to look like this

zipkin-redis/src/test/scala/com/twitter/zipkin/storage/redis/RedisSortedSetMapSpec.scala
```diff
@@ -0,0 +1,85 @@
...
+import scala.util.Random
+
+class RedisSortedSetMapSpec extends RedisSpecification {
+  val ep = Endpoint(123, 123, "service")
+
+  def binaryAnnotation(key: String, value: String) =
+    BinaryAnnotation(
+      key,
+      ByteBuffer.wrap(value.getBytes),
+      gen.AnnotationType.String.toAnnotationType,
```
franklinhu: Should use `com.twitter.zipkin.common.AnnotationType.String` rather than converting the generated thrift struct.

zipkin-redis/src/test/scala/com/twitter/zipkin/storage/redis/RedisStorageSpec.scala
```diff
@@ -0,0 +1,97 @@
...
+
+class RedisStorageSpec extends RedisSpecification {
+
+  var redisStorage: RedisStorage = null
+
+  def binaryAnnotation(key: String, value: String) =
+    BinaryAnnotation(
+      key,
+      ByteBuffer.wrap(value.getBytes),
+      gen.AnnotationType.String.toAnnotationType,
```
franklinhu: Same here

franklinhu:
Just had two small nits. Otherwise lgtm

zipkin-redis/src/test/scala/com/twitter/zipkin/storage/redis/RedisSortedSetMapSpec.scala
mosesn: thanks, fixed all of these.

franklinhu:
I tried running the tests, and didn't realize that they require for redis-server to be running in the background. Ideally this kind of a service dependency should be mocked out.

mosesn:
They don't require redis-server to be running, they do need redis to be installed on the machine that hosts it, though.  RedisCluster spins up a new instance of redis.  RedisCluster is also the way that finagle-redis tests itself, so it seems like right now it is the standard.

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisSetMap.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import com.twitter.finagle.redis.Client
+import com.twitter.util.{Duration, Future}
+import org.jboss.netty.buffer.ChannelBuffer
+
+/**
+ * RedisSetMap is a map from strings to sets.
+ * @database the redis client to use
+ * @prefix the namespace of the set
+ * @defaultTTL the timeout on the set
```
franklinhu: These should be `@param` annotations

franklinhu:
Tests pass for me. Just one last nit

mosesn:
Ok, I think it should be good now.  Weird, I thought I had fixed the last of the @param problems already.

PR #204, opened 2012-11-24
InetAddress.getLocalHost requires the local hostname to be resolveable.

olix0r:
It is safer to simply resolve "localhost".


franklinhu:
Would it be better just to use `127.0.0.1` instead?

franklinhu:
It's probably actually safer to use `localhost` for theoretical IPv6 compatibility

PR #205, opened 2012-11-25
Fix util.Duration json serialization

franklinhu:
Jackson can't serialize `util.Duration`, so convert `Annotation`s to
`JsonAnnotation`s that have a String duration field in microseconds


mosesn:
why do you use a string instead of a long?  I don't think it matters, just curious.

franklinhu:
Javascript only has a `Number` type (64 bit float) that causes problems when you have large Longs since the conversion is lossy. We saw the problem a while back when rendering the trace pages, so I've just been either Strings or Ints when serializing to JSON depending on the case.

johanoskarsson:
+1

PR #209, opened 2012-12-01
Cassandra cleanup

franklinhu:
Yak-shaving to help get to the Builder style configs. This basically fixes the mistake that was passing the config objects into the class instead of actual parameters.


johanoskarsson:
+1

Issue #207, opened 2012-11-29
Simplify configuration


franklinhu:
We've had quite a bit of feedback via the mailing lists that the configurations of the collector, query service, and web UI are quite confusing. Users are unclear which parts they do and do not need, and are often put off by there being too many moving parts. In reality, the basic collector/query setup only requires a backing store (either Cassandra or Redis at the time of this writing).

The base configs are pretty confusing, so I'm proposing moving to a Finagle `Builder` style config. 
Base config:

``` scala
CollectorBuilder(Scribe.Interface)
  .writeTo(Cassandra.static("hostname"))
  .writeTo(Cassandra.zookeeperServerSet("/zookeeper/path/to/cassandra/cluster/serverset"))
```

Or maybe you want to use Redis instead

``` scala
CollectorBuilder(...)
  .writeTo(Redis.static(...))
```

Or maybe you want to listen for JSON on an HTTP port (doesn't exist yet, but is conceivable)

``` scala
CollectorBuilder(Http.Json)
```

Or, if you want something more intense with sampling

``` scala
CollectorBuilder(...)
  .sampler(Sampler.static(0.5)) // sample 50% of spans
```

Or even adaptive sampling

``` scala
CollectorBuilder()
  .sampler(AdaptiveSampler.zookeeper(...))
```

The builders would be similar for the query service and the web UI. This would also allow for users to spin up all three services in the same JVM instance by returning a `Seq[Builder]`

Thoughts?

fspillner:
+1 

This makes the configuration API more easier to understand / use. And you decouple the internal of the configuration classes, so you can feel free to change it without change the interface of this new configuration API. These changes make the configuration more fluent to read and will have positive affect on the documentation of the configuration API. I like the idea, that the whole Zipkin project can be configured on one place with multiple builders.

eirslett:
Closing due to lack of activity. It has also been superseded by other issues about configuration.

PR #210, opened 2012-12-05
Cassandra builders

franklinhu:
- Move Cassandra Storage, Index, Aggregates to immutable builders.
- Slightly hack the configs for ZipkinSpec to work with this builder style


johanoskarsson:
+1

franklinhu:
This is related to #207

PR #214, opened 2013-01-07
Unify storage classes under Store

franklinhu:
To make configuration simpler, introduce a Store case class that wraps Storage, Index, and Aggregates. This is a temporary stopgap to get to Builder style configs.


johanoskarsson:
+1

franklinhu:
Related to #207

PR #219, opened 2013-01-08
web: Builder style config

franklinhu:
- Minor refactoring to clean up interfaces
- Switch to builder style configs for zipkin-web


zipkin-web/src/main/scala/com/twitter/zipkin/builder/WebBuilder.scala
```diff
@@ -0,0 +1,60 @@
...
+import com.twitter.zipkin.gen
+import com.twitter.zipkin.config.{CssConfig, JsConfig}
+import com.twitter.zipkin.web.{Resource, ZipkinWeb, App}
+
+case class WebBuilder(
+  rootUrl: String,
+  queryClientBuilder: Builder[ClientBuilder.Complete[ThriftClientRequest, Array[Byte]]],
+  pinTtl: Duration = 30.days,
+  resourcePathPrefix: String = "/public",
+  serverBuilder: ZipkinServerBuilder = ZipkinServerBuilder(8080, 9902)
```
franklinhu, line 32: I don't really like passing in the `ZipkinServerBuilder` as an argument, but subclassing it causes all sorts of Scala horror and breaks the niceties from using case classes. Not sure if there's a cleaner way.

zipkin-web/src/main/scala/com/twitter/zipkin/builder/QueryClient.scala
```diff
@@ -0,0 +1,53 @@
...
+  def zookeeper(
+    zkClientBuilder: ZooKeeperClientBuilder,
+    serverSetPath: String
+  ) = new Builder[ClientBuilder.Complete[ThriftClientRequest, Array[Byte]]] {
+    def apply() = {
+      val serverSet = new ServerSetImpl(zkClientBuilder.apply(), serverSetPath)
+      val cluster = new ZookeeperServerSetCluster(serverSet) {
+        override def ready() = super.ready
+      }
+      val a = default.cluster(cluster)
```
johanoskarsson: No need for the variable

johanoskarsson:
+1

PR #222, opened 2013-01-10
collector: Move collector admin endpoints

franklinhu:
The methods for modifying the sample rate and storage request rate don't really belong in the collector's thrift interface. Moved them to the ostrich admin interface.


zipkin-collector-core/src/main/scala/com/twitter/zipkin/config/ZipkinCollectorConfig.scala
```diff
@@ -106,8 +107,58 @@ trait ZipkinCollectorConfig extends ZipkinConfig[ZipkinCollector] {
...
+    addConfigEndpoint()
     new ZipkinCollector(this)
   }
+
+  /**
+   * Add endpoints to the Ostrich admin service for configuring the adjustable values
+   *
+   * Methods:
+   *   /config/get/sampleRate
+   *   /config/set/sampleRate?value=0.1
```
johanoskarsson: I assume Ostrich doesn't support POST?

zipkin-collector-core/src/main/scala/com/twitter/zipkin/config/ZipkinCollectorConfig.scala
```diff
@@ -106,8 +107,58 @@ trait ZipkinCollectorConfig extends ZipkinConfig[ZipkinCollector] {
   val serverConfig: CollectorServerConfig
 
   def apply(runtime: RuntimeEnvironment): ZipkinCollector = {
+    addConfigEndpoint()
```
johanoskarsson, line 110: Should probably add some tests for this

franklinhu, line 110: Done

johanoskarsson:
+1

PR #224, opened 2013-01-10
web: Upgrade to Finatra 1.1.1

franklinhu:


zipkin-web/src/main/scala/com/twitter/zipkin/web/ZipkinWeb.scala
```diff
@@ -35,36 +34,19 @@ class ZipkinWeb(
...
-        .codec(Http())
-        .bindTo(new InetSocketAddress(serverPort))
-        .name("ZipkinWeb")
-        .tracerFactory(tracerFactory)
-        .build(service)
-    }
-    log.info("Finatra service started in port: " + serverPort)
-  }
+    /* Finatra uses System properties for configuration */
+    System.setProperty("name", "ZipkinWeb")
```
johanoskarsson: Wow, this feels very wrong.

zipkin-web/src/main/scala/com/twitter/zipkin/web/ZipkinWeb.scala
```diff
@@ -35,36 +34,19 @@ class ZipkinWeb(
...
-
-    server = Some {
-      ServerBuilder()
-        .codec(Http())
-        .bindTo(new InetSocketAddress(serverPort))
-        .name("ZipkinWeb")
-        .tracerFactory(tracerFactory)
-        .build(service)
-    }
-    log.info("Finatra service started in port: " + serverPort)
```
johanoskarsson, line 58: Would be nice to keep this

zipkin-web/src/main/scala/com/twitter/zipkin/web/ZipkinWeb.scala
```diff
@@ -35,36 +34,19 @@ class ZipkinWeb(
...
+    s.start(tracerFactory)
 
-  def shutdown() {
-    server.foreach { _.close() }
   }
 
-  def register(app: AbstractFinatraController[Request, Future[Response], Future[HttpResponse]]) {
-    controllers.add(app)
-  }
+  def shutdown() {}
```
johanoskarsson: How does it shut down now?

zipkin-web/src/main/scala/com/twitter/zipkin/web/ZipkinWeb.scala
franklinhu: Finatra does config through system properties right now. It'd be a stopgap until @capotej gets out a more robust configuration system.

franklinhu:
@johanoskarsson We don't necessarily have to use the `FinatraServer` class, could just keep sort of what we had before and piece it together

franklinhu:
@johanoskarsson This feels much cleaner now.

johanoskarsson:
+1

capotej:
woot!

Yeah, I'm looking to replace my hacky config system with something cleaner in the future, perhaps https://github.com/typesafehub/config

PR #225, opened 2013-01-11
collector: builder style configs

franklinhu:
- `CollectorServiceBuilder` now is the way to build a collector service
- Add `Adaptive`, `Adjustable`, `CollectorInterface` convenience functions for more intuitive configuration files.
- Update `collector-dev.scala`
- Add `collector-zk.scala` that uses Zookeeper for sample rate configuration
- Add `collector-zk-adaptive.scala`, same as above but with an adaptive sampler


zipkin-collector-core/src/main/scala/com/twitter/zipkin/collector/builder/CollectorServiceBuilder.scala
```diff
@@ -0,0 +1,137 @@
...
+import com.twitter.zipkin.storage.Store
+import java.net.InetSocketAddress
+
+/**
+ * Builder for ZipkinCollector
+ *
+ * CollectorInterface[T] stands up the actual server, and its filter is used to process the object
+ * of type T into a Span in a worker thread.
+ *
+ * @param interface
```
johanoskarsson, line 40: These don't do much.

zipkin-collector-core/src/main/scala/com/twitter/zipkin/collector/builder/CollectorServiceBuilder.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  val log = Logger.get()
+
+  def writeTo(sb: Builder[Store]) =
+    copy(storeBuilders = storeBuilders :+ sb)
+
+  def addConfigEndpoint(name: String, builder: Builder[AdjustableRateConfig]) =
+    copy(additionalConfigEndpoints = additionalConfigEndpoints :+ (name, builder))
+
+  def register(s: Builder[(InetSocketAddress, StatsReceiver, Timer) => OstrichService]) =
```
johanoskarsson, line 96: Probably worth adding a short javadoc to some of these methods

johanoskarsson:
Some more javadoc in general would be nice, otherwise +1

zipkin-collector-core/src/main/scala/com/twitter/zipkin/collector/builder/CollectorServiceBuilder.scala
franklinhu, line 40: The implementation is in zipkin-collector-scribe `com.twitter.zipkin.builder.Scribe`. Needed a way to abstract out the thing that stands up the server, since it varies a lot (scribe/thrift, json over http, etc). The `filter` allows the service to dump whatever they want into the queue, and potentially do the expensive stuff in the worker threads.

PR #226, opened 2013-01-15
web: Use routeParams instead of params

franklinhu:
Bug fix from Finatra upgrade where `params` changed to `routeParams`


johanoskarsson:
+1

PR #232, opened 2013-02-26
trying out relative urls

bmdhacks:
This makes all web urls relative so that we can individually hit up servers that are part of a cluster.


bmdhacks:
I just noticed that this would break in a vhost scenario where your root_url was something like http://www.example.com/zipkin/
I doubt anybody would do that if they're running zipkin in a cluster anyways, and if they're not in a cluster than finatra dominates the http port.

franklinhu:
+1 I assume the build failed due to that test being flaky?

bmdhacks:
Yeah, it passed on the initial branch commit.

PR #233, opened 2013-03-08
added an error handler for zipkin-web

bmdhacks:
This doesn't fully work until we integrate with my patch to finatra, but I'm submitting this now because it's done.  Once finatra releases the new version we'll change the build and it will all be happy.


johanoskarsson:
+1 from me, but Franklin is probably a better reviewer for this one

franklinhu:
+1

PR #234, opened 2013-03-12
Created ubuntu quick start documention

eodgooch:
I've created a quick start guide to get zipkin running on ubuntu 12.04.


bmdhacks:
Can you comment on how relevant the existing centos guide is?
Should we delete it or is it still useful?

eodgooch:
You can delete it. There are a few inconsistencies and could be updated but
the ubuntu guide should cover centos install as well..

Aaron

bmdhacks:
Thanks for the docs.  Committed with a bit of editing.  One thing to note is that the default configs don't use zookeeper, so your zookeeper config may or may not be set up correctly.

PR #211, opened 2013-03-13
UI should use unsigned hex trace ID in URL

bmdhacks:


zipkin-web/src/main/scala/com/twitter/zipkin/web/App.scala
```diff
@@ -236,6 +237,18 @@ class App(
   }
 
   /**
+   * Get the id param from the http request (which is a hex string) and convert it
+   * to it's numerical value.
+   */
+
```
franklinhu, line 243: whitespace

franklinhu:
+1

PR #235, opened 2013-03-14
Adding some convenience functions to Span and Endpoint classes.

bmdhacks:


zipkin-common/src/main/scala/com/twitter/zipkin/common/Endpoint.scala
```diff
@@ -32,6 +35,25 @@ object Endpoint {
...
+   */
+  def getInetSocketAddress = {
+    val bytes = ByteBuffer.allocate(4).putInt(this.ipv4).array()
+    new InetSocketAddress(InetAddress.getByAddress(bytes), this.getUnsignedPort)
+  }
+
+  /**
+   * Convenience function to get the string-based ip-address
+   */
+  def getHostAddress = {
```
johanoskarsson: It's easier to read the code if the return type is included in public methods. For private ones I don't have a strong opinion either way.

zipkin-common/src/main/scala/com/twitter/zipkin/common/Endpoint.scala
```diff
@@ -32,6 +35,25 @@ object Endpoint {
 case class Endpoint(ipv4: Int, port: Short, serviceName: String)
   extends Ordered[Endpoint] {
 
+  /**
+   * Return the java.net.InetSocketAddress which contains host/port
+   */
+  def getInetSocketAddress: InetSocketAddress = {
```
johanoskarsson, line 41: Forgot one thing in the last review: would be nice with some quick tests for these new methods too

bmdhacks, line 41: Good call, totally slipped my mind.  On it's way...

johanoskarsson:
+1

PR #236, opened 2013-03-22
Beautify dependency graph and add page titles

bmdhacks:


johanoskarsson:
I'll let Franklin review this one, but to my untrained frontend eyes it looks good.

zipkin-web/src/main/scala/com/twitter/zipkin/web/App.scala
```diff
@@ -383,14 +385,17 @@ class App(
   }
 }
 
+abstract class ZipkinView(val pageTitle:String) extends View
+
 class IndexView(
+  override val pageTitle: String,
```
franklinhu: don't need the override since `pageTitle` is abstract.

zipkin-web/src/main/scala/com/twitter/zipkin/web/App.scala
```diff
@@ -383,14 +385,17 @@ class App(
   }
 }
 
+abstract class ZipkinView(val pageTitle:String) extends View
```
franklinhu: Would prefer to use trait rather than abstract class

franklinhu:
+1

PR #240, opened 2013-04-13
bind to 0.0.0.0 by default

bmdhacks:


franklinhu:
+1, but would be good to give a more in-depth description for why the change is necessary, and how it fixes what's broken

bmdhacks:
If you bind to your primary public IP, then you can't get to the service on localhost.  Vice-versa if you're bound to localhost.  I kept getting confused when running ssh port-forwarding or some combination of development/production instances about which ip the service was listening to.  This also comes into effect when your ip changes, say I take my laptop home running zipkin and get assigned a new IP.  Binding to 0.0.0.0 makes the service reachable by both 127.0.0.1 and the public IP.  Most network services do it this way and rely on firewalls etc to limit access.

PR #248, opened 2013-05-24
Strongly Typed Aggregate Dependencies

bmdhacks:
This is the start of work to make an aggregate dependency map that includes timing information.  It's mostly untested, but I've modified and added unit tests where needed.  Next I plan to test it via some scalding jobs that will submit to this interface.

### DIFF
--- a/zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraAggregates.scala
+++ b/zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraAggregates.scala
@@ -61,7 +61,7 @@ case class CassandraAggregates(
     val realStart = floorTime(startDate)
     val realEnd = floorTime(endDate.getOrElse(startDate))
 
-    val rows = NumericRange[Long](realEnd, realStart, 1.days.inMicroseconds)
+    val rows = new NumericRange.Inclusive[Long](realEnd, realStart, 1.days.inMicroseconds)
 
     val result: Future[Iterable[gen.Dependencies]] =
       dependenciesCF.multigetRows(rows.toSet.asJava, None, None, Order.Normal, Int.MaxValue)
--- a/zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraAggregates.scala
+++ b/zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraAggregates.scala
@@ -48,18 +48,13 @@ case class CassandraAggregates(
   val Delimiter: String = ":"
 
   /**
-   * floor the time to day in milliseconds
-   */
-  private[this] def floorTime(t: Time) = { (t.inMicroseconds / 1.days.inMicroseconds) * 1.days.inMicroseconds }
-
-  /**
    * Get the top annotations for a service name
    */
   def getDependencies(startDate: Time, endDate: Option[Time]) : Future[Dependencies] = {
 
     // floor to nearest day in microseconds
-    val realStart = floorTime(startDate)
-    val realEnd = floorTime(endDate.getOrElse(startDate))
+    val realStart = startDate.floor(1.day).inMicroseconds
+    val realEnd = endDate.getOrElse(startDate).floor(1.day).inMicroseconds
 
     val rows = new NumericRange.Inclusive[Long](realEnd, realStart, 1.days.inMicroseconds)
 
@@ -113,7 +108,7 @@ case class CassandraAggregates(
 
   /** Synchronize these so we don't do concurrent writes from the same box */
   def storeDependencies(deps: Dependencies): Future[Unit] = {
-    store[Long,gen.Dependencies](dependenciesCF, floorTime(deps.startTime), Seq(deps.toThrift))
+    store[Long,gen.Dependencies](dependenciesCF, deps.startTime.floor(1.day).inMicroseconds, Seq(deps.toThrift))
   }
 
   /** Synchronize these so we don't do concurrent writes from the same box */
--- a/zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraAggregates.scala
+++ b/zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraAggregates.scala
@@ -16,9 +16,15 @@
 package com.twitter.zipkin.storage.cassandra
 
 import com.twitter.cassie._
-import com.twitter.util.{Future, Return, Throw}
+import com.twitter.util.{Time, Future, Return, Throw}
+import com.twitter.conversions.time._
 import com.twitter.zipkin.storage.Aggregates
+import com.twitter.zipkin.conversions.thrift._
 import scala.collection.JavaConverters._
+import com.twitter.zipkin.gen
+import scala.collection.immutable.NumericRange
+import com.twitter.zipkin.common.Dependencies
+import com.twitter.algebird.Monoid
 
 /**
  * Cassandra backed aggregates store
@@ -32,7 +38,7 @@ import scala.collection.JavaConverters._
 case class CassandraAggregates(
   keyspace: Keyspace,
   topAnnotations: ColumnFamily[String, Long, String],
-  dependencies: ColumnFamily[String, Long, String]
+  dependenciesCF: ColumnFamily[Long, Long, gen.Dependencies]
 ) extends Aggregates {
 
   def close() {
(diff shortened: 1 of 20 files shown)


zipkin-cassandra/src/test/scala/com/twitter/zipkin/storage/cassandra/CassandraAggregatesSpec.scala
```diff
@@ -122,8 +148,9 @@ class CassandraAggregatesSpec extends Specification with JMocker with ClassMocke
         agg.storeTopAnnotations(serviceName, anns2).apply()
         agg.getTopAnnotations(serviceName).apply() mustEqual anns2
 
-        agg.storeDependencies(serviceName, calls1).apply()
-        agg.getDependencies(serviceName).apply() mustEqual calls1
+        // FIXME - probably shouldn't clobber non-overlapping deps
```
IceCreamYou: Per discussion with Brian, we don't actually care about testing this case since we want to fix it, so we might as well go ahead and remove this test.

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraAggregates.scala
```diff
@@ -42,11 +48,32 @@ case class CassandraAggregates(
   val Delimiter: String = ":"
 
   /**
+   * floor the time to day in milliseconds
+   */
+  private[this] def floorTime(t: Time) = { (t.inMicroseconds / 1.days.inMicroseconds) * 1.days.inMicroseconds }
```
johanoskarsson: There's already a floor method in Time

bmdhacks:
Ok, this should be good-to-go now.

zipkin-thrift/src/main/resources/thrift/zipkinDependencies.thrift
```diff
@@ -0,0 +1,40 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+namespace java com.twitter.zipkin.gen
+namespace rb Zipkin
+
+//********* Zipkin Aggregate Dependency Related Structs ***********
+
+
+// hold statistical information about stuff.  Maps to algebird Moments
+struct Moments {
```
johanoskarsson, line 23: This is a bit unclear. Statistical information about stuff? Why isn't m1 called mean if that's what it is?

zipkin-thrift/src/main/resources/thrift/zipkinQuery.thrift
```diff
@@ -212,7 +213,7 @@ service ZipkinQuery {
     i32 getDataTimeToLive() throws (1: QueryException qe);
 
     /** Aggregates related */
-    list<string> getDependencies(1: string service_name) throws (1: QueryException qe);
+    zipkinDependencies.Dependencies getDependencies(1: i64 start_time, 2: optional i64 end_time) throws (1: QueryException qe);
```
johanoskarsson, line 221: Bring over some of the api comment from the rest api here since that would be converted into javadoc in some generated code (or whatever the language might be).

zipkin-thrift/src/main/resources/thrift/zipkinDependencies.thrift
```diff
@@ -0,0 +1,40 @@
...
+
+struct DependencyLink {
+  1: string parent,  // parent service name (caller)
+  2: string child,   // child service name (callee)
+  3: Moments durationMoments
+  // histogram?
+}
+
+struct Dependencies {
+  1: i64 startTime  // microseconds from epoch
```
johanoskarsson: Thrift convention is start_time

zipkin-query-core/src/main/scala/com/twitter/zipkin/query/QueryService.scala
```diff
@@ -383,10 +383,13 @@ class QueryService(storage: Storage, index: Index, aggregates: Aggregates, adjus
     }
   }
 
-  def getDependencies(serviceName: String): Future[Seq[String]] = {
-    log.debug("getDependencies: " + serviceName)
+  /** Aggregates related */
+  def getDependencies(startTime: Long, endTime: Option[Long]) : Future[gen.Dependencies] = {
+    log.debug("getDependencies: " + startTime)
```
johanoskarsson: endtime too?

zipkin-query-core/src/main/scala/com/twitter/zipkin/query/QueryService.scala
```diff
@@ -383,10 +383,13 @@ class QueryService(storage: Storage, index: Index, aggregates: Aggregates, adjus
...
     call("getDependencies") {
-      aggregates.getDependencies(serviceName)
+      val start = Time.fromNanoseconds(startTime*1000)
```
johanoskarsson, line 390: There's no Time.fromMilliseconds?

zipkin-thrift/src/main/resources/thrift/zipkinDependencies.thrift
bmdhacks, line 23: I'll clarify that.  m1 is called m1 for parity with Algebird.

zipkin-query-core/src/main/scala/com/twitter/zipkin/query/QueryService.scala
bmdhacks, line 390: There is no Time.fromMicroseconds, no.  I'll put up a pull request for util though.

zipkin-common/src/main/scala/com/twitter/zipkin/common/Dependencies.scala
```diff
@@ -0,0 +1,76 @@
...
+  endTime: Time,
+  links: Seq[DependencyLink]
+)
+
+object Dependencies {
+  // used for summing/merging database rows
+  implicit val monoid:Monoid[Dependencies] = new Monoid[Dependencies] {
+    def plus(l: Dependencies, r: Dependencies) = {
+      // new start/end should be the inclusive time span of both items
+      val newStart = Seq(r.startTime, l.startTime).min
```
johanoskarsson: I think there's a min on Time you can use directly

zipkin-common/src/main/scala/com/twitter/zipkin/common/Dependencies.scala
```diff
@@ -0,0 +1,76 @@
...
+ */
+case class Dependencies(
+  startTime: Time,
+  endTime: Time,
+  links: Seq[DependencyLink]
+)
+
+object Dependencies {
+  // used for summing/merging database rows
+  implicit val monoid:Monoid[Dependencies] = new Monoid[Dependencies] {
```
johanoskarsson, line 60: I'm not very familiar with implicits or the Monoid stuff, but if we're just merging two dependencies why not make that a method on the Dependencies instance instead? One could argue the method should be called + too, but I don't know if that's better or worse.

zipkin-collector-scribe/src/main/scala/com/twitter/zipkin/collector/ScribeCollectorService.scala
```diff
@@ -78,19 +79,17 @@ class ScribeCollectorService(
     }
   }
 
-  def storeDependencies(serviceName: String, endpoints:Seq[String]): Future[Unit] = {
-    Stats.incr("collector.storeTopAnnotations")
-    log.info("storeDependencies: " + serviceName + "; " + endpoints)
+  def storeDependencies(dependencies: gen.Dependencies) : Future[Unit] = {
```
johanoskarsson: Nit: we're trying to adhere to ): <return type> without the space before : 
I know we still have this style in some places, but we're trying to clean that up

zipkin-collector-core/src/main/scala/com/twitter/zipkin/collector/CollectorService.scala
```diff
@@ -29,5 +31,9 @@ trait CollectorService extends Service {
   def shutdown() {
     running = false
     writeQueue.shutdown()
+    stores.foreach { store =>
+      // index and storage handles are closed in their respective services
+      store.aggregates.close()
```
johanoskarsson, line 35: Isn't it better if we just close the store here and the store takes care of closing the aggregates if needed?

zipkin-common/src/main/scala/com/twitter/zipkin/common/Dependencies.scala
bmdhacks, line 60: The Monoid stuff works really well inside of Scalding.  I'm going to be reusing this structure in the scalding jobs to calculate aggregate dependencies.

FYI, look forward to a monoid definition for Trace and Span as well for the same reason.

zipkin-collector-core/src/main/scala/com/twitter/zipkin/collector/CollectorService.scala
bmdhacks, line 35: I couldn't decide.  In practice all the services (Index, Collector, Storage) all get shutdown at the same time, so it doesn't matter.  I suppose moving them all into the collector shutdown at least puts it all in the same place.  I'll make that change.

mosesn, line 35: If you really don't care, we could shove them into ServiceTracker and be done with it

bmdhacks, line 35: The problem is that the stores are not ostrich services, they are finagle clients, and thus cannot be stored in ServiceTracker.  The Index and Storage services are ostrich services, so they can close the respective index and storage handles to cassie, but the aggregates one does not have its own service.  My solution to this was to stuff it in the collector, but it's probably best to just close all the handles in the collector service (which is then tracked in the service tracker).

zipkin-common/src/main/scala/com/twitter/zipkin/common/Dependencies.scala
bmdhacks, line 60: Also, there's some implicit sugar you can import from Algebird to get the + operator on classes with implicit monoid conversions, so you can do: Dependency1 + Dependency2, but I chose to use the explicit form for readability.

zipkin-thrift/src/main/resources/thrift/zipkinDependencies.thrift
```diff
@@ -0,0 +1,42 @@
...
+  2: double m1, // mean
+  3: double m2, // variance * count
+  4: double m3,
+  5: double m4
+}
+
+struct DependencyLink {
+  1: string parent,  // parent service name (caller)
+  2: string child,   // child service name (callee)
+  3: Moments durationMoments
```
johanoskarsson: Nit: duration_moments

johanoskarsson:
+1 from me.
I didn't have time to dig deep into the monoid stuff so I'll trust you on that, will try to read up later and we can always have a discussion about it if I have concerns.

PR #249, opened 2013-05-24
Make it easier to show remote data in a local instance

IceCreamYou:
This change allows us to point Zipkin to a remote query source so that it can use live data in a local (testing/dev) setup.

### DIFF
--- a/zipkin-web/config/web-dev.scala
+++ b/zipkin-web/config/web-dev.scala
@@ -17,7 +17,10 @@ import com.twitter.finagle.zipkin.thrift.ZipkinTracer
 import com.twitter.zipkin.builder.{QueryClient, WebBuilder}
 import java.net.{InetSocketAddress, InetAddress}
 
-val queryClient = QueryClient.static(new InetSocketAddress(InetAddress.getLoopbackAddress(), 9411)) map {
+// We'd rather use InetAddress.getLoopbackAddress than 127.0.0.1 but it's JDK
+// 1.7+. Previously we used InetAddress.getLocalHost, but that doesn't let us
+// do SSH port forwarding.
+val queryClient = QueryClient.static(new InetSocketAddress("127.0.0.1", 9411)) map {
   _.tracer(ZipkinTracer.mk())
 }
 WebBuilder("http://localhost:8080/", queryClient)
--- a/zipkin-web/config/web-dev.scala
+++ b/zipkin-web/config/web-dev.scala
@@ -17,7 +17,7 @@ import com.twitter.finagle.zipkin.thrift.ZipkinTracer
 import com.twitter.zipkin.builder.{QueryClient, WebBuilder}
 import java.net.{InetSocketAddress, InetAddress}
 
-val queryClient = QueryClient.static(new InetSocketAddress(InetAddress.getLocalHost, 9411)) map {
+val queryClient = QueryClient.static(new InetSocketAddress(InetAddress.getLoopbackAddress(), 9411)) map {
   _.tracer(ZipkinTracer.mk())
 }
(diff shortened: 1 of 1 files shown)


zipkin-web/config/web-dev.scala
```diff
@@ -17,7 +17,7 @@ import com.twitter.finagle.zipkin.thrift.ZipkinTracer
 import com.twitter.zipkin.builder.{QueryClient, WebBuilder}
 import java.net.{InetSocketAddress, InetAddress}
 
-val queryClient = QueryClient.static(new InetSocketAddress(InetAddress.getLocalHost, 9411)) map {
+val queryClient = QueryClient.static(new InetSocketAddress(InetAddress.getLoopbackAddress(), 9411)) map {
```
bmdhacks: This might be nitpicky, but the scala way™ would be to remove the parens from the function call

mosesn: The ["rule"](http://docs.scala-lang.org/style/method-invocation.html#arity0) is that if it doesn't have side effects, omit the parentheses, if it has side effects, always use parentheses.

Is this done to migrate to java 1.7?  Is this an intentional breaking of backwards compatibility?

bmdhacks: zipkin has worked on jdk 1.7 for a while now.  This change was done because Isaac was trying to get zipkin web working under ssh port forwarding.  It shouldn't break any backward compatibility because a recent change made all services bind to 0.0.0.0
Does this change cause you problems on your machine?

mosesn: getLoopbackAddress is 1.7 only, so it seems like it would break on jdk 1.6.  Is it generally agreed that zipkin is only supposed to work for jdk 1.7?

IceCreamYou: Nope, that's my mistake. Zipkin should work on JDK 1.6 too. And the point about parens is not too nitpicky. Will fix shortly.

zipkin-web/config/web-dev.scala
```diff
@@ -17,7 +17,10 @@ import com.twitter.finagle.zipkin.thrift.ZipkinTracer
 import com.twitter.zipkin.builder.{QueryClient, WebBuilder}
 import java.net.{InetSocketAddress, InetAddress}
 
-val queryClient = QueryClient.static(new InetSocketAddress(InetAddress.getLocalHost, 9411)) map {
+// We'd rather use InetAddress.getLoopbackAddress than 127.0.0.1 but it's JDK
+// 1.7+. Previously we used InetAddress.getLocalHost, but that doesn't let us
+// do SSH port forwarding.
```
bmdhacks: Use /\* */ for multiline comments, but my preference would be to remove this to keep the config simple for noobies.  Talking about JDK versions will scare them away and is not necessary for understanding the existing config.

IceCreamYou: Well presumably we would like to change this in the future when we deprecate JDK 1.6 support, so it would be good to note that somewhere. And we'd also like to avoid someone coming along later and saying "oh, that shouldn't be hardcoded, I should change it to getLocalHost." Besides, I think it's a fair bet that someone capable of installing and maintaining Zipkin would know what the JDK is. Thoughts?

bmdhacks: It's just a config file for development.  It's not vital, and this is not the zipkin code.  It's there for noobies to use when they get started.  With that in mind, keeping things as simple as possible is preferable.  I still vote to take it out.

bmdhacks:
+1
Please run the bin/git-pull-request.rb command to merge this change

PR #257, opened 2013-06-13
Refactor Json

bmdhacks:
Ran down a side-path and refactored the json generation code.  This is a first step with no removal of existing functionality, but it makes it much easier to start removing the Json object conversions.

The main architectural change is to use a Jackson generator object to do the type conversion instead of scala type-pimping and the .toJson method.

Unit tests aren't rewritten, but I wanted to put this up to get feedback before pursuing further.

If this looks good, the next step is to start removing JsonXXX objects and to make the actual classes convert natively.


zipkin-web/src/main/scala/com/twitter/zipkin/common/json/ZipkinJson.scala
```diff
@@ -0,0 +1,39 @@
+package com.twitter.zipkin.common.json
+
+import com.codahale.jerkson.Json
+import org.codehaus.jackson.map.annotate.JsonCachable
+import com.twitter.zipkin.common.BinaryAnnotation
+import com.fasterxml.jackson.databind.{SerializerProvider, JsonSerializer}
+import com.fasterxml.jackson.core.{JsonGenerator=>JacksonGenerator}
+import com.fasterxml.jackson.databind.module.SimpleModule
+
```
franklinhu: whitespace

bmdhacks: do what now?
http://www.youtube.com/watch?v=h06cSTAqAYk

franklinhu:
lgtm

bmdhacks:
Gah, did the wrong operation.  Gonna squash-merge by hand. :-p

PR #259, opened 2013-06-14
support new ca/sa binary annotations

bmdhacks:


zipkin-web/src/main/resources/public/js/zipkin-span.js
```diff
@@ -97,6 +97,12 @@ Zipkin.fromRawSpan = function(rawSpan) {
   });
   var kvAnnotations = $.map(rawSpan.binaryAnnotations, function(b) {
     var ann =  Zipkin.fromRawKvAnnotation(b);
+    // convert ca/sa annotations to more readable
+    if (ann.key == "ca" || ann.key == "sa") {
+      ann.key = (ann.key == "ca") ? "Client Address" : "Server Address";
+      ann.setAnnotationType(6); // string
+      ann.setValue(ann.host.ipv4 + ":" + ann.host.port);
+    }
```
franklinhu, line 105: Are the other core annotations converted on the server? Might make sense to do it all in the same place

zipkin-web/src/main/resources/public/js/zipkin-kv-annotation.js
```diff
@@ -22,17 +22,27 @@
...
   KvAnnotation.prototype.getKey            = function() { return this.key; };
   KvAnnotation.prototype.getValue          = function() { return this.value; };
   KvAnnotation.prototype.getAnnotationType = function() { return this.annotationType; };
   KvAnnotation.prototype.getSpan           = function() { return this.span; };
+  KvAnnotation.prototype.getHost           = function() { return this.host; };
+  KvAnnotation.prototype.getHostAddress    = function() {
+    if (this.host !== undefined) {
+     return this.host.ipv4
+    } else {
+     return "0.0.0.0"
```
franklinhu: semicolons? :trollface:

zipkin-web/src/main/resources/public/js/zipkin-span.js
bmdhacks, line 105: good point.

bmdhacks, line 105: No they're not

PR #260, opened 2013-06-14
Make traces and spans monoids

bmdhacks:
Adding two spans with the same id will merge their annotations.
Adding two traces with the same id will merge their spans.
This is the basis for map/reduce jobs to coalesce spans into traces.
TODO - use the monoid for this behavior in the rest of the code.

### DIFF
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/common/Dependencies.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/common/Dependencies.scala
@@ -17,7 +17,7 @@
 package com.twitter.zipkin.common
 
 import com.twitter.util.Time
-import com.twitter.algebird.{Monoid, Semigroup, Moments}
+import com.twitter.algebird.{Monoid, Moments}
 
 /**
  * Abstraction of a service
@@ -35,23 +35,28 @@ case class DependencyLink(parent: Service, child: Service, durationMoments: Mome
 
 object DependencyLink {
   // this gives us free + operator along with other algebird aggregation methods
-  implicit val sg:Semigroup[DependencyLink] = new Semigroup[DependencyLink] {
+  implicit val monoid:Monoid[DependencyLink] = new Monoid[DependencyLink] {
+    val zero = DependencyLink(Service("zero"), Service("zero"), Monoid.zero[Moments])
     def plus(l: DependencyLink, r: DependencyLink) = {
-      assert(l.child == r.child && l.parent == r.parent)
-      DependencyLink(l.parent, l.child, Monoid.plus(l.durationMoments, r.durationMoments))
+      if (l == zero) r
+      else if (r == zero) l
+      else {
+        assert(l.child == r.child && l.parent == r.parent)
+        DependencyLink(l.parent, l.child, Monoid.plus(l.durationMoments, r.durationMoments))
+      }
     }
   }
 }
 
 /**
  * This represents all dependencies across all services over a given time period.
- * @param startTime the startTime time for this period
- * @param endTime how long the period lasted
+ * @param startTime the beginning of the time span this aggregates in epoch microseconds
+ * @param endTime the end of the time span in epoch microseconds
  * @param links link information for every dependent service
  */
 case class Dependencies(
-  startTime: Time,
-  endTime: Time,
+  startTime: Long,
+  endTime: Long,
   links: Seq[DependencyLink]
 )
 
@@ -63,14 +68,16 @@ object Dependencies {
       val newStart = r.startTime min l.startTime
       val newEnd = r.endTime max l.endTime
 
-      // links are merged by mapping to parent/child and summing corresponding links
-      val lLinkMap = l.links.map { link => (link.parent, link.child) -> link }.toMap
-      val rLinkMap = r.links.map { link => (link.parent, link.child) -> link }.toMap
-      val newLinks = Monoid.plus(rLinkMap, lLinkMap).values.toSeq
+      // links are merged by grouping by parent/child and summing corresponding links
+      val newLinks = (l.links ++ r.links)
+        .groupBy { link => (link.parent, link.child) }
+        .values
+        .map { Monoid.sum(_) }
+        .toSeq
 
       Dependencies(newStart, newEnd, newLinks)
     }
 
-    val zero = Dependencies(Time.Top, Time.Bottom, Seq.empty[DependencyLink])
+    val zero = Dependencies(Long.MaxValue, Long.MinValue, Seq.empty[DependencyLink])
   }
 }
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
@@ -17,6 +17,7 @@
 package com.twitter.zipkin.common
 
 import com.twitter.zipkin.Constants
+import com.twitter.algebird.{Monoid, Semigroup}
 
 /**
  * A span represents one RPC request. A trace is made up of many spans.
@@ -31,6 +32,34 @@ object Span {
 
   def apply(span: Span): Span = Span(span.traceId, span.name, span.id,
     span.parentId, span.annotations, span.binaryAnnotations, span.debug)
+
+  val invalid = new Span(0, "invalid", 0, None, Nil, Nil, true)
+  val zero = new Span(0, "zero", 0, None, Nil, Nil, true)
+
+  implicit val sg:Monoid[Span] = new Monoid[Span] {
+    def plus(l: Span, r:Span) = {
+      if (l == zero || r == invalid) r
+      else if (r == zero || l == invalid) l
+      else {
+        // This is a test for duplicate annotations in a span.  When we merge spans and we find
+        // duplicates, the span becomes invalid and we just ignore any further entries.  This is useful
+        // for filtering out rogue clients logging bogus data into the system.
+        val duplicates = for {
+          annotation <- Seq("cs", "cr", "ss", "sr")
+          lann <- l.annotations
+          rann <- r.annotations
+          if (lann.value == rann.value && rann.value == annotation)
+        } yield annotation
+
+        if (duplicates.isEmpty)
+          l.mergeSpan(r)
+        else
+          invalid
+      }
+    }
+
+    val zero = Span.zero
+  }
 }
 
 /**
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
@@ -30,9 +30,6 @@ import com.twitter.algebird.{Monoid, Semigroup}
  */
 object Span {
 
-  def apply(span: Span): Span = Span(span.traceId, span.name, span.id,
-    span.parentId, span.annotations, span.binaryAnnotations, span.debug)
-
   val invalid = new Span(0, "invalid", 0, None, Nil, Nil, true)
   val zero = new Span(0, "zero", 0, None, Nil, Nil, true)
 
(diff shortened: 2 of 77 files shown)


mosesn:
So the set of spans with the same id is one monoid, (identity is the span without annotations), the set of traces with the same id is another monoid (identity is the trace without any spans)?

bmdhacks:
Yeah, it can logically be viewed that way, but the implementation is a bit different because there's no way of having span ids influence the scala type.

I created a special zero span that invokes identity when added to any span, and the plus operation just throws an assertion if the spans do not have matching id's.

Mostly this is useful for some scalding jobs I'm working on, but I think a second phase of this will be to go back and simplify some of the span aggregation code in the collector.

zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
```diff
@@ -16,10 +16,10 @@
 package com.twitter.zipkin.query
 
 import com.twitter.finagle.tracing.{Trace => FTrace}
-import com.twitter.logging.Logger
 import com.twitter.zipkin.common.{BinaryAnnotation, Endpoint, Span}
 import java.nio.ByteBuffer
 import scala.collection.mutable
+import com.twitter.algebird.Monoid
```
franklinhu: let's sort these imports

zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
```diff
@@ -31,12 +31,44 @@ case class Timespan(start: Long, end: Long)
...
 object Trace {
   def apply(spanTree: SpanTreeEntry): Trace = Trace(spanTree.toList)
+
+  def invalid = Trace(Seq(Span.invalid))
+
+  def zero = Trace(Seq.empty[Span])
+
+  val MAX_SPANS = 10000 // any trace with more spans becomes invalid
+
+  implicit val monoid:Monoid[Trace] = new Monoid[Trace] {
```
franklinhu: space after the colon

zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
```diff
@@ -31,12 +31,44 @@ case class Timespan(start: Long, end: Long)
  */
 object Trace {
   def apply(spanTree: SpanTreeEntry): Trace = Trace(spanTree.toList)
+
+  def invalid = Trace(Seq(Span.invalid))
+
+  def zero = Trace(Seq.empty[Span])
```
franklinhu: should `invalid` and `zero` be `val`s?

zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
```diff
@@ -31,12 +31,44 @@ case class Timespan(start: Long, end: Long)
...
+
+  val MAX_SPANS = 10000 // any trace with more spans becomes invalid
```
franklinhu: Scala convention is to use `CapitalCamelCase` for constants

bmdhacks: What do you tink about making this a var, so it can be changed?

zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
bmdhacks: yes

zipkin-common/src/main/scala/com/twitter/zipkin/common/Dependencies.scala
```diff
@@ -35,23 +35,28 @@ case class DependencyLink(parent: Service, child: Service, durationMoments: Mome
...
- * @param endTime how long the period lasted
+ * @param startTime the beginning of the time span this aggregates in epoch microseconds
+ * @param endTime the end of the time span in epoch microseconds
  * @param links link information for every dependent service
  */
 case class Dependencies(
-  startTime: Time,
-  endTime: Time,
+  startTime: Long,
+  endTime: Long,
```
franklinhu, line 59: Why use Longs rather than Time objects?

zipkin-common/src/main/scala/com/twitter/zipkin/common/Dependencies.scala
```diff
@@ -35,23 +35,28 @@ case class DependencyLink(parent: Service, child: Service, durationMoments: Mome
 
 object DependencyLink {
   // this gives us free + operator along with other algebird aggregation methods
-  implicit val sg:Semigroup[DependencyLink] = new Semigroup[DependencyLink] {
+  implicit val monoid:Monoid[DependencyLink] = new Monoid[DependencyLink] {
+    val zero = DependencyLink(Service("zero"), Service("zero"), Monoid.zero[Moments])
```
franklinhu, line 39: This seems like it belongs outside as a constant and refed here

zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
```diff
@@ -31,6 +32,34 @@ object Span {
 
   def apply(span: Span): Span = Span(span.traceId, span.name, span.id,
     span.parentId, span.annotations, span.binaryAnnotations, span.debug)
+
+  val invalid = new Span(0, "invalid", 0, None, Nil, Nil, true)
+  val zero = new Span(0, "zero", 0, None, Nil, Nil, true)
```
franklinhu, line 34: start with capital for constants

zipkin-common/src/main/scala/com/twitter/zipkin/common/Dependencies.scala
bmdhacks, line 59: com.twitter.util.Time is not serializable :-p

zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
bmdhacks, line 34: The convention for most of these algebird things is to have Class.zero be lowercase

zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
```diff
@@ -31,12 +31,44 @@ case class Timespan(start: Long, end: Long)
  */
 object Trace {
   def apply(spanTree: SpanTreeEntry): Trace = Trace(spanTree.toList)
+
+  def invalid = Trace(Seq(Span.invalid))
```
franklinhu: may make sense to add an `isValid` method to `Trace` instead of having a separate instance

zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
```diff
@@ -31,12 +31,44 @@ case class Timespan(start: Long, end: Long)
...
+
+  val MAX_SPANS = 10000 // any trace with more spans becomes invalid
+
+  implicit val monoid:Monoid[Trace] = new Monoid[Trace] {
+    def plus(l: Trace, r: Trace) = {
+      for {
+        lId <- l.s.headOption.map(_.traceId)
+        rId <- r.s.headOption.map(_.traceId)
+        if (l != Trace.invalid && r != Trace.invalid && lId != rId)
+      } throw new IllegalArgumentException("Trace Ids must match")
```
franklinhu, line 47: This is a bit confusing and conflates a few conditions. Maybe split out the invalid trace condition separately and have the for comprehension only check for ID matching?

```
if (!l.isValid || !r.isValid) {
  // throw an exception
}
```

zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
```diff
@@ -31,12 +31,44 @@ case class Timespan(start: Long, end: Long)
...
+        else
+          new Trace(newspans.toSeq)
+      }
+    }
+
+    val zero = Trace.zero
+  }
 }
 
 case class Trace(private val s: Seq[Span]) {
```
franklinhu, line 70: I'm starting to think it may make sense to split this into two classes since conceptually there's a "raw" trace with an unmerged list of spans and a merged/finalized trace and the monoid only operates on the former.

zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
```diff
@@ -31,6 +32,34 @@ object Span {
 
   def apply(span: Span): Span = Span(span.traceId, span.name, span.id,
```
ghost: The parameter should say gen.Span, otherwise the apply looks quite odd.

zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
```diff
@@ -31,6 +32,34 @@ object Span {
...
     span.parentId, span.annotations, span.binaryAnnotations, span.debug)
+
+  val invalid = new Span(0, "invalid", 0, None, Nil, Nil, true)
+  val zero = new Span(0, "zero", 0, None, Nil, Nil, true)
+
+  implicit val sg:Monoid[Span] = new Monoid[Span] {
+    def plus(l: Span, r:Span) = {
+      if (l == zero || r == invalid) r
```
ghost, line 38: I think you want
if (l == invalid || r == invalid ) invalid else {...other stuff }

zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
```diff
@@ -31,6 +32,34 @@ object Span {
...
+        // duplicates, the span becomes invalid and we just ignore any further entries.  This is useful
+        // for filtering out rogue clients logging bogus data into the system.
+        val duplicates = for {
+          annotation <- Seq("cs", "cr", "ss", "sr")
+          lann <- l.annotations
+          rann <- r.annotations
+          if (lann.value == rann.value && rann.value == annotation)
+        } yield annotation
+
+        if (duplicates.isEmpty)
```
ghost, line 51: If the conditional is empty vs non-empty, perhaps we can check sizes of the incoming annotations instead of iterating through all of them in a forcomp, making a long sequence & then checking if its empty or not. Should probably define Seq("cs", "cr", "ss", "sr") as a private val inside object Span.

zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
```diff
@@ -31,12 +31,44 @@ case class Timespan(start: Long, end: Long)
...
+      for {
+        lId <- l.s.headOption.map(_.traceId)
+        rId <- r.s.headOption.map(_.traceId)
+        if (l != Trace.invalid && r != Trace.invalid && lId != rId)
+      } throw new IllegalArgumentException("Trace Ids must match")
+
+      if (l.s.size + r.s.size > MAX_SPANS)
+        Trace.invalid
+      else {
+        // merge span lists by combining spans with matching ids
```
ghost, line 52: I would check for invalid span in the trace ( I presume a trace is a sequence of spans, and if any one of those spans is invalid, then the trace addition is invalid as well ?) before proceeding with the addition.

zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
johnynek, line 34: in fact, zero is a method that is sometimes overridden with val.

zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
johnynek: don't think that's the safest approach. Can you find a way to a make a factory method tha generates new instances with the value you want, but once generated they are constant?

zipkin-common/src/main/scala/com/twitter/zipkin/common/Dependencies.scala
bmdhacks, line 39: I prefer to keep the implicit pollution constrained within the object

zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
bmdhacks: actually this method is never used

bmdhacks:
Turns out we don't really need to do this.  It was a good exercise, but not enough benefit.

PR #261, opened 2013-06-17
Added an indicator of duration to the Firefox extension visualizations

IceCreamYou:
As discussed on Friday, added an indicator of how long each request took to process. I did it in the easiest possible way. It would probably be better to adjust all visualizations to have relative scale, so that for example a 1ms request would look small next to a 100ms request, but really doing that right would take a disproportionate amount of time.

Here's what it looks like; hovering over the visualization makes the indicator opaque:

temp


zipkin-browser-extension/firefox/chrome/content/generateOutput.js
```diff
@@ -129,7 +129,10 @@ define([
                                             out += '<div class="zipkin-panel-viz-span-wrapper"><div class="zipkin-panel-viz-span" data-span="' + escapeHTML(s.services.join(', ')) + '" style="margin-left: ' + left + '%; margin-right: ' + (100 - right) + '%;"></div></div>'
                                         }
                                     }
-                                    out += '</td></tr>';
+                                    if (hasTraceData) {
+                                        out += '<div class="zipkin-panel-duration">' + Math.round((max - min) / 1000) + 'ms</div>';
+                                    }
+                                    out += '</div></td></tr>';
                                     if (!hasTraceData) {
```
franklinhu, line 136: I guess this wasn't caught in the initial review, but we should we a templates rather than generating HTML by hand. Check out [Hogan](http://twitter.github.io/hogan.js/)

IceCreamYou, line 136: In principal yes, but in practice because this renders in the browser chrome, there are a bunch of limitations on what JS can do and I believe Hogan would have to be significantly altered to work in that environment. Firefox has its own Chrome DOM templating engine called DOMplate which I found significantly more obtuse and wordy than writing the HTML by hand. That said, it would be nice to find a better solution here.

bmdhacks, line 136: How do other extensions do it?

IceCreamYou, line 136: I don't know if there's a clear convention. In writing this extension I found some that used HTML and some that used Domplate. Domplate isn't the same kind of templating library as Hogan because you write the template as code rather than as HTML with substitutions.

Let's address this in a separate issue: https://github.com/twitter/zipkin/issues/264

bmdhacks:
Works for me, go ahead and merge this.

PR #267, opened 2013-06-27
Add openjdk6 to TravisCI config

franklinhu:


theatrus:
+1

bmdhacks:
not exactly the new contributor I was looking for but I'll take it :-)

bmdhacks:
-1 now the build breaks so you should fix the problems :trollface:

PR #270, opened 2013-06-28
removes the java 1.7 only method

mosesn:


zipkin-collector-scribe/src/main/scala/com/twitter/zipkin/builder/Scribe.scala
```diff
@@ -73,7 +73,12 @@ object Scribe {
     def apply() = (address: InetSocketAddress, statsReceiver: StatsReceiver, timer: Timer) => {
       new OstrichService {
         // make sure we register with zookeeper using a reasonable address
-        var boundAddr = if (address.getHostString == "0.0.0.0" || address.getHostString == "127.0.0.1")
+        def getIP(addr: InetSocketAddress): String = Option(addr.getAddress) match {
```
johanoskarsson, line 76: Remind me again why we have to convert the ip to a string to compare? Would be nice to avoid if possible

mosesn, line 76: I think this is to get around the 0.0.0.0 vs. 127.0.0.1 issue, where sometimes 127.0.0.1 isn't supported, so we just want to rely on java to do it right.  However, I didn't add in the getHostString, so I might be mistaken.

PR #272, opened 2013-07-03
Add support for SQL databases

IceCreamYou:
The goal of this branch is to add support for SQL databases using Anorm from the Play framework.

Note that I started by trying to do this with Slick instead of Anorm, but Slick depends on Scala 2.10 and it was too much scope creep to manage upgrading the whole project to be compatible with 2.10.

This branch makes SQLite the default storage mechanism. There are now no setup steps -- if you have Scala 2.9 installed and you download Zipkin, you can immediately run the daemons. Documentation for getting other SQL databases running is in `doc/sql-databases.md`.

Once this is merged, I'll look at using FuturePools to do the DB operations asynchronously, as well as adding Aggregates support.


mosesn:
What is the fancy new autoloading?  I'm just curious.

project/Project.scala
```diff
@@ -163,6 +163,23 @@ object Zipkin extends Build {
...
   ).dependsOn(scrooge)
 
+  lazy val anormDB = Project(
+    id = "zipkin-anormdb",
+    base = file("zipkin-anormdb"),
+    settings = defaultSettings
+  ).settings(
+    libraryDependencies ++= Seq(
+      "play"        %% "anorm"        % "2.1-09142012",
+      "org.xerial"  %  "sqlite-jdbc"  % "3.7.2"
```
mosesn: what's the point of this dependency?

IceCreamYou: It's the SQLite driver for JDBC. We decided that the MVP would be to get Zipkin running on SQLite, so until we can load config dynamically I just put this in directly. But Anorm supports any SQL database in theory, and the DB class already has settings for MySQL, PostgreSQL, and H2 -- so one of the TODOs is to figure out the easiest way for people who want to use something other than SQLite to get the right driver dependencies loaded.

mosesn: Maybe this should be a test-only dependency.

IceCreamYou: What I meant above is that ideally we figure out an easy way for people to choose their database type and then have Zipkin pull in the required dependency for that database. So you would say you wanted to use SQLite or MySQL instead of Cassandra or whatever, and then you'd load the SQLite or MySQL dependency instead of the Cassandra dependencies. But I haven't looked into how to do that yet or even if there's a straightforward way to do it easily. In the mean time, I'm just including SQLite directly so there is something to test with.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/AnormIndex.scala
```diff
@@ -0,0 +1,215 @@
...
+   */
+  def close() { conn.close() }
+
+  /**
+   * Get the trace ids for this particular service and if provided, span name.
+   * Only return maximum of limit trace ids from before the endTs.
+   */
+  def getTraceIdsByName(serviceName: String, spanName: Option[String],
+                        endTs: Long, limit: Int): Future[Seq[IndexedTraceId]] = {
+    val result:List[(Long, Long)] = SQL(
```
theatrus: These Anorm operations should probably ne mapped into a Finagle FuturePool

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/AnormIndex.scala
```diff
@@ -0,0 +1,215 @@
...
+ *
+ * We're ignoring TTL for now since unlike Cassandra and Redis, SQL databases
+ * don't have that built in and it shouldn't be a big deal for most sites.
+ *
+ * The index methods are stubs since SQL databases don't use NoSQL-style
+ * indexing.
+ */
+case class AnormIndex() extends Index {
+  // Database connection object
+  private implicit val conn = DB.getConnection()
```
franklinhu: Is `DB` a singleton? Would be better to pass it into the `AnormIndex` as an argument.

Edit: After looking at the rest, it's probably best to pass in the connection as a parameter.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/AnormIndex.scala
```diff
@@ -0,0 +1,215 @@
...
+        |ORDER BY timestamp DESC
+        |LIMIT {limit}
+      """.stripMargin)
+      .on("service_name" -> serviceName)
+      .on("span_name" -> (if (spanName.isEmpty) "" else spanName.get))
+      .on("end_ts" -> endTs)
+      .on("limit" -> limit)
+      .as((long("trace_id") ~ long("end_ts") map flatten) *)
+    Future(result map { row =>
+      IndexedTraceId(traceId = row._1, timestamp = row._2)
```
franklinhu: Instead of `row` you can do something like

```
result map { case (tId, ts) => ... }
```

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/AnormStorage.scala
```diff
@@ -0,0 +1,220 @@
...
+  def setTimeToLive(traceId: Long, ttl: Duration): Future[Unit] = {
+    Future.Unit
+  }
+
+  /**
+   * Get the time to live for a specific trace.
+   * If there are multiple ttl entries for one trace, pick the lowest one.
+   */
+  def getTimeToLive(traceId: Long): Future[Duration] = {
+    Future(Duration.Top)
```
franklinhu: You likely want `Future.value(...)`

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/AnormStorage.scala
```diff
@@ -0,0 +1,220 @@
...
+   * in that span. First event should be first in the spans list.
+   */
+  def getSpansByTraceIds(traceIds: Seq[Long]): Future[Seq[Seq[Span]]] = {
+    val traceIdsString:String = traceIds.mkString(",")
+    /*
+     * No need to pull duration (Option[Long]) or created_ts (Option[Long])
+     * because they're derived from the annotations and just stored in the spans
+     * table for straightforward lookups.
+     */
+    val spans:List[(Long, Option[Long], Long, String, Boolean)] =
```
franklinhu: The type params for the inner tuple are a bit opaque; should use a case class for better readability.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/AnormStorage.scala
```diff
@@ -0,0 +1,220 @@
...
+      SQL(
+        """SELECT span_id, parent_id, trace_id, span_name, debug
+          |FROM zipkin_spans
+          |WHERE trace_id IN (%s)
+        """.stripMargin.format(traceIdsString))
+        .as((
+          long("span_id") ~ get[Option[Long]]("parent_id") ~ long("trace_id") ~
+            str("span_name") ~ bool("debug") map flatten) *)
+    val annos:List[(Long, Long, String, String, String, Option[Int], Option[Int], Long, Option[Long])] =
```
franklinhu: same here

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/AnormStorage.scala
```diff
@@ -0,0 +1,220 @@
...
+        """SELECT span_id, trace_id, span_name, service_name, value, ipv4,
+          |  port, timestamp, duration
+          |FROM zipkin_annotations
+          |WHERE trace_id IN (%s)
+        """.stripMargin.format(traceIdsString))
+        .as((long("span_id") ~ long("trace_id") ~ str("span_name") ~
+          str("service_name") ~ str("value") ~ get[Option[Int]]("ipv4") ~
+          get[Option[Int]]("port") ~ long("timestamp") ~
+          get[Option[Long]]("duration") map flatten) *)
+    val binAnnos:List[(Long, Long, String, String, String, Array[Byte], Int, Option[Int], Option[Int])] =
```
franklinhu: and here

zipkin-common/src/main/scala/com/twitter/zipkin/common/AnnotationType.scala
```diff
@@ -24,6 +24,27 @@ object AnnotationType {
...
+    case Bool.value   => Bool
+    case Bytes.value  => Bytes
+    case I16.value    => I16
+    case I32.value    => I32
+    case I64.value    => I64
+    case Double.value => Double
+    case String.value => String
+    case _            => String /* Uh... */
+  }
+  def fromString(v:String) = v match {
```
bmdhacks: fromString is not used anywhere

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/DB.scala
```diff
@@ -0,0 +1,278 @@
...
+      "driver" -> "com.mysql.jdbc.Driver"
+    )
+  )
+
+  /**
+   * Configuration variables.
+   */
+  private val dbconfig = dbc getOrElse {
+    // TODO: How do we know this file is actually at the correct location?
+    val f = new File("../../../../../../../../config/dbconfig.scala")
```
bmdhacks: From my reading, this can just be cut out without any horrible problem.

zipkin-anormdb/config/dbconfig.scala
```diff
@@ -0,0 +1,37 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import com.twitter.zipkin.storage.anormdb.DB
+
+Map(
```
bmdhacks: Can you remove this file and move it to the collector config

bmdhacks:
Also, how hard is it to add Aggregates?

bmdhacks:
Lastly, please write some unit tests.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/DB.scala
```diff
@@ -0,0 +1,278 @@
...
+   * interface to the database type.
+   *
+   * Anorm supports any SQL database, so more databases can be added here.
+   *
+   * The other place the database driver needs to be set is in the project
+   * dependencies in project/Project.scala.
+   *
+   * TODO: Figure out the easiest way to get these dependencies loaded for anything other than SQLite
+   */
+  private val dbmap = Map(
```
bmdhacks: This looks structured enough to be a map of case classes, instead of a map of maps.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/DB.scala
```diff
@@ -0,0 +1,278 @@
...
+  )
+
+  /**
+   * Configuration variables.
+   */
+  private val dbconfig = dbc getOrElse {
+    // TODO: How do we know this file is actually at the correct location?
+    val f = new File("../../../../../../../../config/dbconfig.scala")
+    if (f.exists) Eval[Map[String, Map[String, String]]](f)
+    else Map(
```
bmdhacks: It is preferable to move structure to case classes instead of arbitrary maps of key/values.  If configuration is variable between drivers, it's OK, but otherwise let's try to make this as structured as possible.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/DB.scala
```diff
@@ -0,0 +1,188 @@
...
+   *   // Do database updates
+   * }
+   *
+   * This is useful when you only need a connection open for a single block of
+   * code. Because it creates a new connection, if you need to make a lot of
+   * updates at different times or locations, it may be better to hold open a
+   * single connection using getConnection() as long as you remember to
+   * manually close it.
+   */
+  def withConnection[A](block: Connection => A) : A = {
```
bmdhacks: This method, while cool, is not used anywhere in the code.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/AnormIndex.scala
```diff
@@ -0,0 +1,226 @@
...
+   * Get the trace ids for this annotation before endTs. If value is also
+   * passed we expect both the annotation key and value to be present in index
+   * for a match to be returned.
+   * Only return maximum of limit trace ids from before the endTs.
+   */
+  def getTraceIdsByAnnotation(serviceName: String, annotation: String, value: Option[ByteBuffer],
+                              endTs: Long, limit: Int): Future[Seq[IndexedTraceId]] = {
+    // Ignore core annotations. Yay magic names!
+    if (List("cs", "cr", "ss", "sr", "ca", "sa").contains(annotation))
+      return Future.value[Seq[IndexedTraceId]](Seq());
```
bmdhacks: Semicolon.
Also return is gauche, but I don't mind its use here.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/DB.scala
```diff
@@ -0,0 +1,188 @@
...
+        |  key VARCHAR(255) NOT NULL,
+        |  value %s,
+        |  annotation_type_value INT NOT NULL,
+        |  ipv4 INT,
+        |  port INT
+        |)
+      """.stripMargin.format(this.getBlobType)).execute()
+    //SQL("CREATE INDEX trace_id ON zipkin_binary_annotations (trace_id)").execute()
+    if (!keepAlive) con.close() else ()
+    con
```
bmdhacks: You don't need the else here.
Also, I'm not super stoked on returning a closed connection.
Can you research just returning the collection and getting it to auto-close on garbage-collection, thus obviating the need for keepAlive

bmdhacks: Actually, this method is never called with keepAlive=false, so even if you can't figure out how to close the handle on GC it's OK.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/DB.scala
```diff
@@ -0,0 +1,188 @@
...
+
+import anorm._
+import anorm.SqlParser._
+import java.sql.{Blob, Connection, DriverManager}
+
+/**
+ * The companion object is mainly to make installation easier.
+ */
+object DB {
+  def apply() = {
```
bmdhacks: I don't think this method is needed since the case class should provide a default constructor already.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/DB.scala
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  // Get the column the current database type uses for BLOBs.
+  private def getBlobType = dbconfig.description match {
+    case "PostgreSQL" => "BYTEA" /* As usual PostgreSQL has to be different */
+    case "MySQL" => "MEDIUMBLOB" /* MySQL has length limits, in this case 16MB */
+    case _ => "BLOB"
+  }
+
+  // Provide Anorm with the ability to handle BLOBs.
```
bmdhacks: What part of the code does this comment refer to?  Seems like above, but maybe below.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/DB.scala
```diff
@@ -0,0 +1,188 @@
...
+  private def valueToByteArrayOption(value: Any): Option[Array[Byte]] = {
+    try {
+      value match {
+        case bytes: Array[Byte] => Some(bytes)
+        case blob: Blob => Some(blob.getBytes(1, blob.length.asInstanceOf[Int]))
+        case _ => None
+      }
+    }
+    catch {
+      case e: Exception => None
```
bmdhacks: I'd just catch SQLException to be clearer about what's going on here.  Maybe even push the try up into the 'case blob' bit.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/DB.scala
```diff
@@ -0,0 +1,188 @@
...
+        case _ => Left(TypeDoesNotMatch("Cannot convert " + value + ":" + value.asInstanceOf[AnyRef].getClass + " to Byte Array for column " + qualified))
+      }
+    }
+  }
+
+  /**
+   * Build a RowParser factory for a byte array column.
+   */
+  def bytes(columnName: String): RowParser[Array[Byte]] = {
+    get[Array[Byte]](columnName)(implicitly[Column[Array[Byte]]])
```
bmdhacks: you don't need to make rowToByteArray implicit.  Cut out the implicit and call get like this:

``` scala
  def bytes(columnName: String): RowParser[Array[Byte]] = {
    get[Array[Byte]](columnName)(rowToByteArray)
  }
```

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/AnormIndex.scala
```diff
@@ -0,0 +1,226 @@
...
+        |GROUP BY trace_id
+        |ORDER BY a_timestamp DESC
+        |LIMIT {limit}
+      """.stripMargin)
+      .on("service_name" -> serviceName)
+      .on("span_name" -> (if (spanName.isEmpty) "" else spanName.get))
+      .on("end_ts" -> endTs)
+      .on("limit" -> limit)
+      .as((long("trace_id") ~ long("MAX(a_timestamp)") map flatten) *)
+    Future(result map { case (tId, ts) =>
```
bmdhacks: This applies to all usages of this, but it's more idiomatic in scala to supply closure arguments (especially multiline closures) via curly braces:  

``` scala
    Future {
      result map { case (tId, ts) =>
        IndexedTraceId(traceId = tId, timestamp = ts)
      }
    }
```

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/AnormIndex.scala
```diff
@@ -0,0 +1,226 @@
...
+            |LIMIT {limit}
+          """.stripMargin)
+          .on("service_name" -> serviceName)
+          .on("annotation" -> annotation)
+          .on("end_ts" -> endTs)
+          .on("limit" -> limit)
+          .as((long("trace_id") ~ long("MAX(a_timestamp)") map flatten) *)
+      }
+    }
+    Future(result map { case (tId, ts) =>
```
bmdhacks: Curly braces etc. etc.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/DB.scala
```diff
@@ -0,0 +1,188 @@
...
+        |  created_ts BIGINT
+        |)
+      """.stripMargin).execute()
+    //SQL("CREATE INDEX trace_id ON zipkin_spans (trace_id)").execute()
+    SQL(
+      """CREATE TABLE IF NOT EXISTS zipkin_annotations (
+        |  span_id BIGINT NOT NULL,
+        |  trace_id BIGINT NOT NULL,
+        |  span_name VARCHAR(255) NOT NULL,
+        |  service_name VARCHAR(255) NOT NULL,
```
bmdhacks: Why store span_name in the annotation?
Also, service_name is technically nullable isn't it?  The Endpoint structure on the Annotation is optional.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/DB.scala
```diff
@@ -0,0 +1,188 @@
...
+        |  a_timestamp BIGINT NOT NULL,
+        |  duration BIGINT
+        |)
+      """.stripMargin).execute()
+    //SQL("CREATE INDEX trace_id ON zipkin_annotations (trace_id)").execute()
+    SQL(
+      """CREATE TABLE IF NOT EXISTS zipkin_binary_annotations (
+        |  span_id BIGINT NOT NULL,
+        |  trace_id BIGINT NOT NULL,
+        |  span_name VARCHAR(255) NOT NULL,
```
bmdhacks: Same issue here as above. re: span_name and service_name

zipkin-anormdb/src/main/scala/com/twitter/zipkin/util/Util.scala
```diff
@@ -0,0 +1,43 @@
...
+import java.nio.ByteBuffer
+import java.util.Arrays
+
+object Util {
+
+  /**
+   * Get the contents of a ByteBuffer as an Array
+   * Stolen from http://svn.apache.org/viewvc/cassandra/trunk/src/java/org/apache/cassandra/utils/ByteBufferUtil.java?revision=1201726&view=markup
+   */
+  def getArrayFromBuffer(buf: ByteBuffer): Array[Byte] = {
```
bmdhacks: Please no copy/paste.  Bubble this up into zipkin-common if needed.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/DB.scala
IceCreamYou: span_name is in the annotation to make getTraceIdsByName() and getSpanNames() in AnormIndex easy.

Right now we use Annotation#serviceName() to get the service name that is stored in the database. That returns "Unknown service name" if there is no service name.

bmdhacks: Since this is SQL, I'm more inclined to just do subselects instead of denormalizing the data.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/AnormIndex.scala
```diff
@@ -0,0 +1,232 @@
...
+        TraceIdDuration(traceId, duration.getOrElse(0), startTs)
+      }
+    }
+  }
+
+  /**
+   * Get all the service names.
+   */
+  def getServiceNames: Future[Set[String]] = {
+    Future(SQL(
```
bmdhacks: Future {
etc... couple other places to get below this.

mosesn:
The path for storage.anormdb should instead be storage/anormdb.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage.anormdb/DB.scala
mosesn: I can't tell how this ended up being fixed, but I think it's generally a little dirty to use [finalizers](http://stackoverflow.com/questions/2506488/java-finalize-method-call).  There's probably a better way to do it, maybe by using [scala-arm](https://github.com/jsuereth/scala-arm) or something like that.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/DB.scala
```diff
@@ -0,0 +1,146 @@
...
+          case e: SQLException => None
+        }
+      case _ => None
+    }
+  }
+
+  /**
+   * Implicitly convert an Anorm row to a byte array.
+   */
+  implicit def rowToByteArray: Column[Array[Byte]] = {
```
bmdhacks: Delete the `implicit` keyword here and it will still work just fine.

zipkin-anormdb/src/test/scala/com/twitter/zipkin/storage/anormdb/AnormDBSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+        ("debug", "SMALLINT", 1),
+        ("duration", "BIGINT", 0),
+        ("created_ts", "BIGINT", 0)
+      )
+      // The right columns are present
+      val cols_spans: List[(Int, String, String, Int, Option[String], Int)] =
+        SQL("PRAGMA table_info(zipkin_spans)").as((
+          int("cid") ~ str("name") ~ str("type") ~ int("notnull") ~
+            get[Option[String]]("dflt_value") ~ int("pk") map flatten) *)
+      val better_cols = cols_spans.map(col => (col._2, col._3, col._4))
```
bmdhacks: The ( police is back!  cols_spans.map { col => ...

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormStorage.scala
```diff
@@ -0,0 +1,236 @@
...
+          |WHERE trace_id IN (%s)
+        """.stripMargin.format(traceIdsString))
+        .as((long("span_id") ~ long("trace_id") ~ str("service_name") ~ str("key") ~
+          db.bytes("value") ~ int("annotation_type_value") ~
+          get[Option[Int]]("ipv4") ~ get[Option[Int]]("port") map {
+            case a~b~c~d~e~f~g~h => DBBinaryAnnotation(a, b, c, d, e, f, g, h)
+          }) *)
+
+    val results: Seq[Seq[Span]] = traceIds.map(traceId =>
+      spans.filter(_.traceId == traceId).map({span =>
```
bmdhacks: This statement and the one above it.
.map {

bmdhacks: and below it

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormStorage.scala
```diff
@@ -0,0 +1,236 @@
...
+            case (Some(ipv4), Some(port)) => Some(Endpoint(ipv4, port.toShort, anno.serviceName))
+            case _ => None
+          }
+          val duration:Option[Duration] = anno.duration match {
+            case Some(nanos) => Some(Duration.fromNanoseconds(nanos))
+            case None => None
+          }
+          Annotation(anno.timestamp, anno.value, host, duration)
+        })
+        val spanBinAnnos = binAnnos.filter(_.traceId == span.traceId).map({binAnno =>
```
bmdhacks: and here.  Oh, and this is a superduper nitpick, but Anno means year in spanish and keeps confusing me.  Totally ok if you ignore me.  I'm just typing now for no reason.  Seriously I need top stop this isn't helpful.

PR #274, opened 2013-07-15
HBase Storage

elliottneilclark:
Here's the start of an HBase storage module.  It's still in pretty early form but I wanted to get this up so that reviews could start early.
## Tables
- zipkin.traces
- zipkin.duration
- zipkin.idxService
- zipkin.idxServiceSpanName
- zipkin.idxServiceAnnotation
- zipkin.topAnnotations
- zipkin.dependencies
- zipkin.mappings
- zipkin.idGen
## Internals

I created a non-blocking futures based version of HBase's HTable.  This HBaseTable uses a thread pool to execute the calls to HBase.

Currently it's using HBase 0.94.9 and hadoop 1
## Row Keys Construction

Whenever possible I am using a fixed length row key.  That necessitates a set id for any variable length strings that can be on the row key.  This mapping is created using the mapping table and the id gen table. The idGen table holds the last idGenerated.
- There can be up to Long.MaxValue services
- Each service can have Long.MaxValue spanNames and annotations

The mapping table holds the mapping of Array[Byte] to id and the reverse.
## Still ToDo
- counters, counters, more counters
- Allow this to build against current hbase trunk

### DIFF
--- a/project/Project.scala
+++ b/project/Project.scala
@@ -12,6 +12,7 @@ object Zipkin extends Build {
   val SCROOGE_VERSION = "3.3.2"
   val ZOOKEEPER_VERSION = Map("candidate" -> "0.0.41", "group" -> "0.0.44", "client" -> "0.0.35")
   val ALGEBIRD_VERSION  = "0.1.13"
+  val HBASE_VERSION = "0.94.10"
 
   val proxyRepo = Option(System.getenv("SBT_PROXY_REPO"))
   val travisCi = Option(System.getenv("SBT_TRAVIS_CI")) // for adding travis ci maven repos before others
@@ -75,7 +76,7 @@ object Zipkin extends Build {
     Project(
       id = "zipkin",
       base = file(".")
-    ) aggregate(test, queryCore, queryService, common, scrooge, collectorScribe, web, cassandra, anormDB, collectorCore, collectorService, kafka, redis)
+    ) aggregate(test, queryCore, queryService, common, scrooge, collectorScribe, web, cassandra, anormDB, collectorCore, collectorService, kafka, redis, hbase)
 
   lazy val test   = Project(
     id = "zipkin-test",
@@ -99,7 +100,7 @@ object Zipkin extends Build {
         "com.twitter" %% "finagle-exception" % FINAGLE_VERSION,
         "com.twitter" %% "ostrich"           % OSTRICH_VERSION,
         "com.twitter" %% "util-core"         % UTIL_VERSION,
-        "com.twitter" %% "algebird-core"    % ALGEBIRD_VERSION,
+        "com.twitter" %% "algebird-core"     % ALGEBIRD_VERSION,
 
         "com.twitter.common.zookeeper" % "client"    % ZOOKEEPER_VERSION("client")
       ) ++ testDependencies
@@ -144,7 +145,7 @@ object Zipkin extends Build {
       "com.twitter" %% "finagle-thrift"    % FINAGLE_VERSION,
       "com.twitter" %% "finagle-zipkin"    % FINAGLE_VERSION,
       "com.twitter" %% "ostrich"           % OSTRICH_VERSION,
-      "com.twitter" %% "algebird-core"    % ALGEBIRD_VERSION,
+      "com.twitter" %% "algebird-core"     % ALGEBIRD_VERSION,
       "com.twitter" %% "util-core"         % UTIL_VERSION,
       "com.twitter" %% "util-zk"           % UTIL_VERSION,
       "com.twitter" %% "util-zk-common"    % UTIL_VERSION,
@@ -202,7 +203,7 @@ object Zipkin extends Build {
         "com.twitter" %% "finagle-thrift"    % FINAGLE_VERSION,
         "com.twitter" %% "finagle-zipkin"    % FINAGLE_VERSION,
         "com.twitter" %% "ostrich"           % OSTRICH_VERSION,
-        "com.twitter" %% "algebird-core"    % ALGEBIRD_VERSION,
+        "com.twitter" %% "algebird-core"     % ALGEBIRD_VERSION,
         "com.twitter" %% "util-core"         % UTIL_VERSION,
         "com.twitter" %% "util-zk"           % UTIL_VERSION,
         "com.twitter" %% "util-zk-common"    % UTIL_VERSION,
@@ -228,7 +229,7 @@ object Zipkin extends Build {
       base =>
         (base / "config" +++ base / "src" / "test" / "resources").get
     }
-  ).dependsOn(queryCore, cassandra, redis, anormDB)
+  ).dependsOn(queryCore, cassandra, redis, anormDB, hbase)
 
   lazy val collectorScribe =
     Project(
@@ -271,7 +272,7 @@ object Zipkin extends Build {
       base =>
         (base / "config" +++ base / "src" / "test" / "resources").get
     }
-  ).dependsOn(collectorCore, collectorScribe, cassandra, kafka, redis, anormDB)
+  ).dependsOn(collectorCore, collectorScribe, cassandra, kafka, redis, anormDB, hbase)
 
   lazy val web =
     Project(
@@ -311,6 +312,32 @@ object Zipkin extends Build {
       "com.twitter" %% "finagle-redis"     % FINAGLE_VERSION,
       "org.slf4j" % "slf4j-log4j12"          % "1.6.4" % "runtime",
       "com.twitter"     %% "util-logging"      % UTIL_VERSION
+    ) ++ testDependencies,
+
+    /* Add configs to resource path for ConfigSpec */
+    unmanagedResourceDirectories in Test <<= baseDirectory {
+      base =>
+        (base / "config" +++ base / "src" / "test" / "resources").get
+    }
+  ).dependsOn(scrooge)
+
+  lazy val hbase = Project(
+    id = "zipkin-hbase",
+    base = file("zipkin-hbase"),
+    settings = defaultSettings
+  ).settings(
+    parallelExecution in Test := false,
+    libraryDependencies ++= Seq(
+      "org.apache.hbase"      % "hbase"                 % HBASE_VERSION notTransitive(),
+      "org.apache.hbase"      % "hbase"                 % HBASE_VERSION % "test" classifier("tests") classifier(""),
+      "com.google.protobuf"   % "protobuf-java"         % "2.4.1",
+      "org.apache.hadoop"     % "hadoop-core"           % "1.1.2" notTransitive(),
+      "org.apache.hadoop"     % "hadoop-test"           % "1.1.2" % "test",
+      "commons-logging"       % "commons-logging"       % "1.1.1",
+      "commons-configuration" % "commons-configuration" % "1.6",
+      "org.apache.zookeeper"  % "zookeeper"             % "3.4.5" % "runtime" notTransitive(),
+      "org.slf4j"             % "slf4j-log4j12"         % "1.6.4" % "runtime",
+      "com.twitter"           % "util-logging"          % UTIL_VERSION
     ) ++ testDependencies,
 
     /* Add configs to resource path for ConfigSpec */
--- a/zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseAggregates.scala
+++ b/zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseAggregates.scala
@@ -1,0 +1,117 @@
+package com.twitter.zipkin.storage.hbase
+
+import com.twitter.algebird.Monoid
+import com.twitter.scrooge.BinaryThriftStructSerializer
+import com.twitter.util.{Time, Future}
+import com.twitter.zipkin.common.Dependencies
+import com.twitter.zipkin.conversions.thrift._
+import com.twitter.zipkin.gen
+import com.twitter.zipkin.hbase.TableLayouts
+import com.twitter.zipkin.storage.Aggregates
+import com.twitter.zipkin.storage.hbase.mapping.ServiceMapper
+import com.twitter.zipkin.storage.hbase.utils.{HBaseTable, IDGenerator}
+import org.apache.hadoop.hbase.client.{Scan, Put}
+import org.apache.hadoop.hbase.util.Bytes
+import scala.collection.JavaConverters._
+
+trait HBaseAggregates extends Aggregates {
+
+  val dependenciesTable: HBaseTable
+  val topAnnotationsTable: HBaseTable
+
+  val mappingTable: HBaseTable
+  val idGenTable: HBaseTable
+  lazy val idGen = new IDGenerator(idGenTable)
+  lazy val serviceMapper = new ServiceMapper(mappingTable, idGen)
+
+  val serializer = new BinaryThriftStructSerializer[gen.Dependencies] {
+    def codec = gen.Dependencies
+  }
+
+  def close() {
+    mappingTable.close()
+    idGenTable.close()
+
+    topAnnotationsTable.close()
+    dependenciesTable.close()
+  }
+
+  def getDependencies(startDate: Option[Time], endDate: Option[Time]=None): Future[Dependencies] = {
+    val scan = new Scan()
+    scan.setStartRow(Bytes.toBytes(Long.MaxValue - startDate.map(_.inMilliseconds).getOrElse(Long.MaxValue)))
+    endDate.foreach { ed => scan.setStopRow(Bytes.toBytes(Long.MaxValue - ed.inMilliseconds))}
+    scan.addColumn(TableLayouts.dependenciesFamily, Bytes.toBytes("\0"))
+    dependenciesTable.scan(scan, 100).map { results =>
+      val depList = results.flatMap { result =>
+        result.list().asScala.headOption.map { kv =>
+          val tDep = serializer.fromBytes(kv.getValue)
+          tDep.toDependencies
+        }
+      }
+      Monoid.sum(depList)
+    }
+  }
+
+  def storeDependencies(dependencies: Dependencies): Future[Unit] = {
+    val rk = Bytes.toBytes(Long.MaxValue - dependencies.startTime.inMilliseconds)
+    val put = new Put(rk)
+    put.add(TableLayouts.dependenciesFamily, Bytes.toBytes("\0"), serializer.toBytes(dependencies.toThrift))
+    dependenciesTable.put(Seq(put))
+  }
+
+  def getTopAnnotations(serviceName: String): Future[Seq[String]] = {
+    getTopAnnotationInternal(serviceName, TableLayouts.topAnnotationFamily)
+  }
+
+  def getTopKeyValueAnnotations(serviceName: String): Future[Seq[String]] = {
+    getTopAnnotationInternal(serviceName, TableLayouts.topAnnotationKeyValueFamily)
+  }
+
+  def storeTopAnnotations(serviceName: String, annotations: Seq[String]): Future[Unit] = {
+    storeTopAnnotationsInternal(serviceName, annotations, TableLayouts.topAnnotationFamily)
+  }
+
+  def storeTopKeyValueAnnotations(serviceName: String, a: Seq[String]): Future[Unit] = {
+    storeTopAnnotationsInternal(serviceName, a, TableLayouts.topAnnotationKeyValueFamily)
+  }
+
+  private def getTopAnnotationInternal(serviceName: String, fam: Array[Byte]): Future[Seq[String]] = {
+    serviceMapper.get(serviceName).map { serviceMapping =>
+      val startRk = Bytes.toBytes(serviceMapping.id) ++ Bytes.toBytes(0L)
+      val scan = new Scan()
+      scan.addFamily(fam)
+      scan.setCaching(1)
+
+      // This is a very small table and there will
+      // be row key skew.  To get away with this the table
+      // needs to stay in memory as much as possible.
+      scan.setCacheBlocks(true)
+
+      scan.setStartRow(startRk)
+      scan
+    }.flatMap { scan =>
+      val results = topAnnotationsTable.scan(scan, 1)
+      results
+    }.map { results =>
+      val resultOption = results.headOption
+      resultOption.map { result =>
+        result.list().asScala.map { kv =>
+          (Bytes.toInt(kv.getQualifier), Bytes.toString(kv.getValue))
+        }.sortBy {_._1}.map {_._2}
+      }.getOrElse(Seq.empty[String])
+    }
+  }
+
+  private def storeTopAnnotationsInternal(serviceName: String, annotations: Seq[String], fam: Array[Byte]): Future[Unit] = {
+    serviceMapper.get(serviceName).map { serviceMapping =>
+      val put = new Put(Bytes.toBytes(serviceMapping.id) ++ Bytes.toBytes(Long.MaxValue - System.currentTimeMillis()))
+      annotations.zipWithIndex.map { case (anno: String, id: Int)=>
+        put.add(TableLayouts.topAnnotationFamily, Bytes.toBytes(id), Bytes.toBytes(anno))
+      }
+      put
+    }.flatMap { put =>
+      topAnnotationsTable.put(Seq(put))
+    }
+  }
+
+}
--- a/zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseIndex.scala
+++ b/zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseIndex.scala
@@ -1,0 +1,295 @@
+package com.twitter.zipkin.storage.hbase
+
+import com.twitter.logging.Logger
+import com.twitter.util.Future
+import com.twitter.zipkin.Constants
+import com.twitter.zipkin.common.Span
+import com.twitter.zipkin.hbase.TableLayouts
+import com.twitter.zipkin.storage.hbase.mapping.ServiceMapper
+import com.twitter.zipkin.storage.hbase.utils.{HBaseTable, IDGenerator}
+import com.twitter.zipkin.storage.{IndexedTraceId, TraceIdDuration, Index}
+import java.nio.ByteBuffer
+import org.apache.hadoop.hbase.client.{Get, Result, Scan, Put}
+import org.apache.hadoop.hbase.filter.CompareFilter.CompareOp
+import org.apache.hadoop.hbase.filter.{BinaryComparator, ValueFilter}
+import org.apache.hadoop.hbase.util.Bytes
+import scala.collection.JavaConverters._
+import scala.collection.Set
+import com.twitter.zipkin.util.Util
+
+trait HBaseIndex extends Index {
+  val log = Logger.get(getClass.getName)
+
+  val durationTable: HBaseTable
+  val idxServiceTable: HBaseTable
+  val idxServiceSpanNameTable: HBaseTable
+  val idxServiceAnnotationTable: HBaseTable
+
+  val mappingTable: HBaseTable
+  val idGenTable: HBaseTable
+  lazy val idGen = new IDGenerator(idGenTable)
+  lazy val serviceMapper    = new ServiceMapper(mappingTable, idGen)
+
+  /**
+   * Close the index
+   */
+  def close() {
+    idGenTable.close()
+    mappingTable.close()
+
+    //ttl tables.
+    durationTable.close()
+    idxServiceTable.close()
+    idxServiceAnnotationTable.close()
+    idxServiceSpanNameTable.close()
+  }
+
+
+  /**
+   * Get the trace ids for this particular service and if provided, span name.
+   * Only return maximum of limit trace ids from before the endTs.
+   */
+  def getTraceIdsByName(serviceName: String, spanNameOption: Option[String], endTs: Long, limit: Int): Future[Seq[IndexedTraceId]] = {
+    val resultsFuture = spanNameOption match {
+      case None       => getTraceIdsByNameNoSpanName(serviceName, endTs, limit)
+      case Some(spanName) => getTraceIdsByNameWithSpanName(serviceName, spanName, endTs, limit)
+    }
+
+    resultsFuture.map { results =>
+      results.flatMap { result => indexResultToTraceId(result) }.toSeq.distinct.take(limit)
+    }
+  }
+
+  /**
+   * Get the trace ids for this annotation between the two timestamps. If value is also passed we expect
+   * both the annotation key and value to be present in index for a match to be returned.
+   * Only return maximum of limit trace ids from before the endTs.
+   */
+  def getTraceIdsByAnnotation(serviceName: String, annotation: String, value: Option[ByteBuffer], endTs: Long, limit: Int): Future[Seq[IndexedTraceId]] = {
+    val serviceMappingFuture = serviceMapper.get(serviceName)
+    val annoMappingFuture = serviceMappingFuture.flatMap { serviceMapping =>
+      serviceMapping.annotationMapper.get(annotation)
+    }
+
+    annoMappingFuture.flatMap { annoMapping =>
+      val scan = new Scan()
+      val startRk = Bytes.toBytes(annoMapping.parent.get.id) ++ Bytes.toBytes(annoMapping.id) ++ Bytes.toBytes(0L)
+      val endRk = Bytes.toBytes(annoMapping.parent.get.id) ++ Bytes.toBytes(annoMapping.id) ++ getEndScanTimeStampRowKeyBytes(endTs)
+      scan.setStartRow(startRk)
+      scan.setStopRow(endRk)
+      scan.addFamily(TableLayouts.idxAnnotationFamily)
+      value.foreach { bb => scan.setFilter(new ValueFilter(CompareOp.EQUAL, new BinaryComparator(bb.array()))) }
+      idxServiceAnnotationTable.scan(scan, limit).map { results =>
+        results.flatMap { result => indexResultToTraceId(result)}.toSeq.distinct.take(limit)
+      }
+    }
+  }
+
+  /**
+   * Fetch the duration or an estimate thereof from the traces.
+   * Duration returned in micro seconds.
+   */
+  def getTracesDuration(traceIds: Seq[Long]): Future[Seq[TraceIdDuration]] = {
+    val gets = traceIds.map { traceId =>
+      val get = new Get(Bytes.toBytes(traceId))
+      get.setMaxVersions(1)
+    }
+    // Go to hbase to get all of the durations.
+    durationTable.get(gets).map { results => results.map(durationResultToDuration).toSet.toSeq  }
+  }
+
+  /**
+   * Get all the service names for as far back as the ttl allows.
+   */
+  def getServiceNames: Future[Set[String]] = serviceMapper.getAll.map { f => f.map(_.name) }
+
+  /**
+   * Get all the span names for a particular service, as far back as the ttl allows.
+   */
+  def getSpanNames(service: String): Future[Set[String]] = {
+    // From the service get the spanNameMapper.  Then get all the maps.
+    val spanNameMappingsFuture = serviceMapper.get(service).flatMap { _.spanNameMapper.getAll }
+    // get the names from the mappings.
+    spanNameMappingsFuture.map { maps => maps.map { _.name} }
+  }
+
+  /**
+   * Index a trace id on the service and name of a specific Span
+   */
+  def indexTraceIdByServiceAndName(span: Span): Future[Unit] = {
+    // Get the id of services and span names
+    val serviceMappingsFuture =  Future.collect( span.serviceNames.map { sn =>
+      serviceMapper.get(sn)
+    }.toSeq)
+
+    // Figure out when this happened.
+    val timeBytes = getTimeStampRowKeyBytes(span)
+
+    val traceIdBytes = Bytes.toBytes(span.traceId)
+    val putsFuture = serviceMappingsFuture.flatMap { serviceMappings =>
+      Future.collect(serviceMappings.map { serviceMapping =>
+        val putF: Future[Put] = serviceMapping.spanNameMapper.get(span.name).map { spanNameMapping =>
+          val rk = Bytes.toBytes(serviceMapping.id) ++ Bytes.toBytes(spanNameMapping.id) ++ timeBytes
+          val p = new Put(rk)
+          p.add(TableLayouts.idxServiceSpanNameFamily, traceIdBytes, Bytes.toBytes(true))
+          p
+        }
+        putF
+      })
+    }
+
+    // Put the data into hbase.
+    putsFuture.flatMap { puts => idxServiceSpanNameTable.put(puts) }
+  }
+
+  /**
+   * Index the span by the annotations attached
+   */
+  def indexSpanByAnnotations(span: Span): Future[Unit] = {
+    // Get the normal annotations
+    val annoFutures = span.annotations.filter { a =>
+    // skip core annotations since that query can be done by service name/span name anyway 5
+      !Constants.CoreAnnotations.contains(a.value)
+    }.map { a =>
+      val sf = serviceMapper.get(a.serviceName)
+      sf.flatMap { service => service.annotationMapper.get(a.value)}.map { am => (am, Bytes.toBytes(true))}
+    }
+
+    // Get the binary annotations.
+    val baFutures = span.binaryAnnotations.map { ba =>
+      ba.host match {
+        case Some(host) => Some((ba, host))
+        case None => None
+      }
+    }.flatten.map { case (ba, host) =>
+      val sf = serviceMapper.get(host.serviceName)
+      sf.flatMap { service =>
+        service.annotationMapper.get(ba.key)
+      }.map { am =>
+        val bytes = Util.getArrayFromBuffer(ba.value)
+        (am, bytes)
+      }
+    }
+
+    // Store the sortable time stamp byte array.  This will be used for rk creation.
+    val tsBytes = getTimeStampRowKeyBytes(span)
+    val putsFuture = (baFutures ++ annoFutures).map { annoF =>
+      annoF.map { case (anno, bytes) =>
+        // Pulling out the parent here is safe because the parent must be set to find it here.
+        val rk = Bytes.toBytes(anno.parent.get.id) ++ Bytes.toBytes(anno.id) ++ tsBytes
+        val put = new Put(rk)
+        put.add(TableLayouts.idxAnnotationFamily, Bytes.toBytes(span.traceId), bytes)
+        put
+      }
+    }
+
+    // Now put them into the table.
+    Future.collect(putsFuture).flatMap { puts => idxServiceAnnotationTable.put(puts)  }
+  }
+
+  /**
+   * Store the service name, so that we easily can
+   * find out which services have been called from now and back to the ttl
+   */
+  def indexServiceName(span: Span): Future[Unit] = {
+    val futureMappings = Future.collect(span.serviceNames.map { sn => serviceMapper.get(sn)}.toSeq)
+    val timeBytes = getTimeStampRowKeyBytes(span)
+    val putsFuture = futureMappings.map { mappings =>
+      mappings.map { map =>
+        val rk = Bytes.toBytes(map.id) ++ timeBytes
+        val put = new Put(rk)
+        put.add(TableLayouts.idxServiceFamily, Bytes.toBytes(span.traceId), Bytes.toBytes(true))
+        put
+      }
+    }
+    putsFuture.flatMap { puts => idxServiceTable.put(puts) }
+  }
+
+  /**
+   * Index the span name on the service name. This is so we
+   * can get a list of span names when given a service name.
+   * Mainly for UI purposes
+   */
+  def indexSpanNameByService(span: Span): Future[Unit] = {
+    val serviceMappingsFuture = span.serviceNames.map { sn => serviceMapper.get(sn)}.toSeq
+    Future.collect(serviceMappingsFuture.map { smf =>
+      smf.flatMap {_.spanNameMapper.get(span.name)}
+    }).flatMap {
+      snm => Future.Unit
+    }
+  }
+
+  /**
+   * Index a span's duration. This is so we can look up the trace duration.
+   */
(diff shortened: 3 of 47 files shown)


zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseStorage.scala
```diff
@@ -0,0 +1,144 @@
...
+    val p = new Put(rowKeyFromSpan(span))
+    p.add(TableLayouts.storageFamily, Bytes.toBytes(span.id), serializer.toBytes(span.toThrift))
+    hbaseTable.put(p)
+  }
+
+  /**
+   * Set the ttl of a trace. Used to store a particular trace longer than the
+   * default. It must be oh so interesting!
+   *
+   * This is a NO-OP for HBase.
```
mosesn: why is this a no-op?

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseStorage.scala
```diff
@@ -0,0 +1,144 @@
...
+    val spans: Seq[Span] = lists.map { kv =>
+      serializer.fromBytes(kv.getValue).toSpan
+    }
+    spans
+  }
+
+  /**
+   * How long do we store the data before we delete it? In seconds.
+   */
+  def getDataTimeToLive: Int = 7.days inSeconds
```
mosesn: this suffix notation is a little hard to read, I think

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseStorage.scala
```diff
@@ -0,0 +1,144 @@
...
+
+  private[this] def traceIdFromRowKey(bytes:Array[Byte]):Long = {
```
mosesn: for short things like this you don't need braces

mosesn:
this is pretty neat!  have you found that hbase is fast enough to keep a fairly snappy web ui?

project/Project.scala
```diff
@@ -292,5 +293,35 @@ object Zipkin extends Build {
...
   ).dependsOn(scrooge)
+
+  lazy val hbase = Project(
+    id = "zipkin-hbase",
+    base = file("zipkin-hbase"),
+    settings = defaultSettings
+  ).settings(
+    parallelExecution in Test := false,
+    libraryDependencies ++= Seq(
+      "org.apache.hbase"     % "hbase-client"          % HBASE_VERSION notTransitive(),
```
mosesn: why do you use the raw hbase client instead of asynchbase?

project/Project.scala
```diff
@@ -292,5 +293,35 @@ object Zipkin extends Build {
...
+      "org.apache.hbase"     % "hbase-common"          % HBASE_VERSION % "test" classifier("tests") classifier(""),
+      "org.apache.hbase"     % "hbase-server"          % HBASE_VERSION % "test" classifier("tests") classifier(""),
+      "org.apache.hbase"     % "hbase-hadoop-compat"   % HBASE_VERSION % "test" classifier("tests") classifier(""),
+      "org.apache.hbase"     % "hbase-hadoop1-compat"  % HBASE_VERSION % "test" classifier("tests") classifier(""),
+      "com.google.protobuf"  % "protobuf-java"         % "2.4.1",
+      "org.apache.hadoop"    % "hadoop-core"           % "1.1.2" notTransitive(),
+      "org.apache.hadoop"    % "hadoop-test"           % "1.1.2" % "test",
+      "commons-logging"      % "commons-logging"       % "1.1.1",
+      "org.apache.zookeeper" % "zookeeper"             % "3.4.5" % "runtime" notTransitive(),
+      "org.slf4j"            % "slf4j-log4j12"         % "1.6.4" % "runtime",
```
mosesn: do we need slf4j in the project?  or is this just an hbase dependency?

zipkin-hbase/src/main/scala/com/twitter/hbase/HBaseTable.scala
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.hadoop.conf.Configuration
+import java.lang.ThreadLocal
+import org.apache.hadoop.hbase.util.Bytes
+import com.twitter.util.{FutureTask, Future}
+import scala.collection.JavaConverters._
+
+case class HBaseTable(
+  conf: Configuration,
+  tableName:String,
+  executor: ExecutorService = Executors.newCachedThreadPool()
```
mosesn: prefer [NamedThreadPoolFactory](https://github.com/twitter/util/blob/master/util-core/src/main/scala/com/twitter/concurrent/NamedPoolThreadFactory.scala), so we can have useful thread dumps.

zipkin-hbase/src/main/scala/com/twitter/hbase/HBaseTable.scala
```diff
@@ -0,0 +1,101 @@
...
+     }
+  }
+
+  /**
+   * Execute a Put async and return void when done.
+   * @param p HBase put to execute.
+   * @return Future[Unit]
+   */
+  def put(p:Put):Future[Unit] = {
+    val fTask = new FutureTask[Unit]({
```
mosesn: FutureTask + ExecutorService is more for java compat, I think.  Since this is in scala, it should be fine to just wrap the ExecutorService in a FuturePool and use it raw.

zipkin-hbase/src/main/scala/com/twitter/hbase/HBaseTable.scala
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  /**
+   * Call to add an htable to the set.
+   *
+   * This is used to close down the backing thread local htables.
+   * @param ht htable to add to the set.
+   */
+  private[this] def addHTable(ht:HTable) {
+    this.synchronized {
```
mosesn: it seems like we're going to use this so infrequently that it will be fine to just grab the lock.  however, if we start having to lock frequently, it might make sense to just use a concurrentset.

elliottneilclark: Yep that should only be called as the thread pool starts expanding (Probably only on the first few calls to hbase).

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseStorage.scala
elliottneilclark: HBase can't set a ttl on a row.  It has a ttl automatically associated with the column family.  So when the data is initially put it the ttl is calculated from the initial timestamp.

http://hbase.apache.org/book.html#ttl

I'll add some better comments here.

zipkin-hbase/src/main/scala/com/twitter/hbase/HBaseTable.scala
elliottneilclark: Will do

zipkin-hbase/src/main/scala/com/twitter/hbase/HBaseTable.scala
elliottneilclark: will do

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseStorage.scala
elliottneilclark: K, I'll clean it up.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseStorage.scala
elliottneilclark: Will do

project/Project.scala
elliottneilclark: Asynchbase doesn't support the newer versions of hbase.  And the HBase team is hoping to use this to test the newest version of HBase.

@tsuna hasn't finished his newest version of asynchbase. When he does it will probably be pretty easy (and desirable ? ) to move to that.

tsuna: Yup, it's still work in progress.  Should be available for testing within a week or two.

zipkin-hbase/src/main/scala/com/twitter/hbase/HBaseTable.scala
elliottneilclark: I'll clean it up anyway.  No harm in removing locks where possible.

project/Project.scala
elliottneilclark: Thanks Benoit.

project/Project.scala
elliottneilclark: SBT compile fails without it.  I haven't had time to track down why yet.  
Yay for the wonderful world of Hadoop java dependencies (aka everything ever written in java ever).

elliottneilclark:
> this is pretty neat! have you found that hbase is fast enough to keep a fairly snappy web ui?

Yep HBase is happily running as a backend server for lots of different web ui's.  I haven't finished this part yet, but it should be plenty fast enough and scalable.

elliottneilclark:
I changed the hbase version to a released version so it should be good to go on travis-ci now.

dvryaboy:
@ghelmling you and Chris might be interested :)

ghelmling:
Pretty cool, thanks for the pointer!

ghelmling:
I think the Travis CI failure is due to build environment (MiniDFSCluster expects a umask of 022).  Funny enough, I just answered another question on this earlier today: 
http://stackoverflow.com/questions/17625938/hbase-minidfscluster-java-fails-in-certain-environments/17665439#17665439

For another project (hRaven), we had to modify the .travis.yml file to ensure umask was set to "022" prior to the test run by adding the line:

```
script: umask 0022 && mvn test
```

elliottneilclark:
> For another project (hRaven), we had to modify the .travis.yml file to ensure umask was set to "022" prior to the test run

Awesome thanks.  That just saved me a whole lot of debugging time. I'll add that in the next push.

elliottneilclark:
So I pushed out more commits.  Mostly this is just so that everyone can see the idea behind how the indexes will look.

I also added some scaladoc comments.  There are still a lot of places that need more (They're coming I promise).

Thanks everyone for all of the reviews already.

zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
```diff
@@ -64,8 +64,8 @@ case class Span(traceId: Long, name: String, id: Long, parentId: Option[Long],
     if (annotations.isEmpty) {
       None
     } else {
-      val sName = serverSideAnnotations.flatMap(_.host).headOption.map(_.serviceName)
```
bmdhacks, line 67: I don't really mind this change, but I'm just curious about the impetus for it.

elliottneilclark, line 67: So I actually am un-doing that change in the next push.

I put that in there because there is a disconnect between serviceNames and serviceName.  However it looks like serviceName is used for the ui, and It's probably better to just take note of the weirdness and keep the user experience better.

elliottneilclark:
The latest push runs the web ui and is pretty well complete.

I think there's still a timestamp bug and indexing annotation values.

bmdhacks:
It'd be nice to get an aggregates store in there too.  I've been going over the code, I'll give you some more review feedback next week.  This looks really good.  We might even try using this at twitter.

elliottneilclark:
> It'd be nice to get an aggregates store in there too.

I'll give that a shot once I have some more tests and a lot more polish on the code.

> I'll give you some more review feedback next week.

Awesome,  thanks for your help.

bmdhacks:
Yeah, I'm actually fixing up the aggregate store tests anyways, so feel free to wait a bit.
Also, I need to to sit down with Chris Trezzo to learn about hbase schema design to make sure it's good.  Really stoked on this contribution though and I'm chomping to get some more time to help out with it.

zipkin-hbase/src/main/scala/com/twitter/hbase/HBaseTable.scala
```diff
@@ -0,0 +1,125 @@
...
+   * So we're using a thread local to make sure that every thread only talks to one HTable.  This should result in
+   * fewer threads lying around than using HTablePool.
+   */
+  private val threadLocalHTable: ThreadLocal[HTable] = new ThreadLocal[HTable] {
+     override def initialValue() = {
+       val ht = new HTable(conf, tableNameBytes,  executor)
+       htables.add(ht)
+       ht
+     }
+  }
```
franklinhu: May want to take a look at `Local` in twitter/util since that does fancy exceution context stuff:
https://github.com/twitter/util/blob/master/util-core/src/main/scala/com/twitter/util/Local.scala

zipkin-hbase/src/main/scala/com/twitter/hbase/HBaseTable.scala
```diff
@@ -0,0 +1,125 @@
+package com.twitter.hbase
```
franklinhu: We should probably namespace these under com.twitter.zipkin.storage.hbase

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseStorage.scala
```diff
@@ -0,0 +1,144 @@
...
+import com.twitter.zipkin.common.Span
+import com.twitter.zipkin.conversions.thrift._
+import com.twitter.zipkin.gen
+import com.twitter.zipkin.hbase.TableLayouts
+import com.twitter.zipkin.storage.Storage
+import org.apache.hadoop.hbase.KeyValue
+import org.apache.hadoop.hbase.client.{Result, Get, Put}
+import org.apache.hadoop.hbase.filter.KeyOnlyFilter
+import org.apache.hadoop.hbase.util.Bytes
+import scala.collection.JavaConversions._
```
franklinhu: It's recommended to use JavaConverters instead of JavaConversions to avoid invisible implicit magic

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseStorage.scala
```diff
@@ -0,0 +1,144 @@
...
+
+  /**
+   * Set the ttl of a trace. Used to store a particular trace longer than the
+   * default. It must be oh so interesting!
+   *
+   * This is a NO-OP for HBase. when the data is initially put into HBase the ttl starts from the
+   * timestamp. See http://hbase.apache.org/book.html#ttl for more information about HBase's TTL Data model.
+   */
+  def setTimeToLive(traceId: Long, ttl: Duration): Future[Unit] = {
+    Future()
```
franklinhu: `Future.Unit`

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/AnnotationMapping.scala
```diff
@@ -0,0 +1,7 @@
+package com.twitter.zipkin.storage.hbase.mapping
+
+import java.nio.ByteBuffer
+
+
+case class AnnotationMapping(id:Long, value:ByteBuffer, parent:Option[ServiceMapping]) extends Mapping {
+}
```
franklinhu: No need for the trailing braces

franklinhu:
Gave some cursory comments, but will sit down and do a more in-depth review in the near future. 
Also, slight style nit: spaces after colons for type declarations

```
myThing: Array[Byte]
```

rather than

```
myThing:Array[Byte]
```

zipkin-hbase/src/main/scala/com/twitter/hbase/HBaseTable.scala
elliottneilclark: The thought was that this class has nothing specific to zipkin in it.  It's really just an async facade on the hbase client.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseStorage.scala
elliottneilclark: Will do.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseStorage.scala
elliottneilclark: Will Do.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/AnnotationMapping.scala
elliottneilclark: Will Do.

zipkin-hbase/src/main/scala/com/twitter/hbase/HBaseTable.scala
elliottneilclark: I don't think that Local is right here.

Local seems like it should be used to make sure that a single value moves along with an executor's tree.  Local proviceds update/set.  None of these are really what I need.  

Here I'm really using ThreadLocal as a way to ensure that no two threads are accessing the same HTable (HTable has buffers that aren't protected from concurrent access).  

I should probably change this to use HTablePool, but that's a lot more code.

elliottneilclark:
Pushed a new commit that squashed the old ones.  It contains:
- rebase to master.
- Moved to HTablePool so that the usage inside of HBaseTable is a little clearer on why we're doing that.
- Changed annotation index to use annotation values
- Cleaned up most of the source code nits that @franklinhu was talking about.  Thought the test code still could have some.  I'll address that soon.

elliottneilclark:
Pushed aggregates.

zipkin-hbase/src/main/scala/com/twitter/hbase/HBaseTable.scala
franklinhu: It'd be good to avoid potential namespace collisions with other projects (however unlikely)

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseAggregates.scala
```diff
@@ -0,0 +1,117 @@
...
+    scan.setStartRow(Bytes.toBytes(Long.MaxValue - startDate.inMilliseconds))
+    endDate.foreach { ed => scan.setStopRow(Bytes.toBytes(Long.MaxValue - ed.inMilliseconds))}
+    scan.addColumn(TableLayouts.dependenciesFamily, Bytes.toBytes("\0"))
+    dependenciesTable.scan(scan, 100).map { results =>
+      val depList = results.map { result =>
+        result.list().asScala.headOption.map { kv =>
+          val tDep = serializer.fromBytes(kv.getValue)
+          tDep.toDependencies
+        }
+      }.flatten
```
franklinhu: You can use a `.flatMap` here instead of `.map { .. }.flatten`

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseIndex.scala
```diff
@@ -0,0 +1,291 @@
...
+   * Only return maximum of limit trace ids from before the endTs.
+   */
+  def getTraceIdsByName(serviceName: String, spanNameOption: Option[String], endTs: Long, limit: Int): Future[Seq[IndexedTraceId]] = {
+    val resultsFuture = spanNameOption match {
+      case None       => getTraceIdsByNameNoSpanName(serviceName, endTs, limit)
+      case Some(spanName) => getTraceIdsByNameWithSpanName(serviceName, spanName, endTs, limit)
+    }
+
+    resultsFuture.map { results =>
+      results.flatMap { result => indexResultToTraceId(result) }.toSet.toSeq.slice(0, limit)
```
franklinhu: Do we want to impose a sort order here? If not, we can probably simplify the end part to `.toSet.take(limit)`

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseIndex.scala
```diff
@@ -0,0 +1,291 @@
...
+      log.debug("getTraceIdsByAnnotation -> serviceName: %s[%d] annotation: %s[%d]", serviceName, annoMapping.parent.get.id, annotation, annoMapping.id)
+      val scan = new Scan()
+      val startRk = Bytes.toBytes(annoMapping.parent.get.id) ++ Bytes.toBytes(annoMapping.id) ++ Bytes.toBytes(0L)
+      val endRk = Bytes.toBytes(annoMapping.parent.get.id) ++ Bytes.toBytes(annoMapping.id) ++ getEndScanTimeStampRowKeyBytes(endTs)
+      scan.setStartRow(startRk)
+      scan.setStopRow(endRk)
+      scan.addFamily(TableLayouts.idxAnnotationFamily)
+      value.foreach { bb => scan.setFilter(new ValueFilter(CompareOp.EQUAL, new BinaryComparator(bb.array()))) }
+      idxServiceAnnotationTable.scan(scan, limit).map { results =>
+        results.flatMap { result => indexResultToTraceId(result)}.toSet.toSeq.slice(0, limit)
```
franklinhu: same thing here

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseIndex.scala
```diff
@@ -0,0 +1,291 @@
...
+    val baFutures = span.binaryAnnotations.map { ba =>
+      ba.host match {
+        case Some(host) => Some((ba, host))
+        case None => None
+      }
+    }.flatten.map { case (ba, host) =>
+      val sf = serviceMapper.get(host.serviceName)
+      sf.flatMap { service =>
+        service.annotationMapper.get(ba.key)
+      }.map { am => (am, ba.value.array())}
```
franklinhu: I don't think we want the entire underlying array from the ByteBuffer since it could be shared. We'll have to copy it out to a newly allocated `Arrray[Byte]`

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseIndex.scala
```diff
@@ -0,0 +1,291 @@
...
+   * Index the span name on the service name. This is so we
+   * can get a list of span names when given a service name.
+   * Mainly for UI purposes
+   */
+  def indexSpanNameByService(span: Span): Future[Unit] = {
+    val serviceMappingsFuture = span.serviceNames.map { sn => serviceMapper.get(sn)}.toSeq
+    Future.collect(serviceMappingsFuture.map { smf =>
+      smf.flatMap {_.spanNameMapper.get(span.name)}
+    }).flatMap {
+      snm => Future()
```
franklinhu: Future.Unit

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseStorage.scala
```diff
@@ -0,0 +1,144 @@
...
+   * This is a NO-OP for HBase. when the data is initially put into HBase the ttl starts from the
+   * timestamp. See http://hbase.apache.org/book.html#ttl for more information about HBase's TTL Data model.
+   */
+  def setTimeToLive(traceId: Long, ttl: Duration): Future[Unit] = Future.Unit
+
+  /**
+   * Get the time to live for a specific trace.
+   * If there are multiple ttl entries for one trace, pick the lowest one.
+   */
+  def getTimeToLive(traceId: Long): Future[Duration] = Future(7.days)
```
franklinhu: `Future.value(..)`

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseIndex.scala
elliottneilclark: Yea they are returned in time sorted order.  So if we need to limit it I would like to take the newest spans.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseIndex.scala
elliottneilclark: Done.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseAggregates.scala
elliottneilclark: Done

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseStorage.scala
```diff
@@ -0,0 +1,144 @@
...
+      g.addFamily(TableLayouts.storageFamily)
+    }
+  }
+
+  private[this] def resultToSpans(option: Option[Result]): Seq[Span] = {
+    val lists: Seq[KeyValue] = option.map { result =>
+      Option(result.list()) // Extract the key values from the result.  Each will be a span.
+    }.flatten // Flatten out the option
+      .map {_.asScala} // Make it a scala list.
+      .flatten.toSeq // Flatten the wrapper and make it a seq.
```
franklinhu: This may be a bit more readable

```
option match {
  case Some(result) => result.list().asScala
  case None => Seq.empty[KeyValue]
}
```

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/Mapper.scala
```diff
@@ -0,0 +1,164 @@
...
+    val normString = name.toLowerCase
+    if (normString.equals("")) {
+      throw new Exception("Can't get empty string")
+    }
+    get(ByteBuffer.wrap(Bytes.toBytes(normString)))
+  }
+
+  def get(value: ByteBuffer): Future[T] = pool {
+
+    val result = Retry(100) {
```
franklinhu: make this value configurable?

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseStorage.scala
elliottneilclark: Done.  Thanks

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/Mapper.scala
```diff
@@ -0,0 +1,164 @@
...
+      val preCreated: Option[T] = getPreCreated(value)
+      val m = preCreated.getOrElse {
+        val proposedFuture: Future[T] = createProposed(value)
+        val storedFuture: Future[T] = proposedFuture.flatMap { pm =>
+          store(pm)
+        }
+        Await.result(storedFuture)
```
franklinhu: :\
Maybe Retry should operate over Future[T] so it can respond to errors accordingly? Can use

```
future.transform {
  case Return(r) => r
  case Throw(t) => // log the error and try again
}
```

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/Mapper.scala
```diff
@@ -0,0 +1,164 @@
...
+      mappings.foreach { m => cache.put(m.value, m)}
+      mappings.toSeq.toSet
+    }
+  }
+
+
+  private def getPreCreated(value: ByteBuffer): Option[T] = {
+    val fromCache = Option(cache.get(value))
+    val preMade = fromCache match {
+      case None => Await.result(getFromHBase(value))
```
franklinhu: Should try to avoid blocking

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseStorage.scala
elliottneilclark: Done.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/Mapper.scala
```diff
@@ -0,0 +1,164 @@
...
+      case s => s
+    }
+    preMade
+  }
+
+  private def getFromHBase(value: ByteBuffer): Future[Option[T]] = {
+    val valueBytes = value.array()
```
franklinhu: Need to copy out the correct segment of bytes we want

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseIndex.scala
elliottneilclark: Sure.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/Mapper.scala
```diff
@@ -0,0 +1,164 @@
...
+      results.headOption.map(createFromResult(_))
+    }
+  }
+
+  private def store(proposedMapping: T): Future[T] = {
+    // Forwards : Name to Id
+    // backwards: Id to Name
+
+    val idBytes = Bytes.toBytes(proposedMapping.id)
+    val valBytes = proposedMapping.value.array()
```
franklinhu: Same here

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/ServiceMapping.scala
```diff
@@ -0,0 +1,14 @@
...
+import com.twitter.hbase.HBaseTable
+import com.twitter.zipkin.storage.hbase.IDGenerator
+import java.nio.ByteBuffer
+import org.apache.hadoop.hbase.util.Bytes
+
+case class ServiceMapping(id: Long, value: ByteBuffer, mappingTable: HBaseTable, idGen: IDGenerator) extends Mapping {
+  val parent: Option[Mapping] = None
+  val annotationMapper = new AnnotationMapper(this)
+  val spanNameMapper = new SpanNameMapper(this)
+  val name = Bytes.toString(value.array())
```
franklinhu: `Bytes.toString(value.array, value.position, value.remaining)`

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/SpanNameMapping.scala
```diff
@@ -0,0 +1,9 @@
+package com.twitter.zipkin.storage.hbase.mapping
+
+import java.nio.ByteBuffer
+import org.apache.hadoop.hbase.util.Bytes
+
+case class SpanNameMapping(id: Long, value: ByteBuffer, parent: Option[ServiceMapping]) extends Mapping {
+  val mappingType: Byte = 1
+  val name = Bytes.toString(value.array())
```
franklinhu: `Bytes.toString(value.array, value.position, value.remaining)`

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseIndex.scala
franklinhu: Maybe `.toSeq.distinct.take(limit)` then?

zipkin-hbase/src/main/scala/com/twitter/hbase/HBaseTable.scala
elliottneilclark: Ok.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/Mapper.scala
elliottneilclark: Seems like making this blocking is actually better here.  It means there are a lot fewer calls to an ExecutorService on the most used path.

Since most mappings will be in the cache keeping it blocking means that only one future will ever be created on hits.

elliottneilclark:
I pushed a new version.  The mapping code still has some blocking; keeping some blocking just makes the code much cleaner and allows a fully tail recursive version of the retry loop.  In addition it allows me to completely stay off of all thread executors on cache hits.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/Mapper.scala
```diff
@@ -0,0 +1,162 @@
...
+
+      val mapping: Future[T] = cachedOption.getOrElse {
+        // Spin up a pool future here as we need to await fully to make sure that it's durable before returning.
+        pool {
+          val valueBytes: Array[Byte] = ByteBufferUtil.getBytesFromByteBuffer(value)
+          val fromOrToHBase: Future[T] = getFromHBase(valueBytes).flatMap { mo =>
+            mo match {
+              // If there is already a mapping wrap the mapping in a future after putting it in the cache.
+              case Some(mapping) => {
+                cache.put(ByteBuffer.wrap(mapping.value), mapping)
```
mosesn: You never check that this succeeded, is that ok?

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/Mapper.scala
```diff
@@ -0,0 +1,162 @@
...
+    }
+    get(ByteBuffer.wrap(Bytes.toBytes(normString)))
+  }
+
+  def get(value: ByteBuffer): Future[T] = {
+    val result = Retry(100) {
+      val cachedOption: Option[Future[T]] = Option(cache.get(value)).map(Future.value)
+
+      val mapping: Future[T] = cachedOption.getOrElse {
+        // Spin up a pool future here as we need to await fully to make sure that it's durable before returning.
```
mosesn: I don't understand what this comment means.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/Mapper.scala
```diff
@@ -0,0 +1,162 @@
...
+
+    GASP!!!
+
+    This should be a small table and it will greatly help with
+    creating more mappings if this is in cache.
+    */
+    scan.setCacheBlocks(true)
+    mappingTable.scan(scan, maxToLoad).map { results =>
+      val mappings = results.map(createFromResult)
+      mappings.foreach { m => cache.put(ByteBuffer.wrap(m.value), m)}
```
mosesn: you aren't checking whether cache.put has succeeded, is that ok?

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/Mapper.scala
```diff
@@ -0,0 +1,162 @@
...
+    get(ByteBuffer.wrap(Bytes.toBytes(normString)))
+  }
+
+  def get(value: ByteBuffer): Future[T] = {
+    val result = Retry(100) {
+      val cachedOption: Option[Future[T]] = Option(cache.get(value)).map(Future.value)
+
+      val mapping: Future[T] = cachedOption.getOrElse {
+        // Spin up a pool future here as we need to await fully to make sure that it's durable before returning.
+        pool {
```
mosesn: I don't understand why you're doing this.  It seems like you're getting a thread just to block.  Wouldn't it be the same if you just didn't use the futurepool and didn't await?

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/Mapper.scala
elliottneilclark: I could put an assert here.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/Mapper.scala
elliottneilclark: Yeah I think I optimized too much last night and I optimized away the retry.... Oooops.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/Mapper.scala
elliottneilclark: The cache is a concurrent hash map.  If putting into a map fails there are bigger issues.  I could add an assert to make sure that the value returned is what was expected (eg if we're replacing make sure that we're replacing a mapping that is exactly equal to what we just created).

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/Mapper.scala
mosesn: never mind, I thought this was a network call.

elliottneilclark: I actually really liked the idea of adding an assert just to make 100% sure that the mapping table stays consistent.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/Mapper.scala
elliottneilclark: I've got another version that has much better comments.  I'll post that in a second after I do some more local testing.

elliottneilclark:
Hmmm not sure what's going on I'll try and work out how to fix the tests.  They seem like a test only issue on travis.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/mapping/Mapper.scala
```diff
@@ -0,0 +1,183 @@
...
+import com.twitter.zipkin.storage.hbase.utils._
+import java.nio.ByteBuffer
+import java.util.concurrent.ConcurrentHashMap
+import org.apache.hadoop.hbase.client._
+import org.apache.hadoop.hbase.util.Bytes
+import scala.Some
+import com.twitter.zipkin.storage.hbase.utils.HBaseTable
+import com.twitter.zipkin.storage.hbase.utils.IDGenerator
+
+trait Mapper[T >: Null <: Mapping] {
```
bmdhacks, line 20: Isn't the Null bound here redundant to subclassing the Mapping trait?  I can't think of how you could have a non-nullable Mapping.  Or are you just including it here as pseudo-documentation that having Mapper[Null] is ok?

elliottneilclark, line 20: I would have thought so too.  However when passing a Mapping of type T to Retry, the type falls back to java.lang.Object if Null isn't in the parameterized type here

bmdhacks:
Oh hey, I merged my dependencies work and messed up the interface.  Do you have time to rebase?

elliottneilclark:
Sure.  Here's a rebased version.  It has:
- New dependencies
- New version of HBase
- Better threading
- Un constrained thread creation is a bad idea with the # of qps that hbase can generate :-)

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseIndex.scala
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  /**
+   * Index the span by the annotations attached
+   */
+  def indexSpanByAnnotations(span: Span): Future[Unit] = {
+    // Get the normal annotations
+    val annoFutures = span.annotations.filter { a =>
+    // skip core annotations since that query can be done by service name/span name anyway 5
+      !Constants.CoreAnnotations.contains(a.value)
```
bmdhacks, line 152: TODO: we added CoreAddress annotations in Isaac's PR so whatever gets merged first we probably want to fix the other one.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/utils/ByteBufferUtil.scala
```diff
@@ -0,0 +1,12 @@
+package com.twitter.zipkin.storage.hbase.utils
+
+import java.nio.ByteBuffer
+
+object ByteBufferUtil {
+  def getBytesFromByteBuffer(value: ByteBuffer): Array[Byte] = {
```
bmdhacks: com.twitter.zipkin.util.Util.getArrayFromBuffer has recently been added,
although I just noticed the double-util in the class path which I might fix.

elliottneilclark: Sure I'll use that now.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/utils/IdGenerator.scala
```diff
@@ -0,0 +1,25 @@
...
+    // Failing here is fine.  We just waste an id.
+    val idResult = idGenTable.atomicIncrement(inc)
+    val idBytes = idResult.map { r =>
+      val bytes = r.getValue(TableLayouts.idGenFamily, qual)
+      if (Bytes.toLong(bytes) == Long.MaxValue) {
+        throw new Exception("All ID's are used up.")
+      }
+      Bytes.toLong(bytes)
+    }
+    idBytes
```
bmdhacks: UberNit:  don't really need that last idBytes val if you're just gonna turn around and return it.
Yes, I'm trying to look for things to critique.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/utils/Retry.scala
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Retry a given function up to nTries.
+ *
+ * All exceptions will be caught until the retry limit is met.
+ *
+ * No clean up between each invocation is attempted.  It's up to the user to ensue that
+ * supplied function is resilient to this fact.
+ * @param nTries number of retries that will be attempted.
+ */
+case class Retry(nTries: Int) {
```
bmdhacks: Make this an object for less GC churn.

``` scala
object Retry {
  def apply[T](n: Int)(f: => T): T => {
    ...
  }
}
```

bmdhacks: Also this is generic enough to chuck into zipkin-common Util

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/utils/IdGenerator.scala
elliottneilclark: Yep that was left over from debugging so I could put breakpoints.  I'll clean that up.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/utils/Retry.scala
elliottneilclark: Will do.

bmdhacks: [sp]retries exhausted
Also might be more readable to use a for comprehension instead of recursion:

``` scala
  def apply[T](n: Int)(f: => T) : T = {
    var result:Option[T] = None
    var throwable:Option[Throwable] = None

    for (i <- 0 until n if result.isEmpty) {
      try {
        result = Option(f)
      } catch {
        case e:Throwable => { throwable = Some(e) }
      }
    }

    if (result.isEmpty && throwable.isDefined) {
      throw new RetriesExhaustedException("%d retries exhausted".format(n), throwable)
    }

    result.get
  }
```

Got the nitpick meter up to 100% !!!

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/utils/ThreadProvider.scala
```diff
@@ -0,0 +1,12 @@
...
+
+import com.twitter.concurrent.NamedPoolThreadFactory
+import java.util.concurrent.ThreadPoolExecutor.{CallerRunsPolicy, AbortPolicy}
+import java.util.concurrent.{Executors, LinkedBlockingDeque, TimeUnit, ThreadPoolExecutor}
+
+object ThreadProvider {
+  lazy val defaultExecutor = Executors.newFixedThreadPool(32, new NamedPoolThreadFactory("HBase-Executor", true))
+  lazy val clientExecutor = Executors.newFixedThreadPool(32, new NamedPoolThreadFactory("HBase-Client", true))
+
+  val maxThreads = 32
```
bmdhacks: put maxThreads first and pass it in to newFixedThreadPool?

bmdhacks:
Ok, done reviewing.  Lemme know about my nits, but I'm cool with merging it as-is.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/utils/ThreadProvider.scala
elliottneilclark: I broke the executors up a little bit more.  I'll post a version.

zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/utils/Retry.scala
elliottneilclark: Nice, I like the for comprehension.  Makes it more explicit that it's iterative.

elliottneilclark:
Ok, I think that addresses all of your thoughts.

I added more thread pools so they can spin up faster when load is applied.

bmdhacks:
Is this good to merge then from your perspective?

bmdhacks:
Some of the unit tests fail.

elliottneilclark:
Whoops yeah some debugging stuff got left in there.  Lets run this past ci again.

elliottneilclark:
Yep ready to go in.

PR #276, opened 2013-07-23
Dependency Visualizations

bmdhacks:
This is the final work to bubble up the aggregate dependencies to the frontend.


zipkin-web/src/main/resources/public/js/zipkin-aggregates.js
```diff
@@ -0,0 +1,96 @@
...
+ */
+Zipkin.Aggregates.loadJson = function (callback) {
+
+  var nodeMap = {};
+  var graph = {};
+
+  /**
+   * create an empty node
+   */
+  function newNode(name) {
```
IceCreamYou: Not a commit-blocker, but for style points, the typical JS pattern is like this:

```
function Node(name) {
    this.name = name;
    /* ... */
}
// Instantiate a new node with new Node(name);
```

zipkin-web/src/main/resources/public/js/zipkin-block-dependencies.js
```diff
@@ -0,0 +1,328 @@
...
+        node.y=0;
+        node.x0=0;
+        node.y0=0;
+        node.height = boxHeight;
+        node.id = id;
+        id += 1;
+        return node;
+      });
+
+      self.selected = _.find(self.nodes, function (node) {
```
IceCreamYou: Seems like you could do this in the map() above instead of iterating through the nodes again

zipkin-web/src/main/resources/public/js/zipkin-block-dependencies.js
```diff
@@ -0,0 +1,328 @@
...
+
+  var topMargin = 20; // spacing from the top
+  var boxSpacing = 5;
+  var boxHeight = 17;
+  var selectedHeight = 600; // how tall is the middle box
+  var boxWidth = 200;
+  var firstColumn = 0;
+  var secondColumn = 400;
+  var thirdColumn = 800;
+  var fourthColumn = 1200; // only used for animation starting point
```
IceCreamYou: If style properties have to be defined literally in code, I like to make them properties of an object, e.g. `var STYLE = { topMargin: 20, /* ... */ };` just to make it clearer when they're used that they are set, not computed.

zipkin-web/src/main/resources/public/js/zipkin-block-dependencies.js
```diff
@@ -0,0 +1,328 @@
...
+        ;
+
+      // links
+      svg.selectAll(".blockLink")
+        .data(graph.links, function (d) { return d.id })
+        .call(drawLink)
+        ;
+    }
+
+    function click_node(node) {
```
IceCreamYou: Nitpick: camelCase for consistency

zipkin-web/src/main/resources/public/js/zipkin-block-dependencies.js
```diff
@@ -0,0 +1,328 @@
...
+      return combined;
+
+    }
+
+    /**
+     * Given a list of all nodes and one specific selected node, return
+     * a data structure enumerating all nodes connected to our selection,
+     * and all links to connect them.
+     */
+    function zoom_data(selected) {
```
IceCreamYou: Nitpick: camelCase for consistency

zipkin-web/src/main/resources/public/js/zipkin-block-dependencies.js
```diff
@@ -0,0 +1,328 @@
...
+
+      var sortedParentLinks = selected.linksTo.sort(function(a,b) {
+        return b.moments.m0 - a.moments.m0;
+      });
+      var parents = [];
+      _.each(sortedParentLinks, function(link, index) {
+        var node = link.source;
+
+        node.ratio = link.moments.m0 / selected.moments.m0;
+        if (node.generation < currGeneration - 1 || node.class != "blockTarget") {
```
IceCreamYou, line 219: Is there some place magic names like "blockTarget" are defined, and can we set a variable for them so they can be easily identified and changed?

zipkin-web/src/main/resources/public/js/zipkin-block-dependencies.js
```diff
@@ -0,0 +1,328 @@
...
+      _.each(selected.linksFrom, function (link, index) {
+        childDurationTotal += link.moments.mean() * link.moments.count();
+      });
+
+      var sortedChildren = selected.linksFrom.sort(function(a,b) {
+        return b.moments.mean() * b.moments.count() - a.moments.mean() * a.moments.count();
+      });
+
+      var children = [];
+      _.each(sortedChildren,function (link, index) {
```
IceCreamYou, line 251: This looks like the same code that gets executed for parents, except for some alignment variables -- can we abstract it to a function?

zipkin-web/src/main/resources/public/js/zipkin-block-dependencies.js
```diff
@@ -0,0 +1,328 @@
...
+
+    var diagonal = d3.svg.diagonal()
+      .projection(function (d) { return [d.y, d.x]; })
+      .source(function (d) { //noinspection JSSuspiciousNameCombination
+              return {x: d.source.y, y: d.source.x, name: d.source.name }; })
+      .target(function (d) { //noinspection JSSuspiciousNameCombination
+              return {x: d.target.y, y: d.target.x, name: d.target.name }; })
+      ;
+
+    function drawNode(group) {
```
IceCreamYou, line 93: Is there nowhere else a node would ever need to be drawn? If you make node into a class as I suggested above, I would consider making drawNode() into Node#draw().

zipkin-web/src/main/resources/public/js/zipkin-moments.js
```diff
@@ -0,0 +1,69 @@
...
+ */
+Zipkin.Moments = (function() {
+
+  var Moments = function(m) {
+    this.m0 = m.m0;
+    this.m1 = m.m1;
+    this.m2 = m.m2;
+    this.m3 = m.m3;
+    this.m4 = m.m4;
+  };
```
IceCreamYou, line 32: If you're creating a bunch of Moments, all the extra objects that are created briefly just for the purpose of creating a new Moment and then thrown away afterwards can create significant GC pause. Two alternatives are to pass in the moments as parameters directly (e.g. `new Moment(m0, m1, m2, m3, m4)`) or to assign the object to this (`this = m`).

zipkin-web/src/main/resources/public/js/zipkin-radial-dependencies.js
```diff
@@ -0,0 +1,164 @@
...
+/**
+ * radial graph of global dependencies
+ */
+Zipkin.RadialDependencies = (function () {
+
+  var RadialDependencies = function (startTime, endTime) {
+    this.startTime = startTime;
+    this.endTime = endTime;
+
+    this.diameter = 1024;
```
IceCreamYou: Can different RadialDependencies instances have different diameters? If not, I would make this a static property (i.e. `RadialDependencies.diameter = 1024;`)

IceCreamYou:
Added a couple comments on the JS -- mostly nitpicky, in general LGTM.

zipkin-web/src/main/resources/public/js/zipkin-block-dependencies.js
bmdhacks, line 219: I could make a constant, but it's still hard-coded in the css

zipkin-web/src/main/resources/public/js/zipkin-block-dependencies.js
bmdhacks, line 251: I couldn't come up with a function that didn't take a jillion arguments.  Each line does same thing, but with subtle differences.

zipkin-web/src/main/resources/public/js/zipkin-moments.js
bmdhacks, line 32: not possible to assign to `this` in javascript.

zipkin-web/src/main/resources/public/js/zipkin-block-dependencies.js
bmdhacks, line 93: The problem is that d3 sortof caters to a specific style where Node#draw() won't work since it bases things off of selections.

zipkin-query-core/src/main/scala/com/twitter/zipkin/query/QueryService.scala
```diff
@@ -384,10 +384,10 @@ class QueryService(storage: Storage, index: Index, aggregates: Aggregates, adjus
   }
 
   /** Aggregates related */
-  def getDependencies(startTime: Long, endTime: Option[Long]) : Future[gen.Dependencies] = {
+  def getDependencies(startTime: Option[Long], endTime: Option[Long]) : Future[gen.Dependencies] = {
     log.debug("getDependencies: " + startTime + " - " + endTime)
     call("getDependencies") {
-      val start = Time.fromNanoseconds(startTime*1000)
+      val start = startTime.map { t => Time.fromNanoseconds(t*1000) }
       val end = endTime.map { t => Time.fromNanoseconds(t*1000) }
```
franklinhu, line 391: Why are we multiplying by 1000? `t` is in picoseconds?

bmdhacks, line 391: 1 microsecond = 1000 nanoseconds
n microseconds  = n_1000 nanoseconds
Time.fromNanoseconds(n_1000)

bmdhacks, line 391: Although I dislike this conversion.  I submitted fromMicroseconds to util, which should be released now.  Gonna up the artifact versions to pull that in.

PR #281, opened 2013-07-29
Add aggregates support to the Anorm module

IceCreamYou:
A preemptive note about the getDependencies time range -- CassandraAggregates does this:

```
val realStart = startDate.floor(1.day).inMicroseconds
val realEnd = endDate.getOrElse(startDate).floor(1.day).inMicroseconds
```

Whereas AnormAggregates does this:

```
val startMs = startDate.inMicroseconds
val endMs = endDate.getOrElse(startDate + 1.day).inMicroseconds
```

What CassandraAggregates does looks clearly wrong to me, and Franklin confirms that it looks broken. Basically right now it says that if you request a start date but do not pass the optional end date, or if you request a start and end date within the same day, it will search within a single microsecond at the beginning of that day. I would humbly suggest that it is not getDependencies' role to enforce flooring to the day; rather it should return dependencies in the exact time range requested, and the caller should choose whether they want dependencies within a day range or not. In any case, the documentation for floor() indicates that it should not be called for values higher than 1 hour or unexpected results could occur due to time zone differences -- so our current usage of floor has an undefined result.

The places where getDependencies are used confirms that this is broken. Currently it is only called from the CassandraAggregatesSpec test and from the /api/dependencies callback in App.scala. The API endpoint has incorrect documentation (it indicates that the parameters should be in seconds) and nonsensical default behavior (if startTime is undefined it retrieves only dependencies from the first microsecond of the current day). The test passes because it only looks for dependencies in the first microsecond of the specified day.

That's why this pull request attempts to fix those things in addition to implementing Aggregates support in the Anorm module.

Also note, as indicated at the top of AnormAggregates.scala, that the top and KV dependencies methods are stubbed because they are not currently used anywhere (that feature was never completed) so there is no definition of what "working" means for them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,21 +7,15 @@ See [http://twitter.github.com/zipkin](http://twitter.github.com/zipkin)
 
 ## Quick start
 
-To install Zipkin on a single machine, see the
-[Ubuntu Quickstart](https://github.com/twitter/zipkin/blob/master/doc/ubuntu-quickstart.txt) and
-[Mac Quickstart](https://github.com/twitter/zipkin/blob/master/doc/mac-quickstart.md) guides.
-For more in-depth installation instructions with an explanation of the
-dependencies and related services, see
-[install.md](https://github.com/twitter/zipkin/blob/master/doc/install.md).
-
-Zipkin itself provides three services:
+Zipkin provides three services:
 
  - To collect data: `bin/collector`
  - To extract data: `bin/query`
  - To display data: `bin/web`
 
-If all three of these daemons are running, you should be able to visit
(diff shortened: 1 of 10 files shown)


franklinhu:
You may want to rebase this branch against master since the anorm stuff was merged?

Edit: or just send a PR for the changes against the branch this was built on?

IceCreamYou:
@franklinhu this is against master after the anorm commit.

However, I'm getting confused tracking two concurrent but dependent sets of changes with the [async PR](https://github.com/twitter/zipkin/pull/283) and that one is more current so I'm going to close this one.

PR #282, opened 2013-07-29
Regenerating web docs from master

IceCreamYou:
Docs need to be updated from the Anorm module landing

All that is required to do this is to run `python make.py`


bmdhacks:
Does this incorporate the image links patch recently sent by the guy on HN which moves the images to github CDN?

IceCreamYou:
Yes

PR #283, opened 2013-07-30
Anorm module cleanup: async, transactions, aggregates

IceCreamYou:
This branch wraps all the methods that access the DB in a FuturePool, wraps writes in transactions, and adds aggregates.

### DIFF
--- a/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormAggregates.scala
+++ b/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormAggregates.scala
@@ -1,0 +1,139 @@
+/*
+ * Copyright 2013 Twitter Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.twitter.zipkin.storage.anormdb
+
+import com.twitter.util.{Future, Time}
+import com.twitter.conversions.time._
+import com.twitter.zipkin.common.{Service, DependencyLink, Dependencies}
+import com.twitter.zipkin.storage.Aggregates
+import java.sql.Connection
+import anorm._
+import anorm.SqlParser._
+import com.twitter.algebird.Moments
+
+/**
+ * Retrieve and store aggregate dependency information.
+ *
+ * The top annotations methods are stubbed because they're not currently
+ * used anywhere; that feature was never completed.
+ */
+case class AnormAggregates(db: DB, openCon: Option[Connection] = None) extends Aggregates {
+  // Database connection object
+  private implicit val conn = openCon match {
+    case None => db.getConnection()
+    case Some(con) => con
+  }
+
+  /**
+   * Close the index
+   */
+  def close() { conn.close() }
+
+  /**
+   * Get the dependencies in a time range.
+   *
+   * endDate is optional and if not passed defaults to startDate plus one day.
+   */
+  def getDependencies(startDate: Time, endDate: Option[Time]=None): Future[Dependencies] = {
+    val startMs = startDate.inMicroseconds
+    val endMs = endDate.getOrElse(startDate + 1.day).inMicroseconds
+
+    val links: List[DependencyLink] = SQL(
+      """SELECT parent, child, m0, m1, m2, m3, m4
+        |FROM zipkin_dependency_links AS l
+        |LEFT JOIN zipkin_dependencies AS d
+        |  ON l.dlid = d.dlid
+        |WHERE start_ts >= {startTs}
+        |  AND end_ts <= {endTs}
+        |ORDER BY l.dlid DESC
+      """.stripMargin)
+    .on("startTs" -> startMs)
+    .on("endTs" -> endMs)
+    .as((str("parent") ~ str("child") ~ long("m0") ~ get[Double]("m1") ~ get[Double]("m2") ~ get[Double]("m3") ~ get[Double]("m4") map {
+      case parent ~ child ~ m0 ~ m1 ~ m2 ~ m3 ~ m4 => new DependencyLink(
+        new Service(parent),
+        new Service(child),
+        new Moments(m0, m1, m2, m3, m4)
+      )
+    }) *)
+
+    Future {
+      new Dependencies(startDate, Time.fromNanoseconds(endMs*1000), links)
+    }
+  }
+
+  /**
+   * Write dependencies
+   *
+   * Synchronize these so we don't do concurrent writes from the same box
+   */
+  def storeDependencies(dependencies: Dependencies): Future[Unit] = {
+    val dlid = SQL("""INSERT INTO zipkin_dependencies
+          |  (start_ts, end_ts)
+          |VALUES ({startTs}, {endTs})
+        """.stripMargin)
+      .on("startTs" -> dependencies.startTime.inMicroseconds)
+      .on("endTs" -> dependencies.endTime.inMicroseconds)
+    .executeInsert()
+
+    dependencies.links.foreach { link =>
+      SQL("""INSERT INTO zipkin_dependency_links
+            |  (dlid, parent, child, m0, m1, m2, m3, m4)
+            |VALUES ({dlid}, {parent}, {child}, {m0}, {m1}, {m2}, {m3}, {m4})
+          """.stripMargin)
+        .on("dlid" -> dlid)
+        .on("parent" -> link.parent.name)
+        .on("child" -> link.child.name)
+        .on("m0" -> link.durationMoments.m0)
+        .on("m1" -> link.durationMoments.m1)
+        .on("m2" -> link.durationMoments.m2)
+        .on("m3" -> link.durationMoments.m3)
+        .on("m4" -> link.durationMoments.m4)
+      .execute()
+    }
+
+    Future.Unit
+  }
+
+  /**
+   * Get the top annotations for a service name
+   */
+  def getTopAnnotations(serviceName: String): Future[Seq[String]] = {
+    Future.value(Seq[String]())
+  }
+
+  /**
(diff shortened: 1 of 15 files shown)


mosesn:
Future {} runs the task synchronously, but this does it async, so if you're not Await.resulting on every Future, you could have an ordering issue.

bmdhacks:
Do we care about ordering of inserts in zipkin?

IceCreamYou:
I think writes are async with Cassandra and Redis so this wouldn't change behavior.

IceCreamYou:
Anyone know if there is some kind of issue with Await.result() on FuturePools? Not sure why the AnormIndexSpec tests are timing out.

mosesn:
I think this [PR](https://github.com/twitter/zipkin/pull/286) fixes it.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormAggregates.scala
```diff
@@ -0,0 +1,143 @@
...
+        .execute()
+      }
+    })
+  }
+
+  /**
+   * Get the top annotations for a service name
+   */
+  def getTopAnnotations(serviceName: String): Future[Seq[String]] = {
+    Future.value(Seq[String]())
```
franklinhu: `Seq.empty[String]` is slightly clearer

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormAggregates.scala
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  /**
+   * Get the top key value annotation keys for a service name
+   */
+  def getTopKeyValueAnnotations(serviceName: String): Future[Seq[String]] = {
+    Future.value(Seq[String]())
```
franklinhu: same here

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormAggregates.scala
```diff
@@ -0,0 +1,143 @@
...
+   * Close the index
+   */
+  def close() { conn.close() }
+
+  /**
+   * Get the dependencies in a time range.
+   *
+   * endDate is optional and if not passed defaults to startDate plus one day.
+   */
+  def getDependencies(startDate: Time, endDate: Option[Time]=None): Future[Dependencies] = sqlFuturePool[Dependencies] {
```
franklinhu: Is the type annotation necessary here?

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/DB.scala
```diff
@@ -48,6 +48,32 @@ case class DB(dbconfig: DBConfig = new DBConfig()) {
...
   /**
+   * Execute SQL in a transaction.
+   *
+   * Example usage:
+   *
+   * db.withTransaction(conn, { implicit conn: Connection =>
+   *   // Do database updates
+   * })
+   */
+  def withTransaction[A](conn: Connection, code: Connection => A): Either[String, A] = {
```
franklinhu: If `Left` is always an error string, then I think we actually want to return a `Try[A]`. `Try` has two subtypes: `Return` and `Throw`.

https://github.com/twitter/util/blob/master/util-core/src/main/scala/com/twitter/util/Try.scala

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormAggregates.scala
franklinhu: Sorry I should be more clear, is the type annotation on the FuturePool call necessary?

IceCreamYou:
Per [my comment here](https://github.com/twitter/zipkin/pull/281#issuecomment-21886353) I am getting confused tracking two concurrent but dependent sets of changes to the Anorm module (adding async + transactions and adding aggregates support) so I am closing the other PR and using this one for both since this branch is more current.

IceCreamYou:
Heads up, I force-pushed so I could rebase to current master since there were merge conflicts.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormIndex.scala
```diff
@@ -54,7 +61,7 @@ case class AnormIndex(db: DB, openCon: Option[Connection] = None) extends Index
    * Only return maximum of limit trace ids from before the endTs.
    */
   def getTraceIdsByName(serviceName: String, spanName: Option[String],
-                        endTs: Long, limit: Int): Future[Seq[IndexedTraceId]] = {
+                        endTs: Long, limit: Int): Future[Seq[IndexedTraceId]] = sqlFuturePool[Seq[IndexedTraceId]] {
```
bmdhacks: Gratuitous type annotation.  Either annotate the method or the future pool, no need for both.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormIndex.scala
```diff
@@ -84,55 +89,54 @@ case class AnormIndex(db: DB, openCon: Option[Connection] = None) extends Index
...
-            |ORDER BY a_timestamp DESC
-            |LIMIT {limit}
-          """.stripMargin)
-          .on("service_name" -> serviceName)
-          .on("annotation" -> annotation)
-          .on("end_ts" -> endTs)
-          .on("limit" -> limit)
-          .as((long("trace_id") ~ long("MAX(a_timestamp)") map flatten) *)
-      }
+                              endTs: Long, limit: Int): Future[Seq[IndexedTraceId]] = sqlFuturePool[Seq[IndexedTraceId]] {
```
bmdhacks: same here

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormIndex.scala
```diff
@@ -144,7 +148,7 @@ case class AnormIndex(db: DB, openCon: Option[Connection] = None) extends Index
    *
    * Duration returned in microseconds.
    */
-  def getTracesDuration(traceIds: Seq[Long]): Future[Seq[TraceIdDuration]] = {
+  def getTracesDuration(traceIds: Seq[Long]): Future[Seq[TraceIdDuration]] = sqlFuturePool[Seq[TraceIdDuration]] {
```
bmdhacks: etc

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormIndex.scala
```diff
@@ -153,45 +157,39 @@ case class AnormIndex(db: DB, openCon: Option[Connection] = None) extends Index
...
-    Future {
-      SQL(
-        """SELECT service_name
-          |FROM zipkin_annotations
-          |GROUP BY service_name
-          |ORDER BY service_name ASC
-        """.stripMargin)
-        .as(str("service_name") *).toSet
-    }
+  def getServiceNames: Future[Set[String]] = sqlFuturePool[Set[String]] {
```
bmdhacks: go find the rest of them

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormStorage.scala
```diff
@@ -45,6 +46,11 @@ case class AnormStorage(db: DB, openCon: Option[Connection] = None) extends Stor
     case Some(con) => con
   }
 
+  // Cached pools automatically close threads after 60 seconds
+  private val threadPool = Executors.newCachedThreadPool()
```
bmdhacks: Do you nead a future pool for each storage object, or can you define one in an object that is reused.  Gimme the pros/cons of both approaches.  I think one future pool would be easier to tune than many.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormAggregates.scala
```diff
@@ -0,0 +1,143 @@
...
+    .on("endTs" -> endMs)
+    .as((str("parent") ~ str("child") ~ long("m0") ~ get[Double]("m1") ~ get[Double]("m2") ~ get[Double]("m3") ~ get[Double]("m4") map {
+      case parent ~ child ~ m0 ~ m1 ~ m2 ~ m3 ~ m4 => new DependencyLink(
+        new Service(parent),
+        new Service(child),
+        new Moments(m0, m1, m2, m3, m4)
+      )
+    }) *)
+
+    new Dependencies(Time.fromNanoseconds(startMs*1000), Time.fromNanoseconds(endMs*1000), links)
```
bmdhacks: I just committed to master new deps that include Time.fromMicroseconds

bmdhacks:
Please rebase and reopen also.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormStorage.scala
IceCreamYou: Separate FuturePools keep the caller from having to know anything about the implementation; we wouldn't have to put scary thread-related code in the configuration. On the other hand, reusing a single FuturePool is a little more efficient with cached threadpools because threads can be reused for both indexing and storage. (The efficiency difference is ambiguous with fixed thread pools because fewer threads get created total at the expense of less parallel computation.) In principle, we could get the benefit of both approaches by using an optional parameter that spawns a new FuturePool if no existing pool was passed, and not using it by default (so that people who know what they're doing could use the more efficient approach and people who don't wouldn't be intimidated by seeing threads in their config). Thoughts?

I also just realized that cached thread pools are self-closing (or at least, the threads are) and fixed thread pools don't seem to be. The code right now doesn't close the thread pools. I think it should be okay to close them when the database connection is closed since after that the threads wouldn't be able to do anything interesting anyway, but I'm not really sure what is supposed to happen if e.g. AnormStorage is closed synchronously while writes are occurring in a separate thread...

bmdhacks: Does the query and collector service shutdown cleanly when you hit localhost:9901/shutdown.txt etc?

bmdhacks: I vote to have a single global thread pool.

PR #284, opened 2013-07-30
update to latest dependencies

bmdhacks:
This is basically so I could get Time.fromMicroseconds() in util, but it pulled a bunch of other stuff up too.


franklinhu:
lgtm

PR #286, opened 2013-07-31
debug timeouts

mosesn:
The way that future pools work is that they return empty promises immediately, but spin off a thread to fulfill the promise.  Instead of trying to fulfil the promise, returning Future.value hopes to short-circuit the promise entirely, but the promise had already been returned.  I actually don't know exactly what the behavior is when you try to drop a stack frame that doesn't exist anymore, but it doesn't sound good!

In general, we should probably avoid having multiple exit points for a given method.  Having multiple exit points makes functions more difficult to reason about.

TL;DR the future never got finished :anguished:

edit: figured out what [happens](http://stackoverflow.com/questions/17754976/scala-return-statements-in-anonymous-functions/17755988#17755988) when you return elsewhere

### DIFF
--- a/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormIndex.scala
+++ b/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormIndex.scala
@@ -90,14 +90,15 @@ case class AnormIndex(db: DB, openCon: Option[Connection] = None) extends Index 
   def getTraceIdsByAnnotation(serviceName: String, annotation: String, value: Option[ByteBuffer],
                               endTs: Long, limit: Int): Future[Seq[IndexedTraceId]] = sqlFuturePool[Seq[IndexedTraceId]] {
     // Ignore core annotations. Yay magic names!
-    if (List("cs", "cr", "ss", "sr", "ca", "sa").contains(annotation))
-      return Future.value[Seq[IndexedTraceId]](Seq())
-
-    val result:List[(Long, Long)] = value match {
-      // Binary annotations
-      case Some(bytes) => {
-        SQL(
-          """SELECT zba.trace_id, s.created_ts
+    if (List("cs", "cr", "ss", "sr", "ca", "sa").contains(annotation)) {
+      Seq.empty
+    }
+    else {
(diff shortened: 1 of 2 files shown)


mosesn:
The test that failed on jdk6 is just flaky.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormIndex.scala
```diff
@@ -90,14 +90,15 @@ case class AnormIndex(db: DB, openCon: Option[Connection] = None) extends Index
...
     // Ignore core annotations. Yay magic names!
-    if (List("cs", "cr", "ss", "sr", "ca", "sa").contains(annotation))
-      return Future.value[Seq[IndexedTraceId]](Seq())
-
-    val result:List[(Long, Long)] = value match {
-      // Binary annotations
-      case Some(bytes) => {
-        SQL(
-          """SELECT zba.trace_id, s.created_ts
+    if (List("cs", "cr", "ss", "sr", "ca", "sa").contains(annotation)) {
```
franklinhu: We should use `Constants.CoreAnnotations` and also add the client/server address annotations to that list.

mosesn: wasn't exactly sure what you were getting at--I was hesitant to modify Constants.CoreAnnotations because I wasn't sure why it didn't already have "ca" and "sa".

IceCreamYou:
Thanks! One of those "can't believe I missed that" moments... :)

On the "multiple exit points" question -- guard clauses are the only [recommended use](http://twitter.github.io/effectivescala/#Control%20structures-Returns) of returns, but I have no objection to doing it the more idiomatic way.

mosesn:
Good call!  I'm probably wrong then, since effective scala is generally very good advice.  However, I find that having many different exit points for my functions drives me nuts.

bmdhacks:
Strange that this even compiled before.  I wonder where the return call returned to, since there probably wasn't anything up-stack from that.  Perhaps it just evaporated the thread without fulfilling the promise.

PR #287, opened 2013-08-01
Fix MySQL reserved keyword error

IceCreamYou:
As reported by Alex Xu on the user list, "key" is a reserved word in MySQL and that is causing errors when the zipkin_binary_annotations schema is installed. This change fixes it.

### DIFF
--- a/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/DB.scala
+++ b/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/DB.scala
@@ -87,7 +87,7 @@ case class DB(dbconfig: DBConfig = new DBConfig()) {
         |  trace_id BIGINT NOT NULL,
         |  span_name VARCHAR(255) NOT NULL,
         |  service_name VARCHAR(255) NOT NULL,
-        |  key VARCHAR(255) NOT NULL,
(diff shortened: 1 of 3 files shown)


IceCreamYou:
Note that this is a breaking change -- that is, people who have installed Zipkin and are using the Anorm module already (e.g. with SQLite) will get errors after applying this change because their schema will not match the schema expected after this change. We should probably have a plan for how to approach schema migration.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/DB.scala
```diff
@@ -87,7 +87,7 @@ case class DB(dbconfig: DBConfig = new DBConfig()) {
         |  trace_id BIGINT NOT NULL,
         |  span_name VARCHAR(255) NOT NULL,
         |  service_name VARCHAR(255) NOT NULL,
-        |  key VARCHAR(255) NOT NULL,
+        |  a_key VARCHAR(255) NOT NULL,
```
bmdhacks: I'd prefer a more descriptive expansion.  annotation_key, annotation_value isn't that bad.

bmdhacks:
I'm OK with kicking the migrations can down the road.

PR #288, opened 2013-08-02
Enable url hash to link to specific services in aggregates

bmdhacks:


IceCreamYou:
I like this. I think changing the hash works better here than window.history.pushState and friends. LGTM.

PR #289, opened 2013-08-05
Update setup documentation now that bin/collector and bin/query support a DB engine argument

IceCreamYou:


elliottneilclark:
+1

PR #290, opened 2013-08-06
Fix the Ruby gem not honoring incoming X-B3 headers

IceCreamYou:
I keep getting an error about "uninitialized constant Trace" when I try to set up an environment that will talk to the gem, but I expect the code itself to work once I figure out how to test it.


bmdhacks:
Do you have the finagle ruby gems included in your environment?

PR #294, opened 2013-08-21
Enhancements to dependencies for enabling scalding jobs.

bmdhacks:
The primary change is removing util.Time since it cannot be serialized.


franklinhu:
+1, but :frowning:

bmdhacks:
Why sad face?

bmdhacks:
Looks like serialization was added to util.Time so I don't need this anymore

PR #299, opened 2013-09-14
web: UI improvements for zero-length spans

franklinhu:
Two minor improvements for spans with length zero:
- Draw span bars with small width instead of zero
- Show bar labels even if the span is of length zero


theatrus:
+1

bmdhacks:
Ship it!

`Ship It!`

PR #303, opened 2013-10-06
fix split grouped nodes

fulinlin924:
see https://github.com/twitter/zipkin/issues/300


bmdhacks:
Thanks for the contribution!

PR #308, opened 2013-10-17
anormdb: fix span merging

franklinhu:
Annotations and binary annotations were not joined with spans
predicated on span names. This caused for duplicate annotations to show
up in the UI.


bmdhacks:
Ship it!
`Ship it!`

PR #313, opened 2013-12-04
fix cassandra search issue. https://github.com/twitter/zipkin/issues/310

fulinlin924:
see https://github.com/twitter/zipkin/issues/310


fulinlin924:
withdraw and will resubmit the pull request

PR #314, opened 2013-12-05
fix cassandra search issue.

fulinlin924:
see https://github.com/twitter/zipkin/issues/310


fulinlin924:
sbt.ResolveException: unresolved dependency: org.scala-tools.testing#specs_2.9.2;1.6.9: not found

any comments?

bmdhacks:
lgtm

PR #318, opened 2013-12-14
make stores Closable

sprsquish:
Extend stores (and related parts) with Closable. This is the first and smallest step toward moving Zipkin to TwitterServer.

### DIFF
--- a/zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraAggregates.scala
+++ b/zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraAggregates.scala
@@ -16,7 +16,7 @@
 package com.twitter.zipkin.storage.cassandra
 
 import com.twitter.cassie._
-import com.twitter.util.{Time, Future, Return, Throw}
+import com.twitter.util.{Future, FuturePool, Return, Throw, Time}
 import com.twitter.conversions.time._
 import com.twitter.zipkin.storage.Aggregates
 import com.twitter.zipkin.conversions.thrift._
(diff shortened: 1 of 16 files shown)


johanoskarsson:
+1

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraAggregates.scala
```diff
@@ -42,7 +42,7 @@ case class CassandraAggregates(
   dependenciesCF: ColumnFamily[ByteBuffer, Long, gen.Dependencies]
 ) extends Aggregates {
 
-  def close() {
+  def close(deadline: Time): Future[Unit] = FuturePool.unboundedPool {
```
bmdhacks, line 45: Doesn't the anorm code already have it's own future pool?

bmdhacks:
I'm surprised none of these other storages like redis and cassandra have Closable compatible close methods.

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraAggregates.scala
sprsquish, line 45: ah.. didn't see that.

bmdhacks, line 45: Hbase might have one too... I haven't looked.

sprsquish, line 45: They're private to the Mapper trait.

sprsquish:
Redis and Cassie are behind the times, by quite a bit. They don't implement Closable, yet. And updating them isn't a yak I want to shave just yet.

sprsquish:
took this a different direction

Issue #196, opened 2012-11-02
web: better client/server differentiation


franklinhu:
Would be nice to be able to clearly differentiate between the parts of the span

codefromthecrypt:
closing this due to (most sincere) lack of activity

PR #319, opened 2013-12-16
highlight server duration on trace summary page

wadey:
It is helpful to tell at a glance how network latency affected a trace.
This change shows a highlight at the bottom of each span that contains
both client and server timings. The highlight shows what duration of the
span was spent in the server.

I arrived on this highlight style after a few other attempts. It keeps the
span text readable and uses the color for the next depth (which
hopefully makes it more intuitive).

Related to issue #196, but probably not everything envisioned.

example

### DIFF
--- a/zipkin-web/src/main/resources/public/css/application.css
+++ b/zipkin-web/src/main/resources/public/css/application.css
@@ -624,6 +624,18 @@ ul.traces li:hover .timestamp .label {
 .bars.depth5 { fill: #08306B;}
 .bars.selected { fill: #00A8C6; }
 
+/**
+ * Same as .bars, but shifted by 1
+ */
+.server-bars.filtered { fill: #F9F2E7; }
+.server-bars.depth0 { fill: #9ECAE1; }
+.server-bars.depth1 { fill: #4292C6; }
+.server-bars.depth2 { fill: #2171B5;}
+.server-bars.depth3 { fill: #08519C;}
+.server-bars.depth4 { fill: #08306B;}
+.server-bars.depth5 { fill: #C6DBEF; }
+.server-bars.selected { fill: #00A8C6; }
+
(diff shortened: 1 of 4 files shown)


johanoskarsson:
This seems useful to me and not a big change code/ui wise from what I can tell. I'll let someone that knows front end stuff chime in though.

bmdhacks:
Seems reasonable.  Perhaps we can trim the decimal precision down a bit though?
I'll let @sprsquish chime in though.  What about making this a tooltip?

sprsquish:
I played with this a little. I like the idea, but think it needs a bit more. Having the extra line only gives a sense of what's happening but no real data is provided. 

What about @bmdhacks' suggestion of a tooltip? Add a little popover that provides the timing breakdowns?

wadey:
I am not attached to the current approach and I'm willing to try out other ways to visualize this. I do think that it should be readily apparent and not hidden in a tool tip (you can already click the span to see the actual timestamps).

Inspiration from the Dapper paper:

```
Every RPC span is broken down further into time spent within a server
process (green) and time spent on the network (blue).
```

dapper example

bmdhacks:
God I want that +- thread expansion, not the janky pseudo-intellijent fake-spans we have now.

codefromthecrypt:
@wadey what's your current thinking on this? @eirslett etc?

I recognize this is over 1.5years old. should we pull this in or close it?

(closing in 3 days unless I hear back)

wadey:
I still think this is a good idea, but I need to re-do it in a way that makes sense for the new UI. I'll close this pull request and open a new one once I have done that (we are still running a very old version of Zipkin internally and we need to update to latest).

codefromthecrypt:
ack. thanks for the response!

PR #322, opened 2013-12-30
Corrected link to zipkinCore.thrift

frenchfrywpepper:
In the instrument.md documentation.  The zipkinCore.thrift file is no longer located under resources.


franklinhu:
+1 thanks for the fix!

PR #325, opened 2014-01-15
update web runner script

sprsquish:


bmdhacks:
#shipit
Note query and collector too when the time comes.

PR #326, opened 2014-01-15
zipkin storm: kafka spout and span scheme 

suchang:
This is to pave the way for zipkin storm jobs. A new project zipkin-storm is created to contain all the storm related sources. This pull request includes:
1. span scheme to parse the steam data
2. kafka spout to reading data from kafka queue
3. sbt changes to create standalone jar "zipkin-storm" for storm job submission.


zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
```diff
@@ -0,0 +1,64 @@
...
+ */
+class SpanScheme extends Scheme {
+  @transient private val log = Logger.get
+  lazy val deserializer = new BinaryThriftStructSerializer[gen.Span] {
+    def codec = gen.Span
+  }
+
+  override def deserialize(bytes: Array[Byte]) = {
+    val span = try {
+      Stats.time("deserializeSpanSpout") {
```
sprsquish: Are stats for this actually useful?

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
```diff
@@ -0,0 +1,64 @@
...
+import com.twitter.zipkin.gen
+import java.util
+import scala.collection.JavaConverters._
+
+/**
+ * Spout scheme to turn incoming data into Span fields
+ */
+class SpanScheme extends Scheme {
+  @transient private val log = Logger.get
+  lazy val deserializer = new BinaryThriftStructSerializer[gen.Span] {
```
bmdhacks, line 35: Can you biject into the actual span class instead of the scrooge generated one like we do in the hadoop source?

zipkin-storm/src/test/scala/com/twitter/zipkin/SpanSchemeSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+package com.twitter.zipkin.storm
+
+import org.specs.Specification
+import org.specs.mock.{ClassMocker, JMocker}
+import com.twitter.zipkin.gen.{Annotation, Endpoint, Span}
+import scala.collection.JavaConversions._
+
+class SpanSchemeSpec extends Specification with JMocker with ClassMocker {
```
sprsquish: This should be implemented in ScalaTest

bmdhacks:
Also I wouldn't mind getting somebody who knows storm to review this, because I'm only mildly aware of what I'm looking at.

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
suchang, line 35: Hmm, I can use gen.Span#toSpan to convert the scrooge generated one to common.Span.  Seems both approaches do the job here, but bijection will need extra dependencies and make the jar bigger for job submission. Did I missed any benefits bijection provided?

bmdhacks, line 35: You don't have to implement it in terms of the Bijection library (which is quite small) but I would prefer if the Scheme outputs common.Span instead of gen.Span.

project/Project.scala
```diff
@@ -367,5 +370,43 @@ object Zipkin extends Build {
...
+    mergeStrategy in assembly <<= (mergeStrategy in assembly) { (old) =>
+      {
+        case PathList("javax", "servlet", xs @ _*) => MergeStrategy.last
+        case PathList("org", "apache", xs @ _*) => MergeStrategy.last
+        case PathList("com", "twitter", xs @ _*) => MergeStrategy.last
+        case PathList("project.clj") => MergeStrategy.last
+        case x => old(x)
+      }
+    }
+  ).dependsOn(scrooge)
```
franklinhu, line 511: You probably want to also update the dependency for the kafka module up near line 270

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
```diff
@@ -0,0 +1,64 @@
...
+    def codec = gen.Span
+  }
+
+  override def deserialize(bytes: Array[Byte]) = {
+    val span = try {
```
franklinhu: Could make this a `Try` and then match on `Return` and `Throw`

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
```diff
@@ -0,0 +1,64 @@
...
+      }
+    }
+
+    span match {
+      case Some(s) => {
+        Seq(s.traceId, s.id, s.name, s.serviceName.getOrElse(""), s.isClientSide)
+            .asJava.asInstanceOf[util.List[java.lang.Object]]
+      }
+      case None =>
+        Seq("").asJava.asInstanceOf[util.List[java.lang.Object]]
```
franklinhu: Why do these need to be cast into `java.util.List[java.lang.Object]`?

franklinhu: Oh because of Storm. We may as well just create them as Java objects rather than doing the conversions?

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
suchang: Hm, seems Try is only available since Scala 2.10 while we are still on 2.9.2 now..

mosesn: The OG Try is [c.t.u.Try](https://github.com/twitter/util/blob/master/util-core/src/main/scala/com/twitter/util/Try.scala).

project/Project.scala
suchang, line 511: updated.

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
suchang: done

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
suchang: done

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
suchang: dropped this stats

zipkin-storm/src/test/scala/com/twitter/zipkin/SpanSchemeSpec.scala
suchang: changed to ScalaTest

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
suchang, line 35: Yes, it actually outputs the fields of comom.Span - Fields("traceId", "spanId", "name", "serviceName", "isClientSide"). We could expand the field list if needed by the jobs.

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
```diff
@@ -0,0 +1,55 @@
...
+ * Spout scheme to turn incoming data into Span fields
+ */
+class SpanScheme extends Scheme {
+  lazy val deserializer = new BinaryThriftStructSerializer[gen.Span] {
+    def codec = gen.Span
+  }
+
+  override def deserialize(bytes: Array[Byte]) = {
+    Try (deserializer.fromBytes(bytes).toSpan) match {
+      case s: Return[Span] => {
```
sprsquish: case Return(s) =>
that way you wont need to apply it every time.

In fact, sorry to bump you back and forth, but this is basically a try/catch block. There's no need for c.t.u.Try

Also, no need for the braces here or at the throw

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
```diff
@@ -0,0 +1,55 @@
...
+          util.Arrays.asList(
```
sprsquish: Import java.util.Arrays
Arrays.asList

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
```diff
@@ -0,0 +1,55 @@
...
+            s().traceId.asInstanceOf[java.lang.Long],
+            s().id.asInstanceOf[java.lang.Long],
+            s().name,
+            s().serviceName.getOrElse(""),
+            s().isClientSide.asInstanceOf[java.lang.Boolean])
```
sprsquish: You shouldn't need to cast here, were you getting compile errors?

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
```diff
@@ -0,0 +1,55 @@
...
+      }
+      case t: Throw[Span] => {
```
sprsquish: type param isn't necessary

zipkin-storm/src/test/scala/com/twitter/zipkin/SpanSchemeSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.twitter.zipkin.storm
+
+import org.scalatest._
+import com.twitter.zipkin.gen.{Annotation, Endpoint, Span}
+import scala.collection.JavaConversions._
+
+class SpanSchemeSpec extends FlatSpec {
```
sprsquish: For consistency please use FunSuite: https://github.com/twitter/zipkin/commit/34998bf2376da8a64747819cd083530b39670618#diff-cc6e36fb0d77281c22b57e2acc535f0bR26

suchang: Changed to FunSuite.

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
suchang: yes, I'm getting compile errors for Long and Boolean. String seems fine.

type mismatch;
[error]  found   : Long
[error]  required: java.lang.Object
[error] Note: an implicit exists from scala.Long => java.lang.Long, but
[error] methods inherited from Object are rendered ambiguous.  This is to avoid
[error] a blanket implicit which would convert any scala.Long to any AnyRef.
[error] You may wish to use a type ascription: `x: java.lang.Long`.

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
suchang: done

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
suchang: Changed back to try/catch block.

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
```diff
@@ -0,0 +1,60 @@
...
+      Arrays.asList(
+        s.traceId.asInstanceOf[java.lang.Long],
+        s.id.asInstanceOf[java.lang.Long],
+        s.name,
+        s.serviceName.getOrElse(""),
+        s.isClientSide.asInstanceOf[java.lang.Boolean])
+    } catch {
+      case e: Exception => {
+        log.warning(e, "Invalid bytes for deserializer")
+        Stats.incr("spout.invalid_bytes")
```
sprsquish: Are stats exposed anywhere?

zipkin-storm/src/test/scala/com/twitter/zipkin/SpanSchemeSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+  test("SpanScheme deserializes bytes to span" ) {
+      val spanRecovered = spanScheme.deserializer.fromBytes(bytes)
+      assert(spanRecovered === span)
+    }
+
+  test("SpanScheme return correct values of the fields") {
+      val expectedValues = Seq(12345, 666, "methodcall", "service", true)
+      val values = spanScheme.deserialize(bytes).toList
+      assert(expectedValues === values)
+    }
```
sprsquish: indentation nit

zipkin-storm/src/main/scala/com/twitter/zipkin/SpanScheme.scala
suchang: turns out storm provides build-in metrics. removing this.

zipkin-storm/src/test/scala/com/twitter/zipkin/SpanSchemeSpec.scala
suchang: fixed

sprsquish:
lgtm. use the script in bin to pull submit

sprsquish:
Hrm. Can you merge master? You'll probably have to fix the Project file.

PR #327, opened 2014-01-16
TwitterServer based Collector

sprsquish:
Also included is a ScribeReceiver and an example project that shows how
a collector listening to scribe and writing to cassandra would be wired
up.

No idea why this got closed with the last merge.


doc/examples/collector/build.sbt
```diff
@@ -0,0 +1,16 @@
+name := "my-collector"
+
+organization := "com.twitter"
+
+scalaVersion := "2.9.2"
+
+libraryDependencies ++= Seq(
+  "com.twitter" % "zipkin-cassandra" % "1.1.1-SNAPSHOT",
```
bmdhacks, line 8: Please bump the version.  I say 2.0 but I suck at reading the semver tea leaves.

zipkin-receiver-scribe/src/main/scala/com/twitter/zipkin/receiver/scribe/ScribeSpanReceiver.scala
```diff
@@ -0,0 +1,131 @@
...
+  ): SpanReceiver = new SpanReceiver {
+    val addr = InetSocketAddressUtil.toPublic(new InetSocketAddress(scribePort())).asInstanceOf[InetSocketAddress]
+
+    val zkNode: Option[Closable] = scribeZkPath.get.map { path =>
+      val nodeName = "%s:%d".format(addr.getHostName, addr.getPort)
+      val fullPath = path + "/" + nodeName
+      zkClient.createEphemeral(path, nodeName.getBytes)
+    }
+
+    val service = Thrift.serveIface(addr, new ScribeReceiver(scribeCategories().toSet, process, stats))
```
bmdhacks: Probably want to bind to 0.0.0.0, not just the public ip.  I agree with announcing the public ip, but bind to the wildcard.

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
```diff
@@ -0,0 +1,283 @@
+/*
+ * Copyright 2012 Twitter Inc.
```
bmdhacks, line 2: This is missing from the Project.scala file

doc/examples/collector/src/main/scala/com/twitter/my-collector/MyCollectorService.scala
```diff
@@ -0,0 +1,21 @@
...
+
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.util.Future
+import com.twitter.zipkin.cassandra.CassieSpanStoreFactory
+import com.twitter.zipkin.collector.{SpanReceiver, ZipkinQueuedCollectorServer}
+import com.twitter.zipkin.common.Span
+import com.twitter.zipkin.receiver.scribe.ScribeSpanReceiverFactory
+import com.twitter.zipkin.storage.WriteSpanStore
+
+object MyCollectorService
```
mosesn, line 11: Is this the new way to use Zipkin?  Do you think we'll be able to go in a direction where you don't need to write any code to deploy zipkin?

zipkin-collector/src/main/scala/com/twitter/zipkin/collector/ItemQueue.scala
```diff
@@ -0,0 +1,81 @@
...
+  maxSize: Int,
+  maxConcurrency: Int,
+  process: T => Future[_],
+  stats: StatsReceiver = DefaultStatsReceiver.scope("ItemQueue")
+) extends Closable with CloseAwaitably {
+  @volatile private[this] var running: Boolean = true
+
+  private[this] val queue = new ArrayBlockingQueue[T](maxSize)
+  private[this] val queueSizeGauge = stats.addGauge("queueSize") { queue.size }
+  private[this] val activeWorkers = new AtomicInteger
```
mosesn: new AtomicInteger(0) is a little more explicit

doc/examples/collector/src/main/scala/com/twitter/my-collector/MyCollectorService.scala
sprsquish, line 11: Yea. After I get all these things setup I'm going to write an example service that will allow you to exec say "./bin/run" and have a single instance started up with the receiver/collector and web listening for scribe and/or http and storing to sqlite.

zipkin-collector/src/main/scala/com/twitter/zipkin/collector/ItemQueue.scala
sprsquish: fixed

zipkin-collector/src/main/scala/com/twitter/zipkin/collector/ItemQueue.scala
```diff
@@ -0,0 +1,81 @@
...
+  private[this] val activeWorkerGauge = stats.addGauge("activeWorkers") { activeWorkers.get }
+  private[this] val maxConcurrencyGauge = stats.addGauge("maxConcurrency") { maxConcurrency }
+  private[this] val workers = (0 until maxConcurrency) map { _ =>
```
mosesn: This might be a little more clear as

Seq.fill(maxConcurrency)(FuturePool.unboundedPool { loop() })

Why do you want to hold references to these though?

sprsquish: Look to line 68. :)

mosesn: Ahh, I see, Future.join(workers).

zipkin-collector/src/main/scala/com/twitter/zipkin/collector/ItemQueue.scala
```diff
@@ -0,0 +1,81 @@
...
+  def close(deadline: Time): Future[Unit] = closeAwaitably {
+    running = false
+    Future.join(workers)
+  }
+
+  private[this] val QueueFull = Future.exception(new QueueFullException(maxSize))
+  private[this] val QueueClosed = Future.exception(new QueueClosedException)
+
+  def add(item: T): Future[Unit] =
+    if (!running)
```
mosesn, line 73: I'm not sure if you care about this, there's a small race condition here.  If running is true, it will pass this, but if a close happens in between, and then if running is set to false and the queue is empty, then the workers will die, and the item won't drain properly.

sprsquish, line 73: You're right. It's possible, but very unlikely.

zipkin-collector/src/test/scala/com/twitter/zipkin/collector/ItemQueueTest.scala
```diff
@@ -0,0 +1,101 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.twitter.zipkin.collector
+
+import com.twitter.util.{Await, Future}
+import java.util.concurrent.{CountDownLatch, TimeUnit}
```
mosesn, line 19: If you want, there's a c.t.u.CountDownLatch which uses twitter util classes/idioms.

zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
```diff
@@ -16,17 +16,38 @@
...
 import com.twitter.zipkin.common.Span
 import java.nio.ByteBuffer
 import scala.collection.mutable
 
-trait SpanStoreFilter extends Filter[Seq[Span], Unit, Seq[Span], Unit]
-
 trait SpanStore extends WriteSpanStore with ReadSpanStore
 
+object SpanStore {
+  type Filter = FFilter[Seq[Span], Unit, Seq[Span], Unit]
```
mosesn, line 29: This might be a little simpler as

``` scala
type Filter = SimpleFilter[Seq[Span], Unit]
```

zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
```diff
@@ -16,17 +16,38 @@
...
+}
+
+
+/**
+ * A convencience builder to create a single WriteSpanStore from many. Writes
```
mosesn: Convenience

zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
```diff
@@ -16,17 +16,38 @@
...
+ * will be fanned out concurrently. A failure of any store will return a failure.
```
mosesn, line 35: Is the premise that each store should retry by itself?

zipkin-receiver-scribe/src/main/scala/com/twitter/zipkin/receiver/scribe/ScribeSpanReceiver.scala
```diff
@@ -0,0 +1,131 @@
...
+ */
+trait ScribeSpanReceiverFactory { self: App with ZooKeeperClientFactory =>
+  implicit object flagOfURI extends Flaggable[URI] {
+    def parse(v: String) = new URI(v)
+    override def show(c: URI) = c.toString
+  }
+
+  val scribeAddr = flag("zipkin.receiver.scribe.addr", new InetSocketAddress(1490), "the address to listen on")
+  val scribeCategories = flag("zipkin.receiver.scribe.categories", Seq("zipkin"), "a whitelist of categories to process")
+  val scribeZkPath = flag("zipkin.receiver.scribe.zk.path", "", "the zookeeper URI to announce on. blank does not announce")
```
mosesn: should this be a URI instead of a string?

zipkin-receiver-scribe/src/main/scala/com/twitter/zipkin/receiver/scribe/ScribeSpanReceiver.scala
```diff
@@ -0,0 +1,131 @@
...
+import com.twitter.zipkin.collector.SpanReceiver
+import com.twitter.zipkin.common.Span
+import com.twitter.zipkin.conversions.thrift._
+import com.twitter.zipkin.gen.{LogEntry, ResultCode, Scribe, Span => ThriftSpan, ZipkinCollector}
+import com.twitter.zipkin.zookeeper._
+import com.twitter.zk.ZkClient
+import java.net.{InetSocketAddress, URI}
+
+/**
+ * A SpanReceiverFactory that should be mixed into a base ZipkinCollector. This provides `newScribeSpanReceiver` which
```
mosesn: line length?

zipkin-receiver-scribe/src/main/scala/com/twitter/zipkin/receiver/scribe/ScribeSpanReceiver.scala
```diff
@@ -0,0 +1,131 @@
...
+    val zkNode: Option[Closable] = scribeZkPath.get.map { path =>
+      val addr = InetSocketAddressUtil.toPublic(scribeAddr()).asInstanceOf[InetSocketAddress]
+      val nodeName = "%s:%d".format(addr.getHostName, addr.getPort)
+      val fullPath = path + "/" + nodeName
+      zkClient.createEphemeral(path, nodeName.getBytes)
+    }
+
+    val service = Thrift.serveIface(scribeAddr(), new ScribeReceiver(scribeCategories().toSet, process, stats))
+
+    val closer: Closable = zkNode map { Closable.sequence(_, service) } getOrElse { service }
```
mosesn, line 75: Not sure if this would be clearer as Closable.sequence(zkNode.toSeq :+ service: _*)--I think it's a toss-up.

zipkin-receiver-scribe/src/main/scala/com/twitter/zipkin/receiver/scribe/ScribeSpanReceiver.scala
```diff
@@ -0,0 +1,131 @@
...
+  process: Seq[Span] => Future[Unit],
+  stats: StatsReceiver = DefaultStatsReceiver.scope("ScribeReceiver")
+) extends Scribe[Future] {
+  private[this] val deserializer = new BinaryThriftStructSerializer[ThriftSpan] {
+    def codec = ThriftSpan
+  }
+
+  private[this] val log = Logger.get
+
+  private[this] val TryLater = Future.value(ResultCode.TryLater)
```
mosesn: I think it's idiomatic for these constants to be lower camel case, no?

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
```diff
@@ -0,0 +1,283 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+package com.twitter.zipkin.sampler
+
+import com.twitter.conversions.time._
```
mosesn, line 20: sort imports?

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
```diff
@@ -0,0 +1,283 @@
...
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicReference, AtomicInteger}
+import com.twitter.logging.Logger
+import com.twitter.app.App
+import com.twitter.zipkin.common.Span
+
+trait AdaptiveSampler { self: App with ZooKeeperClientFactory =>
+  private[this] val basePath = "/com/twitter/zipkin/adaptiveSampler"
+
+  val asElectionPath =
+    flag("zipkin.sampler.adaptive.electionPath", basePath + "/election", "ZK path to use for running leader elections")
```
mosesn: line length?

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
```diff
@@ -0,0 +1,283 @@
...
+import com.twitter.util._
+import com.twitter.zipkin.zookeeper._
+import com.twitter.zipkin.common.Span
+import com.twitter.zipkin.storage.SpanStore
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicReference, AtomicInteger}
+import com.twitter.logging.Logger
+import com.twitter.app.App
+import com.twitter.zipkin.common.Span
+
+trait AdaptiveSampler { self: App with ZooKeeperClientFactory =>
```
mosesn, line 29: there are a ton of flags here.  I don't think it's worth doing immediately, but can we consolidate them in the future?

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
```diff
@@ -0,0 +1,283 @@
...
+class IsLeaderCheck[T](
+  zkClient: ZKClient,
+  electionPath: String
+) extends (Option[T] => Option[T]) with Closable {
+  private[this] val isLeader = new AtomicBoolean(false)
+  private[this] val leadership = zkClient.offerLeadership(electionPath)
+  leadership.data.changes.register(Witness(isLeader.set(_)))
+
+  def apply(in: Option[T]): Option[T] =
+    in filter { _ => isLeader.get }
```
mosesn, line 168: What is going on here?  I don't understand.

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
```diff
@@ -0,0 +1,283 @@
...
+    in filter { _ => curRate.get > 0 }
+}
+
+class SufficientDataCheck[T](sufficient: Int) extends (Option[Seq[T]] => Option[Seq[T]]) {
+  def apply(in: Option[Seq[T]]): Option[Seq[T]] =
+    in filter { _.length >= sufficient }
+}
+
+class ValidDataCheck[T](validate: T => Boolean) extends (Option[Seq[T]] => Option[Seq[T]]) {
+  def apply(in: Option[Seq[T]]): Option[Seq[T]] =
```
mosesn, line 217: why is this named in?

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
```diff
@@ -0,0 +1,283 @@
...
+      val allow = permit.compareAndSet(true, false)
+      if (allow) timer.doLater(period) { permit.set(true) }
+      allow
+    }
+}
+
+class OutlierCheck(
+  reqRate: Var[Int],
+  requiredDataPoints: Int,
+  threshold: Double = 0.15
```
mosesn, line 238: where does this number come from?

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
sprsquish, line 168: com.twitter.common.zookeeper has a mechanism for running leadership elections with groups. If this particular instance has been elected leader it will allow Some to passthrough, otherwise it'll transform it to a None.

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
sprsquish, line 217: it's the incoming option.

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
sprsquish, line 238: from the current implementation

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
```diff
@@ -0,0 +1,283 @@
...
+
+  def apply(in: Option[T]): Option[T] =
+    in filter { _ =>
+      val allow = permit.compareAndSet(true, false)
+      if (allow) timer.doLater(period) { permit.set(true) }
+      allow
+    }
+}
+
+class OutlierCheck(
```
mosesn, line 235: Could you document how you decided on this outlier check?  What do you expect the distribution to look like?  What ratio do you expect that this will catch?

zipkin-collector/src/test/scala/com/twitter/zipkin/collector/ItemQueueTest.scala
sprsquish, line 19: Yea, it's thin enough that I don't care here.

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
```diff
@@ -0,0 +1,283 @@
...
+  def apply(in: Option[Seq[Int]]): Option[Seq[Int]] =
+    in filter { buf =>
+      buf.segmentLength(isOut(curRate.get, _), buf.length - requiredDataPoints) == requiredDataPoints
+    }
+
+  private[this] def isOut(curRate: Int, datum: Int): Boolean =
+    math.abs(datum - curRate) > curRate * threshold
+}
+
+object DiscountedAverage extends (Seq[Int] => Double) {
```
mosesn, line 252: It looks like you use the invariant here that apply will be called regularly.  Could you document that?

What makes this different from an exponential moving average?  Could you just use that instead?

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
```diff
@@ -0,0 +1,283 @@
...
+  def apply(vals: Seq[Int]): Double =
+    calculate(vals, 0.9)
```
mosesn, line 254: How did you choose 0.9?

zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
sprsquish, line 29: Unfortunately SimpleFilter is an abstract class and not a type alias. Thus any filter would have to be a SimpleFilter. I don't want to make that restriction.

zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
sprsquish, line 35: Yep. I'll clarify in the comments

zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
sprsquish: fixed

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/BoundedBuffer.scala
```diff
@@ -0,0 +1,43 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+package com.twitter.zipkin.sampler
+
+class BoundedBuffer[T: ClassManifest](size: Int) {
+  private[this] val buffer = new Array[T](size)
+  private[this] var curSize = 0
+  private[this] var bufferLoc = -1
```
mosesn: I had trouble figuring out what bufferLoc meant--could you call it bufferIdx, curIdx, headIdx, head or similar instead?

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/BoundedBuffer.scala
```diff
@@ -0,0 +1,43 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+package com.twitter.zipkin.sampler
+
+class BoundedBuffer[T: ClassManifest](size: Int) {
```
mosesn: Could we call this a RingBuffer?

zipkin-receiver-scribe/src/main/scala/com/twitter/zipkin/receiver/scribe/ScribeSpanReceiver.scala
sprsquish: Hold over from old code. s/URI/path/

zipkin-receiver-scribe/src/main/scala/com/twitter/zipkin/receiver/scribe/ScribeSpanReceiver.scala
sprsquish: I don't think so?

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/HttpVar.scala
```diff
@@ -0,0 +1,61 @@
...
+  private[this] val underlying = Var(default)
+
+  val self: Var[Double] with Extractable[Double] = underlying
+
+  HttpMuxer.addRichHandler("/vars/"+name, Service.mk[Request, Response] {
+    case req if req.method == HttpMethod.GET =>
+      req.response.contentString = underlying().toString
+      Future.value(req.response)
+
+    case req if req.method == HttpMethod.POST =>
```
mosesn, line 40: break this up

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
sprsquish, line 2: added

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/HttpVar.scala
```diff
@@ -0,0 +1,61 @@
...
+ *
+ */
+package com.twitter.zipkin.sampler
+
+import com.twitter.finagle.Service
+import com.twitter.finagle.http.{HttpMuxer, Request, Response}
+import com.twitter.util.{Extractable, Future, Var}
+import org.jboss.netty.handler.codec.http.HttpMethod
+
+class HttpVar(name: String, default: Double = 1.0) {
```
mosesn, line 30: It's worth noting that name is restricted by the HttpMux DSL

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
sprsquish, line 29: I don't think so. What would you suggest?

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/Sampler.scala
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * A sampler is essentially a function from Long => Boolean. Using a provided Var[Double] the sampler
+ * will calculate whether the provided value should be sampled or not.
+ */
+class Sampler(
+  rateVar: Var[Double],
+  stats: StatsReceiver = DefaultStatsReceiver.scope("Sampler")
+) extends (Long => Boolean) {
+  private[this] val rate = new AtomicReference[Double](1.0)
+  rateVar.changes.register(Witness(rate.set(_)))
```
mosesn: rateVar.changes.register(Witness(rate))

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/Sampler.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  private[this] val RateGauge = stats.addGauge("rate") { rate.get.toFloat }
+  private[this] val allowedCounter = stats.counter("allowed")
+  private[this] val deniedCounter = stats.counter("denied")
+
+  def apply(traceId: Long): Boolean = {
+    val curRate = rate.get
+
+    val allow = if (curRate == 1) true else {
```
mosesn: (curRate == 1) || {
...
}

might be a little more straight forward, it's a toss-up.

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/SpanSamplerFilter.scala
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * A sampling filter to be placed in front of a SpanStore. This takes a `sample` function that will
+ * calculate whether the span should be sampled based on its traceId. If a span has its `debug` flag
+ * set the function will be bypassed and the span stored regardless of the sampling rate.
+ */
+class SpanSamplerFilter(
+  sample: Long => Boolean,
+  stats: StatsReceiver = DefaultStatsReceiver.scope("SpanSamplerFilter")
+) extends SpanStore.Filter {
+  private[this] val DebugCounter = stats.counter("debugFlag")
```
mosesn, line 34: lower camel case?

mosesn, line 34: Having something scoped to the same name as a counter seems messy . . . like giving a file and a directory the same name.  Can you change one?

zipkin-sampler/src/test/scala/com/twitter/zipkin/sampler/SpanSamplerFilterTest.scala
```diff
@@ -0,0 +1,64 @@
...
+      Span(1, "svc", 123L, None, List.empty[Annotation], Seq.empty[BinaryAnnotation]),
+      Span(2, "svc", 123L, None, List.empty[Annotation], Seq.empty[BinaryAnnotation]),
+      Span(3, "svc", 123L, None, List.empty[Annotation], Seq.empty[BinaryAnnotation]),
+      Span(4, "svc", 123L, None, List.empty[Annotation], Seq.empty[BinaryAnnotation]))
+
+    var rcvdSpans = Seq.empty[Span]
+
+    val svc = new SpanSamplerFilter(_ > 2) andThen Service.mk[Seq[Span], Unit] { spans =>
+      rcvdSpans = spans
+      Future.Unit
```
mosesn: prefer Future.Done

zipkin-sampler/src/test/scala/com/twitter/zipkin/sampler/SpanSamplerFilterTest.scala
```diff
@@ -0,0 +1,64 @@
...
+    val spans = Seq(
+      Span(0, "svc", 123L, None, List.empty[Annotation], Seq.empty[BinaryAnnotation], true),
+      Span(1, "svc", 123L, None, List.empty[Annotation], Seq.empty[BinaryAnnotation], true),
+      Span(1, "svc", 123L, None, List.empty[Annotation], Seq.empty[BinaryAnnotation], false))
+
+    var rcvdSpans = Seq.empty[Span]
+
+    val svc = new SpanSamplerFilter(_  => false) andThen Service.mk[Seq[Span], Unit] { spans =>
+      rcvdSpans = spans
+      Future.Unit
```
mosesn: Future.Done

zipkin-zookeeper/src/main/scala/com/twitter/zipkin/zookeeper/ZooKeeperClient.scala
```diff
@@ -0,0 +1,235 @@
...
+    def close(deadline: Time): Future[Unit] = {
+      if (closed.getAndSet(true)) Future.Unit else pool {
+        val cmd = abdicate.getAndSet(null)
+        if (cmd != null) cmd.execute()
+        ()
+      }
+    }
+  }
+
+  def close(deadline: Time): Future[Unit] = {
```
mosesn, line 232: closeAwaitably?

zipkin-zookeeper/src/main/scala/com/twitter/zipkin/zookeeper/ZooKeeperClient.scala
```diff
@@ -0,0 +1,235 @@
...
+            abdicate.set(cmd)
+            data.update(true)
+          }
+        }
+        def onDefeated() { data.update(false) }
+      })
+    }
+
+    def close(deadline: Time): Future[Unit] = {
+      if (closed.getAndSet(true)) Future.Unit else pool {
```
mosesn, line 224: closeAwaitably?

zipkin-receiver-scribe/src/main/scala/com/twitter/zipkin/receiver/scribe/ScribeSpanReceiver.scala
mosesn: then do we need flagOfURI anymore?  what is it used for?

zipkin-receiver-scribe/src/main/scala/com/twitter/zipkin/receiver/scribe/ScribeSpanReceiver.scala
mosesn: [Scala style guide](http://docs.scala-lang.org/style/naming-conventions.html#values_variable_and_methods) says for classes, vals should be lower camel--I can't find any examples of this style (upper camel) except on objects.

Most of your vals in classes follow lower camel.  Is there something I'm missing that makes these vals special?

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
mosesn, line 29: I wonder if you can scrap force users to use your suffixes (/election, /requestRate etc) and just have them specify a base, and then replace windowSize, threshold etc with precision and accuracy metrics.  Not a fully baked thought, might not be possible with the distribution we're expecting.

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
mosesn, line 168: Ahh, could you add a comment?

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/BoundedBuffer.scala
mosesn: It actually looks like there's a RingBuffer in c.t.u, can you use that one?

doc/examples/collector/build.sbt
mosesn, line 8: +1 2.0

mosesn:
It used to be possible to run zipkin without ZK.  Are we now saying that it's impossible?  Maybe we should provide a built in in-memory ZK for stand-alone clusters?

sprsquish:
It's still possible. ZK is only required for certain uses.

zipkin-zookeeper/src/main/scala/com/twitter/zipkin/zookeeper/ZooKeeperClient.scala
sprsquish, line 224: the watch isn't awaitable

zipkin-zookeeper/src/main/scala/com/twitter/zipkin/zookeeper/ZooKeeperClient.scala
sprsquish, line 232: the watch isn't awaitable

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/SpanSamplerFilter.scala
sprsquish, line 34: This is pretty common for our services. A total count at the root then a count per X (in this case service name) under it.

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/BoundedBuffer.scala
sprsquish: perhaps we could use c.t.u.RingBuffer but they don't have the same semantics. this one is per ordered the way we need it to be and provides a thread safe interface

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
sprsquish, line 254: pre-existing code

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
sprsquish, line 252: This is a re-implementation of existing code. I don't yet know why the original authors chose this method.

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
sprsquish, line 235: also taken from the current implementation.

sprsquish:
Regarding the rate calculate, I'm trying to re-write this in a way that's pretty much a drop in replacement for what already exists. I plan on revisiting later. When we move to the new storage backend, it's likely we wont need to sample anyway.

zipkin-receiver-scribe/src/main/scala/com/twitter/zipkin/receiver/scribe/ScribeSpanReceiver.scala
sprsquish: eh. I just really hate that style.

doc/examples/collector/build.sbt
sprsquish, line 8: That'll happen in a separate commit

zipkin-receiver-scribe/src/main/scala/com/twitter/zipkin/receiver/scribe/ScribeSpanReceiver.scala
sprsquish: nope, that's going away too

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
sprsquish, line 29: Yea.. Initially I thought about just asking for a base path. I'm going to switch to a base path flag and pass the rest of the paths in to newAdaptiveSamplerFilter while providing defaults based on basePath. If someone really needs to use different paths they can do it when they create the filter.

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/BoundedBuffer.scala
sprsquish: I'll re-implement it in terms of the c.t.u.RingBuffer

mosesn:
LGTM

zipkin-receiver-scribe/src/main/scala/com/twitter/zipkin/receiver/scribe/ScribeSpanReceiver.scala
bmdhacks: I'm ok with all caps.

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/HttpVar.scala
```diff
@@ -0,0 +1,64 @@
...
+import com.twitter.util.{Extractable, Future, Var}
+import org.jboss.netty.handler.codec.http.HttpMethod
+
+/**
+ * A wrapper around a Var[Double] that exposes it to an HTTP endpoint
+ * at "/vars/`name`". The Var can be read via a GET request and
+ * updated via a POST request. The body of the POST will be used
+ * as the new value.
+ */
+class HttpVar(name: String, default: Double = 1.0) {
```
bmdhacks, line 30: This isn't used anywhere.

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/HttpVar.scala
```diff
@@ -0,0 +1,64 @@
...
+  private[this] val underlying = Var(default)
+
+  val self: Var[Double] with Extractable[Double] = underlying
```
bmdhacks: I have an issue with "self" as the name.  Self implies a handle to the class of the given object, and as an external api it's kinda misnamed.  I'd prefer something like apply.

zipkin-zookeeper/src/main/scala/com/twitter/zipkin/zookeeper/ZooKeeperClient.scala
```diff
@@ -0,0 +1,235 @@
...
+import com.twitter.common.quantity.{Time => CommonTime}
+
+trait ZooKeeperClientFactory { self: App =>
+  val zkServerLocations = flag("zipkin.zookeeper.location", Seq(new InetSocketAddress(2181)), "Location of the ZooKeeper server")
+  lazy val zkClient = new ZKClient(zkServerLocations())
+}
+
+trait ZkWatch[T] extends Closable {
+  val data: Var[T]
+}
```
bmdhacks, line 44: The above is so close to Var.observe (which returns a closable).  I don't think the semantics are different enough to warrant a whole new class.

zipkin-zookeeper/src/main/scala/com/twitter/zipkin/zookeeper/ZooKeeperClient.scala
```diff
@@ -0,0 +1,235 @@
...
+      register()
+
+      def close(deadline: Time): Future[Unit] = {
+        if (closed.getAndSet(true)) Future.Unit else {
+          zkClient(path).delete().unit
+        }
+      }
+    }
+
+  def watchData(path: String): ZkWatch[Array[Byte]] = new ZkWatch[Array[Byte]] {
```
bmdhacks, line 135: Continuing the idea from above, this is so close to Var.observe

zipkin-zookeeper/src/main/scala/com/twitter/zipkin/zookeeper/ZooKeeperClient.scala
sprsquish, line 44: observe has been deprecated in favor of changes. Though, if it weren't, I still wouldn't be able to use it. Outside of c.t.u.Var.scala it's pretty much impossible to create my own kind of Var and I need to be able to hook the close method to unset the watches.

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/HttpVar.scala
sprsquish: done

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/HttpVar.scala
sprsquish, line 30: No. It was brought in because a version of it existed in the old code. Seems reasonable that someone might want to provide a sampler that can be updated via the admin interface.

bmdhacks:
Ship It!

zipkin-collector/src/main/scala/com/twitter/zipkin/collector/ItemQueue.scala
```diff
@@ -0,0 +1,79 @@
...
+ * This queue is backed by ArrayBlockingQueue of `maxSize`. Items are processed by a number of workers
+ * limited by `maxConcurrency`. Each worker will pull an item from the queue and process it via the
+ * provided `process` function. Workers will Await on `process` thus backpressure is based on the workers'
+ * ability to fully process an item.
+ *
+ * On close the queue will drain itelf before completing the close Future.
+ *
+ * The queue can be awaited on and will not complete until it's been closed and drained.
+ */
+class ItemQueue[T](
```
franklinhu, line 38: I know this is parallel to the original code, but any downsides to just replacing this with an `AsyncSemaphore` and  `FuturePool`?

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
```diff
@@ -0,0 +1,294 @@
...
+
+trait AdaptiveSampler { self: App with ZooKeeperClientFactory =>
+  val asBasePath = flag(
+    "zipkin.sampler.adaptive.basePath",
+    "/com/twitter/zipkin/sampler/adaptive",
+    "Base path in ZooKeeper for the sampler to use")
+
+  val asUpdateFreq = flag(
+    "zipkin.sampler.adaptive.updateFreq",
+    30.seconds,
```
franklinhu, line 37: How are `Flag[Duration]` parsed on the command line? As millis? It may be good to have some documentation here so users don't have to go digging.

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/HttpVar.scala
franklinhu, line 30: I believe the original one was wired up to update the config values in ZK for convenience.

zipkin-zookeeper/src/main/scala/com/twitter/zipkin/zookeeper/ZooKeeperClient.scala
```diff
@@ -0,0 +1,235 @@
...
+import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
+import com.twitter.finagle.util.DefaultTimer
+import scala.collection.mutable
+import com.twitter.common.zookeeper._
+import com.twitter.zk.{Connector, ZkClient, ZNode}
+import com.google.common.base.Supplier
+import org.apache.zookeeper.{CreateMode, KeeperException, Watcher, ZooDefs}
+import scala.collection.JavaConverters._
+import com.twitter.common.quantity.Amount
+import com.twitter.common.quantity.{Time => CommonTime}
```
franklinhu: sort these imports

zipkin-collector/src/main/scala/com/twitter/zipkin/collector/ItemQueue.scala
sprsquish, line 38: how do you mean?

zipkin-zookeeper/src/main/scala/com/twitter/zipkin/zookeeper/ZooKeeperClient.scala
```diff
@@ -0,0 +1,235 @@
...
+trait ZooKeeperClientFactory { self: App =>
+  val zkServerLocations = flag("zipkin.zookeeper.location", Seq(new InetSocketAddress(2181)), "Location of the ZooKeeper server")
+  lazy val zkClient = new ZKClient(zkServerLocations())
+}
+
+trait ZkWatch[T] extends Closable {
+  val data: Var[T]
+}
+
+class ZKClient(
```
franklinhu, line 46: Any reason not to use one of the many existing libs for zk clients?

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/HttpVar.scala
sprsquish, line 30: You mean so you could update the value via the HTTP interface and that would update the value in ZooKeeper?

zipkin-zookeeper/src/main/scala/com/twitter/zipkin/zookeeper/ZooKeeperClient.scala
sprsquish, line 46: I am. This wraps util-zk.

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
sprsquish, line 37: I think it's intuitive. You can write the same thing you'd write in the code: 1.hour, 30.seconds, 100.milliseconds, etc.

zipkin-zookeeper/src/main/scala/com/twitter/zipkin/zookeeper/ZooKeeperClient.scala
sprsquish: sorted

sprsquish:
Last call for comments. I'd like to get this in then get the query service in. That will complete the core components.

PR #328, opened 2014-01-17
s/LoadedStatsReceiver/DefaultStatsReceiver/g

sprsquish:
should have used the DefaultStatsReceiver rather than the LoadedStatsReceiver directly.


sprsquish:
Rolled this into #327

PR #329, opened 2014-01-23
TwitterServer based query service

sprsquish:
this is based on the new collector work (tws-collector branch)


zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
```diff
@@ -35,7 +35,7 @@ object Trace {
 
 case class Trace(private val s: Seq[Span]) {
 
-  lazy val spans = mergeBySpanId(s).toSeq.sortWith {
+  lazy val spans = mergeBySpanId(s).toList.sortWith {
```
bmdhacks, line 38: Why the type change here?

zipkin-query/src/main/scala/com/twitter/zipkin/query/Constants.scala
```diff
@@ -0,0 +1,25 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+package com.twitter.zipkin.query
+
+import com.twitter.conversions.time._
+import com.twitter.util.Duration
+
+object Constants {
```
bmdhacks, line 22: Just a suggestion, but you could make this a package object.

zipkin-query/src/main/scala/com/twitter/zipkin/query/ThriftQueryService.scala
```diff
@@ -0,0 +1,307 @@
...
+    sortedTraceIds(traces.map(_.map(_.traceId)), limit, order)
+
+  private[this] def adjustedTraces(traces: Seq[Seq[Span]], adjusts: Seq[thrift.Adjust]): Seq[Trace] = {
+    val as = adjusts flatMap { adjusters.get(_) }
+    traces map { spans =>
+      as.foldLeft(Trace(spans)) { (t, adjuster) => adjuster.adjust(t) }
+    }
+  }
+
+  private[this] def padTimestamp(timestamp: Long): Long =
```
bmdhacks, line 81: This is unused

zipkin-query/src/main/scala/com/twitter/zipkin/query/ThriftQueryService.scala
```diff
@@ -0,0 +1,307 @@
...
+    timestamp + Constants.TraceTimestampPadding.inMicroseconds
+
+  private[this] def traceIdsIntersect(idSeqs: Seq[Seq[IndexedTraceId]]): Seq[IndexedTraceId] = {
```
bmdhacks, line 84: It'd be nice to bubble up methods like this into the Trace object, or at least in IndexedTraceId object.

bmdhacks:
This looks good, although reading this code makes me pine for a more unified data model for Spans/Traces/Annotations with perhaps a good Monoid definition or at least common manipulation methods for doing merging/intersecting/etc.

zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
sprsquish, line 38: Iterators are Seqs too and I wanted to force evaluation.

zipkin-query/src/main/scala/com/twitter/zipkin/query/ThriftQueryService.scala
sprsquish, line 81: oop. oversite.

zipkin-query/src/main/scala/com/twitter/zipkin/query/ThriftQueryService.scala
```diff
@@ -0,0 +1,307 @@
...
+  traceDurationFetchBatchSize: Int = 500,
+  stats: StatsReceiver = DefaultStatsReceiver.scope("ThriftQueryService"),
+  log: Logger = Logger.get("ThriftQueryService")
+) extends thrift.ZipkinQuery[Future] {
+
+  private[this] val MethodStats = stats.scope("PerMethod")
+
+  private[this] def opt[T](param: T): Option[T] = param match {
+    case null => None
+    case "" => None
```
franklinhu, line 44: case null | "" => None

zipkin-query/src/main/scala/com/twitter/zipkin/query/ThriftQueryService.scala
```diff
@@ -0,0 +1,307 @@
...
+          spanStore.getTraceIdsByName(qr.serviceName, None, qr.endTs, qr.limit) flatMap {
+            queryResponse(_, qr)
+          }
+
+        case slice :: Nil =>
+          querySlices(sliceQueries, qr) flatMap { ids => queryResponse(ids.flatten, qr) }
+
+        case _ =>
+          querySlices(sliceQueries, qr.copy(limit = 1)) flatMap { ids =>
+            val ts = ids.flatMap(_.map(_.timestamp)).min
```
franklinhu, line 176: This will blow up if `ids` is empty

zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
mosesn, line 38: [Iterators](http://www.scala-lang.org/api/current/#scala.collection.Iterator) are not Seqs, you should be safe.

sprsquish, line 38: toSeq on an Iterator is lazy by default.

sprsquish:
had to re-open this as #332

PR #330, opened 2014-01-29
fixes for cassie

sprsquish:
Exclude old versions of finagle and util from cassie's dependencies.
Switch endTs to be where the current setup expects it to be.

### DIFF
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
@@ -160,7 +160,7 @@ class InMemorySpanStore extends SpanStore {
         spans
     }) filter { span =>
       span.lastAnnotation match {
-        case Some(ann) => ann.timestamp <= endTs
+        case Some(ann) => ann.timestamp >= endTs
         case None => false
       }
     } filter(shouldIndex) take(limit) map { span =>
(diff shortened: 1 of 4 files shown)


bmdhacks:
shipit

zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
```diff
@@ -160,7 +160,7 @@ class InMemorySpanStore extends SpanStore {
         spans
     }) filter { span =>
       span.lastAnnotation match {
-        case Some(ann) => ann.timestamp <= endTs
+        case Some(ann) => ann.timestamp >= endTs
```
franklinhu, line 163: What's the reasoning behind this change? The endTs argument is supposed to be an upper bound for "get me all traces before this time" not after.

sprsquish, line 163: It's wrong in the current implementation. The current implementation uses it as the start of the slice [1][2]. I intend to fix it after moving to the new CassieSpanStore.

1) https://github.com/twitter/zipkin/blob/master/zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraIndex.scala#L147
2) https://github.com/twitter/cassie/blob/master/cassie-core/src/main/scala/com/twitter/cassie/ColumnFamily.scala#L148

sprsquish, line 163: @franklinhu am I missing something somewhere though?

PR #331, opened 2014-01-29
cleanup zipkin-web a bit

sprsquish:
Switch to new style of scrooge service types.
Use ThriftMux for the query client
Split ZipkinWeb into a factory that can be mixed into other Apps
Create a stats filter


franklinhu:
lgtm +1

bmdhacks:
me too.  shipit.

PR #334, opened 2014-02-06
ZooKeeperClient/Scribe fixes

sprsquish:
ZK: Allow for optional credentials
ZK: Check for the existance of a node before attempting to create it
Scribe: annouce the proper path (use addr as the node name)


suchang:
lgtm - shipit

bmdhacks:
shipit

PR #337, opened 2014-02-07
fixup collector example

sprsquish:


bmdhacks:
shipit.
Did this break travis before or do we not compile it normally?

sprsquish:
It did not break Travis. I need to add it. Docs in generally need a good updating.

PR #343, opened 2014-02-19
zipkin-web: Specs -> ScalaTest

sprsquish:
Plus re-write QueryExtractor a bit.


suchang:
LGTM - could you also please update the scalatest dependency in zipkin-storm for consistency?

PR #344, opened 2014-02-20
Added more tests to SpanStoreValidator

nshkrob:
Fixed InMemorySpanStore.getTracesDuration.
Print more info in tests.

### DIFF
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
@@ -219,7 +219,7 @@ class InMemorySpanStore extends SpanStore {
   }
 
   def getTracesDuration(traceIds: Seq[Long]): Future[Seq[TraceIdDuration]] = call {
-    traceIds.map { traceId =>
+    traceIds.flatMap { traceId =>
       val timestamps = spans.filter { span => span.traceId == traceId }.flatMap { span =>
         Seq(span.firstAnnotation.map { _.timestamp }, span.lastAnnotation.map { _.timestamp }).flatten
       }
@@ -228,7 +228,7 @@ class InMemorySpanStore extends SpanStore {
         None
       else
         Some(TraceIdDuration(traceId, timestamps.max - timestamps.min, timestamps.min))
-    }.flatten
+    }
   }
 
   def getAllServiceNames: Future[Set[String]] = call {
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
@@ -219,11 +219,14 @@ class InMemorySpanStore extends SpanStore {
   }
 
   def getTracesDuration(traceIds: Seq[Long]): Future[Seq[TraceIdDuration]] = call {
-    spans filter { span => traceIds.contains(span.traceId) } flatMap { span =>
-      span.duration map { d =>
-        TraceIdDuration(span.traceId, d, span.firstAnnotation.get.timestamp)
+    println(spans)
+    traceIds.map { traceId =>
+      val timestamps = spans.filter { span => span.traceId == traceId }.flatMap { span =>
+        Seq(span.firstAnnotation.map { _.timestamp }, span.lastAnnotation.map { _.timestamp }).flatten
       }
-    } toList
+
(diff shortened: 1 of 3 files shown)


zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
```diff
@@ -219,11 +219,13 @@ class InMemorySpanStore extends SpanStore {
...
-      span.duration map { d =>
-        TraceIdDuration(span.traceId, d, span.firstAnnotation.get.timestamp)
+    traceIds.map { traceId =>
+      val timestamps = spans.filter { span => span.traceId == traceId }.flatMap { span =>
+        Seq(span.firstAnnotation.map { _.timestamp }, span.lastAnnotation.map { _.timestamp }).flatten
       }
-    } toList
+
+      TraceIdDuration(traceId, timestamps.max - timestamps.min, timestamps.min)
+    }
```
sprsquish, line 231: It's possible for a span to not have any annotations in which case, this would blow up. FlatMap over traceIds and have the body return an Option[TraceIdDuration] based on the existence of annotations.

zipkin-common/src/main/scala/com/twitter/zipkin/storage/util/SpanStoreValidator.scala
```diff
@@ -143,11 +160,19 @@ class SpanStoreValidator(
...
     }
+  }
+
+    test("get traces duration - 2") {
+      val store = resetAndLoadStore(Seq(span4))
+      assert(Await.result(store.getTracesDuration(Seq(999))) == Seq(TraceIdDuration(999, 1, 6)))
+
+      Await.result(store.apply(Seq(span5)))
+      assert(Await.result(store.getTracesDuration(Seq(999))) == Seq(TraceIdDuration(999, 3, 5)))
+    }
```
sprsquish, line 172: why not roll this into the other test?

zipkin-common/src/main/scala/com/twitter/zipkin/storage/util/SpanStoreValidator.scala
```diff
@@ -65,18 +72,28 @@ class SpanStoreValidator(
...
+
+    val passedCount = results.count(x => x)
+    println("%d / %d passed.".format(passedCount, tests.size))
+
+    if (passedCount < tests.size) {
+      println("Failed tests for %s:".format(spanStoreName))
+      results.zip(tests) collect { case (result, (name, _)) if !result => println(name) }
+    }
+
+    assert(results.count(x => x) == tests.size)
```
sprsquish: This could be put in the if body (with an else clause)

zipkin-common/src/main/scala/com/twitter/zipkin/storage/util/SpanStoreValidator.scala
```diff
@@ -65,18 +72,28 @@ class SpanStoreValidator(
...
     val spanStoreName = newSpanStore.getClass.getName.split('.').last
-    tests foreach { case (name, f) =>
+    val results = tests map { case (name, f) =>
       println("validating %s: %s".format(spanStoreName, name))
       try {
         f(); println("  pass")
+        true
       } catch { case e: Throwable =>
-        passed = false
+        println("fail")
```
sprsquish: prefix the string with two spaces (like the "  pass" println on line 73).

nshkrob:
Fixed the issues above.

zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
```diff
@@ -219,11 +219,16 @@ class InMemorySpanStore extends SpanStore {
...
-    spans filter { span => traceIds.contains(span.traceId) } flatMap { span =>
-      span.duration map { d =>
-        TraceIdDuration(span.traceId, d, span.firstAnnotation.get.timestamp)
+    traceIds.flatMap { traceId =>
+      val timestamps = spans.filter { span => span.traceId == traceId }.flatMap { span =>
+        Seq(span.firstAnnotation.map { _.timestamp }, span.lastAnnotation.map { _.timestamp }).flatten
       }
-    } toList
+
+      if(timestamps.isEmpty)
```
sprsquish: space between "if" and "("

sprsquish:
after it passes ci, shipit

PR #346, opened 2014-02-26
Summingbird: serializer, injections and configs

suchang:
This PR provides additional components needed for summingbird jobs:
1) serialization for span and a monoid used by summingbird sumByKey()
2) injection scheme for span
3) config changes: the scala version is bumped to 2.9.3.

### DIFF
--- a/project/Project.scala
+++ b/project/Project.scala
@@ -83,7 +83,7 @@ object Zipkin extends Build {
   def zipkinSettings = Seq(
     organization := "com.twitter",
     version := zipkinVersion,
-    crossScalaVersions := Seq("2.9.2"),
+    crossScalaVersions := Seq("2.9.3"),
     scalaVersion := "2.9.3",
     crossPaths := false,            /* Removes Scala version from artifact name */
     fork := true, // forking prevents runaway thread pollution of sbt
--- a/project/Project.scala
+++ b/project/Project.scala
@@ -40,10 +40,12 @@ object Zipkin extends Build {
   val twitterServerVersion = "1.4.0"
   val algebirdVersion  = "0.4.0"
   val hbaseVersion = "0.94.10"
+  val summingbirdVersion = "0.3.2"
 
-  def finagle(name: String) = "com.twitter" %% ("finagle-" + name) % finagleVersion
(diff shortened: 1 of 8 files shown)


project/Project.scala
```diff
@@ -82,7 +84,7 @@ object Zipkin extends Build {
     organization := "com.twitter",
     version := zipkinVersion,
     crossScalaVersions := Seq("2.9.2"),
-    scalaVersion := "2.9.2",
+    scalaVersion := "2.9.3",
```
sprsquish, line 87: Is it not possible to build the as a 2.9.2 project with a few 2.9.3 dependencies? I ask because it seems like most of our deps are 2.9.2 while only the storm deps are 2.9.3.

suchang, line 87: I tried to build it against 2.9.2 but there's compile errors in the storm code complaining missing classes.

sprsquish:
LGTM. Someone more knowledgable of Storm should take a look though.

zipkin-storm/src/main/scala/com/twitter/zipkin/Serialization.scala
```diff
@@ -0,0 +1,53 @@
...
+    Injection.connect[(BatchID, V), (V, BatchID), Array[Byte]]
+
+  implicit val spanMonoid: Monoid[Span] = new Monoid[Span] {
+    def plus(l: Span, r: Span) = {
+      new Span(l.traceId, l.name, l.id, l.parentId,
+        l.annotations, l.binaryAnnotations, l.debug)
+    }
+
+    val zero = Span(0L, "", 0L, None, List.empty, Seq.empty, false)
+  }
```
bmdhacks, line 58: The above Monoid does not look correct to me.  I would think you'd use Span.mergeSpan to do plus.  Also, this zero is not commutative.  I wrote a span monoid in a discarded PR, if you want to see:

https://github.com/twitter/zipkin/pull/260

suchang, line 58: Fixed the Monoid for span. The plus will keep merging span as long as it's not invalid.

bmdhacks:
LGTM, but I suppose we'll know for sure when we actually run it in production

PR #347, opened 2014-02-26
Refresh the "Find a trace" section of the web ui

sprsquish:
This will start the migration to Bootstrap3 and Flight. JS package
management is handled with Bower, but packages are picked apart by hand
and committed directly to the repo so as to not require a full suite of
JS tools if someone wants to simply build and run the UI.


sprsquish:
Play around with the UI and example data: http://zipkin.io

bmdhacks:
neat

PR #348, opened 2014-03-01
fix Endpoint JSON serialization

sprsquish:
Getting the host address from an endpoint shouldn't require a round trip through the InetSocketAddress. Ensure any Endpoint being serialized to JSON is first wrapped by the JsonEndpoint.


bmdhacks:
shipit

bmdhacks:
Although once you write your own conversion for the ip addresses now you have to write unit tests for it :-)

sprsquish:
Moved EndpointSpec to EndpointTest, changed nothing.

There's a test already for getHostAddress

PR #350, opened 2014-03-01
zipkin-test -> zipkin-tracegen

sprsquish:
Test coverage is pretty good overall. To clarify the intent of the package it's been moved to zipkin-tracegen. 

TraceGen is set up to produce more varied service and span names by pulling them from a pool of lorem ipsum. Trace durations have also been randomized. This will be very handy for setting up example Zipkin instances.


bmdhacks:
LGTM but can you show me what this produces sometime at work?

sprsquish:
I'm still tweaking this actually. But yes.

PR #353, opened 2014-03-07
Add injection of Map[String, List[Long]] for span duration aggregation

suchang:
In order to support span duration in the summingbird job, an injection of Map[String, List[Long]] is added. Summingbird job can aggregate the duration of the spans(List[Long]) and come up with stats such as avg, p99 etc


sprsquish:
lgtm

bmdhacks:
Isn't that list going to get really large?  Shouldn't you reduce it into some statistical monoids as you read it?

suchang:
Yes, you are right. And I'm actually running the job using a monoid to compute the avg. But so far I've not been able to come up with the proper monoid for percentile :(

bmdhacks:
https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/QTree.scala

suchang:
aha, that's what I'm looking for. will take a look. thanks!

PR #355, opened 2014-03-09
move default adjusters to constants package

sprsquish:
Move the default adjusters to the constants package. This will make it much easier to use them in cases where we want to implement a query service outside of the server factory.


nshkrob:
Ship it!

PR #356, opened 2014-03-09
de-duplicate span records in anormdb

sprsquish:
The AnormDB data model is a bit wrong. Spans don't get stored as unique records. Because of this a span from the server and a span from the server get saved to the span table as two different records. Reading them out then causes duplicate data to be displayed in the UI.


nshkrob:
Tests pass, so Ship it!

bmdhacks:
Does this cause any problems of dropped info from one of the spans, like endpoint or service name data?

sprsquish:
it shouldn't. the spans have to be identical in order to be considered duplicates.

PR #357, opened 2014-03-09
re-write Endpoint.toString in terms of the cheaper getHostAddress

sprsquish:
Using toString from InetSocketAddress will cause a DNS lookup to resolve the IP to a hostname. An endpoint is defined as an IP, port, and service name, it's unnecessary for it to resolve the hostname when printing its string representation. A user should explicitly request the hostname via getInetSocketAddress if they are interested.


nshkrob:
Ship it!

PR #359, opened 2014-03-14
Add getBinaryAnnotation in Span

suchang:
Iterate through list of binaryAnnotations and return the one with the given key.


zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
```diff
@@ -81,6 +81,13 @@ case class Span(traceId: Long, name: String, id: Long, parentId: Option[Long],
   }
 
   /**
+   * Iterate through list of binaryAnnotations and return the one with the given key.
+   */
+  def getBinaryAnnotation(key: String): Option[BinaryAnnotation] = {
+    binaryAnnotations.find {ba => ba.key == key}
```
sprsquish: style: spaces before and after each brace.

suchang: fixed, thanks.

sprsquish:
shipit

PR #362, opened 2014-03-25
create an example all-in-one zipkin project

sprsquish:


zipkin-example/src/main/scala/com/twitter/zipkin/example/Main.scala
```diff
@@ -0,0 +1,43 @@
+package com.twitter.mycollector
+
+import com.twitter.finagle.Http
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.server.{Closer, TwitterServer}
+import com.twitter.util.{Await, Closable, Future}
+import com.twitter.zipkin.anormdb.AnormDBSpanStoreFactory
+import com.twitter.zipkin.collector.{SpanReceiver, ZipkinCollectorFactory}
```
nshkrob: ZipkinCollectorFactory, WriteSpanStore are unused.

nshkrob:
Neat! LGTM other than the extra imports.

bmdhacks:
Wow that's simple.  shipit!

PR #369, opened 2014-04-10
zipkin-hbase scans backwards

sethp-jive:
Since row keys for a particular service start at Long.MAX_VALUE at the epoch and grow towards 0 as time progresses, starting at 0 and scanning until a reversed timestamp actually finds traces after the specified date, and not before.

### DIFF
--- a/zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseIndex.scala
+++ b/zipkin-hbase/src/main/scala/com/twitter/zipkin/storage/hbase/HBaseIndex.scala
@@ -73,8 +73,8 @@ trait HBaseIndex extends Index {
 
     annoMappingFuture.flatMap { annoMapping =>
       val scan = new Scan()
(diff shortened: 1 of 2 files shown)


sprsquish:
This isn't merging cleanly. Can you merge master and push the changes?

sethp-jive:
It looks like only one bug was still valid in tip, so I've backed out the conflicting change. It looks like it merges cleanly locally to me now, but I'm not sure how to confirm that it'll work via the PR?

sprsquish:
I'll run it again. The web change was the only conflicting part, so this should be fine.

sprsquish:
thanks!

sprsquish:
@sethp-jive My bad for not waiting on TravisCI to finish. Can you check out the build failure? https://travis-ci.org/twitter/zipkin/jobs/22993905

sethp-jive:
Whoops, looks like a test failure – I'm on it.

sethp-jive:
Sorry it took so long – I opened #374 to fix the tests I broke.

PR #370, opened 2014-04-12
Fix zipkin-web's timezone problem

synk:
Zipkin-web has problem that browser's timezone is not dealt with correctly.
To fix it, "end time" should be passed as timestamp.


sprsquish:
thanks!

PR #372, opened 2014-04-15
Oss sync

sprsquish:
This syncs the internal view to the OSS view. Mostly the PR is so we can run it through TravisCI


nshkrob:
shipit!

suchang:
LGTM- shipit!

PR #375, opened 2014-04-15
a more robust merge script

sprsquish:
When we can't look up a user's name or email the script fails. This will cause it to default to the user's login.

### DIFF
--- a/bin/git-pull-request.rb
+++ b/bin/git-pull-request.rb
@@ -207,7 +207,9 @@ module PullRequest
       user_login = pull_request['user']['login']
       user_obj = Github.get("users/#{user_login}")
       user_name = user_obj['name']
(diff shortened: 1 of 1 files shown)


bin/git-pull-request.rb
```diff
@@ -207,7 +207,9 @@ def merge(number)
       user_login = pull_request['user']['login']
       user_obj = Github.get("users/#{user_login}")
       user_name = user_obj['name']
+      user_name = user_login if user_name.nil? || user_name.empty?
```
nshkrob, line 210: I think you can do this:
user_name = user_obj['name'] || user_login

sprsquish, line 210: that only works if the value is nil or false

nshkrob, line 210: I see. Shipit!

PR #381, opened 2014-04-21
New Trace UI

sprsquish:
This is a re-write of the main trace UI. I'll get an example up soon.

With this re-write all generation happens server side. Once rendered, we use flight to add behavior to the page. The page is rendered as styled divs instead of via d3. The page will now take up the full width of the browser, popups are also larger and clearer. Individual spans can be expanded and collapsed. Filtering of services has been simplified and expanded as well.


zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
```diff
@@ -69,34 +68,29 @@ case class Trace(private val s: Seq[Span]) {
...
   /**
    * Get the start and end timestamps for this trace.
    */
   def getStartAndEndTimestamp: Option[Timespan] = {
-    spans.flatMap {
-      s => s.annotations.map {
-        a => a.timestamp
-      }
-    } match {
+    spans.flatMap(_.annotations.map(_.timestamp)) match {
```
bmdhacks, line 91: If you're going to refactor this, it's begging for a for comprehension

zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
```diff
@@ -69,34 +68,29 @@ case class Trace(private val s: Seq[Span]) {
...
       case Nil => None // No annotations
-      case s@_ => Some(Timespan(s.min, s.max))
+      case s => Some(Timespan(s.min, s.max))
```
bmdhacks, line 93: Untyped cases are deprecated

zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
```diff
@@ -161,11 +155,8 @@ case class Trace(private val s: Seq[Span]) {
   /**
    * Get all the binary annotations in this trace.
    */
-  def getBinaryAnnotations: Seq[BinaryAnnotation] = {
-    spans.map {
-      _.binaryAnnotations
-    }.flatten
-  }
+  def getBinaryAnnotations: Seq[BinaryAnnotation] =
+    spans.map(_.binaryAnnotations).flatten
```
bmdhacks, line 159: isn't this just flatMap?

zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
```diff
@@ -68,9 +68,9 @@ trait ZipkinWebFactory { self: App =>
       ("/app/", handlePublic(resourceDirs, typesMap, webCacheResources())),
       ("/public/", handlePublic(resourceDirs, typesMap, webCacheResources())),
       ("/", addLayout andThen handleIndex(queryClient)),
-      ("/traces/:id", addLayout(webRootUrl()) andThen handleTraces),
-      ("/static", addLayout(webRootUrl()) andThen handleStatic),
-      ("/aggregates", addLayout(webRootUrl()) andThen handleAggregates),
+      ("/traces/:id", addLayout andThen handleTraces(queryClient)),
+      //("/static", addLayout(webRootUrl()) andThen handleStatic),
+      //("/aggregates", addLayout(webRootUrl()) andThen handleAggregates),
```
bmdhacks: What's your plan for this?

sprsquish: As discussed offline, we're working on a better version.

zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
sprsquish, line 159: yes! :)

zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
sprsquish, line 91: I think it's clear this way. Not everything needs to be a for comprehension.

zipkin-common/src/main/scala/com/twitter/zipkin/query/Trace.scala
sprsquish, line 93: fixed

bmdhacks:
LGTM

sprsquish:
You can play around with example data in the new UI here: http://zipkin.io/

bmdhacks:
I think I found a bug:
http://zipkin.io/traces/77e8019056bbae52?serviceName=semper
Click Expand All twice.  Now try to expand collapse 'vitae'

sprsquish:
So I can get a sense for what people will expect from this UI, what's the behavior you expect?

sprsquish:
This has been merged internally and will show up here soon.

PR #382, opened 2014-04-26
use binaryAnnotation hashCode to construct column name

fr0stbyte:
there is another createSpanColumnName in the CassieSpanStore that needs to use the binaryAnnotations hashCode


sprsquish:
Thanks! Pulled this internally. We're changing our patch/release process. So the fix will show up the next time we do a push.

PR #384, opened 2014-05-07
Add AUTH support, fixes #383.

### LINKED ISSUE #383: Support Redis AUTH (requirepass)
Redis is protected using requirepass, would it be possible to support this in zipkin-redis? Thanks!

analytically:


zipkin-redis/src/main/scala/com/twitter/zipkin/redis/IndexBuilder.scala
```diff
@@ -25,13 +26,15 @@ import com.twitter.zipkin.storage.Index
...
-  ttl: Duration = 7.days
+  ttl: Duration = 7.days,
+  authPassword: Option[String] = None
 ) extends Builder[Index] { self =>
 
   def ttl(t: Duration): IndexBuilder = copy(ttl = t)
 
   def apply() = {
     val client = Client("%s:%d".format(host, port))
+    authPassword.map(p => client.auth(StringToChannelBuffer(p)))
```
mosesn: client.auth is asynchronous, so this isn't in sequence.  flatMap is the sequence operator for Futures.  Here's one possible way of implementing this:

``` scala
val authed = authPassword.map(p => client.auth(StringToChannelBuffer(p))) getOrElse Future.Done
Await.result(authed before Future.value(new RedisIndex {
  val database = client
  val ttl = Some(self.ttl)
}), 10.seconds)
```

zipkin-redis/src/main/scala/com/twitter/zipkin/redis/IndexBuilder.scala
```diff
@@ -25,13 +26,15 @@ import com.twitter.zipkin.storage.Index
 case class IndexBuilder(
   host: String,
   port: Int,
-  ttl: Duration = 7.days
+  ttl: Duration = 7.days,
+  authPassword: Option[String] = None
```
mosesn, line 32: @sprsquish is zipkin following semver?

sprsquish, line 32: It's not released as a library yet. So we don't follow a versioning system very closely. However, it will.

suchang:
Seems Travis CI failed by OOM. You might want to increase the PermGen space for sbt in .travis.yml.

sprsquish:
pulled internally. thanks!

PR #387, opened 2014-05-20
Improve WriteQueueWorkers' pushback

sethp-jive:
For storage engines that do most of their work under the returned future (such as hbase),
having the WriteQueueWorker return immediately means that the write queue is not an
accurate representation of the amount of outstanding work.


sprsquish:
pulled internally. will show up here soon.

Thanks!

PR #389, opened 2014-05-21
Change thrift source folder

eirslett:
Set Scrooge settings so they point to the IDL
files in zipkin-thrift. Then we can compile IDLs without
relying on symlinks (which doesn't play well cross-platform)

The symlinks should be deleted, but I'm not familiar with Pants so I'm not sure what to do with the BUILD file...


sprsquish:
I've pulled this internally. Should show up here soon.

Thanks!

PR #391, opened 2014-06-12
switched to gen.Span and some minor improvements

fr0stbyte:


sprsquish:
Thanks for this @fr0stbyte. Can you make the change against master though? I made some edits on the original PR mostly to match the style of the rest of the project.

fr0stbyte:
I noticed that some code was looking strangely unfamiliar. 
Two questions : 
1. here https://github.com/twitter/zipkin/commit/6107d9b#diff-f70d68ae7ea4c895a87b3be96ca84af6R15 
   why is the trait moved down as an internal type of KafkaProcessor ? Decoder[Option[List[Span]]]] is used somewhere else in the code, plus the implementing collectors would use it. Can we keep it top level and use it throughout the project ? 
2. https://github.com/twitter/zipkin/pull/390/files#diff-edc7ec48aa7cab8b9004553848474e92R22 the flag is a `,` separated sequence of type `a=b`. Without the definition of the flag, how will the system understand `-zipkin.kafka.topics=a=1,b=2,c=3` ?

sprsquish:
1. It's never used as a trait. The process doesn't take a `KafkaDecoder` it takes a `Decoder[Option[List[Span]]]` (I think that's the right approach, btw). Having it as a trait doesn't really mean anything. It's used as a type alias, so I made that explicit. 
2. Flag already provides a map type: https://github.com/twitter/util/blob/master/util-app/src/main/scala/com/twitter/app/Flag.scala#L86

fr0stbyte:
Sure, not as a trait but as a type then ? 
Didn't know about the map type, thanks for that.

sprsquish:
Types have to be defined inside something. We could put it in a package object, but there's little different between doing that and leaving it where it is.

fr0stbyte:
Got it. Please check the updated diff.

fr0stbyte:
@sprsquish bump

sprsquish:
@fr0stbyte hey. sorry. I've been pretty buried. I don't think I'm going to have the change to get to this until next week.

sprsquish:
pulled internally will show up here soon.

PR #392, opened 2014-06-20
Make zipkin-web dist file runnable

gardleopard:
With this patch you can run zipkin-web with "java -cp libs/ -jar zipkin-web-1.2.0-SNAPSHOT.jar -zipkin.web.resourcesRoot=resources/" from /zipkin-web/dist/zipkin-web. Without this patch zipkin-web can't resolve its resources. 

This is useful when you want to run zipkin-web without going through sbt.

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -116,13 +116,21 @@ there is no reason why Scribe has to be the only one.
 
 ### Zipkin UI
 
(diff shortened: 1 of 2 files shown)


sprsquish:
Awesome! Thanks for this. Pulled internally. Will show up here soon.

PR #394, opened 2014-07-02
Fix incorrect parameters to Trace::TraceId.new

Oscil8:
An extra argument is needed by finagle-thrift v1.3.x; this was broken by commit 22a12cff2214e39fe948bd2c1b4a4790e0b5987c

### DIFF
--- a/zipkin-gems/zipkin-tracer/lib/zipkin-tracer.rb
+++ b/zipkin-gems/zipkin-tracer/lib/zipkin-tracer.rb
@@ -23,11 +23,11 @@ module ZipkinTracer extend self
     B3_REQUIRED_HEADERS = %w[HTTP_X_B3_TRACEID HTTP_X_B3_PARENTSPANID HTTP_X_B3_SPANID HTTP_X_B3_SAMPLED]
     B3_HEADERS = B3_REQUIRED_HEADERS + %w[HTTP_X_B3_FLAGS]
 
-    def initialize(app)
+    def initialize(app, config=nil)
       @app = app
       @lock = Mutex.new
 
-      config = app.config.zipkin_tracer
+      config ||= app.config.zipkin_tracer # if not specified, try on app (e.g. Rails 3+)
       @service_name = config[:service_name]
       @service_port = config[:service_port]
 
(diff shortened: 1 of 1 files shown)


Oscil8:
It appears that there are other tests that are failing occasionally in Travis? There aren't actually any tests for the gems.

sprsquish:
@Oscil8 TravisCI hasn't been working reliably for some time and the gem tests don't get run anyway.

I've been completely swamped the last couple weeks and will continue to be for the next couple. I'll take a closer look and pull this in when I can catch my breath. Thanks for the contribution!

zipkin-gems/zipkin-tracer/lib/zipkin-tracer.rb
```diff
@@ -87,6 +87,7 @@ def get_or_create_trace_id(env)
                          end
       trace_parameters[3] = (trace_parameters[3] == "true")
 
+      trace_parameters << ::Trace::Flags::EMPTY
```
franklinhu: Ideally we'd be able to support the debug flag here, but if not just make a note that we'd like to fix it in the future?

Oscil8: Sure, we could take it in as a parameter to get_or_create_trace_id, e.g.

   get_or_create_trace_id(env, flags = ::Trace::Flags::EMPTY)
      ...
      trace_parameters << flags

Although this still requires work to allow configuring the default value of that flag in the plugin.

Oscil8:
@franklinhu Let me know if you've got any additional comments.

Oscil8:
Ah, I see you probably wanted handling of the X-B3-Flags header described in https://github.com/twitter/zipkin/blob/master/doc/collector-api.md ; have an update about to come through.

franklinhu:
+1

sprsquish:
@Oscil8 Please rebase this against master. I tried pulling it in after committing your other change and it failed on me.

Oscil8:
Done, thanks.

sprsquish:
Thanks! Merged.

PR #397, opened 2014-07-06
Fix AnormIndex.getTracesDuration to get the duration of the entire trace.

synk:
If I understand correctly, `Index.getTracesDuration` should return the duration of the entire trace. But the current implementation of `AnormIndex.getTracesDuration` returns a span's duration rather than a trace's one. Because of this, `QueryService` can't sort correctly by duration.


sprsquish:
@synk is it possible to switch to AnormSpanStore? I'm starting to deprecate the split store/index classes.

synk:
I see.
thanks!

PR #398, opened 2014-07-09
Follow sampling rate setting

Oscil8:
Noticed that the sample rate setting has no effect on what gets sent to the collector; need to use ::Trace.should_sample? from the finagle-thrift gem.

### DIFF
--- a/zipkin-gems/zipkin-tracer/lib/zipkin-tracer.rb
+++ b/zipkin-gems/zipkin-tracer/lib/zipkin-tracer.rb
@@ -55,9 +55,9 @@ module ZipkinTracer extend self
     end
 
     def call(env)
-      id = get_or_create_trace_id(env)
       ::Trace.default_endpoint = ::Trace.default_endpoint.with_service_name(@service_name).with_port(@service_port)
       ::Trace.sample_rate=(@sample_rate)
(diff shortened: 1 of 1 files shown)


zipkin-gems/zipkin-tracer/lib/zipkin-tracer.rb
```diff
@@ -55,9 +55,9 @@ def initialize(app)
     end
 
     def call(env)
+      ::Trace.sample_rate=(@sample_rate)
       id = get_or_create_trace_id(env)
       ::Trace.default_endpoint = ::Trace.default_endpoint.with_service_name(@service_name).with_port(@service_port)
-      ::Trace.sample_rate=(@sample_rate)
```
franklinhu, line 60: We may as well move all the variable setting to the beginning of the method for consistency. Also, let's make a note that get_or_create_trace_id depends on sample_rate being already set.

Oscil8, line 60: Sounds good, I'll update my patch

Oscil8:
@franklinhu Let me know if you've got any additional comments.

franklinhu:
+1

sprsquish:
Pulled internally. Thanks!

PR #405, opened 2014-08-18
Add support for sbt-assembly to zipkin-{web,{query,collector}-service}

abesto:
Since the assembly plugin is already included, I assume this is not very far from the intended purpose. The code is longer than strictly necessary and has some repetition to make sure the change has only minimal impact. This can of course be changed by just adding everything to `defaultSettings`.

The conflicts for which I added the `first` merge strategy:
- `com/twitter/zipkin/query/adjusters/*.class` built for `zipkin-query` and `zipkin-query-core` differed
- `com/twitter/common/args/apt/cmdline.arg.info.txt.1` differs between `com.twitter.common.zookeeper/server-set/jars/server-set-1.0.72.jar` and `com.twitter.common/args/jars/args-0.2.6.jar`
- `commons-beanutils-core` has both versions `1.7.0` and `1.8.0` somewhere in the dependency tree; I didn't track down where exactly.

### DIFF
--- a/bin/prezi_build.sh
+++ b/bin/prezi_build.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+set -ueo pipefail
+
+releaseid=$(git rev-parse HEAD)
+if [ -z "$releaseid" ]; then
+    echo "git rev-parse HEAD did not give back anything useful, aborting" >&2
+    exit 1
+fi
+
+$(dirname $0)/sbt zipkin-{web,collector-service,query-service}/assembly zipkin-web/package-dist
+
+temp=$(mktemp -d)
+trap "cd $(pwd); rm -rf ${temp}" EXIT
+
+cp -v zipkin-{web,collector-service,query-service}/target/*-assembly-*-SNAPSHOT.jar $temp/
+cp -v zipkin-web/dist/zipkin-web.zip $temp/
+
+cd $temp/
+for f in *; do
+    sha256sum -b $f | cut -d' ' -f1 > $f.sha256
+done
+
+aws s3 cp --acl public-read --recursive . s3://prezireleases/zipkin/$releaseid/
--- a/project/Project.scala
+++ b/project/Project.scala
@@ -288,7 +288,7 @@ object Zipkin extends Build {
     Project(
       id = "zipkin-query",
       base = file("zipkin-query"),
-      settings = defaultSettings
+      settings = defaultSettings ++ assemblySettings
     ).settings(
       libraryDependencies ++= Seq(
         finagle("thriftmux"),
@@ -341,7 +341,7 @@ object Zipkin extends Build {
     Project(
       id = "zipkin-collector-scribe",
       base = file("zipkin-collector-scribe"),
-      settings = defaultSettings
+      settings = defaultSettings ++ assemblySettings
     ).settings(
       libraryDependencies ++= Seq(
         scroogeDep("serializer")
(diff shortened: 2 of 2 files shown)


abesto:
After a lengthy debugging session I realized this doesn't work for `zipkin-web` as the application fails to load the required resources. The `package-dist` method seems to work.

mosesn:
I suspect that the problem is that far-jarring screws up the META-INF files.  You could try to convince sbt-assembly to concatenate the META-INF files instead of obliterating the extras.

abesto:
Here's the output of `./bin/sbt --debug assembly 2>/dev/null | ag Merging | ag META-INF`:

```
[warn] Merging 'META-INF/MANIFEST.MF' with strategy 'discard'
[warn] Merging 'META-INF/DEPENDENCIES' with strategy 'discard'
[warn] Merging 'META-INF/MANIFEST.MF' with strategy 'discard'
[warn] Merging 'META-INF/DEPENDENCIES' with strategy 'discard'
[warn] Merging 'META-INF/services/com.twitter.finagle.Announcer' with strategy 'filterDistinctLines'
[warn] Merging 'META-INF/services/com.twitter.finagle.Resolver' with strategy 'filterDistinctLines'
[warn] Merging 'META-INF/services/javax.ws.rs.ext.MessageBodyWriter' with strategy 'first'
[warn] Merging 'META-INF/MANIFEST.MF' with strategy 'discard'
[warn] Merging 'META-INF/DEPENDENCIES' with strategy 'discard'
[warn] Merging 'META-INF/INDEX.LIST' with strategy 'discard'
[warn] Merging 'META-INF/services/javax.ws.rs.ext.MessageBodyReader' with strategy 'first'
[warn] Merging 'META-INF/MANIFEST.MF' with strategy 'discard'
[warn] Merging 'META-INF/DEPENDENCIES' with strategy 'discard'
[warn] Merging 'META-INF/services/com.twitter.finagle.Announcer' with strategy 'filterDistinctLines'
[warn] Merging 'META-INF/services/com.twitter.finagle.Resolver' with strategy 'filterDistinctLines'
[warn] Merging 'META-INF/services/javax.ws.rs.ext.MessageBodyWriter' with strategy 'first'
[warn] Merging 'META-INF/MANIFEST.MF' with strategy 'discard'
[warn] Merging 'META-INF/DEPENDENCIES' with strategy 'discard'
[warn] Merging 'META-INF/services/com.twitter.finagle.Announcer' with strategy 'first'
[warn] Merging 'META-INF/INDEX.LIST' with strategy 'discard'
[warn] Merging 'META-INF/services/com.twitter.finagle.Resolver' with strategy 'first'
[warn] Merging 'META-INF/services/javax.ws.rs.ext.MessageBodyReader' with strategy 'first'
```

I don't see anything that may be related to resources, but
- I may have messed up and didn't find all merges, I don't use `sbt` daily
- an interesting line may be in there, I just don't recognize it

Does anything stand out to you?

My gut feeling is that https://github.com/twitter/zipkin/blob/master/zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala#L54 and https://github.com/twitter/zipkin/blob/master/zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala#L191-L200 don't agree with assembly on where the resources should be. I tried passing in `-zipkin.web.resourcesRoot='/'` (since the resources are in the root of the fat-jar), which did nothing, so either I'm on the wrong track, or I messed up the fix.

mosesn:
with strategy "first" is probably wrong.  You probably want to merge with strategy "concat" if it's possible.

sprsquish:
Might be better altogether to switch to sbt-onejar: https://github.com/sbt/sbt-onejar

It wont create an uber jar of all classes, rather package the project into a single jar along with its libraries thus eliminating the conflicts.

abesto:
sbt-assembly already has special handling for stuff under META-INF: https://github.com/sbt/sbt-assembly/blob/0.9.2/src/main/scala/sbtassembly/Plugin.scala#L347-L360

I tried to change from discard to concat on manifest.mf, but then the build fails with `java.util.zip.ZipException: duplicate entry: META-INF/MANIFEST.MF` (which is pretty strange btw).

I feel this is starting to take more effort than what a fat jar is worth; the package-dist method is safe and stable. I'm completely open to trying out anything you guys come up with, but I don't plan to investigate this further.

abesto:
I messed up and created the PR from master. Closing to avoid confusion.

PR #406, opened 2014-08-23
Fix AnormSpanStore#getTracesDuration to get the duration of the entire trace.

synk:
SpanStore.getTracesDuration should return the duration of the entire trace. But the current implementation of AnormSpanStore.getTracesDuration returns a span's duration rather than a trace's one. Because of this, QueryService can't sort correctly by duration.

This PR is the same as https://github.com/twitter/zipkin/pull/397, but to `AnormSpanStore` instead of `AnormIndex`.

### DIFF
--- a/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
+++ b/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
@@ -322,17 +322,17 @@ class AnormSpanStore(
     }
 
   private[this] def byDurationSql(ids: Seq[Long]) = SQL("""
-    |SELECT trace_id, duration, created_ts
-    |FROM zipkin_spans
-    |WHERE trace_id IN (%s) AND created_ts IS NOT NULL
+    |SELECT trace_id, MIN(a_timestamp), MAX(a_timestamp)
+    |FROM zipkin_annotations
+    |WHERE trace_id IN (%s)
     |GROUP BY trace_id
   """.stripMargin.format(ids.mkString(",")))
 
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
This seems testable. mind adding one?

adriancole:
Here's a test I'm adding to `SpanStoreValidator`. main thing is to see how many of the other stores break

```scala
test("get traces duration with children") {
  val spans = Seq(
    span1,
    span2.copy(parentId = Some(span1.traceId)),
    span3.copy(parentId = Some(span1.traceId))
  )
  val storeWithChildren = resetAndLoadStore(spans)
  val durationOfGraph = Await.result(storeWithChildren.getTracesDuration(Seq(span1.traceId)))

  val timestamps = spans.flatMap(s => s.annotations).map(a => a.timestamp)
  assert(eq(durationOfGraph, Seq(TraceIdDuration(span1.traceId, timestamps.max - timestamps.min, timestamps.min))))
}
```

adriancole:
yeah looks like mongo and cassandra are also broke

codefromthecrypt:
Our existing test didn't catch this as `span5`'s minimum timestamp and maximum duration happened to also be the trace's minimum timestamp and maximum duration. I'm correcting the test to not hide problems like this.
https://github.com/twitter/zipkin/blob/master/zipkin-common/src/main/scala/com/twitter/zipkin/storage/util/SpanStoreValidator.scala#L237

codefromthecrypt:
ok I have a test coming and will also merge this.

codefromthecrypt:
added tests via 41a6b665c779fe67f72a9af47b8ade452c5c22c5 closed via e6e90b2930572d3c7b5363b935925ac36c0bbd82

PR #411, opened 2014-09-11
Feature/hadoop

eirslett:
For context, see https://groups.google.com/forum/#!topic/zipkin-user/Z89XJfC3goo
The commit messages explain a bit what's going on in the PR. It's basically just a copy of [this Scalding job](https://gist.github.com/bmdhacks/48f1108a5577dbd2fa3a) discussed in https://groups.google.com/forum/#!msg/zipkin-user/q_EMWaNbHCw/qBCHqgeL3bMJ, with a custom Scalding tap/sink built on top of zipkin-cassandra.

I chose to put the low-level hadoop implementation in zipkin-cassandra, since it's cassandra-specific. That also makes stuff easier if Zipkin is ever ported to the native Cassandra client, since Cassie is deprecated. (see https://groups.google.com/forum/#!topic/zipkin-user/Ul58F0WZfe4)

For some reason, I had to change some stuff in zipkin-scrooge to make it build with sbt. I'm not quite sure why. I included it in this PR as a separate commit. If one can build 3b1e4bf without a493c0b, then I guess one could drop that one.

The module is currently only built with sbt - it's missing pants BUILD files. Somebody with more pants experience has to add that, I guess.

There will also be a web UI, but I'll provide that as a second PR later on, so the PR doesn't grow too big!

There's a README showing how to set it up. It would be great if people who run Zipkin with Cassandra in production could battle-test this, and reply to the PR!

### DIFF
--- a/zipkin-aggregate/src/main/scala/com/twitter/zipkin/aggregate/ZipkinAggregateJob.scala
+++ b/zipkin-aggregate/src/main/scala/com/twitter/zipkin/aggregate/ZipkinAggregateJob.scala
@@ -1,0 +1,55 @@
+package com.twitter.zipkin.aggregate
+
+import java.util.Date
+
+import com.twitter.algebird.{Moments, Monoid, Semigroup}
+import com.twitter.scalding._
+import com.twitter.util.Time
+import com.twitter.zipkin.common.{Dependencies, DependencyLink, Service, Span}
+
+final class ZipkinAggregateJob
+   (args: Args) extends Job(args)
+{
+  val dateRange: DateRange = DateRange(new Date(0L), new Date)
+
+  @transient
+  val (extraConfig, spanSource) = SpanSourceProvider(args)
+
+  override def config = super.config ++ extraConfig
+
+  val allSpans = TypedPipe.from(spanSource)
+    .groupBy { span: Span => (span.id, span.traceId) }
+    .reduce { (s1, s2) => s1.mergeSpan(s2) }
+    .filter { case (key, span) => span.isValid }
+
+  val parentSpans = allSpans
+    .group
+
+  val childSpans = allSpans
+    .filter { case (key, span) => span.parentId.isDefined }
+    .map { case (key, span) => ((span.parentId.get, span.traceId), span)}
+    .group
+
+  val result = parentSpans.join(childSpans)
+    .map { case (_,(parent: Span,child: Span)) =>
+    val moments = child.duration.map { d => Moments(d.toDouble) }.getOrElse(Monoid.zero[Moments])
+    val dlink = DependencyLink(Service(parent.serviceName.get), Service(child.serviceName.get), moments)
+    ((parent.serviceName.get, child.serviceName.get), dlink)
+  }
+    .group
+    .sum
+    .values
+    .map { dlink => Dependencies(Time.fromMilliseconds(dateRange.start.timestamp), Time.fromMilliseconds(dateRange.end.timestamp), Seq(dlink))}
+    .sum
+
+  result.write(spanSource)
+}
+
+object SpanSourceProvider {
+  def apply(args: Args) : (Map[AnyRef,AnyRef], Source with TypedSource[Span] with TypedSink[Dependencies]) = args.required("source") match {
+    case "cassandra" => {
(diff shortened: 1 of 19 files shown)


eirslett:
By the way, if somebody's impatient to try out the frontend UI, you can run it (locally) from https://github.com/eirslett/zipkin/tree/feature/aggregate-web-ui. But the branch is experimantal and wil be rebased without notice!

zipkin-aggregate/src/main/scala/com/twitter/zipkin/aggregate/ZipkinAggregateJob.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  val result = parentSpans.join(childSpans)
+    .toTypedPipe
+    .map { case (_,(parent: Span,child: Span)) =>
+    val moments = child.duration.map { d => Moments(d.toDouble) }.getOrElse(Monoid.zero[Moments])
+    val dlink = DependencyLink(Service(parent.serviceName.get), Service(child.serviceName.get), moments)
+    ((parent.serviceName.get, child.serviceName.get), dlink)
+  }
+    .group
+    .reduce { (d1, d2) => Semigroup.plus(d1, d2) }
```
johnynek: this should be .sum which exactly means what you wrote.

zipkin-aggregate/src/main/scala/com/twitter/zipkin/aggregate/ZipkinAggregateJob.scala
```diff
@@ -0,0 +1,58 @@
...
+  val parentSpans = allSpans
+    .group
+
+  val childSpans = allSpans
+    .filter { case (key, span) => span.parentId.isDefined }
+    .map { case (key, span) => ((span.parentId.get, span.traceId), span)}
+    .group
+
+  val result = parentSpans.join(childSpans)
+    .toTypedPipe
```
johnynek: not usually needed and clutters things a bit (in my view).

zipkin-aggregate/src/main/scala/com/twitter/zipkin/aggregate/ZipkinAggregateJob.scala
johnynek: actually, if you are on a recent version of scalding, .group.sum == .sumByKey

eirslett: So line 41 and 42 should just be replaced with `.sumByKey`?

eirslett:
I updated the PR according to @johnynek's comments. Also had to bump versions of finagle and twitter-util to 6.20.0 because of some classpath crashes in zipkin-web.

eirslett:
@sprsquish @johnynek We've run this job in production for some time now. We had to override some memory settings though: https://github.com/finn-no/zipkin/commit/a75fe749ba6754f3128078d9e5b2117adc39d785
Is it possible to set these hadoop flags (for this job only) without "hardcoding" it in Scalding?

Is there anything else that needs to be fixed before you can merge the PR?

eirslett:
@sprsquish @johnynek Have you had time to review this? Is there something that has to be fixed in this PR? Or is it something that you will end up not merging eventually?

tdrisdelle:
I agree with eirslett, I'd think this is a valuable change that should be merged.

sprsquish:
@eirslett I just pushed a bunch of changes that bring the OSS version of Zipkin in line with our internal version. On top of which, I've updated our sync tools so we'll start to schedule our syncs with the other Twitter libraries (once a week). 

The TravisCI integration has also been fixed and should now be trustworthy.

The biggest change is probably that we've moved to scala 2.10.4 and we're more closely tracking the core Twitter libraries.

Please merge master and we'll take a look and get this merged in.

Thanks again for all your work.

eirslett:
Ok, I rebased it against master, but unfortunately, with the latest changes, the tests fail :-(
Could you please help me out a bit, if you see what the problem is?

sprsquish:
@eirslett looks like you got it fixed! I'll pull this internally and we'll have a look.

sprsquish:
Closed via e217c056
Thank you!

PR #412, opened 2014-09-14
Fix binary annotation search in hbase

synk:
zipkin-hbase's binary annotation search doesn't work. This patch picks out an annotation value from ByteBuffer using `Util.getArrayFromBuffer`.


synk:
@tdrisdelle Are you using HBase? We are using HBase as span storage in our production. And annotation search didn't work until this patch.  It seems that a wrong query is thrown into HBase. The reason is that  `bb` is array backed `ByteBuffer`, then `bb.array()` returns whole of backed array.

tdrisdelle:
I'm sorry for the confusion.... what I meant was "I've checked out this patch and tested it on my local build with my own HBase back end and now it works fine, why isn't it being merged?"

Thanks for providing the patch.

synk:
Thank you for your test!

codefromthecrypt:
sorry.. meant to credit you, but this commit was accidentally squashed into the one I made to fix the hbase build. Thanks all the same! 658f972d1f1389b52facc228c808a3844ef75ec1

PR #413, opened 2014-09-14
Show exact time of each annotations at SpanPanel

synk:
For investigation purpose, we often want to know not only duration but also the exact time of each spans. This patch shows their local date time at Span Panel. This is useful to investigate the cause of  irregularly long span by collating its time with server log file.


codefromthecrypt:
@eirslett you cool with this?

eirslett:
+1

codefromthecrypt:
okie dokie.. on the way!

codefromthecrypt:
Thanks for the help @synk! bcac2a322e0305c1fb43a75ebac8007337fdbd01

PR #418, opened 2014-10-24
Include web resource root in packaged web.sh

schlosna:
When running zipkin-web/scripts/web.sh from a built distribution (i.e. `${zipkin}/bin/sbt zipkin-web/package-dist`), attempting to access `http://<host>:8080/` currently results in an empty response. This change adds the proper resource root to the web.sh start script per `doc/install.md` and resolves this issue.


codefromthecrypt:
I checked internally and we aren't using this script, so I'm just going to merge this!

PR #424, opened 2015-01-12
Add web UI for viewing aggregated trace data

eirslett:
There we go - I've cleaned up the code a bit.
Tell me if there are any issues that have to be addressed! (code style etc)

**Commit message**

> After running the zipkin/aggregate hadoop job, dependency
> links are stored in the database. This allows us to draw a
> graph of the entire system.
> 
> Hovering over nodes will highlight which dependency links it has.
> Clicking a node will show a list of its dependency links, and
> from there you can click on a dependency to view data about the
> dependency link itself. (number of calls, mean time etc)
> 
> The width of an arrow corresponds to the amount of traffic
> between nodes (the number of calls made).


sprsquish:
Sweet!

Style issue: we use two spaces for indentation.

Is there a way to populate some example data so someone could see an example of this in action?

lycplus:
good!

eirslett:
If you want to populate test data:

1) Configure a local Zipkin instance running on Cassandra
2) Run zipkin-tracegen
3) Run the hadoop job
4) Look at the web UI - traces should now be generated.

Alternatively: I'll see if I can make a modified JSON dump of our production data and host it somewhere as a static page.

eirslett:
OK, I (think) I fixed the indentation now.

eirslett:
Ok, I guess it's cumbersome to set up hadoop and everything just for some sample data, so here's an alternative.
1) Start Zipkin with a local MySQL instance
2) Dump this into the database: https://gist.github.com/eirslett/302b070cf6bff9598420
3) You're ready to go!

sprsquish:
Great. Anyway to get this into the zipkin-example project? It's already possible to boot it up with the right flags set to populate it with example data. This could piggyback on that.

eirslett:
Yes, it would definitely be nice to dump that test data into the example project.
But I'm not quite sure how to solve the dependency links between modules: zipkin-anormdb probably shouldn't know about test data, since that's only relevant for zipkin-example. But AnormSpanStore doesn't expose a public API with "importSql" functionality, and I don't think it should? So how should we dump the data into Anorm? (Or, eventually, should we break any of the dependency barriers I mentioned?)

sprsquish:
Hrm. Eventually we'll need some sort of API for this. For now, maybe just put the data in zipkin-tracegen along with some documentation on what to do with it.

eirslett:
com.twitter.zipkin.storage.Aggregates. has a storeDependencies method. I could hardcode the test data in a AggregateTestData Scala object and put that either in zipkin-example or zipkin-tracegen? Then the test data should also work independent of which database backend you're using.

eirslett:
I commited the SQL test dump into zipkin-tracegen for now, let me know if you want me to convert it to a Scala dump!

sprsquish:
Okay. Pulled internally.

srapp:
@sprsquish Is there any intention to pull this into the open source repo?  Seems like a really cool feature that many users could benefit from!

jollychang:
any update?

codefromthecrypt:
mental note check to see if it is indeed pulled internally

codefromthecrypt:
good news and bad news... good news is that OSS is synced to our internal stuff! bad news is that the build is broke, most likely due to us not using sbt to build. carrot to the one who fixes first!

PR #425, opened 2015-02-05
BASH_SOURCE is a bashism, we should use bin/bash

baloo:
I did change the java call with an exec, shell should not lie around
once it did forward the args

Signed-off-by: Arthur Gautier baloo@gandi.net

### DIFF
--- a/zipkin-collector-service/src/scripts/collector.sh
+++ b/zipkin-collector-service/src/scripts/collector.sh
@@ -1,5 +1,5 @@
-#!/bin/sh
+#!/usr/bin/env bash
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
don't mean to be bikesheddy, but do you strongly prefer /bin/bash vs /usr/bin/env bash?

http://stackoverflow.com/questions/21612980/why-is-usr-bin-env-bash-superior-to-bin-bash

baloo:
@adriancole fixed

codefromthecrypt:
Thanks! can you rebase this and squash into a single commit? I'll merge after

codefromthecrypt:
thx.. will get this through now!

codefromthecrypt:
thanks so much!

closed via d09d50b8fb3cdb75f11bffc93825f472c8193791

PR #426, opened 2015-02-05
Bump sbt version for packaging purpose

baloo:
For packaging purpose, i'd like to

```
set every version = "1.2.0~$date+$commitid"
```

which is buggy on 0.13.6 yet fixed in 0.13.7

see: https://github.com/sbt/sbt/issues/1430

Signed-off-by: Arthur Gautier baloo@gandi.net

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=0.13.6
+sbt.version=0.13.7


codefromthecrypt:
looks like other Twitter things are on 0.13.8 want to bump it again?
https://github.com/twitter/scrooge/blob/master/project/build.properties

codefromthecrypt:
closed via b8183ffcd94f6932b44d382fda2eb009df9d9d13

PR #427, opened 2015-02-16
Bump zipkin-tracer gem minor version to 0.2.0

pje:
Since `VERSION` hasn't been updated since 2012.

Also: the [version of the gem hosted on Rubygems](https://rubygems.org/gems/zipkin-tracer) is 0.0.1, which was the initial release of the gem. I see that @franklinhu is currently the only maintainer of it—it'd be great to see it updated.

### DIFF
--- a/zipkin-gems/zipkin-tracer/lib/zipkin-tracer/version.rb
+++ b/zipkin-gems/zipkin-tracer/lib/zipkin-tracer/version.rb
@@ -1,4 +1,4 @@
-# Copyright 2012 Twitter Inc.
+# Copyright 2015 Twitter Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/zipkin-gems/zipkin-tracer/lib/zipkin-tracer/version.rb
+++ b/zipkin-gems/zipkin-tracer/lib/zipkin-tracer/version.rb
@@ -12,6 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 module ZipkinTracer
-  VERSION = "0.1.0"
+  VERSION = "0.2.0"
 end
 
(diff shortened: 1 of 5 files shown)


zipkin-gems/zipkin-tracer/lib/zipkin-tracer/version.rb
```diff
@@ -1,4 +1,4 @@
-# Copyright 2012 Twitter Inc.
+# Copyright 2015 Twitter Inc.
```
codefromthecrypt, line 1: nit: I don't think your change should impact copyrights. Typically projects only affect these when adding new files.

codefromthecrypt:
I'd recommend waiting until this closes, as it will likely affect who would be doing the next release https://groups.google.com/forum/#!topic/zipkin-dev/hug_bWIqdLY

rtyler:
@pje FWIW we've been cutting some forked versions to rubygems for a while using [lookout-zipkin-tracer](https://rubygems.org/gems/lookout-zipkin-tracer)

@jamescway is working on getting some of our changes incorporated into the [openzipkin](/openzipkin) repos

codefromthecrypt:
great. 

my 2p, not necessarily twitter's

I think this repo is too big anyway. If there's a standalone repo for gems, it makes sense for sooo many reasons. I'll try an socialize an internal switch once the gems are in openzipkin. Then, if we're lucky, we can just rm -rf this.

codefromthecrypt:
so here's latest thinking of mine. re-orging gems out of the repo might end up holding some good work hostage. It is probably best to go ahead and get something out. @rtyler you ok with this version number (for now)? we can always bump it again for openzipkin or similar.

codefromthecrypt:
closed via 1c22afb3bb502a49aa2463c39d714860a3a8efb5

PR #429, opened 2015-03-20
Addition of Redis SpanStore + Factory to enable Cake pattern usage

drpacman:
Hi,
I wanted to try out the Redis storage and couldn't find a SpanStore or SpanStoreFactory for it in the codebase. I have added one, including using the SpanStoreValidator tests. Also found a some minor inconsistencies in RedisIndex class when using the SpanStoreValidator to test
- It threw an exception if there were no annotations in the span
- It indexed core annotations which was inconsistent with the Cassandra/AnormDB implementations

Apologies the pull also includes a sample redis-example project based on your zipkin-example I was using to test with which is included in the Project.scala

### DIFF
--- a/zipkin-redis/src/main/scala/com/twitter/zipkin/redis/RedisSpanStoreFactory.scala
+++ b/zipkin-redis/src/main/scala/com/twitter/zipkin/redis/RedisSpanStoreFactory.scala
@@ -13,7 +13,7 @@ trait RedisSpanStoreFactory { self: App =>
   val redisPort = flag("zipkin.storage.redis.port", 6379, "Port for Redis")
   val redisTtl = flag("zipkin.storage.redis.ttl", 168, "Redis data TTL in hours")
 
-  def newRedisSpanStore(): SpanStore = {
+  def newRedisSpanStore(): RedisSpanStore = {
     val storage = StorageBuilder(redisHost(), redisPort(), redisTtl().hours)
     val index = IndexBuilder(redisHost(), redisPort(), redisTtl().hours)
     new RedisSpanStore(index.apply(), storage.apply())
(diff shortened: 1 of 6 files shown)


zipkin-redis/src/main/scala/com/twitter/zipkin/redis/RedisSpanStoreFactory.scala
```diff
@@ -0,0 +1,20 @@
+package com.twitter.zipkin.redis
+
+import com.twitter.app.App
+import com.twitter.conversions.time._
+import com.twitter.zipkin.storage.redis.RedisSpanStore
+
+/**
+ * Created by caporp01 on 19/03/2015.
```
codefromthecrypt, line 8: generally, I'd scrub comments like this. leave it to `git log`

codefromthecrypt:
I'd recommend waiting until this closes, as it affects where new features should land https://groups.google.com/forum/#!topic/zipkin-dev/hug_bWIqdLY

drpacman:
Thanks for the link, a re-energised community would be a good thing.

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisIndex.scala
```diff
@@ -148,22 +148,31 @@ trait RedisIndex extends Index {
...
-      val time = span.lastAnnotation.get.timestamp
-      val binaryAnnos: Seq[Future[Unit]] = span.serviceNames.toSeq flatMap { serviceName =>
-        span.binaryAnnotations map { binaryAnno =>
-          binaryAnnotationsListMap.add(
-            binaryAnnoStringify(binaryAnno, serviceName),
-            time,
-            span.traceId
-          )
-        }
+      span.lastAnnotation.map {
```
codefromthecrypt, line 151: I think I get what you are doing here (guard on no last annotation), but it would be more clear if there was a test.

codefromthecrypt:
@drpacman as we've already merged some other things, feel free to rebase and dust off for review

codefromthecrypt:
I'm going to go ahead and dust this off and merge it. thanks tons for doing this. I've heard others interested in redis.

codefromthecrypt:
closed via 3fdd6d63e4ce9dfdf02faac4b380826009bdaf51

PR #430, opened 2015-03-22
make the SQL compliant an unambiguous

postwait:
These changes make the SQL work with standards-compliant SQL implementations (namely postgres).

I've tested and they still work with the built-in sqlite implementation as well.


codefromthecrypt:
thanks for the fix and manually testing it!

Issue #407, opened 2014-08-27
Many duplicate traces shown in zipkin web UI using Cassandra. SQLite was OK


hzhaofb:
I've been testing my instrumentation using Zipkin service with SQLite persistence store. Everything works well. However, in preparation of production release, I switched to use Cassandra 2.0.9, I see  large number of duplicate traces in the web UI, all pointing to the same trace link. 
![screen shot 2014-08-27 at 4 00 31 pm](https://cloud.githubusercontent.com/assets/7967571/4068928/1f477df6-2e3e-11e4-8bb3-b9d5355790b1.png)

Any help would be greatly appreciated. 

Howard. 

BTW, what Cassandra version is officially supported?

hzhaofb:
Found following log showing duplicate traceIds in the ArrayBuffer.

error DEB [20140827-22:47:00.036] cassie: multiget_slice(Zipkin, [java.nio.HeapByteBuffer[pos=0 lim=8 cap=8], java.nio.HeapByteBuffer[pos=0 lim=8 cap=8], java.nio.HeapByteBuffer[pos=0 lim=8 cap=8], java.nio.HeapByteBuffer[pos=0 lim=8 cap=8], java.nio.HeapByteBuffer[pos=0 lim=8 cap=8], java.nio.HeapByteBuffer[pos=0 lim=8 cap=8], java.nio.HeapByteBuffer[pos=0 lim=8 cap=8], java.nio.HeapByteBuffer[pos=0 lim=8 cap=8], java.nio.HeapByteBuffer[pos=0 lim=8 cap=8], java.nio.HeapByteBuffer[pos=0 lim=8 cap=8]], ColumnParent(column_family:SpanNames), SlicePredicate(slice_range:SliceRange(start:, finish:, reversed:false, count:2147483647)), ONE)[0m
[...]

PR #431, opened 2015-03-24
Don't submit duplicate trace Ids

postwait:
Fixes issue #407


codefromthecrypt:
one day... we'll have nice tests which will catch things like this! Thanks.

PR #433, opened 2015-04-03
Use SBT shortcuts to define local repositories

mauhiz:
This fixes an issue where using SBT on Windows would consistently yield the error :

`[error]         URI is not hierarchical`


mauhiz:
sorry - even like this it does not fully work. Will submit a PR again when I got everything working.

PR #434, opened 2015-04-27
set scalaVersion to 2.10.4

fr0stbyte:
minor change, bringing the scala version in the examples to 2.10 inline with the rest of the project


codefromthecrypt:
I think our other stuff is at 2.10.5. This probably needs to be in a general tune-up PR.

https://github.com/twitter/scrooge/blob/master/project/Build.scala#L67

codefromthecrypt:
closed via b8183ffcd94f6932b44d382fda2eb009df9d9d13

PR #423, opened 2015-01-10
Use new Travis CI infrastructure

caniszczyk:
Should speed up the build a bit... http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

Signed-off-by: Chris Aniszczyk zx@twitter.com

PR #439, opened 2015-06-28
Constrains Travis to working configuration

codefromthecrypt:
The openjdk7 build doesn't work and #423 broke the one that worked. This constrains Travis to what can pass, allowing contributors to make use of it as opposed to be constantly distracted by a half-broke build.

In the future, we'll want to fixup the current build and re-add openjdk.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: scala
 
-sudo: false
-
 scala:
   - 2.10.5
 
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
closed via 86544f7ae39384864a9134fdf3eadaea63e59ce9

Note: this isn't guaranteed to fix anything

PR #443, opened 2015-07-01
Reduce default query limit from 100 to 10

noslowerdna:
We've found that the default of 100 trace results is somewhat excessive, and 10 is more reasonable (and potentially substantially better performing) in most cases. This pull request reduces the default query limit from 100 to 10. One notable downside to this change is that when two or more annotations are specified, the probability of a matching intersection being found can be decreased.

### DIFF
--- a/zipkin-web/src/main/scala/com/twitter/zipkin/web/Constants.scala
+++ b/zipkin-web/src/main/scala/com/twitter/zipkin/web/Constants.scala
@@ -16,5 +16,5 @@
 package com.twitter.zipkin.web
 
 object Constants {
-  val DefaultQueryLimit: Int = 100
+  val DefaultQueryLimit: Int = 10
 }


mosesn:
Historically, we've found that 100 works fine for us.  What is the problem you've been finding with 100?  And why do you think 10 should be the global default, rather than changing your setting to use 10 instead?

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
```diff
@@ -296,7 +296,7 @@ class Handlers(jsonGenerator: ZipkinJson, mustacheGenerator: ZipkinMustache) {
           ("annotationQuery" -> req.params.get("annotationQuery").getOrElse("")),
           ("services" -> svcList),
           ("spans" -> spanList),
-          ("limit" -> req.params.get("limit").getOrElse("100")))
+          ("limit" -> req.params.get("limit").getOrElse(Constants.DefaultQueryLimit.toString)))
```
codefromthecrypt, line 299: @mosesn regardless of the value of the constant, do you agree with this change?

mosesn, line 299: yes, this change is rad

noslowerdna:
> why do you think 10 should be the global default

For the majority of use cases we've encountered for Zipkin, we've found that 10 result traces is more than adequate to answer whatever question is being investigated. It seems like a situation where 100 results would be desired or necessary would be fairly rare (although now I'm curious if the typical usage pattern that we haven't yet employed very often is casting a wide net to have a higher chance of finding outliers with a higher duration? In our system we are more frequently wanting to find out, was the data received and where was it delivered to?) The order of magnitude difference in performance is also fairly compelling. For a query with no annotations, limit=100 takes about 1 to 1.5s but limit=10 takes about 50-75ms (as measured by the new timers separately added in https://github.com/twitter/zipkin/pull/444). So the web UI responsiveness is a lot snappier and looks a little less cluttered with no need for scrolling.

codefromthecrypt:
I can make the change internal to twitter to set our default to 100 on deployment, if necessary. This is less about what works in Twitter, and more about what is generally a better default. Let's get a tie-break?

wanna ping folks here and see what they feel about 100 vs 10? https://gitter.im/openzipkin/zipkin

mosesn:
My concern is that in big distributed systems, the long tail typically ends up very bad–if you end up in a bad situation 20% of the time in the first system, 20% of the time in the second system, and 20% of the time in the third system, and it's really bad when you see all of them, then you'll see it less than 1% of the time.  Using a sample of 10, it's going to be pretty unlikely that you'll see the problem.  For very small systems, which don't have as deep hierarchies, it's not as bad, but in that kind of situation, it calls into question why we use zipkin at all.

With that said, as a basic debugging tool, if you're assuming all queries are alike, 10 clearly makes sense.  My concern is that one of the main use cases of zipkin is for surfacing interesting traces, and it's difficult to do that with such a small sample-size.

I'd be interested in knowing:

A.  Why doesn't latency scale linearly with the number of items we're fetching?
B.  What are the main use-cases of zipkin, and which of them should we optimize for?
C.  How much do zipkin users care about latency?  What is acceptable?

codefromthecrypt:
Configuration is something that should be configurable. Regardless of
whether 100 is arbitrary or a well tuned best default, there is a bug here
we shouldn't lose sight of. Let's get that merged especially if this
conversation is going to get laborious.

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
codefromthecrypt, line 299: added via 31a8ef65c4153795d07f39944e191e1756aa70c9

codefromthecrypt:
ok all, once this rebases on master, it will only include the change to the default constant. hoping we can get closure on this.

noslowerdna:
Rebased.

rtyler:
:+1:

codefromthecrypt:
this should be configurable, as there are valid reasons for different defaults. I opened up this to ensure folks can get what they want https://github.com/openzipkin/zipkin/issues/462

codefromthecrypt:
@a-roberts we used to have 100 then set to 10. It has been 2 years, we can revisit the topic, but review ^^ first?

PR #445, opened 2015-07-05
Removes zipkin-query gem

codefromthecrypt:
The zipkin-query gem is out-of-date, untested and full of tech debt
such as TODOs and hard-coded hosts and paths. Internally at Twitter, I
can find no use of it. Its only change was one day after it was written,
over three years ago.

I tried fixing thrifts, regenerating code, etc., but I think it is clear
this effort could be put on more important tasks. Hence, let's remove
the zipkin-query gem.


mosesn:
+1, looks like it was useful for some zipkin maintainer's workflow, but if it's not useful for yours, reinventing it doesn't look that hard (and git will still have it in the future).

rtyler:
@adriancole LGTM, I'll create a new repo for the gem elsewhere

codefromthecrypt:
@rtyler thx the tracer gem and the browser plugin both seem like they're better off in their own repos

codefromthecrypt:
closed via 743e8470f3b4ed7c950a3610060807b6241e3b7c

codefromthecrypt:
FYI this gem is archived here https://github.com/openzipkin/zipkin-query

PR #452, opened 2015-07-14
button for taking to top of the page. For long trace list and long sp…

nikhil2406:
…an list traces.
We generally have traces with number of spans in range of 200-500. Page up button is useful for us. Hoping that it would be useful for others too. Attached screenshot shows this button on bottom right corner.
Thanks,
Nikhil.

screen shot 2015-07-14 at 2 47 31 pm


codefromthecrypt:
sorry no-one's looked at this yet. I don't have an opinion.

@eirslett wdyt?

eirslett:
+1 Great work!

PR #464, opened 2015-07-16
Adds zipkin.web.query.limit flag

### LINKED ISSUE #462: web DefaultQueryLimit is not overridable
`Constants.DefaultQueryLimit` is a hard-coded val. Seems at least this should be overridable by a flag, right?

https://github.com/openzipkin/zipkin/blob/master/zipkin-web/src/main/scala/com/twitter/zipkin/web/Constants.scala#L19

see #443

codefromthecrypt:
Adds `zipkin.web.query.limit` flag to zipkin-web, which allows users
to override the default value of 10.

closes #462


codefromthecrypt:
Note, I tried to follow existing convention.. this doesn't mean I love it :)

cc @noslowerdna @mosesn @rtyler

zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
```diff
@@ -55,12 +55,15 @@ trait ZipkinWebFactory { self: App =>
...
   val queryDest = flag("zipkin.web.query.dest", "127.0.0.1:9411", "Location of the query server")
+  val queryLimit = flag("zipkin.web.query.limit", 10, "Default query limit for trace results")
+
   def newQueryClient(): ZipkinQuery.FutureIface =
     Thrift.newIface[ZipkinQuery.FutureIface]("ZipkinQuery=" + queryDest())
 
   def newJsonGenerator = new ZipkinJson
   def newMustacheGenerator = new ZipkinMustache(webResourcesRoot(), webCacheResources())
-  def newHandlers = new Handlers(newJsonGenerator, newMustacheGenerator)
+  def newQueryExtractor = new QueryExtractor(queryLimit.apply())
```
mosesn: maybe say `queryLimit()` instead of `queryLimit.apply()`?

codefromthecrypt: > maybe just say queryLimit() instead of queryLimit.apply()?
> 
> thx!

mosesn:
lgtm, :shipit:

zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
codefromthecrypt: done

codefromthecrypt:
thx for the review!

rtyler:
_[shrug](https://www.youtube.com/watch?v=ussCHoQttyQ)_

codefromthecrypt:
confirmed this works by deploying to a less used zone in prod at twitter

codefromthecrypt:
oh yeah.. the syntax I verified is this `-zipkin.web.query.limit=100`

eirslett:
One thing that could potentially cause some confusion is that we have different ways of configuring collector/query (with .Scala files) than web (flags). Would it be worth the effort to rewrite to one "universal" method?

codefromthecrypt:
> One thing that could potentially cause some confusion is that we have
> different ways of configuring collector/query (with .Scala files) than web
> (flags). Would it be worth the effort to rewrite to one "universal" method?
> I noticed the inconsistency too. Refactoring how we do config in general is
> probably due as things have improved in the last few years. Care to open a
> top-level issue on how you think config ought be handled?

PR #467, opened 2015-07-17
Use SSL for maven.twttr.com repository

travisbrown:
We [turned on SSL](https://finagle.github.io/blog/2015/07/16/ssl-now-supported-for-maven-twttr-com/) for this repository this week, and unsecured access will no longer be available after September 30.


codefromthecrypt:
Thanks for the help, Travis!

:shipit:

Issue #440, opened 2015-06-28
Builds are flakey and take way too long


codefromthecrypt:
`./sbt clean test` is long on laptops and extremely long (like 30m) on travis. It often flakes on downloads, too. 

zipkin has a vast and uncached dependency tree. There are likely many ways to improve this.

rtyler:
Hopefully some of @jamescway's work for #457 will help alleviate this, but I think this is a separate task unto itself

rtyler:
I think @abesto's work moving us over to Gradle in  #503 has helped tremendously. I'm going to close this now

PR #470, opened 2015-07-17
Use newer container-based builders in Travis and test JDK8 too

### LINKED ISSUE #468: Migrate to newer, faster Travis CI infra
Looks like we're not using their newer builders, will send a PR imminently to switch us over

rtyler:
Fixes #468

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: scala
+sudo: false
 
 scala:
   - 2.10.5
@@ -9,6 +10,7 @@ services:
 
 jdk:
   - oraclejdk7
+  - oraclejdk8
 
 # These directories are cached to S3 at the end of the build
 cache:


codefromthecrypt:
looking at :8ball: to see if this build will pass

codefromthecrypt:
lgtm on green (and even if not green, I guess.. as long as it is not worse than before :))

rtyler:
Well, the build doesn't look _worse_ so that's something? #440 is still alive and well, these builds failed with:

```
[warn]     ::::::::::::::::::::::::::::::::::::::::::::::
[warn]     ::              FAILED DOWNLOADS            ::
[warn]     :: ^ see resolution messages for details  ^ ::
[warn]     ::::::::::::::::::::::::::::::::::::::::::::::
[warn]     :: io.netty#netty;3.10.1.Final!netty.jar(bundle)
[warn]     ::::::::::::::::::::::::::::::::::::::::::::::
```

:crying_cat_face:

travisbrown:
It's likely you'll run into [this issue](https://github.com/travis-ci/travis-ci/issues/3120), which I reported in January after trying this for Finagle. As far as I know it's not been fixed yet. If you're willing to remove `-Djava.net.preferIPv4Stack=true` from your `sbt` runner that should work as well.

rtyler:
@travisbrown I don't think I understand the pattern that links our build issues with the ones you link. We're not seeing JVMs crater, we're just seeing inconsistencies with the ability of a Travis builder to fetch the same dependency the N times the `sbt` build tries to grab it.

travisbrown:
@rtyler Right, I'm not claiming this is responsible for the current build issues—just that if you switch to the containerized build system, you'll get a buffer overflow if you call `getHostName` anywhere in your tests.

codefromthecrypt:
@travisbrown https://github.com/openzipkin/zipkin/pull/472 on the getHostName thingy

PR #472, opened 2015-07-17
Removes -Djava.net.preferIPv4Stack=true to dodge a buffer overflow

codefromthecrypt:
Via @travisbrown, we learned that travis can buffer overflow if we call
`getHostName` in tests. We do that, so pre-emptively cutting the
`-Djava.net.preferIPv4Stack=true` flag to dodge a potential buffer
overflow.

See https://github.com/travis-ci/travis-ci/issues/3120

### DIFF
--- a/bin/sbt
+++ b/bin/sbt
@@ -24,7 +24,6 @@ test -f ~/.sbtconfig && . ~/.sbtconfig
 java -ea                          \
   $SBT_OPTS                       \
   $JAVA_OPTS                      \
-  -Djava.net.preferIPv4Stack=true \
   -XX:+AggressiveOpts             \
   -XX:+UseParNewGC                \
   -XX:+UseConcMarkSweepGC         \


eirslett:
+1

travisbrown:
I don't remember off the top of my head why Finagle includes this setting—there is a reason, although I'm not sure it's applicable here.

For Finagle and related projects, I'm just waiting to move to the containerized build system until @travis-ci fixes this bug, but this is probably also a reasonable solution.

mosesn:
finagle's unit tests assume that you've on ipv4–in particular, asking for localhost on ipv6 can give you one of four different results, but it's only one on ipv4.  finagle-memcached also spins up a real memcached, and we haven't gotten around to figure out how to get finagle to speak ipv4 to memcached when it's preferring ipv6.  I think zipkin once had a similar localhost issue, but my guess is that this is no longer the case, if tests pass reliably.

PR #473, opened 2015-07-17
Caches .m2 for travis runs

codefromthecrypt:
We've had builds fail downloading things from maven central. There's an
approach noted in https://github.com/travis-ci/travis-ci/issues/1441
that suggests caching `.m2`. Seems worth a shot.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ jdk:
 # These directories are cached to S3 at the end of the build
 cache:
   directories:
+   - $HOME/.m2
    - $HOME/.ivy2/cache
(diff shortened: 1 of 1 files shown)


jamescway:
:+1:  can't hurt!  based on what @eirslett was saying u won't be able to observe this behavior until u merge to master rite?

eirslett:
+1 go for it

travisbrown:
I don't know much about your build config, but if I remember correctly SBT won't put anything in `.m2` unless you're using the experimental (?) non-Ivy resolver.

One other thing you can try is putting `travis_retry sbt update` in a `before_script` script—this will retry resolution and downloading three times without adding retries the rest of the build.

PR #475, opened 2015-07-18
Deduplicate dependency definitions

abesto:
This is a first, opinionated stab at DRYing the dependency definitions in `Project.scala`. I aimed to make reading the file and updating dependencies easier. I chose some heuristics to get closer to this goal (this is the opinionated part):
- the version number for an artifact must appear exactly once in the file
- there should be minimal duplication of group ids, artifact ids and classifiers
- a shorter dependency list is easier to read
- a shorter file is easier to read

After the trivial stuff was out of the way, I made three biggish changes:
- Version numbers are not listed separately in stand-alone variables. They **do** appear in exactly one place, usually together with the group and artifact id.
- DRY'd the hadoop and hdfs dependencies similarly to how many others (finagle, util, ...) were already handled
- Instead of `Seq(finagle("core"), finagle("http"))` let's write `many(finagle, "core", "http")`

Take it apart :) I'll happily update the PR, both in the case of better heuristics and better / more idiomatic code.

Room for further improvement, in a separate PR: splitting into per-project `build.sbt` files would remove some boiler-plate, leading to cleaner, more readable, maintainable code. (Is there a reason not to do that?)

### DIFF
--- a/project/Project.scala
+++ b/project/Project.scala
@@ -28,37 +28,38 @@ import com.typesafe.sbt.site.SphinxSupport.Sphinx
 object Zipkin extends Build {
   val zipkinVersion = "1.2.0-SNAPSHOT"
 
-  val finagleVersion = "6.26.0"
-  val utilVersion = "6.25.0"
-  val scroogeVersion = "3.19.0"
-  val zookeeperVersions = Map(
+  def finagle(name: String) = "com.twitter" %% ("finagle-" + name) % "6.26.0"
+  def util(name: String) = "com.twitter" %% ("util-" + name) % "6.25.0"
+  def scroogeDep(name: String) = "com.twitter" %% ("scrooge-" + name) % "3.19.0"
+  def algebird(name: String) = "com.twitter" %% ("algebird-" + name) % "0.10.2"
+  def hbaseDep(name: String) = "org.apache.hbase" % ("hbase" + (if (name.isEmpty) "" else "-" + name)) % "0.98.3-hadoop2"
+  def hbaseTest(name: String) = hbaseDep(name) classifier("tests") classifier("")
+  def hadoop(name: String) = "org.apache.hadoop" % ("hadoop-" + name) % "2.4.0"
+  def hadoopTest(name: String) = hadoop(name) classifier("tests") classifier("")
+
+  def many(gen: (String => ModuleID), names: String*) = names map gen
+
+  val twitterZookeeperVersions = Map(
     "candidate" -> "0.0.41",
     "group" -> "0.0.44",
     "client" -> "0.0.35",
     "server-set" -> "1.0.36"
   )
-
-  val ostrichVersion = "9.9.0"
-  val algebirdVersion  = "0.10.2"
-  val scaldingVersion = "0.11.2"
-  val hbaseVersion = "0.98.3-hadoop2"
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
I'm happy to see effort here! Thanks so much for helping. I don't know what
is idiomatic is SBT, but clarifying and deduping sounds like a step
forward. I will test this against my WIP code for Kafka tomorrow AM Pacific
time.

project/Project.scala
```diff
@@ -28,37 +28,38 @@ import com.typesafe.sbt.site.SphinxSupport.Sphinx
 object Zipkin extends Build {
   val zipkinVersion = "1.2.0-SNAPSHOT"
 
-  val finagleVersion = "6.26.0"
-  val utilVersion = "6.25.0"
-  val scroogeVersion = "3.19.0"
-  val zookeeperVersions = Map(
+  def finagle(name: String) = "com.twitter" %% ("finagle-" + name) % "6.26.0"
```
eirslett, line 31: On alternative option is to accept  varargs, e.g.

``` scala
libraryDependencies ++= finagle("core", "http", "thrift")
```

which could be implemented with something like

``` scala
def finagle(name: String*) = name map { n => "com.twitter" %% ("finagle-" + n) % "6.26.0" }
```

abesto, line 31: Correct. I went with this because in quite a few places there's only a single package included, which would require another round of redundant concatenation / `flatten`.

eirslett:
Multi-module sbt projects often have monolith build definitions inside project/, I think one of the reasons is that it's easier to share common configuration between modules (versions of Scala, Finagle, for example)
I'm not sure if that's possible to do in any other way. I've had some multi-module projects with separate .sbt files in every module, which I eventually refactored to have everything in the same build file.

eirslett:
Oh, looks like you made it work! Disregard my comment ;-)

abesto:
Disregarded :)

On a more serious note: I have very limited actual experience with the long-term maintenance of multi-project builds. Do you foresee more problems with the many-`build.sbt`s approach, or is it just that it's extra effort to initially get it to work?

project/Project.scala
eirslett, line 31: You could use apply/splash:

``` scala
libraryDependencies ++= Seq(
  junit,
  finagle("core","thrift"):_*,
  hadoop("core"):_*
)
```

but many is fine as well.
I guess :_\* is also quite magic for people to read, so maybe it's better to keep it simple, like with `many`.

abesto, line 31: Ah, that's right, I didn't think of that. Yeah, it's a little bit of magic we'd sprinkle all over the place. I'm fine with that too, don't have a strong opinion at this point.

For now I think let's leave it like this, if anyone feels strongly about changing this we can always go back and do it.

eirslett:
mostly about the effort ;-)

codefromthecrypt:
is this obviated by #478

abesto:
@adriancole nope, #478 continues the work in this, but does very different things and has very different scope, so I thought a separate PR was in order.

project/Project.scala
codefromthecrypt, line 31: let's keep this for now, as it is progress.. thanks!

codefromthecrypt:
:shipit: as currently everything is hosed and this makes things better.

PR #477, opened 2015-07-19
Revert "Use newer container-based builders in Travis and test JDK8 too"

codefromthecrypt:
This reverts commit 88a800dd10e7cf82216956481c8223acb5d47752.

Rationale is that the build is too hosed and the rationale for using the newer containers is speed. Once we get anything passing, we can look at the optimization of using container builds. In the mean time, it is arduous and time consuming to troubleshoot this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: scala
-sudo: false
 
 scala:
   - 2.10.5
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
cc @rtyler @abesto

codefromthecrypt:
ps JDK 8 happened to be in the same commit, but I've no strong feelings about that.

rtyler:
bummer man

PR #478, opened 2015-07-19
Major refactor of Project.scala

abesto:
**Note**: this depends on #475 
### What, why

`Project.scala` was over 600 lines of boilerplate-heavy, hard-to-read code. This PR aims to improve the situation by moving what can be moved to `build.sbt` files under each project, and leaving clean, readable, DRY glue-code in `Project.scala`.
### Result

To my eyes, `Project.scala` is now much easier to read. Common code, dependencies and settings are collected and clearly separated from project definitions. The project declarations and the definition of inter-project dependencies are now clearly separated from the configuration of individual projects. The longest `build.sbt` is 29 lines of code, making the build configuration of each project quite approachable and easy to grasp, modify.

I expect this change to make the maintenance of the SBT-based build system much easier, and also to vastly simplify any attempts around migrating to Gradle.
### Testing

To test the changes, I ran `clean`, `compile` and `package` (where applicable) for each project, and compared the log before and after the changes. Also did a full `sbt test` in the end. All this may still mean that some things are broken by the refactor. I'm specifically unclear on how to test, and so unsure of the correctness, of:
- `addConfigsToResourcePathForConfigSpec`, wrt/ evaluation time (I expect it's re-evaluated in the context of each specific project)
- the `DocTest` configuration in the `doc/build.sbt`

I'd highly appreciate a close look at these specifically, and of course at the whole change generally. I did everything that came to mind around testing what I broke, but you know the saying about the number of bugs. Problem classes that may or may not be introduced by this change:
- an artifact is not build
- an artifact doesn't contain what it should / contains stuff it shouldn't
- some tests are not even started
- incomplete / missing generated documentation

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,0 +1,2 @@
+defaultSettings
+Unidoc.settings
--- a/doc/build.sbt
+++ b/doc/build.sbt
@@ -1,0 +1,29 @@
+import Tests._
+import com.typesafe.sbt.site.SphinxSupport.Sphinx
+
+defaultSettings
+Project.defaultSettings
+site.settings
+site.sphinxSupport()
+
+scalacOptions in doc <++= (version).map(v => Seq("-doc-title", "Zipkin", "-doc-version", v))
+
+includeFilter in Sphinx := ("*.html" | "*.jpg" | "*.png" | "*.svg" | "*.js" | "*.css" | "*.gif" | "*.txt")
+
+// A dummy partitioning scheme for tests
+def partitionTests(tests: Seq[TestDefinition]) = {
+  Seq(new Group("inProcess", tests, InProcess))
+}
+// Workaround for sbt bug: Without a testGrouping for all test configs,
+// the wrong tests are run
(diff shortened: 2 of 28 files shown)


build.sbt
```diff
@@ -0,0 +1,3 @@
+Project.defaultSettings
```
eirslett: Why do we have both `Project.defaultSettings` and `defaultSettings`?

abesto: "That's how I found it" :) Good point, `defaultSettings` includes the other one. I'll remove `Project.defaultSettings` in a sec.

eirslett:
Looks very good! +1 But maybe squash them into 1 commit?

abesto:
Thanks! How about having these three commits?

eirslett:
That looks good to me! :-) +1 for merge, let's just wait for another reviewer then

codefromthecrypt:
:shipit: (also prefer less commits, ideally 1)

PR #479, opened 2015-07-19
Fix Windows issue in ZipkinResolver

eirslett:
Problem:
The current resolver setup configured
a resolver URL that wasn't valid on Windows, leading
to build failures.

Solution:
Replace the resolver with one that is
cross-platform, using Java's APIs that are supposed
to work across platforms.

PS: I added som println statements to debug; **can anybody
on a Unix box please run "sbt" and see what pops up?**
It's important that it should say that the original resolver
and the cross-platform resolver point to the same URL
on a Unix box!!!!!
I will remove the println statements afterwards.

### DIFF
--- a/project/ZipkinResolver.scala
+++ b/project/ZipkinResolver.scala
@@ -49,7 +49,7 @@ object ZipkinResolver extends Plugin {
       }) ++ Seq(
         // the local repo has to be in here twice, because sbt won't push to a "file:"
         // repo, but it won't read artifacts from a "Resolver.file" repo. (head -> desk)
-        "local-lookup" at ("file:" + localRepo.getAbsolutePath),
+        "local-lookup" at localRepo.getAbsoluteFile.toURI.toString,
         Resolver.file("local", localRepo)(Resolver.mavenStylePatterns)
       )
     },


project/ZipkinResolver.scala
```diff
@@ -49,7 +49,13 @@ object ZipkinResolver extends Plugin {
       }) ++ Seq(
         // the local repo has to be in here twice, because sbt won't push to a "file:"
         // repo, but it won't read artifacts from a "Resolver.file" repo. (head -> desk)
-        "local-lookup" at ("file:" + localRepo.getAbsolutePath),
+        "local-lookup" at {
+          val original = "file:" + localRepo.getAbsolutePath
+          val crossPlatform = localRepo.getAbsoluteFile.toURI.toString
+          println("The original statement says that the repository url is: "+original)
```
rtyler: Are these printlns going to stay in here forever? Any reason they can't go away?

rtyler:
@eirslett Travis is running the build on a Linux box, is that not good enough? :)

project/ZipkinResolver.scala
eirslett: No, they were just for testing so one of you guys (or Travis) could verify it was correct without starting sbt in debug mode and step through it until you find the value. According to Travis, it looks correct, I'll remove the println statements and push a new commit to this PR.

eirslett:
I'm going to merge this now, looks like it didn't break anything - but it definitely fixed the Windows build. (Nobody probably cares, except for me)

codefromthecrypt:
thanks!

codefromthecrypt:
:shipit:

PR #480, opened 2015-07-20
Reduces heap to below travis limits

codefromthecrypt:
Travis legacy infrastructure has 3GB mem, while containers have 4. Our
build needs to stay under that limit or risk being killed. This changes
the heap size down to 2GB, knowing that other args may still run over 3.


codefromthecrypt:
one passed! probably should verbose:gc in case the other was in a gc spiral.. will do that tomorrow.

rtyler:
@adriancole I'm going to merge this so we can get more success with other PRs, mind opening another PR for the verbose GC bits?

PR #481, opened 2015-07-20
Zoom the trace view based on user selected time range.

nikhil2406:
Our traces sometimes span across a few minutes. That causes a bunch of spans to get huddled up in a corner. This pull request is to enable users to have a better viewing option in such cases. User can move mouse over the span view to select a time range and entire span view would be zoomed in according to selected time range. User thus gets a better view of the spans they are interested in analyzing.

A zoomout button would pop up which allow users to go back to original view. The button has a fixed position on screen so that even if user scroll up or down the zoomout button is always available at same position on screen.
Please check and let me know your feedback.

Thanks,
Nikhil.


eirslett:
Sounds like a good feature! Could you please post some simple screenshots?

nikhil2406:
screen shot 2015-07-20 at 6 33 47 pm
1. Above screenshot shows user selecting a time range between 23-34 ms.

screen shot 2015-07-20 at 6 34 11 pm
1. This one shows the span view redrawn/zoomed in according to selected time range. All spans are rearranged to show their respective positions and widths in new time range. The new time range and spans which started in that time range are shown with different color. Also the zoomout button appears with same color scheme.

screen shot 2015-07-20 at 6 34 27 pm
3. After zoomout button is clicked original span view appears back.

codefromthecrypt:
Cool!

eirslett:
This looks great!
Only nitpick is maybe the zoom button could be placed in the lower right part of the screen, together with the back-to-top button? (That was in a recently merged commit)

nikhil2406:
screen shot 2015-07-21 at 2 02 39 pm

Done. Please check now.

eirslett:
Sweet! What does it mean that the text on a span is red instead of black?

nikhil2406:
I've kept red color as a common theme to highlight all changes that happened in span view when user selected the zoom-in time range. Red text specifies that the span actually 'started' in selected time range. I think it's going to help users to easily locate the spans they are interested in analyzing.
I've changed border color of selection-rectangle also to red.

codefromthecrypt:
:shipit:

I'm cool on merging this. Even if there are future UX enhancements later,
my guess is that this is largely what we want. For example, zoomable ui is
a key feature of HTrace.

After this goes prod, I can facilitate further feedback. Iterations will be
quite easy once releases are possible, and I suspect we will have a release
process quite soon.

codefromthecrypt:
@nikhil2406 this needs to be rebased against master (and squashed)

eirslett:
+1 for merge! Once it's rebased/squashed

nikhil2406:
Squashed and rebased. Please check now.

codefromthecrypt:
Hi, there.. this is still 2 commits (squashed would be one). I don't want to burden you further :) I will cherry-pick this and squash it on the way in.

codefromthecrypt:
closed via e5be7fc425d87651d01f512997fe64197935816a

Thanks so much!

mosesn:
@nikhil2406 yoooo this is sweet :+1:

Issue #415, opened 2014-09-18
cassandra issue with query by binaryAnnotations


LiuJiaYang:
Zipkin runs on cassandra, but cannot find traces with any specified annotation.
When the serviceName is include Uppercase, than the web can not find anything.

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraIndex.scala
fixed:
273          val key = encode(endpoint.serviceName, ba.key).getBytes
-->          val key = encode(endpoint.serviceName.toLowerCase, ba.key).getBytes

eirslett:
Closing due to lack of activity.

PR #482, opened 2015-07-20
Fixes query by binaryAnnotations in cassandra #415

drax68:
Fixes query by binaryAnnotations in cassandra #415


eirslett:
I think I've actually encountered this issue as well, and applied the same fix. +1 for merge!

codefromthecrypt:
+1

PR #485, opened 2015-07-20
Test split travis build with env

eirslett:
If you look at the Travis build, this is one way to go. Splitting up the build into smaller pieces allows us to isolate failure, and work around the problem of resource usage.
The flip side is that we use more resources, since we have to bootstrap sbt and download dependencies for every submodule.

PS: do not merge

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,24 @@ before_script:
   # default $SBT_OPTS is irrelevant to sbt lancher
   - unset SBT_OPTS
 
+env:
+  - PROJECT="zipkin-aggregate"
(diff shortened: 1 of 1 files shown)


rtyler:
that's some split!

codefromthecrypt:
Hehe love the 'ps'

codefromthecrypt:
I like this, but I'm concerned about the >2hrs feedback time. I think that is too long.. a cure worse than the disease. Do you feel otherwise?

eirslett:
Two hours is too long, yes. This might not be a good option after all...

eirslett:
I'll close this one.

PR #487, opened 2015-07-21
Changes build/package namespace to io.zipkin

codefromthecrypt:
This step starts the work towards io.zipkin, but without altering config
or code packages. Most importantly, it sets the group id to io.zipkin
for artifact publication.


codefromthecrypt:
published locally and it made the expected ivy dirs. cc @jamescway

rtyler:
Oh whoops, I forgot to :+1: this yesterday when I looked at it

PR #488, opened 2015-07-22
Simplifies resolvers and updates dependencies

codefromthecrypt:
Many resolvers are not needed, or scoped globally, but only used for a
couple projects. In any case, the configuration was complex. This
starts over, adding the minimal repositories to get the build to pass.

This also updates certain dependencies, which in some cases makes them
available in standard repositories. These dependencies were checked to
ensure they work on JRE 7.


codefromthecrypt:
@abesto @jamescway @rtyler PTAL the build is still failing, but we've strangled the maven resolvers to allow it to fail faster and without twitter's repo serving up stuff already on central

abesto:
:+1: 
Removed code, cleaner resolvers, jcenter, me likes. I don't see any issues.

PR #494, opened 2015-07-23
Ports specs/jmock tests to scalatest/mockito and cleanup

### LINKED ISSUE #493: refactor mocking approach
Our mocking approach is fragile due to jmock class mocks. I attempted to increase the version and failed. It would be nice to be able to have a less complex build, and looks like a large part of that is reducing class mocks.

ex. here's what we need in order to have the build work at all

```
    "org.jmock"               %  "jmock"        % "2.4.0" % "test",
        // jmock tests mock classes and require additional dependencies.
        "cglib"                   %  "cglib"        % "2.2.2" % "test",
        "asm"                     %  "asm"          % "1.5.3" % "test",
        "org.objenesis"           %  "objenesis"    % "1.1"   % "test"
```

codefromthecrypt:
Before, tests were a mix of scalatest and specs, and a mix of jmock and
mockito. This migrates all tests to scalatest/mockito and removes the
`@RunWith` annotation, which isn't necessary in SBT or Intellij.

All tests are now `FunSuite` except `QueryServiceSpec`, which is vast
and better organized as `WordSpec` for now.

Almost all tests use Matchers, and the ones that don't aren't hurting
anything, so could later be moved to Matchers if desireable.

Finally, doing this simplifies the build: we now have a single test
dependency bundle, and it is far simpler than before.

Fixes #493

### DIFF
--- a/zipkin-collector-core/build.sbt
+++ b/zipkin-collector-core/build.sbt
@@ -4,6 +4,5 @@ libraryDependencies ++= Seq(
   Seq(algebird("core"), twitterServer, ostrich),
   many(finagle, "ostrich4", "serversets", "thrift", "zipkin"),
   many(util, "core", "zk", "zk-common"),
-  many(zk, "candidate", "group"),
-  testDependencies
-).flatten
+  many(zk, "candidate", "group")
+).flatten ++ testDependencies
--- a/project/Project.scala
+++ b/project/Project.scala
@@ -58,22 +58,10 @@ object Zipkin extends Build {
   val slf4jLog4j12 = "org.slf4j" % "slf4j-log4j12" % "1.6.4" % "runtime"
   val ostrich = "com.twitter" %% "ostrich" % "9.9.0"
 
-  lazy val scalaTestDeps = Seq(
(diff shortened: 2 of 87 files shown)


zipkin-collector-core/build.sbt
```diff
@@ -4,6 +4,5 @@ libraryDependencies ++= Seq(
   Seq(algebird("core"), twitterServer, ostrich),
   many(finagle, "ostrich4", "serversets", "thrift", "zipkin"),
   many(util, "core", "zk", "zk-common"),
-  many(zk, "candidate", "group"),
-  testDependencies
-).flatten
+  many(zk, "candidate", "group")
+).flatten ++ testDependencies
```
abesto, line 8: Nice, I like the consistency

abesto:
Tests run OK locally, didn't find even nitpicks in the diff. I'm +1 for merge.
Still, I'd prefer getting another set of eyes on this (the brains behind which would ideally have a closer familiarity with the tests).

jamescway:
+1 looks like this was a lot of work!

PR #496, opened 2015-07-23
Separates out downloading dependencies from running tests

codefromthecrypt:


codefromthecrypt:
rebased

.travis.yml
```diff
@@ -21,10 +21,11 @@ cache:
 before_script:
   # default $SBT_OPTS is irrelevant to sbt lancher
   - unset SBT_OPTS
+  # ensure the build errors if dependencies aren't available
+  - travis_retry ./bin/sbt ++$TRAVIS_SCALA_VERSION update
```
codefromthecrypt, line 25: keyword retry! so whenever maven.twttr.com fails (and yes.. it does often).. we might actually be able to proceed.

jamescway, line 25: how many times does it retry?  Kind of looks like 3 by default? Looking for reference...

jamescway, line 25: yeah its [3 times](http://docs.travis-ci.com/user/build-timeouts/#stq=&stp=0)  "Beyond Bundler, you can wrap commands using the function travis_retry which checks the return code of a command, retrying it three times if the return code is non-zero."

jamescway:
:+1:

.travis.yml
codefromthecrypt, line 25: not easy to find, but yeah 3 tries

http://docs.travis-ci.com/user/build-timeouts/

PR #497, opened 2015-07-23
Refactors SBT plugin configuration

codefromthecrypt:
```
The SBT plugin configuration we used was out-of-date, and for example
caused us to download different versions of scrooge. We also had an
unused copy/pasted unidoc plugin.

This change clears out unused and/or confusing plugin configuration. It
also updates plugins we actually use to current versions.

The sbt-assembly plugin was updated to 0.11.2, as versions beyond that
need significant configuration work.
```


bin/sbt
```diff
@@ -37,7 +38,7 @@ version=$(java -version 2>&1 | awk -F '"' '/version/ {print $2}')
 if [[ "$version" > "1.8" ]]; then
     export VM_OPTS="-Xms2G                          \
                     -Xmx2G                          \
-                    -Xss256k                        \
+                    -Xss512k                        \
```
codefromthecrypt, line 41: this was needed for my mac's JDK 8

jamescway:
+1 cut the fat! :cow2:

zipkin-scrooge/build.sbt
```diff
@@ -1,11 +1,4 @@
-import com.twitter.scrooge.ScroogeSBT
-
-defaultSettings ++ ScroogeSBT.newSettings
-
-ScroogeSBT.scroogeThriftSourceFolder in Compile <<= (
-  (baseDirectory in ThisBuild)
-  (_ / "zipkin-thrift" / "src" / "main" / "thrift" / "com" / "twitter" / "zipkin" )
-)
+scroogeThriftSourceFolder in Compile := file("./zipkin-thrift/src/main/thrift")
```
eirslett, line 1: Does this work if you say `sbt` `project zipkin-scrooge`, that it won't change the working directory?

project/plugins.sbt
```diff
@@ -1,11 +1,5 @@
-libraryDependencies ++= Seq(
-    "com.google.collections" % "google-collections" % "0.8",
-    "org.codehaus.plexus"    % "plexus-utils"       % "1.5.4",
-    "org.slf4j"              % "slf4j-api"          % "1.6.1",
-    "org.slf4j"              % "slf4j-simple"       % "1.6.1")
+resolvers += Resolver.sbtPluginRepo("releases")
 
+addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.13.0")
```
eirslett: The assembly stuff is still quite fragile, and unfortunately we don't have tests for it yet... you'll have to start up Hadoop and everything just to see that it works. Does this version bump break anything, or is it fine?
We should probably have a Hadoop docker image just to test that this job works...

zipkin-scrooge/build.sbt
codefromthecrypt, line 1: can try that now.. I basically just ported the prior config.

codefromthecrypt, line 1: that syntax doesn't work, but it also doesn't work for unrelated projects. I'm hesitant to jump down more build areas, but if you think this is important let's at least log an issue.

Doesn't seem related, or a feature we use in travis or the release process. That's why I think we can separate out "sbt project X" from this change, and from the general work on getting the build green.

eirslett, line 1: Well, not really important at all. Mostly when you work with Scrooge, which isn't that often.

project/plugins.sbt
codefromthecrypt: running assembly now.

+1 to having some sort of automated test for this!

codefromthecrypt:
assembly screwed up, so debating whether to undo that part of the change (depends on if it still works anyway)

codefromthecrypt:
ok sbt-assembly is now only upgrated to 0.11.2 (a bit over a year old), and punts updating it further, as that would require rewriting the corresponding build file.

project/Project.scala
```diff
@@ -76,11 +76,7 @@ object Zipkin extends Build {
     crossPaths := false,            /* Removes Scala version from artifact name */
     fork := true, // forking prevents runaway thread pollution of sbt
     baseDirectory in run := file(cwd), // necessary for forking
-    resolvers := Seq(
-      Resolver.jcenterRepo,
-      Resolver.typesafeRepo("releases"),
-      "Twitter Maven Repository" at "https://maven.twttr.com/" // for thrift 0.5
-    ),
+    resolvers += "twitter-repo" at "http://maven.twttr.com", // for thrift 0.5
```
abesto: Does this drop jcenter and typesafe/releases, or are those included by default?

codefromthecrypt: Let me verify, but I think they are default. If not I will update. Thx!

codefromthecrypt: Typesafe isn't needed. Jcenter isn't either except for "performance", but I
think it would come last so wouldn't help there anyway. Resolver order is
hard to control unless you set a JVM arg to override. I recommend leaving
this as is until gradle.. SG?

codefromthecrypt: Scratch that... Something is falling through to the Twitter repo.. I will
look into this!

```

[error] SERVER ERROR: Service Temporarily Unavailable url=
http://maven.twttr.com/org/slf4j/slf4j-parent/1.6.4/slf4j-parent-1.6.4.jar

```

codefromthecrypt:
verified we are only downloading twitter stuff from https://maven.twttr.com

abesto:
LGTM :rainbow: +1

PR #502, opened 2015-07-24
fixed option tags for sort selector

mzagar:
Hi, 

I'm trying to display zipkin-web in an iframe but having issues due to invalid html in sort select option tag; hopefully this fixes it.

Regards,
Mario


eirslett:
Oops, thanks! Looks like a simple bug, so I'll merge it right away.

mzagar:
Thanks!

codefromthecrypt:
:shipit: thanks for the fix!

PR #503, opened 2015-07-24
Gradle.

abesto:
### Why?

Because the SBT build is a pain.
### What works
- `./gradlew check` runs tests using [maiflai/gradle-scalatest](https://github.com/maiflai/gradle-scalatest) and they pass.
- `./gradlew build` builds normal jars, and an IDL jar for `zipkin-thrift`.
- `./gradlew distTar` builds a tar distribution for the stand-alone services
- `./gradlew installDist` generates a runnable version of the services (basically the contents of the output of `distTar`)
### What doesn't work

I'd appreciate thoughts on how important these are.
- The `doc` project isn't ported yet. I can take a stab at it, but if someone knows what actually happens there, they'll probably have an easier time.
- `BuildProperties` is not ported. Do we need it?
### What might not work

I'd appreciate experienced eyes on these
- Maybe some tests are not picked up by Gradle
- Maybe some artifacts don't look like they should (especially IDL and standalone service stuff)
### Integration test

I used https://github.com/openzipkin/docker-zipkin/tree/gradle to bring up Zipkin, and sent traces to it using

``` sh
./zipkin-tracegen/build/install/zipkin-tracegen/bin/zipkin-tracegen -scribeDest=$(docker-machine ip):9410
```

And I saw traces :smile: 
### Notes
- The SBT build system is still in place to make comparison easy.


rtyler:
This is relevant to my interests :)

codefromthecrypt:
For scalatest, would this be an option?
https://github.com/maiflai/gradle-scalatest/blob/master/README.md

abesto:
Tried `maiflai/gradle-scalatest` before going with the JUnit integration thing, couldn't get it to work :(

zipkin-common/src/main/scala/com/twitter/zipkin/test/FunSuiteWithJUnit.scala
```diff
@@ -0,0 +1,9 @@
+package com.twitter.zipkin.test
+
+import org.scalatest.junit.JUnitRunner
+import org.junit.runner.RunWith
+import org.scalatest.FunSuite
+
+@RunWith(classOf[JUnitRunner])
+class FunSuiteWithJUnit extends FunSuite {
```
codefromthecrypt: adding a common dep here is not so hot imho, as we already have a lot of coupling. Can you just add RunWith to all the tests? Then, in the main build file, put a TODO to replace boilerplate `@RunWith(classOf[JUnitRunner])` with getting scalatest plugin working?

codefromthecrypt: PS I would help with pulling this out.. just don't want anything else depending on common than absolutely must :P

abesto: That works too. Feel free to get started if you want, help is greatly appreciated. I'll probably get around to it this weekend otherwise.

codefromthecrypt: Oh right.. This is a shared branch! OK I will help today but will be away
until Tuesday am Pacific. (FYI)

abesto: Cool, thanks! Also for the heads up.

abesto:
Tests are now run using maiflai/gradle-scalatest, and some manual tests suggest the whole thing working (see details in description).

build.gradle
```diff
@@ -0,0 +1,87 @@
+subprojects {
+    apply plugin: 'java'
+    apply plugin: 'scala'
+
+    apply from: "${rootProject.projectDir}/gradle-scalatest.plugin"
+
+    ext {
```
rtyler, line 8: Generally I try to define things like this, where possible, in `gradle.properties` that way they can be more easily overwritten on the command line with `-P` if need be.

I understand that this doesn't quite work when these attributes are invoking some logic but something to take under consideration

build.gradle
```diff
@@ -0,0 +1,87 @@
...
+    //////////////////////////
+    // Common configuration //
+    //////////////////////////
+
+    // SBT migration note: start of zipkinSettings
+    group "io.zipkin"
+    version zipkinVersion
+
+    repositories {
+        jcenter()
```
rtyler, line 15: The order of these declarations will be the order in which servers are queried for a dependency. Considering how freakin' unreliable maven.twttr.com has been, perhaps move it to the bottom so it's called as little as possible :P

build.gradle
```diff
@@ -0,0 +1,87 @@
...
+        maven { url 'https://maven.twttr.com/' }
+        maven { url 'http://repo.typesafe.com/typesafe/releases/' }
+        // For zipkin-cassandra deps and com.twitter:kafka
+        maven { url 'http://conjars.org/repo' }
```
rtyler: Shouldn't this be declared as a dependency in the `zipkin-cassandra/build.grade` then?

build.gradle
```diff
@@ -0,0 +1,87 @@
...
+        // For storm-kafka
+        maven { url 'https://clojars.org/repo' }
```
rtyler: Shouldn't this be declared as a dependency in the `zipkin-receiver-kafka/build.grade` then?

build.gradle
```diff
@@ -0,0 +1,87 @@
...
+
+        testCompile junit,
+                'org.mockito:mockito-all:1.10.19',
+                "org.scalatest:scalatest_${scalaInterfaceVersion}:2.2.5"
+    }
+    // SBT migration note: end of zipkinSettings
+}
+
+// Quick hack to visualize inter-project dependencies
+task dependencyReport {
```
rtyler, line 36: This task should have a `description`

gradle-scalatest.plugin
```diff
@@ -0,0 +1,21 @@
...
+		maven {
+			url "https://plugins.gradle.org/m2/"
+		}
+	}
+	dependencies {
+		classpath "com.github.maiflai:gradle-scalatest:0.8"
+	}
+}
+
+if (!project.plugins.findPlugin(com.github.maiflai.ScalaTestPlugin)) {
```
rtyler: I'm not sure why this is necessary, .apply plugin`has been idempotent in my experience. The way that this code is being called is from a`subprojects {}` which means the apply call will work on each new (sub)project evaluated.

AFAICT this should be safe to nuke

zipkin-anormdb/build.gradle
```diff
@@ -0,0 +1,16 @@
+// Database drivers
+ext.anormDriverDependencies = [
```
rtyler, line 7: Not sure I understand why this list needs to be declared here instead of just as enumerated dependencies below

zipkin-thrift/build.gradle
```diff
@@ -0,0 +1,11 @@
...
+	repositories {
+		jcenter()
+	}
+
+	dependencies {
+		classpath griddle
+	}
+}
+
+apply plugin: 'idl'
```
rtyler, line 11: This `buildscript {}` block is the example same as the one in `zipkin-scrooge` but this one applies the 'idl' plugin, and the other applies the 'scrooge' plugin.

Are both plugins defined in this `griddle` dep?

rtyler:
Did a cursory review, I might be misunderstanding some things so don't take my comments as blocking or anything like that

abesto:
Thanks a lot, good points! I'll address them in detail tomorrow.

build.gradle
abesto, line 8: I agree in principle, but can't point to any specific line / part of line I'd move. Do you have specifics in mind?

build.gradle
abesto, line 15: Nice catch. Here: https://github.com/openzipkin/zipkin/commit/8d2dd007db6c9ca826a90b216d96fd7ac9ce7db8

build.gradle
abesto: I found that if project A depends on project B, and B has a dependency from resolver R, then A must also have R declared, otherwise compiling A will fail, even if it's marked as being dependent on B. I see two solutions:
1. Current setup
2. Put the appropriate line into the `build.gradle` of each project where it's needed, of course DRY-ing it by declaring the URL once in a common place.

Did I miss something in the "problem statement"? Do you see another approach? Would you prefer the second solution?

I'll definitely document whatever the end result is in a comment, this is cause for confusion.

build.gradle
abesto: Same issue as https://github.com/openzipkin/zipkin/pull/503#discussion_r35494029, let's keep the discussion there.

build.gradle
abesto, line 36: Done in https://github.com/openzipkin/zipkin/commit/5df2d8b306e1d79ee0ac21ec26a76893ff0a4fc3, thanks

gradle-scalatest.plugin
abesto: First saw this in https://raw.githubusercontent.com/akhikhl/gradle-onejar/master/pluginScripts/gradle-onejar.plugin, looked like good practice. Is there a down-side to this being here?

zipkin-anormdb/build.gradle
abesto, line 7: It's just DRY; same reason for all the other helpers using `mkDepGen`. The map is needed because we need different versions for different artifacts from the same group.

What would an alternative implementation look like?

zipkin-thrift/build.gradle
abesto, line 11: That's correct. Would you prefer having the common parts factored out?

gradle-scalatest.plugin
rtyler: Not a technical downside as far as I'm aware, but at least as far as Gradle 2.x is concerned it's definitely not a conventional means of applying plugins which has the potential to confuse those less than familiar with Gradle

zipkin-thrift/build.gradle
rtyler, line 11: No, I was more confused because most Gralde plugins that I've seen define one plugin per dependency/jar

build.gradle
codefromthecrypt: TL;DR; I really prefer these being declared where used, as it highlights tech debt without pushing it to all users. Our build files should be documented well enough to help someone understand why repositories are there.

This isn't special to gradle.. I've noticed that the normal finagle thing requires you to define repositories here or there (ex. like zipkin itself needing to track thrift 0.5.0-1). I've been troubled by this topic. It is a usual concern to have to hunt down repos to get twittery things working, particularly maven.twttr.biz. However, at least inside twitter, there's a move towards all things are sonatype -> maven central. In other words, the extra repos, at least some of them, should become less burdensome. Also, we should consider that in some cases, the extraneous repo definition is at the edge of the graph (ex zipkin-cassandra). Those who need to redefine edges are power users already, or a symptom of a config problem. Power users should be able to read our well-commented build files to get what they need. Config problems should be addressed. Plopping a large list or repositories seems a short-cut, but I'm generally uncomfortable with making things that make tech debt creation or retention easier.

gradle-scalatest.plugin
codefromthecrypt: I agree with tyler that we should go with conventional first. Looking back at SBT problems we've had, I don't want to do anything that would cause confusion for future maintainers.

In other words, if the guard (if-statement) is functional, then comment when it can be removed, hopefully via a github issue being resolved.

abesto: Thanks for the insights. I'm convinced, removed in https://github.com/openzipkin/zipkin/commit/b4e7113e437f0c47731f28f25f3f89dcce6194f7

build.gradle
```diff
@@ -0,0 +1,89 @@
...
+    apply plugin: 'scala'
+
+    apply from: "${rootProject.projectDir}/gradle-scalatest.plugin"
+
+    ext {
+        ////////////////////////////////
+        // Commonly used dependencies //
+        ////////////////////////////////
+
+        mkDepGen = { genOne -> { String... names -> names.collect(genOne) } }
```
codefromthecrypt: In cases that we need to switch on scala version, we should be able to do if statements.

https://github.com/apache/kafka/blob/trunk/build.gradle#L237

I'd avoid making new words in general. Ex. the following won't invite other gradle folks to maintain this even if we are saving characters. We should just declare inline, and avoid special build language like `mkDepGen` as zipkin is hopefully not special enough to need its own build vocab. I'd rather optimize for immediate understanding vs saving character count. DRY is something to watch out for, but build files and unit tests have special concerns which IMHO override that in many cases.

On technical notes.. if the goal is to force a specific version for twitter things, I would do that in a way that is more idiomatic to gradle. ex.

```
configurations.all {
    resolutionStrategy {
        eachDependency { DependencyResolveDetails details ->
            if (details.requested.group == 'com.twitter:finagle-foo') {
                details.useVersion "6.26.0"
            }
        }
    }
}
```

codefromthecrypt:
OK I'm done with the review

I prefer this even as is much more than SBT.

Not blocking, but I feel we should go with conventional gradle vs abbreviations or further normalization.

Ex. fully qualified dependency names vs abbreviations, use resolvers to enforce versions etc. I would optimize more for lower WTF count and higher conventional adoption vs less characters or more effective DRY

build.gradle
abesto: Alright, I'll switch over to inline dependency declarations. I think introducing version rewriting like you suggest will cause confusion, I'd rather keep the version numbers in `gradle.properties`. Would that work for you?

codefromthecrypt: I have projects that put nothing into gradle.properties, so not sure if
that's best. Rewriting is only important when you need to force a version,
so I agree it isn't natural for many things.

build.gradle
abesto: I think I addressed this in https://github.com/openzipkin/zipkin/commit/22424ac363a3cc797e2c970237b14e7d2fd8de41.

codefromthecrypt:
I would like to note that this branch is greener than Boston on St Patty's day

zipkin-anormdb/build.gradle
```diff
@@ -0,0 +1,20 @@
+// Database drivers
```
codefromthecrypt, line 1: maybe explain this a bit more in the comment. Ex. what's expecting a property of this naming convention? when is it loaded (during tests?) how is it correlated to the anormDriverDependencies below (if at all)?

codefromthecrypt:
:shipit: with only a couple nits

only one thing surprises me still, which is how the things in anormdb relate to eachother. Then, there's the supernit of tabs? Almost all projects I've seen who have tabs end up in peril due to people spacing to match. FYI, I'm shortly to carve out mongo

zipkin-anormdb/build.gradle
abesto, line 1: How about this? https://github.com/openzipkin/zipkin/commit/e3619a765069c29b15bc2e3f75c906f4f12345fb

abesto:
I suggest we break out the anormdb thing to a separate issue. Users shouldn't have to recompile to switch the storage they use, but I'd like to keep the scope of this PR minimal.

On tabs: oh my, I just let IntelliJ do whatever it wanted. Spaces incoming.

codefromthecrypt:
@abesto wrt anormdb.. it isn't that I want a refactor.. just that the comment `// Database drivers` doesn't explain what's going on. Is there some better context to put in a comment? Right now, it looks like a magic property, which even if is the case, might be able to be explained at least when it is used.

abesto:
@adriancole plaese check the comments added in https://github.com/openzipkin/zipkin/commit/e3619a765069c29b15bc2e3f75c906f4f12345fb. I hope that explains what's going on.

codefromthecrypt:
@abesto whoot! LGTM

codefromthecrypt:
TODO: bump zipkin-web mustache to 0.8.17, remove build files for mongo

codefromthecrypt:
@abesto so I think you are basically merge-town. might want to consider collapsing commits at this point, yeah? Otherwise this will be perpetually a tracking branch.

abesto:
Yep, I'll take care of just this last TODO, then squash.

rtyler:
I'm going to merge this since the build failures are @adriancole's fault and this change looks like it will dramatically improve everybody's lives :)

codefromthecrypt:
NOTE: this also deleted the pants files

PR #504, opened 2015-07-24
Moves zipkin-hbase to its own repository

codefromthecrypt:
zipkin-hbase integration tests regularly cause our builds to crash. This
has endangered the productivity of the project. Since this has never
been released, this burden isn't justifiable.

This change isolates hbase to its own repository, where it can either
be maintained, or be archived.

It is an acknowledged reality that this repository won't work until we
release zipkin artifacts it depends on, and perhaps perform other
modularity work.

https://github.com/openzipkin/zipkin-hbase


eirslett:
+1 let's move it out :-)

jamescway:
:+1: :+1: :+1:  looks like u got everything

PR #508, opened 2015-07-28
zipkin-mongodb: Replace JSR166y with Java 7 API

codefromthecrypt:
This tracks upstream work in Finagle, by @vkostyukov. We shouldn't be
depending on custom ForkJoinPool, especially as we only target JRE 1.7+.


eirslett:
+1

jamescway:
:+1:  but it doesn't seem like its used anywhere?

codefromthecrypt:
it is, just the simple name didn't change, hence not listed in the diff.

PR #509, opened 2015-07-28
Bumps mustache to 0.8.17

codefromthecrypt:
Pants (and therefore twitter production) was running 0.8.13. SBT was
out-of-sync, specifying 0.8.13. This bridges the gap, and should be a
safe merge.

### DIFF
--- a/zipkin-web/build.sbt
+++ b/zipkin-web/build.sbt
@@ -7,7 +7,7 @@ libraryDependencies ++= Seq(
     zk("server-set"),
     algebird("core"),
     twitterServer,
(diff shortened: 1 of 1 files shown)


eirslett:
Or maybe use 0.8.17?

codefromthecrypt:
@eirslett sgtm.. is there anything about 0.8.17 that would be incompatible with things? I know Twitter is converging at least initially towards 0.8.14, which was what motivated the change.

jamescway:
don't have strong feelings about mustache versions, updating lgtm :100:

codefromthecrypt:
bumping to 0.8.17 as at least scrooge also uses this :)

PR #512, opened 2015-07-28
Local copy of ScalaObjectHandler

codefromthecrypt:
Copy of ScalaObjectHandler from scrooge as otherwise build will break.

Recent upgrade highlights we need this explicitly.

See https://github.com/twitter/scrooge/pull/198


codefromthecrypt:
@abesto fyi the latest bump on mustache broke the build. you'll want to rebase over this.

codefromthecrypt:
merging as the build is broke otherwise. Will take any feedback seriously post-merge.

eirslett:
We should be able to revert back to scala-extensions once we're on Scala 2.11.

eirslett:
Or so I think...

codefromthecrypt:
> Or so I think...
> Tee hee

PR #513, opened 2015-07-29
Moves to finagle.httpx

codefromthecrypt:
This is part of the larger project to upgrade Finagle
to Netty 4, which means getting rid of finagle-http
and moving everybody to finagle-httpx.

### DIFF
--- a/zipkin-example/src/main/scala/com/twitter/zipkin/example/Main.scala
+++ b/zipkin-example/src/main/scala/com/twitter/zipkin/example/Main.scala
@@ -1,20 +1,18 @@
 package com.twitter.zipkin.example
 
-import com.twitter.zipkin.conversions.thrift._
-import com.twitter.finagle.Http
-import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.finagle.Httpx
 import com.twitter.server.{Closer, TwitterServer}
-import com.twitter.util.{Await, Closable, Future}
+import com.twitter.util.{Await, Closable}
 import com.twitter.zipkin.anormdb.AnormDBSpanStoreFactory
-import com.twitter.zipkin.collector.SpanReceiver
 import com.twitter.zipkin.common.Span
-import com.twitter.zipkin.{thriftscala => thrift}
-import com.twitter.zipkin.receiver.scribe.ScribeSpanReceiverFactory
-import com.twitter.zipkin.zookeeper.ZooKeeperClientFactory
-import com.twitter.zipkin.web.ZipkinWebFactory
+import com.twitter.zipkin.conversions.thrift._
 import com.twitter.zipkin.query.ThriftQueryService
 import com.twitter.zipkin.query.constants.DefaultAdjusters
+import com.twitter.zipkin.receiver.scribe.ScribeSpanReceiverFactory
 import com.twitter.zipkin.tracegen.ZipkinSpanGenerator
+import com.twitter.zipkin.web.ZipkinWebFactory
+import com.twitter.zipkin.zookeeper.ZooKeeperClientFactory
+import com.twitter.zipkin.{thriftscala => thrift}
 
 object Main extends TwitterServer with Closer
   with ZooKeeperClientFactory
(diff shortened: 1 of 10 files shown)


codefromthecrypt:
@taylorleese fyi

zipkin-example/src/main/scala/com/twitter/zipkin/example/Main.scala
```diff
@@ -1,20 +1,18 @@
...
-import com.twitter.finagle.Http
-import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.finagle.Httpx
 import com.twitter.server.{Closer, TwitterServer}
-import com.twitter.util.{Await, Closable, Future}
+import com.twitter.util.{Await, Closable}
 import com.twitter.zipkin.anormdb.AnormDBSpanStoreFactory
-import com.twitter.zipkin.collector.SpanReceiver
 import com.twitter.zipkin.common.Span
-import com.twitter.zipkin.{thriftscala => thrift}
```
codefromthecrypt, line 11: sorry about the import change.. the amount of unused ones was driving me nuts. I don't feel like making a pull request just to do that.

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
```diff
@@ -55,7 +53,7 @@ class Handlers(jsonGenerator: ZipkinJson, mustacheGenerator: ZipkinMustache, que
 
     def apply(response: Response) {
       response.setContentType(typ)
-      response.content = ChannelBuffers.wrappedBuffer(content)
+      response.content = Buf.ByteArray(content)
```
codefromthecrypt, line 55: Somewhere else in this file, this `content` is basically the result of an InputStream read into a byte array. Is there a shortcut to create a shareable Buf from an InputStream directly?

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
```diff
@@ -15,7 +14,6 @@ import com.twitter.zipkin.query.{SpanTimestamp, TraceCombo, TraceSummary, QueryR
 import com.twitter.zipkin.thriftscala.{Adjust, ZipkinQuery}
 import java.io.{File, FileInputStream, InputStream}
 import org.apache.commons.io.IOUtils
-import org.jboss.netty.buffer.ChannelBuffers
 import org.jboss.netty.handler.codec.http.{HttpRequest, HttpResponse}
```
taylorleese: @adriancole remove ^^

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
taylorleese, line 55: @dschobel @luciferous Can you help with Adrian's question?

luciferous, line 55: Sorry if I'm narrow on the context. I'll look closer and follow up tomorrow, but here's a quick answer for now.

To get a Buf from an InputStream we have to use a Reader.

```
import com.twitter.io.Reader
val reader: Reader = Reader.fromStream(stream)
val buf: Future[Buf] = Reader.readAll(reader)
```

This gives a Future[Buf], which you could flatMap to get at the Buf.

eirslett:
+1

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
codefromthecrypt, line 55: have to admit that this is more complex than the existing code.. I think I'll leave it.

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
codefromthecrypt: done. thanks!

PR #516, opened 2015-07-29
Parallel test tasks in the Gradle build

rtyler:
This also moves the definition of the group and version into the `gradle.properties` which will properly apply to all subprojects


gradle.properties
```diff
@@ -1,5 +1,6 @@
-zipkinVersion = 1.2.0-SNAPSHOT
+versoin=1.2.0-SNAPSHOT
```
abesto: Typo :)

abesto:
Other than the typo, :+1:

Issue #518, opened 2015-07-29
Migrate docs to asciidoc


codefromthecrypt:
The [zipkin website](http://openzipkin.github.io/zipkin/) is running off the [gh-pages](https://github.com/openzipkin/zipkin/tree/gh-pages) branch of the zipkin repository.

The content is built using sphinx, from source in the doc folder. This process relies on python etc, which isn't a big deal in and of itself. More importantly, the content isn't using code that's already tested.

Let's consider asciidoc as an alternative. Something like https://github.com/jruby-gradle/jruby-gradle-plugin/tree/master/docs

eirslett:
zipkin.io was recently fixed, thanks @abesto ! I'll close this issue, I don't think it's relevant any longer.

PR #519, opened 2015-07-29
Converts broken sphinx process to docs

codefromthecrypt:
The existing sphinx process isn't documented, and also is broken. The
configuration was also invoking sbt to create variables never used. This
converts the doc process to docs, which buys us time to switch to
asciidoc.

See #518


jamescway:
:+1:

PR #520, opened 2015-07-29
Removes SBT from bin scripts

codefromthecrypt:
This ports configuration in existing bin scripts from SBT to gradle. It
also removes the redundant `-redis` scripts and renames `test` to
`tracegen`.

This also configures and explains why query and collector need to start
in the same directory by default (configKey == "dev").


bin/test
```diff
@@ -1,2 +0,0 @@
-#!/usr/bin/env sh
```
adriancole: actually I'll resurrect this

bin/collector
```diff
@@ -1,6 +1,2 @@
 #!/usr/bin/env bash
-
-DEFAULT_TYPE=dev
-SERVICE_TYPE=${1:-$DEFAULT_TYPE}
-
-bin/sbt 'project zipkin-collector-service' "run -f zipkin-collector-service/config/collector-${SERVICE_TYPE}.scala"
+SERVICE_TYPE=${1:-dev} ./gradlew :zipkin-collector-service:run
```
jamescway: is it still possible to run https://github.com/openzipkin/zipkin/blob/master/zipkin-collector-service/config/collector-cassandra.scala ?

codefromthecrypt: yep syntax is `bin/collector cassandra`

codefromthecrypt:
Don't merge, yet.

web is not working correctly via gradle. It seems to be missing something as the response is empty. The old script worked.

``` bash
$ curl -v http://localhost:8080/
* Hostname was NOT found in DNS cache
*   Trying ::1...
* Connected to localhost (::1) port 8080 (#0)
> GET / HTTP/1.1
> User-Agent: curl/7.37.1
> Host: localhost:8080
> Accept: */*
> 
* Empty reply from server
curl: (52) Empty reply from server
```

jamescway:
:+1:

bin/collector
blaines: Hey there I was looking at this pull request with James and I had some comments. Everything looks to be working here, I just noticed the "dev" default set in multiple places. What can be done to improve this a bit is to use the gradle.properties file to set the `serviceType` default. That will eliminate the need to grab system environment variables. Also another thing that can be done is to use `-P` which sets a project property that overrides the default. So you could end up with something more like:
`./gradlew :zipkin-collector-service:run -PserviceType=${1}`
(in this case it may be needed to have a default defined here as well, `-PserviceType=${1:-dev}` so as to not accidentally nullify the value, but I didn't test myself)

```
gradle --help
-P, --project-prop      Set project property for the build script (e.g. -Pmyprop=myvalue).
```

codefromthecrypt: Good point. Will use gradle.properties to default and -P to override.
Nullification concern from bash is still there, but at least the change is
more idiomatic. Thanks!

codefromthecrypt:
@blaines @jamescway PTAL I migrated the base config into gradle.properties. I think the default of "dev" in the two shell scripts is benign enough to leave alone.

I'm still looking into what's wrong with web. Also, sqlite broke some time ago.. I'll try to fix it in this change.

codefromthecrypt:
This won't work until #522 is merged. Still looking into what's up with the web script in gradle.

zipkin-collector-service/build.gradle
```diff
@@ -1,6 +1,13 @@
 apply plugin: 'application'
 mainClassName = 'com.twitter.zipkin.collector.Main'
 
+run {
+    // When configKey is dev, collector and query share ../zipkin.db
+    workingDir rootProject.projectDir
```
codefromthecrypt: ps. this is the 'ah-hah!' which was screwing up collector vs query. Basically each was creating their own DB files.

jamescway:
:+1:

abesto:
Cool! :+1:

zipkin-collector-service/build.gradle
rtyler: I suggest changing this to `rootProject.buildDir` that way the generated DB file will be placed in `build/` and naturally ignored by git and cleaned up by `./gradlew clean`

rtyler:
one suggestion but mostly looks fine to me

zipkin-collector-service/build.gradle
codefromthecrypt: SGTM

codefromthecrypt: so the root build directory isn't created by default.. I'll have to make sure it is.

codefromthecrypt: best way out was to point query to collector's build dir when configKey == "dev". The fundamental relationship is the coupling of these two processes, not to the root project. Ends up less work, too.

codefromthecrypt:
yay.. works now! PTAL

zipkin-web/build.gradle
```diff
@@ -2,6 +2,11 @@ apply plugin: 'application'
 apply plugin: 'com.github.johnrengelman.shadow'
 mainClassName = 'com.twitter.zipkin.web.Main'
 
+run {
+    args "-zipkin.web.resourcesRoot=${projectDir}/src/main/resources"
```
codefromthecrypt, line 6: I bet @dsyer loves the fact we have to do this :)

codefromthecrypt:
Note: I've changed the gradle run config to always start in the build dir. Once we do things like actually creating log files, this should help reduce junk creation.

jamescway:
:+1: :+1:

Issue #511, opened 2015-07-28
Document how to get started with SQLite


codefromthecrypt:
docker-compose is noted in our readme as the easiest way to get started.

However, docker-compose is pinned to cassandra, as is our other [quick-start](https://github.com/openzipkin/zipkin/blob/master/doc/mac-quickstart.md).

@irff and I'm sure others would like to use SQLite. It should have a README file saying how to setup this on your laptop, possibly also using docker-compose. The [existing db doc](https://github.com/openzipkin/zipkin/blob/master/doc/sql-databases.md) should be updated and referenced.

PR #522, opened 2015-07-30
Updates AnormDB and database library versions; Fixes SQLite

codefromthecrypt:
This updates all db libraries to latest, and sets the base version of
PostgreSQL to 9.3. It also corrects HikariCP configuration for SQLite,
which prevents it from erroring out on account of JDBC variant mismatch.

AnormDB was using old versions of just about everything, including very
EOL PostgreSQL. More importantly, SQLite stopped working due to a recent
HikariCP change, which moved the default JDBC variant to 4.

Relates to #511

### DIFF
--- a/zipkin-anormdb/build.gradle
+++ b/zipkin-anormdb/build.gradle
@@ -5,20 +5,21 @@
 // The default value of dbEngine is sqlite-persistent, defined in gradle.properties
 // To override, run the build like this: ./gradlew build -PdbEngine=postgresql
 ext.anormDriverDependencies = [
-        "sqlite-memory": "org.xerial:sqlite-jdbc:3.7.2",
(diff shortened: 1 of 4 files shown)


zipkin-anormdb/build.gradle
```diff
@@ -5,20 +5,21 @@
...
+        "postgresql": "org.postgresql:postgresql:9.3-1103-jdbc4",
+        "mysql": "mysql:mysql-connector-java:5.1.36"
 ]
 
 dependencies {
     compile project(':zipkin-common')
     compile project(':zipkin-scrooge')
 
-    compile "com.typesafe.play:anorm_${scalaInterfaceVersion}:2.3.7"
-    compile "com.zaxxer:HikariCP-java6:2.3.8"
```
codefromthecrypt, line 21: we don't support on JRE6 anyway..

jamescway:
:+1:

PR #528, opened 2015-07-30
Ports example configuration and instructions to gradle

codefromthecrypt:
Formerly, `zipkin-example` and `zipkin-redis-example` were described as
a series of SBT steps. This ports those steps to gradle, and backfills
configuration needed.


jamescway:
:+1:

PR #529, opened 2015-07-30
Removes SBT steps from ubuntu-quickstart; adds GRADLE_OPTS

codefromthecrypt:
ubuntu-quickstart included SBT steps which aren't needed, especially as
the bin scripts implicitly run gradle. Added GRADLE_OPTS to both mac and
ubuntu quick starts.


jamescway:
:+1:  might be good in the mac quick start too :p

codefromthecrypt:
done!

PR #531, opened 2015-07-31
Create AnormDB tables by default.

vprithvi:
- The default storage engine for AnormDB is sqlite::memory which
  starts out with a blank state. Creating tables is a more
  sensible default


codefromthecrypt:
agreed. thanks for the help!

PR #532, opened 2015-07-31
Add gradle idea plugin

vprithvi:
- This adds the gradle idea task which creates an IntelliJ project
  file.
  Invoke `./gradlew idea`, and open `zipkin.ipr` with IntelliJ

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,6 +5,10 @@ buildscript {
     }
 }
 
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
Agreed. this works, particularly if you resist the urge to later link as a gradle project. Thanks for the workaround!

PR #533, opened 2015-07-31
Uses Gradle daemon to assign memory args

codefromthecrypt:
Instead of telling developers to set GRADLE_OPTS, this uses the Gradle
daemon to implicitly provide mem args. This also solves the problem
where Travis builds timeout on PermGen on JDK 7.


.gitignore
```diff
@@ -1,5 +1,7 @@
 .idea/
 *.iml
+*.ipr
```
codefromthecrypt: unrelated, just didn't feel like opening a PR just to do this.

PR #535, opened 2015-07-31
Removes SBT

codefromthecrypt:
We've migrated Zipkin to Gradle for reasons including build speed and
skillset affinity of core contributors. This removes SBT as otherwise
it doubles the maintenance burden of the project.


abesto:
ZOMG

michaelsembwever:
YES!

PR #536, opened 2015-07-31
settings.gradle: Merge include statements into a single one

abesto:
For some reason having several include statements where the included projects depend on each other confuse IntelliJ and cause null-pointer exceptions when importing the project. Merging them into a single `include` fixes the issue.


codefromthecrypt:
Funny.. I noticed this, but didn't think to try it. So happy to have
multiple eyes on zipkin!

:shipit:

PR #540, opened 2015-08-03
Bundles scala config of collector and query services

### LINKED ISSUE #526: package config files for query and collector service into the jar
The query and collector service include pre-canned scala config files. These should be packaged into the jar and read from the classpath.

codefromthecrypt:
This bundles the `config` directory of the collector and query services.
By supporting a `-f` fallback to bundled resources, fat-jars can now
operate without needing to have a separate step to stage a configuration
file.

This also removes untested collector-zk files, one of which doesn't
compile.

Fixes #526


spencergibb:
:+1:

jamescway:
:+1:  nice pattern to follow

codefromthecrypt:
thx for the review @spencergibb @jamescway!

PR #541, opened 2015-08-04
Publish to bintray

### LINKED ISSUE #521: Publish to Bintray
Need to have versioned publishes to Bintray
- github tags (signifying a release) should cause a publish action to publish artifacts to bintray
- should call a script after_success from travis to publish if there are tags

jamescway:
### Problem

We need to release artifacts built in travis to Bintray openzipkin/zipkin repo.  Pushing to master should trigger a snapshot release, pushing with a tag (i.e. version) would trigger a release.
#### How it works
- user creates a tag git tag v1.2.3
- then does git push origin master --tags
- triggers the publish script to be called after travis build is successful
  - if no tag & push to master, then creates snapshot build
  - if tagged release, then pushes to bintray, updates version in gradle.properties 
- based on encrypted values in travis.yml
  - gem install travis
  - travis encrypt BINTRAY_API_KEY=<Bintray API key>--add
  - also BINTRAY_USER, GH_USER_EMAIL, GH_USER, GH_TOKEN
### Issues
- snapshot behavior not tested fully, need access from jfrog) this is done by clicking a button on the bintray package after its published
- Exception which may go away once we have snapshot access, but not sure why it appears, may be solved with access to oss.jfrog (artifactory) 

```
Execution failed for task ':zipkin-aggregate:bintrayUpload'.
> Could not create version '4.0.1-SNAPSHOT': HTTP/1.1 400 Bad Request [message:Unable to create version 4.0.1-SNAPSHOT at 'jamescway/zipkin/zipkin-aggregate']
```
### workarounds
- gradle bintrayUpload on main project is overridden to do nothing
- change git url to https instead of git://
- add git user to travis to allow pushing release commit msg from travis
- change branch to master
### To test this locally
- change all instances of openzipkin to your own org
- set env variables to your values BINTRAY_API_KEY, BINTRAY_USER, GH_USER_EMAIL, GH_USER, GH_TOKEN
- make sure you have a maven zipkin repo in bintray
- turn on github hooks in github for travis
- Test Release publishing
  - git tag 1.2.3  (any version)
  - git push origin --tags
- Test snapshot publishing
  - git push anything to master (no tags)

Fixes #521


build.gradle
```diff
@@ -5,14 +5,41 @@ buildscript {
...
     apply plugin: 'idea'
+    apply plugin: 'java'
+    apply plugin: 'scala'
+    apply plugin: 'maven'
+    apply plugin: 'maven-publish'
+    apply plugin: 'com.jfrog.bintray'
 }
 
-subprojects {
+release {
```
abesto, line 22: For the record (wasn't obvious for me): the README for `net.researchgate.release` says it'll only work in a multi-project build if the plugin is only applied to the root project, so we need to do this here instead of inside `subprojects`.

travis/publish.sh
```diff
@@ -0,0 +1,80 @@
...
+    exit 1
+  else
+    echo "[Publishing] Current JDK is the same as PUBLISH_USING_JDK"
+    echo "[Publishing]   environment variable (${TRAVIS_JDK_VERSION})"
+  fi
+}
+
+function publish_snapshots_to_bintray(){
+  echo "[Publishing] Starting Snapshot Publish..." &&
+  ./gradlew build -x test
```
abesto: Why do we need a `build` here? By this time that should be done, and if not, `bintrayUpload` should pull it in, non?

abesto:
Could you elaborate on what the issue "building from oraclejdk7" means?

build.gradle
jamescway, line 22: yeah thats correct, I realized that only after failing inside of the subprojects block many times.

travis/publish.sh
jamescway: @abesto so this was a question, does the build happen here already?  Are you saying that bintrayUpload should cause the build to happen?

jamescway:
Regarding "Building from oraclejdk7", is this the final version we're going with?

travis/publish.sh
jamescway: also it seems like `-x test` has stopped testing of the entire build O_o

jamescway: ^^^ is not correct, it was the code in travis under `script` tag.  Moved that into a `before_script` section.

vprithvi:
The maven plugins add the  `publishToMavenLocal` and  `publish` tasks which seem to not publish anything to the local maven repository in the `~/.m2` folder

codefromthecrypt:
@jamescway @rtyler mind trying to nail this PR? I'd love to have 1.2 cut
prior to debuildering.

jamescway:
@adriancole @abesto @rtyler any outstanding issues?  I ready to merge it.

abesto:
Don't see any blocking issues. I say :shipit:, refinements will be simpler anyway once the whole pipeline is set up.

jamescway:
@vprithvi publishing to local repo's wasn't something I tested.  If its a concern, lets create an issue for it and we can address in another PR.

PR #544, opened 2015-08-05
Workaround for marking config dirs as resource dir in idea

### LINKED ISSUE #543: ./gradlew idea mistakes zipkin-(query|collector)-service/config as a source folder
A project imported from `./gradlew idea` incorrectly has `zipkin-(query|collector)-service/config` set as a source folder, not a resources folder. This means you have to manually correct it before the project will compile or run. 

Click File, Project Structure, zipkin-collector-service or zipkin-query-service, config, Resources

This is a bad experience and should be fixable.

abesto:
The idea gradle plugin doesn't support marking source directories as resources when translating sourceSet declarations to an idea model. This workaround marks the "config" dir in a module as "resources" in idea, without touching anything else.

Fixes #543 

The workaround is from: https://discuss.gradle.org/t/the-idea-plugin-breaks-the-new-intellij-13-iml-configuration/2456

Related Gradle issues:
- https://issues.gradle.org/browse/GRADLE-3108
- https://issues.gradle.org/browse/GRADLE-2975

### DIFF
--- a/gradle/idea-mark-config-dir-as-resource.gradle
+++ b/gradle/idea-mark-config-dir-as-resource.gradle
@@ -1,0 +1,16 @@
+/**
+ * The idea gradle plugin (as of 2015. Aug. 05) doesn't support marking source directories as resources when
+ * translating sourceSet declarations to an idea model. This workaround marks the "config" dir in a module
+ * as "resources" in idea, without touching anything else.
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
You are my hero

:shipit:

PR #545, opened 2015-08-05
Share redis client across storage and index

gneokleo:
- Make the number of connections in the redis client configurable
  in order to gain considerably more throughput.
- Share the client across the storage and the index builders in order
  to maintain the same connection pool


codefromthecrypt:
pretty...

oh and great catch!

PR #548, opened 2015-08-06
add wait to kafka processor

jamescway:
#### Problem

In this case the "process" method is the ItemQueue and if it gets too full it throws queue full exceptions.  

The retry test code was in Hbase O_o so moving it here.  Going to make another PR on hbase to remove that.  Converted to scalatest.
#### Solution

This retry gives it a second to process spans so that it doesn't overflow, while limiting the total number of retries in case of something catastrophic.


codefromthecrypt:
so retry wasn't backing off before?!

jamescway:
it wasn't sleeping, I figured rather than make a new method, i modified this existing one.  Then added the retry code to the kafka code, because it didn't backoff and retry, it just died

codefromthecrypt:
get the idea, just wondering if some sort of already existing retry thing could work. will take a look in a bit

PR #550, opened 2015-08-06
Addresses deprecation in redis

codefromthecrypt:
Most of this is the deprecation of `Future.apply`. Unfortunately, the
alternative is more boilerplate.

### DIFF
--- a/zipkin-redis-example/src/main/scala/com/twitter/zipkin/example/Main.scala
+++ b/zipkin-redis-example/src/main/scala/com/twitter/zipkin/example/Main.scala
@@ -1,7 +1,7 @@
 package com.twitter.zipkin.example
 
 import com.twitter.finagle.Httpx
-import com.twitter.server.{Closer, TwitterServer}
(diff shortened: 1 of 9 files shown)


zipkin-redis-example/src/main/scala/com/twitter/zipkin/example/Main.scala
```diff
@@ -13,7 +13,7 @@ import com.twitter.zipkin.web.ZipkinWebFactory
 import com.twitter.zipkin.zookeeper.ZooKeeperClientFactory
 import com.twitter.zipkin.{thriftscala => thrift}
 
-object Main extends TwitterServer with Closer
+object Main extends TwitterServer
```
codefromthecrypt, line 16: closer is baked into App, and therefore TwitterServer

codefromthecrypt:
note there are many scala warnings.. this is just the deprecation side of things

jamescway:
:+1:

PR #551, opened 2015-08-06
Adding a blocking queue for waiting on span storer

gneokleo:
- Added a blocking queue that waits for the span storer
  to consume records instead of throwing a queue full exception.
  This queue will help in situations where the you're backfilling
  or when the receiver is a lot faster than the span storer (writer)

### DIFF
--- a/zipkin-collector/src/main/scala/com/twitter/zipkin/collector/BlockingItemQueue.scala
+++ b/zipkin-collector/src/main/scala/com/twitter/zipkin/collector/BlockingItemQueue.scala
@@ -1,0 +1,36 @@
+package com.twitter.zipkin.collector
+
+import com.twitter.conversions.time._
+import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
+import com.twitter.logging.Logger
+import com.twitter.util.{Duration, Future}
+
+/**
+ * This queue inherits from ItemQueue but instead of throwing an exception when its full it sleeps
+ * for a configurable amount of time until the span storer has enough time to consume records.
+ */
+class BlockingItemQueue [A, B](
+  maxSize: Int,
+  maxConcurrency: Int,
+  process: A => Future[B],
+  timeout: Duration = Duration.Top,
+  sleepOnFull: Duration = 1.seconds,
+  stats: StatsReceiver = DefaultStatsReceiver.scope("BlockingItemQueue"))
+    extends ItemQueue[A, B](maxSize, maxConcurrency, process, timeout, stats) {
+
+  val log = Logger(getClass())
+
+  override def add(item: A): Future[Unit] = {
+    if (!running) {
+      QueueClosed
+    }
+    while (!queue.offer(item)) {
+      queueFullCounter.incr()
+      log.debug(f"Queue is full, retrying: ${queue.size()}%d/$maxSize%d")
+      Thread.sleep(sleepOnFull.inMillis)
+      log.debug(f"After sleeping: ${queue.size()}%d/$maxSize%d")
+    }
+    Future.Done
+  }
+
+}
--- a/doc/redis.md
+++ b/doc/redis.md
@@ -16,3 +16,5 @@ If you have Redis installed and configured you can run Zipkin normally, except t
     bin/collector redis
     bin/query redis
     bin/web
+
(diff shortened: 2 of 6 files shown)


doc/examples/collector/src/main/scala/com/twitter/my-blocking-collector/MyBlockingCollectorService.scala
```diff
@@ -0,0 +1,30 @@
...
+import com.twitter.server.TwitterServer
+import com.twitter.util.{Await, Future}
+import com.twitter.zipkin.cassandra.CassieSpanStoreFactory
+import com.twitter.zipkin.collector.{SpanReceiver, ZipkinQueuedCollectorFactory}
+import com.twitter.zipkin.thriftscala.Span
+import com.twitter.zipkin.receiver.scribe.ScribeSpanReceiverFactory
+import com.twitter.zipkin.storage.WriteSpanStore
+import com.twitter.zipkin.zookeeper.ZooKeeperClientFactory
+
+object ZipkinCollectorServer extends TwitterServer
```
codefromthecrypt: I like the formatting etc, but not sure this is compiled.. maybe move to notes in ZipkinBlockingQueuedCollectorFactory's scaladoc? It might also be a decent idea to make this default behavior on collector...

zipkin-collector/src/main/scala/com/twitter/zipkin/collector/BlockingItemQueue.scala
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * This queue inherits from ItemQueue but instead of throwing an exception when its full it sleeps
+ * for a configurable amount of time until the span storer has enough time to consume records.
+ */
+class BlockingItemQueue [A, B](
+  maxSize: Int,
+  maxConcurrency: Int,
+  process: A => Future[B],
+  timeout: Duration = Duration.Top,
+  sleepOnFull: Duration = 30.seconds,
```
jamescway: What was the reason for defaulting to 30s?  I was seeing the queue having room after about 1s.

doc/examples/collector/src/main/scala/com/twitter/my-blocking-collector/MyBlockingCollectorService.scala
jamescway: Maybe moving this into [here (main.scala)](https://github.com/openzipkin/zipkin/blob/master/zipkin-collector-service/src/main/scala/com/twitter/zipkin/collector/Main.scala)? 

And another change to the command line args in build.gradle

zipkin-collector/src/main/scala/com/twitter/zipkin/collector/BlockingItemQueue.scala
gneokleo: This really depends on the application. I chose 30 because our application was seriously falling behind. I'll change it to 1s though. It sounds closer to what a lot of people will face.

doc/examples/collector/src/main/scala/com/twitter/my-blocking-collector/MyBlockingCollectorService.scala
gneokleo: @jamescway @adriancole What was the verdict on builder vs factory? We use the factory method so if that's the case I can move it to the file you suggested.

jamescway: i thought factories? :sweat:

jamescway:
@gneokleo if you have time you could copy your code into the main method of collector. 
Alternatively, we could merge this and file a ticket to migrate collector away from builders with this example.  @adriancole ?

gneokleo:
@jamescway I'll do another commit to move the code into the main method of the collector today.

codefromthecrypt:
Let's drop the last commit as it will certainly break people using "dev" or "redis" We do want to move to factories, but only when we can at least keep "dev/anorm", "redis" and "cassandra" working.

I'm good on the change, just a couple suggestions about test and logging. Good work!

doc/examples/collector/src/main/scala/com/twitter/my-blocking-collector/MyBlockingCollectorService.scala
codefromthecrypt: once this file is dropped.. ready to merge (on my side anyway)

codefromthecrypt:
great job

jamescway:
nice I can update the kafka code based on this commit

PR #553, opened 2015-08-07
update travis encryption org/repo

jamescway:
#### Problem

The encrypted values in travis were using the wrong org/repo.  This was causing the publish to not work.
#### Solution

Re-encrypt them using the right org/repo


rtyler:
i want to believe that this works

Issue #483, opened 2015-07-20
Lots of code duplication?


eirslett:
Looks like zipkin-common has lots of code duplication. Or is it just me not understanding the code?

E.g.
https://github.com/openzipkin/zipkin/blob/master/zipkin-common/src/main/scala/com/twitter/zipkin/storage/Storage.scala
has many of the same methods as
https://github.com/openzipkin/zipkin/blob/master/zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala

I'm wondering if it's https://github.com/openzipkin/zipkin/pull/321 that was merged with new interfaces, without removing the old ones. @sprsquish @bmdhacks @franklinhu could that be right?

codefromthecrypt:
Feel free to dedupe.

sprsquish:
@eirslett SpanStore was me trying to update the core interfaces. I knew many OSS users would still be on the old ones, so I was going to deprecate and remove them over time.

codefromthecrypt:
There's a lot of duplication the following modules, which surprisingly are peers
- zipkin-collector is sortof a subset of zipkin-collector-core
- zipkin-query is sortof a subset of zipkin-query-core

Most things depend on the *-core variants. I'll give a go at unwinding this and deleting the non-core ones.

codefromthecrypt:
For starters, looks like ThriftQueryService and QueryService are sporadically maintained relatives. I'll try to merge the fixes from one into the other.

codefromthecrypt:
ZipkinQueuedCollectorFactory (from zipkin-query) is in use at twitter, while QueryService (from zipkin-query-core) is what's currently in use here.

Similarly, ZipkinQueuedCollectorFactory (from zipkin-collector) is in use at twitter, while ZipkinCollector (from zipkin-query-core) is what's currently in use here.

In both cases, these are flag-based alternatives to the eval-based FooBuilder approach noted as undesirable in #466 

In both cases, there's functionality risk and confusion as both sides are independently maintained, and some actually mask classes (ex. look at adjusters package).

My recommendation is to declare bankruptcy: choose the Factory based classes and delete the others. Delete other things that were intended to be marked as deprecated. There's no advantage to having two partially understood trees of code, and the cost of maintenance of either is high enough.

codefromthecrypt:
punting this to a discussion for now https://groups.google.com/forum/#!topic/zipkin-dev/GY_jEI5pQY8

PR #558, opened 2015-08-07
Removes files duplicated between zipkin-query and zipkin-query-core

codefromthecrypt:
zipkin-query-core contained literal copies of the same files in its
parent zipkin-query. This is beyond confusing.

See #483


abesto:
:clap: screenshot 2015-08-08 08 04 54 :clap: whooooo :clap: :shipit:

PR #560, opened 2015-08-08
fix bintray publish subproject description

jamescway:
Fix by changing single quotes to double quotes, and add some more description.


abesto:
LGTM :+1:

PR #524, opened 2015-07-30
Add build config for executable jars

dsyer:
All the executable projects (the zipkin-*-services and zipkin-web)now have a *-all.jar which is executable. Still to do: package theconfig files for the collector service along with the code (at the
moment you have to have them locally and run with `-f <path_to>/collector-<xxx>/scala`).

```
$ find . -name \*-all.jar
./zipkin-redis-example/build/libs/zipkin-redis-example-1.2.0-SNAPSHOT-all.jar
./zipkin-query-service/build/libs/zipkin-query-service-1.2.0-SNAPSHOT-all.jar
./zipkin-example/build/libs/zipkin-example-1.2.0-SNAPSHOT-all.jar
./zipkin-web/build/libs/zipkin-web-1.2.0-SNAPSHOT-all.jar
./zipkin-collector-service/build/libs/zipkin-collector-service-1.2.0-SNAPSHOT-all.jar
./zipkin-tracegen/build/libs/zipkin-tracegen-1.2.0-SNAPSHOT-all.jar
```

Example launch:

```
$ java -jar zipkin-query-service/build/libs/zipkin-query-service-1.2.0-SNAPSHOT-all.jar -f zipkin-query-service/config/query-redis.scala 
Jul 30, 2015 12:02:27 PM com.twitter.zipkin.query.Main$ main
INFO: Loading configuration
INF [20150730-12:02:34.908] stats: Starting LatchedStatsListener
700 [20150730-12:02:35.041] net: HttpServer created http 0.0.0.0/0.0.0.0:9901
700 [20150730-12:02:35.064] net: context created: /
700 [20150730-12:02:35.065] net: context created: /report/
700 [20150730-12:02:35.066] net: context created: /favicon.ico
700 [20150730-12:02:35.067] net: context created: /static
700 [20150730-12:02:35.068] net: context created: /pprof/heap
700 [20150730-12:02:35.069] net: context created: /pprof/profile
700 [20150730-12:02:35.070] net: context created: /pprof/contention
700 [20150730-12:02:35.071] net: context created: /tracing
700 [20150730-12:02:35.072] net: context created: /admin/registry
700 [20150730-12:02:35.073] net: context created: /health
700 [20150730-12:02:35.073] net: context created: /quitquitquit
700 [20150730-12:02:35.073] net: context created: /abortabortabort
700 [20150730-12:02:35.083] net: context created: /graph/
700 [20150730-12:02:35.084] net: context created: /graph_data
[...]

PR #561, opened 2015-08-08
Upload fat jars to Bintray

abesto:
This change uploads the fat jar generated by the `shadow` plugin while leaving the current files in place. For an example, see https://bintray.com/abesto/zipkin-test/zipkin-web/view#files
- `zipkin-web/1.2.2/zipkin-web-1.2.2.jar` is a plain old jar, with no included dependencies (but the dependencies declared in the corresponding `.pom` file)
- `zipkin-web/1.2.2/zipkin-web-1.2.2.{tar.zip}` are the distributions created by the Gradle `application` plugin, with the dependencies included in a subfolder and shell scripts to launch the application
- `zipkin-web/1.2.2/zipkin-web-1.2.2-all.jar` is the fat jar generated by the Gradle `shadow` plugin

### DIFF
--- a/zipkin-aggregate/build.gradle
+++ b/zipkin-aggregate/build.gradle
@@ -17,6 +17,7 @@ run {
 }
 
 tasks.build.dependsOn(shadowJar)
+artifacts.archives shadowJar
 
 repositories {
     // For dependencies of zipkin-cassandra
--- a/zipkin-collector-service/build.gradle
+++ b/zipkin-collector-service/build.gradle
@@ -10,6 +10,7 @@ run {
 }
 
 tasks.build.dependsOn(shadowJar)
+artifacts.archives shadowJar
 
 repositories {
     // For dependencies of zipkin-cassandra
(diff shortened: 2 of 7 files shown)


abesto:
@jamescway @rtyler

abesto:
Note: the jars in the test bintray repo were built for jre 1.8, won't work on 1.7; the actual build process uploads only from 1.7

rtyler:
@abesto it's worth mentioning that dependencies in the shadow jar are unzipped, so there's a whole lot of class files running around. I'm assuming that you've already run this code to make sure that it loads the application correctly?

Other than that concern, :+1:

abesto:
Yep, and also in https://github.com/openzipkin/zipkin/pull/524. Still, it'd be nice not having to do this manually, let's track that in https://github.com/openzipkin/zipkin/issues/562

jamescway:
:+1: very cool so basically is has all the dependencies

PR #569, opened 2015-08-10
Fixes ./gradle idea "Cannot infer Scala class path because.." failures

### LINKED ISSUE #565: ./gradlew idea failure
other tasks work, but idea isn't

``` bash
$ ./gradlew  idea
:ideaProject FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':ideaProject'.
> Cannot infer Scala class path because no repository is declared in root project 'zipkin'
```

codefromthecrypt:
The idea goal failed due to problems understanding scala. This adds a
repository and the current scala version to the parent to address that.

Fixes #565

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,17 @@ allprojects {
     apply plugin: 'maven'
     apply plugin: 'maven-publish'
     apply plugin: 'com.jfrog.bintray'
+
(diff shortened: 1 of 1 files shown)


rtyler:
:+1:

PR #570, opened 2015-08-10
Replace cassie with datastax cql driver

michaelsembwever:
This work is sill premature but is opened for those that want an early review on the branch.

Currently the to do list is…
- plug in a cassandra test framework (CCMBridge?) so the tests run against an embedded cassandra running (right now you need a cassandra instance running for the tests to work, the schema is self-creating so just a blank cassandra installation started up will work)
- fix the remaining tests. i do suspect CassandraSpanStore.getSpansByTraceIds(..) needs a little work still.
- double check the tests that were against Index/Storage/Aggregates still exist in some form against SpanStorage (see the commented out code in CassieSpanStoreTest)
- rename packages and classes appropriately now that things are no longer built on Cassie
- pass flags in CassandraSpanStoreFactory into the Cluster setup

The branch consists of four commits.
-  first commit is a temporary commit to comment out the redis stuff that was getting in my way, (this will be removed from the branch and PR)
- second comment is moving the hadoop aggregate stuff out of zipkin-cassandra over to zipkin-cassandra-aggregate (the latter is not in the reactor so not built – it's not compilable as it still references cassie)
- third commit is the actual code changes involved in replacing cassie with the datastax driver.
- fourth commit is removing the cassie classes, and also removing the Index/Storage/Aggregate classes.

i plan on squashing the last 3 commits into one eventually.

Otherwise the primary fasade to the datastax cql java driver is in the java class "Repository" found in the zipkin-cassandra-java subproject. (There exists rationale to have it in its own subproject). It allows automatic creation and updating of the schema, and is a good starting place for improving the datamodel. 

The datamodel does change as keyspace and table names have become all lowercase.
I've also added comments to the cql file regarding datamodel improvements that should happened afterwards (once this PR has been stabilised and is in master).


codefromthecrypt:
Good progress! Here are some thoughts:
- I wouldn't block this on embedded cassandra, as that's polish. We can just use travis service/local cassandra for now (like we do today).
- Mixed feelings about the separate `-java` and `-aggregate` projects. The artifact graph is already complex, and something we've been working hard to make sense of. Can we delay this optimization? For example, add a java source tree and make the deps for `-aggregate` optional. 

I can help look at the test, but any chance you can move this branch to openzipkin? Easier for us to collaborate there.

codefromthecrypt:
ps.. given a choice, I'd just delete cassandra-aggregate vs move it to an incompatible peer project. I'm sure @eirslett if not others could make short work of that once the core cassandra is in.

codefromthecrypt:
Based on offline chat, here's my take:
- the `-java` project is already in use for a project that writes traces directly into cassandra vs via scribe etc. Maybe the name could be changed, but understand where that's coming from. Personally, I prefer `-core` suffix for library code.
- we should definitely delete the -aggregate codebase as opposed to orphaning it in a way that doesn't compile. This is part of collateral damage needed to get zipkin running again. Hopefully, @eirslett could help add a new `-aggregate` codebase after this is merged (based on git history or from scratch).

Need some folks to weigh in, as I'd like to depend on this work in order to complete other work (like removing Index/Store, and the Builder stuff).

zipkin-cassandra-java/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -0,0 +1,749 @@
...
+            if (LOG.isDebugEnabled()) {
+                LOG.debug(selectTraces.getQueryString()
+                        .replace(':' + KEY_MARKER, Arrays.toString(traceIds))
+                        .replace(':' + LIMIT_MARKER, String.valueOf(limit)));
+            }
+            Map<Long,List<ByteBuffer>> spans = new HashMap<>();
+            for (Row row : session.execute(bound).all()) {
+                long traceId = row.getBytes("key").getLong();
+                if (!spans.containsKey(traceId)) {
+                    spans.put(traceId, new ArrayList<>());
```
codefromthecrypt: for JDK 7 to work, replace this with `new ArrayList<ByteBuffer>()`

jamescway:
answering @adriancole's points
- the name of the pure-java impl
  - not sure if java needs to be in the name since its self evident?  +1 to core or something based on current name scheme.
- whether it is ok to ditch cassandra-aggregates temporarily
  - :+1: , but don't know if there's people out there that might be in an awkward spot, hopefully they come on gitter and ask for help.

zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassieSpanStoreFactory.scala
```diff
@@ -15,102 +15,22 @@
...
 import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
-import com.twitter.finagle.tracing.DefaultTracer
-import com.twitter.finagle.util.InetSocketAddressUtil
-import com.twitter.util.{Await, Future, Promise, Return, Var}
 import com.twitter.zipkin.thriftscala.{Span => ThriftSpan}
 import com.twitter.zipkin.storage.cassandra._
-import java.net.SocketAddress
-import scala.collection.mutable.HashSet
 
 trait CassieSpanStoreFactory { self: App =>
```
codefromthecrypt: rename to CassandraSpanStoreFactory

michaelsembwever:
the `-core` suffix is a nice choice over `-java`.  so +1 on that.

also +1 on removing `-aggregate`. just so long enough people know where it went if questions are raised later on by those that didn't catch it here.

zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassieSpanStoreFactory.scala
michaelsembwever: +1
 (was intentionally leaving it until later in the PR lifecycle so the diff on the file remains clear)

zipkin-cassandra-java/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
michaelsembwever: thanks for spotting that Adrian. 

java7 isn't specified in the build.gradle yet? or is it just my ide not picking it up maybe.

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
```diff
@@ -226,47 +175,41 @@ class CassieSpanStore(
...
-
-            case cols =>
-              Some(cols.toSeq map { case (_, col) => col.value.toSpan })
-          }
-        }
+      pool {
+        val spans = repository.getSpansByTraceIds(idBatch.toArray.map(Long.box), count)
+          .mapValues { case spans :java.util.List[ByteBuffer] => spans.asScala.map(spanCodec.decode(_).toSpan) }
+
+        traceIds.map(traceId => spans.get(traceId)).filter(_.isDefined).map(_.get).toSeq
```
eirslett: Is it possible to do something like this?
`traceIds.map(traceId => spans.get(traceId).flatten)`

`.flatten` will (example given) turn `Seq(None, None, Some(1), None, Some(2))` into `Seq(1,2)` so you don't need to filter by isDefined and map on .get.

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
```diff
@@ -277,71 +220,48 @@ class CassieSpanStore(
...
   }
 
   def setTimeToLive(traceId: Long, ttl: Duration): Future[Unit] = {
-    Traces.getRow(traceId) flatMap { row =>
-      val traces = Traces.batch()
-      row.values.asScala foreach { col =>
-        traces.insert(traceId, col.copy(timestamp = None, ttl = Some(ttl)))
-      }
-      traces.execute().unit
+    getSpansByTraceId(traceId).get foreach { span =>
```
eirslett, line 244: `.get` is dangerous... (purists say it shouldn't be in the Scala library in the first place)
Maybe `.flatten` works here too?

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
```diff
@@ -277,71 +220,48 @@ class CassieSpanStore(
...
   }
 
   def tracesExist(traceIds: Seq[Long]): Future[Set[Long]] = {
     QueryTracesExistStat.add(traceIds.size)
-    getSpansByTraceIds(traceIds, 1) map {
-      _.flatMap(_.headOption.map(_.traceId)).toSet
+    pool {
+      repository
+        .tracesExist(traceIds.toArray.map(Long.box))
+        .map(_.asInstanceOf[Long])
```
eirslett, line 263: I think this is the same as `.map(Long.unbox)` (mentioning it since `Long.box` was used above)

zipkin-cassandra-java/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
codefromthecrypt: Travis builds w/JDK7 and 8. JDK7 is particularly important to retain due to
current scala version 2.10

eirslett:
Wow, this is some Great work! :grinning: :clap: :+1:

zipkin-cassandra/src/test/scala/com/twitter/zipkin/storage/cassandra/CassieSpanStoreTest.scala
```diff
@@ -15,26 +15,320 @@
...
+import com.twitter.util.Await
 import com.twitter.zipkin.cassandra.CassieSpanStoreFactory
+import com.twitter.zipkin.common._
+import com.twitter.zipkin.query.Trace
 import com.twitter.zipkin.storage.util.SpanStoreValidator
-import org.scalatest.FunSuite
+import org.scalatest.{BeforeAndAfter, FunSuite}
+
 
 class CassieSpanStoreTest extends FunSuite {
```
codefromthecrypt: currently trying to work around our famous `org.apache.thrift:libthrift:0.5.0-1` vs `0.9.2` runtime compat story.

eirslett:
We can run zipkin-aggregate off a released version of zipkin-cassandra (with cassie). For now. But I'm pretty sure @michaelsembwever has some experience working with Cassandra taps?

codefromthecrypt:
Great point. Once our jars are published, we could switch the dep in the
aggregated module until the new code is ported over.

michaelsembwever:
rebased off master and included commits from @adriancole 's branch in https://github.com/thelastpickle/zipkin/pull/1

zipkin-cassandra-aggregate is removed.
zipkin-cassandra-java is now called zipkin-cassandra-core.

CassieSpanStoreFactory renamed to CassandraSpanStoreFactory
CassieSpanStoreSpec renamed to CassandraSpanStoreSpec

from here i'll fix up the tests and incorporate @eirslet's improvements.

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
```diff
@@ -389,27 +313,33 @@ class CassieSpanStore(
...
+    val traceIdSet = traceIds.toArray
+
+    pool {
+      (repository.getTraceDuration(true, traceIdSet)
+      .map { case (traceId :java.lang.Long, ts :java.lang.Long) => (traceId, ("start", ts))}
+      .toSeq ++
+        (repository.getTraceDuration(false, traceIdSet)
+        .map { case (traceId :java.lang.Long, ts :java.lang.Long) => (traceId, ("end", ts))}
+        .toSeq))
+      .groupBy(_._1)
```
eirslett: I suggest we give ._1 and ._2 names, for readability:

``` scala
.groupBy { case (whatevernameofthekey, _) => whatevernameofthekey }
```

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
```diff
@@ -389,27 +313,33 @@ class CassieSpanStore(
...
+      .mapValues(_.map(_._2))
```
eirslett, line 340: I think you can use

``` scala
.mapValues(_.values)
```

or something with .flatten? (I'm not sure if that's possible)

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
eirslett, line 244: Ah, it's a Future. You'll have to use

``` scala
getSpansByTraceId(traceId) map { spans => spans foreach { span =>
```

Then you won't need to return Future.Unit in the last line.

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
```diff
@@ -374,10 +294,14 @@ class CassieSpanStore(
 
     // if we have a span name, look up in the service + span name index
     // if not, look up by service name only
-    val idx: ColumnFamily[String, Long, Long] =
-      spanName.map(_ => ServiceSpanNameIndex).getOrElse(ServiceNameIndex)
-    // TODO: endTs seems wrong here
-    idx.getRowSlice(key, Some(endTs), None, limit, Order.Reversed) map colToIndexedTraceId
+    pool {
+      (spanName match {
```
eirslett, line 296: An intermediary variable could be used here, for readability?

```
val traceIds = spanName match { ... }
traceIds.map { case ... }
```

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
```diff
@@ -277,71 +220,48 @@ class CassieSpanStore(
...
-      val ttl = row.values.asScala.foldLeft(Int.MaxValue) { (ttl, col) =>
-        math.min(ttl, col.ttl.map(_.inSeconds).getOrElse(Int.MaxValue))
-      }
 
-      if (ttl == Int.MaxValue)
-        Future.exception(new IllegalArgumentException("The trace " + traceId + " does not have any ttl set!"))
-      else
-        Future.value(ttl.seconds)
+    pool {
+      Duration(repository.getSpanTtl(traceId), java.util.concurrent.TimeUnit.SECONDS)
```
eirslett: How about `repository.getSpanTtlInSeconds(traceId)`? A longer method name to compansate for the lack of time unit?

michaelsembwever: done

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
michaelsembwever, line 296: i've left this. i can't think of an intuitive variable name here, hence the comment.

the closest variable name i can think of is timestampByTraceId, but i find that not intuitive enough for two lines of code.

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
michaelsembwever, line 340: i played around with `.mapValues(_.values)` and `.flatten` and found nothing that worked :-(

i'm open to working suggestions otherwise.

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
michaelsembwever: done

codefromthecrypt:
Here are some hints on how to retain support for Builder w/o having to implement Index or Storage:

there's a bridge you can use ex here: https://github.com/openzipkin/zipkin/blob/master/zipkin-collector-service/config/collector-redis.scala#L30
also here https://github.com/openzipkin/zipkin/pull/583/files#diff-1b0789c140f012e2404ae48339a50569

codefromthecrypt:
The following commit completes this change https://github.com/adriancole/zipkin/commit/84731aca6a4d162bce404580723be874beaef0d9

I'll raise a PR, squashing that into mck's work.

codefromthecrypt:
raised #585 to merge this in (removing the temp commits and squashing)

PR #571, opened 2015-08-11
Replaces SpanStoreValidator with SpanStoreSpec

codefromthecrypt:
SpanStoreValidator was in src/main and included a bespoke unit test
framework. This is harder to understand than normal unit tests, clutters
the main source tree, and makes debugging test failures custom.

This introduces a similar class, SpanStoreSpec, which uses normal
assertions and supports JUnit annotations. This means that you can use 
IDE support such as rerunning a specific test, override or ignore test 
methods, or even use test rules to start your backends. Finally, this
gets the test code out of the main source tree.


adriancole:
Don't merge, yet. I need to change this to skip the two failing cassie tests.

codefromthecrypt:
ok PTAL

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
```diff
@@ -116,7 +116,7 @@ class AnormSpanStore(
               ("value" -> Util.getArrayFromBuffer(b.value)),
               ("annotation_type_value" -> b.annotationType.value),
               ("ipv4" -> b.host.map(_.ipv4)),
-              ("port" -> b.host.map(_.ipv4)))
+              ("port" -> b.host.map(_.port)))
```
codefromthecrypt, line 119: noticed this bug lying around..

jamescway, line 119: hopefully its not a poorly documented hack O_o

jamescway:
Interesting that you uncovered the testing framework.  I restarted the travis job, but there was a problem with the redis span store spec.  Not sure if this is an intermittent failure or not.

codefromthecrypt:
yeah, well redis didn't actually run the tests prior (only did clear() actually!), so there may be some flakes to squash over time. beats zero coverage, though!

codefromthecrypt:
ok build's green. I will help with flakes or disabling things that flake on certain backends.

Issue #559, opened 2015-08-07
Update gradle-bintray-plugin to 1.3 when its released


jamescway:
See https://github.com/bintray/gradle-bintray-plugin/issues/68

We need update the plugin when it is released.

This should solve the problem, when jars are bintrayUpload'ed they are not published and you need to bintray.com to click the publish link.

codefromthecrypt:
This should hopefully result in publication of the fat jars cc @dsyer

abesto:
Fat jars not being published was a different issue, fixed in https://github.com/openzipkin/zipkin/pull/561

abesto:
According to https://bintray.com/jfrog/jfrog-jars/gradle-bintray-plugin/view 1.3 has been out for a couple of days (just the readme is not updated; but the badge in the readme is). Is there anything holding us back from upgrading?

codefromthecrypt:
Dooo iittt ;)

PR #572, opened 2015-08-11
Use Bintray Gradle plugin 1.3.1, fixing #559

abesto:
Quoting #559:

See bintray/gradle-bintray-plugin#68

We need update the plugin when it is released.

This should solve the problem, when jars are bintrayUpload'ed they are not published and you need to bintray.com to click the publish link.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
 }
 
 plugins {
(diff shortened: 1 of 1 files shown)


jamescway:
:+1:

codefromthecrypt:
:shipit:

PR #573, opened 2015-08-11
Upload to a single package on Bintray

abesto:
This simplifies maintenance and configuration, especially when it comes to syncing to Maven Central.

For an example of what the result looks like: https://bintray.com/abesto/zipkin-test/zipkin/view#files
The artifacts are not published (=made public) on Bintray, expect another PR on that.

Side-note: uploading stuff takes long. Most of the fat jars are ~80MB, and they have a `.zip` and a `.tar` friend.


codefromthecrypt:
:clap:

PR #574, opened 2015-08-11
Removes duplicate adaptive zookeeper sampler

codefromthecrypt:
Collector used to have its own ZooKeeper sampler. This was written in
2012 and has been relatively untouched since. SpanStore introduced a
replacement, `zipkin-adaptive`, which is used inside twitter and can
be used by anyone caking up their collector `with AdaptiveSampler` and
 `override def spanStoreFilter = newAdaptiveSamplerFilter()`.

Removing the redundant implementation significantly reduces tech debt
and also a direct ZK dependency in `zipkin-collector-core`.

See #483


codefromthecrypt:
deploys fine internally at twitter

abesto:
:+1: looks good. So much red!

codefromthecrypt:
will merge on green. Thanks for having a look @abesto!

eirslett:
Is zipkin the most continuously delivered product at Twitter now? :stuck_out_tongue_winking_eye:

codefromthecrypt:
must be in the runnings :)

PR #577, opened 2015-08-11
Fix: the Bintray package description had the name of a subproject

abesto:


codefromthecrypt:
makes sense

jamescway:
:+1:  cool new yield obj

PR #578, opened 2015-08-11
Add support for locally publishing maven artifacts

vprithvi:
- `./gradlew publishToMavenLocal` installs all zipkin artifacts to your
  local maven repository

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,6 +25,7 @@ allprojects {
 
     // - no Scala library Jar was found
     apply from: "$rootDir/gradle/dependencies.gradle"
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
weird.. I thought `./gradlew install` already did this?

vprithvi:
@adriancole I entirely missed reading that task as  `./gradlew tasks`  placed `publishToMavenLocal` before the `install` task. We might just need to hide the `publishToMavenLocal` task

codefromthecrypt:
Sounds confusing for sure.

PR #580, opened 2015-08-11
Adds getDataTimeToLive to ReadOnlySpanStore

codefromthecrypt:
The query thrift api requires support of getDataTimeToLive, which is on
`Storage`, but not yet `SpanStore`. Adding this makes it possible to
implement the legacy `Storage` interface with a `SpanStore`.


jamescway:
:+1: lgtm

PR #581, opened 2015-08-11
Adds SpanStoreStorageWithIndexAdapter to bridge SpanStore to legacy code

codefromthecrypt:
The Index and Storage traits need to be removed, but we need to approach
this is steps. Firstly, we need to get the existing Builder approach to
work with SpanStore. That way, we can keep the query and collector
services.

This makes a bridge called `SpanStoreStorageWithIndexAdapter` and ports
redis to use it. Other ports will follow.


zipkin-redis/src/main/scala/com/twitter/zipkin/redis/StorageWithIndexBuilder.scala
```diff
@@ -0,0 +1,26 @@
...
+import com.twitter.conversions.time._
+import com.twitter.finagle.redis.Client
+import com.twitter.finagle.redis.util.StringToChannelBuffer
+import com.twitter.util.{Await, Duration}
+import com.twitter.zipkin.builder.Builder
+import com.twitter.zipkin.storage.{SpanStoreStorageWithIndexAdapter, Storage, Index}
+import com.twitter.zipkin.storage.redis.RedisSpanStore
+
+/** Allows [[RedisSpanStore]] to be used with legacy [[Builder]]s. */
+case class StorageWithIndexBuilder(
```
codefromthecrypt, line 12: This is the last type in redis that references `Storage` or `Index`!

jamescway:
:+1:

PR #582, opened 2015-08-11
Fix ItemQueue/BlockingItemQueue tests

### LINKED ISSUE #566: Seeing intermitent failures with ItemQueueTest
I've been seeing this test fail (maybe non-deterministically) in travis for various branches.  Often times a retry makes it green.  I hope we're not calling those crazy `rand` methods in finagle-thrift :trollface:.

```
ItemQueueTest:
- processes messages (1 second, 135 milliseconds)
- runs a specified number of concurrent workers (18 milliseconds)
- enforces a max queue size (1 millisecond)
- wont accept items after being closed (37 milliseconds)
- drains after being closed *** FAILED *** (159 milliseconds)
  processed.await(100L, MILLISECONDS) was false (ItemQueueTest.scala:97)
Run completed in 1 second, 688 milliseconds.
Total number of tests run: 5
Suites: completed 2, aborted 0
Tests: succeeded 4, failed 1, canceled 0, ignored 0, pending 0
*** 1 TEST FAILED ***
ItemQueueTest:
- drains after being closed *** FAILED *** (159 milliseconds)
  processed.await(100L, MILLISECONDS) was false (ItemQueueTest.scala:97)
:zipkin-collector:test FAILED
```

jamescway:
### Problem

Have been seeing intermittent build failures in travis, this is a stab at fixing these errors. 
### Solution

ItemQueueTest
- adding a sleep in the Items causes the error to be manifested, it shows the items aren't getting drained out of the queue.  The theory is that in travis sometimes the latch takes longer than 100ms to close.  Thus increasing timeout to 5s, which should cover most cases.

BlockingItemQueueTest
- If you add a sleep after the initial fill(), then test adding an overflow item, you see that the test fails.  I believe the queue sometimes drains too fast, thus we never overflow.  Switching to use a countdown latch which sync's all the items to exit at the same-ish time.  Similar to ItemeQueueTest's usage.

fixes #566

### DIFF
--- a/zipkin-collector/src/test/scala/com/twitter/zipkin/collector/BlockingItemQueueTest.scala
+++ b/zipkin-collector/src/test/scala/com/twitter/zipkin/collector/BlockingItemQueueTest.scala
@@ -2,16 +2,15 @@ package com.twitter.zipkin.collector
 
 import com.twitter.conversions.time._
 import com.twitter.finagle.stats.InMemoryStatsReceiver
-import com.twitter.util.{Await, Future}
+import com.twitter.util.{Await, Future, FuturePool}
+import java.util.concurrent.CountDownLatch
+import scala.util.Try
 import org.scalatest._
 
-/**
- * Tests the BlockingItemQueue to make sure that it can store and consume elements even when adding
- * more elements than what its initial capacity is
- */
 class BlockingItemQueueTest extends FunSuite {
 
   val Item = ()
+  val latch = new CountDownLatch(1)
 
   def fill(queue: BlockingItemQueue[Unit, Unit], items: Int): Future[Boolean] = {
     val results = (0 until items) map { _ =>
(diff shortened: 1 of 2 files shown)


zipkin-collector/src/test/scala/com/twitter/zipkin/collector/BlockingItemQueueTest.scala
```diff
@@ -21,23 +20,48 @@ class BlockingItemQueueTest extends FunSuite {
...
+
     // Add 11 and not 10 because the first one that's going to be added will be consumed right away
     assert(Await.result(fill(queue, 11)))
-    assert(Await.ready(queue.add(Item)).poll.get.isReturn)
+
+    // add an item, expect the queueFull stat to get incremented we need this to return because
+    // the blocking queue will block because its full, so we use a future
+    FuturePool.unboundedPool { queue.add(Item) }
+    // poll for when the queueFull counter gets incremented or timeout
+    poll(queueFull)
```
codefromthecrypt: I'd inline the lambdas, as the val declarations above make reading this complicated.
something like

``` scala
assertEventually(() => stats.counter("queueFull")() >= 1)
```

zipkin-collector/src/test/scala/com/twitter/zipkin/collector/BlockingItemQueueTest.scala
```diff
@@ -21,23 +20,48 @@ class BlockingItemQueueTest extends FunSuite {
...
 
   test("Sleeps on a max queue and waits for the worker to drain") {
-
     val stats:InMemoryStatsReceiver = new InMemoryStatsReceiver()
-    val queue = new BlockingItemQueue[Unit, Unit](10, 1, fallingBehindWorker, 100.millis,
-                                                  100.millis, stats)
+    val queue = new BlockingItemQueue[Unit, Unit](10, 1, fallingBehindWorker, 100.millis, 100.millis, stats)
+    val queueFull = () => stats.counter("queueFull").apply() >= 1
+    val queueSizeZero = () => queue.size() == 0
+    val successSize = () => stats.counter("successes").apply() == 12
```
codefromthecrypt: val expectedItemCount = 12

zipkin-collector/src/test/scala/com/twitter/zipkin/collector/BlockingItemQueueTest.scala
```diff
@@ -21,23 +20,48 @@ class BlockingItemQueueTest extends FunSuite {
...
-    assert(stats.counter("queueFull").apply() >= 1)
-    assert(stats.counter("successes").apply() == 12)
-    assert(queue.size() == 0)
+
+    // All items should be processed
+    poll(successSize)
+    // Final queue should be drained
+    assert(queueSizeZero())
+    assert(queueFull())
+    assert(successSize())
```
codefromthecrypt: remove as this is redundant

zipkin-collector/src/test/scala/com/twitter/zipkin/collector/BlockingItemQueueTest.scala
```diff
@@ -21,23 +20,48 @@ class BlockingItemQueueTest extends FunSuite {
...
     Await.ready(queue.close())
-    assert(stats.counter("queueFull").apply() >= 1)
-    assert(stats.counter("successes").apply() == 12)
-    assert(queue.size() == 0)
+
+    // All items should be processed
+    poll(successSize)
+    // Final queue should be drained
+    assert(queueSizeZero())
+    assert(queueFull())
```
codefromthecrypt: The combination of the above 2 asserts are confusing... the queue has a size zero and it is full?

zipkin-collector/src/test/scala/com/twitter/zipkin/collector/BlockingItemQueueTest.scala
```diff
@@ -21,23 +20,46 @@ class BlockingItemQueueTest extends FunSuite {
...
-    assert(stats.counter("successes").apply() == 12)
+
+    // All items should be processed
+    poll( () => stats.counter("successes").apply() == expectedItemCount )
     assert(queue.size() == 0)
+    assert(stats.counter("queueFull").apply() >= 1)
+  }
+
+  def poll[A](f: () => Boolean): Boolean = {
+    val func = FuturePool.unboundedPool { do_poll(f) }
```
codefromthecrypt: does this need to use futures? why not just loop?

jamescway: For the timeout

jamescway: It could also be done without future and pass the end time

codefromthecrypt: yeah was thinking like `for (1 to count) if (foo.apply()) return true; else sleep 1000` in scala. The thread stuff is distracting.

jamescway:
added longer timeouts for all Awaits in ItemQueueTest, due to failures in travis

codefromthecrypt:
thanks!

PR #583, opened 2015-08-12
Removes duplicate code in AnormDB

codefromthecrypt:
AnormDB had two similar implementations, SpanStore and Storage/Index.
The latter is a legacy, but had more recent maintenance than the former.

This change migrates the most recent code into AnormDB's SpanStore and
removes the code for Storage/Index.

See #483


codefromthecrypt:
verified instructions in README still work, as well.

eirslett:
+1 LGTM! :beers:

Issue #458, opened 2015-07-16
Replace Cassie with Datastax native driver


rtyler:
Currently Zipkin is using a [thrift-based means of talking with Cassandra](https://github.com/openzipkin/zipkin/tree/master/zipkin-cassandra/src/main/scala/com/twitter/cassie). Support for this [Thrift interface](https://wiki.apache.org/cassandra/ThriftInterface) will be removed in future versions of Cassandra.

@michaelsembwever has mentioned the desire to introduce support for the datastax-driver into `zipkin-cassandra`. This ticket is to capture that work

abesto:
Can this be closed?

eirslett:
+1

Issue #579, opened 2015-08-11
Release!


abesto:
- [x] Publish an RC via Travis (by creating a tag)
- [x] Get approval for snapshot releases to oss.jfrog.org
- [x] Push snapshot releases via Travis on each successful build on master
- [x] Generate, upload docs to Bintray
- [x] Upload sources to Bintray
- [x] Get the release into jcenter
  - Verify release by building https://github.com/openzipkin/zipkin-mongodb/pull/1
- [x] Publish `SpanStoreSpec` in a jar with `test` classifier
- [x] Sync the release to Maven Central
- [x] Publish a final release (via Travis, by creating a tag)
- [x] Add the version badge to the README! :)
- [x] Migrate [docker-zipkin](https://github.com/openzipkin/docker-zipkin) to published artifacts
- [ ] Pin and release [zipkin-mongodb](https://github.com/openzipkin/zipkin-mongodb/issues/2)
- [ ] Pin and release [zipkin-hbase](https://github.com/openzipkin/zipkin-hbase/issues/2)
## Minor issues
- [x] The Travis run with `gradle release` and `gradle bintrayUpload` duplicate some work, we can probably save some of it
- [ ] Create, use dedicated user for release commits, or even better, a [deploy key](https://developer.github.com/guides/managing-deploy-keys/#deploy-keys)
- [x] All the packages other than `zipkin` need to be deleted from Bintray (we're uploading all artifacts into one package)
- [x] Don't build artifacts we won't upload (for example, `distZip`). See #635 for details on why this is not trivial.
## Tentative release notes
- Cassandra
  - Replaced internally cloned Cassie with the DataStax CQL driver
  - Change keyspace name from "Zipkin" to "zipkin"
  - Automatic schema creation on first start

codefromthecrypt:
@michaelsembwever added your note for release here. thx!

codefromthecrypt:
https://repo1.maven.org/maven2/io/zipkin/zipkin-query-service/1.2.1-rc24/

codefromthecrypt:
note there's still a glitch as automatic publishing doesn't work #644

codefromthecrypt:
this is pretty much done, right?

abesto:
zipkin-mongodb and zipkin-hbase aren't, and i'm guessing there are a couple of other spun-out projects that also need love? maybe those should be tracked in a separate issue though.

codefromthecrypt:
that reminds me...

[...]

PR #585, opened 2015-08-12
Replace cassie with datastax cql driver

codefromthecrypt:
- remove all cassie classes
- remove Storage/Index/Aggregate classes. Zipkin now only supports the SpanStorage trait.
- remove the hadoop aggregation classes. (this needs to be rewritten from Cassie to using CqlInputFormat).
- introduce a new java "Repository" class that provides the persistence layer using the DataStax CQL driver.
- the Repository class sits in its own zipkin-cassandra-core subproject, a pure java dependency free artifact.
- the schema, found in cassandra-schema-cql3.txt, is automatically created during startup.
- minimal but incompatible changes to the schema as keyspace and table names are now all lowercase.
- comments have been added to cassandra-schema-cql3.txt as to how the datamodel can and should be improved.

ref: https://github.com/openzipkin/zipkin/issues/458


codefromthecrypt:
AFAICT, the only thing left is to make sure tests pass. If they don't, I'll update the PR so that they do!

cc @eirslett @michaelsembwever

codefromthecrypt:
at least locally, this deflakes the test https://github.com/adriancole/zipkin/commit/ab2b69f8c4ce53d2eef91b168afe5a5b49f33abf

codefromthecrypt:
updated with https://github.com/adriancole/zipkin/commit/7cf1880dc70f91f872fae11fa6776fe5ff622fe3 to honor cassandra dest

eirslett:
Sweet!
A last, finishing touch; should we add a migration guide or changelog? Where the TLDR version is; wipe your old Cassandra keyspace and start from scratch?

michaelsembwever:
it's going to start from scratch anyway. any migration guide needs to
- inform that existing data will be "lost" (in a now unused keyspace),
- "zipkin.store.cassie" flags have changed to "zipkin.store.cassandra", and most of them removed,
- we only support Cassandra-2.0+ (i suspect we still support C*-1.2 but it'll be nice to not say that)

michaelsembwever:
i did a force push here to incorporate some small additions (i forgot to keep those commits separate like @adriancole did above, my bad)
- add CassandraSpanStoreFactory.createClusterBuilder() and added into the default implementation retryPolicy and loadBalancingPolicy,
- removed lots of unused or unneeded options,
- renamed all remaining flags to use the "zipkin.store.cassandra." prefix

codefromthecrypt:
ok looks like all left is docs. I'm doing a rebase to see if it can go green

michaelsembwever:
i've done a round with bin/collector|query|web|tracegen  and am happy with runtime.

michaelsembwever:
anything other than README.md and CHANGELOG you'd like me to update?

codefromthecrypt:
@michaelsembwever as far as I'm concerned :shipit:

michaelsembwever:
can't actually see anything in README.md that's relevant.
just CHANGELOG, and there are we talking about changes for "1.2.1" ?

codefromthecrypt:
@michaelsembwever yeah it is more about when we actually release 1.2.whatever. There's more to it than just cassandra.

codefromthecrypt:
need a step here #579 iotw

codefromthecrypt:
ok rebased for clean merge. just awaiting travis

michaelsembwever:
Ok, then i would have added to CHANGES the following 

```
# 1.2.1
* Cassandra
  * Replaced internally cloned Cassie with the DataStax CQL driver
  * Change keyspace name from "Zipkin" to "zipkin"
  * Automatic schema creation on first start
```

but i let you do that in master or in #579

codefromthecrypt:
FYI I'm stripping off the span store spec into a separate PR. The spec is entirely reliable on my machine, but something travis related is crashing it. Rather than thrash this PR, I'll spin it out and work on it independently.

PR #587, opened 2015-08-12
Removes Index and Storage from zipkin-cassandra

codefromthecrypt:
Cassandra is the last backend using Index and Storage instead of
SpanStore. Eventhough there's work in progress to replace Cassie with
an alternative Cassandra impl, removing this as an interim unlocks our
ability to remove legacy code.

See #483


codefromthecrypt:
tested via the following:

``` bash
./gradlew :zipkin-aggregate:run
# and in separate windows
bin/collector cassandra
bin/query cassandra
bin/web
bin/tracegen
open http://localhost:8080/
```

codefromthecrypt:
going to merge on green eventhough #585 will obviate it. Almost done the cleanup, and this makes that work easier.

PR #590, opened 2015-08-12
Removes unimplemented RealtimeAggregates and supporting code

codefromthecrypt:


codefromthecrypt:
couldn't find any implementation that was not a no-op or throws an exception, not even at twitter.

eirslett:
+1 sweet! Real time aggregates is still only a sweet dream, I'm afraid...

spencergibb:
:+1:

codefromthecrypt:
thanks for the look, folks

PR #592, opened 2015-08-13
Adds cassandra-unit powered CassandraSpanStoreSpec

codefromthecrypt:
follow-up from #585


codefromthecrypt:
I suspect the gradle daemon is getting killed on too much memory. Going to switch off gradle daemon on this PR to see if this is memory related or not.

codefromthecrypt:
this passes when not using the gradle daemon. I suspect this is memory related, as watching the test run in jconsole, my JVM spiked to 3.5GB memory.

codefromthecrypt:
@michaelsembwever I may be off in my assessment, but anyway, this "greens", but only if we disable the gradle daemon (which slows the build by 5m)

codefromthecrypt:
per chat in gitter, disabling the daemon during CI is acceptable. We're leaving the config in gradle.properties to help ease developer setup. IOTW, we can adjust the mem args needed without needing to notify all developers to update `~/.gradle/gradle.properties`

PR #593, opened 2015-08-13
Ports Builder-based services to SpanStore and deletes Index and Storage

### LINKED ISSUE #483: Lots of code duplication?

codefromthecrypt:
This completes code de-duplication by removing all references to Index
and Storage, after porting zipkin-collector-service and
zipkin-query-service to use SpanStore-based implementations.

closes #483


codefromthecrypt:
tested using `bin/collector`, `bin/query`, `bin/tracegen`, and `bin/web` on dev cassandra and redis

zipkin-query-service/src/main/scala/com/twitter/zipkin/query/Main.scala
```diff
@@ -35,11 +36,15 @@ object Main {
...
-      val server = builder.apply().apply(runtime)
-      server.start()
-      ServiceTracker.register(server)
+      val query = builder.apply().apply(runtime)
+      Await.ready(query)
+      ServiceTracker.register(new Service() {
+        override def shutdown() = Await.ready(query.close())
+
+        override def start() = {}
+      })
```
jamescway, line 47: I don't have a lot of experience with starting ostrich services, so I hope this is the accept way of doing it.  Perhaps a bit of shoehorning legacy?

jamescway:
Done have a ton of context, but everything lgtm except that one thing, which I'm assuming is necessary.
:+1:  nice deletes :100:

zipkin-query-service/src/main/scala/com/twitter/zipkin/query/Main.scala
codefromthecrypt, line 47: Yeah, it is shoe-horning. Migrating zipkin-collector-server and zipkin-query-server off ostrich is yet another todo :) Where to, is yet another question! finatra, finch etc.

PR #595, opened 2015-08-13
Removes the last of circa 2012 Cassandra

codefromthecrypt:
This cleans up both the build and the source tree of orphaned cassandra
code and dependencies.


michaelsembwever:
Nice! I had forgotten about all this stuff in my haste on #585. Thanks.

Issue #451, opened 2015-07-09
Express key interfaces as java interfaces, not scala traits


codefromthecrypt:
A number of folks have expressed a desire to have zipkin become more friendly to non-scala implementations. We want integrators to not leave zipkin solely because of scala or finagle.

For example, the brave project could implement interfaces defined in zipkin as opposed to similar ones as zipkin only exposes traits. More recently at Lyft, @jpinner needs to implement a Splunk base SpanStore in java.

By making integration points representable and implementable as dependency-free java types, we can slow the amount of divergence in the ecosystem, and head towards convergence. This issue will track the traits and types folks prefer to be java types, and address modeling, for example what to use instead of scala's Seq.

codefromthecrypt:
interested parties include at least @jpinner @dsyer @autoletics @kristofa @eirslett @xylus @michaelsembwever and myself. Many of these folks can write scala, but are working in interop scenarios that prefer java. (correct me if I'm wrong folks)

codefromthecrypt:
@postwait based on your [presentation last year](http://www.ustream.tv/recorded/61488592), I suspect you are also interested in this.

NiteshKant:
If this was the case, we would not have written our own custom dapper implementation inside Netflix, so I certainly support the intent here :)

autoletics:
The Probes Open API approach to API library design would be a good starting point...if not the reference itself. The original API was designed in 2008 and has remained the same except that when I moved from JXInsight (product) to Satoris (technology) I removed the com.jinspired.jxinsight.probes.\* from the namespace and replaced it with org.jinspired.probes.*.

http://autoletics.github.io/probes-api/

The library includes a single Java class for bootstrapping the underlying provider implementation, see SPI, and everything else is expressed as an inner interface. Other than the single class there is not dependency on an actual implementation.

The Probes Open API also shows how best to implementation extension points. Please look at Strategy, Interceptor,...

I am going to write an article that shows how easy it is to extend the metering runtime via the Interceptor Extension API to publish all method call spans to the existing Zipkin Collector. Hopefully this would serve as the inspiration though naturally I would prefer for everyone to instead adopt the Probes API as the new trace instrumentation and then to move on focusing on span collectors/stores/ui.

codefromthecrypt:
[...]

PR #598, opened 2015-08-13
Makes SpanStore and Aggregates implementable in Java

### LINKED ISSUE #584: Move to SpanStore normal java.io.Closeable
As mentioned in #451, some types are not instantiatable due to how they mix in Closable.

There's a pattern of use like this..

``` scala
  with com.twitter.util.Closable
  with com.twitter.util.CloseAwaitably
```

I suggest that we use `java.io.Closeable` instead. Eventhough it is a synchronous call, many of our implementations perform synchronous state cleanup. It is also extremely common practice for `close()` to be synchronous, including being built-into the java language (try-with-resources).

If we decide that asynchronous `.close` is required, we could experiment with how bad it is by adding a `src/test/java` tree and putting an example impl there. Even then, we may want to still support synchronous close, similar to how rxnetty has both `closeNow()` and `close()`.

codefromthecrypt:
This makes the critical storage interfaces, SpanStore and Aggregates,
abstract classes, and adds implementations in the java test tree to
prove they can compile with Java 7.

SpanStore was a mixture of traits, which made it impossible to write in
Java. Moreover, eventhough all implementations close synchronously, it
mixed in Twitter's Closeable, which was both a trait and also async.

Aggregates was a trait and implemented a synchronous close method
already.

See #451
Closes #584

### DIFF
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/storage/Aggregates.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/storage/Aggregates.scala
@@ -23,9 +23,7 @@ import com.twitter.zipkin.common.Dependencies
  * Storage and retrieval interface for aggregates that may be computed offline and reloaded into
  * online storage
  */
-trait Aggregates {
-
-  def close()
+abstract class Aggregates extends java.io.Closeable {
 
   def getDependencies(startDate: Option[Time], endDate: Option[Time]=None): Future[Dependencies]
   def storeDependencies(dependencies: Dependencies): Future[Unit]
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
@@ -17,58 +17,14 @@ package com.twitter.zipkin.storage
 
 import com.twitter.conversions.time._
 import com.twitter.finagle.{Filter => FFilter}
-import com.twitter.util.{Closable, CloseAwaitably, Duration, Future, Time}
+import com.twitter.util.FuturePools._
+import com.twitter.util.{Closable, Duration, Future}
 import com.twitter.zipkin.Constants
 import com.twitter.zipkin.common.Span
 import java.nio.ByteBuffer
 import scala.collection.mutable
 
-trait SpanStore extends WriteSpanStore with ReadSpanStore
-
-object SpanStore {
-  type Filter = FFilter[Seq[Span], Unit, Seq[Span], Unit]
-}
-
-
-/**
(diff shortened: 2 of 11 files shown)


zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
```diff
@@ -17,58 +17,14 @@ package com.twitter.zipkin.storage
...
-  type Filter = FFilter[Seq[Span], Unit, Seq[Span], Unit]
-}
-
-
-/**
- * A convenience builder to create a single WriteSpanStore from many. Writes
- * will be fanned out concurrently. A failure of any store will return a failure.
- * Any store logic should be handled per-store and then wrapped in this.
- */
-object FanoutWriteSpanStore {
```
codefromthecrypt, line 38: never used

zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
```diff
@@ -17,58 +17,14 @@ package com.twitter.zipkin.storage
...
-      Closable.all(stores: _*).close(deadline)
-    }
-  }
-}
-
-/**
- * Write store extends CloseAwaitably so we can close writes and await possible draining
- * of internal queues.
- */
-trait WriteSpanStore
```
codefromthecrypt, line 56: only used once

zipkin-redis-example/src/main/scala/com/twitter/zipkin/example/Main.scala
```diff
@@ -30,6 +30,6 @@ object Main extends TwitterServer
     closeOnExit(closer)
 
     println("running and ready")
-    Await.all(web, receiver, store)
```
codefromthecrypt, line 33: not necessary to await as the spanstore doesn't have an initialization routine.

codefromthecrypt:
@eirslett PTAL

cc @jpinner @NiteshKant

codefromthecrypt:
updated with the trivial change needed to make the other storage interface, Aggregates, java friendlier.

spencergibb:
:+1:

zipkin-common/src/main/scala/com/twitter/zipkin/storage/Aggregates.scala
```diff
@@ -23,9 +23,7 @@ import com.twitter.zipkin.common.Dependencies
  * Storage and retrieval interface for aggregates that may be computed offline and reloaded into
  * online storage
  */
-trait Aggregates {
-
-  def close()
+abstract class Aggregates extends java.io.Closeable {
```
codefromthecrypt, line 26: this one was easy.

Issue #597, opened 2015-08-13
Re-implement zipkin-aggregates Cassandra source


aloon:
when I try execute the agregate hadoop 

``` bash
hadoop jar zipkin-aggregate/build/libs/zipkin-aggregate-twitter-is-here-all.jar --source cassandra --hosts localhost --port 9160
```

return me:

``` bash
Exception in thread "main" java.io.IOException: Mkdirs failed to create /var/folders/75/b_p3cc112y52tvjdfg1bkz580000gp/T/hadoop-unjar2090021427468948969/META-INF/license
    at org.apache.hadoop.util.RunJar.ensureDirectory(RunJar.java:128)
    at org.apache.hadoop.util.RunJar.unJar(RunJar.java:104)
    at org.apache.hadoop.util.RunJar.unJar(RunJar.java:81)
    at org.apache.hadoop.util.RunJar.run(RunJar.java:209)
    at org.apache.hadoop.util.RunJar.main(RunJar.java:136)
```

I have read that it's problem of macos. I try to execute from debian but:

```
root@debian0:~# /root/hadoop-2.7.1/bin/hadoop jar /root/zipkin-aggregate-twitter-is-here-all.jar --source cassandra --hosts 172.16.144.1 --port 9160
Exception in thread "main" java.lang.Throwable: If you know what exactly caused this error, please consider contributing to GitHub via following link.
https://github.com/twitter/scalding/wiki/Common-Exceptions-and-possible-reasons#comtwitterscaldingargsexception
    at com.twitter.scalding.Tool$.main(Tool.scala:130)
    at com.twitter.zipkin.aggregate.RunHadoopJob$delayedInit$body.apply(RunHadoopJob.scala:4)
    at scala.Function0$class.apply$mcV$sp(Function0.scala:40)
    at scala.runtime.AbstractFunction0.apply$mcV$sp(AbstractFunction0.scala:12)
    at scala.App$$anonfun$main$1.apply(App.scala:71)
    at scala.App$$anonfun$main$1.apply(App.scala:71)
    at scala.collection.immutable.List.foreach(List.scala:318)
    at scala.collection.generic.TraversableForwarder$class.foreach(TraversableForwarder.scala:32)
    at scala.App$class.main(App.scala:71)
    at com.twitter.zipkin.aggregate.RunHadoopJob$.main(RunHadoopJob.scala:3)
    at com.twitter.zipkin.aggregate.RunHadoopJob.main(RunHadoopJob.scala)
[...]

PR #599, opened 2015-08-13
Resurrects Cassandra's DependencyStore

codefromthecrypt:
This resurrects Cassandra's DependencyStore, and notes the constraints
of the design. Most importantly, calls to `storeDependencies` will
overwrite values on a day granularity. This is different than anorm, for
example, which doesn't bucket by day.

See #597


codefromthecrypt:
rebased, but no other progress

eirslett:
Oh, sorry for the lack of response!
+1 LGTM!

codefromthecrypt:
To whom it may concern. I looked into resurrecting this. The problem is that there is very little test coverage to say what "should" happen. For example, the tests are trivial and rely mostly on whatever algebird is supposed to do. Best first step would be to backfill tests for the working implementation (anormdb and/or zipkin-hbase), then try to make a cassandra impl that passes those tests.

rtyler:
See also [Working Effectively With Legacy Code](http://objectmentor.com/resources/articles/WorkingEffectivelyWithLegacyCode.pdf) <_<

codefromthecrypt:
PTAL!

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraDependencyStore.scala
```diff
@@ -0,0 +1,39 @@
...
+import com.twitter.zipkin.common.Dependencies
+import com.twitter.zipkin.conversions.thrift._
+import com.twitter.zipkin.storage.DependencyStore
+import com.twitter.zipkin.thriftscala.{Dependencies => ThriftDependencies}
+import org.twitter.zipkin.storage.cassandra.Repository
+
+import scala.collection.JavaConverters._
+
+/**
+ * This implementation of DependencyStore assumes that the job aggregating dependencies
```
codefromthecrypt, line 14: this thing was the reason it took me so long to finish this. eventually, I figured out why everything was flooring to a day!

codefromthecrypt:
merging so I can release, which could help progress https://github.com/openzipkin/zipkin-dependencies-hadoop

PR #601, opened 2015-08-13
Release-related tweaks

abesto:
- In publish.sh, remove -b switch from git checkout master.
  It's slightly confusing that this didn't break previously,
  but there it is; the master branch should definitely already
  exist.
- In .travis.yml, override the default Travis behavior of
  ./gradlew assemble && ./gradlew check with
  running just publish.sh. The tasks run by publish.sh also run
  the tests, and we had duplicate execution between the Gradle runs
  started by Travis and publish.sh.
- If we don't want to release from this build, then just run
  ./gradlew check and exit. This can happen for four reasons:
  if this build was started by a PR, because we're not running
  on the JDK we want to use to build releases, because
  of travis-ci/travis-ci#1675,
  or because we're not on master.
- Fix: TRAVIS_PULL_REQUEST can be "false" or the PR number,
  but never "true".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,12 @@ before_install:
   - git config --global user.email "${GH_USER_EMAIL}"
   - git config --global user.name "${GH_USER}"
 
+# Skip `./gradlew assemble`
+install: /bin/true
+
 before_script:
   - git remote set-url origin "https://${GH_TOKEN}@github.com/openzipkin/zipkin.git"
 
-after_success: ./travis/publish.sh
+script:
+  - ./travis/publish.sh
+
--- a/travis/publish.sh
+++ b/travis/publish.sh
@@ -1,10 +1,6 @@
 #!/usr/bin/env bash
 set -e
 
-echo "********************************"
-echo "***** Publishing Artifacts *****"
-echo "********************************"
-
 declare -r PUBLISH_USING_JDK="oraclejdk7"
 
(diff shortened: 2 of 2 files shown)


codefromthecrypt:
these sound like miles forward. thanks!

jamescway:
Yeah funny -b didn't fail before.
:+1:  looks like the red build is from publishing not building.

abesto:
Thanks for checking! Just pushed another change to handle the last edge-cases I can think of, I think the builds should be green this time (won't even try to publish from building a branch / PR). If that happens, I'll go ahead and merge.

jamescway:
- Don't see the new travis.yml behavior?
- also I think the branch name set was not master (it was the tag name).  So in order to push the release notification commits to master, had to change the branch.

abesto:
Wrt/ new travis.yml behaviour: yeah, don't know what's up with that. Hope it'll go away if this goes into master, will do some research in the meanwhile.

Wrt/ checkout: yep, I've moved the git checkout to just before the call to `./gradlew release` to minimize side-effects and keep the "main" logic clean. AFAICT nothing else should need it; does that sound right?

abesto:
Re: `./gradlew assemble`: http://stackoverflow.com/questions/26566564/override-default-travis-ci-script-for-java-gradle-project :)

codefromthecrypt:
:shipit:

PR #602, opened 2015-08-13
Run checks before snapshot release, call Gradle just once

abesto:
According to https://docs.gradle.org/current/userguide/tutorial_gradle_command_line.html:
You can execute multiple tasks in a single build by listing each of the tasks on the command-line.

Plus, each task is run only once across the whole dependency graph of tasks, which can save some time.

I'm slightly concerned about the `-P` flags bleeding over into `bintrayUpload` and causing confusion, but didn't find evidence in https://github.com/bintray/gradle-bintray-plugin that it would cause issues.


codefromthecrypt:
I'm ok with this. :shipit:

PR #603, opened 2015-08-13
Removes legacy serverset registration

codefromthecrypt:
When Zipkin used to be deployed internally at Twitter, it registered
itself in ZooKeeper using Twitter's serverset library. Nowadays,
services are announced as a side effect of deployment to Aurora.

Removing serverset has no impact to those using docker, either, as
services are hard-wired directly to corresponding ports.

Finally, removing serverset decomplicates the dependency graph and
gets us closer to being able to not have zookeeper version conflicts.

### DIFF
--- a/zipkin-collector-service/src/main/scala/com/twitter/zipkin/collector/builder/CollectorServiceBuilder.scala
+++ b/zipkin-collector-service/src/main/scala/com/twitter/zipkin/collector/builder/CollectorServiceBuilder.scala
@@ -15,11 +15,9 @@
  */
 package com.twitter.zipkin.collector.builder
 
-import com.twitter.finagle.stats.StatsReceiver
 import com.twitter.finagle.Service
 import com.twitter.logging.Logger
-import com.twitter.ostrich.admin.{ServiceTracker, RuntimeEnvironment, Service => OstrichService}
-import com.twitter.util.Timer
+import com.twitter.ostrich.admin.{ServiceTracker, RuntimeEnvironment}
 import com.twitter.zipkin.builder.{ZipkinServerBuilder, Builder}
 import com.twitter.zipkin.collector.filter.{ServiceStatsFilter, SamplerFilter}
 import com.twitter.zipkin.collector.processor.{SpanStoreService, FanoutService}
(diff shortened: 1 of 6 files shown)


zipkin-query-service/src/main/scala/com/twitter/zipkin/builder/QueryServiceBuilder.scala
```diff
@@ -37,24 +32,10 @@ case class QueryServiceBuilder(
...
-
   def apply(): (RuntimeEnvironment) => ListeningServer = (runtime: RuntimeEnvironment) => {
-    val log = Logger.get()
     serverBuilder.apply().apply(runtime)
-
-    val address = new InetSocketAddress(serverBuilder.serverAddress, serverBuilder.serverPort)
     val store = storeBuilder.apply()
 
-    /* Register server sets */
-    serverSetPaths foreach { case (zkClientBuilder, path) =>
```
codefromthecrypt, line 50: we shouldn't take on random responsibilities such as registering arbitrary services.

zipkin-collector-service/src/main/scala/com/twitter/zipkin/collector/builder/CollectorServiceBuilder.scala
```diff
@@ -142,13 +129,6 @@ case class CollectorServiceBuilder[T](
       serverBuilder.adminHttpService map { _.addContext(path, new ConfigRequestHandler(adjustable)) }
     }
 
-    /* Start additional services (server sets) */
-    additionalServices foreach { builder =>
```
codefromthecrypt, line 146: we shouldn't take on random responsibilities such as registering arbitrary services.

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
```diff
@@ -13,7 +14,6 @@ import com.twitter.zipkin.conversions.thrift._
 import com.twitter.zipkin.query.{SpanTimestamp, TraceCombo, TraceSummary, QueryRequest}
 import com.twitter.zipkin.thriftscala.{Adjust, ZipkinQuery}
 import java.io.{File, FileInputStream, InputStream}
-import org.apache.commons.io.IOUtils
```
codefromthecrypt, line 16: was implicit in serverset's dep tree

spencergibb:
:+1:

eirslett:
+1

PR #604, opened 2015-08-13
Simplify the logic used to decide what kind of release to build, if any

abesto:
It turns out the logic for deciding what kind of release we want to do is a bit complicated, and easy to get wrong (so I got it wrong the first time). This PR aims to simplify its implementation to the point where it reads like prose and is trivial to see that it's correct.

Tested the various cases by adding an `echo` to before `$action` at the end. Result (without `is_pull_request`, which I forgot about and added later):

screenshot 2015-08-13 23 05 43


abesto:
CC @jamescway @adriancole @rtyler

codefromthecrypt:
thanks, @abesto.. let's give this a shot!

PR #606, opened 2015-08-13
Migrates ZooKeeper code to its only caller: AdaptiveSampler

codefromthecrypt:
This moves our bespoke ZooKeeper code to call site: AdaptiveSampler, so 
that it is contained. Later, we can refactor to use curator or similar. 

This also removes documentation about zookeeper, and replaces the 
outdated admin api for AdaptiveSampler with the one in use at Twitter


zipkin-sampler/src/test/scala/com/twitter/zipkin/sampler/RequestRateCheckTest.scala
```diff
@@ -0,0 +1,90 @@
...
+ * limitations under the License.
+ */
+package com.twitter.zipkin.sampler
+
+import com.twitter.conversions.time._
+import com.twitter.finagle.stats.NullStatsReceiver
+import com.twitter.util.{MockTimer, Time, Var}
+import org.scalatest.FunSuite
+
+class RequestRateCheckTest extends FunSuite {
```
codefromthecrypt, line 23: this file was incorrectly named AdaptiveSamplerTest before. It is just a rename, but git didn't understand.

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampler.scala
```diff
@@ -58,6 +85,24 @@ trait AdaptiveSampler { self: App with ZooKeeperClientFactory =>
     5.minutes,
     "Amount of time to see outliers before updating sample rate")
 
+  val zkServerLocations = flag(
```
codefromthecrypt, line 88: this is just folding in the ZK trait. I removed the trait as we don't want to encourage its use.

codefromthecrypt:
happy to address any feedback post-merge!

codefromthecrypt:
Note: this is now running in production at Twitter

PR #610, opened 2015-08-15
Graph my current best understanding of the release process

abesto:
Don't merge (yet), release process is not final.

It's not completely trivial :)

I suspect this will be helpful for whenever we (=any Zipkin contributor) needs to touch the release process. Or even just understand it.

Note the question marks in the bottom right; I'm confused about that, and I suspect that's where the last dragons are in the system.


eirslett:
### @abesto brings you:

# The Zipkin Release Process!

[IMAGE ALT TEXT](http://www.youtube.com/watch?v=AuUVBprLGKU"Zipkin release process")

Great work! I haven't been involved in any releasing (yet), the graph was very informative!

abesto:
Another graph will be upcoming with new release process.

PR #611, opened 2015-08-15
Remove some confusion from the release process

abesto:
This change tweaks the current release process to remove some confusion around how `gradle release` works and how it interacts with `gradle bintrayUpload` and the Travis runs started by commits created by `gradle release`.

Overview of the release process:

overview

High-level examples of RC and final releases:

examples

For example runs (with gutted-out builds building only `zipkin-common`) see the recent builds on https://travis-ci.org/abesto/zipkin/builds. The redness is due to still not being able to push SNAPSHOT releases.

**Note**: This WILL result in tests being run twice for a release (once before `gradle release` and once before `bintrayUpload`). As the saying goes: first make it work, then make it fast.


abesto:
+cc @rtyler @jamescway @adriancole

gradle.properties
```diff
@@ -1,5 +1,5 @@
 #Fri, 14 Aug 2015 17:14:51 +0000
-version=1.2.1-rc5
+version=1.2.1-SNAPSHOT
```
abesto, line 2: This just cleans up some mess left behind by the previous release.

travis/publish.sh
```diff
@@ -1,9 +1,10 @@
 #!/usr/bin/env bash
-set -e
+set -ex
```
abesto, line 2: The `x` is candidate for removal of course, but experience shows it'll be useful to have it there for the first few releases.

build.gradle
```diff
@@ -97,6 +95,11 @@ release {
     }
 }
 
+task createReleaseTag(overwrite: true, group: 'Release') {
```
codefromthecrypt, line 98: this sounds lovely

codefromthecrypt:
I like this flow, it is well documented and covers the cases, we've got so far.

PR #615, opened 2015-08-15
Don't upload SNAPSHOT releases to Bintray

abesto:
Also: pave the way for integrating with oss.jfrog.org for publishing SNAPSHOT releases.


codefromthecrypt:
Thanks for chipping the way towards snapshots.

PR #616, opened 2015-08-16
Fix pageTitle bug

gneokleo:
- The pageTitle was incorrectly set in the index handler. Moved that
  to the layout handler


codefromthecrypt:
Can you separate the bug fix from the enhancement?

The enhancement will be confusing to those who don't set this variable, which will be everyone except you. I'd personally prefer no environment in the UI than a false one.

gneokleo:
Sure. Doing that now.

codefromthecrypt:
gracias!

PR #618, opened 2015-08-16
Add environment/app name in the web ui

gneokleo:
- This commit adds an environment or app name in the top right corner of the the UI. This argument defaults to the empty string so there will be no visual change to previous versions if people don't want to set it


gneokleo:
cc @adriancole @eirslett
I will post a screenshot in the zipkin chat room

eirslett:
LGTM! +1

zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
```diff
@@ -55,6 +55,7 @@ trait ZipkinWebFactory { self: App =>
 
   val queryDest = flag("zipkin.web.query.dest", "127.0.0.1:9411", "Location of the query server")
   val queryLimit = flag("zipkin.web.query.limit", 10, "Default query limit for trace results")
+  val environment = flag("zipkin.web.environmentName", "", "The name of the environment Zipkin is running in")
```
codefromthecrypt, line 58: whoot

PR #619, opened 2015-08-16
Publish SNAPSHOT releases to oss.jfrog.com

abesto:
Looks like this: http://oss.jfrog.org/repo/io/zipkin/zipkin-common/


codefromthecrypt:
I can't wait to use this!

PR #620, opened 2015-08-16
Upgrade cassandra's cql schema to use correct types and meaningful names

michaelsembwever:
- column names are meaningful
- columns are of specific types
- remove unused columns
- use correct clustering order
- use DateTieredCompactionStrategy on tables that are time-series

because this is an incompatible change to the previous schema if there has been a release then the keyspace name should be bumped to "zipkin_v0". (but AFAICT there's been no such release, and developers will just need to know to drop any existing testing keyspaces they have).

original branch from https://github.com/thelastpickle/zipkin/tree/feature%2Fcassandra-transparent-persistence

### DIFF
--- a/zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
+++ b/zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
@@ -4,6 +4,7 @@ package org.twitter.zipkin.storage.cassandra;
 import com.datastax.driver.core.BatchStatement;
 import com.datastax.driver.core.BoundStatement;
 import com.datastax.driver.core.Cluster;
+import com.datastax.driver.core.KeyspaceMetadata;
 import com.datastax.driver.core.PreparedStatement;
 import com.datastax.driver.core.ResultSetFuture;
 import com.datastax.driver.core.Row;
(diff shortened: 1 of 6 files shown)


zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -32,10 +33,18 @@
 
 public final class Repository implements AutoCloseable {
 
-    static final String KEYSPACE = "zipkin";
-
-    private static final String KEY_MARKER = "k";
-    private static final String COLUMN_MARKER = "c";
+    public static final String KEYSPACE = "zipkin";
+
+    private static final String TRACE_ID = "trace_id";
```
codefromthecrypt, line 38: I've found most project avoid declaring constants just to snake case words in the same file. Seems the only below that isn't just a lower-snake is ts, but we use ts often in the codebase. By removing these, the file is less arduous to navigate.

codefromthecrypt:
Change looks good! I'd love to have a follow-up that drops the internal constants that just snake case things, but it isn't blocking.

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
michaelsembwever, line 38: follow up on this at https://github.com/openzipkin/zipkin/pull/622

PR #621, opened 2015-08-16
Supports CASSANDRA_CONTACT_POINTS comma-separated environment variable

codefromthecrypt:
Hopefully eliminates the need to make copies of the cassandra scala files in the docker project.

See https://github.com/openzipkin/docker-zipkin/pull/22

ex.

``` bash
$ CASSANDRA_CONTACT_POINTS=google.com,yahoo.com,localhost bin/query cassandra
```


kristofa:
Nice! :+1:

PR #622, opened 2015-08-16
replace constants that represent the same string literal

michaelsembwever:
ref: https://github.com/openzipkin/zipkin/pull/620#discussion_r37146775


codefromthecrypt:
weird.. thought I merged this!

PR #624, opened 2015-08-17
Sync non-SNAPSHOT releases to Maven Central

abesto:


jamescway:
:+1:

codefromthecrypt:
Whoah.. Is this for real?! :)

abesto:
Real, absolutely. We don't yet know if it'll work, but it's most definitely real!

PR #625, opened 2015-08-17
*really don't create the zipkin-cassandra/target directory when running cassandra-unit

michaelsembwever:
follow up to 

| \* ec70 - 2015-08-15 21:32:46 -0700
| |  Adrian Cole  Moves cassandra's test temp directory to gradle's

to *really avoid putting files into the target/ directory we needed the cu-cassandra.yaml file existing and configured so all paths were to
build/ instead of target/

cu-cassandra.yaml is also tuned for lower memory usage.

### DIFF
--- a/zipkin-cassandra/src/test/resources/cu-cassandra.yaml
+++ b/zipkin-cassandra/src/test/resources/cu-cassandra.yaml
@@ -1,0 +1,586 @@
+# Cassandra storage config YAML
+
+# NOTE:
+#   See http://wiki.apache.org/cassandra/StorageConfiguration for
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
Sad we need to check in a big text file to accomplish this :(

:shipit: on green

michaelsembwever:
is there a way to re-run travis builds?

codefromthecrypt:
Yep. You should have a login with github option at the top when you click
into Travis link. Once logged into travis, there should be a circular arrow
you can click to kick a job.

michaelsembwever:
well that was obvious :-)

Issue #623, opened 2015-08-17
New Cassandra schema may create hot spots


yurishkuro:
Hi, I just pulled the recent code, and while it's great to see an upgrade from thrift-based schema in Cassandra to proper CQL, wouldn't the new schema create hot spots by not using the previous bucketing technique?  For example, the new schema for indexing spans by service name

```
CREATE TABLE zipkin.service_name_index (
    service_name text,
    ts timestamp,
    trace_id bigint,
    PRIMARY KEY (service_name, ts)
```

The `service_name` here serves as the row key, so for a heavily loaded service ALL spans will be hitting a single partition. The previous implementation was adding a random bucket number 1-10 to the service name, so that the inserts were uniformly distributed across 10 nodes, at the cost of having to do 10 queries when reading by service name. The previous bucketing would be roughly equivalent to this new schema:

```
CREATE TABLE zipkin.service_name_index (
    service_name text,
    bucket tinyint,
    ts timestamp,
    trace_id bigint,
    PRIMARY KEY ((service_name, bucket), ts)
```

yurishkuro:
Another observation - before the upgrade the Zipkin code had a -D switch to control the number of buckets, such that setting that param to 1 bucket would've been equivalent to the current/new schema. If we extend the schema as suggested above, people will have an option to use or not use the bucketing, with little overhead if they don't want to use it.

michaelsembwever:
Indeed there will be hot spots on the `service_name` in this table.
That there was bucketing happening before I definitely missed.

michaelsembwever:
bucketCount and "zipkin.store.cassie.indexBuckets", as you mention defaulted to ten previously.

yurishkuro:
it's pretty much in all tables, e.g. annotations_index for standard annotations like cs/cr/ss/sr will be always hitting row key `"${serviceName}:${annotation}"`. The old code was using BucketedColumnFamily for most of the tables.

michaelsembwever:
yes. thanks for spotting this. i'll have a fix out in the next hour or so.

yurishkuro:
also, if we go for a dedicated bucket column included in the composite row key, I would suggest changing the schema for annotations_index. Right now it is stored as a composite string `${serviceName}:${annotation} [:$annotationValue]`. I would suggest the following schema instead:

```
CREATE TABLE IF NOT EXISTS zipkin.annotations_index (
    service_name      string,
    annotation_name   string,
[...]

PR #626, opened 2015-08-17
New Cassandra schema may create hot spots #623

michaelsembwever:
I've done this branch off feature/remove-words-as-is-constants as it's rather trivial and i'm expecting it merged soon enough (and it would be conflicts otherwise).

The work here does
- restore bucketing approach
- number of buckets are hardcoded to 10
- buckets exist as a separate column in the partition key to maintain transparency
  ref: https://github.com/openzipkin/zipkin/issues/623

### DIFF
--- a/zipkin-cassandra-core/src/main/resources/cassandra-schema-cql3.txt
+++ b/zipkin-cassandra-core/src/main/resources/cassandra-schema-cql3.txt
@@ -17,23 +17,26 @@ CREATE TABLE IF NOT EXISTS zipkin.top_annotations (
 
 CREATE TABLE IF NOT EXISTS zipkin.service_name_index (
     service_name      text,
+    bucket            int,
     ts                timestamp,
     trace_id          bigint,
-    PRIMARY KEY (service_name, ts)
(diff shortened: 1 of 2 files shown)


zipkin-cassandra-core/src/main/resources/cassandra-schema-cql3.txt
```diff
@@ -17,23 +17,26 @@ CREATE TABLE IF NOT EXISTS zipkin.top_annotations (
 
 CREATE TABLE IF NOT EXISTS zipkin.service_name_index (
```
yurishkuro, line 18: looks like the schema change for service_name table is missing

michaelsembwever, line 18: service_name intentionally does not use buckets.

michaelsembwever:
the `service_span_name_index` table doesn't need buckets i believe (and didn't previously).

the combination of service and span is enough to avoid hot partitions.

you are right about splitting service and span columns. i will tackle that in a subsequent PR.

spencergibb:
:+1:

PR #627, opened 2015-08-18
Fix CassandraSpanStoreSpec.setTimeToLive randomly failing

michaelsembwever:
CassandraSpanStoreSpec.setTimeToLive was randomly failing
 because it has no way to know when the write for the CassandraSpanStore.setTimeToLive(..) was complete
 more accurately Repository.storeSpan(..), which setTimeToLive(..) calls, executes the write asynchronously but does not return any future
that the consumer can wait upon.

This could have been better implemented with properly proxing java futures onto scala/twitter futures, but the initial attempts fell into the
too-hard basket, so all storeXX(..) methods in Repository.java just fire-and-forget.

The hack solution here was to override SpanStoreSpec.setTimeToLive in CassandraSpanStoreSpec so to keep checking for the correct result and
only after 5 seconds give up and presume the test has really failed.

### DIFF
--- a/zipkin-cassandra/src/test/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStoreSpec.scala
+++ b/zipkin-cassandra/src/test/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStoreSpec.scala
@@ -1,8 +1,12 @@
 package com.twitter.zipkin.storage.cassandra
 
 import com.datastax.driver.core.Cluster
+import com.twitter.conversions.time.intToTimeableNumber
+import com.twitter.util.Await.{ready, result}
+import com.twitter.util.Duration
 import com.twitter.zipkin.storage.SpanStoreSpec
 import java.util.Collections
+import junit.framework.Test
 import org.cassandraunit.CQLDataLoader
 import org.cassandraunit.dataset.CQLDataSet
 import org.cassandraunit.utils.EmbeddedCassandraServerHelper.startEmbeddedCassandra
@@ -36,4 +40,19 @@ class CassandraSpanStoreSpec extends SpanStoreSpec {
   override lazy val store = new CassandraSpanStore(new Repository(keyspace, cluster))
 
   override def clear = cluster.connect().execute("DROP KEYSPACE IF EXISTS " + keyspace)
+
+  override def setTimeToLive() {
+    ready(store(Seq(span1)))
+    ready(store.setTimeToLive(span1.traceId, 1234.seconds))
+
+    for( i <- 1 to 100) {
+      // Repository.storeXXX() methods write asynchronously but don't return a Future so we can't reliably test.
+      //  just wait and loop instead
+      java.lang.Thread.sleep(50)
+      if ((result(store.getTimeToLive(span1.traceId)) - 1234.seconds).abs.inMilliseconds <= (i*50) + 100) {
+        return
+      }
+    }
+    throw new AssertionError
+  }
 }


zipkin-cassandra/src/test/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStoreSpec.scala
```diff
@@ -36,4 +40,23 @@ class CassandraSpanStoreSpec extends SpanStoreSpec {
...
+
+  override def setTimeToLive() {
+    ready(store(Seq(span1)))
+    ready(store.setTimeToLive(span1.traceId, 1234.seconds))
+
+    for( i <- 1 to 100) {
+      // @xxx Repository.storeXXX() methods write asynchronously but don't return a Future so we can't reliably test.
+      //  just wait and loop instead
+      java.lang.Thread.sleep(50)
+      // If a store doesn't use TTLs this should return Duration.Top
```
abesto: But we know whether the Cassandra store uses TTLs, no? So we could just have the specialized code here?
(also in the assert at the bottom)

zipkin-cassandra/src/test/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStoreSpec.scala
```diff
@@ -36,4 +40,23 @@ class CassandraSpanStoreSpec extends SpanStoreSpec {
   override lazy val store = new CassandraSpanStore(new Repository(keyspace, cluster))
 
   override def clear = cluster.connect().execute("DROP KEYSPACE IF EXISTS " + keyspace)
+
+  override def setTimeToLive() {
+    ready(store(Seq(span1)))
+    ready(store.setTimeToLive(span1.traceId, 1234.seconds))
+
+    for( i <- 1 to 100) {
+      // @xxx Repository.storeXXX() methods write asynchronously but don't return a Future so we can't reliably test.
```
abesto: That sounds pretty bad; I'd expect it can lead to all kinds of inconsistencies because the Cassandra store is not telling the truth ("truth" as defined by the API of `SpanStore`). I don't know how to formulate this properly, but maybe it's worth opening an issue to track this?

abesto:
Two notes, neither one blocking. Other than those, LGTM :+1: 

Very excited, once this is merged, we might have a release on Maven Central :)

codefromthecrypt:
In jclouds we had stores that acted differently wrt consistency. We created
a method "assertEventually" that invoked an assertion with exponential
backoff (ex invoked both the retrieve and compare) where this was permitted
to be different.

Where it wasn't permitted to be different, the stores themselves blocked
before they returned success. This allowed all stores to have the same
behavior regardless of impl. Otherwise, as @abesto said, people will be
very surprised to find they can't read their writes.

I'd suggest changing this code to subclass CassandraSpanStore upon
initialization. Override setttl with a method that calls super then has the
blocking loop before returning. This allows the test to stay in tact (ex
not subject to copy/paste drift).

Add to the README that this op is eventually consistent.

Wdyt?

zipkin-cassandra/src/test/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStoreSpec.scala
michaelsembwever: yeah in this test we know that the test is writing a TTL and that cassandra honours that.

it's a good catch and i will fix.

zipkin-cassandra/src/test/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStoreSpec.scala
michaelsembwever: It's not as bad as it sounds because it only is a problem in the context of tests.

The cassandra backend is intentionally eventually consistent here. Writes are written with only a consistency level of one, and so are reads. So if you had increased the replication of the cluster for the zipkin keyspace (something you definitely should do in production) then it's quite likely that there's a few microseconds after a write that if you were to read from a different node you wouldn't get the latest value.

And I can't see any reason in the zipkin datamodel that such truth, or strict consistency is ever required. Therefore the fire-and-forget of the writes is a triviality.

Except for the tests. But it's solved here in this pull request.

zipkin-cassandra/src/test/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStoreSpec.scala
michaelsembwever: fix with this commit https://github.com/thelastpickle/zipkin/commit/26991109926b2c070b111277e173ecd4a7d7b235

which has been squashed into the pull request.

zipkin-cassandra/src/test/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStoreSpec.scala
codefromthecrypt: Easiest way our would be to add doc to base span store, say it is
eventually consistent, and change the base test. I think many ops will end
up eventually consistent anyways. Regardless, I'm fine with this.

michaelsembwever:
> Otherwise, as @abesto said, people will be
> very surprised to find they can't read their writes.

How would people know?
Arn't the writes we're talking about here coming from an event system, and the system at large is already asynchronous?
I think trying to make the store strictly consistent is a bit of a pointless exercise here.

> I'd suggest changing this code to subclass CassandraSpanStore upon
> initialization. Override setttl with a method that calls super then has the
> blocking loop before returning. This allows the test to stay in tact (ex
> not subject to copy/paste drift).

My scala remains too young to accomplish that.
Could you provide an example?

zipkin-cassandra/src/test/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStoreSpec.scala
```diff
@@ -36,4 +40,19 @@ class CassandraSpanStoreSpec extends SpanStoreSpec {
...
+
+  override def setTimeToLive() {
+    ready(store(Seq(span1)))
+    ready(store.setTimeToLive(span1.traceId, 1234.seconds))
+
+    for( i <- 1 to 100) {
+      // Repository.storeXXX() methods write asynchronously but don't return a Future so we can't reliably test.
+      //  just wait and loop instead
+      java.lang.Thread.sleep(50)
+      if ((result(store.getTimeToLive(span1.traceId)) - 1234.seconds).abs.inMilliseconds <= 10) {
```
michaelsembwever: obvious bug there, given we've just slept for 50ms. let me fix.

michaelsembwever:
>  people will be very surprised to find they can't read their writes.

maybe you mean by "people" the zipkin developers rather the zipkin users?
if so, is there any reason to not presume that all stores are eventually consistent, since that would be a) the design of the system we're working with at large, and b) beneficial to performance as it will prevent read-before-write anti-patterns creeping into the code.

i don't know… it's a more involved discussion than that in regards to the user.
and yes it would have been nice if the Repository.storeSpan(..) returned a java future that was wrapped into a scala/twitter future.

zipkin-cassandra/src/test/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStoreSpec.scala
michaelsembwever: fixed with https://github.com/thelastpickle/zipkin/commit/b518cbbf69dfb97f68328b9e05d5388b66f7f057

which has been squashed into this PR

codefromthecrypt:
> How would people know?
> Arn't the writes we're talking about here coming from an event system,
> and the system at large is already asynchronous?
> A developer of SpanStore should know what expectations are, especially as
> this is the only time we've hit this.

Maybe just document the method as 'Returns once the backend accepts the ttl
request'. Or document the whole SpanStore type as 'all commands that return
void are processed asynchronously'

> I think trying to make the store strictly consistent is a bit of a
> pointless exercise here.
> Not about that, just saying it is now (else all stores would flake). We
> should document that it isn't.
> 
> My scala remains too young to accomplish that.
> Could you provide an example?
> Will do, but sadly need an hour to be at a laptop :(

codefromthecrypt:
PS I volunteer to address all of this post merge!

abesto:
He said merge. Did you see him say merge? :)

codefromthecrypt:
He clicked merge.. Did you see him click merge? :)

I will fix the root issue later, which is that the SpanStore api needs to be documented and tested assuming writes are async.

PR #629, opened 2015-08-18
Don't build, test during the Travis job doing gradle release

abesto:
## What?

Before this PR, the Travis job that does a build + check, which is re-done by the Travis job that does the upload. This PR changes that so that `gradle release` ONLY creates new commits and nothing else; the Travis job doing the upload also doesn't run tests.
## Why?

Two reasons:
- tests fail in mysterious ways during the build started by `gradle release` on Travis, may be related to OOM, but near impossible to debug
- With the old setup, we run tests four times for a release (the PR job of the last merged PR, the test run on master when the PR is merged, release and upload). With the new setup, that's just two, and we reduce the number of test runs inside the release process from two to zero.
## How?
- Remove `check` from the release and upload part of `publish.sh`
- Override the Gradle task `runBuildTasks` to do nothing, similarly to how `createReleaseTag` was already overridden

This change also upgrades the release plugin, because https://github.com/researchgate/gradle-release/issues/120 is nice to have.

### DIFF
--- a/travis/README.md
+++ b/travis/README.md
@@ -1,6 +1,14 @@
 # Zipkin release process
 
-The Zipkin release process is controlled by [publish.sh](publish.sh). Releases are initiated by pushing a tag `major.minor.revision` or `major.minor.revision-qualifier`, where `qualifier` is usually something like `rc3`. A high-level overview of how the release happens in each case:
+The Zipkin release process is controlled by [publish.sh](publish.sh).
+Releases are initiated by pushing a tag `major.minor.revision` or `major.minor.revision-qualifier`,
+where `qualifier` is usually something like `rc3`.
+
+There is currently no automated testing during the release process. As the person pushing a release tag, you're
+expected to manually check that the test [run on Travis](https://travis-ci.org/openzipkin/zipkin) is green for
+the commit you're tagging. (Building tooling for this would be awesome.)
+
+A high-level overview of how the release happens for RC and final releases:
 
 ![Release example](examples.png)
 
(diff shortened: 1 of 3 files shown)


abesto:
These overrides feel like they'd be better as configurations options inside the release plugin. I may (or may not) get around to creating a PR for that to the plugin itself. As it is, we are on the border of abusing internals of the plugin instead of using the stable API.

codefromthecrypt:
:shipit:

completely on board with not running tests as a part of release process.

travis/README.md
```diff
@@ -1,6 +1,14 @@
 # Zipkin release process
 
-The Zipkin release process is controlled by [publish.sh](publish.sh). Releases are initiated by pushing a tag `major.minor.revision` or `major.minor.revision-qualifier`, where `qualifier` is usually something like `rc3`. A high-level overview of how the release happens in each case:
+The Zipkin release process is controlled by [publish.sh](publish.sh).
+Releases are initiated by pushing a tag `major.minor.revision` or `major.minor.revision-qualifier`,
+where `qualifier` is usually something like `rc3`.
+
+There is currently no automated testing during the release process. As the person pushing a release tag, you're
```
codefromthecrypt, line 7: nit add link https://travis-ci.org/openzipkin/zipkin
nit s/version you're tagging/commit you're tagging/

codefromthecrypt:
LGTM with nits

jamescway:
LGTM :)

travis/README.md
abesto, line 7: Nits addressed (on different line, so the comment stays here for eternity)

PR #630, opened 2015-08-18
Publish to the right oss.jfrog repo on non-SNAPSHOT builds

abesto:


build.gradle
```diff
@@ -13,6 +13,8 @@ plugins {
     id 'net.researchgate.release' version '2.2.1'
 }
 
+ext.isSnapshot = VersionNumber.parse(rootProject.version).qualifier == "SNAPSHOT"
```
codefromthecrypt, line 16: I like how you've extracted this. In case someone knows a better way, they can replace it easily. FWIW this looks good to me.

codefromthecrypt:
:shipit:

PR #635, opened 2015-08-19
Create, use custom configuration for collecting artifacts to upload

abesto:
This lets us control exactly what gets uploaded to Bintray / OJO / Sonatype.
## Context

Disabling the `distTar` or `distZip` tasks confuses the hell out of later tasks that want to use the output of these tasks: just disabling them doesn’t remove their would-be output from the list of artifacts to upload (in gradle-speak: from the configuration ‘archives’). I couldn't find a clean, easy solution for this on the internets. Looking at the source of the bintray plugin (specifically https://github.com/bintray/gradle-bintray-plugin/blob/master/src/main/groovy/com/jfrog/bintray/gradle/BintrayPlugin.groovy#L74 and the fact that it’s applied only if we use configurations)

We could hack further and manually try to remove the right artifacts from the right configurations, but apparently these are internals of the distribute plugin, so that may break in the future. Manually declaring the artifacts to be published would also give us more control and make the intentions explicit. The "old" way is using configurations, which is used in this PR for ease of implementation. The new way would be publications, which would require some more re-working.

For the reasons at the top, this solution builds everything and then uploads a subset. Room for further optimization (adding to #579): let's not build stuff we don't plan to upload.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,11 +14,6 @@ plugins {
 }
 
 ext.isSnapshot = VersionNumber.parse(rootProject.version).qualifier == "SNAPSHOT"
-ext.wantDistTar = [
-        project(':zipkin-collector-service'),
-        project(':zipkin-query-service'),
-        project(':zipkin-web')
-]
 
 allprojects { thisProject ->
     apply plugin: 'idea'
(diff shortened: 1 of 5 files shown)


codefromthecrypt:
Wow. Much simpler! :shipit:

Does this still build the zip or tarball that isnt published? If so, eventually would be nice to not spend time making them. If no, fantastic!

abesto:
Short answer: yes I think, we still build all the zips and tarballs (though I'm not 100% confident, we'll need to see). Added some context to the description.

PR #637, opened 2015-08-19
Disable syncing to Maven Central for now

abesto:
Enabling syncing creates a situation where the plugin needs to wait until publishing of the version finishes on Bintray, but that takes longer than the request timeout, so we get failed builds that look like https://travis-ci.org/openzipkin/zipkin/jobs/76344895

At the same time, re-enable uploading of "big" artifacts (tars, fat jars)


codefromthecrypt:
+1

PR #639, opened 2015-08-20
add authentication to cassandra

jamescway:
Adding authentication to cassandra.  Would have liked to do this inside of Repository.java but I think the entry point is at a higher level in the scala code.  So there are 3 entry points to passing user credentials.  If the user uses the config files (collector-cassandra.scala) they need to add environment variables as written in the submodule readme.

### DIFF
--- a/zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
+++ b/zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
@@ -25,19 +25,20 @@ import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
 import com.twitter.zipkin.storage.cassandra._
 import org.twitter.zipkin.storage.cassandra.Repository
 import org.twitter.zipkin.storage.cassandra.ZipkinRetryPolicy
+import com.twitter.app.Flag
 
 trait CassandraSpanStoreFactory {self: App =>
 
   import com.twitter.zipkin.storage.cassandra.{CassandraSpanStoreDefaults => Defaults}
 
+  val keyspace              = flag("zipkin.store.cassandra.keyspace", Defaults.KeyspaceName, "name of the keyspace to use")
+  val cassandraDest         = flag("zipkin.store.cassandra.dest", "localhost:9042", "dest of the cassandra cluster; comma-separated list of host:port pairs")
+  val cassandraSpanTtl      = flag("zipkin.store.cassandra.spanTTL", Defaults.SpanTtl, "length of time cassandra should store spans")
+  val cassandraIndexTtl     = flag("zipkin.store.cassandra.indexTTL", Defaults.IndexTtl, "length of time cassandra should store span indexes")
+  val cassandraMaxTraceCols = flag("zipkin.store.cassandra.maxTraceCols", Defaults.MaxTraceCols, "max number of spans to return from a query")
+  val cassandraUser: Flag[String]     = flag("zipkin.store.cassandra.user", "cassandra authentication user name")
+  val cassandraPassword: Flag[String] = flag("zipkin.store.cassandra.password", "cassandra authentication password")
 
-  val keyspace = flag("zipkin.store.cassandra.keyspace", Defaults.KeyspaceName, "name of the keyspace to use")
-  val cassandraDest = flag("zipkin.store.cassandra.dest", "localhost:9042", "dest of the cassandra cluster; comma-separated list of host:port pairs")
-
-  val cassieSpanTtl = flag("zipkin.store.cassandra.spanTTL", Defaults.SpanTtl, "length of time cassandra should store spans")
-  val cassieIndexTtl = flag("zipkin.store.cassandra.indexTTL", Defaults.IndexTtl, "length of time cassandra should store span indexes")
-
-  val cassieMaxTraceCols = flag("zipkin.store.cassandra.maxTraceCols", Defaults.MaxTraceCols, "max number of spans to return from a query")
 
   def newCassandraStore(stats: StatsReceiver = DefaultStatsReceiver.scope("CassandraSpanStore")): CassandraSpanStore = {
     val repository = new Repository(keyspace(), createClusterBuilder().build())
@@ -45,14 +46,17 @@ trait CassandraSpanStoreFactory {self: App =>
     new CassandraSpanStore(
       repository,
       stats.scope(keyspace()),
-      cassieSpanTtl(),
-      cassieIndexTtl(),
-      cassieMaxTraceCols())
+      cassandraSpanTtl(),
+      cassandraIndexTtl(),
+      cassandraMaxTraceCols())
   }
 
   def createClusterBuilder(): Cluster.Builder = {
-    addContactPoint(Cluster.builder())
-      .withRetryPolicy(ZipkinRetryPolicy.INSTANCE)
+    val builder = addContactPoint(Cluster.builder())
+    if(cassandraUser.isDefined && cassandraPassword.isDefined)
+      builder.withCredentials(cassandraUser(), cassandraPassword())
+
+    builder.withRetryPolicy(ZipkinRetryPolicy.INSTANCE)
       .withLoadBalancingPolicy(new TokenAwarePolicy(new LatencyAwarePolicy.Builder(new RoundRobinPolicy()).build()))
   }
 
--- a/zipkin-cassandra/src/test/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactorySpec.scala
+++ b/zipkin-cassandra/src/test/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactorySpec.scala
@@ -1,10 +1,12 @@
 package com.twitter.zipkin.cassandra
 
-import com.datastax.driver.core.Cluster
+import com.datastax.driver.core.{AuthProvider, Cluster}
 import com.twitter.app.App
 import java.net.InetSocketAddress
 import java.util.Arrays.asList
 import org.scalatest.{FunSuite, Matchers}
+import com.datastax.driver.core.exceptions.AuthenticationException
+
 
 class CassandraSpanStoreFactorySpec extends FunSuite with Matchers {
   object TestFactory extends App with CassandraSpanStoreFactory
@@ -45,5 +47,29 @@ class CassandraSpanStoreFactorySpec extends FunSuite with Matchers {
     TestFactory.addContactPoint(Cluster.builder()).getContactPoints should be(
       asList(new InetSocketAddress("1.1.1.1", 9143), new InetSocketAddress("2.2.2.2", 9042))
     )
+  }
+
+  test("creatingClusterBuilder with SASL authentication null delimited utf8 bytes") {
+    TestFactory.nonExitingMain(Array(
+      "-zipkin.store.cassandra.user", "user",
+      "-zipkin.store.cassandra.password", "pass"
+    ))
+    val SASLhandshake = Array[Byte](0, 'u', 's', 'e', 'r', 0, 'p', 'a', 's', 's')
+    val authProvider = TestFactory.createClusterBuilder()
+                         .getConfiguration()
+                         .getProtocolOptions()
+                         .getAuthProvider()
+                         .newAuthenticator(new InetSocketAddress("localhost", 8080))
+                         .initialResponse()
+    authProvider should be(SASLhandshake)
+  }
+
+  test("creatingClusterBuilder without authentication") {
+    TestFactory.nonExitingMain(Array())
+
+    TestFactory.createClusterBuilder()
+      .getConfiguration()
+      .getProtocolOptions()
+      .getAuthProvider() should be (AuthProvider.NONE)
   }
 }
--- a/zipkin-collector-service/config/collector-cassandra.scala
+++ b/zipkin-collector-service/config/collector-cassandra.scala
@@ -16,20 +16,32 @@
 
 import com.datastax.driver.core.Cluster
 import com.datastax.driver.core.SocketOptions
+import com.twitter.app.App
 import com.twitter.zipkin.builder.Scribe
 import com.twitter.zipkin.cassandra
+import com.twitter.zipkin.cassandra.CassandraSpanStoreFactory
 import com.twitter.zipkin.collector.builder.CollectorServiceBuilder
 import com.twitter.zipkin.storage.Store
 import org.twitter.zipkin.storage.cassandra.ZipkinRetryPolicy
+import scala.collection.mutable.ListBuffer
 
 val contactPoints: Array[String] = sys.env.get("CASSANDRA_CONTACT_POINTS").getOrElse("localhost")
   .split(",")
 
-val cluster = Cluster.builder()
-  .addContactPoints(contactPoints:_*)
-  .withSocketOptions(new SocketOptions().setConnectTimeoutMillis(10000).setReadTimeoutMillis(20000))
-  .withRetryPolicy(ZipkinRetryPolicy.INSTANCE)
-  .build()
+val cassandraUser = sys.env.get("CASSANDRA_USER")
+val cassandraPass = sys.env.get("CASSANDRA_PASS")
+
+var args = new ListBuffer[String]()
+if (cassandraUser.isDefined && cassandraPass.isDefined) {
+  args += "-zipkin.store.cassandra.user"
+  args += cassandraUser.get
+  args += "-zipkin.store.cassandra.password"
+  args += cassandraPass.get
+}
+
+object CollectorService extends App with CassandraSpanStoreFactory
+CollectorService.nonExitingMain(args.toArray)
+val cluster = CollectorService.createClusterBuilder().build()
 
 val storeBuilder = Store.Builder(new cassandra.SpanStoreBuilder(cluster))
 
--- a/zipkin-query-service/config/query-cassandra.scala
+++ b/zipkin-query-service/config/query-cassandra.scala
@@ -15,19 +15,30 @@
  */
 import com.datastax.driver.core.Cluster
 import com.datastax.driver.core.SocketOptions
+import com.twitter.app.App
 import com.twitter.zipkin.builder.QueryServiceBuilder
 import com.twitter.zipkin.cassandra
+import com.twitter.zipkin.cassandra.CassandraSpanStoreFactory
 import com.twitter.zipkin.storage.Store
-import org.twitter.zipkin.storage.cassandra.ZipkinRetryPolicy
+import scala.collection.mutable.ListBuffer
 
 val contactPoints: Array[String] = sys.env.get("CASSANDRA_CONTACT_POINTS").getOrElse("localhost")
   .split(",")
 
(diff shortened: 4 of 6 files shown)


zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
```diff
@@ -30,14 +30,14 @@ trait CassandraSpanStoreFactory {self: App =>
...
-
-  val cassieSpanTtl = flag("zipkin.store.cassandra.spanTTL", Defaults.SpanTtl, "length of time cassandra should store spans")
-  val cassieIndexTtl = flag("zipkin.store.cassandra.indexTTL", Defaults.IndexTtl, "length of time cassandra should store span indexes")
-
+  val keyspace           = flag("zipkin.store.cassandra.keyspace", Defaults.KeyspaceName, "name of the keyspace to use")
+  val cassandraDest      = flag("zipkin.store.cassandra.dest", "localhost:9042", "dest of the cassandra cluster; comma-separated list of host:port pairs")
+  val cassieSpanTtl      = flag("zipkin.store.cassandra.spanTTL", Defaults.SpanTtl, "length of time cassandra should store spans")
+  val cassieIndexTtl     = flag("zipkin.store.cassandra.indexTTL", Defaults.IndexTtl, "length of time cassandra should store span indexes")
   val cassieMaxTraceCols = flag("zipkin.store.cassandra.maxTraceCols", Defaults.MaxTraceCols, "max number of spans to return from a query")
+  val cassandraAuth      = flag("zipkin.store.cassandra.auth", "false", "Set to true if using cassandra authentication")
```
codefromthecrypt: do we need this flag? could we instead use existence of user to accomplish the same?

zipkin-cassandra/src/test/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactorySpec.scala
```diff
@@ -46,4 +46,38 @@ class CassandraSpanStoreFactorySpec extends FunSuite with Matchers {
...
+  }
+
+  test("creatingClusterBuilder without authentication") {
+    TestFactory.nonExitingMain(Array(
+      "-zipkin.store.cassandra.auth", "false"
+    ))
+
+    val builder = TestFactory.createClusterBuilder()
+    val authProvider     = get_privates(builder, "authProvider")
+    val user = get_privates(authProvider, "username")
```
codefromthecrypt: rename to getField? also no need to declare vals.. `getField(authProvider, "username") should be(empty)`

p.s. I'd be surprised if there was no reflection-based matcher, but I'm often surprised.

codefromthecrypt: I dug through the code a bit.. 

should be able to do this instead. `TestFactory.createClusterBuilder().getConfiguration(). getAuthProvider()`

cast as `PlainTextAuthenticator` then `getCredentials()` should match.

That's more refactor save and doesn't use reflection. If doesn't work, lemme know!

zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
jamescway: I started with that at first, but I figure explicitness over inference, eitherway WFM

zipkin-cassandra/src/test/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactorySpec.scala
jamescway: I guess the method name wasn't too funny _womp womp_, thanks for finding that :smile:

jamescway: @adriancole actually I couldn't find it, am I doing something different?
screen shot 2015-08-19 at 7 50 21 pm

jamescway: nevermind getConfiguration.getProtocolOptions().getAuthProvider()

zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
codefromthecrypt: it isn't typical, so I'd remove it. saying I want to set a username but not authenticate doesn't make sense.

zipkin-cassandra/src/test/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactorySpec.scala
codefromthecrypt: yep and your base case test can simply set no flags and check the type of this.

zipkin-query-service/config/query-cassandra.scala
```diff
@@ -22,9 +22,16 @@ import org.twitter.zipkin.storage.cassandra.ZipkinRetryPolicy
 
 val contactPoints: Array[String] = sys.env.get("CASSANDRA_CONTACT_POINTS").getOrElse("localhost")
   .split(",")
+val cassandraAuth: String = sys.env.get("CASSANDRA_AUTH").getOrElse("")
```
codefromthecrypt: I wouldn't do the getOrElse things unless "cassandra/cassandra" is a builtin-user/password? Even then, I'd have people set it.

iotw use `sys.env.get("CASSANDRA_USER").isSet` or whatever to determine if the flags are set as we aren't using default values for this.

zipkin-cassandra/src/test/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactorySpec.scala
```diff
@@ -46,4 +48,32 @@ class CassandraSpanStoreFactorySpec extends FunSuite with Matchers {
       asList(new InetSocketAddress("1.1.1.1", 9143), new InetSocketAddress("2.2.2.2", 9042))
     )
   }
+
+  test("creatingClusterBuilder with SASL authentication null delimited utf8 bytes") {
+    TestFactory.nonExitingMain(Array(
+      "-zipkin.store.cassandra.user", "user",
+      "-zipkin.store.cassandra.password", "pass"
+    ))
+    val SASLhandshake = Array[Byte](0, 117, 115, 101, 114, 0, 112, 97, 115, 115)
```
codefromthecrypt: nit.. use the character form of 'u', 's', ... that way you don't have to explain what 117 is

zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
```diff
@@ -30,14 +30,13 @@ trait CassandraSpanStoreFactory {self: App =>
...
-  val keyspace = flag("zipkin.store.cassandra.keyspace", Defaults.KeyspaceName, "name of the keyspace to use")
-  val cassandraDest = flag("zipkin.store.cassandra.dest", "localhost:9042", "dest of the cassandra cluster; comma-separated list of host:port pairs")
-
-  val cassieSpanTtl = flag("zipkin.store.cassandra.spanTTL", Defaults.SpanTtl, "length of time cassandra should store spans")
-  val cassieIndexTtl = flag("zipkin.store.cassandra.indexTTL", Defaults.IndexTtl, "length of time cassandra should store span indexes")
-
+  val keyspace           = flag("zipkin.store.cassandra.keyspace", Defaults.KeyspaceName, "name of the keyspace to use")
+  val cassandraDest      = flag("zipkin.store.cassandra.dest", "localhost:9042", "dest of the cassandra cluster; comma-separated list of host:port pairs")
+  val cassieSpanTtl      = flag("zipkin.store.cassandra.spanTTL", Defaults.SpanTtl, "length of time cassandra should store spans")
+  val cassieIndexTtl     = flag("zipkin.store.cassandra.indexTTL", Defaults.IndexTtl, "length of time cassandra should store span indexes")
```
michaelsembwever: why we re-introducing these?

jamescway: justed updated the spacing

zipkin-collector-service/config/collector-cassandra.scala
```diff
@@ -25,11 +25,18 @@ import org.twitter.zipkin.storage.cassandra.ZipkinRetryPolicy
...
+
+val clusterBuilder = if(cassandraUser == "" && cassandraPass == "")
+  Cluster.builder()
+else
+  Cluster.builder().withCredentials(cassandraUser, cassandraPass)
+
+val cluster = clusterBuilder.addContactPoints(contactPoints:_*)
+    .withSocketOptions(new SocketOptions().setConnectTimeoutMillis(10000).setReadTimeoutMillis(20000))
+    .withRetryPolicy(ZipkinRetryPolicy.INSTANCE)
+    .build()
```
michaelsembwever: any reason we're not using the factory here instead?

michaelsembwever:
LGTM   +1

zipkin-collector-service/config/collector-cassandra.scala
jamescway: Not that I can see, I think you implemented the SpanStoreBuilder to be backwards compatible.  Will roll it forward.

jamescway: It looks like if we use the factory we would use it [here](https://github.com/openzipkin/zipkin/blob/master/zipkin-query-service/src/main/scala/com/twitter/zipkin/builder/QueryServiceBuilder.scala#L39)?  Maybe I could do that in another PR.

 or were you thinking something like:

``` scala
object TestFactory extends App with CassandraSpanStoreFactory
cluster = TestFactory.createClusterBuilder()
```

Not sure if this is valid ^^^ or if there's a more direct way to do it.

zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
michaelsembwever: yeah. i read the diff wrong. sorry about that.
these variables should get renamed from cassieXX to cassandraXX someday.

zipkin-collector-service/config/collector-cassandra.scala
michaelsembwever: yes like that, and override createClusterBuilder() and inside the overridden method you can add credentials from the system properties.

that way you don't loose the customisations we make to the default Cluster setup (which can have important behavioural differences).

ultimately though it would make more sense IMO to pass the system variables on to the flags. but i don't know how the flags really work so well…

zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
codefromthecrypt: we should make this today :) @jamescway mind?

jamescway: @adriancole yeah sorry forgot to mention, already done when he said it :smile:

zipkin-collector-service/config/collector-cassandra.scala
```diff
@@ -25,11 +25,18 @@ import org.twitter.zipkin.storage.cassandra.ZipkinRetryPolicy
 val contactPoints: Array[String] = sys.env.get("CASSANDRA_CONTACT_POINTS").getOrElse("localhost")
   .split(",")
 
-val cluster = Cluster.builder()
-  .addContactPoints(contactPoints:_*)
-  .withSocketOptions(new SocketOptions().setConnectTimeoutMillis(10000).setReadTimeoutMillis(20000))
-  .withRetryPolicy(ZipkinRetryPolicy.INSTANCE)
-  .build()
+val cassandraUser: String = sys.env.get("CASSANDRA_USER").getOrElse("")
```
codefromthecrypt: I still think this is an awkward use of optional. instead of getOrElse(""), you could leave that off and check isPresent or whatever below.

zipkin-collector-service/config/collector-cassandra.scala
codefromthecrypt: I'd just set flags the same we do in the unit tests. Forgive the quasi-scala.

```
args = mutableList()

val user: String = sys.env.get("CASSANDRA_USER")
val pass: String = sys.env.get("CASSANDRA_PASS")
if (user.isSet && pass.isSet) {
args.add("-zipkin.store.cassandra.user").add(user.get)
args.add("-zipkin.store.cassandra.password").add(password.get)
}

// similar for other args.

TestFactory.nonExitingMain(args.toArray)
val cluster = TestFactory.createClusterBuilder().build()
```

jamescway:
@michaelsembwever @adriancole tested locally, updated to factory, fixed up getorelse stmts.  Thx for the review!

zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
```diff
@@ -30,14 +30,13 @@ trait CassandraSpanStoreFactory {self: App =>
...
-  val cassieIndexTtl = flag("zipkin.store.cassandra.indexTTL", Defaults.IndexTtl, "length of time cassandra should store span indexes")
-
-  val cassieMaxTraceCols = flag("zipkin.store.cassandra.maxTraceCols", Defaults.MaxTraceCols, "max number of spans to return from a query")
+  val keyspace              = flag("zipkin.store.cassandra.keyspace", Defaults.KeyspaceName, "name of the keyspace to use")
+  val cassandraDest         = flag("zipkin.store.cassandra.dest", "localhost:9042", "dest of the cassandra cluster; comma-separated list of host:port pairs")
+  val cassandraSpanTtl      = flag("zipkin.store.cassandra.spanTTL", Defaults.SpanTtl, "length of time cassandra should store spans")
+  val cassandraIndexTtl     = flag("zipkin.store.cassandra.indexTTL", Defaults.IndexTtl, "length of time cassandra should store span indexes")
+  val cassandraMaxTraceCols = flag("zipkin.store.cassandra.maxTraceCols", Defaults.MaxTraceCols, "max number of spans to return from a query")
+  val cassandraUser         = flag("zipkin.store.cassandra.user", "", "cassandra authentication user name")
```
codefromthecrypt: I think you want to take out the default of ""? iirc this param should just be left out to make optional.

zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
```diff
@@ -45,14 +44,18 @@ trait CassandraSpanStoreFactory {self: App =>
...
-      cassieMaxTraceCols())
+      cassandraSpanTtl(),
+      cassandraIndexTtl(),
+      cassandraMaxTraceCols())
   }
 
   def createClusterBuilder(): Cluster.Builder = {
-    addContactPoint(Cluster.builder())
-      .withRetryPolicy(ZipkinRetryPolicy.INSTANCE)
+    val builder = if(!cassandraUser.isDefined || !cassandraPassword.isDefined)
```
codefromthecrypt: I think this would be clearer to first initalize, then guard entry to withCredentials.

ex. 

```
val builder = addContactPoint(Cluster.builder())
if(cassandraUser.isDefined && cassandraPassword.isDefined) {
  builder.withCredentials(cassandraUser(), cassandraPassword())
}
```

codefromthecrypt: reworded logic to read better, too.

zipkin-cassandra/src/test/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactorySpec.scala
```diff
@@ -46,4 +48,32 @@ class CassandraSpanStoreFactorySpec extends FunSuite with Matchers {
...
+                         .getAuthProvider()
+                         .newAuthenticator(new InetSocketAddress("localhost", 8080))
+                         .initialResponse()
+    authProvider should be(SASLhandshake)
+  }
+
+  test("creatingClusterBuilder without authentication") {
+    TestFactory.nonExitingMain(Array())
+
+    intercept[AuthenticationException] {
```
codefromthecrypt: I like where you're going, but I think it would be clearer to just check it is AuthProvider.NONE?

```
TestFactory.createClusterBuilder()
        .getConfiguration()
        .getProtocolOptions()
        .getAuthProvider() must be (AuthProvider.NONE)
```

jamescway:
@adriancole gtg

zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
```diff
@@ -25,19 +25,20 @@ import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
...
 
   import com.twitter.zipkin.storage.cassandra.{CassandraSpanStoreDefaults => Defaults}
 
+  val keyspace              = flag("zipkin.store.cassandra.keyspace", Defaults.KeyspaceName, "name of the keyspace to use")
+  val cassandraDest         = flag("zipkin.store.cassandra.dest", "localhost:9042", "dest of the cassandra cluster; comma-separated list of host:port pairs")
+  val cassandraSpanTtl      = flag("zipkin.store.cassandra.spanTTL", Defaults.SpanTtl, "length of time cassandra should store spans")
+  val cassandraIndexTtl     = flag("zipkin.store.cassandra.indexTTL", Defaults.IndexTtl, "length of time cassandra should store span indexes")
+  val cassandraMaxTraceCols = flag("zipkin.store.cassandra.maxTraceCols", Defaults.MaxTraceCols, "max number of spans to return from a query")
+  val cassandraUser: Flag[String]     = flag("zipkin.store.cassandra.user", "cassandra authentication user name")
+  val cassandraPassword: Flag[String] = flag("zipkin.store.cassandra.password", "cassandra authentication password")
```
michaelsembwever, line 40: nice!

michaelsembwever:
all :+1:

codefromthecrypt:
fantastico

zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
jamescway, line 40: hard to find documentation indeed, ended up grepping finagle repo

PR #643, opened 2015-08-20
Removes unused canned scripts and tar.gz publication

codefromthecrypt:
We shouldn't be storing things not used, or complicating publication
with duplicate artifacts.

The scripts stored in zipkin's repo are not used at Twitter, not used by
the dist routine, not used by our bin directory, and not used in docker.

At the moment, the tar.gz dist isn't used by anything either. When our
artifact publication works, there's nothing in the tar.gz we can do that
we can't do with the fat jar. Once we have versioned fat jars, we should
change docker to use them.. we don't need a tar.gz IOTW.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ build/
 # https://github.com/openzipkin/zipkin/issues/534
 zipkin-collector-service/zipkin.db
 /.nb-gradle/
+
+# https://github.com/openzipkin/zipkin/pull/642
+travis_wait*.log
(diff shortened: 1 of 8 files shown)


codefromthecrypt:
cc @rtyler @eirslett @kristofa @jamescway

jamescway:
:+1:

kristofa:
Having the versioned and published fat jars will be great!
:+1:

eirslett:
+1

codefromthecrypt:
Thanks for the quick look, folks!

PR #646, opened 2015-08-21
Uses JavaExec directly to avoid spending time on dists

codefromthecrypt:
A few more lines, but hopefully a lot less time and i/o. Merge after https://github.com/openzipkin/docker-zipkin/pull/24


codefromthecrypt:
ok this should be safe to merge, since docker only uses fat jars now.

spencergibb:
:+1:

abesto:
Nice!

I guess it'd be possible to move the common parts into a separate `.gradle` file, but it's questionable whether that would make this easier or harder to read.

Great stuff!

PR #647, opened 2015-08-21
Explicitly declares jdk and vm flag matrixes

codefromthecrypt:
This is an attempt to give more metaspace to jdk 8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,12 @@ sudo: false
 services:
   - redis-server
 
-jdk:
-  - oraclejdk7
-  - oraclejdk8
+matrix:
+  include:
+    - jdk: "oraclejdk7"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,9 +4,12 @@ sudo: false
...
 
-jdk:
-  - oraclejdk7
-  - oraclejdk8
+matrix:
+  include:
+    - jdk: "oraclejdk7"
+      env: GRADLE_OPTS="-Dorg.gradle.daemon=false -XX:+CMSClassUnloadingEnabled -XX:+UseConcMarkSweepGC -Xms2G -Xmx2G -Xss512k -XX:MaxPermSize=256m"
+    - jdk: "oraclejdk8"
+      env: GRADLE_OPTS="-Dorg.gradle.daemon=false -Xms2G -Xmx2G -Xss512k -XX:MetaspaceSize=256m"
```
jamescway, line 12: i'd be curious what happens at this size

jamescway:
:+1:

.travis.yml
codefromthecrypt, line 12: well cool thing is that we can now toy with it! Ex. we can put in another row for actually running the gradle daemon, or only skipping it on a jdk it actually breaks on. I'm soooo excited :P

Issue #644, opened 2015-08-21
hang at bintrayUpload


codefromthecrypt:
so at the binTray upload step, the build times out like this.

```
:zipkin-aggregate:bintrayUpload
:zipkin-anormdb:bintrayUpload
:zipkin-cassandra:bintrayUpload
:zipkin-cassandra-core:bintrayUpload
:zipkin-collector:bintrayUpload
:zipkin-collector-scribe:bintrayUpload
:zipkin-collector-service:bintrayUpload
:zipkin-common:bintrayUpload
:zipkin-example:bintrayUpload
:zipkin-query:bintrayUpload
:zipkin-query-service:bintrayUpload
:zipkin-receiver-kafka:bintrayUpload
:zipkin-receiver-scribe:bintrayUpload
:zipkin-redis:bintrayUpload
:zipkin-redis-example:bintrayUpload
:zipkin-sampler:bintrayUpload
:zipkin-scrooge:bintrayUpload
:zipkin-thrift:bintrayUpload
:zipkin-tracegen:bintrayUpload
:zipkin-web:bintrayUpload
:bintrayUpload
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated
```

https://travis-ci.org/openzipkin/zipkin/jobs/76552907

When you browse to https://bintray.com/openzipkin/zipkin/zipkin/view, you file it waiting for you to click publish. Once you do.. you can actually successfully click sync to maven https://bintray.com/openzipkin/zipkin/zipkin/1.2.1-rc24/view#central

So, what's up with the hang at bintrayUpload?

codefromthecrypt:
similar status

```
Found 1 configuration(s) to publish.
Uploading to https://api.bintray.com/content/openzipkin/zipkin/zipkin/1.2.1-rc25/io/zipkin/zipkin-web/1.2.1-rc25/zipkin-web-1.2.1-rc25.jar...
Uploaded to 'https://api.bintray.com/content/openzipkin/zipkin/zipkin/1.2.1-rc25/io/zipkin/zipkin-web/1.2.1-rc25/zipkin-web-1.2.1-rc25.jar'.
Uploading to https://api.bintray.com/content/openzipkin/zipkin/zipkin/1.2.1-rc25/io/zipkin/zipkin-web/1.2.1-rc25/zipkin-web-1.2.1-rc25-javadoc.jar...
[...]

PR #648, opened 2015-08-21
Make Gradle more verbose during release and upload

abesto:
Travis currently times out the build in the upload phase (see #644). This can have two reasons:
- Gradle is stuck
- Gradle is working, but not producing any output

In either case, this change takes us closer to the truth:
- If Gradle is stuck, then this will tell us where exactly it's stuck
- If Gradle is working, then this should produce output to keep Travis happy

### DIFF
--- a/travis/publish.sh
+++ b/travis/publish.sh
@@ -73,7 +73,7 @@ function want_to_release_from_this_jdk(){
 
 function publish(){
   echo "[Publishing] Publishing..."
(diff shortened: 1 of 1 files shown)


jamescway:
:+1:

codefromthecrypt:
yes.. the more output likely will keep travis happy.. let's watch! https://travis-ci.org/openzipkin/zipkin/jobs/76685357

PR #650, opened 2015-08-21
Allow excluding of non-core projects on publishing

jamescway:
### Problem

Jar files are too big to publish, need to exclude some submodules
### Solution

Create property which controls excluding certain submodules from publishing.
If you pass the gradle property `-Pinclude_noncore_projects=false` to the gradle script it excludes non-core subprojects.  If you leave the property empty it defaults to including all subprojects.
This is allows, releasing smaller jar files.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,61 +1,70 @@
 rootProject.name = 'zipkin'
 
+def coreProjects = new ArrayList()
+def noncoreProjects = new ArrayList()
+
 ////////////////////////////////////
 // Zipkin's mostly common library //
 ////////////////////////////////////
 // Not all projects depend on this!
-include 'zipkin-common',
+coreProjects << 'zipkin-common'
 
 ///////////////////////////////////////////////////////////////////////////////////////////
 // Projects with outputs that are useful on their own: documentation, examples, services //
 ///////////////////////////////////////////////////////////////////////////////////////////
 
 //'doc'
-'zipkin-example',
-'zipkin-redis-example',
-'zipkin-collector-service',
-'zipkin-query-service',
-'zipkin-web',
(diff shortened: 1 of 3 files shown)


settings.gradle
```diff
@@ -4,59 +4,68 @@ rootProject.name = 'zipkin'
...
-'zipkin-redis-example',
-'zipkin-collector-service',
-'zipkin-query-service',
-'zipkin-web',
-'zipkin-aggregate',
+if (include_noncore_projects == 'true') {
+  println("[settings.gradle] Including non-core project: zipkin-example")
+  include 'zipkin-example'
+}
+include 'zipkin-redis-example'
```
codefromthecrypt: this is also very non-core :)

jamescway:
@abesto I remember you removing all the `include` stmts, in favor of `include a,b,c...` 
Hopefully this doesn't cause any problems

settings.gradle
```diff
@@ -4,59 +4,68 @@ rootProject.name = 'zipkin'
 // Zipkin's mostly common library //
 ////////////////////////////////////
 // Not all projects depend on this!
-include 'zipkin-common',
```
codefromthecrypt, line 7: there was some weird thing once upon a time that forced us to use commas instead of multiple lines of include. Maybe it was intellij?

Probably what we want to do is make a list, and then condtionally add to the end of it. After that, "include" that list.

jamescway, line 7: yeah ok, thats what I was asking Zoltan

abesto, line 7: That's correct, having different include statements caused problems for `gradle idea`. I think if we move just some modules to the bottom that are not dependencies of other projects, it should be fine.

PR #651, opened 2015-08-21
Excludes examples and tracegen to artifact repositories

codefromthecrypt:


jamescway:
:+1:

codefromthecrypt:
one passed, so merging.

This is another bad weather day for jcenter

PR #652, opened 2015-08-22
Propagates Cassandra environment variables as flags

codefromthecrypt:
This fixes docker. Note I also am using `CASSANDRA_PASSWORD` not `CASSANDRA_PASS` as the former is more consistent with the existing properties, as password isn't commonly abbreviated.

### DIFF
--- a/zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
+++ b/zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
@@ -16,60 +16,47 @@
 package com.twitter.zipkin.cassandra
 
 import com.datastax.driver.core.Cluster
-import com.datastax.driver.core.policies.LatencyAwarePolicy;
-import com.datastax.driver.core.policies.RoundRobinPolicy;
-import com.datastax.driver.core.policies.TokenAwarePolicy;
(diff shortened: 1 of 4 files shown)


zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
```diff
@@ -16,60 +16,47 @@
...
 import org.twitter.zipkin.storage.cassandra.ZipkinRetryPolicy
-import com.twitter.app.Flag
 
-trait CassandraSpanStoreFactory {self: App =>
-
-  import com.twitter.zipkin.storage.cassandra.{CassandraSpanStoreDefaults => Defaults}
+import scala.collection.JavaConversions
 
-  val keyspace              = flag("zipkin.store.cassandra.keyspace", Defaults.KeyspaceName, "name of the keyspace to use")
-  val cassandraDest         = flag("zipkin.store.cassandra.dest", "localhost:9042", "dest of the cassandra cluster; comma-separated list of host:port pairs")
```
adriancole: oh wait.. I forgot this is a trait. I'll un-rename this back to cassandraFoo

adriancole: done

jamescway:
:+1:

PR #654, opened 2015-08-22
Removes duplicated examples

codefromthecrypt:
The two example projects combine zipkin components into a single JVM.
This is not a realistic use of zipkin, so not a good example actually. 

The README shows how to start the query, collector, and web via the bin
scripts. It also shows how to make trace data and open the UI. We're
better off with a single quick-start story.


jamescway:
got to go :100:

PR #663, opened 2015-08-26
zipkin-collector-service is better configurable

kristofa:
Allow more configuration options to be configurable through
environment variables.

Especially being able to easily configure log level, sample rate
and number of queue workers is crucial for us.

### DIFF
--- a/zipkin-collector-service/README.md
+++ b/zipkin-collector-service/README.md
@@ -11,15 +11,26 @@ store, most typically SQL, Cassandra or Redis.
 ./gradlew :zipkin-collector-service:run
 ```
 
-#### Start with Cassandra Authentication
+#### Configuration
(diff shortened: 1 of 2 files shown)


kristofa:
In our current zipkin setup we tweak `log level` (default debug level logs way too much), `queue num workers` and `sample rate`.  By using environment variables it means we also can configure these options when using `docker-zipkin`.

kristofa:
If the environment variables are not specified I use the current default values.

jamescway:
:+1:, maybe could use some simple docs?

kristofa:
@jamescway Agree. I updated `README.md` with explanation of the new configurable parameters.

jamescway:
looks great :)

PR #664, opened 2015-08-27
reduce duplicates writes in cassandra to service_names and span_names tables

michaelsembwever:
problem:
 storing spans in CassandraSpanStore involves writes to service_names and span_names tables.
 these tables are nothing more than lists of service names and span names active within the index ttl period (default of 3 days)
 there is both at the application level and in the database a large amount of wasted overhead with duplicate writes having to do this for
every span.

solution:
 keep a simple thread-local cache in the write path so within any given hour writes to either table that have already happened are avoided.
 this removes the need for the span_names.bucket column but it's kept for compatibilty and a value of zero always now used in it.

### DIFF
--- a/zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
+++ b/zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
@@ -40,11 +40,15 @@ public final class Repository implements AutoCloseable {
 
     private static final Logger LOG = LoggerFactory.getLogger(Repository.class);
     private static final Random RAND = new Random();
+
     private static final List<Integer> ALL_BUCKETS = Collections.unmodifiableList(new ArrayList<Integer>() {{
         for (int i = 0 ; i < BUCKETS ; ++i) {
             add(i);
         }
     }});
+
+    private static final long WRITTEN_NAMES_TTL
+            = Long.getLong("zipkin.store.cassandra.internal.writtenNamesTtl", 60 * 60 * 1000);
 
     private final Session session;
     private final PreparedStatement selectTraces;
@@ -70,6 +74,26 @@ public final class Repository implements AutoCloseable {
     private final PreparedStatement selectTraceDurationTail;
     private final Map<String,String> metadata;
 
+    private final ThreadLocal<Set<String>> writtenNames = new ThreadLocal<Set<String>>() {
+            private long cacheInterval = toCacheInterval(System.currentTimeMillis());
+
+            @Override
+            protected Set<String> initialValue() {
+                return new HashSet<String>();
+            }
+            @Override
+            public Set<String> get() {
+                long newCacheInterval = toCacheInterval(System.currentTimeMillis());
+                if (cacheInterval != newCacheInterval) {
+                    cacheInterval = newCacheInterval;
+                    set(new HashSet<String>());
+                }
+                return super.get();
+            }
+            private long toCacheInterval(long ms) {
+                return ms / WRITTEN_NAMES_TTL;
+            }
+        };
 
     public Repository(String keyspace, Cluster cluster) {
         metadata = Schema.ensureExists(keyspace, cluster);
@@ -139,7 +163,7 @@ public final class Repository implements AutoCloseable {
                 QueryBuilder.select("span_name")
                     .from("span_names")
                     .where(QueryBuilder.eq("service_name", QueryBuilder.bindMarker("service_name")))
-                    .and(QueryBuilder.in("bucket", QueryBuilder.bindMarker("bucket"))));
+                    .and(QueryBuilder.eq("bucket", QueryBuilder.bindMarker("bucket"))));
 
         insertSpanName = session.prepare(
                 QueryBuilder
@@ -467,18 +491,21 @@ public final class Repository implements AutoCloseable {
     public void storeServiceName(String serviceName, int ttl) {
         Preconditions.checkNotNull(serviceName);
         Preconditions.checkArgument(!serviceName.isEmpty());
-        try {
-            BoundStatement bound = insertServiceName.bind()
-                    .setString("service_name", serviceName)
-                    .setInt("ttl_", ttl);
+        if (writtenNames.get().add(serviceName)) {
+            try {
+                BoundStatement bound = insertServiceName.bind()
+                        .setString("service_name", serviceName)
+                        .setInt("ttl_", ttl);
 
-            if (LOG.isDebugEnabled()) {
-                LOG.debug(debugInsertServiceName(serviceName, ttl));
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug(debugInsertServiceName(serviceName, ttl));
+                }
+                session.executeAsync(bound);
+            } catch (RuntimeException ex) {
+                LOG.error("failed " + debugInsertServiceName(serviceName, ttl), ex);
+                writtenNames.get().remove(serviceName);
+                throw ex;
             }
-            session.executeAsync(bound);
-        } catch (RuntimeException ex) {
-            LOG.error("failed " + debugInsertServiceName(serviceName, ttl), ex);
-            throw ex;
         }
     }
 
@@ -496,7 +523,7 @@ public final class Repository implements AutoCloseable {
 
                 BoundStatement bound = selectSpanNames.bind()
                         .setString("service_name", serviceName)
-                        .setList("bucket", ALL_BUCKETS);
+                        .setInt("bucket", 0);
 
                 if (LOG.isDebugEnabled()) {
                     LOG.debug(debugSelectSpanNames(serviceName));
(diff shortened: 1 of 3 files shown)


michaelsembwever:
fyi @jamescway @yurishkuro

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -70,6 +71,24 @@
     private final PreparedStatement selectTraceDurationTail;
     private final Map<String,String> metadata;
 
+    private final ThreadLocal<Set<String>> writtenNames = new ThreadLocal<Set<String>>() {
+                private final AtomicLong hour
```
eirslett: I guess you don't need thread-safe constructs inside a ThreadLocal? Wouldn't a Long be enough here? Or maybe this field is shared across threads?

michaelsembwever: yeah good catch. (i had the field outside to begin with…)

michaelsembwever: fixed with https://github.com/thelastpickle/zipkin/commit/dd3ba49169298eabc9a90ed342de0080ba1de114
(i'll squash that commit into this PR's one commit)

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -496,7 +516,7 @@ private String debugInsertServiceName(String serviceName, int ttl) {
 
                 BoundStatement bound = selectSpanNames.bind()
                         .setString("service_name", serviceName)
-                        .setList("bucket", ALL_BUCKETS);
+                        .setInt("bucket", 0);
```
yurishkuro, line 526: this removes bucketing functionality for span names - why?

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -70,6 +70,23 @@
     private final PreparedStatement selectTraceDurationTail;
     private final Map<String,String> metadata;
 
+    private final ThreadLocal<Set<String>> writtenNames = new ThreadLocal<Set<String>>() {
+                private long hour = TimeUnit.MILLISECONDS.toHours(System.currentTimeMillis());
```
yurishkuro: a couple of problems with this cache. First, why one hour? Different installations might use different TTL in Cassandra, in most cases at least 1 day, so cache TTL of 1hr is too fine. I have similar logic in Go collector, we use 12hr TTL for cache. In this OSS implementation I'd suggest making this TTL externally configurable.

Second problem - the TTL is on the whole cache, not on individual entries. It means once an hour _all threads_ in _all collectors_ will drop the cache and re-insert unnecessary data. 

Granted, both are not huge issues given the ratio of writes, but the implementation would not be much more difficult if both were addressed.

Also, thread locals are unnecessary here, a concurrent map shared between threads would do.

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
michaelsembwever, line 526: it won't be (afaik) needed.

no service_span combination will be written more than a handful of times each hour (for each collector jvm running).

from a write per span stored to just a few per hour i do think removes the need for bucketing.

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
michaelsembwever: > Different installations might use different TTL in Cassandra, in most cases at least 1 day, so cache TTL of 1hr is too fine.

We're not needing to match the index ttl value here.
For example if the cache ttl matched the index ttl, ie 3 days, then after 3 days all service and service-span names would disappear, despite there having been plenty of spans stored over the past 3 days with these values.

Describing it the other way around, with a cache ttl of one hour it means that the service and service-names will disappear after 2 days and 23 hours rather than the index ttl of 3 days.

So… the cache ttl needs to be some small proportion of the index ttl, and i'm thinking that a cache ttl of one hour is certainly big enough in regards to both performance (the reduction of writes) and to the UI (ensuring the service and spans names are available to choose from), and i don't think it's too small (is anyone setting the index ttl so low as just a few hours?)

> Second problem - the TTL is on the whole cache, not on individual entries.

This isn't a problem. This improvement is not trying to remove duplicate writes, simply improve performance (and remove the need for bucketing).

> Also, thread locals are unnecessary here, a concurrent map shared between threads would do.

concurrency has overhead, here it is a choice between letting the zipkin jvm handle it or letting cassandra handle it (ie duplicates writes). performance wise, which is the context for this improvement, cassandra is superior at handling the concurrency. hence the thread local approach.

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
jamescway, line 526: makes sense :)

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
jamescway: In a heavy write scenario would it be possible for the memory usage to grow to an unwieldy size?  In this case maybe it would be worth using an LRUMap (threadsafe), which evicts based on size and might reduce the need to nuke it periodically.  I also heard that the [Guava cache](https://code.google.com/p/guava-libraries/wiki/CachesExplained) is super cool and full featured.  Another possible scenario would be getting cache misses if the data is in the Set in another thread.

michaelsembwever: the size of the set is a potential concern, and i had considered it, but i can't imagine the number of service and span names within any one hour having such high cardinality.

if it is a problem one simple way to solve it is just limit the size of the set and let writes spill over.
 (if this were to happen i would re-introduce the bucketing, one reason i left it in the schema)

but the easiest approach would be just to bring the cache ttl from one hour down to for example one minute.

>  I also heard that the Guava cache is super cool and full featured. 

a guava cache (or a LRUMap) introduces concurrency which as i tried to explain above is just overhead, which i'm explicitly avoiding here.

> Another possible scenario would be getting cache misses if the data is in the Set in another thread.

not a problem really. a cache miss only means that the write happens. today we have a 100% cache miss and it's "not a problem" :-)

jamescway: Limiting the size sounds good, I think evicting would be important to prevent spill over and cache would always be warm (maybe thread local lrumap O_o).  Sounds like we could just revisit these issues should they surface.

jamescway:
Definitely some room for additional work like you mentioned, but sounds like we could revisit. 
:+1:

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
michaelsembwever: > Limiting the size sounds good, …

What about solving it with Yuri suggestion?

> In this OSS implementation I'd suggest making this TTL externally configurable.

Then in any setup where the size of the thread-local maps become an issue it is just to set the cache ttl to a smaller value.

jamescway: Yeah, I think it would definitely help.  We would have something we could tune, should a problem arise.
Have fun with the flags :stuck_out_tongue_winking_eye:

codefromthecrypt: Fwiw I think most of the span stores accept a TTL, which not sure if the
granularity is right. That said, James raises a good point. We don't have a
dynamic config story. At least having it statically configured (similar to
default ttl in SpanStore) would seem a good start.

michaelsembwever:
I'll add the flag then... :)

michaelsembwever:
flag (just a simple system variable to begin with) added with https://github.com/thelastpickle/zipkin/commit/2f5ac5753e30ac319a3b5f2de2972a89ea4b2839

(will squash to this PR)

jamescway:
:shipit: merge it!

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -70,6 +71,26 @@
     private final PreparedStatement selectTraceDurationTail;
     private final Map<String,String> metadata;
 
+    private final ThreadLocal<Set<String>> writtenNames = new ThreadLocal<Set<String>>() {
+            private long hour = toTtlInterval(System.currentTimeMillis());
```
michaelsembwever: the field name "hour" is misleading

michaelsembwever:
two small fixes coming for the hour field and https://github.com/thelastpickle/zipkin/commit/2f5ac5753e30ac319a3b5f2de2972a89ea4b2839#commitcomment-12938413

codefromthecrypt:
supernit which is to make the first line of your commit message similar to the title, as it currently reads "problem: " :)

LGTM

michaelsembwever:
https://github.com/thelastpickle/zipkin/commit/efbed5b6684dc54d37197432789401aa55cc714d
addresses the field and property naming.

it's been squashed into this PR, with the commit msg given a title.

codefromthecrypt:
great progress!

PR #666, opened 2015-08-28
remove zookeeper from the architecture diagram, and show the collecto…

michaelsembwever:
…r protocol as either scribe or kafka.


jamescway:
:+1:

PR #668, opened 2015-08-29
Obsoletes thrift query for Order and deprecates TraceIdDuration 

codefromthecrypt:
Removing the unused order query removes complex logic from zipkin. It
also removes the only caller of `SpanStore.getTracesDuration`, its
supporting class `TraceIdDuration` and backend-specific indexes that
only supported trace duration.

Traces are ordered in the UI javascript. The only caller of order was
tracegen. This warns if the order argument is passed, which would only
occur during ad-hoc query commands. It renames the thrift fields to
obsolete, which retains binary compatibility while still noting its
status.


eirslett:
Nice! +1

codefromthecrypt:
actually there's more to remove.. the only user of `spanStore.getTracesDuration` was ordering!

codefromthecrypt:
ok this is much better... far more mileage out of removing support for Order!

zipkin-cassandra-core/src/main/resources/cassandra-schema-cql3.txt
```diff
@@ -51,12 +51,6 @@ CREATE TABLE IF NOT EXISTS zipkin.service_names (
     PRIMARY KEY (service_name)
 ) WITH compaction = {'min_threshold': '4', 'class': 'org.apache.cassandra.db.compaction.SizeTieredCompactionStrategy', 'max_threshold': '32'};
 
-CREATE TABLE IF NOT EXISTS zipkin.duration_index (
```
codefromthecrypt, line 54: @michaelsembwever note.. this is no longer needed

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -757,69 +732,6 @@ private String debugInsertTraceIdByAnnotation(ByteBuffer annotationKey, long tim
                 .replace(":ttl_", String.valueOf(ttl));
     }
 
-    public Map<Long,Long> getTraceDuration(boolean head, long[] traceIds) {
```
codefromthecrypt, line 760: @michaelsembwever fyi this no longer has a caller, so it and its supporting things are nuked.

PR #669, opened 2015-08-30
Removes obsolete id query methods

codefromthecrypt:
Source history shows thrift `QueryRequest` has been in use by web since
at least June of 2014 (possibly longer, as source history is truncated).
The methods QueryRequest obsoleted are now removed.

### DIFF
--- a/zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinQuery.thrift
+++ b/zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinQuery.thrift
@@ -68,9 +68,6 @@ struct TraceCombo {
   4: optional map<i64, i32> span_depths # not set if no spans in trace
 }
 
-/** This is obsolete. Use Order.NONE for compatibility. */
-enum Order { TIMESTAMP_DESC, TIMESTAMP_ASC, DURATION_ASC, DURATION_DESC, NONE }
-
 /**
  * The raw data in our storage might have various problems. How should we adjust it before
  * returning it to the user?
@@ -86,9 +83,11 @@ struct QueryRequest {
   2: optional string span_name
   3: optional list<string> annotations
   4: optional list<zipkinCore.BinaryAnnotation> binary_annotations
+  /** results will have epoch microsecond timestamps before this value */
   5: i64 end_ts
+  /** maximum entries to return before "end_ts" */
   6: i32 limit
-  7: Order OBSOLETE_order = Order.NONE
+  # 7: Order OBSOLETE_order = Order.NONE
 }
 
 struct QueryResponse {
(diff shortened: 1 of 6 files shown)


zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinQuery.thrift
```diff
@@ -78,9 +78,6 @@ struct TraceCombo {
   4: optional map<i64, i32> span_depths # not set if no spans in trace
 }
 
-/** This is obsolete. Use Order.NONE for compatibility. */
-enum Order { TIMESTAMP_DESC, TIMESTAMP_ASC, DURATION_ASC, DURATION_DESC, NONE }
```
codefromthecrypt, line 72: only user was obsolete query methods.

jamescway, line 72: out of curiosity, does this break anything for people who used the old definition?

jamescway:
:+1: just had a non-merge-blocking question

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinQuery.thrift
codefromthecrypt, line 72: well, the only caller of this was web. and thrift will ignore fields it doesn't know about (ex. field 7 below).

An experiment to prove this would be to run `bin/query` and `bin/collector` from master, and then run `bin/web` from tag 1.2.2.

codefromthecrypt, line 72: After I rebase, I'll double-check again.

jamescway, line 72: cool.  maybe chat with you later about integration testing offline.

jamescway:
:+1:  :+1:

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinQuery.thrift
```diff
@@ -96,9 +93,10 @@ struct QueryRequest {
   2: optional string span_name
   3: optional list<string> annotations
   4: optional list<zipkinCore.BinaryAnnotation> binary_annotations
+  /** results will have epoch microsecond timestamps before this value */
   5: i64 end_ts
+  /** maximum entries to return before "end_ts" */
   6: i32 limit
-  7: Order OBSOLETE_order = Order.NONE
```
codefromthecrypt, line 91: will change this to a comment so that we don't re-use ordinal 7

codefromthecrypt:
@jamescway yeah.. just verified old UI still works fine (extra unnecessary order param is ignored)

jamescway:
nice :ramen:

PR #670, opened 2015-08-30
Removes unused tracesExist query

codefromthecrypt:
`tracesExist` was only referenced by an unused private method in web.
`getSpansByTraceIds` ignores absent ids anyway.


jamescway:
:+1:

PR #672, opened 2015-08-30
Removes TTL from Query API

codefromthecrypt:
TTLs in zipkin are backend-specific and at best partially supported as
they aren't exposed in the UI. This change removes TTL from the query
api, and removes trace-specific TTL controls. It leaves support for
global TTLs in Cassandra and Redis.

`getDataTimeToLive` was implemented by Cassandra and Redis, but not used
by the UI. `setTimeToLive` by trace was implemented by Cassandra and
Redis, but silently fails when unsupported. Cassandra support of this
feature led to brittle tests and corresponding complexity. When we
consider TTL by trace wasn't used by the UI either, but rather an
undocumented "pin" api, the feature's existence becomes more curious.
This isn't helped by no history of this feature in any GitHub issue.

### DIFF
--- a/zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
+++ b/zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
@@ -6,7 +6,7 @@ import com.twitter.finagle.stats.{StatsReceiver, Stat}
 import com.twitter.finagle.tracing.SpanId
 import com.twitter.finagle.{Filter, Service, SimpleFilter}
 import com.twitter.io.Buf
-import com.twitter.util.{Duration, Future}
+import com.twitter.util.Future
 import com.twitter.zipkin.{Constants => ZConstants}
 import com.twitter.zipkin.common.json._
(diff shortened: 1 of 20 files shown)


zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
```diff
@@ -454,40 +454,4 @@ class Handlers(jsonGenerator: ZipkinJson, mustacheGenerator: ZipkinMustache, que
...
         }
     }
-
-  def handleIsPinned(client: ZipkinQuery[Future]): Service[Request, Renderer] =
-    new NotFoundService {
-      def process(req: Request): Option[Future[Renderer]] =
-        pathTraceId(req.path.split("/").lastOption) map { id =>
-          client.getTraceTimeToLive(id) map { ttl => JsonRenderer(ttl) }
-        }
-    }
```
jamescway, line 447: reminds me of ruby's method_missing LOL

jamescway:
:+1:

PR #673, opened 2015-08-30
Removes top annotation aggregates

codefromthecrypt:
Top annotations was introduced in 2012 and not mentioned since. The
corresponding hadoop jobs at Twitter have been broken since last year.

This removes the dead apis from the UI and the corresponding code in
cassandra (which incidentally wasn't used).

### DIFF
--- a/zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
+++ b/zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
@@ -1,7 +1,6 @@
 
 package org.twitter.zipkin.storage.cassandra;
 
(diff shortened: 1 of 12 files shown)


zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -330,77 +308,6 @@ private String debugSelectTraceTtl(long traceId) {
         return selectTraceTtl.getQueryString().replace(":trace_id", String.valueOf(traceId));
     }
 
-    public List<String> getTopAnnotations(String key) {
```
codefromthecrypt, line 333: wasn't actually used by spanstore

jamescway:
:+1:

PR #674, opened 2015-08-30
Obsoletes unused fields in thrift QueryResponse

codefromthecrypt:
`QueryResponse` required supplying startTs and endTs, fields neither
used by the UI nor tested. This change removes the code around that,
simplifying the interface.


jamescway:
:+1:

PR #676, opened 2015-08-30
Cleans up redis and fixes misleading test

codefromthecrypt:


zipkin-redis/src/test/scala/com/twitter/zipkin/storage/redis/RedisIndexSpec.scala
```diff
@@ -55,56 +56,50 @@ class RedisIndexSpec extends RedisSpecification {
...
   }
 
-  test("index only on annotation in each span with the same value") {
-    result(redisIndex.indexSpanByAnnotations(span3))
-  }
-
   test("getTraceIdsByName") {
-    var ls = List[Long]()
-    result(redisIndex.indexTraceIdByServiceAndName(span1))
-    result(redisIndex.getTraceIdsByName("service", None, 0, 3)) foreach {
```
codefromthecrypt, line 74: this and below it silently failed based on trying to have a timestamp comparison with zero!

PR #680, opened 2015-08-31
remove oraclejdk7 from build

jamescway:
Because 1.4.0 was released using scala 2.11 we can remove jdk7 from the build now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ services:
 
 matrix:
   include:
-    - jdk: "oraclejdk7"
-      env: GRADLE_OPTS="-Dorg.gradle.daemon=false -XX:+CMSClassUnloadingEnabled -XX:+UseConcMarkSweepGC -Xms2G -Xmx2G -Xss512k -XX:MaxPermSize=256m"
     - jdk: "oraclejdk8"
       env: GRADLE_OPTS="-Dorg.gradle.daemon=false -Xms2G -Xmx2G -Xss512k -XX:MetaspaceSize=256m"
 
--- a/build.gradle
+++ b/build.gradle
@@ -20,6 +20,8 @@ allprojects {
     apply plugin: 'com.jfrog.bintray'
     apply plugin: "com.jfrog.artifactory-upload"
     apply from: "$rootDir/gradle/dependencies.gradle"  // common dependency versions
+    sourceCompatibility = 1.7
+    targetCompatibility = 1.7
 }
 
 import org.gradle.util.VersionNumber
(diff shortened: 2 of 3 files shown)


codefromthecrypt:
happy

eirslett:
But are we sure Twitter will be able to stay up to date? (Do we care?)

codefromthecrypt:
@mosesn safe to assume twitter will be able to build on JDK8 this year, right?

codefromthecrypt:
I know zipkin was running on JRE 8 when I left, but yeah building is a different topic. At worst, we could switch the compiler (if not already set) to produce Java 7 bytecode.

codefromthecrypt:
ps if we want to decouple, add this to the main build.gradle after apply plugin java

```
    sourceCompatibility = 1.7
    targetCompatibility = 1.7
```

eirslett:
But Twitter runs on the published fat-jars now anyways, right? Developers working on Zipkin can fork the project from GitHub and build locally with JDK8 and Gradle. (Pants is gone anyways)

codefromthecrypt:
No twitter doesn't use the fat-jars (or the -service projects at all).
There's an internal TwitterServer, configured by flags, which uses the
Factory approach as opposed to the ostrich stuff.

If zipkin's libraries are source-level 8, then pants (the internal build
tool) would need to run on 1.8 to compile them. @stuhood might know if that
migration is complete.

Regardless, twitter can update to 1.4.0 (released today) as that was still
compiled to target Java 7. cc @CaitieM20

eirslett:
Ok, source compatibility with Java 7 is probably a good idea then. (We won't gain anything by not being source compatible)

codefromthecrypt:
@jamescway thanks for updating the settings.. I'll run this locally and see if the classes are built properly.. hang tight!

codefromthecrypt:
LGTM 51 is Java 7

``` bash
$ ./gradlew -version

------------------------------------------------------------
Gradle 2.6
------------------------------------------------------------

Build time:   2015-08-10 13:15:06 UTC
Build number: none
Revision:     233bbf8e47c82f72cb898b3e0a96b85d0aad166e

Groovy:       2.3.10
Ant:          Apache Ant(TM) version 1.9.3 compiled on December 23 2013
JVM:          1.8.0_51 (Oracle Corporation 25.51-b03)
OS:           Mac OS X 10.10.5 x86_64

$ ./gradlew clean install
--snip--
$ file ./zipkin-common/build/classes/main/com/twitter/zipkin/util/Util.class
./zipkin-common/build/classes/main/com/twitter/zipkin/util/Util.class: compiled Java class data, version 51.0
```

PR #685, opened 2015-09-03
Consolidates and cleans up getTraces thrift query

codefromthecrypt:
This deprecates thrift calls for zipkin-web data conversion and
consolidates usage to a getTraces query. We will drop the obsolete
methods in 1.5.0, resulting in a much cleaner thrift api.

Before this change, there was a leaky abstraction between the query and
the web packages. For example, types like TraceCombo were introduced
eventhough they were only a single stage of data conversion and only
used to support one web view. The web component now handles data
conversions directly. Moreover, the web component duplicated work by
doing an id scan which implied looking up spans before looking up spans.

This also cleans up a couple questionable api concerns: adjuster and
binary annotation query.

In practice, there was only one type of adjustment, time-skew, and this
was almost always enabled. The code was confused by an idea that there
would be many types of adjustments, and in some places time-skew was
applied multiple times. This moves it to a boolean flag, just like the
http api.

The binary annotation query in thrift was a weird one. It allows the
user to send a string -> string query only. However, the signature
accepted literally the BinaryAnnotation type, which has other members,
and might not be a string. This clarifies the concern by only accepting
a map of string -> string.


zipkin-common/src/main/scala/com/twitter/zipkin/query/TraceSummary.scala
```diff
@@ -29,29 +29,29 @@ object TraceSummary {
...
+ *
  * @param traceId id of this trace
  * @param startTimestamp when did the trace start?
  * @param endTimestamp when did the trace end?
  * @param durationMicro how long did the traced operation take?
- * @param serviceCounts name of the services involved in the traced operation
- *                      mapped to the number of spans of that service
  * @param endpoints endpoints involved in the traced operation
  */
 case class TraceSummary(
```
codefromthecrypt, line 53: this will move to the web package when we remove the obsolete thrift methods.

zipkin-query/src/main/scala/com/twitter/zipkin/query/ThriftQueryService.scala
```diff
@@ -118,53 +122,74 @@ class ThriftQueryService(
...
+    }
 
-        case _ =>
-          // TODO: timestamps endTs is the wrong name for all this
-          querySlices(sliceQueries, qr.copy(limit = 1)) flatMap { ids =>
-            val ts = padTimestamp(ids.flatMap(_.map(_.timestamp)).reduceOption(_ min _).getOrElse(0))
-            querySlices(sliceQueries, qr.copy(endTs = ts)) flatMap { ids =>
-              queryResponse(traceIdsIntersect(ids), qr)
-            }
+  def traceIds(qr: thriftscala.QueryRequest): Future[Seq[Long]] = {
```
codefromthecrypt, line 131: mostly indent from extracting from getTraceIds

zipkin-query/src/main/scala/com/twitter/zipkin/query/ThriftQueryService.scala
```diff
@@ -118,53 +122,74 @@ class ThriftQueryService(
...
+    val sliceQueries = Seq[Option[Seq[SliceQuery]]](
+      qr.spanName.map { n => Seq(SpanSliceQuery(n)) },
+      qr.annotations.map { _.map { AnnotationSliceQuery(_, None) } },
+      qr.binaryAnnotations.map { _.map { e => AnnotationSliceQuery(e._1, Some(ByteBuffer.wrap(e._2.getBytes(UTF_8)))) }(collection.breakOut) },
```
codefromthecrypt, line 135: this is the only actual change

eirslett:
+1 LGTM

spencergibb:
belated :+1:

jamescway:
:+1:

PR #688, opened 2015-09-03
add readme for docker

jamescway:
thanks to @jfeltesse-mdsol!

### DIFF
--- a/doc/docker-zipkin.md
+++ b/doc/docker-zipkin.md
@@ -1,0 +1,154 @@
+# Docker toolbox & docker-zipkin
+
+## Installation
+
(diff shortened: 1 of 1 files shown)


doc/docker.md
```diff
@@ -0,0 +1,154 @@
+# Docker toolbox & docker-zipkin
```
jfeltesse-mdsol: you might want to add "on Mac OS X" because parts are mac os only (eg, `brew install` stuff)

jfeltesse-mdsol:
:+1:

codefromthecrypt:
Only suggestion I have is that it is probably in the wrong repo. Docs for
docker should be in docker-zipkin else we are surely to duplicate efforts.
Regardless, we can link out to this.

Make sense?

jfeltesse-mdsol:
> Docs for docker should be in docker-zipkin

:+1:

PR #689, opened 2015-09-03
Removes duplicate scribe implementation

codefromthecrypt:
`zipkin-receiver-scribe` was implented as a replacement for
`zipkin-collector-scribe`, but `zipkin-collector-service` was never
updated to use it. This consolidates towards receiver and deletes the
orphaned code.


codefromthecrypt:
FYI twitter doesn't use any of the affected code.

zipkin-collector-service/build.gradle
```diff
@@ -21,8 +21,7 @@ repositories {
 }
 
 dependencies {
-    compile project(':zipkin-collector-scribe')
-    compile project(':zipkin-receiver-kafka')
```
codefromthecrypt, line 25: removed receiver-kafka as it wasn't used. can be added back in the future.

eirslett:
+1 :beers:

jamescway:
:+1:

PR #690, opened 2015-09-03
Removes obsolete or deprecated methods

codefromthecrypt:
since now we are in version 1.5, we can have nice things.


spencergibb:
:+1:

Issue #682, opened 2015-09-01
End-to-End example of HTrace-Zipkin (via Kafka)


codefromthecrypt:
@clehene noted that the architecture diagram doesn't explain kafka well. Regardless of implementation, we should make it clear, what's going on. Maybe by switching to @abesto's dot approach to diagramming.

Here's what we're looking for

```
htrace instrumentation -> kafka <- zipkin-kafka-receiver -> SpanStore <- zipkin-query <- zipkin-web
```

Also.. we actually want this to work.

codefromthecrypt:
For those watching at home.. the collector can now have kafka enabled. There's configuration options here: https://github.com/openzipkin/zipkin/tree/master/zipkin-receiver-kafka

can't wait to see @clehene's end-to-end

clehene:
https://issues.apache.org/jira/browse/HTRACE-235

codefromthecrypt:
PS besides htrace, I think we should inventory other kafka instrumentation. For example, Brave, Ruby and Finagle. At some point, we should be able to trace zipkin itself without scribe, too.

eirslett:
HTrace support is dropped (or at least moved to a separate repository) so I'll close this issue.

codefromthecrypt:
htrace, not hbase :) Anyway, e2e tests aren't likely to happen in this repo.

eirslett:
Ooos, sorry :-P

Issue #466, opened 2015-07-17
Configuration


eirslett:
How should we configure zipkin-collector/query/web?
The present way is to use Scala code with eval, @adriancole says Twitter has done a lot of refactoring in their apps later, now might be the time to use Twitter flags instead?
We would lose the possibility of writing Scala code in the configuration, so it might be a little more difficult to set up backends as people wish. Another option I suggest is to use [typesafe config](https://github.com/typesafehub/config), which has a pretty flexible setup.

Possible alternatives:
- Config as Scala, eval (like today)
- Twitter flags (runtime command-line options)
- Something else

Thoughts?

codefromthecrypt:
At Twitter, static config that varies by environment is passed via flags or
in some cases json files. These are set in pants args when running a
process on laptop or in Aurora config.

The most thought in this area has been in Finatra. A part of this was
spinning out a project to address config, as particularly the lack of
testability has been troublesome. For example the most recent apps can be
spawned safely managed by a test lifecycle (embedded).

https://github.com/twitter/finatra/blob/master/inject/README.md

Hope this helps!

eirslett:
One thing that separates Zipkin from in-house applications is the need to replace backend technologies like Cassandra/MySQL/HBase. So simple flags might not be enough - e.g. MySQL may have other configuration options than Redis. In that regard, configuration as code is actually quite nice...

The way we configure Zipkin today (with Scala files and eval) is actually quite close to Spring XML configuration, except it's Scala, so we actually ship the Scala compiler with the Zipkin distribution (which of course increases artifact size and dependency tree quite a bit). I think we even load the Scala compiler in-memory as part of Zipkin, but I'm not quite sure how the eval module is written.

dsyer:
My comment from #526:

> Spring Boot would also provide a nice abstraction for externalized configuration (properties, YAML, command line, system properties, env vars). You wouldn't need a huge raft of dependencies just to bind to the environment - probably just a couple of extra jars. And it has support for executable jars in a nicer format than the shadow ones.

It wouldn't involve writing XML unless you really wanted to (XML is a valid, but infrequently used, format for Java `Properties`). The only dependencies you would need would be spring-boot and spring-context (plus a few transitive dependencies). Possibly you could even get away with just spring-boot and spring-core (very thin). (Plus the executable jar thing if we use that.)

eirslett:
[...]

PR #691, opened 2015-09-04
Integrates Kafka receiver into the zipkin-collector-service process

### LINKED ISSUE #547: Kafka receiver activation should be configurable
To enable the kafka receiver, you need to update the zipkin/collector/main.scala file with something like [this](https://gist.github.com/jamescway/14f798685e35a15cb489).  It should be configurable.

codefromthecrypt:
Integrates Kafka receiver into the zipkin-collector-service process

This starts integration with kafka by triggering it on presence of the
`KAFKA_ZOOKEEPER` environment variable. Refinements to defaults will 
occur in later change, with a goal to have the minimum viable setup and 
one that is easy to configure with docker.

For example, if you are starting from a zipkin's source tree, you might 
assign this way:

``` bash
$ KAFKA_ZOOKEEPER=127.0.0.1:2181 bin/collector
# or if zipkin isn't the right topic name
$ KAFKA_ZOOKEEPER=127.0.0.1:2181 KAFKA_TOPIC=notzipkin bin/collector
```

Note: Enabling Kafka doesn't disable the Scribe listener as it's part of
the Collector's thrift api alongside aggregate dependencies. Future work
might decouple that relationshop.

See #682
Closes #547

### DIFF
--- a/zipkin-collector-service/config/collector-dev.scala
+++ b/zipkin-collector-service/config/collector-dev.scala
@@ -13,7 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 import com.twitter.zipkin.anormdb.{SpanStoreBuilder, AggregatesBuilder}
+import com.twitter.zipkin.receiver.kafka.KafkaSpanReceiverFactory
 import com.twitter.zipkin.storage.anormdb.DB
 import com.twitter.zipkin.collector.builder.CollectorServiceBuilder
 import com.twitter.zipkin.storage.Store
@@ -21,5 +23,8 @@ import com.twitter.zipkin.storage.Store
 val db = DB()
 
 val storeBuilder = Store.Builder(SpanStoreBuilder(db, true), AggregatesBuilder(db))
+val kafkaReceiver = sys.env.get("KAFKA_ZOOKEEPER").map(
+  KafkaSpanReceiverFactory.factory(_, sys.env.get("KAFKA_TOPIC").getOrElse("zipkin"))
+)
 
-CollectorServiceBuilder(storeBuilder)
+CollectorServiceBuilder(storeBuilder, kafkaReceiver)
--- a/zipkin-collector-service/config/collector-redis.scala
+++ b/zipkin-collector-service/config/collector-redis.scala
@@ -15,10 +15,11 @@
  */
 
 import com.twitter.finagle.builder.ClientBuilder
-import com.twitter.finagle.redis.{Redis, Client}
-import com.twitter.zipkin.redis
+import com.twitter.finagle.redis.{Client, Redis}
 import com.twitter.zipkin.collector.builder.CollectorServiceBuilder
+import com.twitter.zipkin.receiver.kafka.KafkaSpanReceiverFactory
 import com.twitter.zipkin.storage.Store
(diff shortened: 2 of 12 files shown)


zipkin-collector-service/config/collector-dev.scala
```diff
@@ -13,13 +13,25 @@
...
 import com.twitter.zipkin.storage.Store
+import com.twitter.zipkin.thriftscala
 
 val db = DB()
 
 val storeBuilder = Store.Builder(SpanStoreBuilder(db, true), AggregatesBuilder(db))
 
-CollectorServiceBuilder(storeBuilder)
+// experimental support for kafka. copy-pasted temporarily into collector-*
+val rcv = sys.env.get("KAFKA_ZOOKEEPER_CONNECT").map(zookeeperConnect => {
```
clehene: could someone have `KAFKA_ZOOKEEPER_CONNECT` set (for something else) and activate this by mistake?

zipkin-collector-service/config/collector-redis.scala
```diff
@@ -28,4 +31,11 @@ val client = Client(ClientBuilder().hosts("0.0.0.0:6379")
 
 val storeBuilder = Store.Builder(redis.SpanStoreBuilder(client))
 
-CollectorServiceBuilder(storeBuilder)
+// experimental support for kafka. copy-pasted temporarily into collector-*
```
clehene: left this duplication on purpose?

zipkin-collector-service/src/main/scala/com/twitter/zipkin/collector/builder/CollectorServiceBuilder.scala
```diff
@@ -39,6 +40,7 @@ import com.twitter.zipkin.thriftscala
  */
 case class CollectorServiceBuilder[T](
   storeBuilder: Builder[Store],
+  receiver: Option[(Seq[thriftscala.Span] => Future[Unit]) => SpanReceiver] = None,
```
jamescway: I don't suppose theres something we could do to simplify this?  Do you think it would be worth abstracting this to some kind of type `type SpanReceiver = A => B => C`.

zipkin-receiver-kafka/README.md
```diff
@@ -1,11 +1,19 @@
...
- * ```bin/collector cassandra```
+### Enabling the Kafka receiver with zipkin-collector-service
+`zipkin-collector-service` typically receives spans from scribe. To enable kafka, you need to
+assign the `KAFKA_ZOOKEEPER_CONNECT` variable when starting the process.
 
+For example, if you are starting from a zipkin's source tree, you might assign this way:
+```bash
+$ KAFKA_ZOOKEEPER_CONNECT=127.0.0.1:2181 bin/collector
+```
+
```
clehene, line 14: Add a note on the default topic and perhaps how to override that

jamescway:
:+1: thanks for doing this, I left a NIO comment

codefromthecrypt:
PTAL things are adjusted except @clehene valid feedback on env variable naming. Leaving this way to be consistent with how cassandra is being configured. I _really_ want us to do better when we move from working around ostrich to designing our cli interface in #466

Issue #687, opened 2015-09-03
End-to-End example of Zipkin with mysql


codefromthecrypt:
Zipkin currently has baked configs for file-based SQL, redis and cassandra. @jfeltesse-mdso mentions MySQL is an extremely popular storage backend and would be a great alternative for those getting started with Zipkin and I completely agree.

I suspect this won't be too much effort, as anorm already "supports" it, at least including canning scala files for collector and query and making the url settable.

codefromthecrypt:
Here's a start. I managed to get it working locally today and updated this: https://github.com/openzipkin/zipkin/tree/master/zipkin-anormdb

``` bash
$ mysql -uroot -e "create database if not exists zipkin"
$ MYSQL_USER=root ./bin/collector mysql
$ MYSQL_USER=root ./bin/query mysql
$ ./bin/tracegen
...
```

jfeltesse-mdsol:
awesome! :100:

codefromthecrypt:
PS here's a working docker-compose pointing to amazon RDS. cc @abesto

``` yaml
collector:
  image: quay.io/openzipkin/zipkin-collector:1.7.1
  environment:
    - MYSQL_HOST=zipkin.lalalalalalala.us-west-2.rds.amazonaws.com
    - MYSQL_USER=zipkin
    - MYSQL_PASS=NOT_TELLING
  entrypoint: /bin/sh
  command: -c "java -cp ${ANORMPATCH_CLASSPATH}:/zipkin/zipkin-collector.jar com.twitter.zipkin.collector.Main -f /collector-mysql.scala"
  expose:
    - 9410
  ports:
    - 9410:9410
    - 9900:9900
query:
  image: quay.io/openzipkin/zipkin-query:1.7.1
  environment:
    - MYSQL_HOST=zipkin.lalalalalalala.us-west-2.rds.amazonaws.com
    - MYSQL_USER=zipkin
    - MYSQL_PASS=NOT_TELLING
  entrypoint: /bin/sh
  command: -c "java -cp ${ANORMPATCH_CLASSPATH}:/zipkin/zipkin-query.jar com.twitter.zipkin.query.Main -f /query-mysql.scala"
  expose:
    - 9411
  ports:
    - 9411:9411
    - 9901:9901
web:
  image: quay.io/openzipkin/zipkin-web:1.7.1
  links:
    - query
  ports:
    - 8080:8080
    - 9990:9990
```

Issue #655, opened 2015-08-22
Set redis socket with environment variable


codefromthecrypt:
Similar to how we work with cassandra, we should be able to set the redis socket using an environment variable.
https://github.com/openzipkin/zipkin/blob/master/zipkin-collector-service/config/collector-cassandra.scala#L26

This would allow us to make docker files similarly to how we do with cassandra.
https://github.com/openzipkin/docker-zipkin/blob/master/collector/run.sh#L8

Cassandra naturally supports multiple endpoints, hence the name `CASSANDRA_CONTACT_POINTS`. Redis clients typically only connect to a single endpoint. Looking at redis-cli and hiredis, there's no interesting name for this. Perhaps we just choose `REDIS_SOCKET`?

codefromthecrypt:
@dsyer @spencergibb any good choices of env variable name?

dsyer:
Not that I know of. But maybe a `REDIS_[HOST,PORT,PASSWORD]`?

codefromthecrypt:
done

PR #693, opened 2015-09-04
Adds mysql span storage and consolidates configuration docs

### LINKED ISSUE #686: Much confusion about storage
Hi,

The [documentation says](http://zipkin.io/Architecture.html)

> We originally built Zipkin on Cassandra for storage. It’s scalable, has a flexible schema, and is heavily used within Twitter. However, we made this component pluggable, and we now have support for Redis, MySQL, PostgreSQL, SQLite, and H2.

But nowhere is it documented how to achieve this, or at least I couldn't find it.

The [gradle.properties file refers to backend switching](https://github.com/openzipkin/zipkin/blob/master/gradle.properties#L17-L34) but the naming is confusing. 
"configKey=dev" stands for using AnormDB, and if I read that correctly it needs to be set when switching to a different backend. 
Are we to infer from this name that backends other than cassandra or redis are not production-ready?
Are there known caveats associated with using RDBMS backends?

To give some context we are evaluating Zipkin and being able to do without Cassandra would be a big plus. We have lots of MySQL instances around with qualified people to oversee them, we'd rather not introduce yet another storage solution in the mix.

codefromthecrypt:
There have been a number of questions about how to configure zipkin's
storage backends. This consolidates the way that they are presented,
ensuring all storage options have configuration notes.

See #687 #682 #466
Closes #686 #655 #207

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,5 +14,5 @@ master.
 # License
 By contributing your code, you agree to license your contribution under the 
 terms of the Apache Public License v2: 
-https://github.com/twitter/zipkin/blob/master/LICENSE
+https://github.com/openzipkin/zipkin/blob/master/LICENSE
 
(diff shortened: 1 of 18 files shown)


codefromthecrypt:
going to review-after-commit this one.. feel free to tell me things to follow-up on

spencergibb:
:+1:

PR #694, opened 2015-09-04
Sets cassandra connect port when contact points aren't default

codefromthecrypt:
thanks to @espenkm for reporting!


michaelsembwever:
:+1:

michaelsembwever:
there might remain confusion about this, and it's normal for cassandra consumer to define contact points as only ipaddresses and then port separately.

that is "contact point" in cassandra terminology does not include a port.

i guess that confusion couldn't be avoided either as we're also needing to support the existing "zipkin.store.cassandra.dest" flag.

PR #696, opened 2015-09-06
Decouples zipkin thrift interfaces and top-levels Dependency view

codefromthecrypt:
Finagle/Scrooge do not natively support multiple thrift interfaces on
the same port, eventhough other libraries do (such as FB Swift). To work
around this, zipkin used to expose query and collector services via
a thrift service hierarchy. This coupling was accidental and limits our
ability to be modular.

This change breaks up the thrift interfaces, particularly around the
dependency aggregates, addressing the Finagle limitation in code as
opposed to in thrift definitions.

Since we only expose a single aggregate function (dependencies), and
we very often have questions about it, this replaces the abstract term
aggregates with dependency. It also adds mysql instructions which
actually work!


codefromthecrypt:
This actually works now

screen shot 2015-09-06 at 1 44 44 pm

codefromthecrypt:
ps eventhough the thrifts were re-organized, they are completely compatible. Here's an example of running master's web against the endpoints.

screen shot 2015-09-06 at 2 03 41 pm

eirslett:
Finally, the UI is live again! Thanks a lot!
I've been working with the hadoop job, but it's tough to make it work. I'll consider copypasting Cassie into there, just to get something that works....

+1 on this PR!

PR #321, opened 2013-12-24
Create SpanStore + cassie example

sprsquish:
Combine storage and index into a single concept. Stores should handle
indexing however they need and simply present a store/retrieve API.

PR #702, opened 2015-09-09
Removes double-compression in cassandra

codefromthecrypt:
## Note to those who are already running the new (datastax) cassandra

This change implies a one-time hit, removing the legacy of double-compressing at the cost of making existing spans unreadable (iotw either drop and re-create or do data conversion to unsnappy all blobs) 

Those who haven't updated to the new cassandra driver (still using cassie), will need to recreate your schema when you update anyway.

---

20 months ago, client-side snappy compression was called into question
as Cassandra does server-side compression now.

https://github.com/openzipkin/zipkin/pull/321#discussion-diff-8702086

We've pulled this compression along, despite scary comments in the code
and despite compression being default unless it is explicitly disabled.

This removes client-side compression and punts to Cassandra server
configuration.

See #321


codefromthecrypt:
cc @eirslett @michaelsembwever @gneokleo

zipkin-cassandra/build.gradle
```diff
@@ -4,11 +4,6 @@ dependencies {
     compile project(':zipkin-scrooge')
     compile project(':zipkin-cassandra-core')
 
-    compile 'org.iq80.snappy:snappy:0.4'
```
codefromthecrypt, line 7: yay less deps!

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/SnappyCodec.scala
```diff
@@ -1,51 +0,0 @@
...
-import com.twitter.zipkin.storage.cassandra.Codec
-import java.nio.ByteBuffer
-import com.twitter.zipkin.util.Util
-import org.iq80.snappy.Snappy
-import com.twitter.ostrich.stats.Stats
-import com.twitter.finagle.tracing.Trace
-
-/**
- * Cassie codec that wraps another and compresses/decompresses that data.
- * TODO remove this once we are running a Cassandra version with internal compression.
```
codefromthecrypt, line 27: :eyes:

gneokleo:
This is a breaking change for some people. Some spans they might already have stored will not be readable. I would bump the version in the gradle file too

For the record this is what redis is doing. 
https://github.com/openzipkin/zipkin/blob/master/zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisStorage.scala#L23

codefromthecrypt:
@gneokleo NP bumped to 1.7.0-SNAPSHOT

codefromthecrypt:
also updated the description

PR #704, opened 2015-09-10
anormDB: move the DB schema out of the code in the case of MySQL

jfeltesse-mdsol:
The user will have to load the schema manually, just like when using Cassandra.

This is following a discussion on the chat about how docker-zipkin has to fetch the cassandra schema at build time while a mysql based version wouldn't have such a step.
The benefit of manually loading the schema is that we wouldn't need the collector service to start before the query service so that it can setup the schema auto-magically.

Note that the code changes are untested, I am not a scala hacker.
In other words,

enhanced-14836-1414320930-8

@adriancole have a look when you have a minute please.

### DIFF
--- a/zipkin-anormdb/src/main/resources/mysql-indexes.sql
+++ b/zipkin-anormdb/src/main/resources/mysql-indexes.sql
@@ -1,0 +1,18 @@
+ALTER TABLE zipkin_spans ADD PRIMARY KEY(span_id);
+ALTER TABLE zipkin_spans ADD INDEX(trace_id);
+ALTER TABLE zipkin_spans ADD INDEX(span_name(64));
+ALTER TABLE zipkin_spans ADD INDEX(created_ts);
+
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
PS I'm offline until Monday. Anyone else can feel free to look at this,
else I will then.

codefromthecrypt:
@jfeltesse-mdsol so I think we still want this to install the schema, even if it is reading from files. (preserving the install=true contract). I'll add work on your commit to do that.

codefromthecrypt:
ps I think these should be the same file. the only reason the indexes weren't in the default formerly was that index syntax isn't portable.

codefromthecrypt:
actually I see what you mean. mysql needs a separate step to install the database anyway, and probably prod usage wouldn't want to give it access to change its schema.

codefromthecrypt:
#705

codefromthecrypt:
Thanks for the start! I added polish on the other.

PR #705, opened 2015-09-14
Move the DB schema out of the code in the case of MySQL

codefromthecrypt:
Polishes @jfeltesse-mdsol's pull request, by documenting how to install it, and erring if someone expects auto-install.

``` bash
$ mysql -uroot -e "create database if not exists zipkin"
# install the schema and indexes
$ mysql -uroot -Dzipkin < zipkin-anormdb/src/main/resources/mysql.sql
$ MYSQL_USER=root ./bin/collector mysql
...
```

Fixes #704


zipkin-anormdb/src/main/resources/mysql.sql
```diff
@@ -0,0 +1,69 @@
...
+  dlid BIGINT NOT NULL,
+  parent VARCHAR(255) NOT NULL,
+  child VARCHAR(255) NOT NULL,
+  m0 BIGINT NOT NULL,
+  m1 DOUBLE PRECISION NOT NULL,
+  m2 DOUBLE PRECISION NOT NULL,
+  m3 DOUBLE PRECISION NOT NULL,
+  m4 DOUBLE PRECISION NOT NULL
+);
+ALTER TABLE zipkin_spans ADD PRIMARY KEY(span_id);
```
jfeltesse-mdsol: nitpicking here but there should be an empty line above and, having said that, if we're going to have the schema and index defs done at once maybe having the index defs close to the table they change would be easier to understand?

codefromthecrypt: sure thing

jfeltesse-mdsol:
cool stuff! :100:

zipkin-anormdb/src/main/resources/mysql.sql
codefromthecrypt: done

jfeltesse-mdsol:
:ok_woman:

PR #707, opened 2015-09-14
Decouples HikariCP so that MySQL work can complete

codefromthecrypt:
Decouples HikariCP code so we can temporarily override it in Docker.

See https://github.com/brettwooldridge/HikariCP/issues/415
See #687


abesto:
I probably wouldn't notice any deep issues in this (no knowledge of the related tech), but FWIW, it looks good to me.

PR #709, opened 2015-09-14
Add snappy compression to redis

gneokleo:
- This will reduce the footprint for redis by more than 50% according to benchmarks.
- Thrift serializer has also been changed from Binary to Compact

### DIFF
--- a/zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisSnappyCodec.scala
+++ b/zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisSnappyCodec.scala
@@ -1,0 +1,26 @@
+package com.twitter.zipkin.storage.redis
+
+import com.twitter.scrooge.{ThriftStruct, ThriftStructSerializer}
+import org.iq80.snappy.Snappy
+import org.jboss.netty.buffer.{ChannelBuffer, ChannelBuffers}
+
+/**
+ * Snappy serializer and deserializer for thrift structs
+ */
+class RedisSnappyThriftCodec[T <: ThriftStruct](val serializer: ThriftStructSerializer[T]) {
+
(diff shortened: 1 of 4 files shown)


zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisSnappyCodec.scala
```diff
@@ -0,0 +1,37 @@
...
+ */
+class RedisSnappyThriftCodec[T <: ThriftStruct](val serializer: ThriftStructSerializer[T]) {
+
+  val CompressTimeMetricName = "snappycodec.compress"
+  val CompressBytesMetricName = "redissnappythriftcodec.compressed"
+  val UncompressTimeMetricName = "snappycodec.uncompress"
+  val UncompressBytesMetricName = "redissnappythriftcodec.uncompressed"
+
+  def encode(t: T): ChannelBuffer = {
+    Stats.time(CompressTimeMetricName) {
```
codefromthecrypt: not sure we should be doing this in a tight loop. when you are finished testing, mind removing the stats?

codefromthecrypt:
love it except for the stats. if we want, we can start a microbenchmark around compression time, otherwise, I'd leave it out.

zipkin-redis/src/main/scala/com/twitter/zipkin/storage/redis/RedisSnappyCodec.scala
gneokleo: done

eirslett:
Will this break backwards-compatibility for clients that send traces that are not snappy-encoded?

gneokleo:
Yes, it does. Same as the cassandra change a few commits ago which removed snappy compression on the client side

codefromthecrypt:
PS this merged after a minor version increment

PR #710, opened 2015-09-14
Resurrects ability to change collector and query ports

### LINKED ISSUE #659: Cannot change port of the query service
After this commit, it is no longer possible to change the port of the query service:
- https://github.com/openzipkin/zipkin/commit/35a3891e664ceb4d7584a6a808a110299876016b#diff-5e7107cba5b23b0e69b1ae6e8cff978fR60

Previously, you could do this to change the port:

```
val serverBuilder = ZipkinServerBuilder(9416, 9906)
QueryServiceBuilder(storeBuilder, serverBuilder)
```

But this no longer works. From reading the source, it looks like this should work now but it doesn't work either:

```
java -jar … -D zipkin.queryService.port=127.0.0.1:9416
```

codefromthecrypt:
The collector and query ports used to be possible to change. This makes
that possible for all non-dev configurations. It also documents them.

Here are the new properties:

```
* `COLLECTOR_PORT`: Listen port for scribe and dependency apis
* `COLLECTOR_ADMIN_PORT`: Listen port for the ostrich admin http
* `COLLECTOR_LOG_LEVEL`: Log level written to the console
* `COLLECTOR_SAMPLE_RATE`: Updatable via POST /config/sampleRate

* `QUERY_PORT`: Listen port for the query thrift api
* `QUERY_ADMIN_PORT`: Listen port for the ostrich admin http
* `QUERY_LOG_LEVEL`: Log level written to the console
```

Note this removes some extraneous configuration and sets default log
level to INFO, not DEBUG.

Fixes #659

### DIFF
--- a/zipkin-collector-service/src/main/scala/com/twitter/zipkin/collector/builder/CollectorServiceBuilder.scala
+++ b/zipkin-collector-service/src/main/scala/com/twitter/zipkin/collector/builder/CollectorServiceBuilder.scala
@@ -43,35 +43,20 @@ case class CollectorServiceBuilder[T](
   storeBuilder: Builder[Store],
   receiver: Option[SpanReceiver.Processor => SpanReceiver] = None,
   scribeCategories: Set[String] = Set("zipkin"),
+  /**
+   * Endpoints are available via
+   *   GET  /config/<name>
+   *   POST /config/<name>?value=<value>
+   */
   sampleRateBuilder: Builder[AdjustableRateConfig] = Adjustable.local(1.0),
   adaptiveSamplerBuilder: Option[Builder[AdaptiveSamplerConfig]] = None,
-  additionalConfigEndpoints: Seq[(String, Builder[AdjustableRateConfig])] = Seq.empty,
-  queueMaxSize: Int = 500,
-  queueNumWorkers: Int = 10,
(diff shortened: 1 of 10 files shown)


codefromthecrypt:
@wadey this one's for you

zipkin-collector-service/src/main/scala/com/twitter/zipkin/collector/builder/CollectorServiceBuilder.scala
```diff
@@ -43,35 +43,20 @@ case class CollectorServiceBuilder[T](
...
   scribeCategories: Set[String] = Set("zipkin"),
+  /**
+   * Endpoints are available via
+   *   GET  /config/<name>
+   *   POST /config/<name>?value=<value>
+   */
   sampleRateBuilder: Builder[AdjustableRateConfig] = Adjustable.local(1.0),
   adaptiveSamplerBuilder: Option[Builder[AdaptiveSamplerConfig]] = None,
-  additionalConfigEndpoints: Seq[(String, Builder[AdjustableRateConfig])] = Seq.empty,
-  queueMaxSize: Int = 500,
```
codefromthecrypt, line 49: these did nothing.. we can re-introduce them when that's not the case

PR #711, opened 2015-09-15
Removes timing of individual base64+thrift decoding

codefromthecrypt:
This sort of thing is inappropriate to do in a tight loop. Let's benchmark if we are curious.


eirslett:
Timing checks could be moved to unit tests that we can run whenever we need to benchmark changes.

codefromthecrypt:
I agree, except not as unit tests, but microbenchmarks (JMH).

PR #712, opened 2015-09-15
Enables tracing of zipkin itself when SCRIBE_HOST or SCRIBE_PORT are set

codefromthecrypt:
If `SCRIBE_HOST` or `SCRIBE_PORT` are set, Finagle's `DefaultTracer`
sends all traces to the specified destination, often the
`zipkin-collector`. Otherwise, the `DefaultTracer` is disabled.

This applies to the `zipkin-web` and `zipkin-query` services, which are
now labeled correctly as services. `zipkin-collector` has tracing
explicitly disabled to discourage cyclic relationships.

### DIFF
--- a/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
+++ b/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
@@ -224,7 +224,7 @@ class AnormSpanStore(val db: DB, val openCon: Option[Connection] = None) extends
 
   override def getTraceIdsByAnnotation(serviceName: String, annotation: String, value: Option[ByteBuffer],
     endTs: Long, limit: Int): Future[Seq[IndexedTraceId]] = db.inNewThreadWithRecoverableRetry {
-    if ((Constants.CoreAnnotations ++ Constants.CoreAddress).contains(annotation) || endTs <= 0 || limit <= 0) {
+    if ((Constants.CoreAnnotations).contains(annotation) || endTs <= 0 || limit <= 0) {
       Seq.empty
     }
     else {
--- a/zipkin-query-service/src/main/scala/com/twitter/zipkin/builder/QueryServiceBuilder.scala
+++ b/zipkin-query-service/src/main/scala/com/twitter/zipkin/builder/QueryServiceBuilder.scala
@@ -16,6 +16,8 @@
 package com.twitter.zipkin.builder
 
 import com.twitter.finagle.ListeningServer
+import com.twitter.finagle.tracing.{NullTracer, DefaultTracer}
+import com.twitter.finagle.zipkin.thrift.RawZipkinTracer
 import com.twitter.ostrich.admin.RuntimeEnvironment
 import com.twitter.zipkin.query.ZipkinQueryServerFactory
 import com.twitter.zipkin.storage.Store
@@ -28,6 +30,15 @@ case class QueryServiceBuilder(
   def apply(): (RuntimeEnvironment) => ListeningServer = (runtime: RuntimeEnvironment) => {
     serverBuilder.apply().apply(runtime)
     val store = storeBuilder.apply()
+
+    // If a scribe host is configured, send all traces to it, otherwise disable tracing
+    val scribeHost = sys.env.get("SCRIBE_HOST")
+    val scribePort = sys.env.get("SCRIBE_PORT")
+    DefaultTracer.self = if (scribeHost.isDefined || scribePort.isDefined) {
+      RawZipkinTracer(scribeHost.getOrElse("localhost"), scribePort.getOrElse("1463").toInt)
+    } else {
+      NullTracer
+    }
 
     object UseOnceFactory extends com.twitter.app.App with ZipkinQueryServerFactory
     UseOnceFactory.nonExitingMain(Array(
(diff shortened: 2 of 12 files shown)


codefromthecrypt:
cc @kristofa @yurishkuro

zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
```diff
@@ -102,6 +103,16 @@ trait ZipkinWebFactory { self: App =>
...
 object Main extends TwitterServer with ZipkinWebFactory {
   def main() {
+    // If a scribe host is configured, send all traces to it, otherwise disable tracing
+    val scribeHost = sys.env.get("SCRIBE_HOST")
+    val scribePort = sys.env.get("SCRIBE_PORT")
+    if (scribeHost.isDefined || scribePort.isDefined) {
+      host.parse(scribeHost.getOrElse("localhost") + ":" + scribePort.getOrElse("1463"))
+      initialSampleRate.parse("1.0")
+    } else {
+      initialSampleRate.parse("0.0")
```
yurishkuro: Is there a way to completely disable scribe reporter in this case? I am not sure if setting sampling to 0 would prevent it from trying to make connections to scribe. I was able to disable ALL trace reporting with below, but not selectively disable only scribe reporter.

```
    val role = Stack.Role("TraceInitializerFilter")
    val stack = Httpx.Server.stack.remove(role) // hack!!!
    val server = Httpx.Server(stack)
    val web = server.serve(webServerPort(), ServiceFactory.const(webService))
```

codefromthecrypt: Short answer is yes! I have a different impl which I will push when I get
connected later.

codefromthecrypt:
ok much cleaner impl now, based on advise from @kristofa and feedback from @yurishkuro 

zipkin traces now appear as services "zipkin-web" and "zipkin-query"

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
```diff
@@ -224,7 +224,7 @@ class AnormSpanStore(val db: DB, val openCon: Option[Connection] = None) extends
 
   override def getTraceIdsByAnnotation(serviceName: String, annotation: String, value: Option[ByteBuffer],
     endTs: Long, limit: Int): Future[Seq[IndexedTraceId]] = db.inNewThreadWithRecoverableRetry {
-    if ((Constants.CoreAnnotations ++ Constants.CoreAddress).contains(annotation) || endTs <= 0 || limit <= 0) {
```
codefromthecrypt, line 227: this was just wrong

zipkin-query/src/main/scala/com/twitter/zipkin/query/ZipkinQueryServerFactory.scala
```diff
@@ -36,7 +36,9 @@ trait ZipkinQueryServerFactory { self: App =>
     log: Logger = Logger.get("QueryService")
   ): ListeningServer = {
     val impl = new ThriftQueryService(spanStore, dependencyStore, queryServiceDurationBatchSize())
-    ThriftMux.serve(queryServicePort(), composeQueryService(impl, stats))
+    ThriftMux.server
+             .configured(param.Label("zipkin-query"))
```
codefromthecrypt, line 40: fancy syntax compliments of @kristofa!

zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
```diff
@@ -55,8 +57,22 @@ trait ZipkinWebFactory { self: App =>
   val queryLimit = flag("zipkin.web.query.limit", 10, "Default query limit for trace results")
   val environment = flag("zipkin.web.environmentName", "", "The name of the environment Zipkin is running in")
 
-  def newQueryClient() = Thrift.newIface[ZipkinQuery.FutureIface]("ZipkinQuery=" + queryDest())
-  def newDependencySource() = Thrift.newIface[DependencySource.FutureIface]("DependencySource=" + queryDest())
+  // If a scribe host is configured, send all traces to it, otherwise disable tracing
+  val scribeHost = sys.env.get("SCRIBE_HOST")
+  val scribePort = sys.env.get("SCRIBE_PORT")
+  DefaultTracer.self = if (scribeHost.isDefined || scribePort.isDefined) {
```
codefromthecrypt, line 63: This is the best way to make sure "all tracing" gets affected. Otherwise, it is whack-a-mole where some use builders and others Thrift/Httpx.foo

codefromthecrypt:
Tested with:

``` bash
$ bin/collector
$ SCRIBE_PORT=9410 bin/query
$ SCRIBE_PORT=9410 bin/web
$ curl -s localhost:8080/api/services
# now see services pop up!
$ curl -s localhost:8080/api/services
["zipkin-query","zipkin-web"]
```

kristofa:
I was not familiar with the `DefaultTracer.self` approach. Using the `NullTracer` is also better than having sample rate = 0.  :+1:

zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
yurishkuro, line 63: @adriancole note that even though we can override the tracer this way, the `DefaultTracer` still instantiates some default ones via

```
  private[this] val tracers = LoadService[Tracer]()
```

zipkin-query-service/src/main/scala/com/twitter/zipkin/builder/QueryServiceBuilder.scala
```diff
@@ -29,6 +31,15 @@ case class QueryServiceBuilder(
     serverBuilder.apply().apply(runtime)
     val store = storeBuilder.apply()
 
+    // If a scribe host is configured, send all traces to it, otherwise disable tracing
+    val scribeHost = sys.env.get("SCRIBE_HOST")
```
codefromthecrypt, line 35: @yurishkuro here's what to reproduce in custom code

PR #714, opened 2015-09-16
Fixes QUERY_PORT and adds dev configuration to set it

codefromthecrypt:
Thanks to @wadey for catching this.

see #659


zipkin-query-service/src/main/scala/com/twitter/zipkin/builder/QueryServiceBuilder.scala
```diff
@@ -41,9 +41,7 @@ case class QueryServiceBuilder(
     }
 
     object UseOnceFactory extends com.twitter.app.App with ZipkinQueryServerFactory
-    UseOnceFactory.nonExitingMain(Array(
-      "zipkin.queryService.port", serverBuilder.serverAddress + ":" + serverBuilder.serverPort
```
codefromthecrypt, line 45: missing `.getHostAddress`, but also awkard way to set the value.. below is less awkward

PR #716, opened 2015-09-16
Work around Travis builds timing out

abesto:


codefromthecrypt:
neat idea

jamescway:
I guess all travis wants is some output..?

codefromthecrypt:
looks like it is worth a shot. thanks!

PR #717, opened 2015-09-17
Drops support for annotation.duration

codefromthecrypt:
This drops annotation.duration, an optional field rarely used in zipkin.
It causes more problems than it solves.

Implementations handled `annotation.duration` in many ways. Some advised
against using it (circonus), others exposed hooks not used (brave), some
stored the result of timestamp subtraction (spring-cloud-sleuth), or put
constants like 1 (htrace), yet others stored bugs (wrong timeunit in
anormdb). Many didn't support it at all, like ruby and python tracers.

Internally, zipkin has a lot of logic around annotation durations, but
it does timestamp math, in other words, it ignores this value.

On the front end, `Annotation.duration` has space dedicated to it in
the UI, which is often blank. It is not queryable and actually was used
to set an incorrect width.

I could go on, but bottom-line is that this doesn't add value to the
zipkin experience, rather it subtracts from it. Dropping this field is
safe as it was an optional thrift field to begin with, and it wasn't
useful as implemented.

https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/filter/JvmFilter.scala#L23
https://github.com/apache/incubator-htrace/blob/master/htrace-zipkin/src/main/java/org/apache/htrace/zipkin/HTraceToZipkinConverter.java#L190
https://github.com/circonus-labs/libmtev/blob/master/src/utils/mtev_zipkin.h#L215
https://github.com/wadey/go-zipkin/blob/master/trace.go#L144
https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinSpanListener.java#L205
https://github.com/openzipkin/brave/blob/master/brave-core/src/main/java/com/github/kristofa/brave/AnnotationSubmitter.java#L42

The only use of `Annotation.duration` I could find that wasn't derived
from other span data was finagle's gc annotation. Finagle's GC is a neat
annotation, but better stored as a binary one, instead of a semi-
structured string like this:

```
value=Gc(10,0.PSScavenge,2015-09-17 12:57:12 +0000,525.milliseconds+970.microseconds)
duration=525970
```

Note that even when we remove this, the duration is still present in the
string, in millis.

For example, if it were a binary annotation, you could store the entire
GC object as a json, and use the key `jvm.gc` instead. This is easier
than parsing the toString of a scala object.

### DIFF
--- a/zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
+++ b/zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
@@ -40,7 +40,6 @@ struct Annotation {
   1: i64 timestamp                 # microseconds from epoch
   2: string value                  # what happened at the timestamp?
   3: optional Endpoint host        # host this happened on
-  4: optional i32 duration         # how long did the operation take? microseconds
 }
 
 enum AnnotationType { BOOL, BYTES, I16, I32, I64, DOUBLE, STRING }
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 #Sat, 15 Aug 2015 15:15:18 +0000
-version=1.8.2-SNAPSHOT
+version=1.9.0-SNAPSHOT
 group=io.zipkin
 repo=https://github.com/openzipkin/zipkin
 description=A distributed tracing system
--- a/zipkin-anormdb/src/main/resources/mysql.sql
+++ b/zipkin-anormdb/src/main/resources/mysql.sql
@@ -4,7 +4,6 @@ CREATE TABLE IF NOT EXISTS zipkin_spans (
   trace_id BIGINT NOT NULL,
   span_name VARCHAR(255) NOT NULL,
   debug SMALLINT NOT NULL,
-  duration BIGINT,
   created_ts BIGINT
 );
 
@@ -21,8 +20,7 @@ CREATE TABLE IF NOT EXISTS zipkin_annotations (
   value TEXT,
   ipv4 INT,
   port INT,
-  a_timestamp BIGINT NOT NULL,
-  duration BIGINT
+  a_timestamp BIGINT NOT NULL
 );
 
 ALTER TABLE zipkin_annotations ADD FOREIGN KEY(span_id) REFERENCES zipkin_spans(span_id) ON DELETE CASCADE;
--- a/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
+++ b/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
@@ -44,16 +44,15 @@ class AnormSpanStore(val db: DB, val openCon: Option[Connection] = None) extends
         SQL(
           db.getSpanInsertCommand() +
             """ INTO zipkin_spans
-              |  (span_id, parent_id, trace_id, span_name, debug, duration, created_ts)
+              |  (span_id, parent_id, trace_id, span_name, debug, created_ts)
               |VALUES
-              |  ({span_id}, {parent_id}, {trace_id}, {span_name}, {debug}, {duration}, {created_ts})
+              |  ({span_id}, {parent_id}, {trace_id}, {span_name}, {debug}, {created_ts})
             """.stripMargin)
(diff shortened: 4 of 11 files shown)


codefromthecrypt:
Note this bumps version

codefromthecrypt:
cc @spencergibb @wadey @kristofa

codefromthecrypt:
Here's some feedback on this topic.

From @kristofa 

```
About duration for an annotation. This can be used to track duration of certain parts of your app. For example I had the idea of submitting an annotation with duration for measuring Json serialisation time. The annotation is displayed in Zipkin-web on the Span at its time stamp. Unfortunately the actual duration is not displayed in Zipkin-web. That's why in Brave I added the duration, if specified, to the annotation name.
```

From @yurishkuro 

```
I had a meeting today with our mobile team, they are interested in using Zipkin tracing for measuring timing of various functions within the app, ranging across time to open a controller, time to resolve DNS, time to call a service, etc. Not all of these use cases fit directly into Zipkin RPC model, some are more like stages of an operation rather than sub-spans. An annotation's duration might be an interesting one to use.
```

I'd suggest that if folks want to use duration, then let's evaluate it against real use cases. You may find overlaps with sub-service spans, normal metrics, or even find that duration would work better on BinaryAnnotation (which has both key and value). Duration certainly does seem handy, but we do need to face the facts that this is a broken and confusing part of zipkin.

Leaving something handy, but broke in zipkin is worse than removing it, IMHO. Decent use cases, tests and documentation would be the only way we could prevent repeating status quo, which is broken and has led to curious implementations and bugs in an unusable feature.

codefromthecrypt:
Thinking about alternatives here, too. One thing we've tossed around is having a [common data model for annotations](https://docs.google.com/document/d/1pxZ5NQHftJzhlCIT4vMfOT6QT_StikaoDc_NCdfesoA/edit). Let's take the case of GC, for example.

GC happens to be something that isn't caused by the span, it is metadata present, which could help explain timing, but it could be far longer than the span itself and hint at a visualization concern. Nevertheless, let's consider how we might do this.

As opposed to an `Annotation`, which has a timestamp, and a name, but no value like this:

```
value=Gc(10,0.PSScavenge,2015-09-17 12:57:12 +0000,525.milliseconds+970.microseconds)
duration=525970
```

What if this were a `BinaryAnnotation` which doesn't have a timestamp, but does have a value:

```
key=gc.durationMillis
type=I32
value=525970
```

Speaking with @dsyer here, this seems to make more sense, as you can query against it, and its namespace explains itself. Moreover, it is ancillary information to explain the span (ex. it isn't a sub-span, or an activity directly caused by it)

codefromthecrypt:
Merging as I think @kristofa and @yurishkuro are on board:

@kristofa 

```
Yes. I agree. You can also track such metrics with a regular monitoring system and graph them, calculate percentiles etc.
```

@yurishkuro

```
+1 for local span. As alternative to duration, I think it's more structured, and can be rendered nicer. ... kill the durations, kill zem all
```

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
```diff
@@ -40,7 +40,6 @@ struct Annotation {
   1: i64 timestamp                 # microseconds from epoch
   2: string value                  # what happened at the timestamp?
   3: optional Endpoint host        # host this happened on
-  4: optional i32 duration         # how long did the operation take? microseconds
```
yurishkuro, line 43: can we keep it here and just document as deprecated?  I'm concerned this would make historical data in Cassandra incompatible.

codefromthecrypt, line 43: It is optional, so the reader will skip it.. One good thing about thrift
and protobuf ;) that said, I should have left it commented out so we don't
re-use that number for a different purpose. Will add that.

codefromthecrypt, line 43: https://github.com/openzipkin/zipkin/commit/daebe8eb843d942bd49eed6f80889cd178e8384d

codefromthecrypt:
See also https://issues.apache.org/jira/browse/HTRACE-260

PR #720, opened 2015-09-21
Compresses MySQL tables and shrinks binaryannotations.value column

codefromthecrypt:
This adjusts the DDL commands for MySQL to use compression. It also
makes the binary annotation value column smaller, as 16MB was way too
large. Finally, this drops key references as they weren't helping anything.

Tested on local MySQL and also MySQL 5.6 on Amazon RDS


codefromthecrypt:
cc @dsyer @jfeltesse-mdsol

spencergibb:
:+1:

PR #724, opened 2015-09-23
Removes strange edge case when serviceName='client'

codefromthecrypt:
There was awkward code that special-cased the service name of "client". This was carried from cassie into its replacement, but not the other span stores. Better to remove this sort of cognitive hiccup than keep passing it along.


yurishkuro:
+1

PR #725, opened 2015-09-23
use different event names for parent/child popup, they were clashing …

yurishkuro:
…with the main load event


codefromthecrypt:
LGTM. thanks!

PR #726, opened 2015-09-23
Simplifies dependency graph by only focusing on callCount

codefromthecrypt:
One of the challenges of resurrecting the Dependency Graph feature has
been the incidental complexity of Algebird/Moments. The moments api
obfuscated otherise simple relationships such as how many calls are made
between links.

This removes Algebird from the data model and flattens code and schema
around it. This will significantly ease our process to apply dependency
graph to all span stores, while not excluding more advanced analysis
from occurring in the future.


codefromthecrypt:
screen shot 2015-09-23 at 1 04 59 pm

Still works! i.e. the darkness of the line corresponds to call count.

gradle.properties
```diff
@@ -1,5 +1,5 @@
 #Sat, 15 Aug 2015 15:15:18 +0000
-version=1.9.1-SNAPSHOT
+version=1.10.0-SNAPSHOT
```
codefromthecrypt, line 2: bumped as this is a schema affecting change

spencergibb:
Nice, that removes a ton of complexity. :+1:

kristofa:
Besides the build that is not happy (not sure if related to change or flaky test) :+1:

yurishkuro:
LGTM

PR #728, opened 2015-09-25
Converts the communication between web and query to http

codefromthecrypt:
Formerly, Zipkin's UI communicated with the query server via thrift RPC.
This means that any implementation that supports the UI needs to run a
Thrift listener. There's many problems with this.
- Thrift RPC isn't supported in most frameworks, hence a learning curve
- Thrift RPC isn't supported by loadbalancers, so you have to go TCP
- The only thrift RPC tracing implementation is custom to finagle
- Finagle has their own TCP multiplexer, which isn't supported elsewhere

Worst than all of this is that the choice is a poor match to power a UI.
zipkin-web literally converts these calls to http+json, and there's no
other consumer of the query api. Moving this link to normal HTTP/json
will make Zipkin easier to maintain, interop with and troubleshoot.


lycplus:
Can I select  traces by specifying a time interval ?

eirslett:
Great work! :)

zipkin-query/src/main/scala/com/twitter/zipkin/query/ZipkinQueryController.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  get("/api/v1/traces") { request: Request =>
+    queryExtractor(request) match {
+      case Some(qr) => traceIds(qr).flatMap(getTraces(_, qr.adjustClockSkew))
+      case None => Future.value(response.badRequest())
+    }
+  }
+
+  get("/api/v1/trace/:id") { request: GetTraceRequest =>
+    getTraces(Seq(request.id.toLong), request.adjust_clock_skew).map(_.headOption)
```
yurishkuro: Would be nice to separate HTTP API concerns from the business logic into different classes, to allow different implementations of the end-point, possibly with alternative (non-http) protocols / discovery mechanisms.

codefromthecrypt:
@liyichao this keeps the same query capabilities as exist right now. That said, we should be able to change that more easily in the future with a json api (which actually has tests).

I saw from gitter you'd like below.. I'll use this syntax for what's currently supported!

```
params:

start_time -- optional: default to include all traces
end_time -- optional: default to now
annotation -- optional: annotation query
service_name -- optional
method_name -- optional
```

adriancole:
status: rebased on top of some bugs around ordering some cracks in the dependency store api. will polish more tomorrow

codefromthecrypt:
@liyichao FYI I've left the query syntax as-is.. just changed the plumbing from underneath it. Once this is in, let's have the conversation and/or feel free to raise a PR to revise the api.

codefromthecrypt:
PTAL I think this is a good place to progress from

zipkin-query/src/main/scala/com/twitter/zipkin/query/ZipkinQueryController.scala
codefromthecrypt: @yurishkuro so api/v1 is basically moving the existing functionality over to a more supportable framework. I don't think we are ready to do multiple protocols as there's still lots of tech debt left (ex there's still ostrich in the all of the services, and the config story is awful). My take is that get the project into good shape before going towards flexibility. Ex. let's prove we can cleanly swap storage or the known listener concerns (ex kafka vs scribe) before we start optimizing for multiple protocols for a query server.

codefromthecrypt: All that said.. I agree the code isn't well organized, so feel free to send a refactor idea/PR to help with this.. I'd love to have help on the tech debt side.

codefromthecrypt:
summary of the implementation is that we moved the json-converting queries from zipkin-web to zipkin-query and forwarded ones from zipkin-dependencies (as the javascript calls the api directly). The framework used is Finatra 2, which is far more supportable than Ostrich.

I've ideas for follow-up work, including moving the web server also to Finatra and/or replacing it with a pure javascript app, which could be hosted on the query server (and drop the need for zipkin-web entirely). We can also look at removing ostrich completely from zipkin-query-service, porting the query config scala files to finatra/twitter-inject modules. I hope that this work will help others build on zipkin, and help us finish the tech debt clearing.

eirslett:
Zipkin-web should definitely be a javascript-only app, either hosted from zipkin-query or on top of an http server like apache/nginx!

spencergibb:
:+1:

PR #729, opened 2015-09-25
Clarifies contract of DependencyStore and backfills tests

codefromthecrypt:
This standardizing on micros for dependency start/end interval and
backfills tests in a way that doesn't flake on timezones. It also
clarifies docs and converges the DependencySource/Sink as prior to our
move to json.

In the near future, we are going to migrate from Thrift to Json api.
This backfills tests and revises the contract around dependencies so
that future work isn't confusing or difficult to implement portably.

Dependencies implementations were confusing, using a combination of
microseconds, dates, milliseconds, Time, Duration and Calendar
instances to do the job. There were also incorrect docs, confusing
which of startTime or endTime was the default anchor of the interval.

This was very difficult to understand and easy to break, as there were
no tests that cover, for example, what the result value should be. It
was also difficult to understand the difference between the Dependencies
api and limitations of the Cassandra implementation.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 #Sat, 15 Aug 2015 15:15:18 +0000
-version=1.10.1-SNAPSHOT
+version=1.11.0-SNAPSHOT
 group=io.zipkin
 repo=https://github.com/openzipkin/zipkin
 description=A distributed tracing system
--- a/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormDependencyStore.scala
+++ b/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormDependencyStore.scala
@@ -17,13 +17,13 @@
 package com.twitter.zipkin.storage.anormdb
 
 import com.twitter.util.{Future, Time}
-import com.twitter.conversions.time._
-import com.twitter.zipkin.common.{DependencyLink, Dependencies}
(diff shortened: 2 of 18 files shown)


gradle.properties
```diff
@@ -1,5 +1,5 @@
 #Sat, 15 Aug 2015 15:15:18 +0000
-version=1.10.1-SNAPSHOT
+version=1.11.0-SNAPSHOT
```
codefromthecrypt, line 2: because the thrift service definition change. eventhough we soon will move this out of thrift.

codefromthecrypt:
verified the UI operates with no impact

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormDependencyStore.scala
```diff
@@ -35,34 +35,44 @@ case class AnormDependencyStore(
   val db: DB,
   val openCon: Option[Connection] = None) extends DependencyStore with DBPool {
 
-  /**
-   * Get the dependencies in a time range.
-   *
-   * endDate is optional and if not passed defaults to startDate plus one day.
```
codefromthecrypt, line 41: total lie

codefromthecrypt:
going to merge on green as dependency store is not wholly implemented, yet, anyway due to https://github.com/openzipkin/zipkin-dependencies-hadoop/issues/2

PR #730, opened 2015-09-25
Squashes Span/Annotation ordering logic

codefromthecrypt:
Spans and annotation orderering were handled differently, and sometimes
(like redis) in opposite order. This fully specifies and consolidates
ordering logic, as well backfilling a test for out-of-order spans in a
trace.


zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -22,7 +22,7 @@
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Date;
-import java.util.HashMap;
+import java.util.LinkedHashMap;
```
codefromthecrypt, line 25: linked hashmap retains order

PR #731, opened 2015-09-26
Modify kafka defaults to be more inline with defaults

danchia:
I changed the values to be more inline with the defaults of https://kafka.apache.org/08/configuration.html

In particular, having `auto.commit.interval.ms` be 10ms is really dangerous as a default, as it will cause very high ZK loads. We actually observed this in production.


codefromthecrypt:
Yeah this makes loads of sense. cc @jamescway @clehene

codefromthecrypt:
Thanks!

jamescway:
thanks!

Issue #706, opened 2015-09-14
Unable to find compaction strategy class 'org.apache.cassandra.db.compaction.DateTieredCompactionStrategy


lycplus:
I set CASSANDRA_CONTACT_POINTS to nodes that are up
it connects and fails like this:

```
Sep 14, 2015 6:17:52 PM com.twitter.zipkin.collector.Main$ main
INFO: Loading configuration
log4j:WARN No appenders could be found for logger (com.datastax.driver.core.SystemProperties).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
INF [20150914-18:17:57.100] stats: Starting LatchedStatsListener
700 [20150914-18:17:57.212] net: HttpServer created http 0.0.0.0/0.0.0.0:9900
700 [20150914-18:17:57.219] net: context created: /
700 [20150914-18:17:57.220] net: context created: /report/
700 [20150914-18:17:57.221] net: context created: /favicon.ico
700 [20150914-18:17:57.221] net: context created: /static
700 [20150914-18:17:57.222] net: context created: /pprof/heap
700 [20150914-18:17:57.224] net: context created: /pprof/profile
700 [20150914-18:17:57.224] net: context created: /pprof/contention
700 [20150914-18:17:57.225] net: context created: /tracing
700 [20150914-18:17:57.226] net: context created: /admin/registry
700 [20150914-18:17:57.226] net: context created: /health
700 [20150914-18:17:57.226] net: context created: /quitquitquit
700 [20150914-18:17:57.227] net: context created: /abortabortabort
700 [20150914-18:17:57.231] net: context created: /graph/
700 [20150914-18:17:57.232] net: context created: /graph_data
INF [20150914-18:17:57.233] admin: Starting TimeSeriesCollector
INF [20150914-18:17:57.234] admin: Admin HTTP interface started on port 9900.
INF [20150914-18:17:57.235] builder: Building store: <function0>
FAT [20150914-18:17:57.991] concurrent: input future failed.
FAT [20150914-18:17:57.991] concurrent: com.datastax.driver.core.TransportException: [/bad_cassandra_ip:9042] Cannot connect
FAT [20150914-18:17:57.991] concurrent:     at com.datastax.driver.core.Connection$1.operationComplete(Connection.java:156)
[...]

PR #734, opened 2015-09-27
Cassandra tests flake less when cassandra isn't started on demand

codefromthecrypt:
Note: this skips cassandra tests if you aren't running locally. They will always run on pull-request, though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: java
 sudo: false
 
 services:
-  - cassandra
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
this didn't improve things, so rolling it back. Because #706, we'd have to install a custom cassandra. Even after that, the flakes persist. Maybe there's some way to get cassandra to have a higher consistency model when running in single-node mode for tests.

PR #735, opened 2015-09-27
Add content type for Mustache

danchia:
I've noticed the HTML pages for zipkin aren't compressed.

It turns out finagle has auto-gzip, but only applies it to content types it believes are text. Adding the missing content-type header to Mustache generated responses fixes this.

In practices, this has huge improvements in page delivery time when requesting many traces.


codefromthecrypt:
nice one!

PR #736, opened 2015-09-27
Allow partial maches in span names

danchia:
Small usability improvement.

Our span names tend to be pretty long - fully qualified java class names + method.
This tweak allows for search to look for matches contained with the span name.


codefromthecrypt:
Makes sense.. I forgot we have this advantage as we have a separate call to get span names.

PR #739, opened 2015-09-27
use bigint for ts instead of timestamp in cassandra

### LINKED ISSUE #737: New Cassandra store stores invalid timestamps
Zipkin uses microsecond resolution timestamps.

However, we attempt to store these microsecond timestamps directly into a `timestamp` column in C*, which assumes millsecond resolution. In my opinion this is misleading to people.
Additionally, in the newer cassandra `cqlsh` shell, this breaks the ability the query the column.

Suggested solution is to use `long` type for the `ts` column in C*.

Unfortunately, this is not a compatible change - tables would have to be dropped and re-created.

codefromthecrypt:
ts are microsecond resolution, whereas cassandra timestamps should be millisecond

Fixes #737 thanks to @danchia

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 #Sat, 15 Aug 2015 15:15:18 +0000
-version=1.11.1-SNAPSHOT
+version=1.12.0-SNAPSHOT
 group=io.zipkin
 repo=https://github.com/openzipkin/zipkin
 description=A distributed tracing system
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
the reason I think it is best to merge this is because:
- It is is of least surprise, as it matches all other impls
- It draws into focus where dates are actually used: dependencies
- timestamp type in CQL doesn't match SQL default microsecond precision

michaelsembwever:
:+1:

PR #740, opened 2015-09-27
Javac 7 friendly ordering

codefromthecrypt:
Latest polishing on ordering, does the following:
- Uses ordering friendly with javac 7 (implicit ordering broke someone)
- Removes explicit instructions for ordering spans
- Consolidates "List of Spans" ordering to first span vs timestamp
  - First span in a sorted list is already first timestamp
  - This avoids redundant searching through annotations

### DIFF
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/common/Annotation.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/common/Annotation.scala
@@ -16,13 +16,17 @@
  */
 package com.twitter.zipkin.common
 
+import com.google.common.collect.ComparisonChain
+import com.google.common.collect.Ordering.natural
+import java.util.Comparator
+
 /**
  * @param timestamp when was this annotation created? microseconds from epoch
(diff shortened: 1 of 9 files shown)


codefromthecrypt:
@rjbhewei this should fix your javac 7 thing.. I compiled locally on javac 7 w/o failure

zipkin-common/src/main/scala/com/twitter/zipkin/common/Annotation.scala
```diff
@@ -30,8 +34,11 @@ case class Annotation(timestamp: Long, value: String, host: Option[Endpoint])
    */
   def -(annotation: Annotation): Long = timestamp - annotation.timestamp
 
-  import scala.math.Ordered.orderingToOrdered
+  private[this] val nullsFirst: Comparator[Endpoint] = natural().nullsFirst()
 
-  def compare(that: Annotation): Int =
```
codefromthecrypt, line 35: this broke javac 7

codefromthecrypt:
we've redundant checks on travis. the first failed due to maven.twttr.biz flaking out, I consider this a pass.

PR #741, opened 2015-09-27
Reduces core dependencies to dodge conflicts with Finatra

codefromthecrypt:
Finatra has a lot of dependencies that don't match ostrich. This rids
all non-core dependencies and pushes those from ostrich as far to the
edge as possible. After this, the move to Finatra will be easier.


zipkin-collector/src/main/scala/com/twitter/zipkin/collector/WriteQueue.scala
```diff
@@ -1,67 +0,0 @@
...
- *
- */
-package com.twitter.zipkin.collector
-
-import com.twitter.ostrich.admin.{Service => OService}
-import com.twitter.ostrich.stats.Stats
-import java.util.concurrent.ArrayBlockingQueue
-import com.twitter.finagle.Service
-
-class WriteQueue[T](writeQueueMaxSize: Int,
```
codefromthecrypt, line 24: orphaned file

zipkin-collector/src/main/scala/com/twitter/zipkin/collector/WriteQueueWorker.scala
```diff
@@ -1,37 +0,0 @@
...
- *
- */
-package com.twitter.zipkin.collector
-
-import com.twitter.finagle.Service
-import com.twitter.ostrich.admin.BackgroundProcess
-import java.util.concurrent.{TimeUnit, BlockingQueue}
-import com.twitter.util.Await
-
-class WriteQueueWorker[T](queue: BlockingQueue[T],
```
codefromthecrypt, line 24: orphaned file

zipkin-collector/src/main/scala/com/twitter/zipkin/collector/builder/CollectorInterface.scala
```diff
@@ -1,24 +0,0 @@
...
-import com.twitter.zipkin.collector.WriteQueue
-import com.twitter.zipkin.common.Span
-import com.twitter.zipkin.storage.Store
-import java.net.InetSocketAddress
-
-/**
- * Specifies a builder for the input interface of a Zipkin collector
- * @tparam T
- */
-trait CollectorInterface[T]
```
codefromthecrypt, line 17: orphaned file

zipkin-common/src/main/scala/com/twitter/zipkin/common/Dependencies.scala
```diff
@@ -16,10 +16,8 @@
  */
 package com.twitter.zipkin.common
 
-import com.twitter.algebird.{Monoid, Semigroup}
```
codefromthecrypt, line 19: The duties of this class are too simple to depend on algebird. Algebird shouldn't be a core dep, and now isn't.

codefromthecrypt:
tested all backends: cassandra, dev, mysql, redis

Issue #698, opened 2015-09-08
DB error when trying to display 1000 items in the UI


jcarres-mdsol:
To reproduce, in the UI choose Limit to be 1000  and click on Find traces

I get this:

```
com.datastax.driver.core.exceptions.InvalidQueryException: Cannot page queries with both ORDER BY and a IN restriction on the partition key; you must either remove the ORDER BY or the IN and sort client side, or disable paging for this query


com.twitter.zipkin.thriftscala.QueryException$.decode(QueryException.scala:104)
com.twitter.zipkin.thriftscala.ZipkinQuery$GetTraceIds$Result$.readQeValue(ZipkinQuery.scala:483)
com.twitter.zipkin.thriftscala.ZipkinQuery$GetTraceIds$Result$.decode(ZipkinQuery.scala:426)
com.twitter.zipkin.thriftscala.ZipkinQuery$GetTraceIds$Result$.decode(ZipkinQuery.scala:332)
com.twitter.zipkin.thriftscala.ZipkinQuery$FinagleClient.decodeResponse(ZipkinQuery$FinagleClient.scala:58)
com.twitter.zipkin.thriftscala.ZipkinQuery$FinagleClient$$anonfun$getTraceIds$1.apply(ZipkinQuery$FinagleClient.scala:96)
com.twitter.zipkin.thriftscala.ZipkinQuery$FinagleClient$$anonfun$getTraceIds$1.apply(ZipkinQuery$FinagleClient.scala:95)
com.twitter.util.Future$$anonfun$flatMap$1.apply(Future.scala:890)
com.twitter.util.Future$$anonfun$flatMap$1.apply(Future.scala:889)
com.twitter.util.Promise$Transformer.liftedTree1$1(Promise.scala:100)
com.twitter.util.Promise$Transformer.k(Promise.scala:100)
com.twitter.util.Promise$Transformer.apply(Promise.scala:110)
com.twitter.util.Promise$Transformer.apply(Promise.scala:91)
com.twitter.util.Promise$$anon$2.run(Promise.scala:358)
com.twitter.concurrent.LocalScheduler$Activation.run(Scheduler.scala:193)
com.twitter.concurrent.LocalScheduler$Activation.submit(Scheduler.scala:140)
com.twitter.concurrent.LocalScheduler.submit(Scheduler.scala:222)
com.twitter.concurrent.Scheduler$.submit(Scheduler.scala:84)
com.twitter.util.Promise.runq(Promise.scala:342)
[...]

PR #742, opened 2015-09-28
DB error when trying to display 1000 items in the UI #698

michaelsembwever:
quick fix to enable limit on queries that use IN and ORDER BY (CASSANDRA-6722)

see https://github.com/openzipkin/zipkin/issues/698


codefromthecrypt:
cool. thanks!

PR #744, opened 2015-09-28
Reverts code/config that dodges java.beans

codefromthecrypt:
We were dodging access to the `java.beans` package to allow zipkin to
run on JRE compact3 profile, as the OS image becomes quite small. This
has become an unsustainable goal as too many libraries, now also Finatra
aren't designed for this. This rolls back things to not avoid
java.beans.


codefromthecrypt:
fyi logback 1.1.7 now supports compact3 profile

PR #745, opened 2015-09-29
Fixes logging for services; simplifies zipkin-query; defers c* i/o

codefromthecrypt:
The log level flags for all processes were broken, defaulting to DEBUG
level. This resurrects them by using logback directly.

This also squashes some indirection in the query service in preparation
of removing ostrich completely.

Finally, this defers network i/o caused by c*'s Repository ctor.


zipkin-query-service/config/query-cassandra.scala
```diff
@@ -37,18 +37,14 @@ if (username.isDefined && password.isDefined) {
...
-  handlers = List(ConsoleHandler())
-)
+val repository = new Repository(CassandraSpanStoreDefaults.KeyspaceName, cluster)
+val spanStore = new CassandraSpanStore(repository)
+val dependencies = new CassandraDependencyStore(repository)
 
 QueryServiceBuilder(
-  storeBuilder,
-  serverBuilder = ZipkinServerBuilder(serverPort, adminPort).loggers(List(loggerFactory))
+  "0.0.0.0:" + serverPort,
```
jamescway, line 41: [non-blocking] is there any significance to 0.0.0.0 in this case? is the intention the same as localhost?

jamescway:
:+1:  to :green_heart: build

zipkin-query-service/config/query-cassandra.scala
codefromthecrypt, line 41: This was making explicit what the previous value was. Neat, huh?

jamescway, line 41: cool!

codefromthecrypt:
red is because cassandra flaked again. the other green shows that this is mergable as these checks are redundant.

PR #749, opened 2015-09-30
Fixes span name selector

### LINKED ISSUE #748: span name selector UI broken
With the merge of the json api branch, the selector in zipkin-web is broken, because it loads data from `/api/spans?serviceName=xxx` (which returns a nested array of spans), and then tries to display it (as if it was a array of strings). 

Previously, the data was loaded from a span names endpoint, which used `getSpanNames` in the span store to retrieve a list of span names. This approach was superior, because it allows the span store to execute an efficient query to just aggregate all span names across the traces (the aggregates could even be stored separately).

codefromthecrypt:
The span name selector UI was calling getTraces instead of getSpanNames.
This happened during the switch to the json api. This change fixes that
and clarifies the confusing naming that was used in the web handlers.

Fixes #748


codefromthecrypt:
Pushing version 1.13.3 immediately. Will take an hour or so to reach docker

PR #750, opened 2015-09-30
Initial implementation of cached service names

mzagar:
- ui service name dropdown is populated using ajax call to /api/services endpoint using serviceName component
- http response to /api/services includes "Cache-Control: max-age=300, must revalidate" header which makes browser return cached values of service names
- browser cache for /api/services call expires after 5 minutes (300 seconds) by default
- flag "zipkin.web.serviceNameCacheTimeoutInSec" can be used to set this cache timeout to a custom value

### DIFF
--- a/zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
+++ b/zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
@@ -28,6 +28,7 @@ class Handlers(mustacheGenerator: ZipkinMustache, queryExtractor: QueryExtractor
     def apply(response: Response) {
       response.contentString = input.contentString
       response.statusCode = input.statusCode
+      input.headerMap foreach (e => response.headerMap.put(e._1,e._2))
     }
   }
 
@@ -222,8 +223,6 @@ class Handlers(mustacheGenerator: ZipkinMustache, queryExtractor: QueryExtractor
       val serviceName = req.params.get("serviceName").filterNot(_ == "")
       val spanName = req.params.get("spanName").filterNot(_ == "")
 
-      val servicesCall = client.executeJson[Seq[String]](Request(s"/api/v1/services"))
-
       val spansCall = serviceName match {
         case Some(service) => client.executeJson[Seq[String]](Request(s"/api/v1/spans?serviceName=${service}"))
         case None => EmptyStrings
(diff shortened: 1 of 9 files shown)


mzagar:
@adriancole please review the CachedRenderer 'thingy'... I was not really sure how to handle that one better (fighting my way through scala)

Optimization regarding service name count is not implemented yet.

Suggestions welcome!

codefromthecrypt:
Sure will take a look in a bit. Thanks for the help!

codefromthecrypt:
ps did you verify this works? I can regardless.

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
```diff
@@ -383,4 +385,11 @@ class Handlers(mustacheGenerator: ZipkinMustache, queryExtractor: QueryExtractor
           .map(renderTrace(_))
       } getOrElse NotFound
     }
+
+  def handleServices(client: HttpClient, webServiceNameCacheTimeoutInSec: Int): Service[Request, Renderer] =
```
codefromthecrypt: lets call this servicesMaxAge - much shorter and the max-age concept in http caching is already seconds. sg?

codefromthecrypt: actually I think this should move to the query server. You already have a response object, so it is less code. Then all you need to do is forward the request, like we do with `/api/spans`.

```
  get("/api/v1/services") { request: Request =>
    response.ok(spanStore.getAllServiceNames()).cacheControl(servicesMaxAge)
  }
```

codefromthecrypt: plus.. you can actually test it there. ex. in `ZipkinQueryServerFeatureTest` you can look for the default cache control header.

mzagar:
TODO: write zipkin-query test -- figure out how to setup controller with custom parameters so I can test Cache-Control behaviour on 3 test services...

mzagar:
@adriancole I'm stuck writing a test for /api/v1/services that sets Cache-Control header depending on number of services available (for < 5, no Cache-Control set, else cache it); please see my comments [here](https://github.com/openzipkin/zipkin/pull/750/files#diff-0949fd0d2250b984204c05eacbc88f3cR35) and [here](https://github.com/openzipkin/zipkin/pull/750/files#diff-77863e2473d7731918e6a960b0514a2cR87).

Can you help me and point me in the right direction please?

codefromthecrypt:
I don't think we want a configuration parameter for such an edge case as
max services before caching.  I literally meant to hard-code it to 3. That
should be easy to test as all you need to do is sink spans with more than 3
service names (endpoints) in the test.

mzagar:
Hopefully this should do it... ping me if anything else needs fixing. Thanks!

codefromthecrypt:
looks good..  mind squashing these commits?

mzagar:
will do...

codefromthecrypt:
closed via https://github.com/openzipkin/zipkin/commit/fd2cadf62fa1e01cc87b79a30f9d675e9e62eeee

Thanks!

PR #753, opened 2015-10-01
Routes ZK logging to slf4j

codefromthecrypt:
When people set flags that spin up ZK, this avoids a log configuration error.


codefromthecrypt:
This is not a blocker, but this will eliminate clutter in the console output when using kafka or the adaptive sampler

PR #754, opened 2015-10-01
Fixes tracegen against remote query servers

codefromthecrypt:
tracegen was evaluating the queryDest arg before it was parsed.

### DIFF
--- a/zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
+++ b/zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
@@ -73,8 +73,11 @@ trait ZipkinWebFactory { self: App =>
     NullTracer
   }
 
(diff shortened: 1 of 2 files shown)


zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
```diff
@@ -73,8 +73,11 @@ trait ZipkinWebFactory { self: App =>
     NullTracer
   }
 
-  /** Initialize a json-aware Finatra client, targeting the query host */
-  def newQueryClient() = new HttpClient(
+  /**
+   * Initialize a json-aware Finatra client, targeting the query host. Lazy to ensure
+   * we get the host after the [[queryDest]] flag has been parsed.
+   */
+  lazy val queryClient = new HttpClient(
```
codefromthecrypt, line 80: this wasn't a bug, just wanted to be consistent

PR #755, opened 2015-10-02
Merges SpanStore with the defunct Thrift query api and tightens contract

codefromthecrypt:
Before, there was lack of clarity around when spans were merged and how
the endTs should be handled. Moreover, SpanStore lacked the QueryRequest
object, which formerly existed in the thrift impl.

By moving the QueryRequest object to SpanStore and backfilling tests on
previously vague behavior, implementors have a better contract and can
make more efficient storage queries.

`CollectAnnotationQueries` is a crutch that can be used until folks are
able to refactor their storage backends.


zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
```diff
@@ -224,7 +224,7 @@ abstract class CassandraSpanStore(
       })
   }
 
-  override def getSpansByTraceIds(traceIds: Seq[Long]): Future[Seq[Seq[Span]]] = {
+  override def getTracesByIds(traceIds: Seq[Long]): Future[Seq[Seq[Span]]] = {
     QueryGetSpansByTraceIdsStat.add(traceIds.size)
```
yurishkuro, line 229: lest I forget - I found it useful to also measure latency, not just counts

zipkin-common/src/main/scala/com/twitter/zipkin/storage/QueryRequest.scala
```diff
@@ -0,0 +1,22 @@
+package com.twitter.zipkin.storage
+
+import com.twitter.util.Time
+
+/**
+ * @param serviceName Mandatory [[com.twitter.zipkin.common.Endpoint.serviceName]]
+ * @param spanName When present, only include traces with this [[com.twitter.zipkin.common.Span.name]]
+ * @param annotations Include  traces whose [[com.twitter.zipkin.common.Span.annotations]] value is in this list.
+ *                    This is an OR condition against [[binaryAnnotations]]
```
yurishkuro: what about within the collection, is it AND or also OR?

zipkin-common/src/main/scala/com/twitter/zipkin/storage/QueryRequest.scala
```diff
@@ -0,0 +1,22 @@
...
+ * @param binaryAnnotations Include traces whose [[com.twitter.zipkin.common.Span.binaryAnnotations]] include a
+ *                          String whose key and value are an entry in this map.
```
yurishkuro, line 11: given that it's also OR within the collection, a Map would be limiting struct, a `Set[(String, String)]` would be better

codefromthecrypt, line 11: good point.

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
codefromthecrypt, line 229: the size this was a carry-over from an old impl.. the size is always 1.

zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
```diff
@@ -15,50 +15,33 @@
...
    */
-  def getSpansByTraceIds(traceIds: Seq[Long]): Future[Seq[Seq[Span]]]
+  def getTraces(qr: QueryRequest): Future[Seq[Seq[Span]]]
 
   /**
-   * Get the trace ids for this particular service and if provided, span name.
-   * Only return maximum of limit trace ids from before the endTs.
+   * Get the available trace information from the storage system.
+   * Spans in trace are sorted by the first annotation timestamp
+   * in that span. First event should be first in the spans list.
```
yurishkuro, line 40: not clear on this last part - meaning that annotations will be sorted within each span?  Also, should this do a merge of spans with the same ID?  Otherwise it returns redundant data.

zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
```diff
@@ -76,6 +59,10 @@ abstract class SpanStore extends java.io.Closeable {
 
   /**
    * Store a list of spans, indexing as necessary.
+   *
+   * <p/> Spans may come in sparse, for example apply may be called multiple times
```
codefromthecrypt, line 64: @yurishkuro here's some clarity about the span merging

yurishkuro:
lgtm

codefromthecrypt:
ok updated, including backfilled tests for span merges and endTs

yurishkuro:
lgtm

PR #759, opened 2015-10-04
Settles on startTs/endTs and fixes Dependencies UI

codefromthecrypt:
Troubleshooting subtle problems was unnecessarily difficult as we used
many ways to refer to the start and end of a span in epoch microseconds.

This settles on `startTs` and `endTs`, which are the dominant names. By
doing so, it is cognitively easier as, for example, we can see that
`Span.endTs` relates directly to the query parameter `endTs`, as well as
the upper bound of `Dependencies.endTs`.

This bumps the version because it fixes the dependencies api endpoint to
use normal query parameters as opposed to path expressions for `startTs`
and `endTs`. This also renames the thrift fields for dependencies,
although that isn't breaking as names aren't encoded in thrift
serialization.

Most importantly, this fixes the Dependencies UI, which has been broken
for a while.

### DIFF
--- a/zipkin-common/src/test/scala/com/twitter/zipkin/storage/SpanStoreSpec.scala
+++ b/zipkin-common/src/test/scala/com/twitter/zipkin/storage/SpanStoreSpec.scala
@@ -120,8 +120,6 @@ abstract class SpanStoreSpec extends JUnitSuite with Matchers {
     result(store.getTraces(QueryRequest("service", annotations = Set("custom")))) should be(
       Seq(Seq(span1))
     )
-    // should not find any traces since the core annotation doesn't exist in index
-    result(store.getTraces(QueryRequest("service", annotations = Set("cs")))) should be(empty)
 
     // should find traces by the key and value annotation
     result(store.getTraces(QueryRequest("service", binaryAnnotations = Set(("BAH", "BEH"))))) should be(
--- a/zipkin-query/src/test/scala/com/twitter/zipkin/query/ZipkinQueryServerFeatureTest.scala
+++ b/zipkin-query/src/test/scala/com/twitter/zipkin/query/ZipkinQueryServerFeatureTest.scala
@@ -109,7 +109,6 @@ class ZipkinQueryServerFeatureTest extends FeatureTest with MockitoSugar with Be
     assert(response.headerMap.get("Cache-Control") == Some("max-age=300, must-revalidate"))
   }
 
-
   "get span names" in {
     app.injector.instance[SpanStore].apply(allSpans)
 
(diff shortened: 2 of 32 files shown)


zipkin-common/src/test/scala/com/twitter/zipkin/storage/SpanStoreSpec.scala
```diff
@@ -120,8 +120,6 @@ abstract class SpanStoreSpec extends JUnitSuite with Matchers {
     result(store.getTraces(QueryRequest("service", annotations = Set("custom")))) should be(
       Seq(Seq(span1))
     )
-    // should not find any traces since the core annotation doesn't exist in index
```
codefromthecrypt, line 123: goofy and completely unnecessary. Many indexes cannot be configured to exclude certain words, so this is essentially busy-work to enforce.

zipkin-query/src/test/scala/com/twitter/zipkin/query/ZipkinQueryServerFeatureTest.scala
```diff
@@ -131,6 +130,20 @@ class ZipkinQueryServerFeatureTest extends FeatureTest with MockitoSugar with Be
       withJsonBody = "[]")
   }
 
+  "get trace when bad limit" in {
+    server.httpGet(
+      path = "/api/v1/traces?serviceName=service1&limit=0",
+      andExpect = BadRequest,
+      withBody = "limit should be positive: was 0")
```
codefromthecrypt, line 137: look. now we have actual error messages!

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
```diff
@@ -169,22 +169,18 @@ class Handlers(mustacheGenerator: ZipkinMustache, queryExtractor: QueryExtractor
...
+      val (current, next) = stamps.partition { t => t.startTs >= ts.startTs && t.endTs <= ts.endTs }
+      val endTs = current.map(_.endTs).max
+      totalServiceTime(next, acc + (endTs - ts.startTs))
     }
 
   private[this] def traceSummaryToMustache(
     serviceName: Option[String],
     ts: Seq[TraceSummary]
   ): Map[String, Any] = {
-    val (maxDuration, minStartTime, maxStartTime) = ts.foldLeft((Long.MinValue, Long.MaxValue, Long.MinValue)) { case ((maxD, minST, maxST), t) =>
```
codefromthecrypt, line 183: the confusingly named minStartTime and maxStartTime were never used

PR #769, opened 2015-10-10
Flattens AnormDB Schema, enabling on-demand dependency view

codefromthecrypt:
This uses the schema zipkin-java has been using for the last month,
which allows on-demand processing of the dependency view. This means the
default dev configuration can show dependencies without any hadoop job
or database seeding. This also means `zipkin-java` and `zipkin-anorm`
can co-exist.

Here's an example result by running tracegen against the default zipkin servers:
screen shot 2015-10-10 at 2 00 47 pm


gradle.properties
```diff
@@ -1,5 +1,5 @@
 #Sat, 15 Aug 2015 15:15:18 +0000
-version=1.16.2-SNAPSHOT
+version=1.17.0-SNAPSHOT
```
codefromthecrypt, line 2: @jfeltesse-mdsol bumping the version as there's a schema change. Now's a good time to get docker working, since the schema can now support all functions in zipkin

PR #776, opened 2015-10-14
fix span name indexing

danchia:
TraceIds indexed by spanNames were not always showing up if your span name had capital letters. This fixes the ingestion to be consistent.

### DIFF
--- a/zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
+++ b/zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
@@ -114,7 +114,7 @@ abstract class CassandraSpanStore(
 
       Future.join(
         span.serviceNames.toSeq map { serviceName =>
-          FutureUtil.toFuture(repository.storeSpanName(serviceName, span.name.toLowerCase, indexTtl.inSeconds))
+          FutureUtil.toFuture(repository.storeSpanName(serviceName, span.name, indexTtl.inSeconds))
         })
     }
   }
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
@@ -18,6 +18,7 @@ package com.twitter.zipkin.common
 
 import com.twitter.zipkin.Constants
 import scala.collection.breakOut
+import scala.util.hashing.MurmurHash3
 
 /**
  * A span represents one RPC request. A trace is made up of many spans.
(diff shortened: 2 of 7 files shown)


codefromthecrypt:
mind adjusting and adding a test like this to SpanStoreSpec that shows what expectations are? I'm not saying that the below test is what we want, just an example of how to write one.

```
  @Test def spanNamesGoLowercase() {
    result(store(Seq(span1.copy(name = "Yak"))))

    result(store.getSpanNames("service")) should be(List("yak"))
    result(store.getTraces(QueryRequest("service", Some("Yak")))) should be(empty)
    result(store.getTraces(QueryRequest("service", Some("yak")))) should be(
      Seq(span1.copy(name = "yak"))
    )
  }
```

danchia:
Happy to do so :)

danchia:
@adriancole added a test as requested

codefromthecrypt:
Thanks! Lg on green

Nit:

Lets avoid forcing navigation to see what test input thats only used once is. I know many tests in zipkin are like that, but it is bad practice as it adds unnecessary and distracting indirection. In some cases, for example, I've found misnamed variables hiding bugs. Inlining eliminates that problem space.

codefromthecrypt:
To add more to my Nit (digging a deeper hole):

I should say inlining or locally declaring input where only used once. Ex. A misnamed variable is obvious when it is declared and used within the same few lines. Your variable is fine, just people do refactors and break things. Locals or inlining is great as say a future PR breaks this, you can usually see in the diff its context, whereas if the input is declared at the top of the file, you cant.

codefromthecrypt:
Lemme know if I lost you. Apologies if I over communicated. I don't mind
helping to get this merged.

danchia:
Whoops. @adriancole I've updated the diff, but forgot to ping you. I don't know why it's failing though. Can you trigger a retest? It works locally for me.

codefromthecrypt:
sure.. kicking now

codefromthecrypt:
sure you are pushed (and doing ./gradlew clean check from root)? test fail looks legit:

```
*** 1 TEST FAILED ***
AnormSpanStoreSpec:
- spanNamesGoLowercase *** FAILED ***
  org.scalatest.exceptions.TestFailedException: List("SPAN_NAME") was not equal to List("span_name") (SpanStoreSpec.scala:226)
```

danchia:
Ah good catch.. clean surfaced the failure. Silly incremental.
Fixing now.

danchia:
@adriancole I lowercased the spanName in the `Span` constructor to normalize is, as discussed in gitter.

codefromthecrypt:
Looks simple enough. Can you add a note to zipkinCore.thrift under span name that these are lowercase? I know it is separate issues, but extremely similar is the service name. Is that easy enough to address in this PR?

codefromthecrypt:
thanks tons

PR #778, opened 2015-10-16
Only start traces for dynamic content on zipkin-web

codefromthecrypt:
Before, zipkin-web had the default finagle behavior of tracing all
http requests. Even if we were sampling, this would create junk traces
as we'd get spans for static assets like icons and javascript files.
These extra traces make troubleshooting real problems more difficult, at
least due to the distraction of spans that share the same name as spans
of interest.

This hard-codes a filter, as Finagle doesn't provide a visible means to
filter tracing based on uri path.


codefromthecrypt:
@cacoco this was super unintuitive. Does finatra have a way to filter paths that are traced?

PR #784, opened 2015-10-21
Adds support for HTTP POST /api/v1/spans as an alternative to Scribe

codefromthecrypt:
It has come up regularly that people are sending spans directly to the
collector's scribe interface, not because they use scribe, but because
they want a simpler architecture when getting started with zipkin. For
example, they are not yet ready for a data pipeline and just want to run
without any dependencies.

While this works, scribe uses Thrift RPC, which is rarely supported, and
for example doesn't have load balancer support (except at TCP level) or
reverse proxies like HTTP does. The type of users who connect directly
to the Collector's scribe endpoint would be better off with an HTTP
option, this is that.

Details of the HTTP Api:

With no content type (or `application/json`) we accept JSON, and with
`content-type=application/x-thrift` we accept Thrift. In both cases
a List is POSTed to `/api/v1/spans` on the query service. The endpoint
returns immediately with a 202 or 400 if the input was malformed.

In either case, spans are written asynchronously to the span store.

Ex. from https://github.com/adrianco/spigo

``` bash
$ curl -s localhost:9411/api/v1/spans -X POST --data @json_metrics/cassandra_flow.json -H "Content-Type: application/json"
```

### DIFF
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/json/JsonSpan.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/json/JsonSpan.scala
@@ -4,14 +4,12 @@ import com.google.common.io.BaseEncoding
 import com.google.common.primitives.Longs
 import com.twitter.zipkin.common.Span
 
-import scala.util.control.NonFatal
-
 case class JsonSpan(traceId: String, // hex long
                     name: String,
                     id: String, // hex long
-                    parentId: Option[String], // hex long
-                    annotations: List[JsonAnnotation], // ordered by timestamp
-                    binaryAnnotations: Seq[JsonBinaryAnnotation],
+                    parentId: Option[String] = None, // hex long
+                    annotations: List[JsonAnnotation] = List.empty, // ordered by timestamp
+                    binaryAnnotations: Seq[JsonBinaryAnnotation] = Seq.empty,
                     debug: Option[Boolean] = None)
 
 object JsonSpan extends (Span => JsonSpan) {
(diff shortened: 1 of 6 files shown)


codefromthecrypt:
cc @adrianco I'll ping back when this is in a release

zipkin-common/src/main/scala/com/twitter/zipkin/json/JsonSpan.scala
```diff
@@ -30,19 +28,26 @@ object JsonSpan extends (Span => JsonSpan) {
...
+      val correctLength = new Array[Char](16)
+      java.util.Arrays.fill(correctLength, '0') // cause 0 in ASCII is NUL, not '0'
+      System.arraycopy(idInHex.toCharArray, 0, correctLength, 16 - idInHex.length, idInHex.length)
+      hex.decode(new String(correctLength))
+    } else {
+      hex.decode(idInHex)
+    }
     Longs.fromByteArray(array)
-  } catch {
-    case NonFatal(e) => 0L
```
codefromthecrypt, line 46: quietly failing to zero at this level of abstraction is a timebomb

zipkin-common/src/test/scala/com/twitter/zipkin/json/JsonConversionTest.scala
```diff
@@ -7,6 +7,19 @@ import com.twitter.zipkin.common.{Endpoint, AnnotationType, BinaryAnnotation}
 import org.scalatest.{FunSuite, Matchers}
 
 class JsonConversionTest extends FunSuite with Matchers {
+
+  test("span with null annotations, binaryAnnotations") {
```
codefromthecrypt, line 11: it is easier to completely replace the codec system than properly work around this in the version of jackson we are using.

zipkin-scrooge/src/main/scala/com/twitter/zipkin/conversions/thrift.scala
```diff
@@ -130,8 +130,8 @@ object thrift {
     lazy val toThrift = thriftscala.Dependencies(d.startTs, d.endTs, d.links.map(_.toThrift))
   }
   class ThriftDependencies(d: thriftscala.Dependencies) {
-    lazy val toDependencies = Dependencies(d.startTs, d.endTs, d.links.map(_.toDependencyLink))
+    lazy val toDependencies: Dependencies = Dependencies(d.startTs, d.endTs, d.links.map(_.toDependencyLink))
```
codefromthecrypt, line 133: something magical happened where after I added the scrooge dep to the query project, this started to fail compile unless I added a type hint

codefromthecrypt:
This is the scala implementation of https://github.com/openzipkin/zipkin-java/pull/35

codefromthecrypt:
docker release is out. if you clone https://github.com/openzipkin/docker-zipkin, you can do the same spigo instructions as above after running `docker-compose up`. Only change is that you need to use your docker-machine's IP as opposed to localhost to POST the spans.

PR #788, opened 2015-10-22
Eliminates case class tech debt in Span and clarifies debug is an option

codefromthecrypt:
Before, the common Span type was a strange trait, only used as a case
class. It mismatched the thrift Span struct as it treated debug as if it
weren't an option. This cleans up the tech debt and removes all passing
of `Nil` wrt instantiating a Span class. `Nil` should be very uncommon
in scala code.

### DIFF
--- a/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
+++ b/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
@@ -49,7 +49,7 @@ class AnormSpanStore(val db: DB,
           .on("id" -> span.id)
           .on("name" -> span.name)
           .on("parent_id" -> span.parentId)
-          .on("debug" -> (if (span.debug) 1 else 0))
+          .on("debug" -> span.debug.map(if (_) 1 else 0))
           .on("start_ts" -> span.startTs)
           .execute()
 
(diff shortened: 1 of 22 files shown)


zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
```diff
@@ -112,8 +112,8 @@ class AnormSpanStore(val db: DB,
             |ORDER BY start_ts
           """.stripMargin.format(traceIdsString))
           .as((long("id") ~ get[Option[Long]]("parent_id") ~
-          long("trace_id") ~ str("name") ~ int("debug") map {
-          case a~b~c~d~e => DBSpan(a, b, c, d, e > 0)
+          long("trace_id") ~ str("name") ~ get[Option[Int]]("debug") map {
```
codefromthecrypt, line 115: this type of bug was easy as Span lied about it not being optional

zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
```diff
@@ -28,46 +27,7 @@ import scala.collection.breakOut
  *
  * Some are created by users, describing application specific information,
  * such as cache hits/misses.
- */
-object Span {
-  // TODO(jeff): what?!
```
codefromthecrypt, line 33: good riddance crazy code

zipkin-redis/src/test/scala/com/twitter/zipkin/storage/redis/RedisSnappyThriftCodec.scala
```diff
@@ -7,18 +7,13 @@ import com.twitter.zipkin.thriftscala
 import com.twitter.zipkin.conversions.thrift._
 import com.twitter.zipkin.common.{Annotation, BinaryAnnotation, Span}
 
-
-/**
- * Tests the RedisSnappyThriftCodec
```
codefromthecrypt, line 12: pet peeve

PR #789, opened 2015-10-22
Reduces default database connections to 10; adds MYSQL_MAX_CONNECTIONS

### LINKED ISSUE #787: Too many connections to the mysql DB
After running the query and collector for a while I can see the mysql DB maxes out its connections

```
ERROR 1040 (08004): Too many connections
```

It seems our setting is 120 connections. I can see it slowly building up and never releasing even when there is no activity in query and probably almost none in the collector

codefromthecrypt:
Database tuning is site-specific. However, we should include smart
defaults. This moves from 32 to Hikari's default of 10 connections per
instance/pool. This adds a new env variable `MYSQL_MAX_CONNECTIONS` to
adjust that.

Fixes #787


codefromthecrypt:
cc @jcarres-mdsol

PR #790, opened 2015-10-23
Fixes service name resolution of dependency graph

### LINKED ISSUE #781: Dependencies graph is incorrect
I have a test setup where
traceProducerOne sleeps 0.5 then calls traceProducerTwo which sleeps 0.2 and calls traceProducerThree which sleeps 0.3

Clicking on the dependencies graph I would expect

traceProducerOne -> traceProducerTwo -> traceProducerThree

Instead I have the following graph:

screen shot 2015-10-18 at 12 20 56 pm

Which is incorrect as One never calls three directly.
When I click on traceProducerOne I get this
screen shot 2015-10-18 at 12 24 10 pm

Which is incorrect.

codefromthecrypt:
Before this change, the dependency graph would display the _first_
serviceName in a span, and use that as the parent name. The problem is
that there are multiple endpoints in a span.

For example, the endpoint associated with "sr" or "cs" is the local
server, in this case the server _or_ the client. So, just picking any
endpoint and using it as the service name will certainly be inaccurate.

Moreover, sometimes the server isn't instrumented, so in this case it
won't be able to log an "sr" annotation. Unless you look at the "sa" or
Service Addr annotation, you won't be able to name that span, either.

This fixes both of these problems, by considering only service names of
"sr" or "sa" when building dependency links.

This also backfills tests for the previous span name parsing behavior.

Fixes #781

### DIFF
--- a/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormDependencyStore.scala
+++ b/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormDependencyStore.scala
@@ -60,6 +60,7 @@ case class AnormDependencyStore(val db: DB,
         """SELECT DISTINCT trace_id, span_id, endpoint_service_name
           |FROM zipkin_annotations
           |WHERE trace_id IN (%s)
+          |AND a_key in ("sr","sa")
           |AND endpoint_service_name is not null
           |GROUP BY trace_id, span_id
         """.stripMargin.format(parentChild.keys.mkString(",")))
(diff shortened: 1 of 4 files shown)


yurishkuro:
when A->B and B is not instrumented, would you not want to use the name from "cs" emitted by A?

codefromthecrypt:
All these little things make logic more complex. It would be better if the
instrumentation was corrected to log the 'sa'

codefromthecrypt:
PS I don't mind raising a pull request to do that. Is the thing that's
causing the 'cs' open source?

codefromthecrypt:
PS also we will eventually need the 'sa' anyway, for example querying by
destination address. The address on 'cs' is the local one.

codefromthecrypt:
All of this reminds me to update the thrift and associated doc on those
core annotations. All of these rules are very subtle.

jcarres-mdsol:
I do not know what's 'sa' 
In ruby and C# we have the endpoint for cs cr set to the service we are calling. Not the service calling

PR #794, opened 2015-10-24
Adds documentation to annotations as they are often misunderstood

codefromthecrypt:
Instrumentation libraries almost always get some aspect of annotations
wrong. This is an attempt to document them, so that folks have a better
chance at creating portable implementations.

### DIFF
--- a/zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
+++ b/zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
@@ -15,22 +15,121 @@ namespace java com.twitter.zipkin.thriftjava
 #@namespace scala com.twitter.zipkin.thriftscala
 namespace rb Zipkin
 
-#************** Common annotation values **************
+#************** Annotation.value **************
+/*
+ * The client sent ("cs") a request to a server. There is only one send per
+ * span. For example, if there's a transport error, each attempt can be logged
+ * as a WIRE_SEND annotation.
+ *
+ * If chunking is involved, each chunk could be logged as a separate
+ * CLIENT_SEND_FRAGMENT in the same span.
+ *
+ * Annotation.host is not the server. It is the host which logged the send
+ * event, almost always the client. When logging CLIENT_SEND, instrumentation
+ * should also log the SERVER_ADDR.
+ */
 const string CLIENT_SEND = "cs"
+/*
+ * The client received ("cr") a response from a server. There is only one
+ * receive per span. For example, if duplicate responses were received, each
+ * can be logged as a WIRE_RECV annotation.
+ *
+ * If chunking is involved, each chunk could be logged as a separate
+ * CLIENT_RECV_FRAGMENT in the same span.
+ *
+ * Annotation.host is not the server. It is the host which logged the receive
+ * event, almost always the client. The actual endpoint of the server is
+ * recorded separately as SERVER_ADDR when CLIENT_SEND is logged.
+ */
 const string CLIENT_RECV = "cr"
+/*
+ * The server sent ("ss") a response to a client. There is only one response
+ * per span. If there's a transport error, each attempt can be logged as a
+ * WIRE_SEND annotation.
+ *
+ * Typically, a trace ends with a server send, so the last timestamp of a trace
+ * is often the timestamp of the root span's server send.
+ *
+ * If chunking is involved, each chunk could be logged as a separate
+ * SERVER_SEND_FRAGMENT in the same span.
+ *
+ * Annotation.host is not the client. It is the host which logged the send
+ * event, almost always the server. The actual endpoint of the client is
+ * recorded separately as CLIENT_ADDR when SERVER_RECV is logged.
+ */
 const string SERVER_SEND = "ss"
+/*
+ * The server received ("sr") a request from a client. There is only one
+ * request per span.  For example, if duplicate responses were received, each
+ * can be logged as a WIRE_RECV annotation.
+ *
+ * Typically, a trace starts with a server receive, so the first timestamp of a
+ * trace is often the timestamp of the root span's server receive.
+ *
+ * If chunking is involved, each chunk could be logged as a separate
+ * SERVER_RECV_FRAGMENT in the same span.
+ *
+ * Annotation.host is not the client. It is the host which logged the receive
+ * event, almost always the server. When logging SERVER_RECV, instrumentation
+ * should also log the CLIENT_ADDR.
+ */
 const string SERVER_RECV = "sr"
+/*
+ * Optionally logs an attempt to send a message on the wire. Multiple wire send
+ * events could indicate network retries. A lag between client or server send
+ * and wire send might indicate queuing or processing delay.
+ */
 const string WIRE_SEND = "ws"
+/*
+ * Optionally logs an attempt to receive a message from the wire. Multiple wire
+ * receive events could indicate network retries. A lag between wire receive
+ * and client or server receive might indicate queuing or processing delay.
+ */
 const string WIRE_RECV = "wr"
+/*
+ * Optionally logs progress of a (CLIENT_SEND, WIRE_SEND). For example, this
+ * could be one chunk in a chunked request.
+ */
+const string CLIENT_SEND_FRAGMENT = "csf"
+/*
+ * Optionally logs progress of a (CLIENT_RECV, WIRE_RECV). For example, this
+ * could be one chunk in a chunked response.
+ */
+const string CLIENT_RECV_FRAGMENT = "crf"
+/*
+ * Optionally logs progress of a (SERVER_SEND, WIRE_SEND). For example, this
+ * could be one chunk in a chunked response.
+ */
+const string SERVER_SEND_FRAGMENT = "ssf"
+/*
+ * Optionally logs progress of a (SERVER_RECV, WIRE_RECV). For example, this
+ * could be one chunk in a chunked request.
+ */
+const string SERVER_RECV_FRAGMENT = "srf"
 
-#************** Common binary annotation keys **************
-// The endpoint associated with CLIENT_ annotations is not necessarily Annotation.host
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
@jcarres-mdsol @yurishkuro @adrianco fyi

 @theatrus can you critique this, in case I misunderstood something?

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
```diff
@@ -15,18 +15,118 @@ namespace java com.twitter.zipkin.thriftjava
...
+ * receive per span. For example, if duplicate responses were received, each
+ * can be logged as a WIRE_RECV annotation.
+ *
+ * If chunking is involved, each chunk could be logged as a separate
+ * CLIENT_RECV_FRAGMENT in the same span.
+ *
+ * Annotation.host is not the server. It is the host which logged the receive
+ * event, almost always the client.
+ *
+ * To record the actual endpoints of the client and server side of the span,
```
yurishkuro: Perhaps the recommendation may be to leave endpoint logging to CS and SR events ("begin" events), not do it in the "end" events.

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
```diff
@@ -15,18 +15,118 @@ namespace java com.twitter.zipkin.thriftjava
...
+ * WIRE_SEND annotation.
+ *
+ * Typically, a trace ends with a server send, so the last timestamp of a trace
+ * is often the timestamp of the root span's server send.
+ *
+ * If chunking is involved, each chunk could be logged as a separate
+ * SERVER_SEND_FRAGMENT in the same span.
+ *
+ * Annotation.host is not the client. It is the host which logged the send
+ * event, almost always the server.
```
yurishkuro: Perhaps the recommendation may be to leave endpoint logging to CS and SR events ("begin" events), not do it in the "end" events.

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
```diff
@@ -15,18 +15,118 @@ namespace java com.twitter.zipkin.thriftjava
...
+const string SERVER_SEND_FRAGMENT = "ssf"
+/*
+ * Optionally logs progress of a (SERVER_RECV, WIRE_RECV). For example, this
+ * could be one chunk in a chunked request.
+ */
+const string SERVER_RECV_FRAGMENT = "srf"
 
-#************** Common binary annotation keys **************
-// The endpoint associated with CLIENT_ annotations is not necessarily Annotation.host
+#***** BinaryAnnotation.key where value = [1] and annotation_type = BOOL ******
```
yurishkuro, line 110: Shouldn't the value be the actual address, in whichever form user code chose to send it? A URI format could be  recommended: `tcp://hostname:port`.

Also, if this attribute is sent, why have the endpoints in the events?

codefromthecrypt, line 110: Well a new model could choose to do that, but this form has been in place
for a couple years. Main thing is documenting what it is already.

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
codefromthecrypt: Sounds good.

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
codefromthecrypt, line 110: PS if in a new model we made endpoint a value type, we might get away with
not logging it as "host" all the time. I suspect the original decision of
CA/SA as a marker annotation was because the value are thrift primitive
types, not structs.

Also, by doing a marker type the normal
annotation.host/binaryannotation.host fields and related logic could be
reused.

Regardless, this is a great idea to bear in mind for model 2.0. I like the
idea of logging the protocol and also revisiting to not constrain to ipv4.

jcarres-mdsol, line 110: Agreed. What's the meaning of having endpoint sent all the time if there is no use for it and the canonical source is ca sa ?
In fact makes it confusing.

The original ruby code was assuming cr cs should have the servicename and addresses of the server which was called. Kind of makes sense. And the UI looks good following that assumption also.

adriancole: First, if it isn't clear, I am documenting what actually exists, for at
least 2 years. Sadly, a lot of instrumentation including Ruby, wasn't well
maintained before you arrived. It might be based on 3 years ago. Let me try
to explain, and please keep an open mind. We have a separate work to making
a new zipkin model. This is about making existing ones compatible.

Zipkin is a logging system and annotations are log lines in that. The host
of an annotation is the host that logged, in this case a client send.
Because these events are placed in time, I can use the host dimension to
find out more about what hosts were doing, such as making client requests.
Knowing this, it is very surprising to put a host in an annotation that
wasn't the actual host. For example, I cannot group my
log/annotation/events by ip anymore.

Binary annotations are tags on a span. It is where you add extra metadata.
For example, you may have multiple ways of establishing the client address,
for example the x-forwarded-for header. Whatever you can find out, can be
added here, and would apply to 'the entire span'. For example, I could make
a query for all events dealing with a client by looking for a single
key/value pair (CA). That might include several actual hosts!

How a client or server address is encoded is confusing. I agree. Because it
is, feel free to read below only if you really care ;) Otherwise, I hope we
can move past the past. If we want to make zipkin's model better, we need
to at least understand what it does now. The Ruby and also other code have
done incorrect things over time because there was no documentation. This
isnt changing anything except that it is now documented. Make sense?

-- why single byte for CA/SA value

For example, it takes a while to understand why bool (a byte array of size
one) was chosen back in 2013. The reason is that there's already a struct
to hold the endpoint and it is already a member of binary annotation. Since
binary annotation.value is not optional, what's the smallest thing you can
place here? Well.. A single byte is the answer to that.

adriancole:
good news is that the finagle side of this PR validated the interpretations are correct.
bad news is that it is still confusing. I'll give a stab at reducing that.

For example, discussing annotation values has made it clear that there's confusion still about annotations vs binary annotations in general. For example, Annotation.host doc says "The endpoint that recorded this annotation". This is still being confused and conflated with the origin or destination of a call. That may be because that doc isn't in the diff, but regardless, it says there's still opportunity to make this better. I'll take the challenge.

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
adriancole: In zipkin all annotations/binary annotations log the "local" ip/port in the host field.

This allows you to group events by the host creating them (via endpoint.ip). For example, if I have a host that's acting awry, I could narrow that down and see all activity on that host. If we didn't store the local IP in annotation.host, we could not do this.

By using the local ip/port in annotations (log statements), we can also see when a client and server disagree on a value of the same key. For example, if "http.uri" can be logged as "/v1/foo" on the client, and received as "/foo" on the server. If we logged the server's ip as the host of a "http.uri" annotation sent by a client, we wouldn't be see differences like this, or know which was caused by who.

For better or worse, there are two exceptions to the "always use local ip/port in annotation/binary annotations" rule:  `CLIENT_ADDR` and `SERVER_ADDR`. When a binary annotation key is "ca" or "sa", the value of `[1]` is ignored, as the actual value is the host/endpoint. These special annotations allow us to make visible the ip and port of uninstrumented clients or servers. For example, a root span begins with a "sr", but we often can tell what the IP of the client was. For example, we could store the `X-Forwarded-For` value. When sending a call to an uninstrumented destination, such as an Amazon S3 bucket, we can record the ip/port as `SERVER_ADDR`. 

What's not immediately clear is why we put `[1]` as a value, instead of serializing the endpoint in some encoding. The answer is more than just avoiding duplicating binaryAnnotation.host as also binaryAnnotation.value. It is that binaryAnnotation.host is already destructured and indexed in all backends. If we stored the host as binaryAnnotation.value, we would need to do heavy lifting to parse that across potentially large amounts of data.

This optimization comes at a cost of intuitiveness, as the value isn't where we expect it to be. It also requires guidance of log SERVER_ADDR from the client and visa-versa, as without that, interpretation is confusing as well. Theses issues may lead to refactoring in [zipkin v2](https://docs.google.com/document/d/1ixxEs9TvhiGjJObGbRSPhSna3zHdadoUTQIZ5JKgLzU/edit).

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
codefromthecrypt: done

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
codefromthecrypt: done

codefromthecrypt:
ok updated based on feedback from @jcarres-mdsol @yurishkuro and finagle

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
codefromthecrypt, line 110: moved the long discussion here: https://docs.google.com/document/d/1ixxEs9TvhiGjJObGbRSPhSna3zHdadoUTQIZ5JKgLzU/edit#

Added more clarification in the thrift doc about why this is important, hopefully not too much explaining..

codefromthecrypt:
got an ack from finagle that the docs are still valid. plan to merge this tomorrow

michaelsembwever:
very nice.

PR #795, opened 2015-10-26
Reverts Cassandra schema change from timestamp to bigint

codefromthecrypt:
I misunderstood the precision of Cassandra timestamps, distracted by
conflicting CQL documentation and the use of j.u.Date (which has millis
precision). This led to advice to change our schema to bigint to avoid
precision loss storing timestamps.

@yurishkuro's production environment broke on this change, pinning them
to zipkin 1.12. He noted that there's no precision problem, as zipkin
go ingesters are storing micros in timestamp fields!

Upon investigation, and by help of datastax, we found it is indeed
possible with custom code. This change reverts the schema, but without
a precision loss in java.

Reverts 521f209972db8b741cc0ceacc7e42be71420b8e4
See https://groups.google.com/a/lists.datastax.com/forum/#!topic/java-driver-user/PUpJh6cKaRA

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 #Sat, 15 Aug 2015 15:15:18 +0000
-version=1.18.2-SNAPSHOT
+version=1.19.0-SNAPSHOT
 group=io.zipkin
 repo=https://github.com/openzipkin/zipkin
 description=A distributed tracing system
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
cc @danchia @eirslett @michaelsembwever @kristofa

danchia:
LGTM, thanks for digging deeper.

PR #797, opened 2015-10-26
Makes collector optional by re-routing zipkin tracing to http

codefromthecrypt:
This does the following, in order to make collector an optional part of
zipkin.
- `zipkin-web` logs traces to `zipkin-query` when `TRANSPORT_TYPE=http`
- `zipkin-query` logs traces locally when `TRANSPORT_TYPE=http`
- `bin/query` now owns the file-based datastore when `STORAGE_TYPE=dev`
- Documentation no longer encourages users to start `bin/collector`
- Prevents `zipkin-query` from tracing recursively on POST requests

This change relies on custom finagle tracers, and tests have been added.


codefromthecrypt:
tested locally against cassandra and mysql, both scala and java projects

yurishkuro:
does it mean that query service now writes to storage?

codefromthecrypt:
yes, via SpanStore. However that happened in an earlier version as well.
This change is about how self-tracing is performed.

PR #801, opened 2015-10-28
Moves minimum JRE to 8 and updates library versions

### LINKED ISSUE #777: Move minimum JRE to 8
The main zipkin project's artifacts are services, particularly a collector, query and web process. We have done a lot of work to update things to scala 2.11, which is why all major adopters can run on JRE 8. This includes Twitter as I personally moved internal zipkin onto JRE 8 earlier this year.

We wanted to update to JDK 8, as there are a lot of libraries that no longer compile to JRE 7 bytecode. We decided against, as Twitter's internal software still compiled with JDK 7. We can no longer justify holding the project back for a single user. There are many problems in anorm we can't fix as they don't release new versions that work with JRE 7. We need to start thinking about scala 2.12.. we can't freeze forever.

Let's make the call, accept the collateral damage, and update to producing Java 8 bytecode. The primary impact is those still compiling with javac 7 can no longer patch zipkin servers.

Those using zipkin instrumentation are unaffected as this is decoupled completely. For example, both brave and finagle target minimum Java 1.7. This means web apps that generate traces, running on JRE 7 aren't affected at all. It is only the servers that are updated.

codefromthecrypt:
Zipkin servers now require JRE 8 or higher. This also updates libraries,
particularly mustache which removes the need for us to host a patched
class for JRE 7 interop.

Fixes #777


codefromthecrypt:
cutting as the only change in 1.20.0

PR #805, opened 2015-10-30
Enforces span and service names as lowercase

### LINKED ISSUE #767: 'spanName' filter should consider character case in zipkin web UI
I'm using finagle which might record it's rpc method as span name, e.g. "getResult",

in the zipkin UI, there will be a lower case "getresult" show in the dropdown, when I select it I'll find nothing, but if I manually change the url from like:

http://localhost/?serviceName=x&spanName=getresult ....
to 

http://localhost/?serviceName=x&spanName=getResult ....

I could get the right answer.

it looks like we should not transform the span name into lower case in UI for selection

codefromthecrypt:
There have been many subtle issues around span and service name case.
Particularly, not all indexes are case-insensitive, which makes search
behavior inconsistent across storage layers.

We discussed this topic at length on gitter, and decided that the best
route to take is to enforce strict lowercase on span and service names.

See #776
Fixes #767

### DIFF
--- a/zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
+++ b/zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
@@ -56,9 +56,6 @@ abstract class CassandraSpanStore(
   private[this] def createSpanColumnName(span: Span): String =
     "%d_%d_%d".format(span.id, span.annotations.hashCode, span.binaryAnnotations.hashCode)
 
-  private[this] def nameKey(serviceName: String, spanName: Option[String]): String =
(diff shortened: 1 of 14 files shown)


zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
```diff
@@ -56,9 +56,6 @@ abstract class CassandraSpanStore(
   private[this] def createSpanColumnName(span: Span): String =
     "%d_%d_%d".format(span.id, span.annotations.hashCode, span.binaryAnnotations.hashCode)
 
-  private[this] def nameKey(serviceName: String, spanName: Option[String]): String =
```
codefromthecrypt, line 59: not used

PR #806, opened 2015-10-30
Updates twitter util version to match finagle

codefromthecrypt:


yurishkuro:
works for me from cmd line, so lg.

yurishkuro:
IntelliJ is happy now as well

codefromthecrypt:
Yes!

PR #807, opened 2015-11-01
Enables duration queries by promoting Span.duration, timestamp

codefromthecrypt:
A commonly requested feature is querying by duration, for example showing traces longer than 1 minute. 

While the UI supports ordering by duration, it is a game of chance whether the traces returned happen to be long or not. Even if you put a limit of 1000, you still may not find the longest trace (as your storage system may have a million traces in it).

The path proposed is top-level Span.timestamp, and Span.duration, which formerly existed in the scala and mustache models, but not in the thrift.

We could implement duration queries without changing thrifts. For example, this could be an implementation detail of each storage system. However, this limits testing to side-effects and keeps essential details in scala code, making portability harder.

Also, we have other reasons to top-level these fields:
- local spans can be smaller (#808)
- non-rpc spans can't be supported without artificially logging annotations like cs/cr
- reduce redundant span.duration logic in several scala classes and mustache templates
- Dependency graph links can cheaply add duration aggregates

Here's a convenience paste of the definitions of Span.timestamp and Span.duration

``` java
  /**
   * Microseconds from epoch of the creation of this span.
   *
   * This value should be set directly by instrumentation, using the most
   * precise value possible. For example, gettimeofday or syncing nanoTime
   * against a tick of currentTimeMillis.
   *
   * For compatibilty with instrumentation that precede this field, collectors
   * or span stores can derive this via Annotation.timestamp.
   * For example, SERVER_RECV.timestamp or CLIENT_SEND.timestamp.
   *
   * This field is optional for compatibility with old data: first-party span
   * stores are expected to support this at time of introduction.
   */
  10: optional i64 timestamp,
  /**
   * Measurement of duration in microseconds, used to support queries.
   *
   * This value should be set directly, where possible. Doing so encourages
   * precise measurement decoupled from problems of clocks, such as skew or NTP
   * updates causing time to move backwards.
   *
   * For compatibilty with instrumentation that precede this field, collectors
   * or span stores can derive this by subtracting Annotation.timestamp.
   * For example, SERVER_SEND.timestamp - SERVER_RECV.timestamp.
   *
   * If this field is persisted as unset, zipkin will continue to work, except
   * duration query support will be implementation-specific. Similarly, setting
   * this field non-atomically is implementation-specific.
   *
   * This field is i64 vs i32 to support spans longer than 35 minutes.
   */
  11: optional i64 duration
```

### DIFF
--- a/zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
+++ b/zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
@@ -16,7 +16,7 @@ namespace java com.twitter.zipkin.thriftjava
 namespace rb Zipkin
 
 #************** Annotation.value **************
-/*
+/**
  * The client sent ("cs") a request to a server. There is only one send per
  * span. For example, if there's a transport error, each attempt can be logged
  * as a WIRE_SEND annotation.
@@ -29,7 +29,7 @@ namespace rb Zipkin
  * should also log the SERVER_ADDR.
  */
 const string CLIENT_SEND = "cs"
-/*
+/**
  * The client received ("cr") a response from a server. There is only one
  * receive per span. For example, if duplicate responses were received, each
  * can be logged as a WIRE_RECV annotation.
@@ -42,7 +42,7 @@ const string CLIENT_SEND = "cs"
  * recorded separately as SERVER_ADDR when CLIENT_SEND is logged.
  */
 const string CLIENT_RECV = "cr"
-/*
+/**
  * The server sent ("ss") a response to a client. There is only one response
  * per span. If there's a transport error, each attempt can be logged as a
  * WIRE_SEND annotation.
@@ -58,7 +58,7 @@ const string CLIENT_RECV = "cr"
  * recorded separately as CLIENT_ADDR when SERVER_RECV is logged.
  */
 const string SERVER_SEND = "ss"
-/*
+/**
  * The server received ("sr") a request from a client. There is only one
  * request per span.  For example, if duplicate responses were received, each
  * can be logged as a WIRE_RECV annotation.
@@ -74,54 +74,54 @@ const string SERVER_SEND = "ss"
  * should also log the CLIENT_ADDR.
  */
 const string SERVER_RECV = "sr"
-/*
+/**
  * Optionally logs an attempt to send a message on the wire. Multiple wire send
  * events could indicate network retries. A lag between client or server send
  * and wire send might indicate queuing or processing delay.
  */
 const string WIRE_SEND = "ws"
-/*
+/**
  * Optionally logs an attempt to receive a message from the wire. Multiple wire
  * receive events could indicate network retries. A lag between wire receive
  * and client or server receive might indicate queuing or processing delay.
  */
 const string WIRE_RECV = "wr"
-/*
+/**
  * Optionally logs progress of a (CLIENT_SEND, WIRE_SEND). For example, this
  * could be one chunk in a chunked request.
  */
 const string CLIENT_SEND_FRAGMENT = "csf"
-/*
+/**
  * Optionally logs progress of a (CLIENT_RECV, WIRE_RECV). For example, this
  * could be one chunk in a chunked response.
  */
 const string CLIENT_RECV_FRAGMENT = "crf"
-/*
+/**
  * Optionally logs progress of a (SERVER_SEND, WIRE_SEND). For example, this
  * could be one chunk in a chunked response.
  */
 const string SERVER_SEND_FRAGMENT = "ssf"
-/*
+/**
  * Optionally logs progress of a (SERVER_RECV, WIRE_RECV). For example, this
  * could be one chunk in a chunked request.
  */
 const string SERVER_RECV_FRAGMENT = "srf"
 
 #***** BinaryAnnotation.key where value = [1] and annotation_type = BOOL ******
-/*
+/**
  * Indicates a client address ("ca") in a span. Most likely, there's only one.
  * Multiple addresses are possible when a client changes its ip or port within
  * a span.
  */
 const string CLIENT_ADDR = "ca"
-/*
+/**
  * Indicates a server address ("sa") in a span. Most likely, there's only one.
  * Multiple addresses are possible when a client is redirected, or fails to a
  * different server ip or port.
  */
 const string SERVER_ADDR = "sa"
 
-/*
+/**
  * Indicates the network context of a service recording an annotation with two
  * exceptions.
  *
@@ -142,14 +142,20 @@ struct Endpoint {
   3: string service_name
 }
 
-/*
+/**
  * An annotation is similar to a log statement. It includes a host field which
  * allows these events to be attributed properly, and also aggregatable.
  */
 struct Annotation {
-  1: i64 timestamp                 // microseconds from epoch
+  /**
+   * Microseconds from epoch.
+   *
+   * This value should use the most precise value possible. For example,
+   * gettimeofday or syncing nanoTime against a tick of currentTimeMillis.
+   */
+  1: i64 timestamp
   2: string value                  // what happened at the timestamp?
-  /*
+  /**
    * Always the host that recorded the event. By specifying the host you allow
    * rollup of all events (such as client requests to a service) by IP address.
    */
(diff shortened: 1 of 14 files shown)


zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
```diff
@@ -204,5 +204,19 @@ struct Span {
...
+   * SERVER_RECV.timestamp
+   */
+  10: optional i64 start_ts,
+  /**
+   * Duration of the span in microseconds, used to support queries. This value
+   * is usually derived by subtracting the last annotation from the first. For
+   * example, SERVER_SEND.timestamp - SERVER_RECV.timestamp.
+   *
+   * Note that this should be treated unsigned. i32 implies trace durations are
+   * not longer than 35.79 minutes.
```
yurishkuro: I don't think that's a reasonable limitation. Just the other day I had a discussion with the data team where they sometimes have spans measured in hours. I would go for i64.

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
```diff
@@ -204,5 +204,19 @@ struct Span {
   6: list<Annotation> annotations, # all annotations/events that occured, sorted by timestamp
   8: list<BinaryAnnotation> binary_annotations # any binary annotations
   9: optional bool debug = 0       # if true, we DEMAND that this span passes all samplers
+  /**
+   * Microseconds from epoch, usually taken from the first annotation. Ex.
+   * SERVER_RECV.timestamp
+   */
+  10: optional i64 start_ts,
```
yurishkuro: What is the benefit of capturing this explicitly instead of the usual cs/cr? It feels like it's going to complicate other code paths as they need to handle dual representation.

yurishkuro:
> In order to support a query by duration, we need to have an index of (service name, duration) -> traceId limited by startTs. We've always had implicit Span.duration and Span.startTs fields. To ensure we can support duration queries, we need to make these explicit.

The last sentence does not follow from the requirements. Yes, we do need an index, but we don't need to capture start/duration in the collected spans, we just need a way to calculate it in the collector.  We could  use cs/cr for that, provided they reach the collector in a single submission from instrumentation.

adriancole:
So, startTs is needed to satisfy the time constraint on QueryRequest. It is
present in almost all the storage backends. Duration is also needed to
support querying. If we don't capture this way, then things like local
spans still have to artificially log cs/cr, just to log their duration.

While I agree that we can leave this as an implementation detail, it is
kindof a hairy one, if we actually want to support query by duration.

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
codefromthecrypt: PS I don't see this as complicating, particularly as I've touched all the
code in zipkin. The current codebase is a mess as it has implicit logic
around things like startTs and duration. This formalizes it, making it
possible to create an un-hairy implementation.

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
codefromthecrypt: fair enough.

adriancole:
PS there might be something unclear about this change, which is what the impact is on existing code.

Firstly, these fields are marked optional, so existing collectors can continue to operate in absence of values. All SpanStores need to track duration whether or not we formalize this in the data type.

Here's info about the two fields.
* `Span.startTs` already exists, but not in thrift. Existing SpanStores have a randomly named field for `Span.startTs`. This is used to support existing trace and dependency queries. The change is to make this official, which removes the on-demand processing of `Span.startTs`.
* `Span.duration` also exists, but not in thrift. This would be added to SpanStore implementations to support duration queries. It can also be used to clean-up implementation of dependency queries.

We can choose to do nothing, and just keep special casing things. So far, this has led us to very complex logic and sometimes expensive queries. When adding new features I like to look at how we can reduce debt to pay them off.

I believe that if we continue to leave `Span.startTs` and `Span.duration` as implementation details, we just add more hair to the yak. I'm not sure I would want to do this change if that were the case. Rather, we should take this opportunity to first-class span duration *and deprecate* other means to do the same.

IOTW, we could keep using annotation subqueries to derive a spans duration, but we should stop doing that.

adriancole:
added more detail to rationale

yurishkuro:
These are my concerns with adding the fields to thrift: 
- it suggests to the instrumentation libs to send redundant data, since they are already sending the start/end annotations from which startTs/duration can be derived. The Zipkin story about what and how the instrumentation libs have to send is already super convoluted.
- In RPC case the start/duration can be derived either from server annotations or from client annotations, depending on where the instrumentation exists. There isn't really such a thing as a "span" in the ingestion pipeline, more like multiple slices of span. A slice + annotations is not ambiguous (although  "span name" is), but slice + start/duration definitely are. So someone need to resolve that somewhere, and I think it's better resolved using the raw data of annotations than using potentially incorrectly calculated start/duration.
- In cases when instrumentation cannot send the complete span, some other mechanism, like a Spark pipeline, can do a post-processing and calculate start/duration having seen the complete span in the storage (multiple slices stored independently). It would certainly be storage implementation dependent how such post-caculated result would be stored. In Cassandra we'd add a record to the index by duration. If other stores have a record representing a complete span, they would have to make an update to that record. Forcing Cassandra to store an update of post-calculation would be pretty bad for performance.

yurishkuro:
btw, what is your concern with having these two fields be implementation-specific? The span API that query service deals with can have them as def's, and storage implementations fill it. In many cases the actual calculation should be the same, e.g. if it's from annotations, so that logic can be refactored to a base class or composition. I thought right now it it's already in the shared class.

adriancole:
So I've worked on many projects with similar responsibility and the cost of zipkin is high because of all the tacit relationships. It is hard to keep wheels turning and offer efficient implementations when the same data problem is solved in several slightly similar ways. For example, it raises the test burden as all this stuff is only visible via side effects.

Maybe if we rotated the burden of maintaining storage systems we aren't personally interested in, it might  become more obvious how expensive this is.

adriancole:
Also, I think we should stop saying things like Cassandra will be expensive, considering no-one has pitched in to do even basic benchmarking. We shouldn't hold features hostage to fear.

adriancole:
My hope is that we can come together on this. I'm not sure we are hearing eachother, so maybe this isn't today. Maybe tomorrow!

Instrumentation sends what they can. When they don't know explicitly that they are starting a span (unlikely) or they are ending a span, they don't set those fields.

Collectors process what they can. When they have start and/or duration, they pass that directly to the storage tier (like they do now). When they have a well-known start/stop, they set duration. When they don't they don't set duration.

Duration back-filling jobs could be added, or it can be handled at query time like we do time skew... or we choose to support duration only with clean spans.

If we come to a point where we specify this, similar to how htrace does, or how our scala model does, or how our storage implementations do, I think we are far better for it. We are better for it, as the code can now become more maintainable as we have a portable way to describe and implement duration without digging through scala code.

If after this, you still aren't in on it.. well, I'm sad as the change to support your view is dropping this pull request, and making yet more logic implicit and hard to test portably. This is why I hope tomorrow is a better day.

yurishkuro:
ok, I'll shut up, go for it. I believe in my system we may be able to calculate those fields in the collector.

codefromthecrypt:
Glad it might work. Just added some more detail around how I see this fitting in with local spans. PTAL, as it may make more sense within this context https://github.com/openzipkin/zipkin/issues/808

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
codefromthecrypt: done

codefromthecrypt:
Re-summarized in the thrift definition and WIP description. I removed my related redundant comments.

codefromthecrypt:
Note I plan to start on this today, as soon as the supporting changes are released as 1.21.1

codefromthecrypt:
ETA tomorrow!

codefromthecrypt:
expect an update tomorrow pacific AM. Just polishing tests

codefromthecrypt:
ok. this is finally complete. Once merged, I'll raise a pull request for duration query.

PR #809, opened 2015-11-03
Enforces lowercase serviceName and span name by validation

codefromthecrypt:
Scala case class doesn't support custom construction, and the penalty
for working around this is high. For example, not only do we have to do
normal java toString, hashCode, equals, but also copy and apply methods.

Since zipkin already has multiple representation objects, ex for json
and thrift, it is far less tech debt to do lowercasing in their
conversion functions vs writing manual case classes.

See #805

### DIFF
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/common/Endpoint.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/common/Endpoint.scala
@@ -20,23 +20,18 @@ import java.net.{InetAddress, InetSocketAddress}
 import java.nio.ByteBuffer
 
 import com.google.common.collect.ComparisonChain
-
-import scala.util.hashing.MurmurHash3
+import com.twitter.zipkin.util.Util._
 
 /**
  * Represents the client or server machine we traced.
  *
  * @param ipv4 ipv4 ip address.
  * @param port note that due to lack of unsigned integers this will wrap.
- * @param _serviceName the service this operation happened on, in lowercase
+ * @param serviceName the service this operation happened on, in lowercase
  */
-// This is not a case-class as we need to enforce serviceName as lowercase
-class Endpoint(val ipv4: Int, val port: Short, _serviceName: String) extends Ordered[Endpoint] {
-
-  /**  the service this operation happened on, in lowercase. */
-  val serviceName: String = _serviceName.toLowerCase
-
-  override lazy val toString = "%s:%d(%s)".format(getHostAddress, port, serviceName)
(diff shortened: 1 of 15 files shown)


zipkin-web/src/test/scala/com/twitter/zipkin/web/HttpZipkinTracerTest.scala
```diff
@@ -0,0 +1,70 @@
...
+
+object HttpZipkinTracerTest {
+  // Singleton as the test needs to read the actual port in use
+  val server = new MockWebServer()
+
+  // Scala cannot generate fields with public visibility, so use a def instead.
+  @ClassRule def serverDef = server
+}
+
+class HttpZipkinTracerTest extends JUnitSuite {
```
adriancole: this class was hanging around in my workspace and is on a different commit than the lowercase stuff.

yurishkuro:
the usual way to get custom constructors for case classes is to implement conversion logic in the companion's apply method, and make the constructor private. I think the only thing you loose by doing that is the copy method (but not sure).

codefromthecrypt:
at least in intellij, private constructor vs apply with same args often gets confused. feel free to raise a PR to try it out though!

danchia:
private constructor with apply starts going down the manual `case class` route, so may not be the best choice.

One thing I might suggest however, is adding a `require` into the case class body to make sure all the things that should be lowercase are indeed lowercase, so people who skip the conversions are not caught off guard.

zipkin-common/src/main/scala/com/twitter/zipkin/common/Endpoint.scala
```diff
@@ -20,23 +20,18 @@ import java.net.{InetAddress, InetSocketAddress}
...
  */
-// This is not a case-class as we need to enforce serviceName as lowercase
-class Endpoint(val ipv4: Int, val port: Short, _serviceName: String) extends Ordered[Endpoint] {
-
-  /**  the service this operation happened on, in lowercase. */
-  val serviceName: String = _serviceName.toLowerCase
-
-  override lazy val toString = "%s:%d(%s)".format(getHostAddress, port, serviceName)
+case class Endpoint(ipv4: Int, port: Short, serviceName: String) extends Ordered[Endpoint] {
+  checkArgument(serviceName.toLowerCase == serviceName,
```
codefromthecrypt, line 33: @danchia you mean "require" instead of this? mind giving it a go? Note, we do this a few places, so you could do a reference hunt for Util.checkArgument

PR #810, opened 2015-11-03
Clarifies a trace as a list of spans

codefromthecrypt:
`Span.annotations` is a list because it is ordered. A trace is an
ordered list of spans, yet in scala we designate it as a Seq. This led
to numerous bugs where people tried to order seqs. The fixes to those
resulted in all implementations using `List` not `Seq`. This change
formalizes that relationship, which has a side-effect of making trace
transformations more likely to retain order.

### DIFF
--- a/zipkin-scrooge/src/main/scala/com/twitter/zipkin/conversions/thrift.scala
+++ b/zipkin-scrooge/src/main/scala/com/twitter/zipkin/conversions/thrift.scala
@@ -133,7 +133,7 @@ object thrift {
       result += thriftSpanToSpan(thrift).toSpan
     }
     proto.readListEnd()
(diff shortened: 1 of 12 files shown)


zipkin-scrooge/src/main/scala/com/twitter/zipkin/conversions/thrift.scala
```diff
@@ -133,7 +133,7 @@ object thrift {
       result += thriftSpanToSpan(thrift).toSpan
     }
     proto.readListEnd()
-    result
+    result.toList
```
codefromthecrypt, line 136: this changes from mutable to immutable list

PR #812, opened 2015-11-04
Removes the Trace type, moving its methods to their call-site

codefromthecrypt:


codefromthecrypt:
next after this is cleaning up use of Span.timestamp/duration in the scala api before we backfill it with a thrift change

codefromthecrypt:
tested with collector, query, tracegen and web

PR #813, opened 2015-11-04
Enable auto-syncing releases to Maven Central

abesto:
According to https://github.com/bintray/gradle-bintray-plugin/pull/82 the actual sync is now decoupled from the request, so the build can make the request, carry on, and the sync will happen in the background.


codefromthecrypt:
took a while to finish, but not sure it is related. Thanks!

Issue #808, opened 2015-11-02
Support local spans via a LOCAL_COMPONENT("lc") binary annotation


codefromthecrypt:
Currently it is tribal knowledge that local (in-process) spans exist. We need a way to designate span as local, and offer insight into how to use them. This can lead us to answer what status quo could be on this, and how they might be tested.

Firstly, local spans need to be annotated with a namespace that can be used like service. A discussion on github suggests using a LOCAL_COMPONENT("lc") binary annotation for this.

Here's a json view:

``` json
{
  "key" : "lc",
  "value" : "encryption-service",
  "endpoint" : {
    "serviceName" : "proxy",
    "ipv4" : "1.2.3.4",
    "port" : 345
  }
}
```

You'll notice that the binary annotation holds almost all information needed to identify the context of the span. It contains a searchable annotation which contains its namespace, in this case `encryption-service`. Via the existing contract of binary annotations, it also includes the service that contains it, in this case `proxy`.

What's left is the skeleton of the span. How do we capture the duration of the span?

I'm aware of three options:
- Re-use "cs/cr" annotations
- Make up new annotations for the bookends of a local span.
- piggy-back off Span.startTs/duration (#807)

TL;DR; is that I think the latter is best.

If we re-use "cs/cr" annotations, we avoid logic around dependency graphs where server annotations are special-cased. In other words, we avoid accidentally cluttering it with local spans. We can also hang off an existing explanation where we use "cs/cr" to address uninstrumented systems. It might be the least code changes.

``` json
{
  "traceId" : "0000000000000005",
  "name" : "encrypt",
  "id" : "000000000000029a",
  "parentId" : "0000000000000002",
  "annotations" : [
    {
      "timestamp" : 60,
      "value" : "cs",
      "endpoint" : {
        "serviceName" : "proxy",
        "ipv4" : "1.2.3.4",
        "port" : 345
      }
    },
    {
      "timestamp" : 100,
      "value" : "cr",
      "endpoint" : {
        "serviceName" : "proxy",
        "ipv4" : "1.2.3.4",
        "port" : 345
      }
    }
  ],
  "binaryAnnotations" : [
    {
      "key" : "lc",
      "value" : "encryption-service",
      "endpoint" : {
        "serviceName" : "proxy",
        "ipv4" : "1.2.3.4",
        "port" : 345
      }
    }
  ]
}
```

[...]

PR #814, opened 2015-11-04
Settles on a single timestamp for Span

codefromthecrypt:
This moves existing code around the notion of Span.timestamp,
Span.duration discussed in #807.

The impact from a user POV is minimal. `endTs`, used by the query and UI
formerly looked for the last timestamp in a trace. What this meant is
that if someone clicked search, waited, then clicked search again with
the same `endTs`, an in-flight trace may "disappear" if it has new
activity. Since `endTs` is now based on a stable point (the start), a
trace wouldn't disappear anymore. This impact is so subtle that it is
barely worth discussing.

The primary motivation for this change is to simplify the commodity task
of timestamping and duration stamping spans. This is discussed #807, and
directly supports a new minimal design of local spans (#808).


codefromthecrypt:
NOTE: there's zero DDL change needed to support this. However, the interpretation of data is different.

PR #815, opened 2015-11-05
Moves clock skew correction to SpanStore, for portable testing

codefromthecrypt:
Clock skew correction is vital for the UI to appear properly,
particularly in the current model. This is because there's no concept of
sequence of spans except by timestamp.

The type of clock skew adjustment zipkin does is shifting child spans to
a later timestamp when they appear to happen before their parent.

Formerly, there were no tests to show span stores actually do this. This
means you can create implementations which appear portable, but are not
useful as they result in a poor UI.

This moves the implementation of clock skew correction to the common
package, and introduces a new test in SpanStoreSpec which asserts what
expectations are, without limiting how implementations correct clock
skew. For example, this test will pass regardless of whether clock skew
is corrected at storage or query time.


zipkin-common/src/test/scala/com/twitter/zipkin/storage/SpanStoreSpec.scala
```diff
@@ -258,4 +259,56 @@ abstract class SpanStoreSpec extends JUnitSuite with Matchers {
...
+   * the UI supports happens-before without using timestamps. The easiest clock
+   * skew to correct is where a child  appears to happen before the parent.
+   *
+   * It doesn't matter if clock-skew correction happens at storage or query
+   * time, as long as it occurs by the time results are returned.
+   *
+   * Span stores who don't support this can override and disable this test,
+   * noting in the README the limitation.
+   */
+  @Test def correctsClockSkew() {
```
codefromthecrypt, line 274: @dsyer @spencergibb fyi, this is good because it will break the zipkin-java project. I've a partial java impl of clock skew correction somewhere, and will finish that or ignore the test when bumping.

zipkin-common/src/main/scala/com/twitter/zipkin/adjuster/MergeById.scala
```diff
@@ -0,0 +1,47 @@
...
+package com.twitter.zipkin.adjuster
+
+import com.twitter.zipkin.common._
+import scala.collection.mutable
+
+/**
+ * Merge all the spans with the same id. This is used by span stores who
+ * store partial spans and need them collated at query time.
+ *
+ * Spans without a timestamp are filtered out, as they are not possible to
```
yurishkuro: maybe make it a flag, since e.g. for Dependencies job you don't need timestamps, just the relationships.

codefromthecrypt: well, you do in order to query. Ex. the dependencies query is strictly bound by time, right?

zipkin-common/src/main/scala/com/twitter/zipkin/common/Trace.scala
```diff
@@ -0,0 +1,11 @@
+package com.twitter.zipkin.common
+
+/** Utilities for working on traces */
+object Trace {
+
+  /** What's the total duration of the spans in this trace? */
+  def duration(spans: List[Span]): Long = {
+    val endTs = spans.flatMap(_.annotations).map(_.timestamp).reduceOption(_ max _)
+    (endTs.getOrElse(0L) - spans.headOption.flatMap(_.timestamp).getOrElse(0L))
```
yurishkuro: not sure I follow this one. Shouldn't it be like this:

```
start := min(*.timestamp)
end := max(*.timestamp + *.duration)
duration := end - start
```

codefromthecrypt: this logic was moved. The change for this to become duration based is in another PR.

codefromthecrypt: span.duration I mean.. anyway I agree it is weird and will move the logic from that branch here.

zipkin-common/src/main/scala/com/twitter/zipkin/adjuster/MergeById.scala
yurishkuro: I think the dependencies job itself just reads all data from `traces` table. There's no index in C\* implementation to only query for traces within a given timeframe. But anyway, dep job does its own merging now, so ok.

yurishkuro:
lg

codefromthecrypt:
thx for the review, @yurishkuro 

just going to pull in the duration thing.. so eta on merge is w/in the hour

zipkin-common/src/main/scala/com/twitter/zipkin/adjuster/MergeById.scala
codefromthecrypt: Well, it is bucketed by day in the old job, but anyways I agree
dependencies is something that is easier w/ storage api's directly.

codefromthecrypt: clarified the comment, which explains why I think a flag would be instant tech-debt and not actually solve anything.

zipkin-common/src/main/scala/com/twitter/zipkin/common/Trace.scala
codefromthecrypt: moved the code.. it looks terrible in scala (bc these are options)

zipkin-common/src/main/scala/com/twitter/zipkin/common/Trace.scala
```diff
@@ -0,0 +1,17 @@
+package com.twitter.zipkin.common
+
+/** Utilities for working on traces */
+object Trace {
+
+  /** What's the total duration of the spans in this trace? */
+  def duration(spans: List[Span]): Option[Long] = {
```
codefromthecrypt, line 7: @yurishkuro this is the best I could do with scala.

Feel free to offer another solution post-merge

I'm suggesting post-merge because my experience with people's suggestions on scala is that they don't work. After this is merged, we have tests that will pass or fail based on an optimized solution as opposed to what I'm usually subjected to, which is "try this!" where "this" doesn't work for some reason.

PR #820, opened 2015-11-06
Adds minDuration and maxDuration to the query api, implemented by SQL

codefromthecrypt:
This adds minDuration and maxDuration queries that zipkin-web can use to
search for long traces. This is initially supported by the in-memory and
anorm (SQL) span stores.

This adds very basic UI support, which someone better at javascript can
polish up:

screen shot 2015-11-06 at 4 34 56 pm

### DIFF
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
@@ -149,6 +149,23 @@ class InMemorySpanStore extends SpanStore with CollectAnnotationQueries {
       .toList
   }
 
+  override protected def getTraceIdsByDuration(
+    serviceName: String,
+    minDuration: Long,
+    maxDuration: Option[Long],
+    endTs: Long,
+    limit: Int
+  ): Future[Seq[IndexedTraceId]] = call {
+    spansForService(serviceName)
+      .filter(s => s.id == s.traceId) // only root spans
+      .filter(_.timestamp.exists(_ <= endTs))
+      .filter(_.duration.exists(_ >= minDuration))
+      .filter(_.duration.exists(_ <= maxDuration.getOrElse(Long.MaxValue)))
+      .take(limit)
+      .map(span => IndexedTraceId(span.traceId, span.timestamp.get))
+      .toList
+  }
+
   override def getAllServiceNames(): Future[Seq[String]] = call {
     spans.flatMap(_.serviceNames).distinct.toList.sorted
   }
--- a/zipkin-web/src/main/resources/templates/v2/index.mustache
+++ b/zipkin-web/src/main/resources/templates/v2/index.mustache
@@ -26,6 +26,11 @@
     </div>
 
     <div class="form-group">
+      <label for="minDuration">Duration in microseconds >=</label>
+      <input class="form-control input-sm" id="minDuration" name="minDuration" type="text" value="{{minDuration}}">
+    </div>
+
+    <div class="form-group">
       <label for="limit">Limit</label>
       <input class="form-control input-sm" id="limit" name="limit" type="text" value="{{limit}}">
     </div>
--- a/zipkin-anormdb/src/main/resources/mysql.sql
+++ b/zipkin-anormdb/src/main/resources/mysql.sql
@@ -4,7 +4,8 @@ CREATE TABLE IF NOT EXISTS zipkin_spans (
   `name` VARCHAR(255) NOT NULL,
   `parent_id` BIGINT,
   `debug` BIT(1),
-  `start_ts` BIGINT COMMENT 'Used to implement TTL; First Annotation.timestamp() or null'
+  `start_ts` BIGINT COMMENT 'Span.timestamp(): epoch micros used for endTs query and to implement TTL'
+  `s_duration` BIGINT COMMENT 'Span.duration(): micros used for minDuration and maxDuration query'
 ) ENGINE=InnoDB ROW_FORMAT=COMPRESSED;
 
 ALTER TABLE zipkin_spans ADD UNIQUE KEY(`trace_id`, `id`) COMMENT 'ignore insert on duplicate';
--- a/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
+++ b/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
@@ -47,7 +47,7 @@ class AnormSpanStore(val db: DB,
           db.replaceCommand() +
             """ INTO zipkin_spans
               |VALUES
-              |  ({trace_id}, {id}, {name}, {parent_id}, {debug}, {timestamp})
+              |  ({trace_id}, {id}, {name}, {parent_id}, {debug}, {timestamp}, {duration})
             """.stripMargin)
           .on("trace_id" -> span.traceId)
           .on("id" -> span.id)
@@ -55,6 +55,7 @@ class AnormSpanStore(val db: DB,
           .on("parent_id" -> span.parentId)
           .on("debug" -> span.debug.map(if (_) 1 else 0))
           .on("timestamp" -> span.timestamp)
+          .on("duration" -> span.duration)
           .execute()
 
         span.annotations.foreach(a =>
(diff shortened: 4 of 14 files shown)


codefromthecrypt:
@yurishkuro let me know, if you need a hand on cassandra support.

zipkin-web/src/main/resources/templates/v2/index.mustache
```diff
@@ -26,6 +26,11 @@
     </div>
 
     <div class="form-group">
+      <label for="minDuration">Duration in microseconds >=</label>
```
yurishkuro, line 29: I suggest `Duration (&#956;s) &#8805;`

yurishkuro:
has something changed with the styling of UI elements recently?  I just upgraded, and the dropdowns/textboxes look taller / bigger.

Circa 1.20.x:
screen shot 2015-11-07 at 3 37 30 pm

Now:
screen shot 2015-11-07 at 3 39 59 pm

zipkin-web/src/main/resources/templates/v2/index.mustache
codefromthecrypt, line 29: sure.. can bump!

codefromthecrypt:
@yurishkuro I am not aware of any intentional change like that.

zipkin-web/src/main/resources/templates/v2/index.mustache
codefromthecrypt, line 29: https://github.com/openzipkin/zipkin/pull/822

codefromthecrypt:
Between 1.19 and now, mustache changed.. that's the template library. I
don't see any CSS changes, but maybe viewing source on both could identify
what's up

git diff 1.19.0 1.22.1 zipkin-web/build.gradle

yurishkuro:
Looking @ Cassandra. One difficulty there is that the query becomes a composite query, `ts <= X and duration >= Y`, which cannot be satisfied directly by Cassandra. 

```
cqlsh:zipkin> select * from duration_index where service_name='mezzanine' and time_bucket=1 and ts > '2015-11-07T11:00:00' and duration > 100000;
InvalidRequest: code=2200 [Invalid query] message="PRIMARY KEY column "duration" cannot be restricted (preceding column "ts" is restricted by a non-EQ relation)"
```

The only efficient way I can think of doing it is by bucketing the timestamps, say hourly, so that the schema will be 

```
CREATE TABLE IF NOT EXISTS zipkin.duration_index (
    service_name      text,
    time_bucket       int,
    duration          bigint,
    ts                timestamp,
    trace_id          bigint,
    PRIMARY KEY ((service_name, time_bucket), duration, ts)
) WITH CLUSTERING ORDER BY (duration DESC, ts DESC)
```

The drawback of this is that if the time buckets are small, the distribution of returned traces will be skewed. If the buckets are large, like 1d, the partition/row might become too large.

Is there another solution I am missing?

yurishkuro:
small good news, Cassandra allows multi-column restrictions on clustering columns, it just requires special syntax like `(duration, ts) >= (20, '2015-11-07T11:00:00')`. With the table above, and the following data

```
 service_name | time_bucket | duration | ts                       | trace_id
--------------+-------------+----------+--------------------------+----------
            x |           1 |       20 | 2015-11-07 12:00:00-0500 |        1
            x |           1 |       20 | 2015-11-07 11:00:00-0500 |        1
            x |           1 |       10 | 2015-11-07 12:00:00-0500 |        1
            x |           1 |       10 | 2015-11-07 11:00:00-0500 |        1
```

the following query finds both D=20 records at different timestamps:

```
cqlsh:zipkin> select * from duration_index where service_name='x' and time_bucket=1 and (duration, ts) >= (20, '2015-11-07T11:00:00');

 service_name | time_bucket | duration | ts                       | trace_id
--------------+-------------+----------+--------------------------+----------
            x |           1 |       20 | 2015-11-07 12:00:00-0500 |        1
            x |           1 |       20 | 2015-11-07 11:00:00-0500 |        1
```

The question of buckets skewing the distribution still remains, but it could be addressed by asking the user to specify both ends of the timeframe, not just endTs, and searching all relevant buckets. Then the bucket size can be tuned depending on the traffic volume in a given deployment.

codefromthecrypt:
Great progress!

I think we'll eventually hit a point where we need to specify startTs. We
already need this for the dependency query. Should we add this now?

zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
```diff
@@ -149,6 +149,23 @@ class InMemorySpanStore extends SpanStore with CollectAnnotationQueries {
       .toList
   }
 
+  override protected def getTraceIdsByDuration(
+    serviceName: String,
```
yurishkuro, line 153: @adriancole 

Q1 - if I am reading sql query correctly, the duration limit applies to individual span duration, not the overall trace duration, is that correct?

Q2 - what are the semantics of serviceName here? Any service involved in a span satisfying the duration limit? What about services involved in a long trace, but not in the long span in that trace?

Q3 - shouldn't we include span name in the search?

codefromthecrypt, line 153: Thanks for looking deep enough to have questions like this!

> Q1 - if I am reading sql query correctly, the duration limit applies to individual span duration, not the overall trace duration, is that correct?

Since the query only applies to root spans, it is essentially the same as
trace. There's no trace table in SQL as a trace is just where id = traceid

> Q2 - what are the semantics of serviceName here? Any service involved in a span satisfying the duration limit? What about services involved in a long trace, but not in the long span in that trace?

So this doesn't constrain the search to only service names in the root
span. It applies to any service in the trace, at the moment (I think).
Probably we should make a test about this nuance. What's your goal? Limit
to service names only in the root span?

Keep in mind that other queries act similarly, ex annotation query isn't
bound to only annotations from that service iirc. In other words, this may
be a bigger issue.

> Q3 - shouldn't we include span name in the search?
> You are only seeing a partial query. The scala lib does this thing called
> query slice. I don't like that but don't have time to rewrite all the
> backends. This is part of a deprecated class.

Ex in the java port, you'd see this implemented differently (still passing
tests). My hope is that we delete anorm once the java port has a release.

Cassandra should write directly to SpanStore as opposed to via the
deprecated SliceQuery approach. Would love your help with that.

codefromthecrypt, line 153: PS I may be totally wrong on interpreting how the queries work in relation
to service name. If SpanStoreSpec tests don't answer these questions, we
need to make them.

codefromthecrypt, line 153: I will add tests tonight or tomorrow, basically load 2 traces. One has root
span with service A calling child span with service B. The other is visa
versa.

I suspect the intended behavior is that one returns, not both!

yurishkuro:
Q1 - I see that it only does root spans, due to `|WHERE trace_id = id`. Which is probably a bad assumption anyway, nowhere do we say that root span ID must be equal to trace_id (a more accurate check would've been parent_id == nil or 0).

Q2 - the sql duration query checks for service_name == query.sn. In combination with the above, I conclude it's only limited to the services mentioned in the root span.

Q3 - I remember the slice query, iirc it's one of the reasons the limit param doesn't work correctly in combination queries. This is very hard to support in Cassandra, actually. I would go as far as forcing the UI to make the use choose between annotations query and duration queries. If we want to support both, then at least in Cassandra implementation it requires in-memory join of trace IDs from different index queries, and it's fairly hard to make that join work correctly with LIMIT.

It's fairly easy to implement correct LIMIT for serviceName {+span name} + duration combination in Cassandra, as it can all be stored in the duration index I showed above. As for the semantics of such query, I would go for span-level query, i.e. where both duration and service-name (and span name) must apply to the _same span_ in order for the trace ID to qualify. It seems like it would satisfy most of the queries. The one that it would not satisfy is finding long traces by service name whose span was not long.

zipkin-common/src/main/scala/com/twitter/zipkin/storage/SpanStore.scala
yurishkuro, line 153: If you have A->C->B, and any other trace involving B, I think the query by service B should always return all those traces if the B span was long. The owner of service B has no control over the shape of traces above B, so when they want to diagnose performance of B, they should see all traces containing B.

codefromthecrypt, line 153: OK so that basically means remove the root span condition. Makes sense.
Pull request on the way!

PR #821, opened 2015-11-07
Introduces LOCAL_COMPONENT("lc") binary annotation and example use-case

### LINKED ISSUE #808: Support local spans via a LOCAL_COMPONENT("lc") binary annotation

codefromthecrypt:
Local spans have been discussed at length, but have never been
formalized. This formalizes how to tag a span as local, including
rationale documentation. To help demonstrate the point, zipkin-web and
zipkin-query use this for bootstrap tracing.

Fixes #808

Here are example local spans from finatra and twitter-server bootstrap:

screen shot 2015-11-07 at 12 34 17 pm

screen shot 2015-11-07 at 12 35 28 pm

### DIFF
--- a/zipkin-query-service/src/main/scala/com/twitter/zipkin/builder/QueryServiceBuilder.scala
+++ b/zipkin-query-service/src/main/scala/com/twitter/zipkin/builder/QueryServiceBuilder.scala
@@ -15,16 +15,10 @@
  */
 package com.twitter.zipkin.builder
 
-import ch.qos.logback.classic.{Logger, Level}
-import com.twitter.conversions.time._
-import com.twitter.finagle.ListeningServer
-import com.twitter.finagle.stats.DefaultStatsReceiver
 import com.twitter.finagle.tracing.{DefaultTracer, NullTracer}
-import com.twitter.finagle.zipkin.thrift.{SpanStoreZipkinTracer, RawZipkinTracer}
-import com.twitter.ostrich.admin.RuntimeEnvironment
-import com.twitter.zipkin.query.ZipkinQueryServer
+import com.twitter.finagle.zipkin.thrift.{RawZipkinTracer, SpanStoreZipkinTracer}
+import com.twitter.zipkin.query.{BootstrapTrace, ZipkinQueryServer}
 import com.twitter.zipkin.storage.{DependencyStore, NullDependencyStore, SpanStore}
-import org.slf4j.LoggerFactory
 
 case class QueryServiceBuilder(override val defaultFinatraHttpPort: String = "0.0.0.0:9411",
                                override val defaultHttpPort: Int = 9901,
@@ -32,25 +26,24 @@ case class QueryServiceBuilder(override val defaultFinatraHttpPort: String = "0.
                                spanStore: SpanStore,
                                dependencies: DependencyStore = new NullDependencyStore,
                                override val defaultHttpServerName: String = "zipkin-query"
-                                ) extends ZipkinQueryServer(spanStore, dependencies) with
-                                          Builder[RuntimeEnvironment => ListeningServer] {
+                                ) extends ZipkinQueryServer(spanStore, dependencies) {
 
-  override def apply() = (runtime: RuntimeEnvironment) => {
-    LoggerFactory.getLogger(org.slf4j.Logger.ROOT_LOGGER_NAME)
-      .asInstanceOf[Logger].setLevel(Level.toLevel(logLevel))
-
-    /** If the span transport is set, trace accordingly, or disable tracing */
+  /** If the span transport is set, trace accordingly, or disable tracing. */
+  premain {
     DefaultTracer.self = sys.env.get("TRANSPORT_TYPE") match {
       case Some("scribe") => RawZipkinTracer(sys.env.get("SCRIBE_HOST").getOrElse("localhost"), sys.env.get("SCRIBE_PORT").getOrElse("1463").toInt)
-      case Some("http") => new SpanStoreZipkinTracer(spanStore, DefaultStatsReceiver.get)
+      case Some("http") => new SpanStoreZipkinTracer(spanStore, statsReceiver)
       case _ => NullTracer
     }
+  }
 
-    val defaultLookback = sys.env.get("QUERY_LOOKBACK").getOrElse(7.days.inMicroseconds.toString)
-    nonExitingMain(Array(
-      "-local.doc.root", "/",
-      "-zipkin.queryService.lookback", defaultLookback
-    ))
-    adminHttpServer
+  override def warmup() {
+    super.warmup()
+    BootstrapTrace.record("warmup")
+  }
+
+  override def postWarmup() {
+    super.postWarmup()
+    BootstrapTrace.complete()
   }
 }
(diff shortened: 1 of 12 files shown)


codefromthecrypt:
cc @yurishkuro @eirslett @spencergibb @dsyer @kristofa @michaelsembwever

codefromthecrypt:
Note that finagle instrumentation doesn't yet support Span.duration. The workaround until that's not the case is add a ClientRecv("cr") annotation. That's because finagle needs some signal to know the span is complete. https://github.com/twitter/finagle/blob/develop/finagle-zipkin/src/main/scala/com/twitter/finagle/zipkin/thrift/DeadlineSpanMap.scala#L110

yurishkuro:
Does this change require UI/JS changes? When I emit a span with LC only, without RPC-bounding annotations, the UI doesn't show it.

codefromthecrypt:
Make sure you are setting Span.timestamp, duration? The example of
bootstrap tracing displays...

zipkin-query-service/src/main/scala/com/twitter/zipkin/builder/QueryServiceBuilder.scala
```diff
@@ -31,23 +26,24 @@ case class QueryServiceBuilder(override val defaultFinatraHttpPort: String = "0.
...
-                                          Builder[RuntimeEnvironment => ListeningServer] {
+                                ) extends ZipkinQueryServer(spanStore, dependencies) {
 
-  override def apply() = (runtime: RuntimeEnvironment) => {
-    LoggerFactory.getLogger(org.slf4j.Logger.ROOT_LOGGER_NAME)
-      .asInstanceOf[Logger].setLevel(Level.toLevel(logLevel))
-
-    /** If the span transport is set, trace accordingly, or disable tracing */
+  /** If the span transport is set, trace accordingly, or disable tracing. */
+  def premain {
```
codefromthecrypt: @yurishkuro this was a bug!! maybe related to your inability to see spans. pushing a fix now.

codefromthecrypt:
rebased with a fix to a bug where the tracers weren't initializing.

@yurishkuro @kristofa please take a look. I'd like to add support to this to brave, but want to make sure folks are on-board before raising that.

codefromthecrypt:
last call for feedback, I'll merge in a couple hours and address anything else post-merge

yurishkuro:
looks good

yurishkuro:
@adriancole I pulled this, but still not able to see local spans, even though I do send ts and duration. Your screenshots above show "cr" for both examples of local spans. Did you try it without having any RPC annotations on the local span?

codefromthecrypt:
good point. so we have a test that shows spanstore doesn't require annotations, but maybe there's some special casing in the web process subverting this. Should be possible to reproduce by making json and then posting it to the query server. I'll have a look.

yurishkuro:
the json result from query service does not return that local span either, so it might be deeper than web layer

nishantshobhit:
brave user here: The trace_id generate for a span generated from the local tracer is different from other spans for the same trace. may be that's the problem here?

codefromthecrypt:
@nishantshobhit feel free to hop on openzipkin/zipkin even if the questions are about brave.

I have a working example of local tracer in the zipkin-java project. However, there's some important details about the implementation of span storage in brave you need to keep in mind. Look here for details: https://github.com/openzipkin/brave/blob/master/brave-core/src/main/java/com/github/kristofa/brave/LocalSpanState.java#L9

PR #823, opened 2015-11-07
force a single version of mustache

yurishkuro:
without this change, `gradlew idea` generates IntelliJ classpath with both 0.8 and 0.9 versions of mustache.


codefromthecrypt:
thanks!

PR #824, opened 2015-11-08
Clarifies fallbacks for fields as particulary CLIENT_ADDR needs to know

codefromthecrypt:


codefromthecrypt:
since this is documenting things actually the case in finagle, I'm gonna merge :P

PR #825, opened 2015-11-08
Settles on endTs and lookback query parameters

codefromthecrypt:
This settles on endTs as the only time boundary for queries, with a
configurable lookback of either the flag `zipkin.queryService.lookback`
or when running zipkin-query-service, the env variable `QUERY_LOOKBACK`.

We have two queries that need to support a notion of endTs and lookback:
trace and dependency. By supporting lookback, we allow administrators
control over how much data might be subject to search by default.

In the past, the dependency query documentation said it defaulted to
lookback 24hrs. However, the query passed actually applied to all data,
as 0 was passed as the startTs parameter. This changes the default to
be the same as trace query, 7 days, and overridable.

7 days was chosen as that's the length of the data TTL. We would expect
administrators to override this, for example to 2 days.

### DIFF
--- a/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormDependencyStore.scala
+++ b/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormDependencyStore.scala
@@ -16,38 +16,28 @@
 
 package com.twitter.zipkin.storage.anormdb
 
-import java.sql.Connection
-import java.util.concurrent.TimeUnit._
-
 import anorm.SqlParser._
 import anorm._
 import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
-import com.twitter.util.{Future, Time}
+import com.twitter.util.Future
 import com.twitter.zipkin.common.{Dependencies, DependencyLink}
 import com.twitter.zipkin.storage.DependencyStore
+import java.sql.Connection
 
-/**
- * Retrieve and store aggregate dependency information.
- *
- * The top annotations methods are stubbed because they're not currently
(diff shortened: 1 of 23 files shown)


codefromthecrypt:
@yurishkuro this is the change we spoke about for optimizing buckets for duration query

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormDependencyStore.scala
```diff
@@ -16,38 +16,28 @@
...
 import anorm._
 import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
-import com.twitter.util.{Future, Time}
+import com.twitter.util.Future
 import com.twitter.zipkin.common.{Dependencies, DependencyLink}
 import com.twitter.zipkin.storage.DependencyStore
+import java.sql.Connection
 
-/**
- * Retrieve and store aggregate dependency information.
```
codefromthecrypt, line 30: this doc was a combination of obvious and obsolete

yurishkuro:
Look good. The only comment is why not make lookback argument mandatory in the Dependency Store API, same as in SpanStore? Since the upstream query service already has defaulting mechanism for this field, it would make this API simpler. Otherwise it needs a sensible default, and endTs as default means we're looking at dependencies at a single point in time, which is weird.

codefromthecrypt:
The only reason is that there are so many more uses of QueryRequest in
tests vs uses of DependencyStore.getDependencies. Basically it would
"ugly up" probably a hundred or so call-sites.

Have a look and if you feel strongly about this, we can change after
the fact as it would break a lot of files in unrelated ways.

PR #827, opened 2015-11-09
793 introduce ensureSchema flag for cassandra span store factory

pszymczyk:


zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/CannotReadKeyspaceMetadataException.java
```diff
@@ -0,0 +1,10 @@
+package org.twitter.zipkin.storage.cassandra;
+
+import com.datastax.driver.core.Cluster;
+
+public class CannotReadKeyspaceMetadataException extends RuntimeException {
+
+    public CannotReadKeyspaceMetadataException(String keyspace, Cluster cluster) {
```
codefromthecrypt: I'd not expect anyone to recognize this. Please just throw IllegalStateException vs declaring single-use exceptions.

codefromthecrypt:
After addressing the only couple things. please squash and rebase into a single commit. http://rebaseandsqua.sh/

Here's a better commit message I've crafted just for you!

```
Makes cassandra schema commands optional

Before, using cassandra wasn't possible without escalated privileges. 
This was due to implicit use of schema commands. This change moves the
default to assume schema is absent. To run with less privileges, set the
environment variable `CASSANDRA_ENSURE_SCHEMA=true`. 

For users customizing `CassandraSpanStoreFactory` directly, this maps to
the flag `zipkin.store.cassandra.ensureSchema`, which defaults to false.

Fixes #793
```

zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
```diff
@@ -35,6 +35,7 @@ import scala.collection.JavaConverters._
 
 trait CassandraSpanStoreFactory {self: App =>
 
+  val ensureSchema          = flag[Boolean]  ("zipkin.store.cassandra.schema.ensureExists", false, "ensures schema exists")
```
codefromthecrypt: rename this flag to be same depth as others.. `zipkin.store.cassandra.ensureSchema`

codefromthecrypt:
updated the comment as I noticed the environment variable defaults to true, which was going to be a question!

OK Good job.. very little feedback for you. Thanks!

pszymczyk:
I'm not sure about suggested commit message: 'To run with less privileges, set the
environment variable `CASSANDRA_ENSURE_SCHEMA=true`, when you have less privileges you probably don't want to ensure schema and execute script on your DB so CASSANDRA_ENSURE_SCHEMA should be set to false?

codefromthecrypt:
@pawelszymczyk typo.. I screwed up that sentence because I wrote it before realizing the env variable was opposite of the twitter flag. Please correct!

PS if curious, default to true is mostly a laptop thing. Ex. our test docker image installs the schema directly.

codefromthecrypt:
LGTM will merge on green! Also, this change will go out today, probably in a few hours

pszymczyk:
Wow great to hear that! :D thanks for quick review :)

PR #828, opened 2015-11-09
Ensures duration queries all spans by service, not just root spans

codefromthecrypt:
Before, min and maxDuration queries applied only to root spans. Since a
call to a service can happen anywhere in the tree, this artifically
limited usefulness. Now, duration queries will apply to any span for a
given service.


codefromthecrypt:
@yurishkuro ptal

yurishkuro:
+1

PR #830, opened 2015-11-09
Fixes service name query for spans who only have binary annotations

codefromthecrypt:
Local spans are permitted to have a single binary annotation. Logic
needed to be adjusted to allow these to be queried by service, as
formerly the service of a binary annotation wasn't consistently
indexed or presented.


codefromthecrypt:
@yurishkuro PTAL

codefromthecrypt:
Here's the span that now works:

``` bash
$ curl -s localhost:9411/api/v1/spans -X POST -H 'Content-Type: application/json' -d '[
  {
    "traceId": "5a4a9e596c08e861",
    "name": "foo",
    "id": "5a4a9e596c08e861",
    "timestamp": 1447092995596000,
    "duration": 37000,
    "annotations": [ ],
    "binaryAnnotations": [
      {
        "key": "lc",
        "value": "twitter-server",
        "endpoint": {
          "serviceName": "zipkin-web",
          "ipv4": "127.0.0.1"
        }
      }
    ]
  }
]'
```

PR #831, opened 2015-11-10
Filters duration query by span name

codefromthecrypt:
Least code approach to ensure people can get relevant results when looking for a long "get" vs long "post"

### DIFF
--- a/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
+++ b/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
@@ -283,6 +283,7 @@ class AnormSpanStore(val db: DB,
 
   override protected def getTraceIdsByDuration(
     serviceName: String,
+    spanName: Option[String],
     minDuration: Long,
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
cc @yurishkuro

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
```diff
@@ -295,13 +296,15 @@ class AnormSpanStore(val db: DB,
         """SELECT t1.trace_id, start_ts
           |FROM zipkin_spans t1
           |JOIN zipkin_annotations t2 ON t1.trace_id = t2.trace_id AND t1.id = t2.span_id
-          |WHERE t2.endpoint_service_name = {service_name}
+          |WHERE (name = {name} OR {name} = '')
```
codefromthecrypt, line 299: literally same SQL as the name filter earlier in the file

codefromthecrypt:
manually verified also, by putting 30000 into the duration bin. with "all" names selected, bootstrap and get spans returned. selecting either refined as expected

PR #834, opened 2015-11-10
Ensures all span stores apply filters to results bound to endTs

### LINKED ISSUE #826: Cant see latest traces when many in the DB
How the UI works now seems to work:
Stage 1: Search
Here you specify the service name, endtime and limit

Stage 2: Ordering of results
Once there are results there is an ordering dropdown to order these results.

I have ~30.000 traces for a given service. These traces have been sent the last 1 month or so.
I want to see the latest traces, the ones send the last couple of days or so, but it is impossible.

If I try to get more than 1.000 traces in the UI the connection will timeout.
If I get 1000 traces then those seems to be the oldest so I can not see anything that has happen recently.
Looking at the UI in fact, I can not even know if Zipkin keeps receiving traces or no.

And Also I do not want to see 1000 traces in the UI.

Possible solutions:
- Add a start date to the UI
  or
- The ordering dropdown is there before clicking find and the ordering happens before the call to the DB

Or both.

codefromthecrypt:
Filtered traces are now returned in reverse insertion order. This is
because the primary search interface is a timeline view, looking back
from an end timestamp.

The in-memory and anorm span stores were the only ones applying request
filters against traces at the beginning of (endTs - lookback, endTs) vs
looking back from endTs. This created a poor experience where old traces
returned by default.

This enforces the lookback behavior and clarifies it by ordering traces
descending against endTs, A loose test was clarified to ensure limit
applies this way.

Besides trace ordering, which happens after traces return from the
backend, redis and cassandra implementations are unaffected, as they
already had the intended behavior.

Fixed #826

### DIFF
--- a/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
+++ b/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
@@ -179,11 +179,26 @@ class AnormSpanStore(val db: DB,
         .values.toList
         .map(CorrectForClockSkew)
         .map(ApplyTimestampAndDuration)
-        .sortBy(_.head) // sort traces by the first span
+        .sortBy(_.head)(Ordering[Span].reverse) // sort descending by the first span
     } finally {
       returnConn(conn, borrowTime, "getSpansByTraceIds")
     }
   }
+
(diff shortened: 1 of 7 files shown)


codefromthecrypt:
@jcarres-mdsol Thanks tons for reporting this!

yurishkuro:
+1

PR #836, opened 2015-11-10
Adds Local Address to span details when no annotations are present

codefromthecrypt:
It is valid to have a local span without any annotations. This presents
the local endpoint in the Key/Value section of Span detail, similarly to
"Client Address" and "Server Address".

Here's an example of this working, given a pre-canned span without any annotations.

screen shot 2015-11-10 at 11 28 55 am

``` bash
$ curl -s localhost:9411/api/v1/spans -X POST -H 'Content-Type: application/json' -d '[
  {
    "traceId": "5a4a9e596c08e861",
    "name": "foo",
    "id": "5a4a9e596c08e861",
    "timestamp": 1447092995596000,
    "duration": 37000,
    "annotations": [ ],
    "binaryAnnotations": [
      {
        "key": "lc",
        "value": "twitter-server",
        "endpoint": {
          "serviceName": "zipkin-web",
          "ipv4": "127.0.0.1"
        }
      }
    ]
  }
]'
```


yurishkuro:
+1

PR #837, opened 2015-11-10
added cache control to public css and js

jcarres-mdsol:
I'm not scala programmer but looking at this
https://github.com/twitter/finagle/blob/bf4b0ea8b4bb34f4cb98fa6da06723bc58635487/finagle-http/src/test/scala/com/twitter/finagle/http/MessageTest.scala

I think it should work.

The UI is slow. Looking at the cause it is because we have lots of css and js and they are sent everytime.
@adriancole  please run tests xD

### DIFF
--- a/zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
+++ b/zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
@@ -12,6 +12,7 @@ import com.twitter.zipkin.common._
 import com.twitter.zipkin.json._
 import com.twitter.zipkin.web.mustache.ZipkinMustache
 import com.twitter.zipkin.{Constants => ZConstants}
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
yep looks quite sensible.

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
```diff
@@ -57,6 +57,7 @@ class Handlers(mustacheGenerator: ZipkinMustache, queryExtractor: QueryExtractor
 
     def apply(response: Response) {
       response.setContentType(typ)
+      response.cacheControl = 10.minutes
```
codefromthecrypt, line 61: you'll need to insert this add this at line 4 to resolve 10.minutes

``` scala
import com.twitter.conversions.time._
```

PR #839, opened 2015-11-10
Truncates cassandra test data as opposed to dropping the keyspace

### LINKED ISSUE #816: Cassandra micro-integration tests are *really* slow
Cassandra tests are taking orders of minutes to complete, even on a fast laptop. Since a clean build and all other tests pass in <1 minute, this is a big deal and discourages good behavior. Ex. when builds are slow, people are less likely to want to do revisions as the penalty for doing so is long wait times.

Considering these tests are in-memory, I'm not sure why these objects would take so long to start up. We may be able to configure them better, or worst, conditionally run cassandra tests in CI.

For those who are doing big refactors and regularly running clean builds, you may just want to delete or skip these tests until your work is complete.

```
zipkin-cassandra/src/test/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStoreSpec.scala zipkin-cassandra/src/test/scala/com/twitter/zipkin/storage/cassandra/CassandraDependencyStoreSpec.scala
``
```

codefromthecrypt:
Cassandra tests were taking several minutes to complete. @yurishkuro
noticed this was due to redundant keyspace dropping. This changes
behavior to truncate instead.

Fixes #816

### DIFF
--- a/zipkin-cassandra/src/test/scala/com/twitter/zipkin/storage/cassandra/CassandraDependencyStoreSpec.scala
+++ b/zipkin-cassandra/src/test/scala/com/twitter/zipkin/storage/cassandra/CassandraDependencyStoreSpec.scala
@@ -1,35 +1,14 @@
 package com.twitter.zipkin.storage.cassandra
 
-import java.util.Collections
-
-import com.datastax.driver.core.Cluster
 import com.twitter.util.Await._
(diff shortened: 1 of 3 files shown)


yurishkuro:
wow!  +1

codefromthecrypt:
yep. `./gradlew clean check` on my laptop is now back to a tolerable 1.5mins

PR #844, opened 2015-11-12
Support duration-based queries in Cassandra storage

yurishkuro:
Complements https://github.com/openzipkin/zipkin/pull/820

### DIFF
--- a/zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
+++ b/zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
@@ -24,17 +24,26 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.nio.ByteBuffer;
+import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
-import java.util.LinkedHashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.function.IntFunction;
+import java.util.function.LongFunction;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.LongStream;
+import java.util.stream.StreamSupport;
 
 public final class Repository implements AutoCloseable {
 
@@ -53,6 +62,10 @@ public final class Repository implements AutoCloseable {
     private static final long WRITTEN_NAMES_TTL
             = Long.getLong("zipkin.store.cassandra.internal.writtenNamesTtl", 60 * 60 * 1000);
 
+    // Time window covered by a single bucket of the Span Duration Index, in seconds. Default: 1hr
+    private static final long DURATION_INDEX_BUCKET_WINDOW_SECONDS
+            = Long.getLong("zipkin.store.cassandra.internal.durationIndexBucket", 60 * 60);
+
     private final Session session;
     private final PreparedStatement selectTraces;
     private final PreparedStatement insertSpan;
@@ -68,6 +81,8 @@ public final class Repository implements AutoCloseable {
     private final PreparedStatement insertTraceIdBySpanName;
     private final PreparedStatement selectTraceIdsByAnnotations;
     private final PreparedStatement insertTraceIdByAnnotation;
+    private final PreparedStatement selectTraceIdsBySpanDuration;
+    private final PreparedStatement insertTraceIdBySpanDuration;
     private final Map<String,String> metadata;
     private final ProtocolVersion protocolVersion;
 
@@ -205,6 +220,28 @@ public final class Repository implements AutoCloseable {
                     .value("ts", QueryBuilder.bindMarker("ts"))
                     .value("trace_id", QueryBuilder.bindMarker("trace_id"))
                     .using(QueryBuilder.ttl(QueryBuilder.bindMarker("ttl_"))));
+
+        selectTraceIdsBySpanDuration = session.prepare(
+                QueryBuilder.select("d", "ts", "tid")
+                    .from("span_duration_index")
+                    .where(QueryBuilder.eq("s", QueryBuilder.bindMarker("service_name")))
+                        .and(QueryBuilder.eq("sp", QueryBuilder.bindMarker("span_name")))
+                        .and(QueryBuilder.eq("b", QueryBuilder.bindMarker("time_bucket")))
+                        .and(QueryBuilder.lte("d", QueryBuilder.bindMarker("max_duration")))
+                        .and(QueryBuilder.gte("d", QueryBuilder.bindMarker("min_duration")))
+                    .orderBy(QueryBuilder.desc("d")));
+
+        insertTraceIdBySpanDuration = session.prepare(
+                QueryBuilder
+                        .insertInto("span_duration_index")
+                        .value("s", QueryBuilder.bindMarker("service_name"))
+                        .value("sp", QueryBuilder.bindMarker("span_name"))
+                        .value("b", QueryBuilder.bindMarker("bucket"))
+                        .value("d", QueryBuilder.bindMarker("duration"))
+                        .value("ts", QueryBuilder.bindMarker("ts"))
+                        .value("tid", QueryBuilder.bindMarker("trace_id"))
+                        .using(QueryBuilder.ttl(QueryBuilder.bindMarker("ttl_"))));
+
     }
 
     /**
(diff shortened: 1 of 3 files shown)


zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -24,17 +24,26 @@
...
-import java.util.LinkedHashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.function.IntFunction;
```
codefromthecrypt, line 41: cool! first use of java 8!

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -205,6 +220,28 @@ public Repository(String keyspace, Cluster cluster, Boolean ensureSchema) {
                     .value("ts", QueryBuilder.bindMarker("ts"))
                     .value("trace_id", QueryBuilder.bindMarker("trace_id"))
                     .using(QueryBuilder.ttl(QueryBuilder.bindMarker("ttl_"))));
+
+        selectTraceIdsBySpanDuration = session.prepare(
+                QueryBuilder.select("d", "ts", "tid")
+                    .from("span_duration_index")
+                    .where(QueryBuilder.eq("s", QueryBuilder.bindMarker("service_name")))
```
codefromthecrypt, line 227: is there a space or performance advantage to using single-character fields in cassandra?

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -707,6 +744,131 @@ private String debugInsertTraceIdByAnnotation(ByteBuffer annotationKey, long tim
...
+            duration = row.getLong("d");
+            timestamp = row.getLong("ts");
+        }
+        public String toString() {
+            return String.format("trace_id=%d, duration=%d, timestamp=%d", trace_id, duration, timestamp);
+        }
+    }
+
+    /**
+     * @return Returns a map of trace id -> timestamp
```
codefromthecrypt: supernit, but usually I inline doc like this: `/** Returns a map of trace id -> timestamp */`

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -707,6 +744,131 @@ private String debugInsertTraceIdByAnnotation(ByteBuffer annotationKey, long tim
...
+        int startBucket = durationIndexBucket(startTs);
+        int endBucket = durationIndexBucket(endTs);
+        try {
+            if (startBucket > endBucket) {
+                throw new IllegalArgumentException("Start bucket (" + startBucket + ") > end bucket (" + endBucket + ")");
+            }
+            IntFunction<ListenableFuture<List<DurationRow>>> oneBucketQuery = (bucket) -> {
+                BoundStatement bound = selectTraceIdsBySpanDuration.bind()
+                        .setString("service_name", serviceName)
+                        .setString("span_name", spanName == null ? "" : spanName)
```
codefromthecrypt, line 774: this can be addressed later, but sometimes spans come in as "" or "unknown" and later merge with a valid span name. Does this impact anything?

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -707,6 +744,131 @@ private String debugInsertTraceIdByAnnotation(ByteBuffer annotationKey, long tim
...
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug(debugSelectTraceIdsByDuration(serviceName, spanName, minDuration, maxDuration, startTs, endTs, limit));
+                }
+                return Futures.transform(
+                        session.executeAsync(bound),
+                        (ResultSet rs) -> {
+                            Iterable<Row> it = rs::iterator;
+                            return StreamSupport
+                                    .stream(it.spliterator(), false)
+                                    .map(DurationRow::new)
```
codefromthecrypt, line 791: neat

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -707,6 +744,131 @@ private String debugInsertTraceIdByAnnotation(ByteBuffer annotationKey, long tim
...
+            List<ListenableFuture<List<DurationRow>>> futures = IntStream
+                    .rangeClosed(startBucket, endBucket)
+                    .mapToObj(oneBucketQuery)
+                    .collect(Collectors.toList());
+
+            return Futures.transform(Futures.successfulAsList(futures),
+                    (List<List<DurationRow>> input) -> {
+                        return input.stream()
+                                .flatMap(Collection::stream)
+                                .collect( // bloody IntelliJ can't infer types
```
codefromthecrypt, line 808: :)

zipkin-cassandra-core/src/main/resources/cassandra-schema-cql3.txt
```diff
@@ -51,3 +51,15 @@ CREATE TABLE IF NOT EXISTS zipkin.traces (
     span      blob,
     PRIMARY KEY (trace_id, ts, span_name)
 ) WITH compaction = {'class': 'org.apache.cassandra.db.compaction.DateTieredCompactionStrategy', 'max_sstable_age_days': '1'};
+
+// This table uses bigint to store timestamp in microseconds, as Cassandra's timestamp is apparently limited to milliseconds
+CREATE TABLE IF NOT EXISTS zipkin.span_duration_index (
+    s      text,      // service name
```
codefromthecrypt, line 59: all of these keys are different than the prior ones in the same file, which could lead to confusion. I'd suggest the following, if this is ideal.
1. add a line noting why single-character preference is used, as it obfuscates the schema. In other words, make it obvious why we are being cryptic (ex for some performance reason?)
2. add a TODO: to change the other schema to the same naming convention.

codefromthecrypt:
The build is blowing up, guessing something around java8. I can help debug this later, if the cause isn't obvious.

Besides that, I'm happy with helping to support this code. I'd like some clarity on the naming choices, but otherwise LGTM

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
yurishkuro, line 227: shockingly, yes.  https://issues.apache.org/jira/browse/CASSANDRA-4175

It may be solved in v. 2.2+, so maybe I should put it back to long names.

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
yurishkuro, line 774: Can't do much about that, if the correct span name is not available/incorrect at the time of the insert. It will only get indexed by the service name.

yurishkuro:
I have built it on Mac with

```
$ java -version
java version "1.8.0_45"
Java(TM) SE Runtime Environment (build 1.8.0_45-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.45-b02, mixed mode)
```

The CI job runs with "1.8.0_31", maybe that's the reason for compiler failure. Can we bump java version in CI?

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
yurishkuro, line 227: Just checked, according to my teammates who use Cassandra for metrics, it's not solved in 2.2 either.

codefromthecrypt, line 227: OK mind just noting this as a schema declaration comment? Would help answer
future questions.

codefromthecrypt:
mind squashing this into a single commit when you rebase on top of the travis JDK bump?

codefromthecrypt:
I know you know how, but I like pasting this.. http://rebaseandsqua.sh/ :)

zipkin-cassandra-core/src/main/resources/cassandra-schema-cql3.txt
```diff
@@ -51,3 +51,17 @@ CREATE TABLE IF NOT EXISTS zipkin.traces (
     span      blob,
     PRIMARY KEY (trace_id, ts, span_name)
 ) WITH compaction = {'class': 'org.apache.cassandra.db.compaction.DateTieredCompactionStrategy', 'max_sstable_age_days': '1'};
+
+// This table uses bigint to store timestamp in microseconds, as Cassandra's timestamp is apparently limited to milliseconds.
+// The table uses short column names for increased efficiency of Cassandra storage.
```
codefromthecrypt, line 56: thanks!

codefromthecrypt:
I'll cherry-pick this in

codefromthecrypt:
closed via https://github.com/openzipkin/zipkin/commit/9603ca81f22d1b8e8ba17b9f0663603184d154dd

Thanks tons!

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
michaelsembwever, line 227: do you know this to be an actual problem in this use-case?

you are talking about very short lived objects
 (it's not like there's lots of different long column names…)

michaelsembwever:
nice work @yurishkuro

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
yurishkuro, line 227: I am going by the experience of our metrics team. They have tables of similar complexity, if not simpler, and they observed a significant impact of shortening column names on the overall performance of the nodes.  The issue has been addressed in Cassandra 3.0, according to the history in the JIRA above.

yurishkuro, line 227: here's some anecdotal evidence (since it's hard to measure exactly)

> ... as an example, for data file sizes > 24h (and thus at the max level of compaction)
> 168829777409 with long column names
> 31931936664 with short column names

michaelsembwever, line 227: I'm aware of the problem and that it affects both heap and sstable size.

In the heap it's not so much about space but objection creation/deletion, IIRC.
And in the sstable it is kinda crazy that every column value is labeled with its column name in every single row through the file. (It's a hangup from when Cassandra used thrift and was truly schema-free).

And I've been aware that it's getting addressed between CASSANDRA-8099 and CASSANDRA-7447

But those numbers above look surprising for what you say is off a simple schema.
Just how long were the long column names? And was compression enabled (which usually makes up for the difference)?

Otherwise it's evidence on sstable size, not performance.
And if we don't have a problem with that (we want to encourage people to scale out rather than scale up) it can well be a better optimisation of developer time to use meaningful column names.

TL;DR i'd love to hear more around the details of from what your metrics team discovered.

yurishkuro, line 227: Yes, this was with compression. The schema is like below. Only the n(name) field would have longer strings, others are usually a single number, which become comparable in space usage to column names if those are fully spelled out.

The x5 size difference may not be directly attributed to column names, but that's the best measure they had. I don't know how much of intellectual burden short names really are, there are like 10 lines of code that depend on them. Taking extra disk space affects performance in various ways, in streaming, further compactions, etc.

But I am not strongly against changing this to long names.

```
CREATE TABLE events (
  n text,
  rt bigint,
  t bigint,
  v blob,
  PRIMARY KEY ((n, rt), t)
) WITH compression={'sstable_compression': 'LZ4Compressor'};
```

zipkin-cassandra-core/src/main/resources/cassandra-schema-cql3.txt
michaelsembwever, line 56: table comments can also go into the table's schema, to help developers understand it when using `cqlsh`.

For example, 

```
CREATE TABLE … ()
    WITH comment =' This table …';
```

If using the short column names (which I suspect hasn't any real benefit but rather just overhead on developer productivity) then i would add the column name comments in as well. Transparency of data is gold after all.

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
michaelsembwever, line 227: But what were the original length of the columns names?
What are we actually comparing here, single-char column names vs what?

> Only the n(name) field would have longer strings, others are usually a single number, which become comparable in space usage to column names if those are fully spelled out.

Those columns are bigints, their size are fixed.

A 5x difference doesn't sound right, given that compression will do a good job with repeated column names in such a small table.

> The x5 size difference may not be directly attributed to column names…

ok, this kinda worries me. How many sstables were there in each case? Was there variations in data generated too?

> …there are like 10 lines of code that depend on them.

That's fine if you're coming to it from within the code, but that's not always the case.

> But I am not strongly against changing this to long names.

What actually is a long name though? is 5 characters long? there's nothing empirical so far to go off here.

I would like to get more empirical data down, because otherwise it smells like premature optimisation.

I would not bother in changing this, but it would be nice to orientate ourselves so we know what practice we should be pushing.

codefromthecrypt, line 227: I agree that if we are making awkward decisions based on size, these tests
should be transparent because there are a bunch of ways we can screw that
up without considering character length.

Passing thought, but right now Travis tests don't pass on current impl
anyway. Should we have another pass, addressing cryptic names and improving
performance?

codefromthecrypt, line 227: ps issue resolved via https://github.com/openzipkin/zipkin/pull/861 thx @yurishkuro

PR #848, opened 2015-11-13
Completes documentation of all core thrift types

codefromthecrypt:
Also makes sure that the scala model fields have the same summary text as the thrift

### DIFF
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
@@ -20,25 +20,35 @@ import com.twitter.zipkin.Constants
 import com.twitter.zipkin.util.Util._
 
 /**
- * A span represents one RPC request. A trace is made up of many spans.
+ * A trace is a series of spans (often RPC calls) which form a latency tree.
  *
- * A span can contain multiple annotations, some are always included such as
- * Client send -> Server received -> Server send -> Client receive.
+ * <p/>Spans are usually created by instrumentation in RPC clients or servers, but can also
+ * represent in-process activity. Annotations in spans are similar to log statements, and are
+ * sometimes created directly by application developers to indicate events of interest, such as a
+ * cache miss.
  *
- * Some are created by users, describing application specific information,
- * such as cache hits/misses.
+ * <p/>The root span is where [[traceId]] = [[id]] and [[parentId]] is empty. The root span is
+ * usually the longest interval in the trace, starting with [[timestamp]] and ending with
+ * [[timestamp]] + [[duration]].
  *
- * @param traceId random long that identifies the trace, will be set in all spans in this trace
- * @param name name of span, can be rpc method name for example, in lowercase.
- * @param id random long that identifies this span
- * @param parentId reference to the parent span in the trace tree
- * @param timestamp epoch microseconds of the start of this span. None when a partial span.
- * @param duration microseconds comprising the critical path, if known.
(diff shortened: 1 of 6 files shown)


zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
```diff
@@ -20,25 +20,35 @@ import com.twitter.zipkin.Constants
...
- * A span can contain multiple annotations, some are always included such as
- * Client send -> Server received -> Server send -> Client receive.
+ * <p/>Spans are usually created by instrumentation in RPC clients or servers, but can also
+ * represent in-process activity. Annotations in spans are similar to log statements, and are
+ * sometimes created directly by application developers to indicate events of interest, such as a
+ * cache miss.
  *
- * Some are created by users, describing application specific information,
- * such as cache hits/misses.
+ * <p/>The root span is where [[traceId]] = [[id]] and [[parentId]] is empty. The root span is
```
yurishkuro, line 30: What specifically depends on `traceId == id` condition for the root span?  It seems very artificial, why is it not enough to have empty parent ID?

yurishkuro:
LGTM, but I would really like to understand and ideally remove the restriction `traceId == spanId` for root spans.

zipkin-common/src/main/scala/com/twitter/zipkin/common/Span.scala
codefromthecrypt, line 30: If you want to change this, we should log an issue because all core instrumentation currently set  traceid = spanid for root spans (ex finagle, ruby tracer, brave, as well zipkin has logic around this)

Basically, I'm not opposed to changing this, but it is in fact a change, and likely has impact. I don't want to document something that isn't status quo and also breaks things.

codefromthecrypt:
@yurishkuro as noted inline, this is documentation, not a change. Basically, I've found much stronger signal that in zipkin root span is traceid = spanid, vs just absent parent id. Documenting how things work doesn't mean we can't change it.

Changing it will certainly impact code anyway, at least ours, as we need to test things like tree construction don't depend on this. Moreover, I don't know off-hand why traceid=spanid, or if that was just to save a random number generator. If I were to start figuring it out, it would be like many things: searching through git logs of zipkin and finagle, asking people, then dealing with test breaks or missing tests. In other words, this type of issue is better dealt with as a top-level change, than a side bar on documentation polish :)

@theatrus do you remember anything about traceid=spanid?

codefromthecrypt:
ps fwiw I would like to understand traceid=spanid, too

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
```diff
@@ -220,16 +242,28 @@ enum AnnotationType { BOOL, BYTES, I16, I32, I64, DOUBLE, STRING }
  * you can see the different points of view, which often help in debugging.
  */
 struct BinaryAnnotation {
+  /**
+   * Name used to lookup spans, such as "http.uri" or "finagle.version".
+   */
   1: string key,
+  /**
+   * Serialized thrift bytes, in TBinaryProtocol format.
```
codefromthecrypt, line 250: @yurishkuro ps I finally hunted down what that endian stuff was about!! little-endian wasn't added until thrift 0.9.3!

codefromthecrypt:
PS parked the root span.id=traceid here: https://github.com/openzipkin/zipkin/issues/849

PR #851, opened 2015-11-14
Remove constraint that root span's trace id = span id

### LINKED ISSUE #849: Investigate impact of relaxing root span id=traceid
At least finagle, ruby tracer, and brave set span.id=traceid on the root span. We've not audited or formally tested relaxing this to just parent id. Before a change is attempted, it might be nice to understand what people know is relying on this, or if span.id=traceid is just a way to save a random number generator.

If we support root span is just absence of parent_id, we would want to add tests to ensure that works in spanstore, aggregators, and anything that generates a tree.

codefromthecrypt:
This removes any confusing text which might indicate root span's trace
id must equal its span id.

Fixes #849


yurishkuro:
+1

So despite the old docs/comments, doesn't seem like there was any actual code that required traceId==spanId for root span?

codefromthecrypt:
Nope! no code change needed, so yeah glad about that.

Issue #843, opened 2015-11-12
Add a license scan?


mstine:
Looking ahead to potentially packaging the code into Spring Cloud Services at Pivotal, I'm wondering if we could add the Gradle License Plugin (https://github.com/hierynomus/license-gradle-plugin) to determine if there are any poteintal poisonous dependencies in Zipkin. @adriancole

codefromthecrypt:
SGTM

codefromthecrypt:
@abesto if you're interested in this, lemme know.

abesto:
Totally. Licensing is something I'm recently very interested in; if no-one picks it up, I'll have a look this weekend.

abesto:
Wow, rebasing is noisy.

PR #852, opened 2015-11-14
Verify that all dependencies use friendly licenses

abesto:
As suggested in #843, this change checks the licenses of all dependencies transitively against a white-list. I've assumed that everything currently in use is OK; see the white-list near the bottom of the diff for a full list. Note especially that there are dependencies that don't specify their license.

Other than that, the only interesting thing in here is the insane amount of license aliasing. Review of that is especially appreciated. To check that all is well, run the task `downloadLicenses`, open `./build/reports/license/license-dependency.html`, and apply legal knowledge :) Removing the aliases for the duration of the test can maybe help that process.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,6 +9,7 @@ buildscript {
 plugins {
     id 'com.jfrog.bintray' version '1.4'
     id 'net.researchgate.release' version '2.2.1'
+    id 'com.github.hierynomus.license' version '0.12.1'
 }
 
 allprojects {
@@ -257,3 +258,4 @@ task dependencyReport {
     }
 }
 
+apply from: "${rootDir}/gradle/verify-licenses.gradle"
--- a/gradle/verify-licenses.gradle
+++ b/gradle/verify-licenses.gradle
@@ -1,0 +1,80 @@
+/**
+ * Gradle plugin used to verify that all dependencies of a project use allowed licenses.
+ * Usage:
+ *   apply from: "${rootDir}/gradle/verify-licenses.gradle"
+ *
+ * The list of allowed licenses can be modified via `licenseWhiteList` from project definitions.
+ */
+
+ext.licenseWhiteList = [
+		'No license found',
+		'Public Domain',
+		'MIT License',
+		'Apache License, Version 2.0',
+		'BSD License',
+		'BSD 3-Clause',
+		'BSD-Style',
+		'GNU Lesser General Public License',
+		'Common Public License Version 1.0',
+		'LGPL-2.1',
+		'CC0 1.0 Universal',
+		'Eclipse Public License - v 1.0'
+]
+
+// Generate license report
+downloadLicenses {
+	includeProjectDependencies = true
+	dependencyConfiguration = 'compile'
+
+	ext.apache = license('Apache License, Version 2.0', 'http://opensource.org/licenses/Apache-2.0')
+	ext.mit = license('MIT License', 'http://www.opensource.org/licenses/mit-license.php')
+	ext.bsd = license('BSD License', 'http://www.opensource.org/licenses/bsd-license.php')
+	ext.bsd3Clause = license('BSD 3-Clause', 'http://opensource.org/licenses/BSD-3-Clause')
+
+	aliases = [
+			(apache) : [
+					'The Apache Software License, Version 2.0', 'Apache License Version 2.0', 'Apache License, Version 2.0',
+					'Apache 2', 'Apache 2.0', 'Apache License 2.0', 'Apache-2.0',
+					license('Apache License', 'http://www.apache.org/licenses/LICENSE-2.0'),
+					license('Apache Software Licenses', 'http://www.apache.org/licenses/LICENSE-2.0.txt'),
+					license('Apache', 'http://www.opensource.org/licenses/Apache-2.0')
+			],
+			(mit) : ['The MIT License'],
+			(bsd) : ['BSD', 'Berkeley Software Distribution (BSD) License',
+					 license('New BSD License', 'http://www.opensource.org/licenses/bsd-license.php')
+			],
+			(bsd3Clause) : [
+					license('BSD 3-clause', 'http://opensource.org/licenses/BSD-3-Clause'),
+					license('BSD 3-Clause', 'http://www.scala-lang.org/license.html')
+			]
+	]
+
+	licenses = [
+			(group('org.apache.httpcomponents')): apache
+	]
+}
+
+// Verify that all dependency licenses are ones we like
+task verifyLicenses {
+	description "Verify that all dependencies use white-listed licenses."
+	dependsOn ':downloadLicenses'
+
+	doLast {
+		def xml = new XmlParser().parse('build/reports/license/license-dependency.xml')
+		def fail = false
+		xml.each { license ->
+			if (!licenseWhiteList.contains(license.@name)) {
+				def depStrings = []
+				license.dependency.each { depStrings << it.text() }
+				logger.error(
+						"License \"${license.@name}\" is not on the list of allowed licenses. " +\
+                        "The dependencies using it: ${depStrings}")
+				fail = true
+			}
+		}
+		if (fail) {
+			throw new GradleException("License verification failed.")
+		}
+	}
+}
+check.dependsOn verifyLicenses
--- a/zipkin-collector-service/build.gradle
+++ b/zipkin-collector-service/build.gradle
@@ -31,3 +31,4 @@ dependencies {
 
 sourceSets.main.resources.srcDirs += ['config']
 apply from: "${rootDir}/gradle/idea-mark-config-dir-as-resource.gradle"
+apply from: "${rootDir}/gradle/verify-licenses.gradle"
--- a/zipkin-query-service/build.gradle
+++ b/zipkin-query-service/build.gradle
@@ -30,4 +30,4 @@ dependencies {
 
 sourceSets.main.resources.srcDirs += ['config']
 apply from: "${rootDir}/gradle/idea-mark-config-dir-as-resource.gradle"
-
+apply from: "${rootDir}/gradle/verify-licenses.gradle"
--- a/zipkin-web/build.gradle
+++ b/zipkin-web/build.gradle
@@ -28,3 +28,5 @@ dependencies {
     testCompile project(':zipkin-scrooge')
     testCompile 'com.squareup.okhttp:mockwebserver:2.5.0'
 }
+
+apply from: "${rootDir}/gradle/verify-licenses.gradle"


abesto:
CC @mstine @adriancole

build.gradle
```diff
@@ -257,3 +258,73 @@ task dependencyReport {
...
+    ext.apacheTwo = license('Apache License, Version 2.0', 'http://opensource.org/licenses/Apache-2.0')
+    ext.mit = license('MIT License', 'http://www.opensource.org/licenses/mit-license.php')
+    ext.bsd = license('BSD License', 'http://www.opensource.org/licenses/bsd-license.php')
+    ext.bsd3Clause = license('BSD 3-Clause', 'http://opensource.org/licenses/BSD-3-Clause')
+
+    aliases = [
+            (apacheTwo) : [
+                    'The Apache Software License, Version 2.0', 'Apache License Version 2.0', 'Apache License, Version 2.0',
+                    'Apache 2', 'Apache 2.0', 'Apache License 2.0', 'Apache-2.0',
+                    license('Apache License', 'http://www.apache.org/licenses/LICENSE-2.0'),
```
abesto: Added URL to make sure we don't alias earlier Apache licenses to Apache 2

build.gradle
```diff
@@ -257,3 +258,73 @@ task dependencyReport {
...
+                    license('Apache Software Licenses', 'http://www.apache.org/licenses/LICENSE-2.0.txt'),
+                    license('Apache', 'http://www.opensource.org/licenses/Apache-2.0')
+            ],
+            (mit) : ['The MIT License'],
+            (bsd) : ['BSD', 'Berkeley Software Distribution (BSD) License',
+                     license('New BSD License', 'http://www.opensource.org/licenses/bsd-license.php')
+            ],
+            (bsd3Clause) : [
+                    license('BSD 3-clause', 'http://opensource.org/licenses/BSD-3-Clause'),
```
abesto: URL is the same, but I want the `C` capitalized :)

build.gradle
```diff
@@ -257,3 +258,73 @@ task dependencyReport {
     }
 }
 
+// License reporting
+downloadLicenses {
+    includeProjectDependencies = true
+
+    ext.apacheTwo = license('Apache License, Version 2.0', 'http://opensource.org/licenses/Apache-2.0')
```
codefromthecrypt: nit: I'd call this apache as it is a bolder highlight when not 2.0

build.gradle
```diff
@@ -257,3 +258,73 @@ task dependencyReport {
...
+    licenses = [
+            (group('org.apache.httpcomponents')): apacheTwo
+    ]
+}
+
+task verifyLicenses {
+    description "Verify that all dependencies use white-listed licenses."
+    dependsOn ':downloadLicenses'
+
+    ext.whiteList = [
```
codefromthecrypt: is there a way to apply this only to compile deps? I'm concerned w/ GPL here

codefromthecrypt:
Thanks. good start. I suspected there'd be a crapload of licenses (because crapload of dependencies).

I think it is most important to know the licenses "per binary" like, collector, web, query.

Do you mind running this and place the example html output in a comment, image a pastebin? That will lower the barrier to review for folks who don't have gradle setup. (Also, you could raise this as a openzipkin branch, too)

build.gradle
abesto: Addressed with `dependencyConfiguration = 'compile'` in `downloadLicenses`. Unfortunately we still have logback with LGPL and `mysql-connector-java` with GPL (latter one used for anorm).

abesto:
We should be able to do per subproject (=== per binary) simply, I'll give it a go today.

I've added the license reports for review, easy way to check: http://htmlpreview.github.io/?https://github.com/abesto/zipkin/blob/license-check/build/reports/license/license-dependency.html

codefromthecrypt:
I'll raise a pull request to switch out mysql driver for mariadb. There's an exception to the GPL clause, but still unnecessarily scary: https://www.mysql.com/about/legal/licensing/foss-exception/

abesto:
SGTM. In the meanwhile, the check is now run separately for the collector, query and web services, as well as for the root project, just in case.

codefromthecrypt:
https://github.com/openzipkin/zipkin/pull/853 wrt killing the GPL dep

codefromthecrypt:
hehe one last whack-a-mole! "verifyLicensesLicense "LGPL-2.1" is not on the list of allowed licenses. The dependencies using it: [mariadb-java-client-1.3.0.jar"

build/reports/license/dependency-license.html
```diff
@@ -0,0 +1,1358 @@
...
+      </tr>
+      <tr>
+        <td class='dependencies'>org.iq80.snappy:snappy:0.4</td>
+        <td class='licenseName'>snappy-0.4.jar</td>
+        <td class='licenseName'>Apache License, Version 2.0</td>
+        <td class='license'>
+          <a href='http://opensource.org/licenses/Apache-2.0'>Show license agreement</a>
+        </td>
+      </tr>
+      <tr>
```
codefromthecrypt: remove this now :)

build/reports/license/dependency-license.xml
```diff
@@ -0,0 +1,684 @@
...
+  </dependency>
+  <dependency name='org.scala-lang.modules:scala-xml_2.11:1.0.4'>
+    <file>scala-xml_2.11-1.0.4.jar</file>
+    <license name='BSD 3-Clause' url='http://opensource.org/licenses/BSD-3-Clause' />
+  </dependency>
+  <dependency name='org.iq80.snappy:snappy:0.4'>
+    <file>snappy-0.4.jar</file>
+    <license name='Apache License, Version 2.0' url='http://opensource.org/licenses/Apache-2.0' />
+  </dependency>
+  <dependency name='mysql:mysql-connector-java:5.1.37'>
```
codefromthecrypt: remove

build/reports/license/license-dependency.html
```diff
@@ -0,0 +1,346 @@
...
+            <li>stats-0.0.114.jar</li>
+            <li>client-0.0.79.jar</li>
+            <li>candidate-0.0.84.jar</li>
+            <li>quantity-0.0.98.jar</li>
+            <li>zipkin-redis-1.23.4-SNAPSHOT.jar</li>
+            <li>zipkin-collector-1.23.4-SNAPSHOT.jar</li>
+          </ul>
+        </td>
+      </tr>
+      <tr>
```
adriancole: remove

build/reports/license/license-dependency.html
```diff
@@ -0,0 +1,346 @@
...
+          <a href='http://www.eclipse.org/legal/epl-v10.html'>License agreement</a>
+        </td>
+        <td class='dependencies'>
+          <ul>
+            <li>logback-core-1.1.3.jar</li>
+            <li>logback-classic-1.1.3.jar</li>
+          </ul>
+        </td>
+      </tr>
+      <tr>
```
codefromthecrypt: remove

build/reports/license/license-dependency.xml
```diff
@@ -0,0 +1,198 @@
...
+    <dependency>inject-modules_2.11-2.1.1.jar</dependency>
+    <dependency>util-logging_2.11-6.29.0.jar</dependency>
+    <dependency>finagle-core_2.11-6.30.0.jar</dependency>
+    <dependency>inject-request-scope_2.11-2.1.1.jar</dependency>
+  </license>
+  <license name='Eclipse Public License - v 1.0' url='http://www.eclipse.org/legal/epl-v10.html'>
+    <dependency>logback-core-1.1.3.jar</dependency>
+    <dependency>logback-classic-1.1.3.jar</dependency>
+  </license>
+  <license name='The GNU General Public License, Version 2' url='http://www.gnu.org/licenses/old-licenses/gpl-2.0.html'>
```
codefromthecrypt: remove

build/reports/license/license-dependency.xml
```diff
@@ -0,0 +1,198 @@
+<licenses>
+  <license name='CC0 1.0 Universal' url='http://creativecommons.org/publicdomain/zero/1.0/'>
+    <dependency>jsr166e-1.0.0.jar</dependency>
+  </license>
+  <license name='No license found'>
+    <dependency>base-0.0.114.jar</dependency>
+    <dependency>application-action-0.0.89.jar</dependency>
+    <dependency>zipkin-cassandra-core-1.23.4-SNAPSHOT.jar</dependency>
```
codefromthecrypt: guess we need to do something to fix this? suppose changing the jar plugin to add a license file?

build/reports/license/license-dependency.xml
```diff
@@ -0,0 +1,198 @@
...
+    <dependency>group-0.0.90.jar</dependency>
+    <dependency>zipkin-query-1.23.4-SNAPSHOT.jar</dependency>
+    <dependency>collections-0.0.109.jar</dependency>
+    <dependency>stats-provider-0.0.92.jar</dependency>
+    <dependency>zipkin-receiver-scribe-1.23.4-SNAPSHOT-idl.jar</dependency>
+    <dependency>asm-3.1.jar</dependency>
+    <dependency>zookeeper-3.4.6.jar</dependency>
```
codefromthecrypt: some jars like this are very surprising...

build/reports/license/license-dependency.xml
```diff
@@ -0,0 +1,198 @@
...
+    <dependency>stat-0.0.72.jar</dependency>
+    <dependency>stat-registry-0.0.57.jar</dependency>
+    <dependency>net-util-0.0.100.jar</dependency>
+    <dependency>zipkin-common-1.23.4-SNAPSHOT.jar</dependency>
+    <dependency>stats-util-0.0.58.jar</dependency>
+    <dependency>zipkin-receiver-scribe-1.23.4-SNAPSHOT.jar</dependency>
+    <dependency>servlet-api-2.5.jar</dependency>
+    <dependency>util-system-mocks-0.0.103.jar</dependency>
+    <dependency>zipkin-receiver-kafka-1.23.4-SNAPSHOT.jar</dependency>
+    <dependency>zookeeper-3.5.0-alpha.jar</dependency>
```
codefromthecrypt: wat?! we have an alpha zk dep?! lovely

build/reports/license/license-dependency.xml
abesto: I think the plugin works from the POM file, I'll verify that soon.

build/reports/license/dependency-license.html
abesto: Hm? This dependency is not gone.

build/reports/license/license-dependency.html
abesto: Same here, we still depend on logback.

abesto:
Whackamole is down, GPL removed from the whitelist :fireworks: The license listing is also updated.

abesto:
Moving a conversation here from a changed / removed line: there's a couple of dependencies with no license specified.
- Some of these are Zipkin projects; for those, this is a good opportunity to add license information to the POM files (https://github.com/hierynomus/license-gradle-plugin/blob/master/src/main/groovy/nl/javadude/gradle/plugins/license/LicenseResolver.groovy#L192 shows that's what the plugin uses, as opposed to LICENSES files or some other heuristic)
- The rest are external dependencies, where we can track down licenses where available and tell the plugin about those explicitly in the build definition of Zipkin, as shown in https://github.com/hierynomus/license-gradle-plugin#license-reporting

I propose we break these out into two separate issues.

codefromthecrypt:
sgtm. Let's merge this when you're ready.

abesto:
New issues created, commits squashed, reports removed. Ready for final review and merge.

PR #854, opened 2015-11-16
Uses shorter lookback and proper endTs to speed up Cassandra duration unit test

yurishkuro:


codefromthecrypt:
Thanks, @yurishkuro!

Issue #857, opened 2015-11-16
Cassandra does not support microsecond precision for timestamp fields


yurishkuro:
I think we need to revert the "setunsafe" code to use native timestamp methods. We should only use millisecond timestamps for the directly exposed timestamp fields, as those should only be used for time window lookups and sorting, where microseconds are not that critical. The timestamps in the thrift blobs will remain at microsecond precision,

codefromthecrypt:
Sounds fine to me, as long as we make it obvious in the code and DDL that these are millisecond granularity as opposed to micros.

codefromthecrypt:
I think we should apply this globally. ex change QueryRequest and the HTTP api to millis, so it is consistent. The logic is sound, and a change like this is better to do before we document the HTTP api.

yurishkuro:
sgtm

michaelsembwever:
for the tracing within cassandra having microsecond accuracy is important.

yurishkuro:
Should we close this as fixed by my PR?

michaelsembwever:
+1

PR #860, opened 2015-11-16
Changes query endTs and lookback from micros to millis

codefromthecrypt:
Time units of query parameters endTs and lookback are now milliseconds
as opposed to microseconds, the grain of Span and Annotation timestamp.
Milliseconds is a more familiar and supported granularity for query,
index and windowing functions.

This change does not change implementation schema. Span stores who use
custom indexes (like cassandra and redis) probably want to update to use
millis now.

See #857

### DIFF
--- a/zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinDependencies.thrift
+++ b/zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinDependencies.thrift
@@ -21,16 +21,16 @@ struct DependencyLink {
   /** child service name (callee) */
   2: string child
   # 3: Moments OBSOLETE_duration_moments
-  /** calls made during the duration (in microseconds) of this link */
+  /** calls made during the duration of this link */
   4: i64 callCount
   # histogram?
 }
 
 /* An aggregate representation of services paired with every service they call. */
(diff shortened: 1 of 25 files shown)


codefromthecrypt:
manual testing of the web app works with queries and dependency graph

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinDependencies.thrift
```diff
@@ -21,16 +21,16 @@ struct DependencyLink {
...
-  /** calls made during the duration (in microseconds) of this link */
+  /** calls made during the duration of this link */
   4: i64 callCount
   # histogram?
 }
 
 /* An aggregate representation of services paired with every service they call. */
 struct Dependencies {
-  /** microseconds from epoch */
+  /** milliseconds from epoch */
```
yurishkuro, line 31: Added https://github.com/openzipkin/zipkin-dependencies-spark/issues/3

codefromthecrypt, line 31: Thx for logging this

yurishkuro:
Fixed #857

PR #861, opened 2015-11-16
Stores timestamps in milliseconds, native format for Cassandra.

yurishkuro:
Uses long column names for duration index table.

@michaelsembwever

### DIFF
--- a/zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
+++ b/zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
@@ -87,14 +87,14 @@ public final class Repository implements AutoCloseable {
 
             @Override
             protected Set<String> initialValue() {
-                return new HashSet<String>();
+                return new HashSet<>();
             }
             @Override
             public Set<String> get() {
                 long newCacheInterval = toCacheInterval(System.currentTimeMillis());
                 if (cacheInterval != newCacheInterval) {
                     cacheInterval = newCacheInterval;
-                    set(new HashSet<String>());
+                    set(new HashSet<>());
                 }
                 return super.get();
             }
@@ -107,7 +107,7 @@ public final class Repository implements AutoCloseable {
      * Note: This constructor performs network I/O to the {@code cluster}.
      */
     public Repository(String keyspace, Cluster cluster, Boolean ensureSchema) {
-        if (ensureSchema.booleanValue()) {
+        if (ensureSchema) {
             Schema.ensureExists(keyspace, cluster);
         }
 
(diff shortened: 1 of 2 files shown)


zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -755,18 +733,22 @@ private String debugInsertTraceIdByAnnotation(ByteBuffer annotationKey, long tim
     private class DurationRow {
         Long trace_id;
         Long duration;
-        Long timestamp;
+        Long timestamp; // inflated back to microseconds
```
codefromthecrypt, line 736: nice comment!

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -848,8 +831,8 @@ private int durationIndexBucket(long ts) {
                 .setString("service_name", serviceName)
                 .setString("span_name", spanName)
                 .setInt("bucket", durationIndexBucket(timestamp))
-                .setLong("ts", timestamp)
-                .setLong("duration", duration)
+                .setBytesUnsafe("ts", serializeTs(timestamp))
+                    .setLong("duration", duration)
```
codefromthecrypt: nit: fuzz

codefromthecrypt:
can you verify the field names are the same as others in the same file? If so, LGTM with only nits

yurishkuro:
field names are consistent with the rest of the tables.

codefromthecrypt:
nice. :shipit:

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -848,7 +831,7 @@ private int durationIndexBucket(long ts) {
                 .setString("service_name", serviceName)
                 .setString("span_name", spanName)
                 .setInt("bucket", durationIndexBucket(timestamp))
-                .setLong("ts", timestamp)
+                .setBytesUnsafe("ts", serializeTs(timestamp))
```
danchia, line 834: do we still need to use this workaround now that we're writing millis – we should be able to just do `.setDate`? Unless we're just trying to avoid allocating the Date object.

codefromthecrypt, line 834: I prefer we leave the serialization hack for now, as there's a comment
where this goes away in datastax 3 (as you can use long directly)

codefromthecrypt, line 834: ps we want this because the current datastax client over-eagerly wants a j.u.Date type. So it complains if you serialize as long until you can control this in v3.

PR #863, opened 2015-11-17
Requires units for min duration param

yurishkuro:


codefromthecrypt:
-1 (first time ever in zipkin)

I don't like this for a number of reasons:
- The syntax is parsed and validated on the server side. This is an unnecessary round trip.
- Web developers now need to learn scala to troubleshoot this parameter.
- Alternative implementations of the query api now need to be compatible with scala duration parsing
- It is a well-understood goal to move zipkin-web towards a pure-js app, and this is the opposite

I'd suggest using moments.js or some other javascript library instead.

PR #865, opened 2015-11-20
Adds basic troubleshooting instructions for SQL

codefromthecrypt:
Found this to be helpful, debugging a problem with @jcarres-mdsol


codefromthecrypt:
cc @spencergibb @dsyer I'll eventually reproduce this doc, in terms of spring boot, but in case this is interesting to you

spencergibb:
:+1:

PR #869, opened 2015-11-27
Sums up parent-child dependency edges from multiple records

### LINKED ISSUE #868: Dependencies view should deduplicate edges
Due to the fact, that the UI does not specify the `startTs` param calling the query-service, duplicate edges might be added to the graph, which results in an exception on the js side:

```
Uncaught Error: Graph already has edge 'serviceA->serviceB'
```

The problem is caused by running the dependencies job on different days.

In terms of the UI, assuming a range of more than one day is displayed, the edge counts could be added and one unique edge with the sum would exist in the graph.
It could also be implemented on the java side, probably in `CassandraDependencyStore` to be analogous to `AnormDependencyStore`, which does the sum operation and provides unique edges.

yurishkuro:
Fixes #868


codefromthecrypt:
Remember to cut a release, push a tag like 1.25.2

codefromthecrypt:
releasing now as 1.25.2 cc @prat0318

Issue #832, opened 2015-11-10
Configurable cassandra driver's read timeout 


drax68:
On large datasets I've noticed frequent com.datastax.driver.core.OperationTimedOutException
Default timeout is 12 seconds, which is quite small.
Please add configuration flag for this parameter (for query service), so it can be overriden.
Or provide a way to override it in config.

drax68:
Anyone can help with this? @adriancole?

codefromthecrypt:
Do you know what command results in a response of >12s? Ex. can you post a
stack trace? That sounds like a long time.. I suspect we could update it or
punch a hole, but it would be nice to know what is taking so long..

drax68:
Here is a stacktrace, it times out strictly after 12 seconds:

`06:53:38.952 [cluster1-nio-worker-0] ERROR c.t.f.h.i.e.FinatraDefaultExceptionMapper - internal server error
com.datastax.driver.core.exceptions.NoHostAvailableException: All host(s) tried for query failed (tried: cassandra.example.com/1.2.3.4:9042 (com.datastax.driver.core.OperationTimedOutException: [cassandra.example.
com/1.2.3.4:9042] Operation timed out))
        at com.datastax.driver.core.exceptions.NoHostAvailableException.copy(NoHostAvailableException.java:84) ~[zipkin-query-service-1.23.2-SNAPSHOT-all.jar:na]
        at com.datastax.driver.core.DriverThrowables.propagateCause(DriverThrowables.java:37) ~[zipkin-query-service-1.23.2-SNAPSHOT-all.jar:na]
        at com.datastax.driver.core.ArrayBackedResultSet$MultiPage.prepareNextRow(ArrayBackedResultSet.java:285) ~[zipkin-query-service-1.23.2-SNAPSHOT-all.jar:na]
        at com.datastax.driver.core.ArrayBackedResultSet$MultiPage.isExhausted(ArrayBackedResultSet.java:245) ~[zipkin-query-service-1.23.2-SNAPSHOT-all.jar:na]
        at com.datastax.driver.core.ArrayBackedResultSet$1.hasNext(ArrayBackedResultSet.java:126) ~[zipkin-query-service-1.23.2-SNAPSHOT-all.jar:na]
        at org.twitter.zipkin.storage.cassandra.Repository$6.apply(Repository.java:439) ~[zipkin-query-service-1.23.2-SNAPSHOT-all.jar:na]
[...]

PR #874, opened 2015-12-08
Limit select statement in getSpanNames(..) to avoid timeouts

michaelsembwever:
Some users can abuse Zipkin's design and insert crazy numbers of span names underneath each service name.

This will initially manifest itself as a problem through timeouts when trying to read all span names for a given service name back out again.

The only purpose to the span_names table is for the UI component displaying all span names.

No one is going to scroll through such a large list. Limiting the select statement, as this patch does, is the smallest change in the code to tackle the immediate problem.

It may be that the frequent writing to the same service name rows creates "hot" partitions and causes problems to the Cassandra cluster. This problem can be tackled when and if it arises.

ref: https://github.com/openzipkin/zipkin/issues/832

### DIFF
--- a/zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
+++ b/zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
@@ -155,7 +155,8 @@ public final class Repository implements AutoCloseable {
                 QueryBuilder.select("span_name")
                     .from("span_names")
                     .where(QueryBuilder.eq("service_name", QueryBuilder.bindMarker("service_name")))
-                    .and(QueryBuilder.eq("bucket", QueryBuilder.bindMarker("bucket"))));
+                    .and(QueryBuilder.eq("bucket", QueryBuilder.bindMarker("bucket")))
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
agreed. zipkin v1 design uses span name similar to a dimensional modeling label. Its only intent is users to scroll through and should not be so big.

Any future design could alter this, particularly if path-based things are implemented, but that's pie-in-the-cloud https://docs.google.com/document/d/1ixxEs9TvhiGjJObGbRSPhSna3zHdadoUTQIZ5JKgLzU/edit

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -463,7 +466,7 @@ private String debugInsertServiceName(String serviceName, int ttl) {
                             for (Row row : input) {
                                 spanNames.add(row.getString("span_name"));
                             }
-                            return spanNames;
+                            return spanNames.size() < 1000 ? spanNames : Collections.singleton("too many span names");
```
codefromthecrypt, line 469: neat trick!

Issue #872, opened 2015-12-04
[1.25.1] too many tombstones in zipkin.span_duration_index


lycplus:
Currently, we encounter this error in our cassandra:

```
ERROR 03:24:53 Scanned over 100000 tombstones in zipkin.span_duration_index; query aborted (see tombstone_failure_threshold)
ERROR 03:24:53 Scanned over 100000 tombstones in zipkin.span_duration_index; query aborted (see tombstone_failure_threshold)
ERROR 03:24:53 Scanned over 100000 tombstones in zipkin.span_duration_index; query aborted (see tombstone_failure_threshold)
WARN  03:24:53 Uncaught exception on thread Thread[SharedPool-Worker-59,5,main]: {}
java.lang.RuntimeException: org.apache.cassandra.db.filter.TombstoneOverwhelmingException
```

Is this because of updating of schema? what usage can cause this error?

ccl0326:
Can anyone have a look it?Any optimize solution?Thx.

codefromthecrypt:
@michaelsembwever @yurishkuro any thoughts?

yurishkuro:
beats me. The primary key has both a timestamp and a trace ID

```
PRIMARY KEY ((service_name, span_name, bucket), duration, ts, trace_id)
```

Since we index each span individually, perhaps we can add a span ID as well?

michaelsembwever:
(still not getting github notifications, quite frustrating)

@yurishkuro (i do believe) you can avoid this problem by `startTs = max(startTs,  TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis()) - TimeUnit.SECONDS.toMicros(ttl));` in Repository.getTraceIdsByDuration(..)

That is you can avoid queries going into those buckets that we already know are only going to be filled with tombstones.

yurishkuro:
@michaelsembwever not sure I follow. The table's TTL should simply expire the records after N days and drop them during compactions. But they shouldn't create tombstones any more than with any other table in zipkin schema. Where I do see tombstones being possible is if the spans are sub-millisecond, i.e. true microseconds timestamps are different, but they are reported from a millisecond-precision language like JS, thus producing non-unique PKs.

Is there a way in C\* to read tombstoned data and compare it with live keys, to understand what exactly is different between the two and what causes the PKs to be overwritten?

michaelsembwever:
[...]

PR #875, opened 2015-12-08
Avoid span_duration_index queries asking for data we already know is expired

michaelsembwever:
Avoid queries going into span_duration_index buckets that we already know are only going to be filled with tombstones,

 by ensuring neither startTs and endTs extend back in time when all data would be TTL's

 ref: https://github.com/openzipkin/zipkin/issues/872

### DIFF
--- a/zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
+++ b/zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
@@ -752,15 +752,25 @@ public final class Repository implements AutoCloseable {
     }
 
     /** Returns a map of trace id to timestamp (in microseconds) */
-    public ListenableFuture<Map<Long, Long>> getTraceIdsByDuration(String serviceName, String spanName,
-                                                                   long minDuration, long maxDuration,
-                                                                   long endTs, long startTs, int limit) {
-        int startBucket = durationIndexBucket(startTs);
-        int endBucket = durationIndexBucket(endTs);
+    public ListenableFuture<Map<Long, Long>> getTraceIdsByDuration(String serviceName,
+                                                                   String spanName,
+                                                                   long minDuration,
+                                                                   long maxDuration,
+                                                                   long endTs,
+                                                                   long startTs,
+                                                                   int limit,
+                                                                   int ttl) {
+
+        long oldestData = TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis() - TimeUnit.SECONDS.toMillis(ttl));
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
+1 looks sensible to me even if the context to know to do this required a bit of experience!

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
```diff
@@ -749,15 +749,25 @@ private static int compareDurationRowTimestamp(DurationRow d1, DurationRow d2) {
...
+                                                                   long minDuration,
+                                                                   long maxDuration,
+                                                                   long endTs,
+                                                                   long startTs,
+                                                                   int limit,
+                                                                   int ttl) {
+
+        long oldestData = TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis() - TimeUnit.SECONDS.toMillis(ttl));
+        long safeStartTs = 0 == startTs ? startTs : Math.max(startTs, oldestData);
+        long safeEndTs = 0 == endTs ? endTs : Math.max(endTs, oldestData);
```
michaelsembwever, line 766: there's something odd there with the tests and how they sometimes invoke with with empty timestamps. i didn't get my head properly around that, rather just accepted zero as a value that won't get adjusted. i think that's correct but am not entirely sure.

codefromthecrypt:
build fixed via https://github.com/openzipkin/zipkin/pull/877

codefromthecrypt:
whoah.. just learned something!
so we have 2 travis checkers on the pull request
the one ending in "pr" uses the travis config from that pull request
the one ending in "push" uses the previous travis config
so .. if you raise a pull request, which affects travis config, you might end up with a christmas status
the "pr" passes", but the "push" fails

Merging because this actually makes sense, even though it is surprising enough to warrant revisiting. cc @abesto

abesto:
Huh. That... is not something that should happen. Opened https://github.com/openzipkin/zipkin/issues/878

zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
codefromthecrypt, line 766: https://github.com/openzipkin/zipkin/pull/879 thanks for reporting!

yurishkuro:
lgtm, let's see if it makes a difference for @liyichao

lycplus:
Have this released to https://repo1.maven.org/maven2/io/zipkin/ ?

codefromthecrypt:
yes 1.25.3

PR #877, opened 2015-12-08
Only run micro-integration tests when cassandra is already listening

codefromthecrypt:
The volume of logs and resources used makes it difficult to debug c*
test failures. This change makes sure tests don't attempt to start
cassandra.


.travis.yml
```diff
@@ -1,14 +1,17 @@
 language: java
-sudo: false
+# Trusty Beta requires sudo per https://docs.travis-ci.com/user/ci-environment/
+sudo: required
 dist: trusty
-addons:
-  apt:
-    packages:
-      - oracle-java8-installer
```
codefromthecrypt, line 7: don't need to manually update the JDK, since the dist is quite new

PR #879, opened 2015-12-08
Adjusts SpanStoreSpec to use realistic timestamps

codefromthecrypt:
SpanStoreSpec was using odd timestamps, which interfere with logic
around lookback and TTL. This uses today floored to midnight, which is
stable and won't go negative, if TTL or lookback is subtracted from it.

See #875


codefromthecrypt:
cc @michaelsembwever

zipkin-common/src/main/scala/com/twitter/zipkin/storage/CollectAnnotationQueries.scala
```diff
@@ -79,7 +79,7 @@ trait CollectAnnotationQueries {
         // TODO: timestamps endTs is the wrong name for all this
         querySlices(sliceQueries, qr.copy(limit = 1)) flatMap { ids =>
           val ts = padTimestamp(ids.flatMap(_.map(_.timestamp)).reduceOption(_ min _).getOrElse(0))
-          querySlices(sliceQueries, qr.copy(endTs = ts)) flatMap { ids =>
+          querySlices(sliceQueries, qr.copy(endTs = ts / 1000)) flatMap { ids =>
```
codefromthecrypt, line 82: this found a bug!

PR #886, opened 2015-12-16
Allow Kafka consumer id to be overwritten

fedj:
We are running the collector-service in Mesos and we faced a problem where a zombie instance of the service was still running leading to a read problem on the index for the freshly created instance.
In order to facilitate debugging, the ability to identify instances would have greatly minimize the debugging cost as we would have directly seen that a zombie instance was still running.

This fix allows to override/specify the consumer id


jamescway:
:+1:

PR #889, opened 2015-12-23
Mounts query endpoints on zipkin-web

codefromthecrypt:
zipkin-web is due to be removed, once a few pages are migrated off
mustache templates in favor of javascript. This begins that process by
making zipkin-web api compatible with its replacement (zipkin-query).


codefromthecrypt:
cc @yurishkuro

yurishkuro:
Thanks!

Issue #760, opened 2015-10-04
Javascript library updates


codefromthecrypt:
It would be great if someone can update our javascript libraries. They are all pretty old.

codefromthecrypt:
I hate to say it, but I think we are nearly a point where recreating a UI would be less work than maintaining it.

Good news: It is possible to update all libraries with the exception of dagre-d3.

Bad news: there's custom trimming involved, and this isn't represented in the bower file. This means I can't update it easily unless we want to check in a lot of junk.

Ugly news:  dagre-d3 is very tangled. If you try to update it, you'll get in a mess between versions of dagre and d3. After a few hours, I was unable to sort problems such as undefined references to functions with requirejs shims. See #772, which says all that work is wasted as we need another means to create sankey or similar diagram.

Next step:

If we don't want to throw away the UI, we need to maintain it far better than we are now. We probably want to stop checking in assets into src/main/resources and using something like [bower-installer](https://github.com/craigburke/bower-installer-gradle) to install them at build time like other deps.

cc @abesto @eirslett

codefromthecrypt:
in case anyone wants to pick this up, here's the diff that works as long as you don't touch anything related to d3.

``` diff
diff --git a/zipkin-web/src/main/resources/bower.json b/zipkin-web/src/main/resources/bower.json
index acdf7cf..9595949 100644
--- a/zipkin-web/src/main/resources/bower.json
+++ b/zipkin-web/src/main/resources/bower.json
@@ -2,14 +2,14 @@
   "name": "zipkin",
   "version": "0.0.0",
   "dependencies": {
-    "flight": "~1.1.0",
-    "requirejs": "~2.1.5",
-    "bootstrap": "~3.1.1",
-    "bootstrap-datepicker": "~1.3.0",
-    "jquery-timeago": "~1.3.1",
-    "chosen": "https://github.com/harvesthq/chosen/releases/download/v1.1.0/chosen_v1.1.0.zip",
-    "jquery-cookie": "~1.4.0",
-    "momentjs": "~2.5.1",
+    "flight": "~1.5.1",
+    "requirejs": "~2.1.20",
+    "bootstrap": "~3.3.5",
+    "bootstrap-datepicker": "~1.4.1",
+    "jquery-timeago": "~1.4.3",
+    "chosen": "~1.4.2",
+    "jquery-cookie": "~1.4.1",
+    "momentjs": "~2.10.6",
     "dagre-d3": "~0.2.9"
   }
 }
```

[...]

PR #890, opened 2015-12-23
Reduces customness of zipkin-web by eliminating filesystem args

codefromthecrypt:
Direct references to the filesystem led to custom code and limited
options for handling javascript dependencies.

By removing the flags `zipkin.web.cacheResources` and
`zipkin.web.resourcesRoot`, we have less complexity and the assumption
that the asset directory is mutable. This means we can use bower to
add javascript depedencies at package time, which reduces manual work
maintaining them.

Users who want to override the static assets packaged in zipkin-web will
need to make a custom jar, or override resources by placing replacements
before them in the classpath.

Removing this flag means that during development, you may need to bounce
zipkin-web when you change assets such as mustache templates. Since the
goal is to move off mustache and onto pure javascript, this is
acceptable.

See #760

### DIFF
--- a/zipkin-web/src/main/scala/com/twitter/zipkin/web/mustache/ZipkinMustache.scala
+++ b/zipkin-web/src/main/scala/com/twitter/zipkin/web/mustache/ZipkinMustache.scala
@@ -1,39 +1,18 @@
 package com.twitter.zipkin.web.mustache
 
-import com.github.mustachejava.DefaultMustacheFactory
 import java.io._
+
+import com.github.mustachejava.DefaultMustacheFactory
 import com.twitter.mustache.ScalaObjectHandler
 
-import collection.JavaConversions.mapAsJavaMap
-
-class ZipkinMustache(templateRoot: String, cache: Boolean) {
-  import java.io.Reader
-  class ZipkinMustacheFactory extends DefaultMustacheFactory() {
-    override def getReader(rn: String): Reader = {
-      // hack to get partials to work properly
-      val name = if (rn.startsWith("public")) rn else "templates/" + rn
-
-      if (cache) super.getReader(name) else {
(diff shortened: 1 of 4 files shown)


zipkin-web/src/main/scala/com/twitter/zipkin/web/mustache/ZipkinMustache.scala
```diff
@@ -1,39 +1,18 @@
...
-import collection.JavaConversions.mapAsJavaMap
-
-class ZipkinMustache(templateRoot: String, cache: Boolean) {
-  import java.io.Reader
-  class ZipkinMustacheFactory extends DefaultMustacheFactory() {
-    override def getReader(rn: String): Reader = {
-      // hack to get partials to work properly
-      val name = if (rn.startsWith("public")) rn else "templates/" + rn
 
-      if (cache) super.getReader(name) else {
```
codefromthecrypt, line 16: classpath isn't mutable, so caching is effectively always on, now, as that's the mustache default.

zipkin-web/src/main/scala/com/twitter/zipkin/web/mustache/ZipkinMustache.scala
```diff
@@ -1,39 +1,18 @@
...
+import com.github.mustachejava.DefaultMustacheFactory
+import com.twitter.mustache.ScalaObjectHandler
 
-    def invalidateCaches() {
-      mustacheCache.clear()
-      templateCache.clear()
-    }
-  }
-  private[this] val mf = new ZipkinMustacheFactory
-  //TODO: why isn't the scala handler coercing maps properly?
```
codefromthecrypt, line 28: it does now :P

PR #891, opened 2015-12-28
Removes httpvar config for collector sample rate

codefromthecrypt:
This removes the POST http config for the adaptive sample rate, after
noticing ancient comments suggest this wasn't used.

This also refactors the AdaptiveSampler code until it has consistent
names and is testable.

See https://github.com/openzipkin/zipkin-java/issues/52#issuecomment-167402624
See https://github.com/openzipkin/zipkin/pull/327/files#r9215131


yurishkuro:
Does this approach only work in deployments with a single collector process, i.e. it's not scalable to multiple collectors?

codefromthecrypt:
> Does this approach only work in deployments with a single collector
> process, i.e. it's not scalable to multiple collectors?

I presume your question is about the adaptive sampler module in general,
since this change doesn't impact the topic.

The adaptive sampler works on multiple collectors as a leader is elected to
manage updates to the sampleRate based on the storeRate of each member. Ex.
There is code to review store rates under a zookeeper path and then
calculate what the sample rate should be.

Next step is to add a test that proves this actually works ;) that said,
this code is used at Twitter with a dozen collector instances (or was half
a year ago), so I suspect it does.

codefromthecrypt:
releasing as 1.28.0

PR #897, opened 2016-01-03
Do not replace empty strings with "unknown" and do not return them to UI

### LINKED ISSUE #896: UI should not prefer empty service name in annotations over non-empty ones
When instrumenting outgoing HTTP calls, the target service name is not always known, so the client can emit ServerAddress annotation with the correct host:port, but empty service name. This results in the spans displayed in the UI with service name of "unknown", even though when the server is instrumented it emits a valid service name in sr/sr annotations.

Proposed fix:
- do not replace empty service names with "unknown" string, as it makes further processing/filtering difficult
- make sure the empty service name does not take precedence if the spans contains annotations with valid service names.

yurishkuro:
Fixes #896

### DIFF
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/common/Endpoint.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/common/Endpoint.scala
@@ -28,6 +28,9 @@ import com.twitter.zipkin.util.Util._
  * @param ipv4 IPv4 host address packed into 4 bytes
  * @param port IPv4 port or 0, if unknown
  * @param serviceName classifier of a source or destination in lowercase, such as "zipkin-web".
+ *                    Can be an empty string "" if the service name is not known, which would
+ *                    make the span not queryable by the service_name unless some other
+ *                    annotation provides a real value.
  */
 case class Endpoint(ipv4: Int, port: Short, serviceName: String) extends Ordered[Endpoint] {
   checkArgument(serviceName.toLowerCase == serviceName,
@@ -63,5 +66,4 @@ case class Endpoint(ipv4: Int, port: Short, serviceName: String) extends Ordered
 
 object Endpoint {
   val Unknown = Endpoint(0, 0, "")
-  val UnknownServiceName = "unknown"
 }
--- a/zipkin-common/src/test/scala/com/twitter/zipkin/common/SpanTest.scala
+++ b/zipkin-common/src/test/scala/com/twitter/zipkin/common/SpanTest.scala
@@ -31,6 +31,17 @@ class SpanTest extends FunSuite {
       BinaryAnnotation("sa", true, Some(Endpoint(1, 2, "sa"))),
       BinaryAnnotation("lc", "foo", Some(Endpoint(1, 2, "lc")))
     ))
+  val spanWithEmptyServiceNames = Span(12345, "methodcall", 666, None, None, None,
+    List(
+      Annotation(1, "cs", Some(Endpoint(1, 2, "cli"))),
+      Annotation(1, "sr", Some(Endpoint(1, 2, "srv"))),
+      Annotation(1, "ss", Some(Endpoint(1, 2, "srv"))),
+      Annotation(1, "cr", Some(Endpoint(1, 2, "cli")))
+    ),
+    List(
+      BinaryAnnotation("ca", true, Some(Endpoint(1, 2, ""))), // empty service name
+      BinaryAnnotation("sa", true, Some(Endpoint(1, 2, ""))) // empty service name
+    ))
 
   /** Representations should lowercase on the way in */
   test("name cannot be lowercase") {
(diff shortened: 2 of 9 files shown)


codefromthecrypt:
Hi, yuri. I'm in favor of this (ex making the de-facto label "unknown" something a user or instrumentation must opt into).

The changes you reverted were added for consistency, particularly to ensure that we don't end up with data that isn't queryable. In your experience, you've found it better to make "" unqueryable, which I can certainly see the logic in (basically this is effectively making this field optional)

Can you add a summary here about the semantics of empty string? I know there are places in docs where we say empty is prohibited, so please do a quick sweep of serviceName accessors to ensure this change doesn't end up confusing ppl

https://github.com/openzipkin/zipkin/blob/master/zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift#L179

Alternatively (and because this is in my timezone, and a branch on the main repo), I can make these changes and fold them into this. I expect my suggestions to be documentation affecting only (maybe another test, but yeah).

wdyt?

yurishkuro:
I replaced the references to "unknown" with "" in `zipkinCore.thrift`.

FWIW, we've been sending empty strings as service name without any ill effects, the only exception being the UI issue #896. And I recall seeing somewhere, maybe in the old Hadoop job, a code that was checking for empty strings.

zipkin-common/src/main/scala/com/twitter/zipkin/common/Endpoint.scala
```diff
@@ -63,5 +63,5 @@ case class Endpoint(ipv4: Int, port: Short, serviceName: String) extends Ordered
 
 object Endpoint {
   val Unknown = Endpoint(0, 0, "")
-  val UnknownServiceName = "unknown"
+  val UnknownServiceName = ""
```
codefromthecrypt: I'd say use this or remove it. This was used to coerce empty string to unknown.

zipkin-common/src/main/scala/com/twitter/zipkin/common/Endpoint.scala
```diff
@@ -63,5 +63,5 @@ case class Endpoint(ipv4: Int, port: Short, serviceName: String) extends Ordered
 
 object Endpoint {
```
codefromthecrypt, line 67: higher in the file, on the serviceName field, I'd add a note that "" implies the span won't be indexed. (or similar succinct thing)

zipkin-common/src/test/scala/com/twitter/zipkin/common/SpanTest.scala
```diff
@@ -31,6 +31,17 @@ class SpanTest extends FunSuite {
...
     ))
+  val spanWithEmptyServiceNames = Span(12345, "methodcall", 666, None, None, None,
+    List(
+      Annotation(1, "cs", Some(Endpoint(1, 2, "cli"))),
+      Annotation(1, "sr", Some(Endpoint(1, 2, "srv"))),
+      Annotation(1, "ss", Some(Endpoint(1, 2, "srv"))),
+      Annotation(1, "cr", Some(Endpoint(1, 2, "cli")))
+    ),
+    List(
+      BinaryAnnotation("ca", true, Some(Endpoint(1, 2, ""))), // empty service name
```
codefromthecrypt, line 42: interesting use case.. basically this is when you know the IP/port, but not its name (pretty common, esp w/ X-Forwarded-For)

zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift
```diff
@@ -168,13 +168,17 @@ struct Endpoint {
...
    * This is the primary parameter for trace lookup, so should be intuitive as
    * possible, for example, matching names in service discovery.
    *
+   * Conventionally, when the service name isn't known, service_name = "unknown".
+   * However, it is also permissible to set service_name = "" (empty string).
+   * The difference in the latter usage is that the span will not be queryable
+   * by service name unless more information is added to the span with non-empty
+   * service name, e.g. an additional annotation from the server.
+   *
    * Particularly clients may not have a reliable service name at ingest. One
```
codefromthecrypt, line 180: good revision.. I agree that when something isn't reliable, empty is better than pasting "unknown". For example, "" means.. wait a bit longer.. this will clear up! vs "unknown" saying "this may be as good as it gets"

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
```diff
@@ -315,7 +315,7 @@ class Handlers(mustacheGenerator: ZipkinMustache, queryExtractor: QueryExtractor
         "parentId" -> span.parentId.filter(spanMap.get(_).isDefined).map(SpanId(_).toString),
         "spanName" -> span.name,
         "serviceNames" -> span.serviceNames.mkString(","),
-        "serviceName" -> span.serviceName,
+        "serviceName" -> span.serviceName.getOrElse(""),
```
codefromthecrypt, line 318: by contract, this shouldn't return null (as we shouldn't return spans w/o a service name) Any reason for this change?

codefromthecrypt, line 318: guess what I'm wondering is ...
- if we known we won't return spans who have no service name..
- and we fix prioritization to deprioritize empty string..

wouldn't it be a bug for this to be null (or empty)?

yurishkuro, line 318: I am not sure we do have a contract to never return spans without service name. We said we can't _find_ such spans by service name (obviously), but if they are retrieved as part of a larger trace, I think we need to return them. They may contain lots of interesting information useful even without a service name.

codefromthecrypt, line 318: excellent point. Trace.serviceName shouldn't ever be empty, but this could.

PR #899, opened 2016-01-04
Fixes IOException in jar -x on shadow jars

codefromthecrypt:
The shadowJar plugin makes a single file out of all the dependencies of
our services. Occasionally, this leads to file conflicts.

Jackson's META-INF/LICENSE conflicts with the directory META-INF/license

While we can run the shadowJars, you cannot extract them with `jar -x`,
as it ends up with an IOException extracting META-INF/license.

This excludes Jackson's META-INF/LICENSE, until we have a better
solution, ideally with one of the proposals in the shadow jar project.

See https://github.com/johnrengelman/shadow/pull/86
See https://github.com/johnrengelman/shadow/pull/102

### DIFF
--- a/zipkin-collector-service/build.gradle
+++ b/zipkin-collector-service/build.gradle
@@ -15,6 +15,10 @@ jar.manifest.attributes 'Main-Class': mainClassName
 tasks.build.dependsOn(shadowJar)
 artifacts.zipkinUpload shadowJar
 
+shadowJar {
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
Thanks @joshlong @dsyer for finding the root problem.

codefromthecrypt:
christmas build is unrelated

PR #900, opened 2016-01-04
Ensures empty string isn't returned by getAllServiceNames

codefromthecrypt:


codefromthecrypt:
cc @yurishkuro this is the minimum test I was looking for

PR #901, opened 2016-01-06
Introduces KAFKA_STREAMS to control count of consumer threads

codefromthecrypt:
@prat0318 noticed that collectors always hit the same partition, which
meant throughput didn't increase by adding instances. This uses the
"High Level Consumer" approach, exposing a consumer count, as an attempt
to remedy this situation. For example setting `KAFKA_STREAMS=4` should
result in 4 consumers reading different partitions.

See https://cwiki.apache.org/confluence/display/KAFKA/Consumer+Group+Example

### DIFF
--- a/zipkin-receiver-kafka/src/main/scala/com/twitter/zipkin/receiver/kafka/KafkaSpanReceiver.scala
+++ b/zipkin-receiver-kafka/src/main/scala/com/twitter/zipkin/receiver/kafka/KafkaSpanReceiver.scala
@@ -17,11 +17,12 @@ object KafkaSpanReceiverFactory {
    *
    * @param zookeeper the zookeeper connect string, ex. 127.0.0.1:2181
    * @param topic  the topic zipkin spans will be consumed from
+   * @param streams the count of consumer threads consuming the topic
    */
-  def factory(zookeeper: String, topic: String) = {
+  def factory(zookeeper: String, topic: String, streams: Int = 1) = {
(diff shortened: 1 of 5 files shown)


zipkin-receiver-kafka/src/main/scala/com/twitter/zipkin/receiver/kafka/KafkaSpanReceiver.scala
```diff
@@ -17,11 +17,12 @@ object KafkaSpanReceiverFactory {
    *
    * @param zookeeper the zookeeper connect string, ex. 127.0.0.1:2181
    * @param topic  the topic zipkin spans will be consumed from
+   * @param streams the count of consumer threads consuming the topic
    */
-  def factory(zookeeper: String, topic: String) = {
+  def factory(zookeeper: String, topic: String, streams: Int = 1) = {
```
prat0318, line 22: just a doubt : so `streams` would be passed in `string`. won't that be an issue in casting. (scala noob)

prat0318:
:+1:

zipkin-receiver-kafka/src/main/scala/com/twitter/zipkin/receiver/kafka/KafkaSpanReceiver.scala
codefromthecrypt, line 22: it is converted before this, so if not an int, would bomb before here.

PR #902, opened 2016-01-06
Pares down kafka config, notably to prevent NODEEXISTS exception

codefromthecrypt:
This change removes settings that impact stream count, and also removes
settings that hard-code default values. The goal is the least possible
configuration needed to get a multiple-stream collector operating.

The kafka configuration set "consumer.id" and "num.consumer.fetchers",
which interfered with customizing the number of streams per topic.

Notably, setting the stream count above one ends up raising an exception
like below:

```
03:25:14.568 [main] INFO  kafka.utils.ZKCheckedEphemeral - Creating /consumers/zipkin/ids/zipkin_zipkin (is it secure? false)
03:25:14.586 [main] INFO  kafka.utils.ZKCheckedEphemeral - Result of znode creation is: NODEEXISTS
org.I0Itec.zkclient.exception.ZkNodeExistsException: org.apache.zookeeper.KeeperException$NodeExistsException: KeeperErrorCode = NodeExists
```

The "num.consumer.fetchers" is a confusing setting, which seems to limit
the effect of stream count (KAFKA-2008).


prat0318:
:+1:

PR #908, opened 2016-01-20
Fixes error where Kafka stream processing chokes on malformed data

### LINKED ISSUE #907: Collector chokes on getting a malformed kafka trace
### span-of-death

Collector just gets stuck on that trace and does not ever move forward (should instead be logging the malformed trace and moving on). Hence, collector is now dead.

codefromthecrypt:
`msg.message()` implicitly decodes thrifts via scrooge. We now guard the
operation as decode failures were hanging things.

Fixes #907

### DIFF
--- a/zipkin-receiver-kafka/build.gradle
+++ b/zipkin-receiver-kafka/build.gradle
@@ -1,3 +1,5 @@
+test.maxParallelForks = 1
+
 dependencies {
     compile project(':zipkin-collector')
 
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
cc @prat0318

prat0318:
:+1: travis is failing though. fix-n-ship

codefromthecrypt:
took some surgery in the tests. basically we need to isolate topics from another

PR #909, opened 2016-01-21
Adds constants for common binary annotation keys to http

codefromthecrypt:
Zipkin systems quite often trace http systems. However, the keys used
vary. This makes it hard to effectively query across tracers written by
different people.

This encourages common keys, like "http.path", for http. This doesn't
preclude other conventions or limit the choices. The naming convention
is path-based, which is most similar to former discussions around a
common data model, and the exact naming convention used in Google Cloud
Trace. The keys chosen actually match fields stored by existing tracers.

Most notably, this replaces the convention "http.uri" with "http.path",
as there was an implicit understanding formerly that "http.uri" strips
query parameters.

The migration path to these keys is a one-time problem, as zipkin has no
OR query. Ex. you cannot simultaneously query for either "http.uri" or
"http.path". Here are some scenarios:

Customize collectors to duplicate "http.uri" as "http.path"; tell users
to query for "http.path". This costs more storage temporarily. For
example, this duplication could apply while tracers are being updated.
Once old data expires out, users can reliably search by "http.path". At
that point, the customization could move to just renaming "http.uri" to
"http.path" until all tracers are updated.

Another option is to simply implement the new keys. Apologize to users
who search by "http.uri", that they have to also search by "http.path"
until tracers are updated. This is impactful for the period of the TTL,
so for example, this could be scheduled over a holiday weekend to have
the least impact to users.

See https://groups.google.com/forum/#!topic/distributed-tracing/dnzo02hO8KY

### DIFF
--- a/zipkin-query/src/main/scala/com/twitter/zipkin/query/FilteredHttpEntrypointTraceInitializer.scala
+++ b/zipkin-query/src/main/scala/com/twitter/zipkin/query/FilteredHttpEntrypointTraceInitializer.scala
@@ -4,6 +4,7 @@ import com.twitter.finagle.http.{Method, Request, Response}
 import com.twitter.finagle.tracing.{DefaultTracer, Trace}
 import com.twitter.finagle.{Filter, ServiceFactory, Stack, param}
 import com.twitter.util.Future
+import com.twitter.zipkin.thriftscala.Constants
 
 /**
  * Hacked variant of the private `com.twitter.finagle.http.codec.HttpServerTraceInitializer`
@@ -35,7 +36,7 @@ object FilteredHttpEntrypointTraceInitializer extends Stack.Module1[param.Tracer
         case -1 => req.uri
         case n => req.uri.substring(0, n)
       }
-      Trace.recordBinary("http.uri", withoutQuery)
+      Trace.recordBinary(Constants.HTTP_PATH, withoutQuery)
       svc(req)
     }
   }
--- a/doc/src/sphinx/Architecture.rst
+++ b/doc/src/sphinx/Architecture.rst
@@ -54,7 +54,7 @@ The line above will add an annotation with the string attached to the point in t
 when it happened. You can also add a key value annotation. It could look like this:
 
 .. parsed-literal::
-    Trace.recordBinary("http.response.code", "500")
+    Trace.recordBinary("http.status_code", "500")
 
 Ruby Thrift
 -----------
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/common/BinaryAnnotation.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/common/BinaryAnnotation.scala
@@ -21,18 +21,18 @@ import com.twitter.io.Charsets.Utf8
 
 /**
  * Binary annotations are tags applied to a Span to give it context. For example, a binary
- * annotation of "http.uri" could the path to a resource in a RPC call.
+ * annotation of "http.path" could the path to a resource in a RPC call.
  *
  * <p/>Binary annotations of type [[AnnotationType.String]] are always queryable, though more a
  * historical implementation detail than a structural concern.
  *
  * <p/>Binary annotations can repeat, and vary on the host. Similar to Annotation, the host
  * indicates who logged the event. This allows you to tell the difference between the client and
- * server side of the same key. For example, the key "http.uri" might be different on the client and
+ * server side of the same key. For example, the key "http.path" might be different on the client and
  * server side due to rewriting, like "/api/v1/myresource" vs "/myresource. Via the host field, you
  * can see the different points of view, which often help in debugging.
  *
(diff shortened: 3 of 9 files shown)


codefromthecrypt:
@kristofa @jcarres-mdsol, this is the promised follow-up to https://github.com/spring-cloud/spring-cloud-sleuth/pull/115 I have an update readying for brave.

zipkin-query/src/main/scala/com/twitter/zipkin/query/FilteredHttpEntrypointTraceInitializer.scala
```diff
@@ -35,7 +35,7 @@ object FilteredHttpEntrypointTraceInitializer extends Stack.Module1[param.Tracer
         case -1 => req.uri
         case n => req.uri.substring(0, n)
       }
-      Trace.recordBinary("http.uri", withoutQuery)
+      Trace.recordBinary("http/path", withoutQuery)
```
jcarres-mdsol: Shouldn't you be using the constants you have added?

codefromthecrypt: hah extremely good point! will change all these call-sites

yurishkuro:
Why use slash as a separator instead of dot?
- for those accustomed to urls and xpaths, having slash in the middle but not at the front looks weird
- in informal written English slash is often used to mean "or", as in "If/when Mary ever shows up"
- slash needs to be encoded when passed in http query string `http%2Fpath`, dot doesn't

codefromthecrypt:
Slashes were used in the common model doc, which a lot of this came out of,
so that is how it started. Slashes fit nicely with URNs, for example, as a
relative path. Ex google cloud trace keys are exactly the same as these,
but have a different implied URN namespace.

I didn't hear any resistance to slash encoding on the distributed-tracing
thread that led to this, nor from Jordi, Kristof or others that saw the
preperatory pull request. Maybe they weren't considering the URL encoding
aspect, or not considering it as a major concern. We should find out it if
is widely shared before merging.

Specifically about URL encoding. You are right that this implies encoding
the slashes. That said, you'll notice though the query format of zipkin v1
already requires this due to " and " syntax. I didn't consider this a
driving factor and was hoping to be more common to be others on these keys.
I didn't enumerate this tradeoff eventhough I thought about it. Thanks for
bringing it up.

Anyone else have an opinion about this?

codefromthecrypt:
Ex in google cloud trace

"http/status_code" is implicitly
 "trace.cloud.google.com/http/status_code" and you can use either in their
system.

codefromthecrypt:
@CodingFabian FYI, I have a WIP for the changes in brave, and in case I don't raise that before I fly out.

prat0318:
i don't have any strong opinions on the name. Though i am curious if these binary annotations (http/path, http/staus_code) will have special behaviour in the zipkin system? in other words, if my tracer still pushes `http.uri`, will it lose something? (other than being different from java/scala tracers).

codefromthecrypt:
There's no impact. Your tracer can store "mary had a little lamb" if it
wanted to :P

These are suggested labels, with a goal that folks want to use them. For
example, if these are logged commonly (eventually), users don't need to
remember different rules per language or framework. In many cases, there's
only one or two frameworks, this is more valuable when there are several
different tracers contributing to the same system.

prat0318:
right. i missed the point about different tracers contributing to the same system.

codefromthecrypt:
I haven't heard anyone voice an opinion in favor of the slash delimited syntax. There have been two questions about the former syntax, and one request to retain it. The URN thing is contrived, as we don't have any requirement to support this. I'll see if I can smoke out any strong opinion for slashes, and if not, switch to dots.

codefromthecrypt:
I think summary in favor of '/' is that it is probably a better delimiter in isolation. However, this isn't truly green field. There's prior art in zipkin using dots. Plus, "http/path=/foo/bar" indeed looks worse than "http.path=/foo/bar", regardless of URL encoding.

Long story short.. will change to dots.

codefromthecrypt:
OK I switched the convention to dots.

I also added "http.response.size" and "http.request.size", convention borrowed from google cloud trace (except dots not slashes).

While not widely supported in existing tracers, http size is widely supported in log-based analysis, including [common logfile format](https://www.w3.org/Daemon/User/Config/Logging.html#common-logfile-format) (which is over 20years old at this point). It would be nice to show folks how to use zipkin instead.

Any opinions? @yurishkuro @kristofa @jcarres-mdsol @prat0318

yurishkuro:
I like it.

prat0318:
++

On Saturday, January 23, 2016, Yuri Shkuro notifications@github.com wrote:

> I like it.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/openzipkin/zipkin/pull/909#issuecomment-174200253.

## 

Prateek Agarwal

PR #911, opened 2016-01-21
Fix: Groovy only interpolates on ", not ', leading to invalid POM

### LINKED ISSUE #910: POMs in Maven Central have unresolvable versions for some dependencies
I assume this is a bug in the gradle configuration. For example from http://repo1.maven.org/maven2/io/zipkin/zipkin-web/1.30.0/zipkin-web-1.30.0.pom:

```
    <dependency>
      <groupId>com.github.spullara.mustache.java</groupId>
      <artifactId>compiler</artifactId>
      <version>${commonVersions.mustache}</version>
      <scope>compile</scope>
    </dependency>
```

abesto:


codefromthecrypt:
Good eye!

Issue #912, opened 2016-01-25
Overly sensitive test in DependencyStoreSpec


yurishkuro:
@adriancole I noticed in zipkin-dependency-spark, the test `getDependenciesAllInstrumented` is disabled because it fails. I think it fails because it's the only one that calls `store.getDependencies` with a specific timestamp, while others call it with `today`. Cassandra job saves the DepLinks record with `Time.now` floored to a day, so while it produces the correct results, they cannot be retrieved by `store.getDependencies`. 

Would it make sense to redefine this test to use timestamps like `(today + 250) * 1000`, and then query at `today + 1000` like all other tests do?

codefromthecrypt:
I didn't look deeply at why the job fails. If you can find a way for the
test to pass meaningfully for the 2 dep stores and spark, please do!

Thanks for having a look.

yurishkuro:
@adriancole I will do a PR.

May also add more tests, per https://github.com/openzipkin/zipkin/issues/914

codefromthecrypt:
thanks

codefromthecrypt:
made a comment in the PR. I think the flooring concern shouldn't be conflated with the "today" one.

cassandra's impl will take an input date and floor it to a day. So, they can be retrieved, subject to the window which is a documented aspect of the dependency api. If the tests don't show that working, we can adjust them.

What I'm concerned about is losing a test that shows we can process data older than a day. For example, if _all_ of our tests look at _today_, we can easily make an impl that cannot process, for example, the data from 2 days ago. After looking at the test in question, I can tell that this aspect is non-obvious.

The test data should be explicitly a relative date such as 2 or 3 days ago (which avoids problems around any lookback defaults), as opposed to constants (which it has now). Also, we can rename the test to something like lookback3days which is doubley calls out the nature of the concern.

yurishkuro:
sorry, replied on PR before seeing this. If the primary concern was "test that shows we can process data older than a day", then the fix is simple - I can probably leave the trace timestamps the same, but only change the query time to "today", because it's how Cassandra job always stores the result.

PR #915, opened 2016-01-25
Changes test getDependenciesAllInstrumented to run the query as of 'today', because Cassandra Spark job only saves computed dependencies as-of 'now' floored to a day.

yurishkuro:
Per #912

### DIFF
--- a/zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
+++ b/zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
@@ -121,10 +121,7 @@ abstract class DependencyStoreSpec extends JUnitSuite with Matchers {
     processDependencies(trace)
 
     val traceDuration = Trace.duration(trace).get
-    result(store.getDependencies(
-      (trace(0).timestamp.get + traceDuration) / 1000,
-      Some(traceDuration / 1000)
-    )).sortBy(_.parent) should be(
+    result(store.getDependencies(today + 1000)).sortBy(_.parent) should be(
       List(
         new DependencyLink("trace-producer-one", "trace-producer-two", 1),
         new DependencyLink("trace-producer-two", "trace-producer-three", 1)
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/common/Dependencies.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/common/Dependencies.scala
@@ -20,7 +20,7 @@ import scala.util.hashing.MurmurHash3
 
 /**
  * A representation of the fact that one service calls another. These should be unique across
- * the set of (parent, child)
+ * the set of (parent, child) for a given time period.
  *
  * @param _parent the calling [[com.twitter.zipkin.common.Endpoint.serviceName]]
  * @param _child the callee's [[com.twitter.zipkin.common.Endpoint.serviceName]]


zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
```diff
@@ -103,28 +103,25 @@ abstract class DependencyStoreSpec extends JUnitSuite with Matchers {
...
+        Annotation((today + 225) * 1000, Constants.ClientRecv, Some(two.copy(port = 3002)))))
+    ))
     processDependencies(trace)
 
     val traceDuration = Trace.duration(trace).get
-    result(store.getDependencies(
-      (trace(0).timestamp.get + traceDuration) / 1000,
-      Some(traceDuration / 1000)
-    )).sortBy(_.parent) should be(
+    result(store.getDependencies(today + 1000)).sortBy(_.parent) should be(
```
codefromthecrypt, line 124: This test is subtly different than before? The former test took the inputs derived from the trace, and now they are constants, and for example, not using the lookback arg anymore. That means the test might be passing for the wrong reasons. (maybe it is fine, but it changed a lot, not only the trace timestamps, but the test inputs too)

Can we use the same semantic inputs, even if they are extracted to locals for readability

zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
```diff
@@ -103,28 +103,25 @@ abstract class DependencyStoreSpec extends JUnitSuite with Matchers {
     val two = Endpoint(127 << 24 | 2, 9410, "trace-producer-two")
     val three = Endpoint(127 << 24 | 3, 9410, "trace-producer-three")
 
-    val trace = List(
-      Span(10L, "get", 10L, None, Some(1445136539256150L), Some(1152579L), List(
```
codefromthecrypt, line 107: overall question. Are we saying that the spark job cannot aggregate data earlier than today? One possibly un-obvious aspect of this test is that it is explicitly older than _today_. Ex. it is easy to accidentally create an impl that cannot re-process historical spans.

If these stamps were redone to be "last week" or "3 days ago" I think it could call out the intent more explicitly.

yurishkuro, line 107: There is one known issue in the spark job that it always aggregates _everything_ in the DB, i.e. it does not filter out any spans by timestamps. I am looking to fix it in https://github.com/openzipkin/zipkin-dependencies-spark/issues/7. But that's not the problem here.

It is that the job takes no inputs, and always saves the computed result as-of _today_. That's why the tests with hardcoded _query time_ are going to fail.

zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
yurishkuro, line 124: What was the objective of the test? I thought it was testing the shape of the resulting graph, sounds like you think it was doing more than that. There are certain caveats with testing the specific timestamps & lookback values I described in the previous comment. E.g. the as-of query, with/out lookback, is unlikely to work without making the Spark job taking the startTs/endTs parameters.

yurishkuro, line 124: if we want to test query param, I would rather do different tests (after passing timestamps to the Spark job), where we have traces A->B, C->D, etc. placed in or out of the time range.  For traces that are on both sides of the time range, the behavior is not even clearly defined today, e.g. Spark job might read the trace partially (after  openzipkin/zipkin-dependencies-spark#7 is done)

yurishkuro:
@adriancole I've restored the odd timestamps, as they don't affect how the test runs in the Spark job. The main fix is to make the query as-of today, else it doesn't find the dependencies data stored in Cassandra.

codefromthecrypt:
Why wouldn't the fix be to have the spark job store the data with
timestamps according to the spans? It seems like you are saying the data is
timestamped for when it was processed, which doesn't make sense at all.

codefromthecrypt:
Ex. here's the parameter.. it is clearly about the timestamps of the
spans, not the time the job happened to run
- @param endTs only return links from spans where
  [[com.twitter.zipkin.common.Span.timestamp]]
-              are at or before this time in epoch milliseconds.

zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
codefromthecrypt, line 107: I'm sure that the spark job can be written in a way that it can retain the
timestamps and that we can also find a way to allow it to run more flexibly.

For example, to test the spark job at all required refactoring. I know
because I did that work. The current form of the spark job is in no way
good enough to dictate changes here.

I'm quite sure you can open issues and find other defences that would
support dumbing down this test to allow the current spark job to pass.

I think it is a better choice to allow the spark job to become better and
pass tests on its own virtue rather than changing the tests so they pass as
is.

For example, this test we are talking about directly supports the current
pull request to support a different endTs.

The back and forth with you is more effort than fixing the code

codefromthecrypt, line 107: I will raise an PR to make the date range concern a separate one and more
explicit. This might result in another test to skip, but at least it will
isolate the topic to a specific test. Even if that test also needs to be
skipped, I hope this is progress.

yurishkuro:
@adriancole ok I am going to take a look at https://github.com/openzipkin/zipkin-dependencies-spark/issues/7, and then perhaps this might just work out of the box.

zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
codefromthecrypt, line 107: Hopefully this addresses the concerns you had. As long as this passes,
or is explicitly skipped, I don't mind if the other test input
parameters are changed.

https://github.com/openzipkin/zipkin/pull/922

codefromthecrypt:
Thanks, yuri

yurishkuro:
Replaced by https://github.com/openzipkin/zipkin-dependencies-spark/pull/16

michaelsembwever:
:+1:

PR #920, opened 2016-01-26
Fixes selected end time no effect

### LINKED ISSUE #786: zipkin-web end time set
image

I find i can not select end time (zipkin version is 1.17.2) 

url send end time is not my select time ，but the OS current time

t-yuki:
End time handling seems to be corrupted and this changes Fixes #786

### DIFF
--- a/zipkin-web/src/main/resources/app/js/component_ui/timeStamp.js
+++ b/zipkin-web/src/main/resources/app/js/component_ui/timeStamp.js
@@ -30,7 +30,7 @@ define(
       this.dateChanged = function (e) {
         var time = moment(e.date);
         time.add(moment.duration(this.$time.val()));
-        this.setTimestamp(time);
+        this.setTimestamp(moment.utc(time));
       }
 
       this.timeChanged = function () {
--- a/zipkin-web/src/main/resources/app/js/component_ui/timeStamp.js
+++ b/zipkin-web/src/main/resources/app/js/component_ui/timeStamp.js
@@ -11,11 +11,11 @@ define(
 
     function timeStamp() {
       this.init = function () {
-        this.$timestamp = this.$node.find("[name=timestamp]");
+        this.$timestamp = this.$node.find(".timestamp-value");
         this.$date = this.$node.find(".date-input");
         this.$time = this.$node.find(".time-input");
         var ts = this.$timestamp.val();
-        this.setDateTime((ts) ? moment(ts / 1000) : moment());
+        this.setDateTime((ts) ? moment(Number(ts)) : moment());
       }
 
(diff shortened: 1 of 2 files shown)


zipkin-web/src/main/resources/templates/v2/index.mustache
```diff
@@ -22,7 +22,7 @@
       <label for="timeStamp">End time</label>
       <input class="form-control input-sm date-input" type="text">
       <input class="form-control input-sm time-input" type="text">
-      <input class="" id="endTs" name="endTs" type="hidden" value="{{endTs}}">
+      <input class="timestamp-value" id="endTs" name="endTs" type="hidden" value="{{endTs}}">
```
yurishkuro, line 25: you could've used the existing `endTs` element ID

t-yuki, line 25: Does it mean to find by element ID in timestamp.js?
I think it is too specific and no other element like start time field can use timestamp class.

yurishkuro, line 25: I see, you want this function to apply to all timestamps, not just this one `endTs`.

zipkin-web/src/main/resources/app/js/component_ui/timeStamp.js
```diff
@@ -28,7 +28,9 @@ define(
       }
 
       this.dateChanged = function (e) {
-        this.setTimestamp(moment(e.date));
+        var time = moment(e.date);
+        time.add(moment.duration(this.$time.val()));
```
yurishkuro, line 32: why do we need to add duration?

t-yuki, line 32: I found that `e.date` contains date only, no time.
So before this change, if you changed date, time will be cleared to 00:00

zipkin-web/src/main/resources/templates/v2/index.mustache
yurishkuro, line 25: how is the `timeStamp` function scoped? When it does `this.$node.find(".timestamp-value");`, is it guaranteed to only search children of some element, e.g. the div? If it does a global search it could find other timestamp fields with the same class.

yurishkuro, line 25: I guess it does the same for date-input and time-input, so nvm.

yurishkuro:
ok, LGTM

codefromthecrypt:
@t-yuki I don't think you realize how much of a hero you are to me for helping with UI fixes.

prat0318:
<3 @t-yuki

Issue #791, opened 2015-10-23
Allow viewing of older dependencies


codefromthecrypt:
While the dependencies api accepts a stop-point, `?endTs=1445470517980000`, the UI doesn't expose this. This means people making demos with canned data will see nada, and have no means to.

In the mean time, folks using spigo need to re-run their simulations before posting to zipkin. If you see nothing, maybe it is just old data. cc @adrianco

eirslett:
I think #921 fixed this.

Issue #727, opened 2015-09-23
Service Dependencies UI enhancements


yurishkuro:
Each item might need to be split into separate issue, but I want to just list them all in one place. Without some of these features the dependency graph is practically unusable for large systems (e.g. 500+ services)
1. There should be a combobox with service names, and selecting a service should act similar to what happens when you mouse-over the service in the graph, i.e. highlight connected services, dim the rest. 
2. There should be a way (button, checkbox?) to trim the graph to only the services connected to a selected service. It can actually be a dropdown with "distance", i.e. trimming the graph to services that are within specified number of hops from the selected service (currently highlighting is only one hop).
3. The edges are currently highlighted on mouse-over, but not clickable - clicking an edge should bring the same second-level popup as when you click on a service first and then on one of its up/downstream friends

codefromthecrypt:
the dependency graph has been rewritten completely in lens so closing out issues in the classic ui

PR #921, opened 2016-01-26
Add dependencies takeback button

t-yuki:
Add experimental takeback button to dependencies view.
It also depends on #920 because timestamp.js was not working.
Refs #791 #727

### DIFF
--- a/zipkin-web/src/main/resources/app/js/component_data/dependency.js
+++ b/zipkin-web/src/main/resources/app/js/component_data/dependency.js
@@ -70,7 +70,13 @@ define(
           }.bind(this));
         });
 
-        this.getDependency(Date.now());
+        var endTs = document.getElementById('endTs').value || moment().valueOf();
+        var startTs = document.getElementById('startTs').value;
+        var lookback;
+        if (startTs && endTs > startTs) {
+          lookback = endTs - startTs;
+        }
+        this.getDependency(endTs, lookback);
       });
 
       this.getServiceData = function (serviceName, callback) {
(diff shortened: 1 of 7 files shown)


t-yuki:
Is adding new interface acceptable?
Preview: 
image

yurishkuro:
@t-yuki enhancing the interface is great! Only it's not clear to me what "take back" button means. Should it be just called "Run" or "Query" or "Execute"?

t-yuki:
I changed the label to "Analyze Dependencies".
Also I added a missing file.
PTAL.

codefromthecrypt:
Can you paste another image?

So the endTs and lookback is the same concept here as in the span query (main screen). Can you verify the fields are named the same and look similar? (for continuity's sake across the screens). Hard to tell, and they may already be.

I'm quite excited as I was interested in being able to see dependencies from a period besides yesterday.

I wouldn't necessarily say "analyze dependencies" as not all stores will do that online, but it is a minor point and I don't know a better word either!

@jcarres-mdsol @eirslett @kristofa any thoughts?

t-yuki:
It queries to /dependency?endTs=1234&startTs but no effect for server side.
Instead, dependency.js converts `startTs` to `lookback` and sends a query with that parameter.
However, error cases like startTs > endTs is not handled well. (it ignores `startTs` simply)

I have no idea for wording of the button. Find traces page has `Find Traces` so I write `Analyze Dependencies` in Dependencies page. Also `Query` is nice.

Screens here:

image
image

codefromthecrypt:
Thanks.

As far as I'm concerned, let's merge.

I did ping others, so will merge tomorrow unless feedback says otherwise.

codefromthecrypt:
1.31.0 is on the way with this

PR #922, opened 2016-01-27
Adds tests to support Dependency Queries not anchored to today

codefromthecrypt:
This adds tests that directly support #921.

This also helps decouple concerns about other tests raised in #912 and #915.


yurishkuro:
lgtm

Issue #929, opened 2016-01-29
Relationship between CLIENT_SEND and SERVER_RECV annotations of spans


t-yuki:
I'm working on AnormDependencyStore to fix some missing dependencies which have intermediate spans between RPC.
For more details, see wip test https://github.com/t-yuki/zipkin/commit/6bd69e365be4a4bf6a260c4935ad1af3d105f5a5#diff-206cff9258cbda55156ecb3e85c2862dR191

Now, I wonder the relationship between CLIENT_SEND and SERVER_RECV annotations of spans.
I surveyed several implementation and test spec but they varies.

DependencyStoreSpec says cross service requests share a span.
~~However, zipkin-tracer (Ruby) seems to be created new span on SERVER_RECV that is a child of CLIENT_SEND.~~
And annotations reported by zipkin-web/query itself are no way.

How should I handle implementation varieties? or Have I misinterpreted?
Or it is not specified so should DependencyStore handle all possible patterns?
### DependencyStoreSpec

```
[service1] Span0: SR
[service1]   Span1: CS-SA[service2]
[service2]   Span1: SR-CA[service1]
```

It  says cross service requests share a span.
### zipkin-web/query

 Finagle?: https://github.com/twitter/finagle/blob/150bbf14ccff580bf156e6ef185f54fd4fc75e76/finagle-core/src/main/scala/com/twitter/finagle/tracing/Trace.scala#L238

```
[service1] Span0: SR-SA[service1]-CA[service1]
[service2]   Span1: CS-SA[service1]-CA[service1]
```

Unless I'm mistaken, it seems to be corrupted.
### zipkin-tracer (Ruby)

 SR: https://github.com/openzipkin/zipkin-tracer/blob/master/lib/zipkin-tracer/rack/zipkin-tracer.rb#L41
 CS: https://github.com/openzipkin/zipkin-tracer/blob/master/lib/zipkin-tracer/faraday/zipkin-tracer.rb#L52

```
[service1] Span0: SR
[service1]   Span1: CS-SA[service2]
[service2]     Span2: SR
```

~~It creates new span on Server Receive.~~

**Actually it has the same output to test spec, OK.**
### Go

 SR: https://github.com/spacemonkeygo/monitor/blob/master/trace/manager.go#L106
 CS: https://github.com/spacemonkeygo/monitor/blob/master/trace/http.go#L37

[service1] Span0: SR
[service1]   Span1: CS
[service2]   Span1: SR

It is the same to TestSpec.

[...]

Issue #917, opened 2016-01-26
Test dependencies_headlessTrace is incorrect


yurishkuro:
This test runs against this trace:

``` scala
  val trace = ApplyTimestampAndDuration(List(
    Span(1L, "get", 2L, Some(1L),
      annotations = List(
        Annotation((today + 50) * 1000, Constants.ClientSend, Some(zipkinWeb)),
        Annotation((today + 100) * 1000, Constants.ServerRecv, Some(zipkinQuery.copy(port = 0))),
        Annotation((today + 250) * 1000, Constants.ServerSend, Some(zipkinQuery.copy(port = 0))),
        Annotation((today + 300) * 1000, Constants.ClientRecv, Some(zipkinWeb))),
      binaryAnnotations = List(
        BinaryAnnotation(Constants.ClientAddr, true, Some(zipkinWeb)),
        BinaryAnnotation(Constants.ServerAddr, true, Some(zipkinQuery)))),
    Span(1L, "query", 3L, Some(2L),
      annotations = List(
        Annotation((today + 150) * 1000, Constants.ClientSend, Some(zipkinQuery)),
        Annotation((today + 200) * 1000, Constants.ClientRecv, Some(zipkinQuery))),
      binaryAnnotations = List(
        BinaryAnnotation(Constants.ClientAddr, true, Some(zipkinQuery)),
        BinaryAnnotation(Constants.ServerAddr, true, Some(zipkinJdbc))))
  ))
```

and it expects the output `List(zipkin-query->zipkin-jdbc(1))`.

I think the expected output is wrong, as from the first span it is clear (to a human) that there should be `zipkinWeb -> zipkinQuery`.

After my change (WIP) to https://github.com/openzipkin/zipkin/issues/914, the actual output from Cassandra store in this test is `List(zipkin-web->zipkin-query(1), zipkin-query->zipkin-jdbc(1))`.

Thoughts?  It's easy to change the test condition, but I am not sure if it's going to be very difficult to support spans like this in other stores.

t-yuki:
I think AnormDB implementation also have same problem and I'm working on to fix it with the other problem.

codefromthecrypt:
Thanks for keeping an eye, out, yuki. Whatever you find in anorm will
likely need to be repeated in zipkin-java (which I can do, if you aren't
into java)

I can't comment too much on this today, but glad the both of you are
looking at it.

yurishkuro:
[...]

Issue #931, opened 2016-01-29
RFC: Introduce backend store Capabilities API


yurishkuro:
It seems #914 is highlighting the need to introduce some sort of Capabilities API for zipkin backend stores so that people who need new features can progress independently without waiting for every single backend store implementation to support a new feature before the feature can be moved into core.

This is especially evident in tests, e.g. #917, where a new feature cannot proceed without upgrading the tests, and the tests cannot be upgraded easily because other backends would not support that.

Proposal: extend SpanStore and DependencyStore API with `getCapabilities()` method returning something like this:

``` scala
trait DependencyStoreCapabilities {
    def canBuildEdgeForSingleSidedSpan: Boolean
    def canInferInstrumentationForEdge: Boolean
}

object DefaultDependencyStoreCapabilities extends DependencyStoreCapabilities {
    val canBuildEdgeForSingleSidedSpan = false
    val canInferInstrumentationForEdge = false
}
```

codefromthecrypt:
is there anyone besides @yurishkuro interested in this?

michaelsembwever:
:-1: 
Until more interest is seen on this I'd rather promote kept-close forks for this type of "additions".

Forking the project isn't a negative thing always. If you've got a company that requires a faster cadence that it can rationalise and get patches accepted then it's perfectly normal that the company keeps a fork so it gets what it wants and can take patience in arguing for and moving ideas back into the main repository. The cadence of open source projects and companies internal development simply don't match, and we can't force that. Maintaining a forked repository so it stays close to the original involves a little effort, but a rather trivial amount if you ask me.

In this manner, a company fork that's kept close to the original, is a Capabilities Interface, without weighing the community with the effort of the API part of it.

It also means that proposals into the main repository have well implemented and tested solutions. The success of implementations eases many of the concerns as to whether the added complexity will pay off.

eirslett:
I'll close this due to lack of activity and disagreement on whether this is the right direction for upstream Zipkin or not.

Issue #930, opened 2016-01-29
RFC: extend DependencyLink data model to include inferred instrumentation


yurishkuro:
## Background

 Adoption is one of the biggest challenges to successful deployment of distributed tracing in an organization. Developers often need to upgrade dependencies of their services and change configuration (in the best case), or even change the application code (in the worst case). Therefore, automatic tracking of adoption stats is very important for distributed tracing teams. The Dependencies graph partially provides these stats, however the edges in the graph sometimes represent a link between two services where only one of the services is fully instrumented for tracing, and the other service name is simply inferred from RPC fields (cf. #914). This is useful information to track in the Dependencies data model, since that information is explicitly available in the traces that are aggregated into dependency graph. 
## Proposal

Add an optional `instrumentation` field to `DependencyLink`.
### Thrift

``` thrift
struct DependencyLink {
  /** parent service name (caller) */
  1: string parent
  /** child service name (callee) */
  2: string child
  #3: Moments OBSOLETE_duration_moments
  /** calls made during the duration of this link */
  4: i64 callCount
  # histogram?
  /** which service(s) actually submitted span data that resulted in this link */
  5: optional Instrumentation instrumentation
}

/** 
 * Describes which service(s) submitted span data that resulted in a DependencyLink. 
 */
enum Instrumentation {
    /** Both parent and child services submitted span data */
    PARENT_AND_CHILD,

    /** Only the parent service submitted span data; the child service name was inferred from it */
    PARENT_ONLY,

    /** Only the child service submitted span data; the parent service name was inferred from it */
    CHILD_ONLY
}
```
### Scala model

In the object model we need a way to "merge" two links with the same `(parent, child)` pair, e.g. when merging Dependency sets from several days. The proposal is to give `ParentAndChild` a priority over the other two values.

``` scala
  sealed trait Instrumentation {
    def +(other: Instrumentation): Instrumentation
  }
  case object ParentAndChild extends Instrumentation {
    override def +(other: Instrumentation): Instrumentation = this
  }
  case object ParentOnly extends Instrumentation {
    override def +(other: Instrumentation): Instrumentation = other match {
      case ParentAndChild => other
      case ChildOnly => ParentAndChild
      case _ => this
    }
  }
  case object ChildOnly extends Instrumentation {
    override def +(other: Instrumentation): Instrumentation = other match {
      case ParentAndChild => other
      case ParentOnly => ParentAndChild
      case _ => this
    }
  }
```

[...]

PR #932, opened 2016-01-29
RFC: Fix anormdb dependencies

t-yuki:
As described in #929 shortly, AnormDependencyStore does not report dependencies that have intermediate spans between RPC like:
- span1: SR SS
  - span2: call
    - span3: CS SR SS CR -> **Dependency 1**

This change also adds dependency using Client Addr and Server Addr like:
- span1: CA SR SS **Dependency 1**
  - span2: call
    - span3: CS CR SA **Dependency 2** #917 case

However, it doesn't pass tests yet by 2 reasons.
One is the same to #917 : `dependencies_headlessTrace`

The other is sorting of results.
It reports:

```
- canSearchForIntervalsBesidesToday *** FAILED ***                                                                                
  org.scalatest.exceptions.TestFailedException:
 List(zipkin-query->zipkin-jdbc(1), zipkin-web->zipkin-query(1)) was not equal to
 List(zipkin-web->zipkin-query(1), zipkin-query->zipkin-jdbc(1)) (DependencyStoreSpec.scala:248)
```

I wonder it must be topologically-sorted so I want to fix tests to ignore order of results, instead of implementing topological sort in the store.
Also this change includes complex-costly SQL, I want comments too.

### DIFF
--- a/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormDependencyStore.scala
+++ b/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormDependencyStore.scala
@@ -34,43 +34,23 @@ case class AnormDependencyStore(val db: DB,
 
     implicit val (conn, borrowTime) = borrowConn()
     try {
-      val parentChild = SQL(
-        """SELECT trace_id, parent_id, id
-          |FROM zipkin_spans
+      SQL(
+        """SELECT parent.endpoint_service_name parent_name, child.endpoint_service_name child_name, COUNT(DISTINCT span.id) count
+          |FROM zipkin_spans       span
+          |JOIN zipkin_annotations parent ON parent.span_id = span.id
+          |JOIN zipkin_annotations child  ON child.span_id  = parent.span_id
           |WHERE start_ts BETWEEN {startTs} AND {endTs}
-          |AND parent_id is not null
+          |AND parent.a_key IN ("cs","ca")
+          |AND child.a_key  IN ("sr","sa")
+          |AND parent.endpoint_service_name IS NOT NULL
+          |AND child.endpoint_service_name IS NOT NULL
+          |GROUP BY parent_name, child_name
         """.stripMargin)
-        .on("startTs" -> startTs * 1000)
-        .on("endTs" -> endTs * 1000)
-        .as((long("trace_id") ~ long("parent_id") ~ long("id") map {
-          case traceId ~ parentId ~ id => (traceId, parentId, id)
-        }) *).groupBy(_._1)
-
-      if (parentChild.isEmpty) {
-        Seq.empty[DependencyLink]
-      } else {
-        val traceSpanServiceName: Map[(Long, Long), String] = SQL(
-          """SELECT DISTINCT trace_id, span_id, endpoint_service_name
-            |FROM zipkin_annotations
-            |WHERE trace_id IN (%s)
-            |AND a_key in ("sr","sa")
-            |AND endpoint_service_name is not null
-            |GROUP BY trace_id, span_id
-          """.stripMargin.format(parentChild.keys.mkString(",")))
-        .as((long("trace_id") ~ long("span_id") ~ str("endpoint_service_name") map {
-          case traceId ~ spanId ~ serviceName => (traceId, spanId, serviceName)
-        }) *).map(r => (r._1, r._2) -> r._3).toMap
-
-        parentChild.values.flatMap(identity).flatMap(r => {
-          // parent can be empty if a root span is missing
-          for (
-            parent <- traceSpanServiceName.get((r._1, r._2));
-            child <- traceSpanServiceName.get((r._1, r._3))
-          ) yield (parent, child)
-        })
-        .groupBy(identity).mapValues(_.size) // sum span count
-        .map{ case ((parent, child), count) => DependencyLink(parent, child, count)}.toSeq
-      }
+      .on("startTs" -> startTs * 1000)
+      .on("endTs" -> endTs * 1000)
+      .as((str("parent_name") ~ str("child_name") ~ long("count") map {
+        case a~b~c => DependencyLink(a, b, c)
+      }) *)
     } finally {
       returnConn(conn, borrowTime, "getDependencies")
     }
--- a/zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
+++ b/zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
@@ -76,12 +76,12 @@ abstract class DependencyStoreSpec extends JUnitSuite with Matchers {
   @Test def getDependencies() = {
     processDependencies(trace)
 
-    result(store.getDependencies(today + 1000)) should be(dep.links)
+    result(store.getDependencies(today + 1000)) should contain theSameElementsAs(dep.links)
   }
 
   /** Edge-case when there are no spans, or instrumentation isn't logging annotations properly. */
   @Test def getDependencies_empty() = {
-    result(store.getDependencies(today + 1000)) should be(Seq.empty)
+    result(store.getDependencies(today + 1000)) should be(empty)
   }
 
   /**
@@ -92,7 +92,7 @@ abstract class DependencyStoreSpec extends JUnitSuite with Matchers {
   @Test def getDependencies_traceIdIsOpaque() = {
     processDependencies(trace.map(_.copy(traceId = Long.MaxValue)))
 
-    result(store.getDependencies(today + 1000)) should be(dep.links)
+    result(store.getDependencies(today + 1000)) should contain theSameElementsAs(dep.links)
   }
 
   /**
@@ -124,7 +124,7 @@ abstract class DependencyStoreSpec extends JUnitSuite with Matchers {
     result(store.getDependencies(
       (trace(0).timestamp.get + traceDuration) / 1000,
       Some(traceDuration / 1000)
-    )).sortBy(_.parent) should be(
+    )) should contain theSameElementsAs(
       List(
         new DependencyLink("trace-producer-one", "trace-producer-two", 1),
         new DependencyLink("trace-producer-two", "trace-producer-three", 1)
@@ -138,7 +138,7 @@ abstract class DependencyStoreSpec extends JUnitSuite with Matchers {
   @Test def getDependenciesMultiLevel() = {
     processDependencies(trace)
 
-    result(store.getDependencies(today + 1000)) should be(dep.links)
+    result(store.getDependencies(today + 1000)) should contain theSameElementsAs(dep.links)
   }
 
   @Test def dependencies_loopback {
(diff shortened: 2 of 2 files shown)


zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
```diff
@@ -188,6 +188,51 @@ abstract class DependencyStoreSpec extends JUnitSuite with Matchers {
...
+        annotations = List(
+          Annotation((today + 150) * 1000, Constants.ClientSend, Some(zipkinQuery)),
+          Annotation((today + 200) * 1000, Constants.ClientRecv, Some(zipkinQuery))),
+        binaryAnnotations = List(
+          BinaryAnnotation(Constants.ClientAddr, true, Some(zipkinQuery)),
+          BinaryAnnotation(Constants.ServerAddr, true, Some(zipkinJdbc))))
+    ))
+
+    processDependencies(trace)
+    val dep = new Dependencies(today, today + 1000, List(
```
yurishkuro, line 270: iirc this is the same as the const at the top of the class

zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
```diff
@@ -188,6 +188,51 @@ abstract class DependencyStoreSpec extends JUnitSuite with Matchers {
...
+    val zipkinJdbc = Endpoint(172 << 24 | 17 << 16 | 2, 0, "zipkin-jdbc")
+
+    val trace = ApplyTimestampAndDuration(List(
+      Span(20L, "get", 20L,
+        annotations = List(
+          Annotation(today * 1000, Constants.ServerRecv, Some(zipkinWeb)),
+          Annotation((today + 350) * 1000, Constants.ServerSend, Some(zipkinWeb)))),
+      Span(20L, "call", 21L, Some(20L),
+        binaryAnnotations = List(
+          BinaryAnnotation(Constants.LocalComponent, "depth2", Some(zipkinWeb)))),
```
yurishkuro, line 207: ironically, Cassandra Spark job might succeed on this, by accident. But in general we need #931 to allow adding per-store features, so that tests can be checking capabilities before setting expectations.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormDependencyStore.scala
```diff
@@ -34,43 +34,23 @@ case class AnormDependencyStore(val db: DB,
...
+          |GROUP BY parent_name, child_name
         """.stripMargin)
-        .on("startTs" -> startTs * 1000)
-        .on("endTs" -> endTs * 1000)
-        .as((long("trace_id") ~ long("parent_id") ~ long("id") map {
-          case traceId ~ parentId ~ id => (traceId, parentId, id)
-        }) *).groupBy(_._1)
-
-      if (parentChild.isEmpty) {
-        Seq.empty[DependencyLink]
```
codefromthecrypt, line 50: I am happy about the red lines!

zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
```diff
@@ -188,6 +188,51 @@ abstract class DependencyStoreSpec extends JUnitSuite with Matchers {
...
+    val today = Time.now.floor(Duration.fromMilliseconds(day)).inMillis
+    val zipkinWeb = Endpoint(172 << 24 | 17 << 16 | 3, 8080, "zipkin-web")
+    val zipkinQuery = Endpoint(172 << 24 | 17 << 16 | 2, 9411, "zipkin-query")
+    val zipkinJdbc = Endpoint(172 << 24 | 17 << 16 | 2, 0, "zipkin-jdbc")
+
+    val trace = ApplyTimestampAndDuration(List(
+      Span(20L, "get", 20L,
+        annotations = List(
+          Annotation(today * 1000, Constants.ServerRecv, Some(zipkinWeb)),
+          Annotation((today + 350) * 1000, Constants.ServerSend, Some(zipkinWeb)))),
```
codefromthecrypt, line 204: Add a comment might explain your intent. Ex. `// Intermediate local spans are not service calls and so shouldn't increase a dependency links call count`

zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
codefromthecrypt, line 207: Which store specific feature does this change imply?

codefromthecrypt:
Definitely agree sort in the test assertion vs the store. In zipkin-java we don't have this issue anyway as assertj has a means to express contents equal without implying order.

I like this change. I think a version of your PR description should be javadoc on the test as it will be understandable later this way.

zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
yurishkuro, line 207: Well, this test is artificially constructed to allow stores to "cheat" (i.e. they pass it by coincidence). So technically it does not introduce a new feature, but if we strictly follow its _intent_ , i.e. building A->B link when there are local spans between A and B, then I can easily construct traces that will fail in most stores. And if we want Anorm to pass after those _harder_ traces, it would be a store-specific feature.

codefromthecrypt, line 207: I think I understand, though maybe not share that capabilities api is
always the answer to this problem. At the moment, this repo is the master,
and implementations follow that or skip test.

Divergence within zipkin has so far been transient, ex people not making
time to adjust leaf implementations.

Capabilities api is a better solution for things that are relatively
permanent, and notable enough to justify the complexity of such an api, and
moreover actionable outside tests (ex users actually invoke the api).

It isn't even needed to share these scala interfaces, it is just a
convenience. Ex java doesn't use these at all except to run tests.
Similarly if you wrote a python dependency store you wouldn't be calling a
capabilities api written in scala.

My 2p is I wouldn't write a capabilities api only to drive tests.. Ignoring
tests that don't pass and keeping issues or README is more appropriate and
better use of the limited time.

IOTW, there's nothing about this change that leads directly to a
capabilities api. We may just feel differently about that.

yurishkuro, line 207: I agree that capabilities API is not needed if you can simply ignore the test. Unfortunately, that's not always the case, e.g. the `headless_trace` test actually needs a different comparison target if the dependencies job can detect uninstrumented clients.

I also think introducing capabilities that require the actual business code doing if-then checks is generally a very bad idea, so I would carefully watch for that. The preferred way is when the store without the capability degrades gracefully into accepted default, e.g. in case of #930, the default null value (or `ParentAndChild` at the domain model) does not break anything and allows the rest of the code work as is.

codefromthecrypt:
@t-yuki the side-discussion is not about your change, and isn't blocking it in any way

to put this back on track, one thing I'd like to see is just a general comment on the test as even-though the pull request explains what this is doing, people looking at the test may not be able to guess that from the method name.

If you don't have time, I can add the comments and fix the order thing on the way in.

ps @jcarres-mdsol if you want to eye-ball the SQL feel free. I know that we've an outstanding concern, which is to have a solution for when the data set size gets bigger.. something that indexes may not be able to solve. We can leave that sort of concern to another pull request, IMHO (unless there's an obvious index you can think of)

t-yuki:
Ok, I add description with  refactoring.
Also I try SQL with our dataset (n GB per day for complete tracing)

zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
t-yuki, line 207: Another demand for capability API is constructing   switches in UI such as counting failed traces, advanced query condition, tenary call counts and etc.
So a HTTP API is potentially required but not yet.

codefromthecrypt, line 207: > Another demand for capability API is constructing switches in UI such as
> counting failed traces, advanced query condition, tenary call counts and
> etc.
> So a HTTP API is potentially required but not yet
> 
> This is the golden point.. The actionable api is the http api, as this is
> the only way trace or dependency elements receive data. Maybe you can
> contribute thoughts here, yuki
> https://docs.google.com/document/d/1ixxEs9TvhiGjJObGbRSPhSna3zHdadoUTQIZ5JKgLzU/edit

codefromthecrypt:
ping when this is squashed and ready. I'll merge and release (also fix zipkin-java to do the same thing)

t-yuki:
Tests are passed.
PTAL with fixing tests including a fix for #917
Should I squash it?
Note that, maybe, default lookback time 7 days is too long for this kind of SQL in usually loaded hosts.

zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
```diff
@@ -76,12 +76,12 @@ abstract class DependencyStoreSpec extends JUnitSuite with Matchers {
   @Test def getDependencies() = {
     processDependencies(trace)
 
-    result(store.getDependencies(today + 1000)) should be(dep.links)
+    result(store.getDependencies(today + 1000)) should contain theSameElementsAs(dep.links)
```
codefromthecrypt, line 79: hah! didn't know that one

codefromthecrypt:
very nice work. squash and merge time!

codefromthecrypt:
@t-yuki about the default lookback.. right now, it is a single parameter that controls both the query and also the dependency tree. I agree that many sites only store 3 days of data. We can re-open the defaults question in another issue, and I don't mind re-summarizing how we got to 7 days

zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
```diff
@@ -160,20 +160,20 @@ abstract class DependencyStoreSpec extends JUnitSuite with Matchers {
   @Test def dependencies_headlessTrace {
     processDependencies(List(trace(1), trace(2)))
 
-    result(store.getDependencies(today + 1000)) should be(Dependencies.toLinks(List(trace(1), trace(2))))
+    result(store.getDependencies(today + 1000)) should contain theSameElementsAs(dep.links)
```
t-yuki, line 163: This change is linked to https://github.com/openzipkin/zipkin-dependencies-spark/pull/18

zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
```diff
@@ -189,6 +189,93 @@ abstract class DependencyStoreSpec extends JUnitSuite with Matchers {
...
+   * SERVER_ADDR and CLIENT_ADDR.
+   * In some cases an RPC call is made where one of the two services is not instrumented.
+   * However, if the other service is able to emit "sa" or "ca" annotation with a service
+   * name, the link can still be constructed.
+   *
+   *   span1: CA SR SS: Dependency 1 by a not-instrumented client
+   *     span2: intermediate call
+   *       span3: CS CR SA: Dependency 2 to a not-instrumented server
+   */
+  @Test def getDependencies_notInstrumentedClientAndServer() = {
```
t-yuki, line 244: This change is linked to https://github.com/openzipkin/zipkin-dependencies-spark/pull/18

t-yuki:
I've squashed and split tests of intermediate spans and not-instrumented clients and servers.

codefromthecrypt:
1.32.0 on the way

PR #934, opened 2016-01-30
Fixes tracegen creates inappropriate CLIENT_SEND and CLIENT_RECV

t-yuki:
CLIENT_SEND and CLIENT_RECV should be annotated with the caller endpoint, not the callee.
Refs #929


codefromthecrypt:
Thank you for this. Tracegen's annotations certainly would have confused someone (and probably confused me when I started). Now, they won't!

PR #936, opened 2016-02-02
Added step to start collector server in README

roshweb:
The steps to start zipkin from source was missing a step to start the collector server.

Added the step to start the collector server before generating dummy traces.


codefromthecrypt:
eek.. you're right, though I think the better fix is to make the tracegen
not use scribe

https://github.com/openzipkin/zipkin/blob/master/zipkin-tracegen/src/main/scala/com/twitter/zipkin/tracegen/Main.scala#L68

PR #938, opened 2016-02-03
Fixes alias column handling on mysql

t-yuki:
Since default behavior of mysql driver can't treat alias column with anorm, DependencyStore on mysql returns weird results.


t-yuki:
Both instrumented spans are good.
zipkin-web and zipkin-query are still bad.

image

codefromthecrypt:
thanks.. incremental progress is good!

PR #942, opened 2016-02-04
Fixes to detect missing dependency by ServerRecv annotations

t-yuki:
#932 allows to detect missing dependency with intermediate spans.

But it introduces another missing dependency by ServerRecv only spans usually generated by finagle including zipkin-web and zipkin-query itself.

This change reduces the impact of this regression by detecting the dependency without intermediate spans.
Walking trace tree within zipkin-query may detect such dependencies. However, it will have serious performance regression.

This change also reduces performance regression by modifying the query to use already existing index `trace_id, span_id` on zipkin_spans table.

Note that it avoids loopback dependency tests by ugly query because finagle generated annotations have no-meaningful CA and SA indicating unintentional loopback.
Please tell me if I misunderstand.

### DIFF
--- a/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormDependencyStore.scala
+++ b/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormDependencyStore.scala
@@ -37,13 +37,16 @@ case class AnormDependencyStore(val db: DB,
       SQL(
         """SELECT parent.endpoint_service_name parent_name, child.endpoint_service_name child_name, COUNT(DISTINCT span.id) count
           |FROM zipkin_spans       span
-          |JOIN zipkin_annotations parent ON parent.span_id = span.id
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
Hmm I wonder if the CA and SA pointing to themselves is a configuration bug?

codefromthecrypt:
Looking at the finagle code, it seems finagle indeed logs ClientAddr as the same service name of the local process.. It records the client-side of the socket.

Yeah seems if we look at "CA" (ex to track uninstrumented clients of the root span), we likely have to deal with this curious "SA" = "CA" cas of finagle..

codefromthecrypt:
I agree that supporting finagle and also supporting "ca" to resolve uninstrumented root spans requires work like this.

I do feel we are digging a hole and more edge cases will make this SQL quite unwieldy. I'm cool merging for now as we can clean that up later. Ex I have alternative SQL (with client side joins) readying for a pull request soon on the java project.

codefromthecrypt:
Thanks for the help, Yuki!

PR #943, opened 2016-02-04
Build frontend assets with Webpack

eirslett:
This lets us use features like ES2015 syntax, commonjs modules,
and makes it easier for developers, with tools like hot-reload.

Bower is replaced by npm, for all libraries except dagre-d3.
We must upgrade to the newest version of dagre-d3 before we can
import it fron npm, but there are breaking changes we need to
account for.

The new tooling requires node/npm on the developer machine, but
this is not a requirement for building zipkin-web. Gradle will
take care of downloading a local copy of node and npm that is
used during the build step.


zipkin-web/src/build.js
```diff
@@ -0,0 +1 @@
+console.log('IN JAVASCRIPT BUILD');
```
eirslett: Oops, I'll remove this, it wasn't intented to be commited

zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
```diff
@@ -52,9 +53,10 @@ trait ZipkinWebFactory { self: App =>
     "js" -> "application/javascript"
   )
 
+  // brewery-zipkin-server.cfapps.io:80
   val webServerPort = flag("zipkin.web.port", new InetSocketAddress(8080), "Listening port for the zipkin web frontend")
   val webRootUrl = flag("zipkin.web.rootUrl", "http://localhost:8080/", "Url where the service is located")
-  val queryDest = flag("zipkin.web.query.dest", "127.0.0.1:9411", "Location of the query server")
+  val queryDest = flag("zipkin.web.query.dest", "localhost:1236", "Location of the query server")
```
eirslett: Oops, this was also not meant to be commited.

zipkin-web/src/main/resources/app/js/component_ui/serviceName.js
```diff
@@ -2,16 +2,17 @@
...
-  function (defineComponent) {
-
-    return defineComponent(serviceName);
+  function (flight, Cookies, chosen) {
+    return flight.component(serviceName);
 
     function serviceName() {
       this.onChange = function() {
-        $.cookie('last-serviceName', this.$node.val());
+        Cookies.set('last-serviceName', this.$node.val());
```
eirslett, line 15: jquery-cookie is replace with js-cookie: https://github.com/carhartl/jquery-cookie

jcarres-mdsol:
This is a great step forward.
I see we will compile the js into one file now. Looking forward the faster UI. The datepicker.css was not needed? The datepicker UI still working?

codefromthecrypt:
So happy to see this! I can test manually in a couple hours.

codefromthecrypt:
Tests run alright, just it hangs on `bin/web`

```
$ bin/web 
:zipkin-common:compileJava UP-TO-DATE
:zipkin-common:compileScala UP-TO-DATE
:zipkin-common:processResources UP-TO-DATE
:zipkin-common:classes UP-TO-DATE
:zipkin-common:jar UP-TO-DATE
:zipkin-web:compileJava UP-TO-DATE
:zipkin-web:compileScala UP-TO-DATE
:zipkin-web:processResources UP-TO-DATE
:zipkin-web:classes UP-TO-DATE
:zipkin-web:nodeSetup
:zipkin-web:npmSetup UP-TO-DATE
> Building 78% > :zipkin-web:npmInstall
```

eirslett:
> The datepicker.css was not needed? The datepicker UI still working?

Oops, I didn't check for that! Thanks for noticing! I guess I was too focused on the JavaScript ;)

> Tests run alright, just it hangs on bin/web

Can you check what's going on inside the `zipkin-web/node_modules` folder? npmInstall is supposed to run `npm install` and fetch a ton of modules.

eirslett:
All the stylesheets are included again now! They too are packaged with Webpack into a nice bundle.

eirslett:
Once this is merged, I'll try to work a bit more on moving server-side logic to the client. (But I didn't want to include everything in the same PR)

codefromthecrypt:
Thanks for the progress, Eirik. I still am having trouble locally, but
maybe I'm the only one. cc @abesto (in case your local setup works)

One thing I want to make sure is that the end distribution actually ends up
working.. I tried this and it seemed to be missing static content.

$ ./gradlew :zipkin-web:clean :zipkin-web:shadowJar

$ java -jar ./zipkin-web/build/libs/zipkin-web-*-all.jar

abesto:
This looks awesome, I'm trying it out now. First note: collector and query fail to build due to `An exception has occurred in the compiler (1.8.0_31)` in `:zipkin-cassandra-core:compileJava`. I suspect it's unrelated, gonna look at just the web without data for now and debug that later.

abesto:
Static assets are served correctly for me locally both with gradle run and from the shadowjar. @adriancole which assets fail to load for you?

eirslett:
The static resources should end up in `zipkin-web/src/main/resources/dist`. With gradle, it puts it in `zipkin-web/build/resources/main/dist` too. And from there it should be packaged into the shadow jar.
Are you sure the node/npm commands were run successfully?
It should do "npm install" and "npm run-script build". (You can find that in build.gradle)

codefromthecrypt:
I can debug what's going on locally, but curiously the build succeeds yet
without the static resources. Seems it should fail (ex we would want it to
fail vs publish bad jar)

In my local env, the only way I can get it to hang is by doing bin/web. I
will look into it as I definitely want this change.

eirslett:
Is it possible to add a task in Gradle that verifies that "main.min.js" and "main.min.css" exist? That could be a fail-safe option.

eirslett:
I think I could reproduce it locally now, I deliberately added a syntax error in the javascript code to make the frontend build fail - but it fails silently, and gradle reports it as a build success - although without the compiled assets.

eirslett:
I think this fix should help!

codefromthecrypt:
Progress, as bin/web now errors out :)

We should add `zipkin-web/node_modules` to the cache section in `.travis.yml`, right? cc @abesto

codefromthecrypt:
after upgrading my local node to 5.5 and nuking `zipkin-web/node_modules`.. `bin/web` works (albeit slower to start than before due to node processing things)!

codefromthecrypt:
ps another thought.. since node was crapping out due to version-things.. might want to check that travis' node doesn't crap out. right now, running tests doesn't actually invoke node, so a failure would wait until post merge. maybe we can do something to make node do things eagerly on pull request?

abesto:
Adding `node_modules` to the cache section sounds like a good idea. Slightly related: @eirslett how do you feel about `npm shrinkwrap`?

Adding `npm test` to the `zipkin-web:check` also sounds like a great idea to me; maybe it'll encourage front-end tests in the future. Edit: wait, `build` depends on `webpack`, so the build should error out if `webpack` fails; I expect the same to happen on Travis.

abesto:
On caching: https://docs.npmjs.com/cli/cache says `npm` has a download cache, `~/.npm` by default. Adding that to the cache section sounds like an even better idea.

eirslett:
> after upgrading my local node to 5.5 and nuking zipkin-web/node_modules.. bin/web works (albeit slower to start than before due to node processing things)!

Actually, you shouldn't need node at all in order to build the project. Try renaming your "node" binary" to "nodedfgdsfg", and see if the gradle still builds?

> We should add zipkin-web/node_modules to the cache section in .travis.yml, right?

Sounds like a good idea!

> On caching: https://docs.npmjs.com/cli/cache says npm has a download cache, ~/.npm by default. Adding that to the cache section sounds like an even better idea.

That is possible (if Travis doesn't cache it by default?), but then we will have an increase in build time, because npm modules will have to be copied into the zipkin-web/node_modules folder. But it might be safer - if we remove a module from package.json, it won't be removed from node_modules automatically, so I guess caching ~/.npm might be the best idea in that regard.

>  @eirslett how do you feel about npm shrinkwrap?

It's a natural next step to think of. Build stability is key!
But I can make another pull request with shrinkwrap later on, so this one doesn't get too big?

> Adding npm test to the zipkin-web:check also sounds like a great idea to me; maybe it'll encourage front-end tests in the future. 

Yes! We should have browser tests. I think using karma/phantomjs is a good fit, since it has good webpack support. Again, I'd want to put that in a separate PR.

eirslett:
> right now, running tests doesn't actually invoke node, so a failure would wait until post merge. maybe we can do something to make node do things eagerly on pull request?

Maybe Travis should run the whole build process instead of just the tests? Might take a little longer, though. But in the long run, I imagine zipkin-web being in a separate git repository, since it slows down the build quite a bit.

codefromthecrypt:
adding travis caching and will watch the build and merge on green (as looks like the other comments are different PRs). Great job @eirslett and thanks for the advice, @abesto!

.travis.yml
```diff
@@ -20,6 +20,9 @@ matrix:
 # These directories are cached to S3 at the end of the build
 cache:
   directories:
+    # The following two directories are for zipkin-web, which gets dependencies via NPM
+    - $HOME/.npm
+    - zipkin-web/node_modules
```
eirslett, line 25: Can we remove this line without a too big impact on build time? It's a relatively rare case that we want to remove node modules we've introduced, but still, we might want to have them removed from node_modules anyways. Some node.js tools do sanity checks on the whole node_modules folder for whatever reasons, even if some modules aren't being used.

abesto, line 25: Since `~/.npm` is cached, I don't think it'll have a big impact; I expect the majority of the build time to come from downloading the packages from NPM. The only thing I can think of that can slow us down is if any compilation takes place, say if there are native extensions to build.

codefromthecrypt, line 25: removed via https://github.com/openzipkin/zipkin/commit/08bd0ee6d95ced8c9f6f11c24269b64e887bc62a

thx for thinking this through!

Issue #945, opened 2016-02-05
CorrectForClockSkew drops local span


yurishkuro:
The attached trace has a local span called `get_root`. This span gets dropped from the trace by `CorrectForClockSkew`.  

[trace.json.txt](https://github.com/openzipkin/zipkin/files/118505/trace.json.txt)

The following diff seems to fix it, but I would like a second opinion about what's the best approach:

```
around @ -249,11 +262,25


+      // Local spans may have no annotations, so above calculation may produce None.
+      // But if the timestamp is already defined, we just need to adjust it by the skew.
+      val lcTimestamp: Option[Long] = span.timestamp.flatMap {
+        ts =>
+          span.binaryAnnotations.find(_.key == Constants.LocalComponent).map {
+            anno => anno.host match {
+              case Some(host) if isHost(host, anno.key) => ts - clockSkew.skew
+              case _ => ts
+            }
+          }
+      }
+
       new SpanTreeEntry(span.copy(
-        timestamp = firstOption,
-        duration = duration,
+        timestamp = lcTimestamp orElse firstOption,
+        duration = span.duration orElse duration,
         annotations = annotations.sorted), spanTree.children)
```

yurishkuro:
Incidentally, there may be another issue (maybe with our instrumentation), because _all_ spans in this trace are actually from the same host, so the local span should have `skew=0`, which would've caused the adjustment logic to be skipped, yet that didn't happen. I haven't investigated yet why it finds a non-zero skew (most likely because of the parent spans).

codefromthecrypt:
logic sounds sensible. ack that it should have skipped (vs dropping the local spans) in the first place

yurishkuro:
@adriancole currently the change is limited to LC spans. It breaks one test in the span store spec that expects span-reported duration to be overwritten by the adjuster.

Should this actually apply to ALL spans? I.e. I can't think of a good reason for the adjuster to _ever_ override the span's timestamp / duration based on the annotations, it should only be allowed to shift the timestamp by the skew.

codefromthecrypt:
> @adriancole currently the change is limited to LC spans. It breaks one
> test in the span store spec that expects span-reported duration to be
> overwritten by the adjuster.
> At first thought, I don't think duration should change, either... Either
> there's a non-intuitive reason why duration is changing, or there's a bug
> in the adjuster.
> 
> Should this actually apply to ALL spans? I.e. I can't think of a good
[...]

PR #947, opened 2016-02-07
Adjusts test expectations to be more correct

codefromthecrypt:
This corrects the clock skew test so that it doesn't require
implementations to adjust a predefined span duration.

This also fixes a dependency store spec which didn't align ids properly.

See #945


codefromthecrypt:
cc @yurishkuro

yurishkuro:
lgtm

PR #948, opened 2016-02-07
Fixes test bug in getDependencies_intermediateSpans

codefromthecrypt:
getDependencies_intermediateSpans declared local spans without a
timestamp. This makes them unqueryable without relying on implementation
specific behavior. For example, the SQL span store assigns a fallback of
current time, but not every implementation does.

### DIFF
--- a/zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
+++ b/zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala
@@ -1,5 +1,7 @@
 package com.twitter.zipkin.storage
 
+import java.util.concurrent.TimeUnit._
+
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
cc @t-yuki @yurishkuro this was hard to notice!

the local spans in this test were unqueryable in implementations that don't set a fallback span.timestamp

yurishkuro:
@adriancole do you mean they were unqueryable because the storage did not index them by the timestamp since it was missing?

codefromthecrypt:
> @adriancole do you mean they were unqueryable because the storage did not
> index them by the timestamp since it was missing?
> Yep.

t-yuki:
Oops, nice catch. `get` spans have Annotation with timestamp but `call` spans are not.

PR #949, opened 2016-02-10
Converges duplicate before_install in travis

codefromthecrypt:
Duplicate before_install stanzas ended up ignoring the first, which
started Cassandra.


codefromthecrypt:
awesome.. this breaks the build :P Thanks for noticing, @yurishkuro 

cc @abesto looks like the main thing here is we should be careful to not duplicate yaml keys as travis silently chooses the last

abesto:
Huh. Nice find.

PR #950, opened 2016-02-10
Uses Anorm in CassandraDependencyStoreSpec

codefromthecrypt:
This removes `Dependencies.toLinks` in favor of re-using Anorm for
Cassandra dependency link tests.

Previously, we had a partially implemented aggregator:
`Dependencies.toLinks`. This was added only to make it possible to test
`CassandraDependencyStore`, which doesn't include an aggregator (ex.
zipkin-dependencies-spark is not in this repo).

`Dependencies.toLinks` hasn't been updated for the various changes in
tests and tracking them here is not valuable maintenance. Since the only
purpose was to feed `CassandraDependencyStore`, we can instead use anorm
which is becoming better maintained.


codefromthecrypt:
cc @t-yuki @yurishkuro I think this will fix the build

codefromthecrypt:
I'm going to merge this as the build is broken otherwise. Happy to take feedback after commit.

yurishkuro:
ingenious solution

lgtm

PR #954, opened 2016-02-10
Fixes CorrectForClockSkew which dropped local spans

codefromthecrypt:
This uses code from @yurishkuro to fix dropped local spans. Note that
this doesn't generally fix adjustment of explicitly set Span.duration,
as that's a [different issue](https://github.com/openzipkin/zipkin/issues/955).

See #945


zipkin-common/src/main/scala/com/twitter/zipkin/adjuster/CorrectForClockSkew.scala
```diff
@@ -251,9 +251,22 @@ object CorrectForClockSkew extends ((List[Span]) => List[Span]) {
...
+              case _ => ts
+            }
+          }
+      }
+
       new SpanTreeEntry(span.copy(
-        timestamp = firstOption,
-        duration = duration,
+        timestamp = lcTimestamp orElse firstOption,
+        duration = duration orElse span.duration, // fallback to original if no annotations
```
codefromthecrypt, line 269: this looks backwards, but it is actually correct as the derived duration isn't present when there are no annotations

PR #957, opened 2016-02-10
Fix #953: Run webpack before preparing resources for fat jar

### LINKED ISSUE #953: zipkin-web-all.jar isn't getting javascript assets
This might be a travis or release related config. The zipkin-web all jar is missing javascript assets. We can't release another version until this is fixed.

ex. download a recent jar here, and java -jar it (or jar -tf it) http://oss.jfrog.org/artifactory/oss-snapshot-local/io/zipkin/zipkin-web/1.32.4-SNAPSHOT/

abesto:
Initially I used `shadowJar.dependsOn`, but it turns out that's too late in the build process.

When testing, keep in mind that currently `zipkin-web/src/main/resources/dist` isn't deleted by `zipkin-web:clean`.

### DIFF
--- a/zipkin-web/build.gradle
+++ b/zipkin-web/build.gradle
@@ -59,6 +59,7 @@ task webpack(type: NpmTask) {
 }
 webpack.dependsOn(npmInstall)
 build.dependsOn(webpack)
+processResources.dependsOn(webpack)
 run.dependsOn(webpack)
 
 apply from: "${rootDir}/gradle/verify-licenses.gradle"


eirslett:
Thanks a lot for finding that bug!
Like you say, we might want to wipe dist as well. That could be done with https://github.com/johnagan/clean-webpack-plugin

codefromthecrypt:
Thanks for the fix! I'm curious why files are going into zipkin-web/src/main/resources/ vs target/something.. anyone know why?

eirslett:
It's just the way I configured webpack. We could change it to target, if that's better.
(You could try changing this line: https://github.com/openzipkin/zipkin/blob/master/zipkin-web/webpack.config.js#L30)

codefromthecrypt:
I think the main idea would be that it could keep "clean" working
conventionally in gradle world, where transient stuff is under build/ OTOH,
maybe the current location is more idiomatic outside gradle?

eirslett:
Moving the output to target/ is more gradle-idiomatic, while keeping it in src/main/resources/dist is more node.js-idiomatic, if you think of that part of the project as a "separate" build that only interacts with the parent project via the gradle-node-plugin.
I don't have a strong opinion about it. If somebody wants to move it to target/ and their build still works, I'm fine with changing it.

codefromthecrypt:
NP for me. I commented here in case someone else
cares 6e2e84521304f57b6f476b42cb2d2161b44c1101

PR #958, opened 2016-02-11
Adds CASSANDRA_MAX_CONNECTIONS to control pooled connections per host

### LINKED ISSUE #956: Cassandra driver's connections per host option
Please add configurable option to manage maxConnectionPerHost cassandra driver parameter.
Defaults is quite low ([2 for HostDistance.REMOTE](https://docs.datastax.com/en/developer/java-driver/2.0/common/drivers/reference/connectionsOptions_c.html?scroll=reference_ds_z1m_2k1_zj__description_unique_5)) and causes multiple errors in collector logs.

codefromthecrypt:
This setting should be used with CASSANDRA_LOCAL_DC, as that indicates
the local datacenter, which CASSANDRA_MAX_CONNECTIONS applies to.

Fixes #956

### DIFF
--- a/zipkin-cassandra/README.md
+++ b/zipkin-cassandra/README.md
@@ -12,6 +12,7 @@ Here are the Cassandra-specific environment variables:
    * `CASSANDRA_USERNAME` and `CASSANDRA_PASSWORD`: Cassandra authentication. Will throw an exception on startup if authentication fails. No default
    * `CASSANDRA_CONTACT_POINTS`: Comma separated list of hosts / ip addresses part of Cassandra cluster. Defaults to localhost
    * `CASSANDRA_LOCAL_DC`: Name of the datacenter that will be considered "local" for latency load balancing. When unset, load-balancing is round-robin.
+   * `CASSANDRA_MAX_CONNECTIONS`: Max pooled connections per datacenter-local host. Defaults to 8
 
 Example usage:
 
--- a/zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
+++ b/zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
@@ -17,7 +17,7 @@ package com.twitter.zipkin.cassandra
 
 import java.net.InetSocketAddress
 
-import com.datastax.driver.core.Cluster
+import com.datastax.driver.core.{HostDistance, PoolingOptions, Cluster}
 import com.datastax.driver.core.policies.{RoundRobinPolicy, DCAwareRoundRobinPolicy, LatencyAwarePolicy, TokenAwarePolicy}
 import com.google.common.net.HostAndPort
 import com.twitter.app.App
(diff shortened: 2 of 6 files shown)


zipkin-cassandra/README.md
```diff
@@ -6,21 +6,12 @@
...
 * [zipkin-query-service](https://github.com/openzipkin/zipkin/blob/master/zipkin-query-service/README.md)
 
-These services apply configuration through environment variables:
+Here are the Cassandra-specific environment variables:
 
-   * `CASSANDRA_ENSURE_SCHEMA`: Ensuring that schema exists, if enabled tries to execute script /zipkin-cassandra-core/resources/cassandra-schema-cql3.txt 
-   * `CASSANDRA_USERNAME` and `CASSANDRA_PASSWORD`: Cassandra authentication. Will throw an exception on startup if authentication fails
-   * `CASSANDRA_CONTACT_POINTS`: Comma separated list of hosts / ip addresses part of Cassandra cluster
-   * `COLLECTOR_SAMPLE_RATE`: Sample rate. Double value between 0.0 (nothing ends up in back-end store) and 1.0 (everything ends up in back-end store)
-   * `COLLECTOR_QUEUE_NUM_WORKERS`: Number of worker threads that pick spans from internal bounded queue and write to back-end store
```
codefromthecrypt: removed the copy/pasting here as it is subject to drift (and unlike other READMEs)

zipkin-collector-service/config/collector-cassandra.scala
```diff
@@ -40,6 +40,8 @@ if (username.isDefined && password.isDefined) {
   Factory.cassandraPassword.parse(password.get)
 }
 
+sys.env.get("CASSANDRA_MAX_CONNECTIONS").foreach(Factory.cassandraMaxConnections.parse(_))
```
codefromthecrypt, line 44: we don't have a great way of testing env variables at the moment, so I tested this via a debugger

zipkin-cassandra/README.md
```diff
@@ -6,21 +6,12 @@
...
-   * `COLLECTOR_ADMIN_PORT`: Collector admin port. Port for admin http service. Admin service provides operational metrics for zipkin-collector-service.
-   * `COLLECTOR_LOG_LEVEL`: Collector log level. Valid values: OFF, FATAL, CRITICAL, ERROR, WARNING, INFO, DEBUG, TRACE, ALL
-
-For default values see:
-* [zipkin-collector-service/config/collector-cassandra.scala](https://github.com/openzipkin/zipkin/blob/master/zipkin-collector-service/config/collector-cassandra.scala).
-* [zipkin-query-service/config/query-cassandra.scala](https://github.com/openzipkin/zipkin/blob/master/zipkin-query-service/config/query-cassandra.scala).
+   * `CASSANDRA_ENSURE_SCHEMA`: Ensuring that schema exists, if enabled tries to execute script /zipkin-cassandra-core/resources/cassandra-schema-cql3.txt. Defaults to true
+   * `CASSANDRA_USERNAME` and `CASSANDRA_PASSWORD`: Cassandra authentication. Will throw an exception on startup if authentication fails. No default
+   * `CASSANDRA_CONTACT_POINTS`: Comma separated list of hosts / ip addresses part of Cassandra cluster. Defaults to localhost
+   * `CASSANDRA_MAX_CONNECTIONS`: Max pooled connections per host. Defaults to 2
```
danchia: I think the default is actually 8 in the Datastax driver.

zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
```diff
@@ -69,6 +70,7 @@ trait CassandraSpanStoreFactory {self: App =>
       builder.withCredentials(cassandraUsername(), cassandraPassword())
     builder.withRetryPolicy(ZipkinRetryPolicy.INSTANCE)
     builder.withLoadBalancingPolicy(new TokenAwarePolicy(new LatencyAwarePolicy.Builder(new RoundRobinPolicy()).build()))
+    builder.withPoolingOptions(new PoolingOptions().setMaxConnectionsPerHost(HostDistance.REMOTE, cassandraMaxConnections()))
```
danchia: I think we should be setting `HostDistance.LOCAL`. REMOTE is only for nodes not in the same DC (as determined by the first node you contact), and the driver should prefer LOCAL nodes.

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
```diff
@@ -35,6 +35,7 @@ object CassandraSpanStoreDefaults {
   val SpanTtl = 7.days
   val IndexTtl = 3.days
   val MaxTraceCols = 100000
+  val MaxConnections = 2
```
danchia: If we change to LOCAL, then the default is 8.

codefromthecrypt: Hmm the original complaint was about the remote pool. Do we want two
variables, then?

danchia: Other C\* people can chime in too, but I'd like to understand better why the OP ended up using what the driver thought were REMOTE nodes. In C\* these typically are the nodes that are in a whole different datacenter (potentially across the US, for example).

codefromthecrypt: great question @drax68 care to chime in?

zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
yurishkuro: the default load balancing is not ideal, we ran into issue with it and changed to 

```
    val lbPolicy = new TokenAwarePolicy(new LatencyAwarePolicy.Builder(new DCAwareRoundRobinPolicy(dc)).build())
```

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
drax68: I guess driver threats cassandra installation as non-local, due to non-default dc name we use. Hope this will be changed after #967

zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
codefromthecrypt: done https://github.com/openzipkin/zipkin/pull/980

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
codefromthecrypt: will change this to affect local, not remote.

codefromthecrypt:
Updated to control local host connections per discussions

PR #959, opened 2016-02-11
Prefers client when deriving duration

### LINKED ISSUE #955: CorrectForClockSkew shouldn't adjust explicitly set Span.duration
`CorrectForClockSkew` currently changes span duration even if it was set by instrumentation. This is a problem.

In order to not meddle with explicitly set duration, we may have to change behavior a bit. Most collectors invoke `ApplyTimestampAndDuration` _before_ storing the spans. `CorrectForClockSkew` is applied at query time, and so can't tell if Span.duration was set by instrumentation, or by `ApplyTimestampAndDuration`! The thing is.. we need duration to be set in order for duration query to work!

One way to unwind this is to not assign missing duration at ingest. Rather, only do this at query time. Note: this would limit duration query support to recently maintained tracers, as old ones don't set Span.duration.

codefromthecrypt:
Previously, the clock skew corrector incorrectly derived duration. This
was inaccurate as the distance between annotations on the span's host is
not subject to skew.

When zipkin spans cross multiple hosts, it is usually due to a client+
server span. By preferring the client's duration, we avoid having to
consider skew when deriving duration from annotations. Regardless, we
only derive duration when it isn't explicitly set.

Fixes #955

### DIFF
--- a/zipkin-common/src/main/scala/com/twitter/zipkin/adjuster/ApplyTimestampAndDuration.scala
+++ b/zipkin-common/src/main/scala/com/twitter/zipkin/adjuster/ApplyTimestampAndDuration.scala
@@ -1,5 +1,6 @@
 package com.twitter.zipkin.adjuster
 
+import com.twitter.zipkin.Constants
 import com.twitter.zipkin.common._
 
 /**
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
cc @yurishkuro

yurishkuro:
seems reasonable

PR #964, opened 2016-02-11
bugfix: make sure that the span is wrapped in jquery

eirslett:
I'm not sure if it's affecting the end user experience in any way, but we get a JavaScript exception that blocks execution when you click the "Expand All" button to show all spans in a trace. (You must open the browser debugger to see it)

Apparently, JavaScript 's weak type system has been ~~abused~~ leveraged. `self.spansByService` is a map from service name to list of span ids (strings), but that list of ids will eventually be converted/lazy-initialised into a list of jquery objects (that wrap the spans' DOM elements). The lazy-initialisation mostly happens inside `getSpansByService`. I think we should refactor the code so that it creates jquery objects up-front. For now, this patch should fix the exception.

### DIFF
--- a/zipkin-web/src/main/resources/app/js/component_ui/trace.js
+++ b/zipkin-web/src/main/resources/app/js/component_ui/trace.js
@@ -287,7 +287,9 @@ define(
             $.each(self.parents[id], function(i, pId) { self.spans[pId].openChildren += 1; });
           });
           $.each(self.spansByService, function(svc, spans) {
-            $.each(spans, function(i, $span) { $span.inFilters += 1; });
+            $.each(self.getSpansByService(svc), function(i, $span) {
+              $span.inFilters += 1;
+            });
           });
           self.triggerForAllServices('uiAddServiceNameFilter');
         });


eirslett:
``` javascript
      this.getSpansByService = function(svc) {
        var spans = this.spansByService[svc];
        if (spans === undefined)
          this.spansByService[svc] = spans = $();
        else if (spans.jquery === undefined)
          this.spansByService[svc] = spans = $('#' + spans.join(',#'));
        return spans;
      };
```

https://github.com/openzipkin/zipkin/blob/3523407844e9d2e09ed305426145996911e27539/zipkin-web/src/main/resources/app/js/component_ui/trace.js#L63
This is the craziest getter function I've ever seen in my life...

zipkin-web/src/main/resources/app/js/component_ui/trace.js
```diff
@@ -287,7 +287,9 @@ define(
             $.each(self.parents[id], function(i, pId) { self.spans[pId].openChildren += 1; });
           });
           $.each(self.spansByService, function(svc, spans) {
-            $.each(spans, function(i, $span) { $span.inFilters += 1; });
+            $.each(self.getSpansByService(svc), function(i, $span) {
```
yurishkuro, line 290: from L289 it looks like svc->spans has already been resolved, but 290 is doing it again via `getSpansByService(svc)`?

eirslett, line 290: Yes, `getSpansByService(svc)` basically returns `spans` (the argument from L289), _except_ when `spans` is a list of strings - then it will convert the list of strings into a jquery object, before returning that jquery object. (like you can see in its implementation)

eirslett, line 290: What the code is trying to do is, `getAllSpans`, which isn't implemented, so it loops through all the service names, and then gets all spans for every service name.

yurishkuro:
lgtm

zipkin-web/src/main/resources/app/js/component_ui/trace.js
codefromthecrypt, line 290: I'll make an attempt at documenting ^^ in a followup. thanks for the explanation!

PR #966, opened 2016-02-12
Documents logic recently discussed in #964

codefromthecrypt:


codefromthecrypt:
@eirslett can you verify this is correct? cc @yurishkuro

eirslett:
Looks good to me!

PR #969, opened 2016-02-12
Frontend routing

eirslett:
Introduce crossroads, a JavaScript URL router. This will let us
perform logic client-side based on which page is served. The first
use-case is to only mount the relevant flight.js components for
which page we are looking at.

query-string is an npm module that parses query strings. We
replace the home-baked jquery plugin with this npm module instead.

main.js is simplified and rewritten to ES6.

### DIFF
--- a/zipkin-web/src/main/resources/app/js/main.js
+++ b/zipkin-web/src/main/resources/app/js/main.js
@@ -1,4 +1,10 @@
-'use strict';
+import {compose, registry, advice, logger as withLogging, debug} from 'flight';
+import $ from 'jquery';
+import crossroads from 'crossroads';
+import queryString from 'query-string';
+import initializeDefault from './page/default';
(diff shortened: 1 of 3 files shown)


zipkin-web/src/main/resources/app/js/main.js
```diff
@@ -1,4 +1,10 @@
-'use strict';
+import {compose, registry, advice, logger as withLogging, debug} from 'flight';
```
eirslett, line 1: Finally, ES6 destructuring (similar to unapply/pattern matching in Scala) and harmony modules!

codefromthecrypt, line 1: This is much nicer

codefromthecrypt:
Nice progress.. Thanks!

eirslett:
Thanks for the quick review!

PR #972, opened 2016-02-13
Render page title and environment name client-side

eirslett:
Add a config.js endpoint that sets relevant zipkin-web
configuration in window.config. (environment name,
default query limit etc)

Moving rendering of title and environment to the client
side means that layout.mustache has no other data to be
injected, so it can act as a pure decorator. (Which helps
us move from server-side rendering to client-side rendering)

### DIFF
--- a/zipkin-web/src/main/resources/app/js/page/dependency.js
+++ b/zipkin-web/src/main/resources/app/js/page/dependency.js
@@ -3,7 +3,9 @@
 define(
   [
     'moment',
+    'query-string',
     '../component_data/dependency',
+    '../component_ui/environment',
     '../component_ui/dependencyGraph',
     '../component_ui/serviceDataModal',
     '../component_ui/timeStamp',
@@ -11,7 +13,9 @@ define(
   ],
 
   function (moment,
+            queryString,
             DependencyData,
+            {environment: EnvironmentUI},
             DependencyGraphUI,
             ServiceDataModal,
             TimeStampUI,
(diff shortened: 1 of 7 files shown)


zipkin-web/src/main/resources/templates/v2/layout.mustache
```diff
@@ -26,9 +26,7 @@
         </div>
         <div class="navbar-right" style="width: 50%">
           <div class="navbar-right" style='padding-left: 15px'>
-            <div class="navbar-brand">
-              <span class='muted'>{{environment}}</span>
-            </div>
+            <p class="navbar-text muted" id="environment"></p>
```
eirslett, line 29: I changed the markup while I was here... navbar-text is more correct than navbar-brand, according to the Bootstrap docs.

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
```diff
@@ -40,6 +40,13 @@ class Handlers(mustacheGenerator: ZipkinMustache, queryExtractor: QueryExtractor
     }
   }
 
+  case class ConfigRenderer(config: Map[String, _]) extends Renderer {
+    def apply(response: Response) {
+      response.contentType = "application/javascript"
+      response.contentString = "window.config = " + ZipkinJson.writeValueAsString(config) + ";"
```
jcarres-mdsol: maybe 

```
s"window.config =  ${ZipkinJson.writeValueAsString(config)};"
```

zipkin-web/src/main/resources/app/js/page/dependency.js
```diff
@@ -20,16 +24,13 @@ define(
     return initialize;
 
     function initialize() {
-      // parse query string to set form
-      var params = {}, postBody = location.search.substring(1), regex = /([^&=]+)=([^&]*)/g, m;
```
codefromthecrypt, line 24: nice

zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
```diff
@@ -99,9 +99,13 @@ trait ZipkinWebFactory { self: App =>
...
       ("/api/v1/traces", handleRoute(queryClient, "/api/v1/traces")),
       // TODO: Once the following are javascript-only, we can move remove zipkin-web
-      ("/", addLayout("Index", environment()) andThen handleIndex(queryClient)),
-      ("/traces/:id", addLayout("Traces", environment()) andThen handleTraces(queryClient)),
-      ("/dependency", addLayout("Dependency", environment()) andThen handleDependency())
+      ("/", addLayout andThen handleIndex(queryClient)),
+      ("/traces/:id", addLayout andThen handleTraces(queryClient)),
+      ("/dependency", addLayout andThen handleDependency()),
+      ("/config.js", handleConfig(Map(
+        "environment" -> environment(),
```
codefromthecrypt, line 106: cool! so basically this endpoint exports the defaults (and environment flag) that previously were done server-side.

codefromthecrypt:
LGTM

zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
eirslett, line 106: That's right. And you can read them in window.config.

PR #973, opened 2016-02-13
Move all mustache rendering to the browser

eirslett:
This means the zipkin-web backend is now only
basically a JSON api + delivering static assets.
There's a lot of complexity going on in the
transformation from trace/span JSON data to
data that is fed into mustache templates. These
transformations are still written in Scala, but
hosted at /modelview endpoints. Ideally, they
should be ported to Scala.

Some code had to be moved around - event handlers
must be bound after the HTML is rendered, which now
happens at a later stage than it used to (when it
was server-side rendered).

### DIFF
--- a/zipkin-web/src/main/resources/app/js/component_data/default.js
+++ b/zipkin-web/src/main/resources/app/js/component_data/default.js
@@ -1,0 +1,15 @@
+import {component} from 'flight';
+import $ from 'jquery';
+
+export const DefaultData = component(function defaultData() {
+  this.after('initialize', function() {
+    $.ajax('/modelview' + window.location.pathname + window.location.search, {
+      type: "GET",
+      dataType: "json",
+      context: this,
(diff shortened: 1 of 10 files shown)


yurishkuro:
@eirslett since you're actively working on renovation of the UI, do you mind creating an issue with a high level roadmap, like
- [ ] do X
- [ ] do Y
  ?

This will provide better visibility into how far the work is / going to be and when it's best to upgrade the forks.  I had a couple of minor UI fixes in our fork that I planned to push back to openzipkin, but with these and upcoming changes they may not work.

eirslett:
That's a good idea, I'll make one! If you show me the fixes you did, I'll try to make them work together with my changes as well!

codefromthecrypt:
@yurishkuro good idea. I'm also guilty of not logging an overall "what I'm working towards" thing pretty often.

@eirslett LGTM

PR #976, opened 2016-02-15
Adds all-jar and Intellij instructions

codefromthecrypt:


codefromthecrypt:
cc @yurishkuro @joshlong

README.md
```diff
@@ -71,6 +71,32 @@ Developers: [https://groups.google.com/group/zipkin-dev](https://groups.google.c
...
+
+*Note* Do not import as a Gradle project. If you do, you'll likely see classpath or other build related problems. That's why we use the command-line.
+
+To build the project file, close any open window for zipkin and (re)generate the content like so:
+
+```bash
+$ ./gradlew cleanIdea idea
+```
+
+Import the the result via File, Open, (path you invoked gradlew). Dismiss any pop-ups about unlinked Gradle projects, as clicking those will likely break your project.
```
yurishkuro, line 98: not sure if other people get that, but when I import the project on a clean repo, IntelliJ cannot build it because of configuration files in `zipkin-collector-service/config` and `zipkin-query-service/config`. These dirs are somehow added as both test resource and some other type, and I have to open project settings and unselect them to make sure they are only included as test resource (so that IntelliJ doesn't try to compile them).

codefromthecrypt, line 98: > > +Import the the result via File, Open, (path you invoked gradlew). Dismiss any pop-ups about unlinked Gradle projects, as clicking those will likely break your project.
> 
> not sure if other people get that, but when I import the project on a
> clean repo, IntelliJ cannot build it because of configuration files in
> zipkin-collector-service/config and zipkin-query-service/config. These
> dirs are somehow added as both test resource and some other type, and I
> have to open project settings and unselect them to make sure they are only
> included as test resource (so that IntelliJ doesn't try to compile them).
> 
> You get this when you do import? or file, open? If using the offline
> generation and open (plus never clicking a popup), I've not had to correct
> anything to get build or test working.

yurishkuro, line 98: I open the project file generated by gradle. It then offers to convert the project file to a project dir, that's the only popup. And then it always fails to build unless I remove config files from being considered "source"

codefromthecrypt, line 98: looks like there's an existing fix for this issue.. maybe it needs to be revised? let's punt to #978 for now.

PR #980, opened 2016-02-15
Adds CASSANDRA_LOCAL_DC for latency-based load-balancing

### LINKED ISSUE #967: Change cassandra to use latency-aware load balancing
Per https://github.com/openzipkin/zipkin/pull/958/files#r52631161, we should consider changing the load balancing policy to be latency aware. Perhaps @yurishkuro or @danchia could shed light on how this is impactful.

The change to CassandraSpanStoreFactory would look like this:

``` scala
new TokenAwarePolicy(new LatencyAwarePolicy.Builder(new DCAwareRoundRobinPolicy(dc)).build())
```

One thing to consider is what to default `dc` to? I suppose this will end up another flag, like `CASSANDRA_LOCAL_DC`

codefromthecrypt:
This allows collector and query services to avoid calling C\* hosts in
remote datacenters.

Fixes #967

### DIFF
--- a/zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
+++ b/zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
@@ -18,9 +18,7 @@ package com.twitter.zipkin.cassandra
 import java.net.InetSocketAddress
 
 import com.datastax.driver.core.Cluster
-import com.datastax.driver.core.policies.LatencyAwarePolicy
-import com.datastax.driver.core.policies.RoundRobinPolicy
-import com.datastax.driver.core.policies.TokenAwarePolicy
+import com.datastax.driver.core.policies.{RoundRobinPolicy, DCAwareRoundRobinPolicy, LatencyAwarePolicy, TokenAwarePolicy}
 import com.google.common.net.HostAndPort
 import com.twitter.app.App
 import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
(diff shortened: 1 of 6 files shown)


zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
```diff
@@ -68,7 +67,12 @@ trait CassandraSpanStoreFactory {self: App =>
     if (cassandraUsername.isDefined && cassandraPassword.isDefined)
       builder.withCredentials(cassandraUsername(), cassandraPassword())
     builder.withRetryPolicy(ZipkinRetryPolicy.INSTANCE)
-    builder.withLoadBalancingPolicy(new TokenAwarePolicy(new LatencyAwarePolicy.Builder(new RoundRobinPolicy()).build()))
+    builder.withLoadBalancingPolicy(new TokenAwarePolicy(
+      if (cassandraLocalDc.isDefined)
+        new LatencyAwarePolicy.Builder(DCAwareRoundRobinPolicy.builder().withLocalDc(cassandraLocalDc()).build()).build()
+      else
+        new RoundRobinPolicy()
```
yurishkuro, line 74: why skip latency-aware policy in this case?

codefromthecrypt, line 74: mistake!

codefromthecrypt, line 74: fixed. thanks for the catch

yurishkuro:
lgtm

PR #982, opened 2016-02-16
adds ruby trace and c# trace links to the README

jcarres-mdsol:
adding links to the other tracing clients


codefromthecrypt:
Thanks!

ps the anorm tests flake out sometimes.. haven't looked into why, clicking to rebuild travis on fail usually works.

Issue #718, opened 2015-09-19
Allow caching of getServices in the UI


codefromthecrypt:
From @yurishkuro 

> the query to get all service names in Cassandra needs to hit every node in the cluster. We've been seeing strange huge latencies with it recently, still investigating. To mitigate the impact, I've added a cache of service names in Cassandra span storage, which periodically refreshes itself.

I think we should solve this at the HTTP abstraction, ex. sending cache headers back, so that the client doesn't call so much.

There are two endpoints on zipkin-web that call `getServices`
- `/` - part of a massive thing that includes enumerating all services, to pass to `index.mustache`
- `/api/services` - possibly unused json string list

I see two things to do:
- change `index.mustache` to use javascript to GET /api/services and fill the `Service Name` selector by itself
- change `handleServices` to return cache headers

codefromthecrypt:
@mzagar @eirslett either of you have some cycles to help with this by chance?

mzagar:
I believe I can take a look at this one.

codefromthecrypt:
Great. I suspect this won't impact you if you are only working in
JavaScript layer. But just in case:
https://github.com/openzipkin/zipkin/pull/728

michaelsembwever:
> the query to get all service names in Cassandra needs to hit every node in the cluster. 

best to increase RF of the zipkin keyspace.
(answers more to the availability than performance)

> I think we should solve this at the HTTP abstraction, ex. sending cache headers back, so that the client doesn't call so much.

:+1:

codefromthecrypt:
so realizing that the getServices isn't actually called by the javascript yet. Also, I noticed a special case. WDYT about this?

if services.size < 3 don't cache
otherwise set TTL on the http response to 30 seconds?

mzagar:
Isn't 30 seconds a bit low - are new services created that often? I was thinking about adding a property allowing to set this timeout to a custom value at web startup and setting the default to 5 minutes, is that too long?

Optimization when not to cache sounds good.

I'll make a PR fpr the js side later today. I noticed /api/services was removed from available endpoints... should I put that back in?

codefromthecrypt:
Honestly I don't know what's the right value.. there's a special case of
bootstrapping, which I think the minimum count would cover.. Perhaps 5m
after that is a decent value. Probably only way to find out is to spike it!

Thanks for the offer for the javascript change! Note that ideally we'd
remove the call from Handlers.scala while at it.

[...]

PR #983, opened 2016-02-16
Remove all server-side mustache code from zipkin-web

eirslett:
The server now only delivers a static index.html page,
which in turn loads JavaScript/CSS to run the application.
- Remove mustache dependencies from Gradle
- Add index.html "container page" for hosting the application
- Add new Flight UI component: "common", for rendering the
  application layout
- The EnvironmentUI component is now rendered by the "common"
  component
- Use events to trigger loading of serviceName/spanName to
  populate dropdowns


zipkin-web/src/main/resources/templates/v2/index.mustache
```diff
@@ -2,19 +2,11 @@
   <form class="form-inline" role="form">
     <div class="form-group">
       <select data-placeholder="Service Name" class="form-control input-sm" style="width: 234px;" name="serviceName" id="serviceName">
-        <option value=""></option>
```
yurishkuro, line 5: in our installation, populating service names takes from one to several seconds.  Adrian tried to add caching control some time ago, but it doesn't look like it helped in our case. Does it makes sense to cache the whole HTML of the dropdown in the UI?

codefromthecrypt, line 5: Ps happy to dig deeper into the cache-control header thing if you want to
open an issue on it (or reopen the old one).

eirslett, line 5: It's all rendered based on JSON API calls, so we should rather add cache-control headers to the API. (In zipkin-query, and zipkin-web should forward the cache headers in CopyRenderer)

yurishkuro, line 5: mine concern is that retrieving 100s of service names even from cache and rebuilding the DOM is expensive.

eirslett, line 5: What is the expensive part of it? The call between the query server and the storage engine/cassandra or the browser rendering? It sounds strange that it takes such a long time... is it due to very many service names?
JavaScript converting a list of names into html is very quick (as quick as if it were done in Java) so it sounds like either IO or an expensive algorithm is to blame...

Does it take several seconds before this commit? Or only after this commit? Or regardless?

codefromthecrypt, line 5: This thread is about #718, so let's keep it contained there. There's been overwhelming support to undo the server-side mustache templates and I don't want to halt this PR on account of an old issue

PR #985, opened 2016-02-17
Uses constants for core annotations and corrects zipkin-web service name

codefromthecrypt:


zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
```diff
@@ -67,7 +67,7 @@ trait ZipkinWebFactory { self: App =>
   * we get the host after the [[queryDest]] flag has been parsed.
   */
   lazy val queryClient = new HttpClient(
-    httpService = Http.client.configured(param.Label("zipkin-query"))
+    httpService = Http.client.configured(param.Label("zipkin-web"))
```
codefromthecrypt, line 70: client annotations should be labeled with the name of the instrumented client. In this case, I mistakenly used the name of the destination. If you looked at the span data before this change, you'd see "cs" labeled with the endpoint zipkin-query. This is incorrect as core annotations should be labeled with the local service name.

zipkin-common/src/test/scala/com/twitter/zipkin/common/SpanTest.scala
```diff
@@ -51,50 +52,50 @@ class SpanTest extends FunSuite {
...
     span = span.copy(binaryAnnotations = List.empty)
 
-    assert(span.serviceName === Some("cs"))
+    assert(span.serviceName === Some("instrumented_client"))
 
     // Finally is the label of the local component's endpoint
-    span = span.copy(binaryAnnotations = List(this.span.binaryAnnotations.last))
+    span = span.copy(annotations = List.empty, binaryAnnotations = List(this.span.binaryAnnotations.last))
 
-    assert(span.serviceName === Some("cs"))
```
codefromthecrypt, line 92: test bug

PR #986, opened 2016-02-17
Rip out parts of modelview for the index page

eirslett:
Some of the properties (query string parsing)
were trivial to move to JavaScript.
.babelrc was added to load plugins, namely
the object spread transform plugin, which lets us
use that syntax (like the diff in page/default.js)

### DIFF
--- a/zipkin-web/src/main/resources/app/js/page/default.js
+++ b/zipkin-web/src/main/resources/app/js/page/default.js
@@ -4,6 +4,7 @@ define(
   [
     'flight',
     'timeago',
+    'query-string',
     '../component_data/default',
     '../component_data/spanNames',
     '../component_data/serviceNames',
@@ -22,6 +23,7 @@ define(
   function (
     {component},
     timeago,
+    queryString,
     {DefaultData},
     SpanNamesData,
     ServiceNamesData,
@@ -42,8 +44,22 @@ define(
         window.document.title = 'Zipkin - Index';
         DefaultData.attachTo(document);
 
+        const query = queryString.parse(window.location.search);
+
         this.on(document, 'defaultPageModelView', function(ev, modelView) {
-          this.$node.html(defaultTemplate(modelView));
+          const limit = query.limit || window.config.queryLimit;
+          const minDuration = query.minDuration;
+          const endTs = query.endTs || new Date().getTime();
+          const serviceName = query.serviceName || '';
+          const annotationQuery = query.annotationQuery || '';
+          this.$node.html(defaultTemplate({
+            limit,
+            minDuration,
+            endTs,
+            serviceName,
+            annotationQuery,
+            ...modelView
+          }));
 
           SpanNamesData.attachTo(document);
           ServiceNamesData.attachTo(document);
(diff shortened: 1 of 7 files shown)


zipkin-web/src/main/resources/app/js/page/default.js
```diff
@@ -42,8 +44,22 @@ define(
...
+          const endTs = query.endTs || new Date().getTime();
+          const serviceName = query.serviceName || '';
+          const annotationQuery = query.annotationQuery || '';
+          this.$node.html(defaultTemplate({
+            limit,
+            minDuration,
+            endTs,
+            serviceName,
+            annotationQuery,
+            ...modelView
```
eirslett, line 61: object spread syntax: for each key in modelView, put the key and its value into this object.

zipkin-web/src/main/resources/app/js/page/default.js
```diff
@@ -42,8 +44,22 @@ define(
...
+
         this.on(document, 'defaultPageModelView', function(ev, modelView) {
-          this.$node.html(defaultTemplate(modelView));
+          const limit = query.limit || window.config.queryLimit;
+          const minDuration = query.minDuration;
+          const endTs = query.endTs || new Date().getTime();
+          const serviceName = query.serviceName || '';
+          const annotationQuery = query.annotationQuery || '';
+          this.$node.html(defaultTemplate({
+            limit,
```
eirslett, line 56: shorthand for `{ limit: limit, minDuration: minDuration, endTs: endTs }` etc

codefromthecrypt:
I'm excited

zipkin-web/src/main/resources/app/js/page/default.js
codefromthecrypt, line 56: I like how the defaults are explicit

codefromthecrypt:
not sure why, but travis is failing on `:zipkin-web:compileTestScala FAILED`

eirslett:
I think I found it! I changed the signature of a method.

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
```diff
@@ -221,25 +215,13 @@ class Handlers(queryExtractor: QueryExtractor) {
...
-          span => Map("name" -> span, "selected" -> (if (Some(span) == spanName) "selected" else ""))
-        }
-
+      for (traces <- tracesCall) yield {
+        val (annotations, binaryAnnotations) = queryExtractor.getAnnotations(req)
         var data = Map[String, Object](
-          ("serviceName" -> serviceName),
-          ("endTs" -> queryExtractor.getTimestampStr(req)),
-          ("annotationQuery" -> req.params.get("annotationQuery").getOrElse("")),
-          ("spans" -> spanList),
```
codefromthecrypt, line 233: was this obviated prior to this change?

eirslett, line 233: Yes, the spans are loaded async from the frontend code. (I think they were loaded twice)

eirslett:
Looks like the build is fixed now.

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
codefromthecrypt, line 233: cool.. I should be able to see this in self-tracing!

codefromthecrypt:
Thanks for the help, eirik!

PR #990, opened 2016-02-19
Fix UI bug - trace placeholder wasn't shown

eirslett:
After the UI refactoring, a bug in the handling of
queryResults surfaced. This means the "placeholder" that is
shown before you have made any queries was never rendered.

This fix adds a "queryWasPerformed" property that is
computed client-side. The structure of the JSON returned
from the server is also simplified a bit.

@yurishkuro I think this should fix the bug you reported.

### DIFF
--- a/zipkin-web/src/main/resources/app/js/page/default.js
+++ b/zipkin-web/src/main/resources/app/js/page/default.js
@@ -52,12 +52,15 @@ define(
           const endTs = query.endTs || new Date().getTime();
           const serviceName = query.serviceName || '';
           const annotationQuery = query.annotationQuery || '';
+          const queryWasPerformed = serviceName && serviceName.length > 0;
           this.$node.html(defaultTemplate({
             limit,
             minDuration,
             endTs,
             serviceName,
             annotationQuery,
+            queryWasPerformed,
+            count: modelView.traces.length,
             ...modelView
           }));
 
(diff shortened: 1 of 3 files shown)


zipkin-web/src/main/resources/app/js/page/default.js
```diff
@@ -52,12 +52,15 @@ define(
           const endTs = query.endTs || new Date().getTime();
           const serviceName = query.serviceName || '';
           const annotationQuery = query.annotationQuery || '';
+          const queryWasPerformed = serviceName && serviceName.length > 0;
```
yurishkuro, line 55: perhaps is would be better to define a flag based on `modelView.traces.length`?  What happens if I GET the home page with `?serviceName=x`?

eirslett, line 55: This merely replicates the existing behavior :wink:
I agree that we should replace it with traces.length > 0.

codefromthecrypt:
Thanks for the continued line of bug fixes @eirslett and thanks for the watchful eye and engaging comments, @yurishkuro!

zipkin-web/src/main/resources/app/js/page/default.js
eirslett, line 55: > What happens if I GET the home page with ?serviceName=x?

The backend performs a query against the storage engine, for traces with serviceName x. So the query is performed.

PR #991, opened 2016-02-20
Fix UI navbar highlighting

eirslett:
The "Find a trace" menu element was hard-coded
to always being selected.

This fix creates a NavbarUI component, which
listens to "navigate" events and highlights the
correct menu item, depending on the route.

### DIFF
--- a/zipkin-web/src/main/resources/app/js/page/dependency.js
+++ b/zipkin-web/src/main/resources/app/js/page/dependency.js
@@ -3,6 +3,7 @@
 define(
   [
     'moment',
+    'flight',
     'jquery',
     'query-string',
     '../component_data/dependency',
@@ -14,6 +15,7 @@ define(
   ],
 
   function (moment,
+            {component},
             $,
             queryString,
             DependencyData,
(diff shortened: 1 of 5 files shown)


zipkin-web/src/main/resources/app/js/page/dependency.js
```diff
@@ -24,22 +26,28 @@ define(
...
-      window.document.title = 'Zipkin - Dependency';
-      $('.content').html(dependenciesTemplate());
+        this.$node.html(dependenciesTemplate());
 
-      const {startTs, endTs} = queryString.parse(location.search);
-      $('#endTs').val(endTs || moment().valueOf());
-      $('#startTs').val(startTs || moment().valueOf() - 7*24*60*60*1000); // set default startTs 7 days ago;
+        const {startTs, endTs} = queryString.parse(location.search);
+        $('#endTs').val(endTs || moment().valueOf());
+        $('#startTs').val(startTs || moment().valueOf() - 7*24*60*60*1000); // set default startTs 7 days ago;
```
yurishkuro, line 38: is this the same behavior as before?  I thought the lookback interval was configurable via env variable.

eirslett, line 38: It seems to be configurable as a flag in zipkin-query. To be honest, I'm not 100 % sure what it does.

codefromthecrypt, line 38: So before Yuki made a change, we had no way to specify the interval of
dependencies query in the UI. If we want to poke a hole and default it from
env on the web or query server, I suppose that would be in the other config
endpoint?

codefromthecrypt, line 38: In general I think now is a good time to think of things like this as
config servers vs abstraction leaks (like JavaScript reading env values of
things). That's why I prefer the config endpoint approach as at least it
contains it until there's a better answer.

eirslett, line 38: Could somebody find a commit in the history (before I made the refactorings) with the Scala logic that sets it up? I'm might have overseen it.

codefromthecrypt, line 38: I'll look it up and ping you. This change made a historical one obvious and shouldn't be held back on account of moving it two spaces to the right.

PR #992, opened 2016-02-21
Add support for JavaScript unit tets

eirslett:
Use karma as a test runner, mocha/chai as
the test framework, and run tests in
PhantomJS.


eirslett:
@abesto the node/npm/webpack/karma stuff slows down the build quite a lot. Could you please help out with some gradle magic to improve the build time? Incremental builds etc.?

codefromthecrypt:
last build before this was 6s faster, but most recent ones are 30s faster than this.

abesto:
Sure, sounds fun. The next few days look kinda chaotic, no promises on the timeline.

codefromthecrypt:
thinking we merge this to encourage test writing... I'm personally not deterred by the build time creep.

eirslett:
:-)

I'm also working on top of this patch to add unit tests while rewriting logic from Scala to JavaScript.
Eventually, I think the zipkin-web submodule will be packaged as a tar.gz or zip artifact, and then imported as a gradle dependency into zipkin-query and zipkin-java. (Which is a consideration when thinking about optimizing the build)

codefromthecrypt:
Ok so cool to merge this, of do you want to add more to it?

eirslett:
I'm good.

Issue #939, opened 2016-02-03
Zipkin v2 span model


codefromthecrypt:
We should not try to do all things as zipkin v2.. the large amount of interesting things may lead us to never even start. I believe that if we pare down do basics, creating a model that only includes features that work in zipkin 1 backends, we can start and finish something.

For example, zipkin 1 only really supports String -> String binary annotations.. let's make them String -> String for v2 even if they become fancy later.

Zipkin v1 allows "contributory" spans, where instrumentation are allowed to place multiple hosts into the same span. Let's prevent that, and use a portability technique google cloud trace does. Google Cloud Trace allows multiple spans with the same id, but they vary on span.kind. For example, core server annotations really imply a span.kind=server, where "sr" is span.startTs and "ss" is span.endTs.

We could probably literally use the google cloud trace model as zipkin v2 as it is almost exactly what we need.

https://cloud.google.com/trace/api/reference/rest/v1/projects.traces#TraceSpan

If we do something like this, I'm confident we can get v2 out in relatively short order, and still not prevent more interesting (and severe) change from occurring in later versions of zipkin.

codefromthecrypt:
here's an issue where binary annotations were a bit confusing (as they allow repeats on keys) https://github.com/openzipkin/brave/issues/136

yurishkuro:
> here's an issue where binary annotations were a bit confusing (as they allow repeats on keys)

is it fair to say that ^^^ is only confusing because the UI does not associate binary annotation with the "span kind" that contains them? For example, I made a fix that adds service name to SA/CA annotations

![image](https://cloud.githubusercontent.com/assets/3523016/13147772/3d2cc928-d628-11e5-93ea-b06f67e9e436.png)

Another common scenario I've seen with same keys is the `http.url`, which is sent by both client and server, and the value may or may not be the same.

If we adopt the suggestion that (spanID, spanKind, emittingService) combination is unique (which I think implies ditching core annotations and moving the Endpoint on the Span itself), then seemingly identical binary annotations may still be distinguishable by the service that emits them.

codefromthecrypt:
The above concern on repeating keys is more comprehensively reviewed here: https://github.com/openzipkin/zipkin/issues/989#issuecomment-198825497

If the span included a copy of the endpoint, it would be possible to support this, though let's bear in mind this isn't a popular feature.

codefromthecrypt:
When we get to actually implementing this, we should be very explicit about the safety any issues that remain.

[...]

PR #995, opened 2016-02-23
Supports reading multiple spans per Kafka message

### LINKED ISSUE #979: multiple spans and kafka collectors
`SpanStore.apply` is invoked by collectors to store spans in bulk (even if not in the same trace). The kafka receiver only sends a single span to this method. That means if optimizations are made on span stores (such as #961 for cassandra), kafka users won't necessarily benefit.
## Options to consider

The kafka transport encodes a span in a single KV message (key is null) at the moment. Let's see if there's a way to leverage the bulk storage api: `SpanStore.apply`.
### Send multiple spans per message and adjust kafka decoding

Kafka messages are currently single-span, and a single-entry list is not the same thrift bytes as a single-entry. To support multiple-span messages, we need to either peek at the bytes or vary on message key (adjusting collectors accordingly)
### separate out reading off kafka from writing to the span store

@yurishkuro mentions that reading from kafka into a local queue might help. For example, a span store reader could consume N spans off that queue. https://github.com/openzipkin/zipkin/issues/961#issuecomment-182956930
### Scribe

The scribe transport encodes a span in a single log entry. However the scribe api allows you to log multiple entries, which means that the scribe receiver can handle a bunch of spans at the same time (same network request).
### Http

The http transport encodes a list of spans in json or thrift format, these are passed directly to the span store.
### TChannel

TChannel-based communication between clients and central collectors batch multiple spans into one request.

codefromthecrypt:
Kafka messages have binary payloads and no key. The binary contents are
serialized TBinaryProtocol thrift messages. This change peeks at thei
first bytes to see if it is a List of structs or not, reading
accordingly.

This approach would need a revision if we ever add a Struct field to
Span. However, that is unlikely. At the point we change the structure of
Span, we'd likely change other aspects which would make it a different
struct completely (see #939). In such case, we'd add a key to the kafka
message of the span version, and not hit the code affected in this
change.

Fixes #979


codefromthecrypt:
cc @clehene @kristofa @prat0318 

so after this.. instrumentation can choose to either TBinaryProtocol encode a single span, or a list of spans.

eirslett:
I must admit this feels very hacky and brittle to me... but it could work as a temporary migration path. In the future, I suggest we should only support collecting a list of spans. For Zipkin v2, we could provide a deprecation warning - add warning log messages whenever the collector receives single spans, and not a list of spans?

yurishkuro:
Did anyone do any benchmarks to show that writing/reading Kafka messages with a single span is slowed than messages with batch of spans?

I do agree with @eirslett that APIs that accept just a single span instead of a collector are suboptimal. Although logging a warning would just spam the collector's log, it can't do much about what clients submit.

codefromthecrypt:
I doubt there is an existing benchmark for any Kafka zipkin reporter, much
less single span vs list. I think the key here is that list gives more
flexibility for instrumentation, including a means to benchmark and report
back.

Personally, I am cool chasing instrumentation to have them switch to list,
as list of one is only several bytes overhead. Kafka is still newish
(except Ruby).

Also happy to add a "log once" message that notes the span id and endpoint
of single-span Kafka messages. That could prevent logs from cluttering.

I agree this is hacky just don't know a cheaper way to help folks move off
single span without breaking them.

Adding a key to the message would seem less hacky, but take more discussion
for example. I was hoping to save that energy for v2 (ack list-only
.)

I also think benchmarks would be helpful, but there's a chicken egg.
Ideally, I would like Prat to respond back when he can try list.

I'm unsurprised about meh reactions.. Anyone feel we shouldn't go down the
peek path with above context in mind?

eirslett:
Don't get me wrong, I'm +1 for this change. I don't like the solution, but I still think it's the best solution - there's no smooth migration path. One possible alternative would be to consume span collections on a separate kafka topic, but then we get lots of additional complexity from handling two topics.

prat0318:
I will try running the collector with this patch to check the perf

On Wednesday, February 24, 2016, Eirik Sletteberg notifications@github.com
wrote:

> Don't get me wrong, I'm +1 for this change. I don't like the solution, but
> I still think it's the best solution - there's no smooth migration path.
> One possible alternative would be to consume span collections on a separate
> kafka topic, but then we get lots of additional complexity from handling
> two topics.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/openzipkin/zipkin/pull/995#issuecomment-188353240.

## 

Prateek Agarwal

yurishkuro:
is there indeed a lot of setup overhead to use a different topic for different message format?

prat0318:
From usability perspective, maintaining multiple topics would be hard imo. For testing perf and maintaining backwards compat, it will be helpful though

prat0318:
In my experiment, each Trace has 9 spans, which means without patch 9 kafka messages per trace.

| - | Traces/s/partition | Spans/s/partition | Messages/s/partition |
| --- | --- | --- | --- |
| Production | 174.5 | 1570 | 1570 |
| Consumption | **40.5** | 364 | 364 |

**With bundling**, all 9 spans are batched up in 2 kafka messages.

| - | Traces/s/partition | Spans/s/partition | Messages/s/partition |
| --- | --- | --- | --- |
| Production | 200 | 1800 | 400 |
| Consumption | **175** | 1580 | 351 |

As we can see, we straight away get a perf improvement of **~4.5 times**. This should
be ~9 times if i batch up all spans in a single kafka message (it just needed some
more code changes at my end, so i just went ahead with 2 messages).

So, what i see is Message consumption rate is around constant to ~350 but we can increase
our Trace consumption rate by bundling together. So, i am super excited to have this
patch merged to master. :+1: 

cc @adriancole @yurishkuro

codefromthecrypt:
I've not heard any feedback against from a kafka transport user, and this clearly will help @prat0318 and move us in the right direction of moving towards lists as the defacto unit-of-transport.

merging

Issue #994, opened 2016-02-22
support /heath endpoint


jcarres-mdsol:
In our testing tools we are checking that the zipkin query server is running but doing a HEAD request to the API.
This will return 404
But if we do a GET request it will respond normally with 200.
That's an extra burden (to be sending data when it does need to) and the API should be responding to HEAD also.

Optionally or in addition the zipkin query server could have some /status or similar endpoint to check its health.

yurishkuro:
@jcarres-mdsol do you have suggestion on what the health endpoint should be checking?  I often see those implemented as `return OK`, which only tells the server is running, not that it's healthy. If we decide, for example, to include some db query in the health check, then it may require extending span storage api.

eirslett:
It would be nice for the frontend UI to poll on a health endpoint as well. I think it sould be sufficient to return OK, to verify that the server is running and reachable. We could then show a simple "query server unreachable" message, which is more helpful than a UI that doesn't work.

codefromthecrypt:
My guess is the most important  state is if the spanstore connection is
OK.  I made a interface like this at Netflix and then exported it to the
health check handler. It is better than just http and simpler than checking
many things.

Basically we add a method span stores need to implement that returns true
if OK.  Span stores use the cheapest valid way to do that. We can export
that to an http call (which load balancers etc can check)

Ex from a DNS project

https://github.com/Netflix/denominator/blob/master/route53/src/main/java/denominator/route53/HostedZonesReadable.java

jcarres-mdsol:
Yes, usually is either "OK"  or a DB check. 
We can add the route now with a simple "OK" and later as the check in the spanstores are implemented upgrade it to check these.

codefromthecrypt:
looks like there's an OK in the query server already:

https://github.com/twitter/finatra/blob/master/inject/inject-server/src/main/scala/com/twitter/inject/server/TwitterServer.scala#L112

$ curl localhost:9901/health

OK
Do we want to map health onto the main port? (9411)

jcarres-mdsol:
It will be useful, else you need to open firewall ports just because of that

codefromthecrypt:
ok, so localhost:9411/health?

jcarres-mdsol:
Sounds good!

On Tue, Feb 23, 2016 at 10:39 AM, Adrian Cole notifications@github.com
wrote:

> ok, so localhost:9411/health?
> 
> —
> Reply to this email directly or view it on GitHub
[...]

PR #997, opened 2016-02-23
Adds /health endpoint to zipkin-query and zipkin-web

codefromthecrypt:
Various platforms have admin instances that expose health, and Twitter's
does so on a different port. By introducing a "/health" endpoint, we can
start a portable conversation about health that doesn't rely on platform
conventions or opening more ports on the firewall. The use case for this
is load balancer health checks and the zipkin javascript UI.

See #994

### DIFF
--- a/zipkin-query/src/main/scala/com/twitter/zipkin/query/ZipkinQueryController.scala
+++ b/zipkin-query/src/main/scala/com/twitter/zipkin/query/ZipkinQueryController.scala
@@ -24,6 +24,10 @@ class ZipkinQueryController @Inject()(spanStore: SpanStore,
                                       @Flag("zipkin.queryService.servicesMaxAge") servicesMaxAge: Int,
                                       @Flag("zipkin.queryService.lookback") val defaultLookback: Long) extends Controller {
 
+  get("/health") { request: Request =>
+    "OK\n" // TODO: expose SpanStore.ok? or similar per #994
+  }
+
   post("/api/v1/spans") { request: Request =>
     val spans: Try[List[Span]] = try {
       Success(request.mediaType match {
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
cc @jcarres-mdsol @eirslett @yurishkuro

zipkin-query/src/main/scala/com/twitter/zipkin/query/ZipkinQueryController.scala
```diff
@@ -24,6 +24,10 @@ class ZipkinQueryController @Inject()(spanStore: SpanStore,
                                       @Flag("zipkin.queryService.servicesMaxAge") servicesMaxAge: Int,
                                       @Flag("zipkin.queryService.lookback") val defaultLookback: Long) extends Controller {
 
+  get("/health") { request: Request =>
+    "OK\n" // TODO: expose SpanStore.ok? or similar per #994
```
codefromthecrypt, line 28: ps this TODO can be satisfied later, as even this check is somewhat decent. Ex guice wiring must have worked.

zipkin-query/src/test/scala/com/twitter/zipkin/query/ZipkinQueryServerFeatureTest.scala
```diff
@@ -876,4 +876,11 @@ class ZipkinQueryServerFeatureTest extends FeatureTest with MockitoSugar with Be
       andExpect = Ok,
       withJsonBody = "[ ]")
   }
+
+  "health checks OK" in {
+    server.httpGet(
+      path = "/health",
+      andExpect = Ok,
+      withBody = "OK\n")
```
codefromthecrypt, line 884: `\n` is the same convention as twitter server.. also helps when using curl.

eirslett:
:shipit:

zipkin-query/src/test/scala/com/twitter/zipkin/query/ZipkinQueryServerFeatureTest.scala
jcarres-mdsol, line 884: I usually see only `OK` but does nor really matter, I do not think any tool would check anything other than the 200 response

jcarres-mdsol:
LGTM

jfeltesse-mdsol:
nice! :confetti_ball:

PR #1000, opened 2016-02-24
set the default for dependencies to 1 day of data

jcarres-mdsol:
In our system we had 2.5million annotations in the last week, when we go to the dependency page we need to wait almost 4 minutes to get the result, people think the page does not work.

Leaving the default in the last one day, I hope that will be better for us and also a sane default as at least the page returns faster and users can change the date if they want.
@adriancole  @eirslett


codefromthecrypt:
absolutely agree. This is the older behavior. At some point, we should move this default to /config.js

codefromthecrypt:
this default was accidental for the dependencies view, so reverting to the previous default of 1 day wfm

PR #1011, opened 2016-03-02
Fleshes out log message when cassandra spans are missing timestamp

### LINKED ISSUE #1010: Add more details to warning: "span with no first or last timestamp"
Currently the log line looks like so:

```
08:02:23.456 [pool-2-thread-5] WARN  o.t.z.storage.cassandra.Repository - span with no first or last timestamp. if this happens a lot consider switching back to SizeTieredCompactionStrategy for zipkin.traces
```

It will be helpful to find more details about the trace, if we can log `traceId` and `spanName` in the log line.

codefromthecrypt:
Before, the log message about spans missing a timestamp had no context.
This made it hard to narrow down instrumentation that might be
responsible. In this change, the span's name and trace id are logged.

Ex. `Span 456_473519988_1241441653 in trace 123 had no timestamp...`

Fixes #1010

### DIFF
--- a/zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
+++ b/zipkin-cassandra-core/src/main/java/org/twitter/zipkin/storage/cassandra/Repository.java
@@ -253,9 +253,9 @@ public final class Repository implements AutoCloseable {
 
         try {
             if (0 == timestamp && metadata.get("traces.compaction.class").contains("DateTieredCompactionStrategy")) {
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
verified by running CassandraSpanStoreSpec

```
[ScalaTest-run] WARN org.twitter.zipkin.storage.cassandra.Repository - Span 456_473519988_1241441653 in trace 123 had no timestamp. If this happens a lot consider switching back to SizeTieredCompactionStrategy for zipkin.traces
```

prat0318:
:+1:

PR #1012, opened 2016-03-02
Documents thrift encoding of Kafka messages

codefromthecrypt:


codefromthecrypt:
cc @abesto @kristofa @clehene @jcarres-mdsol

prat0318:
sg2m

zipkin-receiver-kafka/README.md
```diff
@@ -20,6 +20,25 @@ $ KAFKA_ZOOKEEPER=127.0.0.1:2181 bin/collector
 $ KAFKA_ZOOKEEPER=127.0.0.1:2181 KAFKA_TOPIC=notzipkin bin/collector
 ```
 
+### Encoding spans into Kafka messages
```
codefromthecrypt, line 23: @abesto PS this is the exact encoding of the POST /api/v1/spans endpoint (when content-type is application/x-thrift). Maybe we can move this to the new website and link out to there at some point.

zipkin-receiver-kafka/README.md
```diff
@@ -20,6 +20,25 @@ $ KAFKA_ZOOKEEPER=127.0.0.1:2181 bin/collector
...
+  writeTBinaryProtocol(spans(i))
+}
+```
+
+If using [zipkin-java](https://github.com/openzipkin/zipkin-java), `Codec.THRIFT.writeSpans(spans)`
+implements the above.
+
+#### Legacy encoding
+Versions before 1.35 accepted a single span per message, as opposed to a list per message. This
+practice is deprecated, but still supported.
```
codefromthecrypt, line 40: ps there isn't a likely scenario when we will remove support for single-span-per-message prior without a new model (like zipkin 2). In other words, this is just a request to please not write new code that does this.

codefromthecrypt, line 40: reason is that instrumentation is hard to change. ex htrace is out in the wild and uses this encoding.

zipkin-receiver-kafka/README.md
abesto, line 23: Sounds good, I've made a note of it. Thanks for the mention.

PR #1014, opened 2016-03-03
Span should get children IDs

### LINKED ISSUE #999: [1.33.2] Collapse + button in Trace detail page not working
Using the data in tracegen, when clicking the + button at the left of  service `natoque`, the service does not expand.

Before clicking +:

After clicking +:

eirslett:
It looks like there was a bug in renderTrace, probably an oversight after a refactoring.
In the web UI, (the data/modelview prepared for trace rendering) every span has a list of
children/child span IDs. This list only contained a single element the ID of the span itself.
I'm sure the intention is to get the actual list of children.

It looks like this may have fixed #999, at least in some cases.


codefromthecrypt:
Thanks, Eirik!

PR #1016, opened 2016-03-08
Changes default cassandra load balancing policy to dc-aware round-robin

codefromthecrypt:
Before, the default cassandra load balancing policy was to consider all
datacenters local. This could lead to high latency when storing or
receiving spans. This changes the default policy to be datacenter aware.

The impact is that the first host's datacenter will be chosen as the
local one, and any hosts that aren't in that datacenter will be ignored.

As before, the local datacenter can be pinned via `CASSANDRA_LOCAL_DC`.

### DIFF
--- a/zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
+++ b/zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala
@@ -18,7 +18,7 @@ package com.twitter.zipkin.cassandra
 import java.net.InetSocketAddress
 
 import com.datastax.driver.core.{HostDistance, PoolingOptions, Cluster}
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
For those listening at home.. the following log from the datastax driver says which datacenter won

```
17:17:01.645 [main] INFO  c.d.d.c.p.DCAwareRoundRobinPolicy - Using data-center name 'datacenter1' for DCAwareRoundRobinPolicy (if this is incorrect, please provide the correct datacenter name with DCAwareRoundRobinPolicy constructor)
```

codefromthecrypt:
pinging cassandra peeps @michaelsembwever @kristofa @danchia @yurishkuro @prat0318 @eirslett

abesto:
Or even @openzipkin/cassandra

yurishkuro:
lgtm

prat0318:
lgtm. we didn't have any pinned local dc till now, so half of our inserts would be going to remote dc. this change should improve the query perf.

codefromthecrypt:
Thanks for the review, folks!

kristofa:
:+1:

PR #1004, opened 2016-02-26
Render the index/default page completely client-side

eirslett:
> The default page used to leverage a /modelview endpoint
> on the Scala server to generate a model to pass into its
> Mustache template. The logic from that /modelview endpoint
> is now ported to JavaScript, so that all the pre-rendering
> business logic happens on the client side.
> 
> Since this is a high-risk change, the /modelview endpoint
> still exists on the server. Every time we request a page,
> the UI will actually fetch a processed version from that
> endpoint as well, and compare the Scala version with
> the JavaScript version. The idea is that we can compare
> outputs of the two, and if there's a difference, we can
> assume that's a bug in the JavaScript port.
> The endpoint, and the UI logic to compare implementations,
> will soon be removed, after testing.

Please don't merge this PR yet - I really need people to test
this change manually!

It would be great if you can checkout this branch, start
zipkin-web and look at a few trace summaries. (Not traces,
but the list of trace summaries/search results on the front
page).
In the browser debugger, there's a console.log thing that
logs whether the JavaScript implementation gives the same
result as the Scala implementation, given the same input.

I would love to hear if anybody experiences a situation where
the two implementations produce different output - and in
that case, what the input was.

If you've tried running zipkin-web with this change, and you
found no bugs, be sure to +1 the PR. Whenever we're sure
that most bugs are cleared out, I'll update with another commit
that removes the automatic scala-vs-javascript-comparison
on every page load.

PR #1017, opened 2016-03-08
Feature/move business logic to client side

eirslett:
This is a clean-up of #1004, I also removed the comparison code that was used for manual testing.


eirslett:
It's a wrap! :beers:

zipkin-web/src/main/scala/com/twitter/zipkin/web/TraceSummary.scala
```diff
@@ -18,7 +18,6 @@ package com.twitter.zipkin.web
 
 import com.twitter.finagle.tracing.SpanId
 import com.twitter.zipkin.common.{Trace, Endpoint, Span, SpanTreeEntry}
-import com.twitter.zipkin.web.Util.getRootMostSpan
 
 case class SpanTimestamp(name: String, timestamp: Long, duration: Long) {
```
codefromthecrypt: TraceSummary and TraceSummaryTest can be deleted now

codefromthecrypt: actually I'll delete these things.

codefromthecrypt:
I tidied up deletes and rebased on master. tested locally with a cassandra backend and still works.

:shipit:

zipkin-web/src/main/scala/com/twitter/zipkin/web/TraceSummary.scala
eirslett: I was thinking (but it's a separate/later issue) that we should consider sending trace summaries over the wire instead of complete traces, in the default/index/search results page. And only send the complete trace data when you're looking at a single trace.
Anyways, should we go that direction, the code is still there in git.

codefromthecrypt: Agreed.. Fwiw if we accept gzip (in the javascript client) and the limit is
10, probably have some time to sort out what's best.

zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
```diff
@@ -53,7 +44,6 @@ trait ZipkinWebFactory { self: App =>
   )
 
   val webServerPort = flag("zipkin.web.port", new InetSocketAddress(8080), "Listening port for the zipkin web frontend")
-  val webRootUrl = flag("zipkin.web.rootUrl", "http://localhost:8080/", "Url where the service is located")
```
codefromthecrypt, line 56: removed this from docker as twitter server is very sensitive about being passed more properties than needed https://github.com/openzipkin/docker-zipkin/commit/d97ac3402e20bd4aca295fdead11b48d40416345

eirslett, line 56: Oh sorry, I didn't know that :-/ We should probably fix that (allow zipkin to ignore properties it hasn't heart of) so we don't deployment scripts...

PR #1018, opened 2016-03-08
Documentation migrated to zipkin.io (repo openzipkin.github.io)

abesto:
1,971 deletions


abesto:
Be advised, after this is merged, I'll also delete the `gh-pages` branch (it was built from content deleted in this PR)

codefromthecrypt:
:shipit: thanks for the help!

eirslett:
Looks like this might have broken the README.md file on Github...  I guess we could simply hotlink the images.

abesto:
Ouch. I should've noticed that, sorry. Could hot-link, but that'll break in the far future and be even harder to notice. I'll pull that image back.

PR #1022, opened 2016-03-10
Add back doc images used in the README

abesto:
As noted in https://github.com/openzipkin/zipkin/pull/1018#issuecomment-194567526, #1018 broke the README. This un-breaks it. We could hot-link the images from zipkin.io, but that'll break in the far future and be even harder to notice. I prefer having the images used in the README in this repo.


codefromthecrypt:
:+1:

PR #1023, opened 2016-03-10
Uses TArrayByteTransport to reduce copying reading Kafka messages

codefromthecrypt:
Kafka messages come in the form of byte arrays. Before, a single-span
message was decoded via a input stream wrapper as opposed to reading
the byte array directly.


codefromthecrypt:
@prat0318 I don't think this has anything to do with your memory leak, but anyway...

PR #1026, opened 2016-03-11
Updates screen shot with more a more realistic call graph

codefromthecrypt:
Thanks to @yurishkuro for the content


abesto:
:shipit:

PR #1027, opened 2016-03-11
Supports viewing raw traces via GET /api/v1/trace/:id?raw

codefromthecrypt:
When troubleshooting instrumentation, we sometimes need to hop to the
database to see what the collector stored (which is closer to what the
instrumentation sent).

In some cases, we are hunting down "clock skew" problems, and sometimes
that's a red herring. Some times we are looking to see if applications
are sending the same span multiple times. The reason we have to go to
the database is we can't quite tell if query-time adjustment logic is
helping or hurting.

This introduces a new query flag "raw" to the get trace endpoint. When
specified, no query time adjustments are made after data is returned
from the backend.

### DIFF
--- a/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
+++ b/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
@@ -104,7 +104,13 @@ class AnormSpanStore(val db: DB,
     }
   }
 
-  override def getTracesByIds(traceIds: Seq[Long]): Future[Seq[List[Span]]] = db.inNewThreadWithRecoverableRetry {
+  override def getTracesByIds(traceIds: Seq[Long]) = getSpansByTraceIds(traceIds).map(
+    _.map(CorrectForClockSkew)
+     .map(ApplyTimestampAndDuration)
+     .sortBy(_.head)(Ordering[Span].reverse) // sort descending by the first span
+  )
+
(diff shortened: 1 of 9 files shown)


codefromthecrypt:
cc @yurishkuro who requested this and a few others I think might find this useful (instrumentation-peeps): @marcingrzejszczak @dsyer @kristofa @jcarres-mdsol @prat0318

codefromthecrypt:
Incidentally, this came in handy troubleshooting a duration issue with @fedj today. By running the raw query, we were able to isolate input data that shows there's a duration bug!

yurishkuro:
LGTM

Thanks for putting this in!  It will be very useful during our instrumentation work.

Hint: the new trace image on zipkin.io is actually partially broken, if you pay attention - the root (local) span is not shown, so the first span does not encompass the whole trace.  I think with ^^^ change I can dig down easier why that happens (I'm pretty sure adjusters drop it for some reason).

eirslett:
Are you sure `?raw` is the best format? How about `?clock_skew=false`?

yurishkuro:
from my perspective, I wanted the "raw" as in "no adjustments", which includes not only clock skew, but all other adjustments that can potentially drop or alter the spans recorded in the storage.

codefromthecrypt:
> Are you sure ?raw is the best format? How about ?clock_skew=false?
> 
> yeah, we used to have clock_skew, but the thing is that there are several
> adjustments. For example, there's a current bug with choosing the correct
> duration while merging spans at query time.

Yuri convinced me pretty quickly that "raw" is best, basically don't
manipulate the data at query time when this flag is present, regardless of
whether this is merging, correcting skew, filling timestamps, etc.

PR #1029, opened 2016-03-14
Excludes javadoc and source from zipkin-ui (static asset) publishing

codefromthecrypt:


codefromthecrypt:
javadoc and source jars don't make sense in terms of static asset jars.. I noticed this when browsing snapshot repo http://oss.jfrog.org/artifactory/oss-snapshot-local/io/zipkin/zipkin-ui/1.36.1-SNAPSHOT/

cc @eirslett @abesto

codefromthecrypt:
merging to test (by looking at jfrog)

codefromthecrypt:
reverted by https://github.com/openzipkin/zipkin/commit/4f75efe852319c49bc34ecbba2b6ec38bb0a109b

Basically, this is an unnecessary thing required by maven central policy. Easiest thing is just to ignore the extra assets. http://central.sonatype.org/pages/requirements.html#supply-javadoc-and-sources

PR #1030, opened 2016-03-14
Reverts default Cassandra host policy to RoundRobin

codefromthecrypt:
When changing loadbalancer policy, we didn't consider a scenario brought
up by @michaelsembwever:

`DCAwareRoundRobinPolicy` will by default prefer the first contact node's
datacenter. If that node happens to be in a remote datacenter, latency
will be high, and the `LatencyAwarePolicy` cannot correct it (since
local nodes will be filtered out).

The previous logic is better because eventhough `RoundRobinPolicy` can
select a remote datacenter, it also selects local ones: the
`LatencyAwarePolicy` will eventually correct to choose only the local
ones.

This reverts commit 813f94d2fc2040c9296fb13075080c5d0c07cf19.

See https://github.com/openzipkin/zipkin-java/pull/92/files#r55933918


codefromthecrypt:
cc @danchia @yurishkuro

PR #1033, opened 2016-03-14
[zipkin-ui] Change nomenclature of tasks calling npm

abesto:
The current naming implies knowledge of the implementation of the npm tasks it's calling. I expect in the far future we'll replace karma with another test runner, and forget to change the name of the task, leading to confusion and inconsistency.

(Even so there's still some coupling between webpack and Gradle in the name of the "dist" dir, but if the npm build output dir changes, this will automatically break. If it doesn't need an update, this will be just fine)

### DIFF
--- a/zipkin-ui/build.gradle
+++ b/zipkin-ui/build.gradle
@@ -31,14 +31,14 @@ node {
     nodeModulesDir = file("${project.projectDir}")
 }
 
-task webpack(type: NpmTask) {
+task npmBuild(type: NpmTask) {
     args = ['run-script', 'build']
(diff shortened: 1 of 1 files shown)


abesto:
This assumes the "vision" is to stay with NPM for the duration. If instead we want to replace NPM and keep webpack + karma, the same argument says we need to NOT include `npm` in the name of the tasks, and this should not be merged.

@eirslett?

eirslett:
9f2fc588c27de0ebd2aae933a900121fdbcdd2d1

PR #1034, opened 2016-03-15
Load configuration from /config.json

eirslett:
Renames the /config.js endpoint. Returns json
instead of setting window.config. This is more
flexible, but also somewhat more involved, since
we have to load configuration asynchronously
from the client code.

Also adds source map generation for js and css
(easier debugging)

### DIFF
--- a/zipkin-ui/js/component_ui/environment.js
+++ b/zipkin-ui/js/component_ui/environment.js
@@ -1,8 +1,7 @@
 import flight from 'flightjs';
-import config from '../config';
 
 export default flight.component(function environmentUI() {
   this.after('initialize', function() {
-    this.$node.text(config('environment'));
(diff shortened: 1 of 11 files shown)


codefromthecrypt:
I was going to suggest this! Thanks for implementing before I had the
chance to.

codefromthecrypt:
Do we want to call this config.json or just config? I can see the case for the former, as it looks like a file and could be implemented by one. The latter looks more like other things like /healthcheck (assuming content-type is correct)

Thoughts?

eirslett:
The idea was that, like you said, people could host the ui from a web server like apache httpd, and simply configure it by creating their own config.json.

codefromthecrypt:
WFM

PR #1035, opened 2016-03-15
Put assets in /static/ directory inside the zipkin-ui jar

### LINKED ISSUE #1031: consider moving zipkin-ui assets under /static
currently, zipkin-ui's assets are included in the root directory. This has an interesting side-effect of making all contents of the zipkin-web jar fetchable by the static asset handler. For example, you can download classes and other properties in the jar via GET.

If we packaged zipkin-ui's assets under "/static", then only those assets would be served. Also, it has a side-effect of being automatically exported in spring-boot apps.

Alternatively, we could get fancier and make zipkin-ui a WebJar by moving the static assets under a path  "/META-INF/resources/webjars/zipkinui/1.37.0" (or whatever version) http://www.webjars.org/

eirslett:
This makes it easier to use from Spring boot.
Somebody else needs to take it from here; I can't figure out how to make zipkin-web work with the new directory layout...

Fixes #1031.

### DIFF
--- a/zipkin-ui/build.gradle
+++ b/zipkin-ui/build.gradle
@@ -5,7 +5,6 @@ artifacts.zipkinUpload jar
 sourceSets {
     main {
         resources {
-            srcDir 'static'
             srcDir 'dist'
         }
     }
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
Whoot!

PR #1038, opened 2016-03-16
Removes ostrich dependency from zipkin-query-service

### LINKED ISSUE #1036: Exclude ostrich from zipkin-query-service
Inside the zipkin-query-service shadowJar, we get a /static/ folder with assets from [ostrich](https://github.com/twitter/ostrich). These resources clash with zipkin-ui assets, that are also in /static/. (After #1035 is merged).
We need to find some way to exclude ostrich from our project - is anybody up for that task?

codefromthecrypt:
Ostrich was used to parse the '-f' argument in `zipkin-query-service`.
However, it also brought in a '/static/' directory of resources that
will clash with `zipkin-ui` static content. This parses `-f` manually,
and removes the dep.

Fixes #1036

### DIFF
--- a/zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
+++ b/zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
@@ -33,10 +33,6 @@ import com.twitter.zipkin.json.ZipkinJson
 import org.slf4j.LoggerFactory
 
 trait ZipkinWebFactory { self: App =>
-  private[this] val resourceDirs = Set(
-    "/"
-  )
-
   private[this] val typesMap = Map(
     "css" -> "text/css",
     "png" -> "image/png",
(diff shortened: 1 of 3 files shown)


zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
```diff
@@ -33,10 +33,6 @@ import com.twitter.zipkin.json.ZipkinJson
 import org.slf4j.LoggerFactory
 
 trait ZipkinWebFactory { self: App =>
-  private[this] val resourceDirs = Set(
```
codefromthecrypt, line 36: unused and was bothering me :)

codefromthecrypt:
cc @eirslett @abesto 

I tested this locally with both explicit paths and bundled ones.. the flag is compatible with before, but better worded and with a better error message.

eirslett:
Great, thanks a lot!! +1

PR #1039, opened 2016-03-16
zipkin-web: Serve from /static/ following #1035

abesto:
This change also revamps the way resources are resolved.

The old logic was "find the static resource, and if you can't, then serve index.html".

The new logic is "if the request is a file (through heuristic 'it has an extension'), serve it. otherwise serve index.html"

This makes the intention cleaner, and gives us 404 where a 404 is needed instead of the contents of index.html

### DIFF
--- a/zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
+++ b/zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
@@ -96,6 +96,13 @@ class Handlers {
     new Service[Request, Renderer] {
       private[this] var rendererCache = Map.empty[String, Future[Renderer]]
 
+      /**
+        * Resolves the resource to return for the path requested by the browser
+        * @param path The path requested by the browser. Always starts with `/`
+        * @return The URL of the file referenced by the path if it has an extension (ie. contains a dot), `static/index.html`
+        *         otherwise (we use the same HTML to serve all the pages, the page generation logic is in JS based on
+        *         `window.location`). Returns `None` if the requested file doesn't exist.
+        */
       private[this] def getResource(path: String): Option[URL] =
         Option(getClass.getResource(path match {
           case _ if path.split("/").last.contains(".") => s"/static$path"
--- a/zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
+++ b/zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
@@ -1,5 +1,7 @@
 package com.twitter.zipkin.web
 
+import java.net.URL
+
 import com.google.common.io.ByteStreams
 import com.twitter.finagle.http.{Request, Response}
 import com.twitter.finagle.stats.{Stat, StatsReceiver}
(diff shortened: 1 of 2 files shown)


abesto:
@adriancole @eirslett The Scala is slowly coming back to me, please check to see if I made a mess or not.

zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
```diff
@@ -77,8 +77,7 @@ trait ZipkinWebFactory { self: App =>
     import handlers._
 
     Seq(
-      ("/", handlePublic(Set("/"), typesMap)),
```
codefromthecrypt, line 80: thx Set("/") was confusing

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
```diff
@@ -90,31 +92,25 @@ class Handlers {
...
-        val resource = getClass.getResource(path)
-        if (resource == null || new File(resource.getPath).isDirectory) {
-          None
-        } else {
-          Option(resource.openStream())
-        }
-      }
+      private[this] def getResource(path: String): Option[URL] =
+        Option(getClass.getResource(path match {
+          case _ if path.split("/").last.contains(".") => s"/static$path"
```
codefromthecrypt, line 108: maybe add a comment what's going on here (ex why we are looking for a dot)

abesto:
Side-note: I kind of feel like changing `index.html` to refer to `/static/*.js` instead of `/*.js`. That would make the handlers considerably simpler, and just feels nice. How would that play together with Spring etc.?

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
codefromthecrypt, line 108: guessing `// if the path includes a file name (ex index.html), serve it relatively from static`

and the otherwise case is.. `// if there was no file name in the path, boot them to the home page`

zipkin-web/src/main/scala/com/twitter/zipkin/web/Handlers.scala
```diff
@@ -90,31 +92,25 @@ class Handlers {
...
         rendererCache.get(path) orElse {
           synchronized {
             rendererCache.get(path) orElse {
-              resourceDirs find(path.startsWith) flatMap { _ =>
-                getStream(path) map { input =>
-                  val typ = typesMap find { case (n, _) => path.endsWith(n) } map { _._2 } getOrElse("text/plain")
-                  val renderer = Future.value(StaticRenderer(input, typ))
-                  rendererCache += (path -> renderer)
-                  renderer
-                } orElse Option(Future(StaticRenderer(getClass.getResourceAsStream("/index.html"), "text/html")))
```
codefromthecrypt, line 117: cool. this was also weird, even if understandable

codefromthecrypt:
LGTM: only comment would be to add a comment from your PR near the code

codefromthecrypt:
I don't think it would play nice :) basically they would be nested in the
jar under static/static. In spring and other containers, "static/" or
"public/" basically acts as a content chroot.

eirslett:
We could serve them under "dist", so the folder structure would be /static/index.html, /static/dist/app.min.js and so on.

abesto:
SGTM. Would we want to set up the same folder structure inside the jar? I imagine that'd make routing easier on users (see https://github.com/openzipkin/zipkin-java/pull/98#issuecomment-197276575)

abesto:
Also. @adriancole @eirslett I propose we merge this to un-break master and re-think the folder structure if need be, in another PR. Go?

eirslett:
+1

PR #1042, opened 2016-03-16
Host zipkin ui from query-service

eirslett:
We're getting closer!


abesto:
There’s some special logic in `zipkin-web` to not trace static assets, may be worth porting here. See https://github.com/openzipkin/zipkin/blob/master/zipkin-web/src/main/scala/com/twitter/zipkin/web/FilteredHttpEntrypointTraceInitializer.scala and https://github.com/openzipkin/zipkin/blob/master/zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala#L116-L121. (Not sure it applies, or applies directly)

codefromthecrypt:
@abesto yeah that could be joined with FilteredHttpEntrypointTraceInitializer in zipkin-query

codefromthecrypt:
works as advertised!

codefromthecrypt:
FilteredHttpEntrypointTraceInitializer thing done.. merge on green!

PR #1045, opened 2016-03-16
DEPRECATED zipkin-{ui,web}: Reorganize webpack structure 

abesto:
Closed in favor of #1049 (same changeset initially, but on a shared branch)


zipkin-ui/webpack.config.js
```diff
@@ -27,9 +27,9 @@ module.exports = {
         }]
     },
     output: {
-        path: __dirname + '/dist/static',
+        path: __dirname + '/build/resources/main/static/assets',
```
eirslett: Is it possible to use /static here and assets/app.min.js in "filename"?

abesto: Yes, but everything else will still go into `static`, not `static/assets`:

```
zipkin-ui/build/resources/main/
└── static
    ├── 448c34a56d699c29117adc64c43affeb.woff2
    ├── 614fad616d014daf5367e068505cad35.png
    ├── 89889688147bd7575d6327160d64e760.svg
    ├── 8b55a822e72b8fd5e2ee069236f2d797.png
    ├── app.min.css
    ├── app.min.css.map
    ├── assets
    │   ├── app.min.js
    │   └── app.min.js.map
    ├── e18bbf611f2a2e43afc071aa2f4e1512.ttf
    ├── f4769f9bdb7466be65088239c12046d1.eot
    ├── fa2772327f55d8198301fdb8bcfc8158.woff
    └── index.html
```

eirslett: Ok, I see. Is it possible to set HtmlWebpackPlugin's filename in terms of __dirname maybe? It doesn't feel quite right to put a generated resource in the parent directory of where it's supposed to be :-/

Before we merge this, maybe we could open a PR on zipkin-query and zipkin-java to see what the corresponding diff would be there? I'm not so convinced it will be so much easier, but then again I haven't looked to much at it, so I might be wrong. I'm a bit -1 for this change, unless it gives us real benefits on the other side...

abesto: Sure thing. I'll be putting down Zipkin soon for the day; I'd say let's wait for more input on https://github.com/openzipkin/zipkin-java/pull/98#issuecomment-197361662, that'll probably affect this as well.

abesto:
@adriancole @eirslett I think this is ready for another round of reviews.

@dsyer FYI, this implements the "rename `static` to `zipkin-ui` in the jar" KISS solution mentioned in https://github.com/openzipkin/zipkin-java/pull/98#issuecomment-197453093.

eirslett:
> Generates separate index.html for /dependency/

This would break in other cases, e.g. the /trace/:traceId endpoint. And it wouldn't be very forwards-compatible once we add more routes.

codefromthecrypt:
on rename of "/static" to "/zipkin-ui" I think this was a nudge to try to use webjar format, which neither are. It probably will help, at the end of the day, and the amount of custom code in zipkin-java doesn't change all that much I think.

codefromthecrypt:
ps for things like this, it is kindof more convenient to raise on a openzipkin/zipkin so no need to fetch and pick commits to try it (plus I can contribute there, too)

abesto:
Moved to shared branch in #1049

abesto:
@eirslett Indeed, I missed that. Will revert that change.

@adriancole on webjar: I agree it was a nudge in that direction. I don't have a strong opinion on the webjar-vs-not debate, and don't know enough to form one. I'm deferring to @eirslett on this one, I was under the impression we had a consensus.

Issue #859, opened 2015-11-16
Hunt down licenses for dependencies


abesto:
The license verification introduced in https://github.com/openzipkin/zipkin/pull/852 highlights that some of the dependencies we use don't specify their license in their POM file. The license plugin can be explicitly told the license of each dependency via `downloadLicenses.licenses` (https://github.com/hierynomus/license-gradle-plugin#license-reporting)

eirslett:
Is this fixed?

abesto:
Negative. ATM there are 35 dependencies with no license, 11 of which are other subprojects of `openzipkin/zipkin`.

abesto:
Closed via #1048

PR #1048, opened 2016-03-16
Fix #858, #859: add license info for deps as needed; enforce verification

### LINKED ISSUE #858: Fix: verifyLicenses thinks subprojects don't have a license
The dependency license verification introduced in https://github.com/openzipkin/zipkin/pull/852 fails to recognize that the subprojects (like `zipkin-common`) have a license. Not surprising, since there's no POM file accompanying the jars. I see three options for attacking this:
- Generating a POM file and making sure the license plugin reads it
- Add metadata to the jars that the plugin reads
- Hard-code (or generate with gradle) the license for each subproject into the configuration for the plugin via `downloadLicenses.licenses` (https://github.com/hierynomus/license-gradle-plugin#license-reporting)

abesto:


eirslett:
:grinning: Go for it! :beer:

codefromthecrypt:
thx

PR #1049, opened 2016-03-17
zipkin-{ui,web}: Reorganize webpack structure

abesto:
(Some discussion originally happened on #1045)

**Moves the generated assets from dist to build/resources/main**, following Gradle conventions. Everything there gets automatically packed up into the jar we generate from zipkin-ui.

**Renames the dir `static` to `zipkin-ui`** to avoid conflicts with other packages. See https://github.com/openzipkin/zipkin/pull/1038 and https://github.com/openzipkin/zipkin-java/pull/98#issuecomment-197352561

**Simplifies `zipkin-web` handler registration**. The code originally pretended to support URL placeholders like `:id` in `/foo/:id`. That's not, in fact, the case (it just replaced `:id` with `/`). The new routes directly use the specification of the Finagle HttpMuxer.

**Refactors `Handlers.handlePublic`** to allow reusing its caching logic when serving `index.html`, without putting logic for two semantically different handlers into one piece of code.

**Generate `index.html`** using `html-webpack-plugin`. This lets us move things around without worrying about updating the contents of `index.html` manually.


eirslett:
I'm still not convinced we need an assets directory - finatra solves this in zipkin-query quite elegantly, so we could simply forward all HTTP traffic directly from zipkin-web to zipkin-query as-is. (See my comment in #1050)

abesto:
Agree that if we want to provide zero support for hosting without defaulting to `index.html`, we don't need an assets directory. I'd note that not having an assets subdirectory not only doesn't provide support, but actively makes it harder.

abesto:
As discussed in #1050, defaulting to `index.html` will be the only supported way of hosting, at the very least for the time being. This means we can remove the assets dir with no consequences.

abesto:
@eirslett methinks this is ready for final review. Will squash before merge.

eirslett:
The code looks good to me! (but I haven't tested it)

codefromthecrypt:
bin/query looks fine, but bin/web results in an empty screen. If you enter /index.html  you get the banner like so.. same issue with the shadowJar

screen shot 2016-03-17 at 10 52 00 pm

eirslett:
What shows up in the browser debugger? Do you get 404 on any resources? Any console error messages? What is the HTML that is served?

abesto:
Oh crap I forgot again. The longer URL rule wins. "/" is longer than "". This was the reason for moving both the index and the static asset handling code into the same handler, and decide which one to serve in there, instead of with routing rules.

Won't be able to write the code today, but wanted to give you a heads up. Sorry :D Also, no idea how I didn't notice this. Should probably just cherry-pick #1051 into this one...

codefromthecrypt:
works after rebase!

zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
```diff
@@ -38,10 +39,11 @@ trait ZipkinWebFactory { self: App =>
...
     .asInstanceOf[Logger].setLevel(Level.toLevel(logLevel))
 
- /**
-  * Initialize an HTTP proxy service, targeting the query host.
-  */
-  def newWebServer(): Service[Request, Response] = Http.newService(queryDest())
+  /**
+    * Initialize an HTTP proxy service, targeting the query host.
+    */
+  def proxyClient(): Service[Request, Response] = Http.client.configured(param.Label("zipkin-web"))
```
codefromthecrypt, line 45: this change prevented self-tracing from logging the proxied requests under the service 127.0.0.1:9411

codefromthecrypt, line 45: basically it was a small regression, as the annotations should be logged under the local service name (in this case zipkin-web)

abesto, line 45: Oops.

Do we care enough to figure out how to fix it?

abesto, line 45: Oh, I see you did that already. Thanks!

PR #1050, opened 2016-03-17
Document, improve Zipkin-UI routing rules

abesto:
The README is available at https://github.com/openzipkin/zipkin/blob/zipkin-ui-routing/zipkin-ui/README.md

Before merging:
- [x] Merge #1049, this PR is based off of that
- [x] Add reasoning on the packaging format choice to README. Maybe @eirslett could explain it best?
- [x] Verify `zipkin-query` serves `index.html`; I can't compile it locally ATM
## Hosting methods

Initially this PR described four ways of hosting Zipkin-UI assets:
1. Programatically read `index-routes.json` and set up explicit routes based on it. Example: [zipkin-web](https://github.com/openzipkin/zipkin/blob/5dec252e4c562b21bac5ac2f9d0b437d90988f79/zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala#L92-L95).
   **Pro**: serves `index.html` for exactly the right routes. **Con**: needs active server component. **Suggested** method when
   embedding in a server application.
2. Match all unknown paths, and instead of returning 404, route them back to `index.html`. Example: [zipkin-query](https://github.com/openzipkin/zipkin/blob/5dec252e4c562b21bac5ac2f9d0b437d90988f79/zipkin-query/src/main/scala/com/twitter/zipkin/query/ZipkinQueryController.scala).
   **Pro**: never need to touch the routing rules again. **Con**: no 404, may interfere with other route definitions.
   **Suggested** method for static webservers.
3. A variation on the above: only serve `index.html` for paths where the last `/`-separated
   segment doesn't have a `.` in it. **Pros**: never need to touch the routing rules again; has 404 for paths with an extension.
   **Con**: the heuristic is not very robust.
4. Manually hard-code all the routes that need to serve `index.html`. Example: [zipkin-java](https://github.com/openzipkin/zipkin-java/blob/d031021769432f81d60690356249fee56003c5d1/zipkin-server/src/main/java/zipkin/server/ZipkinUI.java)
   **Pro**: doesn't interfere with other route definitions, correct 404s. **Con**: needs updating when new UI endpoints are
   added to Zipkin-UI.

After discussion (see in-line comments), the conclusion is that we're
- Dropping 1, because let's embrace that we're a single-page app.
- Dropping 2, because it's an inferior version of 3 (option 3 is directly supported, for example, by Finatra)
- Dropping 4, because it's a temporary hack in `zipkin-java` that'll go away, let's not propagate it.


zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
```diff
@@ -70,10 +73,9 @@ trait ZipkinWebFactory { self: App =>
     stats: StatsReceiver = DefaultStatsReceiver.scope("zipkin-web")
   ): Service[Request, Response] = {
     val handlers = newHandlers
-    import handlers._
+    import handlers.{handleIndexHtml, handleConfig, handlePublic, handleRoute, handleTrace, collectStats, renderPage, catchExceptions, checkPath}
```
abesto: Needed because otherwise we get an "ambiguous symbol" error for `getClass`

codefromthecrypt:
verified everything works in this repo!

zipkin-ui/README.md
```diff
@@ -0,0 +1,34 @@
...
+UI should do. This is handled by a route api defined in the crossroads library.
+
+The full list of
+such routes is defined in [`resources/zipkin-ui/index-routes.json`](src/main/resources/zipkin-ui/index-routes.json).
+Routes ending with `/` should be used in prefix matching (`/foo/` should match `/foo/`, `/foo/bar`, and `/foo/bar/baz`).
+Routes not ending with `/` should be used in exact matching (`/foo` should match `/foo` and nothing else).
+
+Based on all these, there are several ways for serving zipkin-ui:
+
+ * Programatically read `index-routes.json` and set up explicit routes based on it. Example: [zipkin-web](https://github.com/openzipkin/zipkin/blob/5dec252e4c562b21bac5ac2f9d0b437d90988f79/zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala#L92-L95).
```
eirslett: I'd rather not maintain index-routes.json... I firmly believe that serving index.html as a fallback for every route that has no other route handler should be a basic feature of any web server; if it's not, then it should be that server's responsibility to bear the burden of finding a workaround.
(I'm sorry that I'm so stubborn)

zipkin-ui/README.md
```diff
@@ -0,0 +1,34 @@
...
+   **Pro**: serves `index.html` for exactly the right routes. **Con**: needs active server component. **Suggested** method when
+   embedding in a server application.
+ * Match all unknown paths, and instead of returning 404, route them back to `index.html`. Example: [zipkin-query](https://github.com/openzipkin/zipkin/blob/5dec252e4c562b21bac5ac2f9d0b437d90988f79/zipkin-query/src/main/scala/com/twitter/zipkin/query/ZipkinQueryController.scala).
+   **Pro**: never need to touch the routing rules again. **Con**: no 404, may interfere with other route definitions.
```
eirslett: ~~404~~ "route not found" will be handled by crossroads on the client side. If crossroads cannot find a matching route in window.location, we show an error message. (Technically it's not implemented in zipkin-ui today, but should be on the short-term todo list)

zipkin-ui/README.md
```diff
@@ -0,0 +1,34 @@
...
+   **Suggested** method for static webservers.
+ * A variation on the above: only serve `index.html` for paths where the last `/`-separated
+   segment doesn't have a `.` in it. **Pros**: never need to touch the routing rules again; has 404 for paths with an extension.
+   **Con**: the heuristic is not very robust.
+ * Manually hard-code all the routes that need to serve `index.html`. Example: [zipkin-java](https://github.com/openzipkin/zipkin-java/blob/d031021769432f81d60690356249fee56003c5d1/zipkin-server/src/main/java/zipkin/server/ZipkinUI.java)
+   **Pro**: doesn't interfere with other route definitions, correct 404s. **Con**: needs updating when new UI endpoints are
+   added to Zipkin-UI.
+
+## What? Why? Why not a .* package?
```
eirslett: What's a .\* package?

abesto: I meant: Why not create the package in an existing package format? Good feedback, that this is not clear.

zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
```diff
@@ -70,31 +73,36 @@ trait ZipkinWebFactory { self: App =>
...
-      val path = p.split("/").toList
-      val handlePath = path.takeWhile { t => !(t.startsWith(":") || t.startsWith("?:")) }
-      val suffix = if (p.endsWith("/") || p.contains(":")) "/" else ""
+    )
 
-      m.withHandler(handlePath.mkString("/") + suffix,
-        collectStats(handlePath.foldLeft(stats) { case (s, p) => s.scope(p) }) andThen
+    val indexHtmlRoutes = (new ObjectMapper() with ScalaObjectMapper)
+      .registerModule(DefaultScalaModule)
+      .readValue(getClass.getResource("/zipkin-ui/index-routes.json").openStream(), classOf[Seq[String]])
```
eirslett: Since we're sunsetting zipkin-web, I suggest that we hard-code routes for serving index.html in here?
We could do the same with static assets.
Finatra, which is used in zipkin-query, solves this quite elegantly.

Another alternative is to rewrite zipkin-web to use the http proxy from here: https://twitter.github.io/finagle/guide/Quickstart.html
Since assets are served from zipkin-query, then zipkin-web could forward _all_ HTTP traffic to zipkin-query, as-is, without any route logic.

zipkin-ui/README.md
abesto: Watch out, wall-of-text incoming ;)

Serving index.html is possible with any server; that's exactly what's documented in the second option. The question is more whether we want to do that. In the end, all users of zipkin-ui are free to do with the package whatever they want, what we need to decide is the use-cases we support / document.

To me "serve index.html always" feels icky; if we know something's not there, I have a slight preference to let the server already tell me so. Not doing so breaks strong conventions. I've already seen several times that it makes debugging harder (I look at the network tab of my browser, see 200s, but nothing works; I guess that one of the JS files is wrong, click through, and _then_ see that the contents of `app.min.js` are the contents of `index.html`, _and then_ from that infer that this is in fact a 404, and to verify that, I'll need to open the URL in a new browser tab)

Those are my arguments for "support this use-case".

Arguments I see so far against "support this use-case": I see that maintaining `index-routes.json` is extra work. Based on what I've seen so far, I expect that to mean adding an extra line every few months at most. I understand that defaulting to `index.html` is the way many modern single-page apps are built. Dropping this use-case also does away with the extra complexity in users that'd use `index-routes.json`, for the price of not allowing them to implement that extra complexity, even if they want to. (I'd want to, but today I'm playing maintainer of Zipkin as opposed to user, so that's mostly off-topic)

Did I miss any arguments on either side? If no: based on these, I have a weak preference to provide `index-routes.json`. I understand your evaluation is more firmly on the other side. I'd love to get more opinions.

PS. I'm totally OK getting rid of the relevant code if we decide to not support `index-routes.json`, that "wasted work" should not be a factor here.

zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
abesto: +1 for turning zipkin-web into a dumb proxy. You can consider this code a proof-of-concept to see how `index-routes.json` works in practice.

zipkin-ui/README.md
eirslett: > if we know something's not there, I have a slight preference to let the server already tell me so.

The problem is that we often don't know that on beforehand on the server side. E.g. if you hit /trace/nonexistenttrace, the server would say that route exists. Except we want to show an error message, "trace not found". But in order to do that, we need to do the API call from the client, to find out that the trace indeed doesn't exist.

The problem with debugging here, is that our configuration of crossroads in zipkin-ui isn't configured to handle scenarios where it gets a location and it has no route handler for that location. We should present a generic "not found" message in the UI. In the network tab, we would see that we got a 404 from the api request.

Some initial head-scratching is unfortunately the cost of migrating to pure client-side applications... but on the other hand, what we gain in the long term, is the complete separation of frontend and backend code, so there are no "leaks" between them.

zipkin-ui/README.md
eirslett: "Why is it packaged as a .jar and not a .zip?"
Which could be answered by something like "since many zipkin servers are java-based, it's convenient to distribute the ui as a jar, which can be imported by the Gradle build tool. A .jar file is really only a .zip file, and can be treated as such. It can be opened by any program that can extract zip files."

zipkin-ui/README.md
abesto: All good points. After these, and more discussion with @adriancole, I'm convinced we'll be better off dropping `index-routes.json`.

I'll also simplify the README to describe only one hosting method (default to index.html) to simplify things, and keep the options in the description of this issue for history.

zipkin-ui/README.md
abesto: I'll just copy-paste this into the README, if you don't mind :)

zipkin-web/src/main/scala/com/twitter/zipkin/web/Main.scala
abesto: (Will do the proxy in a separate PR)

codefromthecrypt:
closed via e01f942f7b3df548f65b272582cbf5470580bb14

nice job

PR #1052, opened 2016-03-17
Allows receiver-kafka to consume JSON encoded spans

### LINKED ISSUE #1037: Feature Request: Allow receiver-kafka to consume JSON encoded spans
It seems that JSON support has been added to the HTTP API, would be nice to publish JSON encoded spans to Kafka and have the receiver consume that (would also simplify coding on the client side and scaling zipkin).

codefromthecrypt:
By allowing JSON encoded span lists as Kafka message values, we enable
users to send the same payload as they can over the http transport. It
also allows a very easy debug story, such as sending test spans via CLI.

This implementation peeks at the first byte in the Kafka message, and
attempts to decode json, if it is a '[', or decimal 91. When using
TBinaryProtocol (thrift) encoding, we know that the first byte will be
the Type, which is decimal <=16, so there's no risk of clash.

Fixes #1037

### DIFF
--- a/zipkin-receiver-scribe/src/main/scala/com/twitter/zipkin/receiver/scribe/ScribeSpanReceiver.scala
+++ b/zipkin-receiver-scribe/src/main/scala/com/twitter/zipkin/receiver/scribe/ScribeSpanReceiver.scala
@@ -22,6 +22,8 @@ import com.twitter.logging.Logger
 import com.twitter.scrooge.BinaryThriftStructSerializer
 import com.twitter.util.{Base64StringEncoder, Future, NonFatal, Return, Throw, Time}
 import com.twitter.zipkin.collector.{QueueFullException, SpanReceiver}
+import com.twitter.zipkin.common.Span
+import com.twitter.zipkin.conversions.thrift
 import com.twitter.zipkin.thriftscala.{LogEntry, ResultCode, Scribe, Span => ThriftSpan}
 import java.net.InetSocketAddress
 import java.util.concurrent.CancellationException
@@ -83,8 +85,8 @@ class ScribeReceiver(
     (cat, messagesStats.scope("perCategory").counter(cat))
   }.toMap
 
-  private[this] def entryToSpan(entry: LogEntry): Option[ThriftSpan] = try {
-    Some(deserializer.fromString(entry.message))
+  private[this] def entryToSpan(entry: LogEntry): Option[Span] = try {
+    Some(thrift.thriftSpanToSpan(deserializer.fromString(entry.message)).toSpan)
   } catch {
     case e: Exception => {
       // scribe doesn't have any ResultCode.ERROR or similar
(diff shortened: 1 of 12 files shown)


codefromthecrypt:
cc @gena01 @eirslett

zipkin-receiver-scribe/src/main/scala/com/twitter/zipkin/receiver/scribe/ScribeSpanReceiver.scala
```diff
@@ -83,8 +85,8 @@ class ScribeReceiver(
     (cat, messagesStats.scope("perCategory").counter(cat))
   }.toMap
 
-  private[this] def entryToSpan(entry: LogEntry): Option[ThriftSpan] = try {
-    Some(deserializer.fromString(entry.message))
+  private[this] def entryToSpan(entry: LogEntry): Option[Span] = try {
+    Some(thrift.thriftSpanToSpan(deserializer.fromString(entry.message)).toSpan)
```
eirslett, line 89: thrift span to span to span? :-/

codefromthecrypt, line 89: this is an unfortunate naming train wreck. thriftSpanToSpan returns a
wrapper.

eirslett:
Great work! The code looks good to me! (But I haven't tested it)
Did you try spinning it up manually, with kafka and all?

codefromthecrypt:
nope.. didn't manually test kafka, but I'm interested enough to do that.
Will paste some goodies tomorrow.

codefromthecrypt:
tested with the following things..

In one tab, start up kafka and post a span list to it

``` bash
$ docker run -d -p 2181:2181 -p 9092:9092 --env ADVERTISED_HOST=192.168.99.100  --env AUTO_CREATE_TOPICS=true spotify/kafka
$ kafka-console-producer.sh --broker-list 192.168.99.100:9092 --topic zipkin
[{"traceId":"1","name":"bang","id":"2","timestamp":1234,"binaryAnnotations":[{"key":"lc","value":"bamm-bamm","endpoint":{"serviceName":"flintstones","ipv4":"127.0.0.1"}}]}]
```

In two tabs, run the query and the collector:

``` bash
$ bin/query
# in a separate tab, as query cannot consume kafka, yet...
$ KAFKA_ZOOKEEPER=192.168.99.100:2181 bin/collector
```

Finally, curl and see the span arrived!

``` bash
$ curl -s localhost:9411/api/v1/trace/1
[{"traceId":"0000000000000001","name":"bang","id":"0000000000000002","timestamp":1234,"annotations":[],"binaryAnnotations":[{"key":"lc","value":"bamm-bamm","endpoint":{"serviceName":"flintstones","ipv4":"127.0.0.1"}}]}]
```

prat0318:
thats so cool! will try it out some time.

PR #1057, opened 2016-03-19
Filters undefined duration when calculating totalServiceTime

### LINKED ISSUE #1056: zipkin-web 1.37 and zipkin-query 1.38 (Web UI) returns errors in web console
I am unable to use the latest 1.37 docker container for zipkin-web or just tried zipkin-query 1.38 with the Web UI. When I select a service from the dropdown list I am seeing nothing loading and errors in Web Console in both Firefox and Chrome.

Firefox reports: "too much recursion"
Chrome reports: 

``` text
lodash.js:13824Uncaught RangeError: Maximum call stack size exceeded
object.(anonymous function) @ lodash.js:13824
(anonymous function) @ traceSummary.js:15
6totalServiceTime @ traceSummary.js:161
(anonymous function) @ traceSummary.js:161
totalServiceTime @ traceSummary.js:161
(anonymous function) @ traceSummary.js:161
totalServiceTime @ traceSummary.js:161
(anonymous function) @ traceSummary.js:161
totalServiceTime @ traceSummary.js:161
(anonymous function) @ traceSummary.js:161
```

eirslett:
The zipkin api is supposed to return Span.timestamp and Span.duration.
There's a test `getSpansByTraceIds_doesntPerformQueryTimeAdjustment`,
which was added recently, which should insure that this is set.

Until all span store tests update to running this test, notably
zipkin-java, we'll need to guard as opposed to recursing indefinitely.

Fixes #1056

### DIFF
--- a/zipkin-ui/js/component_ui/traceSummary.js
+++ b/zipkin-ui/js/component_ui/traceSummary.js
@@ -147,12 +147,15 @@ export function traceSummary(spans = []) {
   }
 }
 
-function totalServiceTime(stamps, acc = 0) {
-  if (stamps.length === 0) {
+export function totalServiceTime(stamps, acc = 0) {
+  // This is a recursive function that performs arithmetic on duration
+  // If duration is undefined, it will infinitely recurse. Filter out that case
+  const filtered = stamps.filter((s) => s.duration);
+  if (filtered.length === 0) {
     return acc;
   } else {
-    const ts = _(stamps).minBy((s) => s.timestamp);
-    const [current, next] = _(stamps)
+    const ts = _(filtered).minBy((s) => s.timestamp);
+    const [current, next] = _(filtered)
         .partition((t) =>
           t.timestamp >= ts.timestamp
           && t.timestamp + t.duration <= ts.timestamp + ts.duration)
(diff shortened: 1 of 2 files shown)


zipkin-ui/js/component_ui/traceSummary.js
```diff
@@ -117,7 +117,9 @@ export function getServiceName(span) {
 }
 
 function getSpanTimestamps(spans) {
-  return _(spans).flatMap((span) => getServiceNames(span).map((serviceName) => ({
+  return _(spans)
+  .filter(span => span.timestamp != null && span.duration != null)
```
codefromthecrypt: the only scenario I can think of for this is that a span with only binary-annotations was returned by a query. I'll update this commit to add the comment.

codefromthecrypt:
very subtle, but the issue isn't about null, rather undefined. A span that has timestamp set, but duration undefined causes this. Will adjust the PR

codefromthecrypt:
OK rewrote implementation after finding a span that could cause this. When zipkin-java updates to and passes latest span store tests, this won't likely be possible. At any rate, the problem is dangerous, so we should leave the check in.

codefromthecrypt:
releasing a subminor as this bug hits very hard when present

PR #1062, opened 2016-03-23
Customizes fileOrIndex logic to include cache max-time

### LINKED ISSUE #1058: Feature Request: Update response headers in zipkin-query (for zipkin-ui assets)
Please add/update response headers to include "Cache-Control" and/or "Expires" headers when serving zipkin-ui? static files. I see them being reloaded on every page/click.

Also seems like /8b55a822e72b8fd5e2ee069236f2d797.png which might be a favicon returns the following response header

``` text
Content-Type:"image/png;charset=utf-8"
```

This is an image, so no need for charset?

codefromthecrypt:
This sets a hard-coded cache policy for the following resources:
- 1 minute for index.html
- 10 minutes for /config.json
- 365 days for hashed resources

Fixes #1058


eirslett:
LGTM!

abesto:
Nice!

PR #1076, opened 2016-03-30
Fix #1074: Don't distribute non-shadow jar for *-service

### LINKED ISSUE #1074: Don't distribute non-shadowed jars of collector-service and query-service
https://jcenter.bintray.com/io/zipkin/zipkin-query-service/1.39.0/

When people download the wrong jar, the UI assets won't be bundled (neither will any of the jar dependencies)
I think @mansu had this problem, getting the wrong jar? (Or so it seems from the gitter conversation)

If there's no good reason why we're publishing the not-shadowed jars, maybe we could just as well not distribute them in the first place? @abesto is it possible to configure Gradle so it doesn't build the jars, only the shadowed jars? (Or, at least doesn't upload the jars)

abesto:
Not totally sure this will work, but the only (easy) way to check is to ask Travis to run it.

### DIFF
--- a/zipkin-collector-service/build.gradle
+++ b/zipkin-collector-service/build.gradle
@@ -14,6 +14,7 @@ jar.manifest.attributes 'Main-Class': mainClassName
 
 tasks.build.dependsOn(shadowJar)
 artifacts.zipkinUpload shadowJar
+configurations.zipkinUpload.artifacts.removeAll { it.archiveTask.is jar }
 
 shadowJar {
     exclude 'META-INF/LICENSE' // jackson's META-INF/LICENSE conflicts with the directory META-INF/license
--- a/zipkin-query-service/build.gradle
+++ b/zipkin-query-service/build.gradle
@@ -14,6 +14,7 @@ jar.manifest.attributes 'Main-Class': mainClassName
 
 tasks.build.dependsOn(shadowJar)
 artifacts.zipkinUpload shadowJar
+configurations.zipkinUpload.artifacts.removeAll { it.archiveTask.is jar }
 
 shadowJar {
     exclude 'META-INF/LICENSE' // jackson's META-INF/LICENSE conflicts with the directory META-INF/license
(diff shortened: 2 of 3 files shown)


codefromthecrypt:
applies to zipkin-web, too. ps once we merge we need to adjust the
dockerfiles

https://jcenter.bintray.com/io/zipkin/zipkin-web/1.39.0/

abesto:
Added `zipkin-web`

Why do we need to adjust the Dockerfiles? They use the fat jar, non? https://github.com/openzipkin/docker-zipkin/blob/master/collector/Dockerfile#L4

codefromthecrypt:
nevermind me.. as long as the fat jar artifact name doesn't change, which
seems the case

codefromthecrypt:
oops.. looks like release needs an update?

```
Caused by: org.gradle.api.GradleException: Could not sync 'openzipkin/maven/zipkin/1.39.1' to Maven Central: HTTP/1.1 400 Bad Request [messages:[Missing: no main jar artifact found in folder '/io/zipkin/zipkin-web/1.39.1', Missing: no main jar artifact found in folder '/io/zipkin/zipkin-collector-service/1.39.1', Missing: no main jar artifact found in folder '/io/zipkin/zipkin-query-service/1.39.1', Dropping existing partial staging repository.], status:Validation Failed]
```

https://travis-ci.org/openzipkin/zipkin/builds/119740137

eirslett:
Looks like nexus has a strict requirement that a "main jar" must be present. This is a hack, but we could do something like echo "This is the wrong jar, use the -all jar instead" > README.txt && tar -gzxfxxz README.txt > zipkin-query-service.jar, and then upload that.
Or we could rename the -all jar to just the "main" jar name, or we could duplicate it (publish the same jar under two different names) for backwards compatibility.
Or we could just stop syncing zipkin-query-service to maven central, and only publish it to jcenter, if that's a possibility?

codefromthecrypt:
I think renaming the jar to be the primary artifact would be a great
option, as polishing the docker files is pretty easy. I don't know how easy
this is in gradle.

eirslett:
I guess this should work:
https://github.com/johnrengelman/shadow#publishing-the-shadow-jar-as-a-standalone-artifact
(without the -all extension)

codefromthecrypt:
reverted until we sort out a solution that works

PR #1082, opened 2016-04-02
Fixes crash when storage isn't available at bootstrap

### LINKED ISSUE #1007: zipkin-query crashes when span store is accessed before cassandra is available
This came up in https://github.com/openzipkin/docker-zipkin/issues/74, but it is easy to reproduce by running `bin/query cassandra` when cassandra isn't running.

```
15:41:49.978 [main] INFO  c.t.z.builder.QueryServiceBuilder - http server started on port: 9411
15:41:50.044 [main] ERROR  - Flag zipkin.store.cassandra.keyspace read before parse.
15:41:50.073 [main] ERROR  - Flag zipkin.store.cassandra.maxConnections read before parse.
15:41:50.175 [main] INFO  com.datastax.driver.core.NettyUtil - Did not find Netty's native epoll transport in the classpath, defaulting to NIO.
Exception in thread "main" com.datastax.driver.core.exceptions.NoHostAvailableException: All host(s) tried for query failed (tried: localhost/127.0.0.1:9042 (com.datastax.driver.core.TransportException: [localhost/127.0.0.1:9042] Cannot connect))
```

codefromthecrypt:
Self-tracing during Finatra bootstrap when cassandra was down caused the
process to crash. This makes it a log warning instead.

``` bash
16:29:18.435 [main] WARN  c.t.z.builder.QueryServiceBuilder - Unable to trace bootstrap due to NoHostAvailableException(All host(s) tried for query failed (tried: localhost/127.0.0.1:9042 (com.datastax.driver.core.TransportException: [localhost/127.0.0.1:9042] Cannot connect)))
```

Note: api calls will fail w/ HTTP 500 until storage is available.

Fixes #1007
Fixes https://github.com/openzipkin/docker-zipkin/issues/74

### DIFF
--- a/zipkin-query-service/src/main/scala/com/twitter/zipkin/builder/QueryServiceBuilder.scala
+++ b/zipkin-query-service/src/main/scala/com/twitter/zipkin/builder/QueryServiceBuilder.scala
@@ -44,6 +44,11 @@ case class QueryServiceBuilder(override val defaultFinatraHttpPort: String = "0.
 
   override def postWarmup() {
     super.postWarmup()
-    BootstrapTrace.complete()
+    try {
+      BootstrapTrace.complete()
+    } catch {
+      case e => log.warning("Unable to trace bootstrap due to %s(%s)"
+        .format(e.getClass.getSimpleName, if (e.getMessage == null) "" else e.getMessage))
+    }
   }
 }


codefromthecrypt:
PS the reason this works is cassandra's repository is already lazy wired. the only thing dereferencing it at bootstrap was self-tracing.

https://github.com/openzipkin/zipkin/blob/master/zipkin-cassandra/src/main/scala/com/twitter/zipkin/cassandra/CassandraSpanStoreFactory.scala#L47

PR #1098, opened 2016-04-25
Fixes bug where adaptive rate was basing decisions on local store rate

codefromthecrypt:
The adaptive sampler has a thread that sets a new rate when it sees
outliers compared against a target (target is the amount of requests
the storage layer is capable of).

This logic had a bug where it compared outliers against the local
storage rate of the leader, not the target rate.

In normal case, if there were 10 balanced nodes, with a target rate of
10k spans/second, each node could write 1k spans/second. Since the input
to span collectors is a percentage, not a budget, the adaptive sampler
needs to re-calculate the sampling percentage based on how much
collectors are writing. Ex, if they are writing more, the percentage
would need to be lowered to drop enough spans to meet a target rate. If
they are writing less, the percentage could be raised to not
unnecessarily drop spans.

The outlier check compares the aggregate rate of all collectors against
a target: its `Seq[Int]` input is a sequence of `memberVals.sum`

The algorith is.. if a count of aggregate rate measurements went up or
down beyond a threshold (15%), a new rate could be calculated.

For example, if the target was 10k, a count of inputs <8.5k or >11.5k
would lead to a new rate.

If you substitute a local node here, you'll notice that this would fire
much more often. For example, if the target was 10k, a specific member
would be writing at 1k. If so, inputs <850 or >1150 would fire a
calculation. <850 is impossible as the local member is writing above
that. >1150 would happen any time there's more than 1 node. In other
words, it would fire always.

While non-critical (firing more is better than firing less), this small
bug caused busy work and made the outlier check almost useless.

### DIFF
--- a/zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampleRate.scala
+++ b/zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampleRate.scala
@@ -129,10 +129,10 @@ class AdaptiveSampleRate(
 
   val calculator =
     { v: Int => Some(buffer.pushAndSnap(v)) } andThen
-      new StoreRateCheck[Seq[Int]](storeRate, stats.scope("storeRateCheck"), log) andThen
+      new TargetStoreRateCheck[Seq[Int]](targetStoreRate, stats.scope("targetStoreRateCheck"), log) andThen
       new SufficientDataCheck[Int](sufficientWindowSize / updateFreq, stats.scope("sufficientDataCheck"), log) andThen
       new ValidDataCheck[Int](_ > 0, stats.scope("validDataCheck"), log) andThen
-      new OutlierCheck(storeRate, outlierThreshold / updateFreq, stats = stats.scope("outlierCheck"), log = log) andThen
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
cc @mosesn

zipkin-sampler/src/main/scala/com/twitter/zipkin/sampler/AdaptiveSampleRate.scala
```diff
@@ -129,10 +129,10 @@ class AdaptiveSampleRate(
 
   val calculator =
     { v: Int => Some(buffer.pushAndSnap(v)) } andThen
-      new StoreRateCheck[Seq[Int]](storeRate, stats.scope("storeRateCheck"), log) andThen
+      new TargetStoreRateCheck[Seq[Int]](targetStoreRate, stats.scope("targetStoreRateCheck"), log) andThen
```
codefromthecrypt, line 132: the smoking gun was local references ending up in the global storage rate calculator. It makes no sense to check if the local store rate is set, then later blindly access targetStoreRate. In chasing this through, I found the more interesting bug.

codefromthecrypt:
cc @jtrobec

mosesn:
Good catch!  LGTM

PR #1102, opened 2016-04-29
Make serviceName an Option[String] in QueryRequest

esbie:
issue: zipkin query currently has no way to request traces across services, making it difficult to show any sort of inter-service overview ui.

proposed solution: change serviceName from being required to being optional.

I'm new to CQL, so feedback on Repository.java changes would be appreciated.

### DIFF
--- a/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
+++ b/zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
@@ -350,31 +350,19 @@ class AnormSpanStore(val db: DB,
     }
   }
 
-  override def getSpanNames(_service: Option[String]): Future[Seq[String]] = db.inNewThreadWithRecoverableRetry {
-    val service = _service.map(_.toLowerCase) // service names are always lowercase!
+  override def getSpanNames(_service: String): Future[Seq[String]] = db.inNewThreadWithRecoverableRetry {
+    val service = _service.toLowerCase // service names are always lowercase!
     implicit val (conn, borrowTime) = borrowConn()
     try {
-      service match {
-        case Some(s) => SQL(
-          """SELECT DISTINCT name
-            |FROM zipkin_spans t1
-            |JOIN zipkin_annotations t2 ON (t1.trace_id = t2.trace_id and t1.id = t2.span_id)
-            |WHERE t2.endpoint_service_name = {service} AND name <> ''
-            |GROUP BY name
-            |ORDER BY name
-          """.stripMargin)
-          .on("service" -> s)
-          .as(str("name") *)
-        case None => SQL(
-          """SELECT DISTINCT name
-            |FROM zipkin_spans t1
-            |JOIN zipkin_annotations t2 ON (t1.trace_id = t2.trace_id and t1.id = t2.span_id)
-            |WHERE name <> ''
-            |GROUP BY name
-            |ORDER BY name
-          """.stripMargin)
-          .as(str("name") *)
-      }
+      SQL(
+        """SELECT DISTINCT name
+          |FROM zipkin_spans t1
+          |JOIN zipkin_annotations t2 ON (t1.trace_id = t2.trace_id and t1.id = t2.span_id)
+          |WHERE name <> ''
+          |GROUP BY name
+          |ORDER BY name
+        """.stripMargin)
+        .as(str("name") *)
 
     } finally {
       returnConn(conn, borrowTime, "getSpanNames")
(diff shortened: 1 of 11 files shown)


codefromthecrypt:
Hi, Sarah.

Thanks for proposing this. Before we start, the scala server is on the way out and this is a fairly sweeping change. To review and merge this would take take effort away from finishing the server transition, and still require us to do it again there. Any chance we can implement it there in zipkin-java first place? I'd offer to help.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
```diff
@@ -321,21 +350,31 @@ class AnormSpanStore(val db: DB,
     }
   }
 
-  override def getSpanNames(_service: String): Future[Seq[String]] = db.inNewThreadWithRecoverableRetry {
-    val service = _service.toLowerCase // service names are always lowercase!
+  override def getSpanNames(_service: Option[String]): Future[Seq[String]] = db.inNewThreadWithRecoverableRetry {
```
codefromthecrypt: fwiw I don't think this needs to be an Option unless you need to get a list of all span names for some reason (ex the http api wouldn't call this)

esbie:
Hi @adriancole! Thanks for looking at the PR :)

I know it can be difficult to complete a migration when folks are still adding features to legacy code.  Since I am more familiar with Scala than Java and I already have a system that uses this scala version, I figured this was a good place for me to add & test the feature.

I'd be happy to work on this feature in the zipkin-java version! Is that really a blocker for addressing this in scala?

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
esbie: I don't have a personal use case for span names across services, so I don't mind backing out of this change

codefromthecrypt:
> I'd be happy to work on this feature in the zipkin-java version! Is that
> really a blocker for addressing this in scala?
> 
> Hi, sarah, it is mostly a matter of time and attention, particularly as the
> impact of the slice query thing is hard to grok sometimes. Also anorm is a
> royal pain and usually any time it is touched == an hour lost on some weird
> thing.

As long as you don't mind doing revisions until tests on all that backends,
and sharing the pain during that, I don't mind supporting by looking into
it and giving advice.

zipkin-anormdb/src/main/scala/com/twitter/zipkin/storage/anormdb/AnormSpanStore.scala
codefromthecrypt: > I don't have a personal use case for span names across services, so I
> don't mind backing out of this change
> 
> cool. I'd suggest backing it out to lower the scope of things to do :)

codefromthecrypt:
here's a hint. cassandra tests run when it is already running on localhost. If you start it similar to how travis runs (or start it however you like), you can get tests to fail locally instead of only in travis 

https://github.com/openzipkin/zipkin/blob/master/.travis.yml#L12

codefromthecrypt:
added this which might help us collaborate. Ex I can steal ideas from you, and in the case I figure something out, pitch them back https://github.com/openzipkin/zipkin-java/issues/182

codefromthecrypt:
My advice would be to add a new test case to `SpanStoreSpec` that can smoke out issues around sliceQuery and priority. By running this directly, you might find issues faster.

Here's an idea:

make 2 timestamps
make 10 endpoints (with different serviceNames)
store 20 single-span traces that vary on above. Ex. 10 traces have one timestamp and another 10 have the other

query the span store with serviceName=None and limit 10. You should see one of each serviceName, and only the most recent timestamp.

query the span store with serviceName=None, limit 10, and endTs before the more recent timestamp, but after the less recent one. You should see one of each serviceName, and only the less recent timestamp.

esbie:
Ok I added another test and fixed the currently failing tests. My cassandra queries still seem pretty bad. For getTraceIdsByAnnotation I don't think it's even possible to do in a single query since the binary annotations include serviceName

zipkin-cassandra/src/main/scala/com/twitter/zipkin/storage/cassandra/CassandraSpanStore.scala
```diff
@@ -292,27 +298,44 @@ abstract class CassandraSpanStore(
...
+        repository
+          .getTraceIdsByAnnotation(annotationKey(name, annotation, value), endTs * 1000, lookback * 1000, limit))
+        .map { traceIds =>
+          traceIds.asScala
+            .map { case (traceId, ts) => IndexedTraceId(traceId, timestamp = ts) }
+            .toSeq
+        }
+      case None => FutureUtil.toFuture(repository.getServiceNames).flatMap { names =>
+        val futures: Seq[Future[util.Map[java.lang.Long, java.lang.Long]]] = names.asScala.toSeq.map { name =>
+          FutureUtil.toFuture(repository
```
codefromthecrypt: this will likely cause some havoc, as there's sometimes thousands of service names. Maybe throw UnsupportedException for now? Then we can open up another issue to add an index for annotation w/o service name

esbie: OK will do

PR #1107, opened 2016-04-30
Updates to gradle 2.13

codefromthecrypt:


yurishkuro:
lgtm

any particular new feature you are after?

codefromthecrypt:
Sorry I should have mentioned. Mainly this is maintenance, ie not falling
too far behind. Also there's supposedly a performance improvement. I'll
check logs to see if that helped at all.

codefromthecrypt:
performance has been erradic in general, so this might not change things on that note.

codefromthecrypt:
reverting this. Even with latest idea, the project no longer imports properly. will rev-lock for now.

Issue #952, opened 2016-02-10
Should we update to DataStax Cassandra Driver 3.0?


codefromthecrypt:
cassandra-driver-core 3.0.0 uploaded a couple weeks ago. Should we update to it? If so, are there any opportunities therein?

Notes:
Since we use CQL, using driver 3.0 shouldn't affect [supported versions](http://datastax.github.io/java-driver/manual/native_protocol/). For example, those using cassandra 2.1 should still operate.

codefromthecrypt:
cc c\* folks @drax68 @michaelsembwever @yurishkuro @prat0318 @danchia

danchia:
https://docs.datastax.com/en/developer/java-driver/3.0/java-driver/reference/customCodecs.html

Custom codecs seems like it would solve the ugly hack we had to do for the us timestamps?

yurishkuro:
I don't have a strong opinion about the new driver, if anything I prefer not to upgrade because it might be incompatible with our Cassandra version (I started getting non-fatal exceptions in the Spark job after the last upgrade). But don't let me stop you if there are perf improvements - we are not exposed to perf on the writing side since our collectors are in Go.

prat0318:
We currently run Cassandra 2.1.4 and i think driver 3.0 supports 2.2+ versions. But yeah, if the driver is more performant we can look into upgrading our cluster to 2.2+.

codefromthecrypt:
Prateek, good point. so the current driver supports up to 2.2 (seems to be
the crossover version). Is there any reason (beyond effort) why you are
running 2.1? Ex. is there a problem in 2.2 folks should be aware of?

prat0318:
Just that it was working fine. But if we have plans to upgrade the driver, i can work with ops to upgrade the cluster.

codefromthecrypt:
Yeah I suspect it is a matter of when, not if to upgrade the driver. I
don't want to black-hole you.

prat0318:
ok. i will prioritise the upgrade from my end and will keep posted here for the updates.

danchia:
Actually, driver 3.0 supports all versions of Cassandra with CQL.

http://datastax.github.io/java-driver/manual/native_protocol/

prat0318:
[...]

PR #1108, opened 2016-04-30
Updates to Datastax Cassandra Driver 3.0

codefromthecrypt:
By updating to Datastax Driver 3, we still support Cassandra 2.2+, but
also 3.0.

Tested against Cassandra 2.2, 3.0 and 3.5

See #952

### DIFF
--- a/zipkin-cassandra-core/build.gradle
+++ b/zipkin-cassandra-core/build.gradle
@@ -5,11 +5,5 @@ repositories {
 }
 
 dependencies {
-    compile(group: 'com.datastax.cassandra', name: 'cassandra-driver-core', version: '2.1.10'){
-        exclude group: 'com.github.jnr', module: 'jnr-posix'
-    }
-    // above is packaged with 3.0.27, which is licensed under CPL; 3.0.29 is licensed under EPL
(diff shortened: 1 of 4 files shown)


zipkin-cassandra-core/build.gradle
```diff
@@ -5,11 +5,5 @@ repositories {
 }
 
 dependencies {
-    compile(group: 'com.datastax.cassandra', name: 'cassandra-driver-core', version: '2.1.10'){
-        exclude group: 'com.github.jnr', module: 'jnr-posix'
-    }
-    // above is packaged with 3.0.27, which is licensed under CPL; 3.0.29 is licensed under EPL
-    compile group: 'com.github.jnr', name: 'jnr-posix', version: '3.0.29'
-    testCompile group: 'com.datastax.cassandra', name: 'cassandra-driver-core', classifier: 'tests', version: '2.1.10'
```
codefromthecrypt, line 13: test deps were never used (as there were no tests)

codefromthecrypt:
here's the related change in zipkin-java https://github.com/openzipkin/zipkin-java/pull/185

Note: I've left travis alone, which makes sure we still work with old cassandra

PR #1109, opened 2016-04-30
Adds kafka property to control fetch.message.max.bytes

### LINKED ISSUE #1101: Make fetch.message.max.bytes configurable
Recently, we migrate our transport from scribe to kafka, but our zipkin seems not accept any message. Our kafka server have a `message.max.bytes` that are greater than the default 1024*1024 byte, and it is not easily to change it. It seems `KafkaSpanReceiver.receiverProps` is the ConsumerConfig, but we can not pass `fetch.message.max.bytes` via environment variable, which is inconvenient.

codefromthecrypt:
This adds `KAFKA_MAX_MESSAGE_SIZE` -> `fetch.message.max.bytes`

Fixes #1101


codefromthecrypt:
tested by exporting KAFKA_MAX_MESSAGE_SIZE=-1 and watched it blow up

Issue #989, opened 2016-02-19
Improve UX when multiple binary annoations exist in the same span, with the same key


codefromthecrypt:
Why does the same binary annotation pop up twice? In zipkin, it is possible to have multiple hosts in the same span. Each of them can log the same binary annotation (tag). ex. `http.status`. When they match, this looks like a duplicate. When they don't match, this could be helpful (ex the client perceived an error even if the server didn't perceive one. Single-host spans by definition don't have this problem (since only one host ever logs anything in a span).

Here's an option @yurishkuro discussed about uncovering why it looks like a dupe. Basically, if we add a column for the origin of the value, it might make more sense. I've pasted some other thoughts of his below as well.

![image](https://cloud.githubusercontent.com/assets/3523016/13147772/3d2cc928-d628-11e5-93ea-b06f67e9e436.png)

> Another common scenario I've seen with same keys is the `http.url`, which is sent by both client and server, and the value may or may not be the same.
> 
> If we adopt the suggestion that (spanID, spanKind, emittingService) combination is unique (which I think implies ditching core annotations and moving the Endpoint on the Span itself), then seemingly identical binary annotations may still be distinguishable by the service that emits them.

jcarres-mdsol:
we at least are careful to not send some stuff we know will be duplicated when the upstream system is also zipkin
https://github.com/openzipkin/zipkin-tracer/blob/master/lib/zipkin-tracer/rack/zipkin-tracer.rb#L55

codefromthecrypt:
per @CodingFabian

> we also see dupes. because we do not know if the dupes are guaranteed, we think the UI should filter dupes (our on ui does that)

yurishkuro:
I think it's ok to filter dups, provided that high fidelity data is used for comparisons.

```
span id=1
   anno:
     sr, host=1.2.3.4
   bin_anno:
     x = y
span id=1
   anno:
     cs, host=5.6.7.8
   bin_anno:
     x = y
```

here the two identical binary annotations are actually sent by different servers, and may have different meaning. This correlation gets lost once the spans are merged in the UI into one.

codefromthecrypt:
@yurishkuro do you have any particular use case where the key and value is the same for different hosts in the same span, and knowing that would have helped solve a latency issue?

yurishkuro:
no

codefromthecrypt:
[...]

PR #1114, opened 2016-05-13
Display service name when showing addresses in the span detail element

yurishkuro:
Example screenshot:
image

Why: when a request comes from or goes to uninstrumented service, the instrumented part may log CA/SA annotations that include not only the peer IP address, but also the service name. However, the current UI, while it has a special handing for CA/SA annotations, does not display the service name.

### DIFF
--- a/zipkin-ui/js/component_ui/traceToMustache.js
+++ b/zipkin-ui/js/component_ui/traceToMustache.js
@@ -79,7 +79,10 @@ function toSpanDepths(spans) {
   return treeDepths(entry, 1);
 }
 
-export function formatEndpoint({ipv4, port = 0}) {
+export function formatEndpoint({ipv4, port = 0, serviceName = ''}) {
+  if (serviceName) {
+    return `${ipv4}:${port} (${serviceName})`;
+  }
   return `${ipv4}:${port}`;
 }
 
--- a/zipkin-ui/js/component_ui/traceToMustache.js
+++ b/zipkin-ui/js/component_ui/traceToMustache.js
@@ -153,7 +153,6 @@ export default function traceToMustache(trace) {
           value: ConstantNames[a.value] || a.value,
           timestamp: a.timestamp,
           relativeTime: mkDurationStr(a.timestamp - traceTimestamp),
-          serviceName: a.endpoint && a.endpoint.serviceName ? a.endpoint.serviceName : null,
           width: 8
         })),
         binaryAnnotations
--- a/zipkin-ui/templates/trace.mustache
+++ b/zipkin-ui/templates/trace.mustache
@@ -104,15 +104,13 @@
           <thead><tr>
             <th>Date Time</th>
             <th>Relative Time</th>
-            <th>Service</th>
             <th>Annotation</th>
-            <th>Host</th>
+            <th>Address</th>
           </tr></thead>
           <tbody>
             <tr>
               <td data-key='timestamp' class="local-datetime"></td>
               <td data-key='relativeTime'></td>
-              <td data-key='serviceName'></td>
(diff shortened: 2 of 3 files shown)


yurishkuro:
@adriancole are there unit tests for what gets displayed in the UI?

codefromthecrypt:
yeap formatEndpoint has unit tests in traceToMustache.test.js

probably, just test that passing blank has the intended effect. Also I suspect the callsite needs to be updated to actually pass serviceName, right? Otherwise it is dead code.

codefromthecrypt:
oh I remember this issue! We discussed and performed no action on this in the past.

Last time we discussed this, we noted that having endpoint information for binary annotations isn't something we expect to carry forward in the model.

Making a UI change to special case when multiple services report the same value formalizes a feature that will look "missing" when we finally get time to fix the model to disallow this.

Next time you are raising an issue we have discussed and took no action on, please make reference to the former discussion and mention what has changed which would lead to us taking different action now.

codefromthecrypt:
closing this as the issue itself hasn't changed https://github.com/openzipkin/zipkin/issues/989

yurishkuro:
this PR is unrelated to that discussion. It is here to display service name normally encoded in SA/CA/LC, which is otherwise hidden (LC had a special handling, which with this PR becomes unnecessary). The service name in the SA/CA annotations is respected in other parts of the code, specifically in the logic that derives a single service name for a span: https://github.com/openzipkin/zipkin/blob/master/zipkin-ui/js/component_ui/traceSummary.js#L53

codefromthecrypt:
@yurishkuro you're right. re-opened and changing the title as this is related to "address" annotations, which are special-cased.

codefromthecrypt:
Ps can you put in a shot of the whole span detail element? It would be
easier to tell what the result looks like (ex if we are still consistent)

codefromthecrypt:
ps I caused this to slow down, and am sorry about that. In repentance, I'll add the missing tests etc.

eirslett:
I would rather not show the parentheses at all if serviceName is empty, isn't that better?

codefromthecrypt:
@eirslett agreed

codefromthecrypt:
Added tests and a more complete screen shot. One glitch is that this causes the Annotations pane to show redundant info, as it has service listed already...

screen shot 2016-05-14 at 6 53 13 pm

codefromthecrypt:
so I don't think this is shippable without revisions because it looks really chatty now. If this change was isolated to just the bottom part, or we took out the Service column in the top part, it would be shippable imho

codefromthecrypt:
I guess another option would be to change the code so that it didn't apply to both sections.

codefromthecrypt:
Yet anothet option could be to just add a service column at the bottom like there is at the top

codefromthecrypt:
Added a commit with one option, which is to make the top match the bottom. Here's a screenshot

screen shot 2016-05-14 at 7 21 13 pm

eirslett:
LGTM!

> Adrian Cole @adriancole 12:50
> @eirslett on endpoint formatting..
> 150.151.152.153:0 vs 150.151.152.153?
> we have a test to show the former, but often zero is absence of a value (at least in code I've noticed)
> I can see it both ways, but kindof prefer just dropping port from the label when zero

No strong opinions here. Leaving it as it is (showing the zero) indicates that Zipkin is indeed able to show which port a service answered on - which can be handy for people to know, when they try to improve their instrumentation/pass in more data. (port, in this case)

yurishkuro:
lgtm. Thanks for improvements & tests. I was porting this change from our pre-JS fork where it was only affecting the bottom part.

PR #1115, opened 2016-05-14
Fixes UI display of json values

### LINKED ISSUE #1113: UI does not display binary annotation value if it contains a JSON string
Our instrumentation emits some binary annotations as JSON strings. When returned from query service, it looks all properly escaped:

```
{ "key": "haproxy.LogMessage",
  "value": "{\"msg\":{\"Tw\":\"0\",\"Tt\":\"8625\",\"Tr\":\"0\",\"Tq\":\"8624\",\"Tc\":\"1\",\"backend\", ....
```

But in the span popup it is displayed like this:

```
haproxy.LogMessage  [object Object]
```

cc @eirslett

codefromthecrypt:
Before, json in a binary annotation value resulted in the UI displaying
'[object Object]'. This guards and stringifies objects, while leaving
primitives alone.

Fixes #1113

### DIFF
--- a/zipkin-ui/js/component_ui/spanPanel.js
+++ b/zipkin-ui/js/component_ui/spanPanel.js
@@ -34,7 +34,10 @@ export default component(function spanPanel() {
       const $row = self.$binaryAnnotationTemplate.clone();
       $row.find('td').each(function() {
         const $this = $(this);
-        $this.text(anno[$this.data('key')]);
+        const maybeObject = anno[$this.data('key')];
+        // In case someone is storing escaped json as binary annotation values
+        // TODO: this class is not testable at the moment
+        $this.text($.type(maybeObject) === 'object' ? JSON.stringify(maybeObject) : maybeObject);
       });
       $binAnnoBody.append($row);
     });


codefromthecrypt:
@yurishkuro @eirslett fyi, this class doesn't have a test yet, and usually it takes several hours for me to refactor things until they are testable with my current experience level. Your call whether to merge with manual testing or not. I've manually tested with below, and also tested old annotations don't get quoted.

``` bash
$ curl -s -X POST -H 'Content-Type: application/json' localhost:9411/api/v1/spans -d '[{"traceId":"1","name":"bang","id":"2","timestamp":1234,"duration":100,"binaryAnnotations":[{"key":"json","value":"{\"msg\":{\"detail\":\"foo\"}}","endpoint":{"serviceName":"flintstones","ipv4":"127.0.0.1"}}]}]'
$ open http://localhost:9411/traces/1
```

codefromthecrypt:
screen shot 2016-05-14 at 10 28 45 am

zipkin-ui/js/component_ui/spanPanel.js
```diff
@@ -34,7 +34,10 @@ export default component(function spanPanel() {
       const $row = self.$binaryAnnotationTemplate.clone();
       $row.find('td').each(function() {
         const $this = $(this);
-        $this.text(anno[$this.data('key')]);
+        const maybeObject = anno[$this.data('key')];
+        // In case someone is storing escaped json as binary annotation values
+        // TODO: this class is not testable at the moment
+        $this.text($.type(maybeObject) === "object" ? JSON.stringify(maybeObject) : maybeObject);
```
yurishkuro: looks like travis's linter wants single quotes around `"object"`

yurishkuro:
lgtm

eirslett:
:+1:

I'd rather scrap that class and rewrite it than make it testable!

anandshah05:
Thanks adrian

PR #1118, opened 2016-05-22
Removes tracegen

codefromthecrypt:
Tracegen was a tool we used to use to populate zipkin when someone had
no instrumented apps. It would create a number of traces based on latin
words.

Since then, we have increasingly sophisticated and relevant self-tracing
setup, which in the java project extends all the way to cassandra.

Moreover, we have spigo, which now includes zipkin scripts to create
large amounts of realistic traces, and instrumentation-specific examples
from Brave and Sleuth.

Considering the better alternatives, we should stop maintaining tracegen

https://github.com/adrianco/spigo
https://github.com/openzipkin/brave-resteasy-example
https://github.com/spring-cloud-samples/brewery


README.md
```diff
@@ -34,16 +34,10 @@ Here's how to start zipkin using the default file-based backend and view traces.
...
 $ git clone https://github.com/openzipkin/zipkin; cd zipkin
-# start the query server in a new terminal session or tab
+# start the query server
 $ ./bin/query
-# start the collector server in a new terminal session or tab
-$ ./bin/collector
-# start the web server in a new terminal session or tab
-$ ./bin/web
-# create dummy traces
-$ ./bin/tracegen
```
abesto, line 44: Not blocker: I'd love to see a one (two?) liner to generate traces into Zipkin with Spigo here to replace `tracegen`.

codefromthecrypt, line 44: spigo demos always use docker, so I've moved this to a new issue https://github.com/openzipkin/docker-zipkin-java/issues/6 I think there are some ways we can help with spigo to reduce steps (possibly making it report directly as opposed to generating files similar to https://github.com/adrianco/spigo/issues/67)

PR #1120, opened 2016-05-26
Two separate bugfixes.

NegatioN:
1. Bugfix goToTrace not triggering on other pages than default, creating a malformed url.
2. webpack devproxy only targeted default-page locally, and got resources from upstream api if you attached an external dev-server. The external zipkin-server can now be configured with an environment-variable.

### DIFF
--- a/zipkin-ui/README.md
+++ b/zipkin-ui/README.md
@@ -6,7 +6,7 @@ When looking at a trace, the browser is sent to the path `/traces/{id}`. For the
 app to serve that route, the server needs to forward the request to `index.html`. The same
 forwarding applies to `/dependencies` and any other routes the UI controls. 
 
-Under the scenes the JavaScript code looks at `window.location` to figure out what the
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
thanks!

mind putting in the README how to connect to an existing zipkin server (ex via docker or some other specified ip)?

NegatioN:
I'm not sure if this is exactly what you were looking for. Please give me a shout if it wasn't. :)

eirslett:
:shipit:

codefromthecrypt:
Thanks!

virtuald:
Heh. Ran into this today when I tried to modify the UI, if only if I had seen it before my flight...

codefromthecrypt:
What one  can synchronise before a flight is indeed important!

eirslett:
Flight attendants should remind people to git pull as part of the preflight instructions...

codefromthecrypt:
nice once :)

Issue #1093, opened 2016-04-19
Add "save a trace" functionality.


mjbryant:
It'd be really nice if individual traces could be tagged to not age out of Cassandra through the UI.

codefromthecrypt:
neat idea. One way could be to save them off to a user profile (even
browser storage), since traces are so small. Ex,
zipkin/api/v1/trace/trace_id becomes searchable locally. Just an idea.

PS this is only an issue in cassandra at the moment, since others don't
support TTLs anyway. It might be tricky to do this on the storage side
portably, since we don't have a portable feature for TTL. That doesn't
preclude custom logic, just explanation.

yurishkuro:
Since the UI has been decoupled from the server side, another possible solution is to have a Save As, so that the user can save the trace json as a file on local disk and later load it into the UI. It's especially useful considering that tracing ultimately is used to troubleshoot perf issues, so one could save a trace, attach it to a ticket, and someone else can later load the trace and see it in the UI.

We already have the JSON button, so Save As is there, but we don't have a Load function in the UI.

codefromthecrypt:
> We already have the JSON button, so Save As is there, but we don't have a
> Load function in the UI.
> 
> "We" are a privileged minority until #1060 :P

yurishkuro:
huh, I didn't realize, thought it was in already.

codefromthecrypt:
PS there's now a download button on a trace.

codefromthecrypt:
We had a note in #1222 about just making a separate non-bucketed index for elasticsearch and just move docs to it. There's no TTL support in mysql anyway, so noop there. Cassandra might take some thinking.

We can address this by making it possible to query the "infinite" index routinely, and we could also make a "request save" api, which either moves the trace there or returns a message if unsupported.

cc @openzipkin/elasticsearch @openzipkin/cassandra

codefromthecrypt:
also in mysql I suppose we could double the table-count to provide an "infinite" index separate from the routine one (cc @jcarres-mdsol)

michaelsembwever:
**Cassandra**
Simplest approach is to rewrite the data with new TTL=-1

Technical: This will leave some old sstables around and prevent them from being wiped off disk in one go (tombstone compactions will instead be required to clean them out). But I can't see this being a visible issue to the zipkin operator.

codefromthecrypt:
thx @michaelsembwever 

[...]

PR #1124, opened 2016-05-28
Add JSON button to trace page

### LINKED ISSUE #1060: Add JSON button to trace detail view
The screenshot on zipkin.io includes a blue JSON button which is a link to /api/v1/trace/id. This would be very helpful when folks are trying to debug problems, as manually instructing is not intuitive.

virtuald:
Well, since the "[privileged minority](https://github.com/openzipkin/zipkin/issues/1093#issuecomment-213216090)" hasn't done it yet, I figured I would since I keep finding that I need it. :) 
- Shows prettified JSON in modal dialog
- Fixes #1060

I'm very new to the various frameworks that the UI is using, so I'm open to suggestions if there's anything that I did that was a bit wonky.

### DIFF
--- a/zipkin-ui/js/component_ui/jsonPanel.js
+++ b/zipkin-ui/js/component_ui/jsonPanel.js
@@ -1,0 +1,16 @@
+import {component} from 'flightjs';
+
+export default component(function jsonPanel() {
+  this.show = function(e, data) {
+    this.$node.find('.modal-title').text(data.title);
+
+    this.$node.find('.save').attr('href', data.link);
+    this.$node.find('.modal-body pre').text(JSON.stringify(data.obj, null, 2));
+    this.$node.modal('show');
+  };
+
+  this.after('initialize', function() {
+    this.$node.modal('hide');
+    this.on(document, 'uiRequestJsonPanel', this.show);
+  });
+});
--- a/zipkin-ui/js/component_ui/traceToMustache.js
+++ b/zipkin-ui/js/component_ui/traceToMustache.js
@@ -88,6 +88,7 @@ export function formatEndpoint({ipv4, port = 0, serviceName = ''}) {
 
 export default function traceToMustache(trace) {
   const summary = traceSummary(trace);
+  const traceId = summary.traceId;
   const duration = mkDurationStr(summary.duration);
   const groupedTimestamps = getGroupedTimestamps(summary);
   const serviceDurations = getServiceDurations(groupedTimestamps);
@@ -167,6 +168,7 @@ export default function traceToMustache(trace) {
   const spansBackup = spans;
 
   return {
+    traceId,
     duration,
     services,
     depth,
--- a/zipkin-ui/js/page/trace.js
+++ b/zipkin-ui/js/page/trace.js
@@ -3,6 +3,7 @@ import $ from 'jquery';
 import TraceData from '../component_data/trace';
 import FilterAllServicesUI from '../component_ui/filterAllServices';
 import FullPageSpinnerUI from '../component_ui/fullPageSpinner';
+import JsonPanelUI from '../component_ui/jsonPanel';
 import ServiceFilterSearchUI from '../component_ui/serviceFilterSearch';
 import SpanPanelUI from '../component_ui/spanPanel';
 import TraceUI from '../component_ui/trace';
@@ -17,18 +18,26 @@ const TracePageComponent = component(function TracePage() {
     TraceData.attachTo(document, {
       traceId: this.attr.traceId
     });
-    this.on(document, 'tracePageModelView', function(ev, modelview) {
-      this.$node.html(traceTemplate(modelview));
+    this.on(document, 'tracePageModelView', function(ev, data) {
+      this.$node.html(traceTemplate(data.modelview));
 
       FilterAllServicesUI.attachTo('#filterAllServices', {
         totalServices: $('.trace-details.services span').length
       });
       FullPageSpinnerUI.attachTo('#fullPageSpinner');
+      JsonPanelUI.attachTo('#jsonPanel');
       ServiceFilterSearchUI.attachTo('#serviceFilterSearch');
       SpanPanelUI.attachTo('#spanPanel');
       TraceUI.attachTo('#trace-container');
       FilterLabelUI.attachTo('.service-filter-label');
       ZoomOut.attachTo('#zoomOutSpans');
+
+      this.$node.find('#traceJsonLink').click(e => {
+        e.preventDefault();
+        this.trigger('uiRequestJsonPanel', {title: `Trace ${this.attr.traceId}`,
+                                            obj: data.trace,
+                                            link: `/api/v1/trace/${this.attr.traceId}`});
+      });
 
       $('.annotation:not(.core)').tooltip({placement: 'left'});
     });
--- a/zipkin-ui/templates/trace.mustache
+++ b/zipkin-ui/templates/trace.mustache
@@ -4,6 +4,7 @@
     <li class=''><a href='#'><strong>Services:</strong> <span class='badge'>{{services}}</span></a></li>
     <li class=''><a href='#'><strong>Depth:</strong> <span class='badge'>{{depth}}</span></a></li>
     <li class=''><a href='#'><strong>Total Spans:</strong> <span class='badge'>{{totalSpans}}</span></a></li>
+    <li class='navbar-right'><a href='/api/v1/trace/{{ traceId }}' id='traceJsonLink'><span class='btn btn-primary btn-xs badge'>JSON</span></a></li>
   </ul>
 
   <form class='form-inline' role='form'>
@@ -197,5 +198,20 @@
     {{/binaryAnnotations}}
   </div>
   {{/spansBackup}}
+</div>
+
+<div class='modal fade' id='jsonPanel'>
+  <div class='modal-dialog modal-lg'>
+    <div class='modal-content'>
+      <div class='modal-header'>
+        <button type='button' class='close' data-dismiss='modal' aria-hidden='true'>&times;</button>
+        <a href='#' class='save'><span class='glyphicon glyphicon-save' aria-hidden='true' aria-label='save'></span></a>
+        <h4 class='modal-title'></h4>
+      </div>
+      <div class='modal-body'>
+        <pre></pre>
+      </div>
+    </div>
+  </div>
 </div>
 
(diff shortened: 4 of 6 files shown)


zipkin-ui/js/component_ui/jsonPanel.js
```diff
@@ -0,0 +1,16 @@
+import {component} from 'flightjs';
+import $ from 'jquery';
```
eirslett: Defined, but never used (try running `npm run lint`)

zipkin-ui/js/component_ui/traceToMustache.js
```diff
@@ -87,7 +87,9 @@ export function formatEndpoint({ipv4, port = 0, serviceName = ''}) {
 }
 
 export default function traceToMustache(trace) {
+  const raw = trace;
```
eirslett: The raw data doesn't belong inside the modelview; this function is supposed to transform the incoming raw data into a format that is ready for mustache rendering.
Can you try refactoring, so that raw is not passed as part of the modelview?

zipkin-ui/js/page/trace.js
```diff
@@ -24,9 +25,10 @@ const TracePageComponent = component(function TracePage() {
         totalServices: $('.trace-details.services span').length
       });
       FullPageSpinnerUI.attachTo('#fullPageSpinner');
+      JsonPanelUI.attachTo('#jsonPanel');
       ServiceFilterSearchUI.attachTo('#serviceFilterSearch');
       SpanPanelUI.attachTo('#spanPanel');
-      TraceUI.attachTo('#trace-container');
+      TraceUI.attachTo('#trace-container', {raw: modelview.raw});
```
eirslett: The parameter that is called "modelview" today could be renamed to "data", and then "data" would contain two properties: "modelview" and "raw"?

eirslett:
I had some comments; otherwise, great work!

codefromthecrypt:
Thanks for the effort and review folks. This is very popular!

virtuald:
Thanks for the comments, I'll address these tonight or tomorrow.

virtuald:
Comments addressed.

codefromthecrypt:
running it locally and will merge (and release) after!

virtuald:
I've got another set of changes coming too dealing with better error messages, just testing it now.

codefromthecrypt:
ah ok thx for the heads-up. this is a different PR, right?

virtuald:
Yep. Different PR.

zipkin-ui/js/component_ui/trace.js
```diff
@@ -443,7 +448,11 @@ export default component(function trace() {
     this.on(document, 'uiZoomInSpans', this.zoomInSpans);
     this.on(document, 'uiZoomOutSpans', this.zoomOutSpans);
 
+    this.on(document, 'uiShowJson', this.showJson);
+
     const self = this;
+    self.raw = data.raw;
```
codefromthecrypt: how is this controlled?

codefromthecrypt:
fyi here are some screen shots:

screen shot 2016-05-31 at 10 49 49 am

screen shot 2016-05-31 at 10 49 36 am

zipkin-ui/js/component_ui/trace.js
virtuald: It comes through the trigger in `js/component_data/trace.js`

codefromthecrypt: sorry I meant how do you control this via the UI. Ex if the user wants to
view the raw json (api/v1/trace/id?raw), what do they do? Or does raw mean
something else?

virtuald: This PR does not provide a mechanism to do that. I suspect that is best saved for future work -- if there isn't a way to show an uploaded trace, no sense downloading the trace? 

I could be convinced otherwise, it would be easy to add an actual raw download link somewhere.

yurishkuro:
is json raw or merged?

virtuald:
The JSON should be whatever was retrieved from the call to `/api/v1/traces/XXX`, as it's the same object.

codefromthecrypt:
@yurishkuro you have this feature already.. do you default to raw json? any thoughts about what the default should be?

zipkin-ui/js/component_ui/trace.js
codefromthecrypt: "raw" in the code here confused me as it is a parameter GET /api/v1/trace
api

virtuald: Oh, ha. Didn't realize that.

codefromthecrypt: ex not sure why the website doesn't render it, but here's the definition
- name: raw
  in: query
  required: false
  description: |
  Note this flag has no value. Ex. /trace/{traceId}?raw
  Normally, the trace endpoint cleans trace data. For example, it merges
  spans by id, adds missing timestamp or duration, corrects clock skew..
  Specifying this flag is a debug case, when you are debugging zipkin
  logic or zipkin instrumentation, and want to see the input to these
  adjusters. For example, this might explain or rule out clock skew.
  type: boolean

https://github.com/openzipkin/zipkin-api/blob/master/zipkin-api.yaml#L146

virtuald: Interesting. I suppose we could add a mechanism that could display either. I chose the default since it was already retrieved by the server -- so it would show the JSON that's currently being displayed without another call to the server. Theoretically, if you made another call down, the JSON could show something different which wasn't currently being displayed.

codefromthecrypt: good point.

Also, principle of least surprise wins for not defaulting to "raw" flag:
without disclaimers, it would confuse people when timestamps don't match
what the UI says.

I'm game for as-is.. I'd prefer that variable to be named something besides
raw (or clarified that it doesn't imply the raw parameter)

virtuald: Renamed to `traceJson`

yurishkuro:
@adriancole  we added two buttons, raw and regular json, as both come in handy. But we just have them as plain links to json documents, not popups, and we leave pretty rendering to browser plugin. That way it's easy to search inside the json, and do a Save As.

codefromthecrypt:
interesting. search works fine with the pop-up, but save-as doesn't. It is
a common support case... @virtuald https://github.com/virtuald thoughts
on save-as etc?

codefromthecrypt:
screen shot 2016-05-31 at 10 49 36 am

Here's a thought. you can add a save or download link to the pop-up. this could point to /api/v1/trace/id for example.

eirslett:
+1 for adriancole's suggestion!

virtuald:
I added a save icon, and also the ability to right click on the JSON button and do a save as (as that's an intuitive option anyways).

zipkin-ui/js/component_ui/traceToMustache.js
```diff
@@ -167,6 +168,7 @@ export default function traceToMustache(trace) {
   const spansBackup = spans;
 
   return {
+    traceId,
```
eirslett, line 171: I guess this isn't needed anymore?

zipkin-ui/templates/trace.mustache
```diff
@@ -4,13 +4,14 @@
     <li class=''><a href='#'><strong>Services:</strong> <span class='badge'>{{services}}</span></a></li>
     <li class=''><a href='#'><strong>Depth:</strong> <span class='badge'>{{depth}}</span></a></li>
     <li class=''><a href='#'><strong>Total Spans:</strong> <span class='badge'>{{totalSpans}}</span></a></li>
+    <li class='navbar-right'><a href='/api/v1/trace/{{ traceId }}' id='traceJsonLink'><button type="button" class="btn btn-primary btn-xs badge" value="uiShowJson">JSON</button></a></li>
```
eirslett: You don't need `<button>`here, you can just put the classes on the A directly, like this: `<a class="btn btn-primary btn-xs badge">` (I think)

zipkin-ui/templates/trace.mustache
```diff
@@ -4,13 +4,14 @@
...
   </ul>
 
   <form class='form-inline' role='form'>
     <div class='btn-group btn-group-sm' id='filterAllServices'>
       <button type='button' class='btn btn-default' value='uiExpandAllSpans'>Expand All</button>
       <button type='button' class='btn btn-default active' value='uiCollapseAllSpans'>Collapse All</button>
-    </div>
+    </div>
```
eirslett: This extra space isn't needed

zipkin-ui/js/component_ui/traceToMustache.js
virtuald, line 171: It's used in trace.mustache line 7.

zipkin-ui/templates/trace.mustache
virtuald: Hm, can you use the `value` thing on an `<a>` element? Using the same event mechanism as everything else is convenient.

zipkin-ui/js/component_ui/traceToMustache.js
eirslett, line 171: Ah, I see it now! Sorry.

zipkin-ui/templates/trace.mustache
eirslett: No, it doesn't have the value property. How about using an explicit click handler on the a element? I think that's cleaner overall.

virtuald: Ok, done.

zipkin-ui/js/page/trace.js
```diff
@@ -17,19 +18,27 @@ const TracePageComponent = component(function TracePage() {
...
       });
       FullPageSpinnerUI.attachTo('#fullPageSpinner');
+      JsonPanelUI.attachTo('#jsonPanel');
       ServiceFilterSearchUI.attachTo('#serviceFilterSearch');
       SpanPanelUI.attachTo('#spanPanel');
       TraceUI.attachTo('#trace-container');
       FilterLabelUI.attachTo('.service-filter-label');
       ZoomOut.attachTo('#zoomOutSpans');
 
+      this.$node.find('#traceJsonLink').click(e => {
```
virtuald, line 35: @eirslett you're right, this ends up being cleaner anyways, it never really belonged in trace.js

eirslett:
Otherwise, LGTM! ship it

codefromthecrypt:
yay json button!

PR #1125, opened 2016-05-31
Show errors

virtuald:
(Includes #1124 because of potential merge conflicts)

This is 3 separate pieces that have 3 commits associated with them:
- Show an error when config.json isn't present
- Provide mechanisms to show errors if one of the ajax calls loading data fails
- Show the user an error when a query fails (this uses a separate mechanism to ensure that it's placed in a semantically meaningful way... a query that fails is something like `foo==bar`)

### DIFF
--- a/zipkin-ui/css/main.scss
+++ b/zipkin-ui/css/main.scss
@@ -39,6 +39,10 @@ body {
   cursor: pointer;
 }
 
+#errorPanel {
+  display: none;
+}
+
 #fullPageSpinner {
   display: none;
   position: absolute;
--- a/zipkin-ui/js/component_data/default.js
+++ b/zipkin-ui/js/component_data/default.js
@@ -1,4 +1,5 @@
 import {component} from 'flightjs';
+import {errToStr} from '../../js/component_ui/error';
 import $ from 'jquery';
 import queryString from 'query-string';
 import {traceSummary, traceSummariesToMustache} from '../component_ui/traceSummary';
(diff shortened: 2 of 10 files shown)


virtuald:
Oops, forgot to lint.

codefromthecrypt:
can you post an example screen shot of config missing and also bad request?

codefromthecrypt:
ps great work!

virtuald:
### config.json

When config.json is missing, it's a completely blank screen, with text to the effect of

```
Error loading config.json: server error (Bad Gateway)
```

The rationale for this is that [the mechanism for showing the entire UI](https://github.com/openzipkin/zipkin/blob/master/zipkin-ui/js/main.js#L9) relies on config.json, so if it's not there then loading the UI doesn't make sense. Typically you only run into this with a misconfiguration anyways -- but right now, you get a blank page, which is confusing. This is better than a blank page.

This can be triggered by running `npm run dev` without launching a zipkin query server.

### ajax errors

Screenshot for bad gateway (this can be triggered by runnning `npm run dev` without launching a zipkin server, just make sure to add a `{}` to `static/config.json` so you don't trigger the config.json error instead).

span_loads

### query errors

Screenshot for bad query such as `foo==bar`... the error message could be better if the query server returned a better message. That's something for a future PR.
bad_query

codefromthecrypt:
:thumbsup:

codefromthecrypt:
needs a rebase, then ready to go

virtuald:
Oops, thought I had done that. Here you go.

eirslett:
Great work!
Maybe the only thing missing, that we should add, is to catch exceptions in traceToMustache.js, so that any bugs in my sloppy JavaScript code will be shown with a stack trace, instead of rendering a black page :stuck_out_tongue:

codefromthecrypt:
pink is the new black, I guess

PR #1128, opened 2016-06-04
Makes CASSANDRA_ENSURE_SCHEMA work when schema is absent

codefromthecrypt:
Our cassandra docker image comes with the schema installed, which hid a
bug where we failed to properly guard state 0 (keyspace didn't exist).

This fixes the bug and also backfills known upgrade scenarios. Tested
on Cassandra 2.1 and 2.2.

Thanks to @gena01 for reporting

### DIFF
--- a/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/Schema.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/Schema.java
@@ -63,10 +63,8 @@ final class Schema {
   }
 
   static KeyspaceMetadata getKeyspaceMetadata(Session session) {
-    return getKeyspaceMetadata(session.getLoggedKeyspace(), session.getCluster());
-  }
-
-  private static KeyspaceMetadata getKeyspaceMetadata(String keyspace, Cluster cluster) {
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
ps the error would bubble as `Cannot read keyspace metadata for give keyspace`, and before this change, the only way out would be to set `CASSANDRA_ENSURE_SCHEMA=false`

codefromthecrypt:
fyi @michaelsembwever I'm going to merge this as it is not safe to assume folks have installed the keyspace

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/Schema.java
```diff
@@ -78,10 +76,12 @@ private static KeyspaceMetadata getKeyspaceMetadata(String keyspace, Cluster clu
   }
 
   static void ensureExists(String keyspace, Session session) {
-    KeyspaceMetadata keyspaceMetadata = getKeyspaceMetadata(keyspace, session.getCluster());
```
codefromthecrypt, line 81: this was the chicken-egg. if the keyspace wasn't installed, getKeyspaceMetadata would throw

PR #1136, opened 2016-06-21
Permits json null token on nullable fields

### LINKED ISSUE #1130: Can't register spans with zipkin-tracer gem
I use json-api with zipkin-tracer gem.
Zipkin-tracer send a request with null parentId, but zipkin 1.1.4 is not accepted null parentId.
I tried to remove parentId when parentId is null and send a request, so it was successful.

zipkin error meessage:

> Cannot decode spans due to IllegalArgumentException(Expected a string but was NULL at path $[0].parentId reading List<Span> from json:

versions:
- zipkin 1.1.4
- zipkin-tracer 0.18.1 json-api tracer.

codefromthecrypt:
While conventionally, we omit fields rather than passing them as null,
null is valid json and was emitted by at least one tracer. This changes
the parser to ignore fields declared like `"parentId": null`

Fixes #1130

### DIFF
--- a/zipkin/src/main/java/zipkin/internal/JsonCodec.java
+++ b/zipkin/src/main/java/zipkin/internal/JsonCodec.java
@@ -110,11 +110,16 @@ public final class JsonCodec implements Codec {
       }
       writer.endObject();
     }
-  };
+  }.nullSafe();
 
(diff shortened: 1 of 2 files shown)


eirslett:
+1 LGTM

PR #1144, opened 2016-06-24
Expose prometheus compatible metrics on /prometheus

klette:
We run Zipkin in several kubernetes clusters and thus we need a sane way of collecting the metrics.
As Prometheus has all the built-in support for dealing with the moving parts of Kubernetes, it seems
like a rational choice.

This PR exposes Spring Actuator metrics on `/prometheus`
using the 0.0.4 version of their text format specified
here https://prometheus.io/docs/instrumenting/exposition_formats/.

Not crazy about the path, but adding new controllers into /metrics/\* seems like more complexity than needed.

### DIFF
--- a/zipkin-autoconfigure/metrics-prometheus/README.md
+++ b/zipkin-autoconfigure/metrics-prometheus/README.md
@@ -1,0 +1,18 @@
+# Prometheus Metrics
+
+Exposes [Spring Actuator metrics](http://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-metrics.html)
+on `/prometheus` using the Prometheus exposition [text format version 0.0.4](https://prometheus.io/docs/instrumenting/exposition_formats/).
+
+All metrics defaults to the `gauge` type, unless a known format is specified by a prefix in the metric name. Currently the only two detected types are `gauge_` and `counter_`.
+
+## Scrape configuration example
+
+```yaml
+  - job_name: 'zipkin'
+    scrape_interval: 5s
+    metrics_path: '/prometheus'
+    static_configs:
+      - targets: ['localhost:9411']
+
+```
+
--- a/zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin/autoconfigure/metrics/PrometheusMetricsAutoConfiguration.java
+++ b/zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin/autoconfigure/metrics/PrometheusMetricsAutoConfiguration.java
@@ -33,10 +33,10 @@ public class PrometheusMetricsAutoConfiguration {
     private static final Pattern SANITIZE_PREFIX_PATTERN = Pattern.compile("^[^a-zA-Z_]");
     private static final Pattern SANITIZE_BODY_PATTERN = Pattern.compile("[^a-zA-Z0-9_]");
 
-    private Collection<PublicMetrics> publicMetrics;
+    private final Collection<PublicMetrics> publicMetrics;
 
     @Autowired
-    public PrometheusMetricsAutoConfiguration(final Collection<PublicMetrics> publicMetrics) {
+    public PrometheusMetricsAutoConfiguration(Collection<PublicMetrics> publicMetrics) {
         this.publicMetrics = publicMetrics;
     }
 
--- a/zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin/autoconfigure/metrics/PrometheusMetricsAutoConfiguration.java
+++ b/zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin/autoconfigure/metrics/PrometheusMetricsAutoConfiguration.java
@@ -1,0 +1,102 @@
+/**
+ * Copyright 2015-2016 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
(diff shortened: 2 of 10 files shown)


eirslett:
@negatioN interesting for FINN?

codefromthecrypt:
cool. I didn't expect this to be possible without adding a dependency (though we ought to have a test dependency to ensure the results are parseable by prometheus).

Metrics export is something that is quite deployment-specific.. would you mind creating an autoconfigure module for this instead? you can then make that module an optional dependency of zipkin-server. This allows it to get into the all-jar (for docker etc), but not pulled in automatically for those running custom builds.

You could do still do everything in one class, just it reduces the scope a bit. That module should include a README saying what version of prometheus format it works with, and ideally a test parser.

I think this is fair to add to core as modularity in docker is still not worked out, this is only a small dependency, and assuming @kristofa also wants this, we have 2 core committers who work at different companies interested in prometheus (and might be able to help support bug-fixes).

Sound reasonable?

kristofa:
Prometheus support would be great.
Definitely interested.

Op zondag 26 juni 2016 heeft Adrian Cole notifications@github.com het
volgende geschreven:

klette:
Alright-y,

moved the code into a auto configure module, and added some docs. Couldn't find any Java based parser, and building and running go code as a part of the test suite seemed a bit overkill :)

zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin/autoconfigure/metrics/PrometheusMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,102 @@
...
+@RequestMapping("/prometheus")
+public class PrometheusMetricsAutoConfiguration {
+
+    private static final Pattern SANITIZE_PREFIX_PATTERN = Pattern.compile("^[^a-zA-Z_]");
+    private static final Pattern SANITIZE_BODY_PATTERN = Pattern.compile("[^a-zA-Z0-9_]");
+
+    private Collection<PublicMetrics> publicMetrics;
+
+    @Autowired
+    public PrometheusMetricsAutoConfiguration(final Collection<PublicMetrics> publicMetrics) {
```
codefromthecrypt: nit: if you're looking for final, I'd move it to the field

zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin/autoconfigure/metrics/PrometheusMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,102 @@
...
+        return SANITIZE_BODY_PATTERN.matcher(
+                SANITIZE_PREFIX_PATTERN.matcher(metricName).replaceFirst("_")
+        ).replaceAll("_");
+    }
+
+    @RequestMapping(method = RequestMethod.GET)
+    public ResponseEntity<String> prometheusMetrics() {
+        StringBuilder sb = new StringBuilder();
+
+        for (PublicMetrics publicMetrics : this.publicMetrics) {
```
codefromthecrypt, line 53: interesting.. so this isn't just zipkin, this is all spring metrics!

zipkin-autoconfigure/metrics-prometheus/src/test/java/zipkin/autoconfigure/metrics/PrometheusMetricsAutoConfigurationTest.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Collections;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.junit.Assert.assertThat;
+
+public class PrometheusMetricsAutoConfigurationTest {
+
+    @Test
+    public void correctHttpResponse() throws Exception {
+        final PublicMetrics publicMetrics = () -> Collections.singleton(new Metric<Number>("mem.free", 1024));
```
codefromthecrypt: no big deal, but we don't put final on variables even if it is explicit and there are good reasons why some folks make conventions like that.

leaving them out keeps line shorter and focuses more attention on the test code (at risk of someone doing the shell-game with references)

zipkin-autoconfigure/metrics-prometheus/README.md
```diff
@@ -0,0 +1,18 @@
+# Prometheus Metrics
```
codefromthecrypt, line 1: should probably make a link from zipkin-server's README to this, under the metrics section. Saying something like: Zipkin also exports metrics in prometheus format (link)

codefromthecrypt:
No required changes on my end, except linking to the server's README (to make it more discoverable)

nice work!

zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin/autoconfigure/metrics/PrometheusMetricsAutoConfiguration.java
klette: Fixed :)

zipkin-autoconfigure/metrics-prometheus/README.md
klette, line 1: Done!

zipkin-autoconfigure/metrics-prometheus/src/test/java/zipkin/autoconfigure/metrics/PrometheusMetricsAutoConfigurationTest.java
klette: Ah, cool. Just my brain that does that automatically after x years with that as a coding standard :)

codefromthecrypt:
nice work. squashed on entry

codefromthecrypt:
hi all. I just noticed that prometheus' java client now ships with spring boot support. any thoughts on whether we should switch? https://github.com/prometheus/client_java/tree/master/simpleclient_spring_boot/src/main/java/io/prometheus/client/spring/boot

cc @shakuzen

Issue #1141, opened 2016-06-24
Why is the data lost in elasticsearch ?


liangman:
I used the zipkin, kafka, and elasticsearch for testing.  The elasticsearch was only a node. It wasn't problem that the data was transferred from kafka to zipkin, but the data is lost from zipkin to elasticsearch. I wrote 500000 log-data to kafka, but it was only 212162 in the elasticsearch. 
The data:

``` json
[{"traceId":"ffffffffffffffff","name":"fermentum","id":"ffffffffffffffff","annotations":[{"timestamp":1466386794757000,"value":"sr","endpoint":{"serviceName":"semper","ipv4":"113.29.89.129","port":2131}},{"timestamp":1466386794757000,"value":"sagittis","endpoint":{"serviceName":"semper","ipv4":"113.29.89.129","port":2131}},{"timestamp":1466386794758000,"value":"montes","endpoint":{"serviceName":"semper","ipv4":"113.29.89.129","port":2131}},{"timestamp":1466386794758000,"value":"augue","endpoint":{"serviceName":"semper","ipv4":"113.29.89.129","port":2131}},{"timestamp":1466386794759000,"value":"malesuada","endpoint":{"serviceName":"semper","ipv4":"113.29.89.129","port":2131}},{"timestamp":1466386794760000,"value":"ss","endpoint":{"serviceName":"semper","ipv4":"113.29.89.129","port":2131}}],"binaryAnnotations":[{"key":"mollis","value":"hendrerit","endpoint":{"serviceName":"semper","ipv4":"113.29.89.129","port":2131}}]}]
```

"ffffffffffffffff" was replaced in "1-5000000";
It was no regular!!!

codefromthecrypt:
there's a "/metrics" endpoint. It would show how many spans were accepted by kafka on the zipkin-server side.  can you send relevant output from that?

also, I don't understand what you mean by

>  "ffffffffffffffff" was replaced in "1-5000000";

do you mean that when you post a span where the id is all bits set, it comes back not in hex?

liangman:
I set id from 1-500000 in hex. I think that the index have a conflict.

liangman:
The lost data is different for storing in the elasticsearch  every, when I wrote data in the kafka.

codefromthecrypt:
wondering if this is to do with the data being older? Ex. if I post this, I have to set lookback to a relatively high value to see it.

``` bash
[...]

PR #1146, opened 2016-06-26
Uses ES bulk api only when there's more than one span

codefromthecrypt:
During a test where 100 single-span messages are sent to Kafka at the
same time, I noticed only 58-97 of them would end up in storage
eventhough all messages parsed properly and no operations failed.

After a 100ms/message pause was added, the store rate of this test went
to 100%, so figured it was some sort of state issue. I noticed the code
was using Bulk operations regardless of input size, so as a wild guess
changed the special-case single-span messages. At least in this test, it
raised the success rate to 100% without any pausing needed.

I don't know why this worked, but it seems sensible to not use bulk apis
when there's no bulk action to perform.

I started to write a unit test to validate single-length lists don't use
bulk, but the Mockito involved became too verbose as the Elasticsearch
client uses chaining and other patterns that are tedious to mock.

Instead, we should make a parallel integration test and apply them to
all storage components.

See #1141

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/ElasticsearchSpanConsumer.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/ElasticsearchSpanConsumer.java
@@ -16,6 +16,7 @@ package zipkin.storage.elasticsearch;
 import com.google.common.base.Function;
 import com.google.common.base.Functions;
 import com.google.common.util.concurrent.AsyncFunction;
+import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
@@ -28,6 +29,7 @@ import zipkin.internal.ApplyTimestampAndDuration;
 import zipkin.storage.guava.GuavaSpanConsumer;
 
 import static com.google.common.util.concurrent.Futures.transform;
+import static zipkin.storage.elasticsearch.ElasticFutures.toGuava;
 
 final class ElasticsearchSpanConsumer implements GuavaSpanConsumer {
   private static final Function<Object, Void> TO_VOID = Functions.<Void>constant(null);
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
ping @anuraaga

zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/ElasticsearchSpanConsumer.java
```diff
@@ -42,24 +44,35 @@
 
   @Override
   public ListenableFuture<Void> accept(List<Span> spans) {
-    BulkRequestBuilder request = client.prepareBulk();
-    for (Span span : spans) {
-      request.add(createSpanIndexRequest(ApplyTimestampAndDuration.apply(span)));
+    if (spans.isEmpty()) return Futures.immediateFuture(null);
+
+    // Create a bulk request when there is more than one span to store
+    ListenableFuture<?> future;
```
anuraaga, line 50: Recommend `final` since it's a branched initialization

anuraaga:
LGTM - thanks for fixing (and feel the pain on the difficulty of unit testing ES client...).

zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/ElasticsearchSpanConsumer.java
codefromthecrypt, line 50: sadly can't unless we refactor further due to the later if
(ElasticsearchStorage.FLUSH_ON_WRITES)

codefromthecrypt, line 50: oh.. actually that looks not too bad.. will do

codefromthecrypt, line 50: nope was too bad.. leaving un-final :P

PR #1147, opened 2016-06-26
Add useful documentation for doing zipkin-ui development

virtuald:
I keep forgetting this, and it's not very easy to discover, so I documented it.

### DIFF
--- a/zipkin-ui/README.md
+++ b/zipkin-ui/README.md
@@ -31,3 +31,42 @@ program that can extract zip files.
 
 By specifying the `proxy` environment variable, you can point the zipkin-ui to a different backend, allowing you to access real data while developing locally.
 An example to run with npm would be `proxy=http://myzipkininstance.com:9411 npm run dev`. (note that prefixing with http:// and suffixing the port is mandatory)
+
+## What's the easiest way to develop against this locally?
+
+The maven install process already downloads everything needed to do development,
+so you don't need to install node/npm or whatever. Instead, you can use the
+`./npm.sh` shell script to perform npm operations. Here's how you launch zipkin
+server and webapp to work together:
+
+* In one terminal, go to the root of the zipkin repo and run this to build zipkin:
+
+```bash
+# In one terminal, build the server and also make its dependencies (run from the root of the zipkin repo)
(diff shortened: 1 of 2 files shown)


eirslett:
Thanks a lot, good work!
Note that you will need npm (and thus node) installed in order to run "npm run dev".

virtuald:
Nope, that's actually the point -- when you run the `mvn` command, it downloads node/npm for you, and it's sitting in the `target/node` directory.

eirslett:
I know - I wrote that code ;-) what I mean is that we need npm on $PATH somewhere so that you can "npm run dev" from the command line.
One option is to use a helper script, so you can run "./npm run dev" or something like that?

virtuald:
The instructions that I put in there have a `PATH` in it, so it works. :) 

A helper script would make a lot of sense, however -- I can add that to this PR if you want.

zipkin-ui/README.md
```diff
@@ -31,3 +31,37 @@ program that can extract zip files.
...
+
+#### What's an easy way to create new spans for testing?
+
+Using this setup, if you open the web UI and find a trace from zipkin-server,
+you can download a JSON blob by right clicking on the JSON button on the trace 
+page (top right) and doing a "Save As". Modify the span to your heart's content,
+and then you can submit the span(s) via curl:
+
+```
+$ curl -H "Content-Type: application/json" --data @span.json http://localhost:9411/api/v1/spans
```
codefromthecrypt, line 71: heh didn't even think that `--data` implies POST. nice shortcut!

codefromthecrypt, line 71: in case you want some more characters chopped.. `-s` lets you skip the `http://`. Though I can see value in not doing this since it is intuitive to see the whole url

codefromthecrypt:
👍

zipkin-ui/README.md
virtuald, line 71: Yeah, that seems a bit much.

virtuald:
Updated docs, added `./npm.sh`

codefromthecrypt:
LGTM

eirslett:
Nice!

PR #1150, opened 2016-06-27
Adds properties that affect Elasticsearch index scalability

codefromthecrypt:
This adds the following properties, which help control the scalability
of Zipkin's index in Elasticsearch.

```
* `ES_INDEX_SHARDS`: The number of shards to split the index into. Each shard and its replicas
                     are assigned to a machine in the cluster. Increasing the number of shards
                     and machines in the cluster will improve read and write performance. Number
                     of shards cannot be changed for existing indices, but new daily indices
                     will pick up changes to the setting. Defaults to 5.
* `ES_INDEX_REPLICAS`: The number of replica copies of each shard in the index. Each shard and
                       its replicas are assigned to a machine in the cluster. Increasing the
                       number of replicas and machines in the cluster will improve read
                       performance, but not write performance. Number of replicas can be changed
                       for existing indices. Defaults to 1. It is highly discouraged to set this
                       to 0 as it would mean a machine failure results in data loss.
```

### DIFF
--- a/zipkin-autoconfigure/storage-elasticsearch/src/main/java/zipkin/autoconfigure/storage/elasticsearch/ZipkinElasticsearchStorageProperties.java
+++ b/zipkin-autoconfigure/storage-elasticsearch/src/main/java/zipkin/autoconfigure/storage/elasticsearch/ZipkinElasticsearchStorageProperties.java
@@ -20,21 +20,16 @@ import zipkin.storage.elasticsearch.ElasticsearchStorage;
 
 @ConfigurationProperties("zipkin.storage.elasticsearch")
 public class ZipkinElasticsearchStorageProperties {
-  /**
-   * The elasticsearch cluster to connect to, defaults to "elasticsearch".
-   */
+  /** @see ElasticsearchStorage.Builder#cluster(String) */
   private String cluster = "elasticsearch";
-
-  /**
-   * A list of elasticsearch hostnodes to connect to, in host:port format. The port should be the
-   * transport port, not the http port. Defaults to "localhost:9300".
-   */
+  /** @see ElasticsearchStorage.Builder#hosts(List) */
   private List<String> hosts = Collections.singletonList("localhost:9300");
-
-  /**
-   * The index prefix to use when generating daily index names. Defaults to zipkin.
-   */
+  /** @see ElasticsearchStorage.Builder#index(String) */
   private String index = "zipkin";
+  /** @see ElasticsearchStorage.Builder#indexShards(int) */
+  private int indexShards = 5;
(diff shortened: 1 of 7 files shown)


codefromthecrypt:
This came to mind when troubleshooting #1141

@anuraaga PTAL, particularly on the descriptions.

zipkin-autoconfigure/storage-elasticsearch/src/main/java/zipkin/autoconfigure/storage/elasticsearch/ZipkinElasticsearchStorageProperties.java
```diff
@@ -36,6 +36,24 @@
    */
   private String index = "zipkin";
 
+
+  /**
+   * Affects scalability by limiting the number of nodes in the Elasticsearch cluster that can
```
anuraaga: While in some sense it "limits", I think it's a bit too semantic.

`The number of shards to split the index into. Each shard and its replicas are assigned to a machine in the cluster. Increasing the number of shards and machines in the cluster will improve read and write performance. Number of shards cannot be changed for existing indices, but new daily indices will pick up changes to the setting. Defaults to 5.`

zipkin-autoconfigure/storage-elasticsearch/src/main/java/zipkin/autoconfigure/storage/elasticsearch/ZipkinElasticsearchStorageProperties.java
```diff
@@ -36,6 +36,24 @@
...
+   * write to the daily index. Defaults to 5.
+   *
+   * <p>Corresponds to <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html">index.number_of_shards</a>
+   */
+  private int indexShards = 5;
+
+  /**
+   * Affects scalability by increasing the number of nodes in the Elasticsearch cluster that can
```
anuraaga: `The number of replica copies of each shard in the index. Each shard and its replicas are assigned to a machine in the cluster. Increasing the number of replicas and machines in the cluster will improve read performance and redundancy, but not write performance. Number of replicas can be changed for existing indices. Defaults to 1. It is highly discouraged to set this to 0 as it would mean a machine failure results in data loss.`

codefromthecrypt: thanks!

codefromthecrypt:
updated with better comments thx to @anuraaga

PR #1153, opened 2016-06-28
Fixes Cassandra indexes that lost traces in the same millisecond

### LINKED ISSUE #1142: The limit problem
zipkin-pic1
zipkin-pic2
hello,when i query in zipkin-ui(the data storage in cassandra), 
if the limit param is 10 ,it shows Showing: 6 of 6 ;
if the limit param is 20 ,it shows Showing: 12 of 12;
What is the problem?  Thank you!

codefromthecrypt:
A schema bug resulted in Cassandra not indexing more than bucket count
(10) trace ids per millisecond+search input. This manifested as less
traces retrieved by UI search or Api query than expected. For example,
if you had 1000 traces that happened on the same service in the same
millisecond, only 10 would return.

The indexes affected are `service_span_name_index`, `service_name_index`
and `annotations_index` and this was a schema-only change. Those with
existing zipkin installations should recreate these indexes to solve the
problem.

Fixes #1142

### DIFF
--- a/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraUtil.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraUtil.java
@@ -23,6 +23,7 @@ import java.nio.CharBuffer;
 import java.nio.charset.CharacterCodingException;
 import java.nio.charset.CharsetEncoder;
 import java.text.SimpleDateFormat;
+import java.util.Collections;
 import java.util.Date;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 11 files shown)


zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraUtil.java
```diff
@@ -100,6 +102,10 @@ static String iso8601(long timestamp) {
   }
 
   static List<String> annotationKeys(QueryRequest request) {
+    if (request.annotations.isEmpty() && request.binaryAnnotations.isEmpty()) {
+      return Collections.emptyList();
+    }
+    checkArgument(request.serviceName != null, "serviceName needed with annotation query");
```
codefromthecrypt, line 108: noticed this validation bug while writing tests

codefromthecrypt:
Releasing 1.2.1 which reverts this change due to https://github.com/openzipkin/zipkin/issues/1142#issuecomment-229254901

PR #1157, opened 2016-07-01
Extracts DeduplicatingExecutor, preventing redundant Cassandra indexing

codefromthecrypt:
The previous code had a mechanism to reduce writes to two indexes:
`service_name_index` and `service_span_name_index`. This mechanism would
prevent writing the same names multiple times. However, it is only
effective on a per-thread basis (as names were stored in thread locals).

In practice, this code is invoked at collection, and collectors have
many request threads per transport. By changing to a shared loading
cache, we can extend the deduplication to all threads. By extracting a
class to do this, we can test the edge cases and make it available for
future work, such as the other indexes.

See #1142

### DIFF
--- a/zipkin-storage/cassandra/src/test/java/zipkin/storage/cassandra/CassandraTestGraph.java
+++ b/zipkin-storage/cassandra/src/test/java/zipkin/storage/cassandra/CassandraTestGraph.java
@@ -20,11 +20,6 @@ import zipkin.internal.LazyCloseable;
 enum CassandraTestGraph {
   INSTANCE;
 
-  static {
-    // Ensure the repository's local cache of service names expire quickly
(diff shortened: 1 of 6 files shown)


codefromthecrypt:
ok this works and is thoroughly unit tested

zipkin-storage/cassandra/src/test/java/zipkin/storage/cassandra/CassandraTestGraph.java
```diff
@@ -20,11 +20,6 @@
 enum CassandraTestGraph {
   INSTANCE;
 
-  static {
-    // Ensure the repository's local cache of service names expire quickly
-    System.setProperty("zipkin.store.cassandra.internal.writtenNamesTtl", "1");
```
codefromthecrypt, line 25: cache is now cleared along with the keyspace during tests

PR #1163, opened 2016-07-04
Resurrects index skipping on core annotations in Cassandra

codefromthecrypt:
In the old scala implementation, we used to skip "core annotations" when
performing api queries, and we skipped them on custom indexes like Redis
or Cassandra. Skipping very common annotations like "cs" results in a
lot less storage operations, without affecting meaningful user queries.

### DIFF
--- a/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraUtil.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraUtil.java
@@ -22,15 +22,13 @@ import java.nio.ByteBuffer;
 import java.nio.CharBuffer;
 import java.nio.charset.CharacterCodingException;
 import java.nio.charset.CharsetEncoder;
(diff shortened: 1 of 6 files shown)


codefromthecrypt:
@danchia @eirslett @yurishkuro @michaelsembwever blast from the past

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraUtil.java
```diff
@@ -59,18 +57,6 @@ static ByteBuffer toByteBuffer(String string) throws CharacterCodingException {
     return UTF8_ENCODER.get().encode(CharBuffer.wrap(string));
   }
 
-  static final ThreadLocal<SimpleDateFormat> DATE_FORMATTER = new ThreadLocal<SimpleDateFormat>() {
```
codefromthecrypt, line 62: dead code removal as partial down payment on new code

PR #1167, opened 2016-07-06
Documents key indexes, affected by recent code and changes in-flight

codefromthecrypt:
This documents certain index tables used in Cassandra, particularly
those implicated in the more common queries.

### DIFF
--- a/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraStorage.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraStorage.java
@@ -220,11 +220,11 @@ public final class CassandraStorage
     for (String cf : ImmutableList.of(
         "traces",
         "dependencies",
-        "service_names",
-        "span_names",
-        "service_name_index",
-        "service_span_name_index",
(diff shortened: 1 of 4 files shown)


zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraStorage.java
```diff
@@ -220,11 +220,11 @@ public Session session() {
...
-        "span_names",
-        "service_name_index",
-        "service_span_name_index",
-        "annotations_index",
+        Tables.SERVICE_NAMES,
+        Tables.SPAN_NAMES,
+        Tables.SERVICE_NAME_INDEX,
+        Tables.SERVICE_SPAN_NAME_INDEX,
+        Tables.ANNOTATIONS_INDEX,
         "span_duration_index"
```
codefromthecrypt, line 228: not documenting this one, yet, as it has a whole lot of analysis pending per https://github.com/openzipkin/zipkin-java/issues/200

codefromthecrypt:
going to merge this as I'd like to do some cleanups on top. if there are any comments, happy to change the docs post commit

PR #1172, opened 2016-07-07
Adds cassandra index optimization

codefromthecrypt:
### What this does

This skips index operations that don't wouldn't improve api queries. It does this by only indexing when it would change the gap of a trace's time interval.

For example, a trace that fans out into N spans against the same service
will end up with a fixed amount of cassandra inserts as opposed to a
function of N.
### What this doesn't do

This doesn't affect duration indexing, as that's a tricky part of the code. This also doesn't improve anything for single-span traces (as there's no redundant indexing),
### How this is configured

You can configure this via the [CassandraStorage](https://github.com/openzipkin/zipkin/blob/master/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraStorage.java) or by environment variables (if using zipkin-server).
- `CASSANDRA_INDEX_CACHE_MAX`: Maximum trace index metadata entries to cache. Zero disables caching. Defaults to 100000.
- `CASSANDRA_INDEX_CACHE_TTL`: How many seconds to cache index metadata about a trace. Defaults to 60.
### When this works best

This particularly helps when traces with a lot of spans touch a small amount of services. It helps even more when the cardinality of annotation or binary annotation values are small (ex http.method = "GET").

This removes extra requests when a collector receives a large amount of span data for a single trace. That could be the case when one or more of the following are true:
- instrumentation bundles all local spans in a trace into the same message
- traces are routed consistently to a collector using this storage component
- there's only one collector
### What this costs

This costs memory, as it caches the state of index rows that this storage component wrote. Unless your annotations are very large, the overhead of default settings should be orders of tens of megabytes heap.
### Example

Here's results from running [brave's example](https://github.com/openzipkin/brave-resteasy-example)

```
2016-07-07 20:43:53.744 DEBUG 89424 --- [nio-9411-exec-1] zipkin.storage.cassandra.TraceIndexer    : optimized out 2/4 inserts into service_name_index
2016-07-07 20:43:53.745 DEBUG 89424 --- [nio-9411-exec-1] zipkin.storage.cassandra.TraceIndexer    : optimized out 2/4 inserts into service_span_name_index
2016-07-07 20:43:53.747 DEBUG 89424 --- [nio-9411-exec-1] zipkin.storage.cassandra.TraceIndexer    : optimized out 2/12 inserts into annotations_index
```

And here's results of self-tracing that (self-tracing uses a lot of local spans to indicate cassandra calls)

```
2016-07-07 20:43:54.127 DEBUG 89424 --- [pool-1-thread-1] zipkin.storage.cassandra.TraceIndexer    : optimized out 33/37 inserts into service_name_index
2016-07-07 20:43:54.130 DEBUG 89424 --- [pool-1-thread-1] zipkin.storage.cassandra.TraceIndexer    : optimized out 14/37 inserts into service_span_name_index
2016-07-07 20:43:54.136 DEBUG 89424 --- [pool-1-thread-1] zipkin.storage.cassandra.TraceIndexer    : optimized out 23/40 inserts into annotations_index
```

### DIFF
--- a/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/InsertTraceIdByAnnotation.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/InsertTraceIdByAnnotation.java
@@ -1,0 +1,59 @@
+/**
+ * Copyright 2015-2016 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin.storage.cassandra;
+
+import com.datastax.driver.core.BoundStatement;
+import com.datastax.driver.core.querybuilder.Insert;
+import com.datastax.driver.core.querybuilder.QueryBuilder;
+import java.nio.charset.CharacterCodingException;
+import java.util.Set;
+import java.util.concurrent.ThreadLocalRandom;
+import zipkin.Span;
+
+import static zipkin.storage.cassandra.CassandraUtil.annotationKeys;
+import static zipkin.storage.cassandra.CassandraUtil.toByteBuffer;
+
+// QueryRequest.annotations/binaryAnnotations
+final class InsertTraceIdByAnnotation implements Indexer.IndexSupport {
+  private static final ThreadLocalRandom RAND = ThreadLocalRandom.current();
+  private final int bucketCount;
+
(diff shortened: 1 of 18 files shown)


zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/InsertTraceIdByAnnotation.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Random;
+import java.util.Set;
+import zipkin.Span;
+
+import static zipkin.storage.cassandra.CassandraUtil.annotationKeys;
+import static zipkin.storage.cassandra.CassandraUtil.toByteBuffer;
+
+// QueryRequest.annotations/binaryAnnotations
+final class InsertTraceIdByAnnotation implements TraceIndexer.IndexSupport {
+  private static final Random RAND = new Random();
```
yurishkuro: java.util.concurrent.ThreadLocalRandom ?

codefromthecrypt: sure

codefromthecrypt:
OK I'm pretty done with this (and out of time). Unless there are other comments, I'll merge soon.

PR #1178, opened 2016-07-11
Adds Endpoint.ipv6 field with supporting changes to MySQL and UI

### LINKED ISSUE #306: IPv6 only hosts support
If I'm right host identifier in trace is currently named `ipv4` and encoded as `int32` and used as integer throughout the system (column types, etc). 
This automaticaly disables `ipv6` only hosts from being compatible with zipkin. Are there any plans or thoughts how one can add such support or maybe workaround can be found?

codefromthecrypt:
This adds `Endpoint.ipv6` as a fixed-width byte array (16 bytes). This
formalizes `Endpoint.ipv4 == 0` implying there's no ipv4 address.

In thrift, this remains a byte array (String) at field 4.
In json, this is normal string formatting.
- normal java utilities are used for codec (Inet6Address)

In the UI, the ipv6 address is preferred and bracketed when present.
Ex. [2001:db8:0:0:0:0:0:c002]:8080

In MySQL, this is mapped to a `BINARY(16)` field named
`zipkin_annotations.endpoint_ipv6`. When this column is missing a
warning like below is printed:

```
Jul 11, 2016 4:08:53 PM zipkin.storage.mysql.HasIpv6 compute
WARNING: zipkin_annotations.ipv6 doesn't exist, so Endpoint.ipv6 is not supported. Execute: alter table zipkin_annotations add `endpoint_ipv6` BINARY(16)
```

Fixes #306


codefromthecrypt:
cc @eirslett @abesto

zipkin/src/main/java/zipkin/internal/ThriftCodec.java
```diff
@@ -110,6 +110,7 @@ public Span readSpan(byte[] bytes) {
     final Field IPV4 = new Field(TYPE_I32, 1);
     final Field PORT = new Field(TYPE_I16, 2);
     final Field SERVICE_NAME = new Field(TYPE_STRING, 3);
+    final Field IPV6 = new Field(TYPE_STRING, 4);
```
yurishkuro, line 113: @adriancole I do not see a change in the thrift file. Are there unit tests verifying that this manual serialization is compatible with the native Thrift serialization done by classes generated from `.thrift` IDL file?

codefromthecrypt, line 113: First, I'll own any bug on this addition.

To answer your question, there aren't compiler (libthrift, scrooge, etc)
interop tests for thrift (or openapi interop tests for json) in this repo.

The thrifts live in a separate repo
https://github.com/openzipkin/zipkin-api/tree/master/thrift, and making
such tests would add maintenance due to factors including the race
condition of adding new fields like this. Such a thing would also be a
heavier build as we'd need to choose which generator (assuming apache maybe
0.93?), hook it up for the test source (or make a module ignored from
publishing etc), then design some tests that pass data through. That's
guaranteed to be a lot more to maintain in this repo than exists now.

Since we add fields very rarely, I'm not sure I want to add all of that
right now. Regardless, it would be a large enough change to warrant a
separate pull request.

codefromthecrypt, line 113: moved this to https://github.com/openzipkin/zipkin/issues/1179

codefromthecrypt, line 113: ah I think I understand why you asked! TType.STRING is used (because there's no TType.BINARY). the binary keyword in IDL is a hint, but the actual thrift type is string.

Ex. thrift makes this out of `4: optional binary ipv6`

``` java
  private static final org.apache.thrift.protocol.TField IPV6_FIELD_DESC = new org.apache.thrift.protocol.TField("ipv6", org.apache.thrift.protocol.TType.STRING, (short)4);
```

codefromthecrypt, line 113: found a way to cheaply help with this. added a commit

codefromthecrypt:
4.5K lines of this change is comparing w/libthrift (checking in generated files as opposed to introducing a build step)

PR #1180, opened 2016-07-12
Adds SCRIBE_ENABLED flag to control whether scribe starts or not

codefromthecrypt:
Scribe starts over a dozen threads using heap and nonheap memory.

By setting `SCRIBE_ENABLED=false`, users can recover those resources
for productive tasks.


codefromthecrypt:
This came up in troubleshooting with @liyichao. Basically, we can eliminate potential sources of memory problems by making the disable setting easier to find.

PR #1182, opened 2016-07-13
Adds SELF_TRACING_SAMPLE_RATE and fixes a C* self-tracing exception

codefromthecrypt:
Production sites should never run 100% self-tracing. This adds a
parameter to control the rate.

This also fixes a mistake where self-tracing could result in a state
exception. By logging instead, self-tracing is less dangerous in 
production.

### DIFF
--- a/zipkin-autoconfigure/storage-cassandra/src/main/java/zipkin/autoconfigure/storage/cassandra/brave/TracedSession.java
+++ b/zipkin-autoconfigure/storage-cassandra/src/main/java/zipkin/autoconfigure/storage/cassandra/brave/TracedSession.java
@@ -26,6 +26,7 @@ import com.github.kristofa.brave.ServerSpan;
 import com.github.kristofa.brave.ServerSpanThreadBinder;
 import com.github.kristofa.brave.SpanCollector;
 import com.github.kristofa.brave.SpanId;
+import com.google.common.collect.Maps;
 import com.google.common.reflect.AbstractInvocationHandler;
 import com.google.common.reflect.Reflection;
 import com.twitter.zipkin.gen.Annotation;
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
This was another subtle bug found by @liyichao!

```
2016-07-12 20:19:30.317  WARN 14973 --- [2-nio-worker-57] io.netty.channel.DefaultChannelPipeline  : An exception was thrown by a user handler's exceptionCaught() method while handling the following exception:

java.lang.IllegalStateException: select-trace-ids-by-span-duration not in the cache eventhough tracing is on
    at com.google.common.base.Preconditions.checkState(Preconditions.java:197) ~[guava-18.0.jar!/:na]
```

codefromthecrypt:
ps this still ends up a warning in the log, just not as severe as it isn't relegated to `exceptionCaught()` handler

codefromthecrypt:
tested with `SELF_TRACING_SAMPLE_RATE=0.01` (exact rate) and `SELF_TRACING_SAMPLE_RATE=0.001`(probabilistic rate)

zipkin-autoconfigure/storage-cassandra/src/main/java/zipkin/autoconfigure/storage/cassandra/brave/TracedSession.java
```diff
@@ -203,8 +202,12 @@ public String toString() {
 
     @Override public void addListener(Runnable listener, Executor executor) {
       delegate.addListener(() -> {
-        threadBinder.setCurrentSpan(currentSpan);
-        listener.run();
+        threadBinder.setCurrentSpan(parent);
+        try {
+          listener.run();
+        } finally {
+          threadBinder.setCurrentSpan(null);
```
codefromthecrypt, line 209: found this bug, too.

PR #1183, opened 2016-07-13
Makes types used in spark RDDs serializable

codefromthecrypt:
This makes Span and DependencyLink serializable. This uses thrift
encoding, as we've benchmarked it to be the fastest.

The approach used here is a rip-off of square wire's (protobuf library) approach https://github.com/square/wire/pull/192


codefromthecrypt:
@yurishkuro FYI

adriancole:
The approach used here is a rip-off of square wire's (protobuf library) approach https://github.com/square/wire/pull/192

yurishkuro:
lgtm

PR #1187, opened 2016-07-17
Allows dependency link jobs to be replayed

codefromthecrypt:
Before, the experimental Elasticsearch dependency link query would
double-count as there was no way to know if there were duplicates or
not. This does two things to address it:
- Sets the id of a dependency-link document to its parent-child
  - This makes documents overwritable in a bucket
- Moves summing of links from an aggregation query to client-side
  - This logic is simpler and matches the Cassandra implementation


codefromthecrypt:
cc @anuraaga ps I tried to get the aggregation query working, but somehow it was double-counting. I punted by doing client side, as at the moment less moving parts is helpful.

anuraaga:
Surprised by the double counting but anyways LGTM

PR #1188, opened 2016-07-17
Makes elasticsearch index buckets based on UTC time

codefromthecrypt:
We can't assume all collectors set their timezone to UTC, so this makes bucketing explicitly UTC.


codefromthecrypt:
cc @anuraaga another thing noticed :)

anuraaga:
LGTM Could have sworn I kept UTC in mind when writing it but maybe it was indeed all in the mind :)

codefromthecrypt:
thanks! real pull request coming in a minute in the spark repo

PR #1190, opened 2016-07-18
Uses default Elasticsearch port of 9300 instead of throwing exception

codefromthecrypt:
noticed this making a docker image


codefromthecrypt:
cc @anuraaga

anuraaga:
LGTM

PR #1191, opened 2016-07-18
Skips indexing of binary annotations longer than 256 characters

codefromthecrypt:
It is unlikely that someone has a valid use case to paste a key/value
match of a value longer than 256 characters. By limiting the length of
binary annotations indexed, we can reduce the size of indexes without
limiting users from storing larger values, like query data.

### DIFF
--- a/zipkin-storage/cassandra/README.md
+++ b/zipkin-storage/cassandra/README.md
@@ -42,8 +42,10 @@ Redundant requests to store service or span names are ignored for an hour to red
 Indexing of traces are optimized by default. This reduces writes to Cassandra at the cost of memory
 needed to cache state. This cache is tunable based on your typical trace duration and span count.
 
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
cc @openzipkin/cassandra

danchia:
That seems reasonable to me.
Maybe if someone really wants this in future, maybe we could index the hash of the really long value instead?

codefromthecrypt:
@michaelsembwever noted that the highest impact is more about the key count, not the length. Ex there's more benefit to reducing the (irrelevant) cardinality in the table vs the length.

@danchia for example, I'm not sure anyone would do a key lookup by md5 of a long url :)

This indirectly reduces count by stripping things that are presumably volatile from being long. Ex query parameters in a signature tend to be (but are not always) long, so stripping them on account of length, may very likely reduce the count as well.

Another option could be to make a blacklist, ex `Signature`. Pro is that this could weed out urls unlikley to be search keys. Problem is that this requires special knowledge of the namespace, and adds maintenance the zipkin project can't likely help with...

codefromthecrypt:
I'm going to merge this as I see more meh or in favor than against. thanks for the feedback!

PR #1196, opened 2016-07-20
Conditionally supports pre-aggregated dependencies in MySQL

codefromthecrypt:
This adds the table `zipkin_dependencies`, which is read when present
to return pre-aggregated dependencies (from the zipkin-dependencies job)

If there is no data in the `zipkin_dependencies` table, results will be
aggregated on-demand. This allows backwards compatibility for those who 
haven't added the new table, or run the spark job yet.


codefromthecrypt:
@jcarres-mdsol @marcingrzejszczak @shakuzen @eirslett need a second opinion on leaving the "compute on demand" around in mysql. Pro is that demos will show the aggregation without running any spark job. Con is that the data gets too big to return quickly.

zipkin-storage/mysql/src/main/java/zipkin/storage/mysql/HasDependencies.java
```diff
@@ -0,0 +1,62 @@
...
+
+  HasDependencies(DataSource datasource, DSLContexts context) {
+    this.datasource = datasource;
+    this.context = context;
+  }
+
+  @Override protected Boolean compute() {
+    try (Connection conn = datasource.getConnection()) {
+      DSLContext dsl = context.get(conn);
+      dsl.select(count()).from(ZIPKIN_DEPENDENCIES).fetchAny();
```
codefromthecrypt: For example, we could count this table, and if empty aggregate on-demand...

codefromthecrypt:
Current approach is to aggregate on demand only if a zipkin-dependencies job has never been run. This means the both have to be true...
- zipkin_dependencies table exists
- zipkin_dependencies table has data

jcarres-mdsol:
Sounds good. I think most serious users will aggregate the data as it gets unusable quite quickly to calculate dependencies realtime.
Maybe as you said still aggregate realtime is good for demos, if maintenance is real high, killing that code and displayed an empty graph if the spark job has not created the graph is not terrible either methinks.

codefromthecrypt:
ok this will end up as zipkin 1.5 after I make sure the dependencies job works fine, etc

PR #1199, opened 2016-07-23
Adding a TraceKeys constant 'jdbc.query'

ewhauser:


zipkin/src/main/java/zipkin/TraceKeys.java
```diff
@@ -99,6 +99,12 @@
    */
   public static final String HTTP_RESPONSE_SIZE = "http.response.size";
 
+  /**
+   * The query executed for JDBC call.  Ex. "select * from customers where id = ?"
```
codefromthecrypt: I know I suggested this.. :) but wondering if we shouldn't say "sql.query" instead of "jdbc.query"
That way, the annotation would be valid for non-java

codefromthecrypt: @jcarres-mdsol @basvanbeek @eirslett what say you? (asking for your non-java opinion)

basvanbeek: I agree with you that it should be `sql.query` as the data payload itself has more to do with `sql` than `jdbc`. With the added bonus that regardless of your database driver / engine you would annotate as `sql.query` payload for consistency. If needing to let know that JDBC is used i'd rather have a separate annotation which then could provide the used jdbc driver / version.

ewhauser:
@adriancole Agreed that's a better choice. Updated the PR.

codefromthecrypt:
thanks!

PR #1204, opened 2016-07-28
Refactor duration queries in Cassandra

michaelsembwever:
(from https://github.com/openzipkin/zipkin-java/issues/200 )

The following three tables can be merged into one.

```
service_name_index
service_span_name_index
span_duration_index
```

by using the following one table:

```
CREATE TABLE zipkin.service_span_name_index (
    service_name  text,      // service name
    span_name     text,      // span name, or blank for queries without span name
    bucket        int,       // time bucket, calculated as ts/interval (in microseconds), for some pre-configured
interval like 1 day.
    ts            timeuuid, // start timestamp of the span, truncated to millisecond precision
    trace_id      bigint,    // trace ID
    duration      bigint,    // span duration, in microseconds
    PRIMARY KEY ((service_name, span_name, bucket), ts)
)
```

Actual query constraints against a duration range would no longer be possible, but I suspect that doing this at run-time is not going to be worse than what we have today, especially if we can reduce so much writes, and put more read queries into the cache.


zipkin-storage/cassandra/src/main/resources/cassandra-schema-cql3.txt
```diff
@@ -66,15 +66,15 @@ CREATE TABLE IF NOT EXISTS zipkin.traces (
     WITH compaction = {'class': 'org.apache.cassandra.db.compaction.DateTieredCompactionStrategy', 'max_window_size_seconds': '86400'}
     AND default_time_to_live =  604800;
 
-CREATE TABLE IF NOT EXISTS zipkin.span_duration_index (
+CREATE TABLE zipkin.service_span_name_index (
```
codefromthecrypt, line 69: did you mean to drop IF NOT EXISTS?

zipkin-storage/cassandra/src/main/resources/cassandra-schema-cql3.txt
```diff
@@ -66,15 +66,15 @@ CREATE TABLE IF NOT EXISTS zipkin.traces (
...
     span_name     text,      // span name, or blank for queries without span name
-    bucket        int,       // time bucket, calculated as ts/interval (in microseconds), for some pre-configured interval like 1h.
+    bucket        int,       // time bucket, calculated as ts/interval (in microseconds), for some pre-configured interval like 1 day.
+    ts            timeuuid, // start timestamp of the span, truncated to millisecond precision
+    trace_id      bigint,    // trace ID
     duration      bigint,    // span duration, in microseconds
-    ts            timestamp, // start timestamp of the span, truncated to millisecond precision
-    trace_id      bigint,    // trace ID. Included as a clustering column to avoid clashes (however unlikely)
-    PRIMARY KEY ((service_name, span_name, bucket), duration, ts, trace_id)
+    PRIMARY KEY ((service_name, span_name, bucket), ts)
```
codefromthecrypt, line 76: trace id isn't here bc now we're timeuuid (just saying out loud)

zipkin-storage/cassandra/src/test/java/zipkin/storage/cassandra/CassandraSpanConsumerTest.java
```diff
@@ -96,8 +96,8 @@ public void skipsRedundantIndexingInATrace() {
     });
 
     accept(trace);
-    assertThat(rowCount(Tables.SERVICE_SPAN_NAME_INDEX)).isEqualTo(4L);
-    assertThat(rowCount(Tables.SERVICE_NAME_INDEX)).isEqualTo(4L);
+    assertThat(rowCount(Tables.SPAN_DURATION_INDEX)).isGreaterThanOrEqualTo(4L);
```
codefromthecrypt, line 99: if possible, I'd tighten this up, as 20 (in the later assertion) is also greaterThanOrEqualTo(4)

or put a TODO and I can look into it post merge

zipkin-storage/cassandra/src/test/java/zipkin/storage/cassandra/IndexerTest.java
```diff
@@ -1,87 +0,0 @@
...
-import com.google.common.collect.Maps;
-import java.util.concurrent.ConcurrentMap;
-import org.junit.Test;
-import zipkin.internal.Pair;
-
-import static org.assertj.guava.api.Assertions.assertThat;
-import static zipkin.storage.cassandra.Tables.SERVICE_NAME_INDEX;
-import static zipkin.storage.cassandra.Tables.SERVICE_SPAN_NAME_INDEX;
-
-public class IndexerTest {
```
codefromthecrypt, line 26: Indexer is still used for the annotation rows, right? If so, I'd leave the test in for now.

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CompositeIndexer.java
```diff
@@ -38,8 +38,6 @@
         CacheBuilder.from(spec).<PartitionKeyToTraceId, Pair<Long>>build().asMap();
     Indexer.Factory factory = new Indexer.Factory(session, indexTtl, sharedState);
     this.indexers = ImmutableSet.of(
-        factory.create(new InsertTraceIdByServiceName(bucketCount)),
-        factory.create(new InsertTraceIdBySpanName()),
         factory.create(new InsertTraceIdByAnnotation(bucketCount))
```
codefromthecrypt, line 41: since this makes the indexer thing really only used for a single table, now, you can put in a TODO for me to cleanup and simplify this

codefromthecrypt:
love the idea of merging these.. I think that leaves us with only 2 indexes (to write to), which makes maintenance far simpler. nice work!

yurishkuro:
> Actual query constraints against a duration range would no longer be possible

That was the reason duration index was introduced in the first place. My users often run queries where `duration > XXXms`, and the timeframe of the results it much less important than finding the actual long traces.

michaelsembwever:
> My users often run queries where duration > XXXms, and the timeframe of the results it much less important than finding the actual long traces.

@yurishkuro I don't think this will be a problem though, because queries are still implicitly constrained by a time range. Either way I suspect there's still a lot to read and filter out at runtime, the real benefit here is to reduce the write amplification and to improve kernel filesystem and cassandra key cache hits.

Further improvements beyond this tactic I think will need a look into either/both Materialized Views or SASI. SASI is an interesting one as it allows free-text searches on any column. But it is not available until Cassandra-3.4

yurishkuro:
>  I don't think this will be a problem though, because queries are still implicitly constrained by a time range.

Well, that's the problem, if my primary goal is finding long traces, applying a time range constraint first may produce no results. The duration index was duration constraint first, time range second.

michaelsembwever:
I so wish github had threaded comments…

> Well, that's the problem, if my primary goal is finding long traces, applying a time range constraint first may produce no results. The duration index was duration constraint first, time range second.

The implied time range is 3 days, since we don't keep more than that already.
How specific are your duration ranges? How many traces are already found within such?

yurishkuro:
Duration query usually would be D > Nms. The 3days time range has little impact on the results because of the LIMIT, which defaults to 100 in our installation. If a long span happens say every 10min, then at our traffic volume you'd need to run a limit of about 1,000,000 to find a matching span with desired duration.

michaelsembwever:
> Duration query usually would be D > Nms. The 3days time range has little impact on the results because of the LIMIT, which defaults to 100 in our installation. 

Not sure I understand.  There is no LIMIT applied in the code today.

> If a long span happens say every 10min, then at our traffic volume you'd need to run a limit of about 1,000,000 to find a matching span with desired duration.

Yes I can see this problem.
My questioning and concerns is that this seems overly tailored to one installation and the user  knowing a value of `D` so that `D > Nms` cuts the query down to a more manageable number of columns to read through.

If the default implied time-range is one day, then one third of the rows will need to be read.
If one third of traces took longer than `D` than one third of the rows will need to be read.

In your particular situation I'm curious about just how slow it's going to be to read through millions of columns. Along with trying to find a correct match between general UI behaviour and appropriate cql queries to be executing, I think this highlights the need to better stress test different scenarios and be working with some concrete numbers.

codefromthecrypt:
@yurishkuro while I see the goal, we neither document nor test the filter order you've mentioned. 

> if my primary goal is finding long traces, applying a time range constraint first may produce no results. The duration index was duration constraint first, time range second.

Right now, it is the first N spans (looking back from a timestamp) that breach the duration threshold.

Regardless of what we decide to do, we have to make it possible for this query to complete. Right now, it is a scar on zipkin as our "recommended" storage option includes a duration query that even if correct doesn't work. It is tiring to re-explain to every new cassandra user (that you cannot use the duration query). Its correctness only helps if it is usable.

codefromthecrypt:
by "doesn't work" I'm referring to the the voluminous requests made with with current bucketing scheme+query plan that results in a long chain of requests that cannot complete (from here https://github.com/openzipkin/zipkin-java/issues/200). I'd almost rather have no duration query feature vs a correct but unusable duration query feature

codefromthecrypt:
I opened this PR, which should buy us time to address this change in whatever way ends up best. By switching the default lookback to 1hr, the existing implementation should be performant enough for us to make a non-hasty decision here. https://github.com/openzipkin/zipkin/pull/1212

michaelsembwever:
Thanks @adriancole for buying us time here.

I agree with your sentiments.
And I see @yurishkuro's need. It's going round in my head how we tackle this properly.

I suspect the right way forward for us is SASI, even though that means such functionality is not available unless Cassandra is running 3.5.

In the short term with the lookback reduced to one hour I think we can go ahead with developing this patch.

@yurishkuro will you be happy if I at the same time at this patch also develop a patch for SASI which I believe is more appropriate for your situation? (Will you be able to upgrade your Cassandra to 3.5 before you next need to upgrade your Zipkin?)

yurishkuro:
@michaelsembwever don't hold on my account, we haven't even upgraded to the Java version of Zipkin, we're still running Scala build.

However, since you mentioned SASI, would it make sense to reimplement _all_ indexing of Zipkin spans via secondary indices? The existing manual indexing doesn't work well anyway if a query contains clauses that must be resolved by different manual index tables today.

michaelsembwever:
> However, since you mentioned SASI, would it make sense to reimplement all indexing of Zipkin spans via secondary indices? The existing manual indexing doesn't work well anyway if a query contains clauses that must be resolved by different manual index tables today.

Yes that's the idea. I'll look into it.

michaelsembwever:
Superseded by https://github.com/openzipkin/zipkin/pull/1252

codefromthecrypt:
I'd prefer the changes to the old code to stay separate from adding a new model. It will keep review easier as there are some unanswered things here that would be burdensome to move to #1252. Hopefully, the latter can just focus on adding the new model and therefore not need any changes from here.

codefromthecrypt:
cassandra3 is the answer to this. in other words, let's stop spending effort on the old one as when duration query is gone, it works well enough right now.

Issue #1210, opened 2016-07-29
is 7 days lookback a terrible default for the UI?


codefromthecrypt:
Something raised by @mansu was that the default interval of 7 days ago until now was troubling him. He believes that this should be a smaller value, like 1 day, so that he doesn't need to set a property like `-Dzipkin.ui.default-lookback=86400000`

I'm not sure what the actual problem is, since the data is limited to 10 results, so any traces now would be returned first. Perhaps it is a distraction? or maybe makes a query take longer?

Anyway, 1 day or even smaller sounds like a fine default for me.

cc others in the room, like @prat0318 @yurishkuro

codefromthecrypt:
incidentally, the performance of duration query in cassandra (as implemented today) would likely be far better if people didn't search 7 days by default.

codefromthecrypt:
I agree this was a terrible default (even if a good limit), particularly as it is the number 1 reason why cassandra duration queries are unreasonable. Setting to 1hr in https://github.com/openzipkin/zipkin/pull/1212

Issue #1185, opened 2016-07-14
Change dependencies view to a day range, not a time range


codefromthecrypt:
The spark job only works on daily buckets, and that's likely to be the way out for all storage services. It would be more clear if we chose day ranges as opposed to time ranges, I think.

cc @eirslett @virtuald @jcarres-mdsol @yurishkuro @travisby @anuraaga

codefromthecrypt:
by "only works on daily buckets" I mean that in cassandra and also elasticsearch, we store links in daily buckets. when we get to pre-aggregating mysql, it would be simplest to follow that pattern

adriancole:
ps I think the best would be to set the default day to yesterday. This will help ensure data is available

adriancole:
yesterday is semantically equivalent to today - 1 DAY lookback

jcarres-mdsol:
Good idea, I do see much value in using time-ranges smaller than one day

PR #1212, opened 2016-07-31
Decouples maximum lookback from default lookback

codefromthecrypt:
This does a number of things to improve performance.

QUERY_LOOKBACK prevents the server from having to construct queries that
are longer than indexes. This value was used by default which led to a
couple major problems.
- When looking for performance problems, you aren't usually looking for 7 day old data, and even if you were, it wouldn't fit into a limit of 10. Changing to a much smaller amount was requested in #1210
- Cassandra duration queries convert into lookback/1hr requests. A
  default of 7 days means 168 requests. This led to most users being
  unable to use the duration query function.
- By changing to 1hr, our 10 trace limit is very likely to apply, and in
  the case of Cassandra, we will only pull a single bucket.

In #1185, we know we want to change to daily granularity. By switching
lookback to 1 day (decoupled from the 1hr lookback on traces), we better
prepare for that change.

I also noticed that we document queryLimit to 10, but config.js says 15.
Regardless of the correct value, they should be aligned.

### DIFF
--- a/zipkin-server/README.md
+++ b/zipkin-server/README.md
@@ -83,7 +83,7 @@ When the UI loads, it reads default configuration from the `/config.json` endpoi
 Attribute | Property | Description
 --- | --- | ---
 environment | zipkin.ui.environment | The value here becomes a label in the top-right corner. Not required.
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
got a nod from @klingerf

defaults are a work in progress, but this answers a public complaint (about bad defaults) and a long-standing performance problem. merging

Issue #1208, opened 2016-07-29
how can i get zipkin version from zipkin-ui?


dragontree101:
how can i get zipkin version from zipkin-ui?
i update zipkin, but don't know how to get zipkin-ui version from web.

shakuzen:
@dragontree101 Are you running Zipkin server (the self-contained executable jar `zipkin.jar`), or serving the `zipkin-ui` some other way?
Also to clarify your question, you want to know what the current version number of a running Zipkin instance is, or are you looking for something else?

dragontree101:
@shakuzen 
i run newest version, use zipkin.jar  (zipkin-server-1.5.1-exec.jar),  
i just want to get current zipkin version from  web ui,  i have not found any version in web.

shakuzen:
@dragontree101 thank you for the quick response! I understand what you're expecting and I don't think there is any way to get that information currently. We should be able to add it easily enough though. I will submit a pull request for review with a potential solution tomorrow or the next day.

dragontree101:
@shakuzen thanks!

PR #1213, opened 2016-07-31
Add Zipkin version to the /info endpoint

shakuzen:
In order to easily identify the version deployed and running, this exposes the Zipkin version from the /info endpoint.

Related to #1208
/cc @dragontree101

Before:

``` bash
curl -s http://localhost:9411/info | jq .
{}
```

Since we can add arbitrary properties to expose on the `/info` endpoint through configuration, including Maven variables (with Maven resource filtering), this pull request does the following.

``` yaml
info:
  zipkin:
    version: "@project.version@"
```

Which would show up in the `/info` endpoint as:

``` json
{
  "zipkin": {
    "version": "1.6.1-SNAPSHOT"
  }
}
```

See [this section](http://docs.spring.io/spring-boot/docs/1.4.0.RELEASE/reference/htmlsingle/#production-ready-application-info) of the Spring Boot documentation regarding this and related feature.

### DIFF
--- a/zipkin-server/README.md
+++ b/zipkin-server/README.md
@@ -12,6 +12,7 @@ The following endpoints are defined for Zipkin:
 * /config.json - [Configuration for the UI](#configuration-for-the-ui)
 * /api/v1 - [Api](http://zipkin.io/zipkin-api/#/)
 * /health - Returns 200 status if OK
+* /info - Provides the version of the running instance
 * /metrics - Includes collector metrics broken down by transport type 
 
 There are more [built-in endpoints](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html) provided by Spring Boot, such as `/metrics`. To comprehensively list endpoints, `GET /mappings`.
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -200,6 +200,12 @@
   </profiles>
 
   <build>
+    <resources>
+      <resource>
+        <directory>src/main/resources</directory>
+        <filtering>true</filtering>
+      </resource>
+    </resources>
     <plugins>
       <plugin>
         <groupId>org.springframework.boot</groupId>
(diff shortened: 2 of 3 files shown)


codefromthecrypt:
I like the latter better. thanks for availing the choice!

shakuzen:
@adriancole I have updated the commit and the description of the pull request. Let me know if there is anything else to do before merging.

codefromthecrypt:
Maybe add something to the README of zipkin-server? under endpoints, you could mention that the /info one reports the version of zipkin you are running

dragontree101:
if possible， add a version tag to zipkin-server's web-ui?

shakuzen:
@dragontree101 Thank you for the quick feedback on this. It really helps to make these enhancements as useful as possible.

I discussed this with @adriancole a little bit. There is no clear cut place to add this to the UI now and if we were going to make such a change, I'd rather someone more familiar than me with UI coding (specifically Zipkin's and in general) to be directly involved. If you would like to see that, can you make a separate issue for that, possibly suggesting where you envision such information would go?

In the meantime, @adriancole had a great suggestion for a way you could accomplish this after the changes in this pull request are released. Zipkin already [supports specifying an environment string](https://github.com/openzipkin/zipkin/tree/master/zipkin-server#configuration-for-the-ui) that is displayed in the top right corner of the UI. You can include the version there using the property `info.zipkin.version`. So, for instance, you can run the following `java -jar zipkin.jar --zipkin.ui.environment=\${info.zipkin.version}` (you can, of course, also set the property in any of the other ways [supported by Spring Boot](https://docs.spring.io/spring-boot/docs/1.4.0.RELEASE/reference/html/boot-features-external-config.html))

I have tested this locally (you need the changes in this pull request to test). Let me know if that makes sense and is something that works for you.

codefromthecrypt:
PS technical difficulties on my side, but anyone with karma should merge
IMHO

PR #1217, opened 2016-08-07
Ensures dependencies can be derived when raw spans lack a timestamp

codefromthecrypt:


codefromthecrypt:
cc @travisby who noticed this.

PR #1236, opened 2016-08-14
Reduces the core jar from 272 to 188k by tuning codec and shade

codefromthecrypt:
Before, we used more reflection than necessary in codec. This implied
use of larger and more classes than we actually need. By slightly
changing codec, and manually controlling shade, we dropped the jar size
by 45%.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Check out the [`zipkin-server`](/zipkin-server) documentation for configuration 
 ## Core Library
 The [core library](https://github.com/openzipkin/zipkin/tree/master/zipkin/src/main/java/io/zipkin) requires minimum language level 7. While currently only used by the server, we expect this library to be used in native instrumentation as well.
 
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
Motivation for this is to lower barrier for tracers to include zipkin's standard library without resorting to extracting a separate "model jar"

README.md
```diff
@@ -26,7 +26,7 @@ Check out the [`zipkin-server`](/zipkin-server) documentation for configuration
 ## Core Library
 The [core library](https://github.com/openzipkin/zipkin/tree/master/zipkin/src/main/java/io/zipkin) requires minimum language level 7. While currently only used by the server, we expect this library to be used in native instrumentation as well.
 
-This includes built-in codec for both thrift and json structs. Direct dependencies on thrift or moshi (json library) are avoided by minifying and repackaging classes used. The result is a 256k jar which won't conflict with any library you use.
```
codefromthecrypt, line 29: lie.. we actually crept up to 272K! I wrote in 190, not 188, to give a little room :)

PR #1238, opened 2016-08-15
Adds span.timestamp_millis when storing in Elasticsearch

### LINKED ISSUE #1209: View Zipkin Data in Kibana as Event-based Data
Currently Zipkin stores the `timestamp` field in Elastsearch as a long value. It does so to preserve the microsecond precision of the timestamp value.

However as a side effect, `timestamp` cannot be indexed as a date/time field and cannot be used as a time field when creating an index view in Kibana.

I'm not sure how to fix this issue in such away as to preserve the microsecond precision and allow for date/time indexing. One idea was to create a new field (`timestamp_millis`?) that would contain just the milliseconds from the `timestamp` field.

codefromthecrypt:
In order to allow systems like Kibana to search by timestamp, we add a
field "timestamp_millis" when storing. The cheapest way to do this
without changing the codec is prefixing it to the json.

For example:
`{"traceId":"...` becomes `{"timestamp_millis":12345,"traceId":"...`

Fixes #1209

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/ElasticsearchSpanConsumer.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/ElasticsearchSpanConsumer.java
@@ -13,6 +13,7 @@
  */
 package zipkin.storage.elasticsearch;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.Functions;
 import com.google.common.util.concurrent.AsyncFunction;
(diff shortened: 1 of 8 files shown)


codefromthecrypt:
TODO: needs README and integration test (on the actual index)

also needs manual testing

zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/ElasticsearchSpanConsumer.java
```diff
@@ -73,14 +75,39 @@
...
+        .setSource(spanBytes);
+  }
+
+  /**
+   * In order to allow systems like Kibana to search by timestamp, we add a field "timestamp_millis"
+   * when storing. The cheapest way to do this without changing the codec is prefixing it to the
+   * json. For example. {"traceId":"... becomes {"timestamp_millis":12345,"traceId":"...
+   */
+  @VisibleForTesting
+  static byte[] prefixWithTimestampMillis(byte[] input, long timestampMillis) {
```
anuraaga, line 98: Heh - modifying the JSON was supposed to be the annoying part of this, but I didn't think of this reasonable(-ish) solution :)

codefromthecrypt, line 98: heavy on the ish :D

codefromthecrypt:
Manually tested with Kibana. timestamp_millis automagically showed up, and with no effort graphs worked.

screen shot 2016-08-16 at 10 38 45 am

codefromthecrypt:
added tests and docs

Issue #1221, opened 2016-08-10
publish javadoc and jdiff reports


codefromthecrypt:
_From @adriancole on May 30, 2016 0:4_

Currently, we package, but don't publish javadocs or jdiff reports
### javadocs

This is a bit of a problem because there's often good javadocs these days, but pointing to source isn't helpful as they aren't rendered. I'd prefer to be able to refer to javadoc links as opposed to source lines.
### jdiff

we don't "do jdiff" yet, but it is a tool to report api differences. Now that we are 1.0, we ought to be able to answer the question of which apis have changed or not.
### Example implementations

both of these tools are used in guava, and they setup a build step to publish them
https://github.com/google/guava/blob/65e6bd2efb54b9f6a94de0771db698cea8d2a1ca/util/update_snapshot_docs.sh

_Copied from original issue: openzipkin/zipkin-java#253_

codefromthecrypt:
cc @eirslett @kristofa @abesto

abesto:
Both sound like worthy causes. I'll have limited internet access over the weekend, also No Man's Sky was just released, but I'll look into this late next week (unless someone beats me to it).

codefromthecrypt:
I'd conserve your internet for No Man's Sky :)

abesto:
I'll have limited access to my PS4 as well, so that won't be an issue :)

codefromthecrypt:
jdiff I think is still a neat idea though not sure if anyone else feels like this :) If we did it, we should exclude the "internal" package and any package-private types and methods as they are inherently internal. I know some tools show changes to non-public symbols as drift..

If you want jdiff, upvote this? or maybe we should open a new issue? wdyt @jorgheymans

jorgheymans:
I can generate a few jdiff reports to see what it would look like. What packages / repositories should i include ?

codefromthecrypt:
first could be "io.zipkin.zipkin2:zipkin" and only the public non-internal types (and symbols inside them). I've struggled in the past getting jdiff to ignore internal things..

maybe after that we can roam to others?

jorgheymans:
having looked a bit at this, the simplest no-code way to generate a jdiff report between 2 arbitrary zipkin versions is (thanks to our friends over at https://github.com/zolyfarkas/spf4j/) : 

```
[...]

PR #1240, opened 2016-08-16
Upload javadocs to gh-pages

abesto:
Creates a directory structure that looks like this:

```
javadoc-builddir/
├── index.html
└── 1.6.1-SNAPSHOT
    ├── index.html
    ├── zipkin
    │   ├── META-INF
    │   │   └── MANIFEST.MF
    │   ├── allclasses-frame.html
    │   ├── allclasses-noframe.html
    │   ├── constant-values.html
    │   ├── deprecated-list.html
...
```

And makes the contents available via GitHub Pages. The URL is, somewhat counter-intuitively, http://zipkin.io/zipkin/ (where `zipkin.io` is the domain used by all GitHub Pages sites of the `openzipkin` org, and `/zipkin` is the name of this repo).

Uploads javadocs only when NOT on a pull request build.

Did some testing, and tried to avoid problems, but we might still hit problems around `git checkout`, `git commit`, and uncommitted files. If all else fails, we can always use a separate  clone in, say, `/tmp`.

(Also did some research; there doesn't seem to exist a Maven plugin for doing this.)

### DIFF
--- a/travis/publish.sh
+++ b/travis/publish.sh
@@ -72,8 +72,12 @@ check_release_tag() {
     fi
 }
 
+print_project_version() {
+  ./mvnw help:evaluate -N -Dexpression=project.version|grep -v '\['
+}
+
 is_release_commit() {
-  project_version=$(./mvnw help:evaluate -N -Dexpression=project.version|grep -v '\[')
+  project_version="$(print_project_version)"
   if [[ "$project_version" =~ ^[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+$ ]]; then
     echo "Build started by release commit $project_version. Will synchronize to maven central."
     return 0
@@ -100,6 +104,38 @@ safe_checkout_master() {
   fi
 }
 
+javadoc_to_gh_pages() {
+  version="$(print_project_version)"
+  rm -rf javadoc-builddir
+  builddir="javadoc-builddir/$version"
+
+  # Collect javadoc for all modules
+  for jar in $(find . -name "*${version}-javadoc.jar"); do
+    module="$(echo "$jar" | sed "s~.*/\(.*\)-${version}-javadoc.jar~\1~")"
+    this_builddir="$builddir/$module"
+    mkdir -p "$this_builddir"
+    unzip "$jar" -d "$this_builddir"
+    # Build a simple module-level index
+    echo "<li><a href=\"${module}/index.html\">${module}</a></li>" >> "${builddir}/index.html"
+  done
+
+  # Update gh-pages
+  git checkout gh-pages
+  rm -rf "$version"
+  mv "javadoc-builddir/$version" ./
+  rm -rf "javadoc-builddir"
+
+  # Update simple version-level index
+  if ! grep "$version" index.html 2>/dev/null; then
+    echo "<li><a href=\"${version}/index.html\">${version}</a></li>" >> index.html
+  fi
+
+  git add "$version"
+  git add index.html
+  git commit -m "Automatically updated javadocs for $version"
+  git push
+}
+
 #----------------------
 # MAIN
 #----------------------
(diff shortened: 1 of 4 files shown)


abesto:
Implementing (the simpler) half of https://github.com/openzipkin/zipkin/issues/1221. CC @adriancole

travis/publish.sh
```diff
@@ -122,11 +159,13 @@ elif is_travis_branch_master; then
...
   if is_release_commit; then
     ./mvnw --batch-mode -s ./.settings.xml -nsu -N io.zipkin.centralsync-maven-plugin:centralsync-maven-plugin:sync
+    javadoc_to_gh_pages
   fi
 
 # If we are on a release tag, the following will update any version references and push a version tag for deployment.
 elif build_started_by_tag; then
   safe_checkout_master
   ./mvnw --batch-mode -s ./.settings.xml -Prelease -nsu -DreleaseVersion="$(release_version)" -Darguments="-DskipTests" release:prepare
+  javadoc_to_gh_pages
```
codefromthecrypt, line 169: ah ok so this is for publishing against snapshots..

codefromthecrypt:
build failure unrelated.. rebase would fix it

codefromthecrypt:
maybe update RELEASE.md? Otherwise LGTM. Thanks for the help!

codefromthecrypt:
Also, probably putting a javadoc link reference to http://zipkin.io/zipkin/ at the bottom of the README with the other things.

abesto:
Rebased, added notes to `{RELEASE,README}.md`.

Happy to halp! In other news: just realized I can probably retroactively generate the javadocs for all previous releases, will see to that now.

codefromthecrypt:
Thanks tons!

abesto:
Done with the time traveling, http://zipkin.io/zipkin/ now has tons of yummy content. Built with https://github.com/openzipkin/zipkin/blob/gh-pages/rebuild-all-javadocs.sh

codefromthecrypt:
hit a snag on a release tag:

https://travis-ci.org/openzipkin/zipkin/builds/153257593

```
+javadoc_to_gh_pages
++print_project_version
++./mvnw help:evaluate -N -Dexpression=project.version
++grep -v '\['
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=192m; support was removed in 8.0
+version=1.7.1-SNAPSHOT
+rm -rf javadoc-builddir
+builddir=javadoc-builddir/1.7.1-SNAPSHOT
++find . -name '*1.7.1-SNAPSHOT-javadoc.jar'
+git checkout gh-pages
error: pathspec 'gh-pages' did not match any file(s) known to git.
The command "./travis/publish.sh" exited with 1.
store build cache
```

codefromthecrypt:
ps thanks for making the step last! wise

codefromthecrypt:
little red here, too.. https://travis-ci.org/openzipkin/zipkin/builds/153262205

abesto:
Oops. I was wondering why I didn't see any new entries. Checking.

abesto:
So! The problem is that the script finds and extracts all the javadoc jars of the right version, but it looks like we have two of some of them, for instance:
- ./zipkin-server/target/local-repo/io/zipkin/java/zipkin-autoconfigure-ui/1.7.0/zipkin-autoconfigure-ui-1.7.0-javadoc.jar
- ./zipkin-autoconfigure/ui/target/zipkin-autoconfigure-ui-1.7.0-javadoc.jar -d javadoc-builddir/1.7.0/zipkin-autoconfigure-ui

More importantly: is it safe to just ignore all but the first jar per module?

codefromthecrypt:
local-repo is for integration testing the jar you just made. so, you
can safely ignore that path.

PR #1245, opened 2016-08-17
Moves zipkin-core language level to JRE 6

codefromthecrypt:
This is an api compatible change to allow code sharing between zipkin
server and instrumentation, and without having to publish 2 jar files.

We accidentally broke the ability to make agent-based instrumentation by
compiling our model classes against Java 7 bytecode. Particularly, this
causes a regression in Brave.

To move back to Java 6 bytecode, we had to do a few things:
- replace moshi internal dependency with gson, which is java 6 compliant
- remove okio dependency and copy its Base64 class 
  - incidentally a class it also copied!
- do variable-size buffer writes w/ byte arrays instead of okio
- fix a few nags about diamond ops etc.

See https://github.com/openzipkin/brave/issues/225

### DIFF
--- a/zipkin/src/main/java/zipkin/Annotation.java
+++ b/zipkin/src/main/java/zipkin/Annotation.java
@@ -13,7 +13,6 @@
  */
 package zipkin;
 
-import zipkin.internal.JsonCodec;
 import zipkin.internal.Nullable;
 
 import static zipkin.internal.Util.checkNotNull;
@@ -101,11 +100,6 @@ public final class Annotation implements Comparable<Annotation> {
   }
 
   @Override
-  public String toString() {
-    return JsonCodec.ANNOTATION_ADAPTER.toJson(this);
-  }
-
-  @Override
   public boolean equals(Object o) {
     if (o == this) {
       return true;
@@ -135,7 +129,7 @@ public final class Annotation implements Comparable<Annotation> {
   @Override
   public int compareTo(Annotation that) {
     if (this == that) return 0;
-    int byTimestamp = Long.compare(timestamp, that.timestamp);
+    int byTimestamp = timestamp < that.timestamp ? -1 : timestamp == that.timestamp ? 0 : 1;
     if (byTimestamp != 0) return byTimestamp;
     return value.compareTo(that.value);
   }
--- a/README.md
+++ b/README.md
@@ -24,9 +24,9 @@ Once you've started, browse to http://your_host:9411 to find traces!
 Check out the [`zipkin-server`](/zipkin-server) documentation for configuration details, or [`docker-zipkin`](https://github.com/openzipkin/docker-zipkin) for how to use docker-compose.
 
 ## Core Library
-The [core library](https://github.com/openzipkin/zipkin/tree/master/zipkin/src/main/java/io/zipkin) requires minimum language level 7. While currently only used by the server, we expect this library to be used in native instrumentation as well.
+The [core library](https://github.com/openzipkin/zipkin/tree/master/zipkin/src/main/java/io/zipkin) is used by both Zipkin instrumentation and the Zipkin server. Its minimum Java language level is 6, in efforts to support those writing agent instrumentation.
 
-This includes built-in codec for both thrift and json structs. Direct dependencies on thrift or moshi (json library) are avoided by minifying and repackaging classes used. The result is a 190k jar which won't conflict with any library you use.
+This includes built-in codec for both thrift and json structs. A direct dependency on gson (json library) is avoided by minifying and repackaging classes used. The result is a 155k jar which won't conflict with any library you use.
 
 Ex.
 ```java
--- a/pom.xml
+++ b/pom.xml
@@ -51,7 +51,6 @@
 
     <brave.version>3.9.0</brave.version>
     <cassandra-driver-core.version>3.1.0</cassandra-driver-core.version>
-    <moshi.version>1.2.0</moshi.version>
     <okio.version>1.9.0</okio.version>
     <jooq.version>3.8.4</jooq.version>
     <spring-boot.version>1.4.0.RELEASE</spring-boot.version>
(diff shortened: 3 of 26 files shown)


codefromthecrypt:
cc @ivansenic

adriancole:
`SpanCodec.THRIFT` in zipkin-model is faster at writing than libthrift (and json), but significantly slower writing vs `Codec.THRIFT` (which uses okio buffers). Okio is minimum JRE 7, so I couldn't use it.

If write performance of zipkin-model's thrift codec is of issue, I could geek out and make it at least significantly better. In the mean time, bigger fish to fry I think.

```
Benchmark                                          Mode  Cnt     Score    Error   Units
CodecBenchmarks.readClientSpan_json               thrpt   15   100.430 ±  3.759  ops/ms
CodecBenchmarks.readClientSpan_thrift_core        thrpt   15   554.605 ± 29.872  ops/ms
CodecBenchmarks.readClientSpan_thrift_libthrift   thrpt   15   522.420 ± 40.683  ops/ms
CodecBenchmarks.readClientSpan_thrift_model       thrpt   15   552.305 ± 35.710  ops/ms
CodecBenchmarks.readLocalSpan_json                thrpt   15   158.819 ±  7.420  ops/ms
CodecBenchmarks.readLocalSpan_thrift_core         thrpt   15   872.564 ± 44.372  ops/ms
CodecBenchmarks.readLocalSpan_thrift_libthrift    thrpt   15   865.447 ± 43.493  ops/ms
CodecBenchmarks.readLocalSpan_thrift_model        thrpt   15   880.617 ± 39.463  ops/ms
CodecBenchmarks.readRpcSpan_json                  thrpt   15    40.085 ±  4.663  ops/ms
CodecBenchmarks.readRpcSpan_thrift_core           thrpt   15   202.961 ±  8.345  ops/ms
CodecBenchmarks.readRpcSpan_thrift_libthrift      thrpt   15   213.566 ±  9.165  ops/ms
CodecBenchmarks.readRpcSpan_thrift_model          thrpt   15   204.944 ±  7.969  ops/ms
CodecBenchmarks.readRpcV6Span_json                thrpt   15    42.304 ±  1.967  ops/ms
CodecBenchmarks.readRpcV6Span_thrift_core         thrpt   15   185.336 ±  7.524  ops/ms
CodecBenchmarks.readRpcV6Span_thrift_libthrift    thrpt   15   183.405 ±  5.702  ops/ms
CodecBenchmarks.readRpcV6Span_thrift_model        thrpt   15   188.565 ±  8.920  ops/ms
CodecBenchmarks.writeClientSpan_json              thrpt   15   108.915 ±  6.446  ops/ms
CodecBenchmarks.writeClientSpan_thrift_core       thrpt   15   832.911 ± 46.442  ops/ms
CodecBenchmarks.writeClientSpan_thrift_libthrift  thrpt   15   469.571 ± 17.487  ops/ms
CodecBenchmarks.writeClientSpan_thrift_model      thrpt   15   537.464 ± 18.230  ops/ms
CodecBenchmarks.writeLocalSpan_json               thrpt   15   201.776 ±  7.071  ops/ms
CodecBenchmarks.writeLocalSpan_thrift_core        thrpt   15  1239.498 ± 67.045  ops/ms
CodecBenchmarks.writeLocalSpan_thrift_libthrift   thrpt   15   765.944 ± 41.224  ops/ms
CodecBenchmarks.writeLocalSpan_thrift_model       thrpt   15   866.032 ± 43.732  ops/ms
CodecBenchmarks.writeRpcSpan_json                 thrpt   15    37.635 ±  1.560  ops/ms
CodecBenchmarks.writeRpcSpan_thrift_core          thrpt   15   276.072 ± 17.234  ops/ms
CodecBenchmarks.writeRpcSpan_thrift_libthrift     thrpt   15   180.952 ±  9.838  ops/ms
CodecBenchmarks.writeRpcSpan_thrift_model         thrpt   15   204.083 ± 16.670  ops/ms
CodecBenchmarks.writeRpcV6Span_json               thrpt   15    31.568 ±  2.500  ops/ms
CodecBenchmarks.writeRpcV6Span_thrift_core        thrpt   15   261.056 ±  8.964  ops/ms
CodecBenchmarks.writeRpcV6Span_thrift_libthrift   thrpt   15   168.900 ±  6.065  ops/ms
CodecBenchmarks.writeRpcV6Span_thrift_model       thrpt   15   189.057 ±  9.052  ops/ms
```

adriancole:
going to do this a different way.. key part is finding performant libraries that still build against java6. okio and moshi build against java7

eirslett:
How about we make a Java 6-compliant fork of the instrumentation library? zipkin-java6? Separate repo? Supporting Java 5 or 6 is an additional burden to contributors; it could be trivial, it could be nontrivial (I'm not sure), but things like okio/moshi only working with Java 7+ is a problem.

Maintaining the fork would be on a best-effort basis, possibly driven by Zipkin users who still use Java 5/6?

adriancole:
nearly done with some rejigging which significantly lowers the impact I
think..

codefromthecrypt:
@eirslett so found that gson is still actively supporting java 6, so wasn't bad after using that instead of moshi.

zipkin/src/main/java/zipkin/Annotation.java
```diff
@@ -101,11 +100,6 @@ public Annotation build() {
   }
 
   @Override
-  public String toString() {
-    return JsonCodec.ANNOTATION_ADAPTER.toJson(this);
```
codefromthecrypt, line 105: removed toStrings on leave nodes while leaving json toString on entrypoint ones (Span and DependencyLink). This reduces the amount of internal class leak

codefromthecrypt:
OK.. done toying around. The new implementation uses gson instead of moshi, and a normal buffer instead of okio.

I ran benchmarks before and after:
- Thrift reads were unaffected. Thrift writes (which used to use okio), are slightly faster with small spans and the same with spans with several annotations.
- Json reads (which now use gson instead of moshi) are 20-35% faster, but writes are about half as fast.

While I'm sure I'll nerd out on json writes, in the future, I think this performance is fine. Those concerned with performance should be using thrift anyway, as it is an order of magnitude faster than json before and after this change.

PR #1246, opened 2016-08-19
[zipkin-ui] Fix install-node-and-npm

abesto:
As discussed in
https://github.com/eirslett/frontend-maven-plugin/issues/294


abesto:
cc @eirslett

PR #1247, opened 2016-08-20
Adds Codec.gather for use in reporters

codefromthecrypt:
This adds `Codec.gather`, which allows you to gather a list of byte
arrays into a single encoded message.

Reporters like HTrace dequeue spans reported into a list of byte arrays.

This allows insight into the size of message before you send it. For
example, you can pop until you have up to 1MB of spans as opposed to
100 spans (which might end up as 100MB!).

Note this is slightly less efficient than writing a list directly, so
it is special-cased for the reporting use case, where spans are encoded
in a separate scope than sent.

```
Benchmark                                       Mode  Cnt  Score   Error   Units
CodecBenchmarks.write100Spans_json             thrpt   15  0.539 ± 0.009  ops/ms
CodecBenchmarks.write100Spans_thrift           thrpt   15  8.315 ± 0.377  ops/ms
CodecBenchmarks.writeAndGather100Spans_json    thrpt   15  0.549 ± 0.021  ops/ms
CodecBenchmarks.writeAndGather100Spans_thrift  thrpt   15  8.964 ± 0.163  ops/ms
```

See https://github.com/openzipkin/zipkin-reporter-java/issues/1

### DIFF
--- a/benchmarks/src/main/java/zipkin/benchmarks/CodecBenchmarks.java
+++ b/benchmarks/src/main/java/zipkin/benchmarks/CodecBenchmarks.java
@@ -17,6 +17,8 @@ import com.google.common.io.ByteStreams;
 import com.twitter.zipkin.thriftjava.Annotation;
 import com.twitter.zipkin.thriftjava.BinaryAnnotation;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 import org.apache.thrift.TDeserializer;
 import org.apache.thrift.TException;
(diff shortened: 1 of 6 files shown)


codefromthecrypt:
@nicmunroe you might also be interested in this option. Especially when using a fast encoder like thrift (~1us to encode a span), spans could be encoded on the calling thread, or you could incrementally decode spans coming off a queue until a certain size threshold where you send them.

codefromthecrypt:
merged into https://github.com/openzipkin/zipkin-reporter-java/pull/6

Issue #1237, opened 2016-08-14
Consider moving off travis, or putting up a redundant job on CircleCI


codefromthecrypt:
Travis gives free builds, so we certainly have to thank them for that. That said, if there's competition, we might be better off evaluating it.

For example, the competition might..
- actually update things (vs indefinitely lagging on c\* https://github.com/travis-ci/travis-ci/issues/3952)
- not build tags by default (https://github.com/travis-ci/travis-ci/issues/1532)
- not black-hole builds for 1.5hrs with no status update
- have less severe outages (https://www.traviscistatus.com/incidents/11hp8bhkrkn7)
- ... build faster!

We do use travis features, such as credential encryption. This allows us to do tag-driven deploys. We also use docker. When considering an alternative, we should ensure we have feature parity. That said, travis has been so bad lately, we might want to throw up another build checker just to eliminate false-negatives.

cc the interested.. @abesto @shakuzen @marcingrzejszczak (your_name_here)

-- example black hole from today
![screen shot 2016-08-14 at 10 25 05 pm](https://cloud.githubusercontent.com/assets/64215/17649762/fbe08d78-626e-11e6-9e2a-63e93352d9d5.png)

codefromthecrypt:
I rewrote a cassandra test that was timing out to use mocks and the build went green. I'm assuming that we're bumping limits of our slave, and that becomes more obvious when travis is degraded.

abesto:
The big draw of Circle CI (for me at least) is running tests in parallel. In theory that should allow us to significantly cut down on build times. Don't know about reliability; if nothing else, a redundant system sounds like a reasonable idea (don't forget about the increased maintenance burden though). They also offer a nice packag for OSS projects (see bottom of https://circleci.com/pricing/)

codefromthecrypt:
Had to rewrite another test, mainly as something in the travis image ended up consistently changing a failure case behavior. Note that this isn't something we'd necessarily avoid in circleci.

abesto:
Played around a bit with CircleCI. Set up test dependencies, though there's probably something messed up around MySQL because those tests fail with MySQL-related errors. I've also set up parallel test runs, with support for the time-based auto-balancing of CircleCI. The build https://circleci.com/gh/openzipkin/zipkin/24 took 9:23 (compare to the ~14 minutes of Travis); though maybe the failing tests skew things somewhat. There's nothing around publishing wired in yet, though it should be pretty easy to reuse the existing logic.

[...]

PR #1248, opened 2016-08-20
Build, publish using CircleCI

abesto:
_openzipkin/zipkin on CircleCI_: https://circleci.com/gh/openzipkin/zipkin

Why: https://github.com/openzipkin/zipkin/issues/1237
## Points of interest
- Support for the test-runtime-based balancing of tests CircleCI does for running tests in parallel
- Caching the `node_modules` dir of `zipkin-ui` saves around 90 seconds
- Overall build time is around 6-7 minutes (about half of that on Travis)
- Release process drafted in #1250, still needs thinking.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -41,6 +41,7 @@ dependencies:
 database:
   override:
     - mysql -u ubuntu -e 'SET GLOBAL innodb_file_format=Barracuda'
+    - mysql -u ubuntu -e "SET GLOBAL sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''))"
     - mysql -u ubuntu circle_test < zipkin-storage/mysql/src/main/resources/mysql.sql
 
 test:
--- a/circle.yml
+++ b/circle.yml
@@ -54,3 +54,7 @@ test:
     - mkdir -p $CIRCLE_TEST_REPORTS/junit/
     - find . -type f -regex ".*/target/surefire-reports/.*xml" -exec cp {} $CIRCLE_TEST_REPORTS/junit/ \;
 
+# Send notifications to Gitter
+notify:
+  webhooks:
+    - url: https://webhooks.gitter.im/e/e63be97ca10d2fbd09c3
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,55 @@
+#
+# Copyright 2015-2016 The OpenZipkin Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+# in compliance with the License. You may obtain a copy of the License at
+#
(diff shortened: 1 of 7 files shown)


abesto:
CC @adriancole @shakuzen @marcingrzejszczak @eirslett

circleci/publish-stable.sh
```diff
@@ -0,0 +1,56 @@
...
+#     This commit will not trigger a separate build due to the [skip ci] in the
+#     commit message.
+./mvnw --batch-mode -s ./.settings.xml -Prelease -nsu -pl -:benchmarks \
+       -DskipTests -DscmCommentPrefix='[maven-release-plugin][skip ci] ' \
+       deploy
+
+# And this syncs it to Maven Central. Note: this needs to be done once per
+# project, not module, hence -N
+./mvnw --batch-mode -s ./.settings.xml -nsu -N \
+       io.zipkin.centralsync-maven-plugin:centralsync-maven-plugin:sync
```
abesto: AFAICT on Travis, this runs on the commit with the non-SNAPSHOT version. But here by the time we get to this point, we're on version (release + 0.0.1)-SNAPSHOT. I suspect we'll need to expose a `sync.version` property so we can pass in the version to sync. @adriancole WDYT?

codefromthecrypt:
Thanks for putting this together. I've a head-start on this as toyed a
little last couple days. Here's one thought..

separate out the generic changes that make tests not flake into a commit.
separate out the non-publish changes into a separate commit (ex ones that
make circle work)
add a new PR for the publishing things.. (or separate ^^ into a different
PR and keep this one alive)

Thought process is that we can merge having dual pull request checkers
(existing one and now circleci). We could then carry on a separate
discussion around the publish oriented things, including the more crude
credential tooling etc.

We didn't find travis to be a problem until time passed. My overall concern
is quickly moving without experience in CircleCI, we end up worse than
travis. One way we could address this is have pull request checkers here,
but use a less critical repo for guinea-pig testing release (ex
zipkin-dependencies, which doesn't change that often)

Thoughts?

marcingrzejszczak:
That's a good idea. In Spring Cloud we've basically only done simple builds in Travis so our transition could have been extreme (create a CircleCI build and disable Travis ASAP). Here the process can be more gradual definitely.

abesto:
I like this approach. Will separate out publishing into a separate PR, and squash this into two commits (one the CircleCI setup, the other changes to Zipkin code for the Circle env).

Side question: the flakes that we had on Circle, are those things our users would experience in their (non-test) environments?

codefromthecrypt:
> Side question: the flakes that we had on Circle, are those things our
> users would experience in their (non-test) environments?
> 
> The flake fixes for things that happen in travis and circle might be OS
> specific or resource specific. OSX laptop builds don't fail for me often.

PR #1249, opened 2016-08-20
Improve javadocs publish

abesto:
As discussed in #1240. Plus a little bit of added goodness.


codefromthecrypt:
oops should have merged this :) thanks!

PR #1259, opened 2016-08-30
Cuts thrift encoding overhead by allocating the correct sized array

codefromthecrypt:
This computes the serialized length prior to serializing thrifts. Then,
a buffer of exactly the right size is allocated, avoiding growth via
rebuffering. The resulting array can be returned directly as it is the
correct length.

This is the last low-hanging fruit of thrift write optimization, and
results in ~4x higher efficiency than typical libthrift usage.

```
Benchmark                                         Mode  Cnt  Score   Error  Units
CodecBenchmarks.writeClientSpan_json_zipkin       avgt   15  17.165 ± 0.785  us/op
CodecBenchmarks.writeClientSpan_thrift_libthrift  avgt   15   2.112 ± 0.073  us/op
CodecBenchmarks.writeClientSpan_thrift_zipkin     avgt   15   0.488 ± 0.026  us/op
```

### DIFF
--- a/zipkin/src/main/java/zipkin/internal/Buffer.java
+++ b/zipkin/src/main/java/zipkin/internal/Buffer.java
@@ -19,8 +19,16 @@ import java.util.Arrays;
 /** Similar to {@link java.io.ByteArrayInputStream}, except specialized and unsynchronized */
 final class Buffer extends OutputStream {
   static final int MAX_ARRAY_LENGTH = Integer.MAX_VALUE - 8;
-  private byte[] buf = new byte[128];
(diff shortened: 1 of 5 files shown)


eirslett:
Could this be merged upstream to libthrift as well? Or is it too Zipkin-specific?

codefromthecrypt:
libthrift doesn't have a "sizeOf" feature, so doing this automatically
would need a change to the compiler. This might be harder to do when you
have an open set of fields, though not impossible. Filed an issue
regardless https://issues.apache.org/jira/browse/THRIFT-3913

PR #1264, opened 2016-08-31
Leniently reads span.annotations and span.binaryAnnotations in UI

### LINKED ISSUE #1263: Uncaught TypeError: Cannot read property 'map' of undefined
Zipkin-1.8.0 UI page will throw javascript exception when any span missing a `binaryAnnotations` field: `Uncaught TypeError: Cannot read property 'map' of undefined`.

Maybe do some null check [here](https://github.com/openzipkin/zipkin/blob/master/zipkin-ui/js/component_ui/traceSummary.js#L10)?

``` javascript
// js/component_ui/traceSummary.js
function endpointsForSpan(span) {
  return _.union(
    span.annotations.map(a => a.endpoint),
    span.binaryAnnotations.map(a => a.endpoint)
  ).
```

codefromthecrypt:
Before, only some branches of code guarded reads to `span.annotations`
and `span.binaryAnnotations`. This guards the rest.

The impact is that we now tolerate json that omits the array fields on
empty.

Fixes #1263

### DIFF
--- a/zipkin-ui/js/component_ui/spanPanel.js
+++ b/zipkin-ui/js/component_ui/spanPanel.js
@@ -15,7 +15,7 @@ export default component(function spanPanel() {
     this.$node.find('.service-names').text(span.serviceNames);
 
     const $annoBody = this.$node.find('#annotations tbody').text('');
-    $.each(span.annotations, (i, anno) => {
+    $.each((span.annotations || []), (i, anno) => {
       const $row = self.$annotationTemplate.clone();
       if (anno.value === Constants.ERROR) {
         $row.addClass('anno-error-transient');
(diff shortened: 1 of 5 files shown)


codefromthecrypt:
also manually tested with the following spans, then went into the UI to check..

``` bash
$ curl -s localhost:9411/api/v1/spans -H'Content-Type: application/json' -d '[{"traceId":"1","name":"no-annotations","id":"1","timestamp":1234,"duration":100,"binaryAnnotations":[{"key":"lc","value":"bamm-bamm","endpoint":{"serviceName":"flintstones","ipv4":"127.0.0.1"}}]}]'
$ curl -s localhost:9411/api/v1/spans -H'Content-Type: application/json' -d '[{"traceId":"2","name":"no-binary-annotations","id":"2","timestamp":1234,"duration":100,"annotations":[{"timestamp":1234,"value":"sr","endpoint":{"serviceName":"flintstones","ipv4":"127.0.0.1"}}]}]'
```

PR #1267, opened 2016-09-04
Fix: pathspec 'gh-pages' did not match any file(s) known to git

abesto:
As seen at https://travis-ci.org/openzipkin/zipkin/builds/156529896, javadoc publishing is currently broken. This is an attempt to fix it.


codefromthecrypt:
Thanks! meant to ping you about this.

PR #1269, opened 2016-09-05
Fixes regression writing I64 binary annotations

codefromthecrypt:
Before, we special-cased writing numbers, except we had no tests to
cover I64 binary annotations. Eventhough these are niche, they should
work!

### DIFF
--- a/zipkin/src/main/java/zipkin/internal/Buffer.java
+++ b/zipkin/src/main/java/zipkin/internal/Buffer.java
@@ -13,8 +13,6 @@
  */
 package zipkin.internal;
 
-import static zipkin.internal.Util.checkArgument;
-
 final class Buffer {
   interface Writer<T> {
     int sizeInBytes(T value);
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
cc @tristanpenman

tristanpenman:
This is definitely an improvement. I can now access the traces via the UI.

Now there is one issue that I'm observing, but I'm not yet sure if it something about the demo code (which I modified slightly) but the annotated values that I see in the UI seem to have different trailing digits to those reported by the demo.

e.g.
179750381162269401 (generated long) -> 179750381162269400 (UI)
-7437073205494488636 (generated long) -> -7437073205494489000 (UI)

So it looks like the values might be getting truncated, and then filled in with random digits, somewhere.

codefromthecrypt:
@tristanpenman interesting.. I wonder if this is a javascript number precision issue?

codefromthecrypt:
I mean strictly speaking we can't fit a I64 into a javascript (JSON) number
(even if we can leniently read it in java). I'd probably change the demo to
do something less confusing, such as generate a UUID and store as a text
binary annotation, or write the long as a string.

codefromthecrypt:
ps when you click the JSON button on the trace detail screen.. do the
numbers look truncated there? https://gitter.im/openzipkin/zipkin

tristanpenman:
Thanks for the tip about the JSON button. That view shows the same number that is reported by the UI. So assuming it is the raw trace data received from the demo, your explanation makes perfect sense.

PR #1273, opened 2016-09-05
Handles RFC 7159 json escaping

codefromthecrypt:
We accept data that might violate json escaping rules. This escapes
any UTF-8 String we accept, even binary annotations.

Fixes #1272

PR #1275, opened 2016-09-05
Adds missing patch for #1273

codefromthecrypt:


zipkin/src/main/java/zipkin/internal/JsonCodec.java
```diff
@@ -278,7 +278,7 @@ public BinaryAnnotation fromJson(JsonReader reader) throws IOException {
           b.writeAscii(value.value[0] == 1 ? "true" : "false");
           break;
         case STRING:
-          b.writeByte('"').write(value.value).writeByte('"');
+          b.writeByte('"').writeJsonEscaped(value.value).writeByte('"');
```
codefromthecrypt, line 281: this was the only field that wasn't guarded in the 1.8.3

PR #1282, opened 2016-09-09
Only write good span timestamp guesses at ingest

### LINKED ISSUE #1281: Stop adding timestamp and duration to spans at ingest
ApplyTimestampAndDuration was a stop-gap measure created to allow a duration query on spans created by old instrumentation. This causes problems, such as #1280, when server spans arrive before client ones.

The correct approach to the timestamp duration problem is to have instrumentation projects update themselves to support these fields, as opposed to guessing incorrectly at ingest time. Here are docs on how to do this https://github.com/openzipkin/openzipkin.github.io/pull/46

codefromthecrypt:
Previously, Zipkin wrote the wrong timestamp when receiving spans that
are missing timestamp data. It is still the case that instrumentation
libraries aren't updated to add this. Until then, we have to guess a
little more smartly so as to not record incorrect timestamps.

This does the following when Span.timestamp is missing:
- If there is a "cs", use that (because client always is authoritative)
- If there is an "sr" in a root span, use that
- Otherwise, return null instead of choosing an arbitrary low value.

This also is more careful when updating timestamps, for example, in
MySQL, we only update if the span's timestamp is set or on client send.

This change has some impact, notably that duration queries now require
instrumentation to update to actually set Span.duration. Also,
elasticsearch "timestamp_millis" field is now used directly for queries.

To facilitate better instrumentation, I've added more documentation:
https://github.com/openzipkin/openzipkin.github.io/pull/46

Fixes #1281

### DIFF
--- a/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
@@ -33,12 +33,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import zipkin.Codec;
 import zipkin.Span;
-import zipkin.internal.ApplyTimestampAndDuration;
 import zipkin.internal.Nullable;
 import zipkin.internal.Pair;
 import zipkin.storage.guava.GuavaSpanConsumer;
 
 import static com.google.common.util.concurrent.Futures.transform;
+import static zipkin.internal.ApplyTimestampAndDuration.guessTimestamp;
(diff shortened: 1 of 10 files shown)


rogeralsing:
> If there is a "cs", use that (because client always is authoritative)

Does this mean that if client sends a request at t0, and the server starts receiving at t100, then the span will be rendered from t0?

codefromthecrypt:
yep

codefromthecrypt:
one broken test.. looking into it.

codefromthecrypt:
ok I think I got the fix now.

codefromthecrypt:
alright I'll plan on merging this and cutting a release tomorrow morning my time (+~12hrs) unless someone says don't.

PR #1283, opened 2016-09-09
Removes duration query support from Cassandra 2.2

### LINKED ISSUE #1224: Refactor duration query in cassandra
_From @adriancole on May 6, 2016 0:59_

The duration query is the most expensive query in cassandra, as it turns into 1 request per hour of lookback, limited to indexTtl. The UI's default lookback is a day, so this means 24 requests to the backend. If someone sets the ui to an early time, it limits at index TTL, which defaults to 7 days or 168 requests!

We should look into alternatives, perhaps doing a "peek" at the first bucket and only proceeding to fan out if that doesn't meet the response limit. Since the response limit is often 10, it might actually be fine.

See #199 

cc @eirslett @yurishkuro @danchia

_Copied from original issue: openzipkin/zipkin-java#200_

codefromthecrypt:
The duration query is unusable for most people, yet adds load every time
a span is written. It is also complicated code that hasn't improved much
since being written.

This removes this feature, punting to cassandra3 which can address the
problem more effectively via SASI indexing.

Fixes #1224
Obviates #1204

### DIFF
--- a/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
@@ -40,7 +40,6 @@ import zipkin.storage.guava.GuavaSpanConsumer;
 
 import static com.google.common.util.concurrent.Futures.transform;
 import static zipkin.storage.cassandra.CassandraUtil.bindWithName;
-import static zipkin.storage.cassandra.CassandraUtil.durationIndexBucket;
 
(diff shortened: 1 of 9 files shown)


codefromthecrypt:
@openzipkin/cassandra I applaud the efforts, but this feature has been non-stop support, and most people can't use it. Let's move on and put efforts behind cassandra3 instead?

danchia:
+1 from me

codefromthecrypt:
merging due to the history of this issue, comments here and also an out-of-band thumbs up from @michaelsembwever a few days ago.

PR #1289, opened 2016-09-11
 Store annotations in the indexed text column individually

michaelsembwever:
- store annotations in the indexed text column individually, providing better search usability in the UI, and less disk space used,
- remove fetch size to statements. that's not needed against this datamodel.

### DIFF
--- a/zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraUtil.java
+++ b/zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraUtil.java
@@ -21,6 +21,7 @@ import com.google.common.collect.Sets;
 import java.math.BigInteger;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
@@ -73,19 +74,20 @@ final class CassandraUtil {
       if (Constants.CORE_ANNOTATIONS.contains(a.value)) continue;
 
       if (a.endpoint != null && !a.endpoint.serviceName.isEmpty()) {
-        annotationKeys.add(a.endpoint.serviceName + ":" + a.value);
+        annotationKeys.add(a.endpoint.serviceName);
+        annotationKeys.add(a.value);
       }
     }
     for (BinaryAnnotation b : span.binaryAnnotations) {
-      if (b.type == BinaryAnnotation.Type.STRING
-          && b.endpoint != null
-          && !b.endpoint.serviceName.isEmpty()
-          && b.value.length <= LONGEST_VALUE_TO_INDEX * 4) { // UTF_8 is up to 4bytes/char
-        String value = new String(b.value, UTF_8);
-        if (value.length() > LONGEST_VALUE_TO_INDEX) continue;
-
-        annotationKeys.add(b.endpoint.serviceName + ":" + b.key);
-        annotationKeys.add(b.endpoint.serviceName + ":" + b.key + ":" + new String(b.value, UTF_8));
+      if (b.endpoint != null && !b.endpoint.serviceName.isEmpty() && !Constants.CORE_ANNOTATIONS.contains(b.key)) {
+        annotationKeys.add(b.endpoint.serviceName);
+        if (b.type == BinaryAnnotation.Type.STRING) {
+            String value = new String(b.value, UTF_8);
+            value = value.substring(0, Math.min(value.length(), LONGEST_VALUE_TO_INDEX));
+            annotationKeys.add(b.key + ":" + value);
+        } else {
+            annotationKeys.add(b.key);
+        }
       }
     }
     return annotationKeys;
@@ -96,12 +98,13 @@ final class CassandraUtil {
       return Collections.emptyList();
     }
     checkArgument(request.serviceName != null, "serviceName needed with annotation query");
-    Set<String> annotationKeys = new LinkedHashSet<>();
+    Set<String> annotationKeys = new HashSet<>();
+    annotationKeys.add(request.serviceName);
     for (String a : request.annotations) { // doesn't include CORE_ANNOTATIONS
-      annotationKeys.add(request.serviceName + ":" + a);
+      annotationKeys.add(a);
     }
     for (Map.Entry<String, String> b : request.binaryAnnotations.entrySet()) {
-      annotationKeys.add(request.serviceName + ":" + b.getKey() + ":" + b.getValue());
+      annotationKeys.add(b.getKey() + ":" + b.getValue());
     }
     return sortedList(annotationKeys);
   }
--- a/zipkin-storage/cassandra3/src/test/java/zipkin/storage/cassandra3/CassandraUtilTest.java
+++ b/zipkin-storage/cassandra3/src/test/java/zipkin/storage/cassandra3/CassandraUtilTest.java
@@ -52,7 +52,7 @@ public class CassandraUtilTest {
         .serviceName("service")
         .addAnnotation(Constants.ERROR)
         .addBinaryAnnotation(TraceKeys.HTTP_METHOD, "GET").build()))
-        .containsExactly("service:error", "service:http.method:GET");
+        .containsExactly("error", "http.method:GET", "service");
   }
 
   @Test
@@ -62,7 +62,7 @@ public class CassandraUtilTest {
         .serviceName("service")
         .addAnnotation(Constants.ERROR)
         .addAnnotation(Constants.ERROR).build()))
-        .containsExactly("service:error");
+        .containsExactly( "error", "service");
   }
 
   @Test
@@ -78,7 +78,7 @@ public class CassandraUtilTest {
         .containsOnly(Constants.SERVER_ADDR, Constants.CLIENT_ADDR);
 
     assertThat(CassandraUtil.annotationKeys(span))
-        .isEmpty();
+        .containsExactly("web", "ca", "app", "sa");
   }
 
   @Test
@@ -95,6 +95,9 @@ public class CassandraUtilTest {
     )).build();
 
     assertThat(CassandraUtil.annotationKeys(span))
-        .containsOnly("web:aws.arn", "web:aws.arn:" + arn);
+        .containsOnly(
+                "web",
+                "aws.arn:" + arn,
+                TraceKeys.HTTP_URL + ":" + url.substring(0, CassandraUtil.LONGEST_VALUE_TO_INDEX));
   }
 }
--- a/zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraSpanConsumer.java
+++ b/zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraSpanConsumer.java
@@ -99,8 +99,7 @@ final class CassandraSpanConsumer implements GuavaSpanConsumer {
         if (timestamp != null) {
           // Contract for Repository.storeServiceSpanName is to store the span twice, once with
           // the span name and another with empty string.
-          futures.add(storeServiceSpanName(serviceName, span.name, timestamp, span.duration,
-              span.traceId));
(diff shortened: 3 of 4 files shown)


michaelsembwever:
This looks like a good idea. But I don't have much experience on how annotation search works (and is typically used) in the UI layer.

codefromthecrypt:
Will take a look.

zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraUtil.java
```diff
@@ -73,19 +73,20 @@ public static int durationIndexBucket(long ts_micro) {
...
-        annotationKeys.add(b.endpoint.serviceName + ":" + b.key);
-        annotationKeys.add(b.endpoint.serviceName + ":" + b.key + ":" + new String(b.value, UTF_8));
+      if (b.endpoint != null && !b.endpoint.serviceName.isEmpty() && !Constants.CORE_ANNOTATIONS.contains(b.key)) {
+        annotationKeys.add(b.endpoint.serviceName);
+        if (b.type == BinaryAnnotation.Type.STRING) {
+            String value = new String(b.value, UTF_8);
+            value = value.substring(0, Math.min(value.length(), LONGEST_VALUE_TO_INDEX));
+            annotationKeys.add(b.key + ":" + value);
+        } else {
+            annotationKeys.add(b.key);
```
codefromthecrypt, line 89: we don't have to do this, and probably shouldn't. We don't support non-string binary annotations on search

zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraUtil.java
```diff
@@ -73,19 +73,20 @@ public static int durationIndexBucket(long ts_micro) {
...
-          && b.endpoint != null
-          && !b.endpoint.serviceName.isEmpty()
-          && b.value.length <= LONGEST_VALUE_TO_INDEX * 4) { // UTF_8 is up to 4bytes/char
-        String value = new String(b.value, UTF_8);
-        if (value.length() > LONGEST_VALUE_TO_INDEX) continue;
-
-        annotationKeys.add(b.endpoint.serviceName + ":" + b.key);
-        annotationKeys.add(b.endpoint.serviceName + ":" + b.key + ":" + new String(b.value, UTF_8));
+      if (b.endpoint != null && !b.endpoint.serviceName.isEmpty() && !Constants.CORE_ANNOTATIONS.contains(b.key)) {
+        annotationKeys.add(b.endpoint.serviceName);
```
codefromthecrypt, line 83: please guard on `BinaryAnnotation.Type.STRING` before adding to the things to index. I don't want to accidentally increase scope to include or promote problematic annotation types

zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraUtil.java
```diff
@@ -96,14 +97,17 @@ public static int durationIndexBucket(long ts_micro) {
       return Collections.emptyList();
     }
     checkArgument(request.serviceName != null, "serviceName needed with annotation query");
-    Set<String> annotationKeys = new LinkedHashSet<>();
```
codefromthecrypt, line 99: set dedupes. maybe I should have put in a comment, but I think we need this (probably also a unit test to show it works)

codefromthecrypt, line 99: basically it is an edge case, where request.annotations has a key that's in request.binaryAnnotations.. perhaps we can just dedupe in the request object instead.. or not bother and switch to list with a risk of accidentally making the same query twice.. wdyt

zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraUtil.java
```diff
@@ -96,14 +97,17 @@ public static int durationIndexBucket(long ts_micro) {
...
+    List<String> annotationKeys = new ArrayList<>();
+    annotationKeys.add(request.serviceName);
     for (String a : request.annotations) { // doesn't include CORE_ANNOTATIONS
-      annotationKeys.add(request.serviceName + ":" + a);
+      if (!annotationKeys.contains(a))
+          annotationKeys.add(a);
     }
     for (Map.Entry<String, String> b : request.binaryAnnotations.entrySet()) {
-      annotationKeys.add(request.serviceName + ":" + b.getKey() + ":" + b.getValue());
+      if (!annotationKeys.contains(b.getKey() + ":" + b.getValue()))
```
codefromthecrypt: ok if we are doing this, then set seems simpler. and less likely for someone to accidentally change one line and not the other.. is there a bigger reason why you aren't using set for this?

zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraUtil.java
```diff
@@ -96,14 +97,17 @@ public static int durationIndexBucket(long ts_micro) {
       return Collections.emptyList();
     }
     checkArgument(request.serviceName != null, "serviceName needed with annotation query");
-    Set<String> annotationKeys = new LinkedHashSet<>();
+    List<String> annotationKeys = new ArrayList<>();
+    annotationKeys.add(request.serviceName);
     for (String a : request.annotations) { // doesn't include CORE_ANNOTATIONS
-      annotationKeys.add(request.serviceName + ":" + a);
+      if (!annotationKeys.contains(a))
```
codefromthecrypt: yeah really not a fan of defensive list add vs just using a set.

zipkin-storage/cassandra3/src/test/java/zipkin/storage/cassandra3/CassandraUtilTest.java
```diff
@@ -52,7 +52,7 @@ public void annotationKeys() {
         .serviceName("service")
         .addAnnotation(Constants.ERROR)
         .addBinaryAnnotation(TraceKeys.HTTP_METHOD, "GET").build()))
-        .containsExactly("service:error", "service:http.method:GET");
+        .containsExactly("service", "error", "http.method:GET");
```
codefromthecrypt: interesting.. surprised this works, but maybe I shouldn't be

codefromthecrypt:
I'd really like to not use defensive list add, and instead go back to the set approach to ensure unique keys. it is easier to not have to think about pairing if-add, and instead just assert a set add.

WRT the change itself.. honestly, I'm pleasantly surprised it works! Either our SpanStoreTest isn't good enough to show this doesn't work, or it works and it is a far more elegant index. Regardless, I'm inclined to go with it until someone screams.

zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraUtil.java
michaelsembwever: but we can…
i've tested this and it means that searching on non-string binary annotations is possible.

feasibility of it aside, has it value?

zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraUtil.java
michaelsembwever: went to list with manual dedupe to keep order. (i think the reason was to have some consistency when reading the data in `cqlsh`. ie you're always reading it as service, annotations, binaryAnnotations)

zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraUtil.java
codefromthecrypt: > feasibility of it aside, has it value?
> 
> No, this is value-reducing. For example, we would have to increase scope
> of other things to make sure special-case boolean annotations like "sa"
> aren't accidentally now indexed.

Moreover, there are numerous problems in the subtleties of thrift types.
When we start indexing them by default, we are reversing our position (that
discouraged them), and would have to take responsibility to support them
despite their problems. For example, this makes transitioning to a new
model that doesn't have an I16 type more difficult. It raises questions on
lossiness of I64 when rendered in json (happened last week!) Right now, the
api is pretty clear about text annotations only, and this is what we should
be doing.

zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraUtil.java
codefromthecrypt: LinkedHashSet retains order.. how about using that?

codefromthecrypt: I think the "sortedList" thing at the bottom was the real problem
underneath the behavior your raised. We could copy-out a linkedHashSet to a
non-sorted list.

michaelsembwever: i have no objection to changing this back to a set.

am more interested in the higher level ramifications of this PR.
it does bring us closer to full-text searching.

michaelsembwever:
> Either our SpanStoreTest isn't good enough to show this doesn't work, or it works and it is a far more elegant index. Regardless, I'm inclined to go with it until someone screams.

I've tested it through the UI, and it does work well (making search more usable imho), and significantly reduces the index size on disk.

i'll restore the set.

codefromthecrypt:
yeah this is a good change.. I'll timebox an hour to see if I can break it (ex if we are simply missing a test). otherwise, will merge :)

zipkin-storage/cassandra3/src/test/java/zipkin/storage/cassandra3/CassandraUtilTest.java
```diff
@@ -78,7 +78,7 @@ public void annotationKeys_skipsCoreAndAddressAnnotations() throws Exception {
         .containsOnly(Constants.SERVER_ADDR, Constants.CLIENT_ADDR);
 
     assertThat(CassandraUtil.annotationKeys(span))
-        .isEmpty();
+        .containsExactly("web", "ca", "app", "sa");
```
codefromthecrypt, line 81: this one should fail I think

codefromthecrypt:
sorry to report, but we were missing a test... will raise a PR with it later.

codefromthecrypt:
here's the missing test, in case it helps you between now and when I fix the in-memory storage which this fails on (old cassandra passes) https://github.com/openzipkin/zipkin/pull/1292

zipkin-storage/cassandra3/src/test/java/zipkin/storage/cassandra3/CassandraUtilTest.java
```diff
@@ -95,6 +95,9 @@ public void annotationKeys_skipsBinaryAnnotationsLongerThan256chars() throws Exc
     )).build();
 
     assertThat(CassandraUtil.annotationKeys(span))
-        .containsOnly("web:aws.arn", "web:aws.arn:" + arn);
```
codefromthecrypt, line 98: ps I only now noticed that for a year we were indexing binaryAnnotation.key in the same space as annotation.value. This goes against all of our docs, and probably people just didn't notice.

Took it out in https://github.com/openzipkin/zipkin/pull/1292

codefromthecrypt:
so good news and bad news.

Good news is that we didn't need to index binaryAnnotation.key independently, so in any case, writes should be less (only one per annotation/binaryAnnotation+service)

Bad news is that the search is imprecise without the service name. I've changed delimiters of binary annotations to ';' to avoid a LIKE clash with annotations (which are delimited with ':')

codefromthecrypt:
gonna close this as after looking closely and adding a missing test, I don't think it will work as-is. Feel free to re-open if this is not the case.

PR #1290, opened 2016-09-12
Rewrite of self-tracing in Cassandra3 storage

michaelsembwever:
Self Tracing used in combination with a Zipkin-tracing Cassandra installation does not work (recursion of traces).

This rewrites the self-tracing implementation from using a java.lang.reflect.InvocationHandler to subclassing the CQL driver's AbstractSession class.

Avoiding the repeated self-tracing of self-tracing generated spans is done through a combination of adding a binary annotation to the span and a flag to the statement's
outgoing payload.

### DIFF
--- a/zipkin-autoconfigure/storage-cassandra3/src/main/java/zipkin/autoconfigure/storage/cassandra3/brave/TracedSession.java
+++ b/zipkin-autoconfigure/storage-cassandra3/src/main/java/zipkin/autoconfigure/storage/cassandra3/brave/TracedSession.java
@@ -11,204 +11,182 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  */
+
 package zipkin.autoconfigure.storage.cassandra3.brave;
 
-import com.datastax.driver.core.BoundStatement;
+import com.datastax.driver.core.AbstractSession;
+import com.datastax.driver.core.CloseFuture;
 import com.datastax.driver.core.Cluster;
-import com.datastax.driver.core.Host;
-import com.datastax.driver.core.LatencyTracker;
+import com.datastax.driver.core.PreparedStatement;
 import com.datastax.driver.core.ProtocolVersion;
-import com.datastax.driver.core.ResultSet;
+import com.datastax.driver.core.RegularStatement;
 import com.datastax.driver.core.ResultSetFuture;
 import com.datastax.driver.core.Session;
 import com.datastax.driver.core.Statement;
+import com.datastax.driver.core.exceptions.DriverException;
+import com.datastax.driver.core.exceptions.DriverInternalError;
+import com.datastax.driver.core.querybuilder.QueryBuilder;
 import com.github.kristofa.brave.Brave;
-import com.github.kristofa.brave.ServerSpan;
-import com.github.kristofa.brave.ServerSpanThreadBinder;
+import com.github.kristofa.brave.ClientTracer;
 import com.github.kristofa.brave.SpanCollector;
 import com.github.kristofa.brave.SpanId;
-import com.google.common.collect.Maps;
-import com.google.common.reflect.AbstractInvocationHandler;
-import com.google.common.reflect.Reflection;
-import com.twitter.zipkin.gen.Annotation;
-import com.twitter.zipkin.gen.BinaryAnnotation;
-import com.twitter.zipkin.gen.Endpoint;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
+import com.google.common.util.concurrent.Uninterruptibles;
 import com.twitter.zipkin.gen.Span;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
 import java.nio.ByteBuffer;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import zipkin.Constants;
+import zipkin.storage.cassandra3.Cassandra3Storage;
 
-import static java.util.Collections.singletonMap;
-import static zipkin.internal.Util.checkNotNull;
+final class TracedSession extends AbstractSession {
 
-/**
- * Creates traced sessions, which write directly to brave's collector to preserve the correct
- * duration.
- */
-public final class TracedSession extends AbstractInvocationHandler implements LatencyTracker {
-  private static final Logger LOG = LoggerFactory.getLogger(TracedSession.class);
+    private static final Executor EXECUTOR = MoreExecutors.directExecutor();
+    private final Session session;
+    private final Brave brave;
+    private final SpanCollector collector;
+    private final ClientTracer clientTracer;
+    private final long useKeyspaceTimeout;
+    private volatile boolean keyspaceSet = false;
+    private final ProtocolVersion version;
 
-  private final ProtocolVersion version;
-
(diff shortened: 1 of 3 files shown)


michaelsembwever:
This is as much an experiment and alternative approach as an improvement.
A point for discussion, if you like.

Whether it is merged or not I have no care about.

zipkin-autoconfigure/storage-cassandra3/src/main/java/zipkin/autoconfigure/storage/cassandra3/brave/TracedSession.java
```diff
@@ -11,204 +11,182 @@
...
-    Span span = cache.remove(statement);
-    if (span == null) {
-      if (statement.isTracing()) {
-        LOG.warn("{} not in the cache eventhough tracing is on", statement);
-      }
-      return;
+
+    @Override
+    public String getLoggedKeyspace() {
+        return session.getLoggedKeyspace();
```
codefromthecrypt, line 72: formatting is off. zipkin is 2 spaces, etc.  Easiest way (if using intellij) is to use this: https://github.com/square/java-code-styles

2 space tab and indent, with 4 space continuation indent... https://github.com/openzipkin/zipkin/blob/master/CONTRIBUTING.md

zipkin-autoconfigure/storage-cassandra3/src/main/java/zipkin/autoconfigure/storage/cassandra3/brave/TracedSession.java
```diff
@@ -11,204 +11,182 @@
...
+            payload.put("zipkin", ByteBuffer.wrap(spanId.bytes()));
+            statement.setOutgoingPayload(payload);
+
+            final Span span = null != spanId ? brave.clientSpanThreadBinder().getCurrentClientSpan() : null;
+            ResultSetFuture future = session.executeAsync(statement);
+            if (null != spanId) {
+                future.addListener(new Runnable() {
+                    @Override
+                    public void run() {
+                        brave.clientSpanThreadBinder().setCurrentSpan(span);
```
codefromthecrypt, line 112: albeit unlikely the middle bits will throw, I'd add a try-finally. Also note server-modules are language level 8, so you can do a lambda

zipkin-autoconfigure/storage-cassandra3/src/main/java/zipkin/autoconfigure/storage/cassandra3/brave/TracedSession.java
```diff
@@ -11,204 +11,182 @@
...
 
-    @Override public ResultSet get(long timeout, TimeUnit unit)
-        throws InterruptedException, ExecutionException, TimeoutException {
-      return delegate.get(timeout, unit);
+    @Override
+    protected ListenableFuture<PreparedStatement> prepareAsync(String query, Map<String, ByteBuffer> customPayload) {
+        return session.prepareAsync(query);
     }
 
-    @Override public void addListener(Runnable listener, Executor executor) {
```
codefromthecrypt, line 203: I think the reason I did this (created a wrapped future vs chaining) was I a heck of a time getting async to actually work properly. I presume you've verified this patch actually works as expected (versions have changed, since so maybe a driver bug I hit was been squashed)

zipkin-autoconfigure/storage-cassandra3/src/main/java/zipkin/autoconfigure/storage/cassandra3/brave/TracedSession.java
```diff
@@ -11,204 +11,182 @@
...
+        // don't trace any already self-traced spans
+        boolean alreadySelfTraced = null != statement.getOutgoingPayload()
+                && statement.getOutgoingPayload().containsKey(Cassandra3Storage.SELF_TRACING_KEY);
+
+        // o.a.c.tracing.Tracing.newSession must use the same format for the key zipkin
+        boolean supported = version.compareTo(ProtocolVersion.V4) >= 0;
+        if (supported && !alreadySelfTraced) {
+            useKeyspace(session, statement.getKeyspace());
+            SpanId spanId = null;
+            spanId = clientTracer.startNewSpan(statement.toString());
```
codefromthecrypt, line 91: please carry-forward the comments here and elsewhere unless they are invalid now. Ex. this one had..

```
// via an internal class z.s.cassandra3.NamedBoundStatement, toString() is a nice name
```

We don't want default BoundStatement toString as it breaks most advice about span naming

zipkin-autoconfigure/storage-cassandra3/src/main/java/zipkin/autoconfigure/storage/cassandra3/brave/TracedSession.java
```diff
@@ -11,204 +11,182 @@
...
-
-  @Override
-  public boolean equals(Object obj) {
-    if (obj instanceof TracedSession) {
-      TracedSession other = (TracedSession) obj;
-      return target.equals(other.target);
+
+    @Override
+    public ResultSetFuture executeAsync(final Statement statement) {
+        // don't trace any already self-traced spans
```
codefromthecrypt, line 82: I noticed statements like this.. how is this possible?

codefromthecrypt, line 82: I mean how is it now possible to have a cycle in self-tracing?

codefromthecrypt, line 82: before, I added this for a couple reasons.. one is to address perhaps your concern about the server amplifying itself, and another is to special-case it so that we don't have as much of a test burden. What's worrying me about trying to generify this is that it is getting past the point where we should kite without tests..

Ex the below was a heavily guarded thing that ensured we don't have to worry about things like local collectors amplifying, nor any other traces we don't expect to make. A generic cassandra tracer would have to be more... generic, but then it would have a higher burden of testing etc. If that's the goal, I might copy/paste the work and then make a cassandra client tracer out of it

```
-    // Only join traces, don't start them. This prevents LocalCollector's thread from amplifying.
 -    if (brave.serverSpanThreadBinder().getCurrentServerSpan() != null &&
 -        brave.serverSpanThreadBinder().getCurrentServerSpan().getSpan() != null
 -        && method.getName().equals("executeAsync") && args[0] instanceof BoundStatement) {
```

codefromthecrypt:
Thanks for putting this together. It made me think! long thoughts about the thing in general

### Client-tracing in C\* is not yet transparent

Regardless of how we coordinate, we need to coordinate. Ex you introduce a special binary annotation, before we also had to do work to control names (especially as datastax nixed the ability to have client-side metadata). I needed to use a proxy to make it drop-in (without collaboration code from the caller). This is a similar problem in Kafka, except easier in Cassandra as at least there's a server-hook.

### I should have written tests on server tracing

A lot of things had to happen before the storage layer could be decoupled from the trace layer explicitly, with short, low-cardinality span names and without async bugs. It took more than a couple days to get this working. When thinking about tests, I realized it would likely take a few days to create them (as it needs to read back and forth to figure out things). I special-cased with flags, smaller entry scope instead, and zero coordination instead of investing more time to make this a real tracer. If I wrote tests, we'd know that this is a drop-in or not. Now, we don't.. at least not without eye-balling spans and running manually.

### End notes

Regardless of which approach is used, it is probably mutually obvious that the code involved is probably better off as a library, particularly when it aims to act generically (ex work on any Statement). If such a library could drive changes in the datastax api to make it more transparent, far more than self-tracing would benefit.

codefromthecrypt:
gonna close this as from your description you seemed to be most interested in feedback. I'd like to keep the PR queue clean as possible. If I closed this too early, please re-open!

Issue #1284, opened 2016-09-09
UI: Visualize client vs server time feature request


rogeralsing:
As proposed by #1282 my impression is that we will see the entire span from client send to client receive in the UI.

It would be really nice if the span could be visualized with different colors/appearance.

e.g.

```
Client: ################################################################
Server:            --------------####################-----------          
```

Where `-` is latency between client and server and `#` is server time.
This would give a much better view of where time is being consumed.

eirslett:
We use a `<div>` now for every span in the visualisation. I suppose we could replace it with `<svg>` and draw a trapezoid instead...
Or even render the entire visualisation with SVG?

rogeralsing:
After inspecting the DOM, I noticed that this is sort of already in place. there are DIV's for each annotation, but they are hidden..
Changing them to display:block makes them appear.
(The white circles on the 2nd span)

Is there any way in the UI to enable them or are those work in progress or why are they hidden?

![annotations](https://cloud.githubusercontent.com/assets/647031/18411633/78b5e8ea-777b-11e6-8120-0f67a3fb51cf.png)

rogeralsing:
``` css
#trace-container .span .duration .annotation.core {
    display: none; //<- setting this to block shows annotations
}
```

eirslett:
They are supposed to be hidden, because only non-core annotations are supposed to be shown. These annotations are actually a separate feature.

michaelsembwever:
The idea of light grey area on each side of a span indicating the network time to and fro that call would be very cool. (The white circles are best reserved for non core annotations imho).

codefromthecrypt:
I was wondering about why dots dont show up for core annotations :) thanks,
eirik! (I will push a comment commit)

codefromthecrypt:
ps there are some [related annotations](https://github.com/openzipkin/zipkin-api/blob/master/thrift/zipkinCore.thrift#L82) that are less common you'd want to consider. For example, wire send ("ws") which is there to help show actual network time. Here's an example to consider.

``` json
[
  {
    "traceId": "cfe7bb9e10d37dac",
    "name": "get",
    "id": "cfe7bb9e10d37dac",
    "timestamp": 1473723216351000,
    "duration": 5000,
    "annotations": [
      {
        "timestamp": 1473723216351000,
        "value": "sr",
        "endpoint": {
          "serviceName": "frontend",
          "ipv4": "127.0.0.1",
          "port": 8081
        }
      },
      {
[...]

PR #1291, opened 2016-09-12
Documents why core annotations are not shown in the UI

codefromthecrypt:
See https://github.com/openzipkin/zipkin/issues/1284#issuecomment-246141155


codefromthecrypt:
cc @eirslett

PR #1292, opened 2016-09-12
Qualifies annotation queries against a service

codefromthecrypt:
This ensures that annotation or binary annotation queries are qualified
against a service.

This also removes the binary annotation key indexing used in cassandra,
but nowhere else. There's nothing in the api, javadoc or web UI docs
that suggest binary annotation key should substitute for annotation
value. Removing this indexing halves writes per binary annotation
without violating our api contract.

### DIFF
--- a/zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraUtil.java
+++ b/zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraUtil.java
@@ -57,8 +57,15 @@ final class CassandraUtil {
   }
 
   /**
-   * Returns keys that concatenate the serviceName associated with an annotation, a binary
-   * annotation key, or a binary annotation key with value.
+   * Returns keys that concatenate the serviceName associated with an annotation or a binary
+   * annotation.
+   *
+   * <p>Note: Annotations are delimited with colons while Binary Annotations are delimited with
+   * semi-colons. This is because the returned keys are joined on comma and queried with LIKE. For
+   * example, a span with the annotation "foo" and a binary annotation "bar" -> "baz" would end up
+   * in a cell "service:foo,service:bar:baz". A query for the annotation "bar" would satisfy as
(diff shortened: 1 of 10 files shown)


adriancole:
only test fail locally is elasticsearch. I'm looking into it now.

zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/ElasticsearchSpanStore.java
```diff
@@ -111,16 +112,18 @@ public int compare(List<Span> left, List<Span> right) {
       filter.must(termQuery("name", request.spanName));
     }
     for (String annotation : request.annotations) {
-      filter.must(termQuery("annotations.value", annotation));
+      filter.must(nestedQuery("annotations", boolQuery()
```
codefromthecrypt: fyi @devinsba @dan-tr @anuraaga I had to adjust the index a bit to fix this

zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraUtil.java
```diff
@@ -101,7 +101,7 @@ public static int durationIndexBucket(long ts_micro) {
       annotationKeys.add(request.serviceName + ":" + a);
     }
     for (Map.Entry<String, String> b : request.binaryAnnotations.entrySet()) {
-      annotationKeys.add(request.serviceName + ":" + b.getKey() + ":" + b.getValue());
+      annotationKeys.add(request.serviceName + ";" + b.getKey() + ";" + b.getValue());
```
codefromthecrypt, line 111: cc @michaelsembwever on "cassandra3" I had to change delimiters of binary annotations to ';' to avoid a LIKE clash with annotations (which are delimited with ':')

michaelsembwever: ok

Issue #1253, opened 2016-08-24
Support AWS hosted Elasticsearch service


devinsba:
In order to use the AWS hosted Elasticsearch service we will need to support the REST api instead of the native transport API.

I'm wondering if this should be a separate module or a flag in the ElasticSearch storage builder if I were to create a PR for this.

codefromthecrypt:
Unless the libraries change or the code ends up way different, I would make
it a flag or similar. I presume there's feature parity across the two?

devinsba:
I don't know about feature parity. I do think it is a different client
library to talk over REST. I only gave it a quick glance at the end of the
day though since I had some time to kill at work. I'll follow up when I get
my computer back in the morning

codefromthecrypt:
In the mean time, pinged on twitter about this as this is a feature requested for them many times. It would be nice to understand if this is a backlog item on the product or not.

https://twitter.com/adrianfcole/status/768616436948271104

I noticed that maggieaws is the one doing product launch announcements on the forum:
https://forums.aws.amazon.com/ann.jspa?annID=3946

This might be the correct maggie, or maybe just there are multiple AWS product managers named maggie
https://www.linkedin.com/in/yuanzhuang

If someone knows the product team, please reach out and see if this is on the roadmap as rewriting would be fruitless if it is.

anuraaga:
There would never be a feature difference between the Transport client and REST since non-Java clients have to use the REST client ;)

The Transport client is a lot more efficient thanks to somewhat more compact serialization and multiplexing over the wire. ES unfortunately doesn't support HTTP2 on the rest port so the lack of multiplexing can matter at high loads.

I think it's possible to convert a request object that we construct already directly into a json object that can be POSTed, so supporting both may not be so bad. Though I generally don't recommend AWS managed ES since going by history, it'll be a year behind the times again when 5.0 comes out...

codefromthecrypt:
PS elastic just mentioned that their cloud (which runs on AWS) supports transport client. So, that's one way to get collocated data in AWS. https://twitter.com/elastic/status/768621013063995392

codefromthecrypt:
there's a new impl in town, via @sethp-jive https://github.com/openzipkin/zipkin/pull/1302

codefromthecrypt:
Will be released in zipkin 1.12

``` bash
# make sure your cli credentials are setup as zipkin will read them
[...]

PR #1294, opened 2016-09-13
Work in progress Elasticsearch HTTP transport

codefromthecrypt:
See #1253

### DIFF
--- a/zipkin-storage/elasticsearch-rest/src/main/java/zipkin/storage/elasticsearch/ElasticsearchRestSpanConsumer.java
+++ b/zipkin-storage/elasticsearch-rest/src/main/java/zipkin/storage/elasticsearch/ElasticsearchRestSpanConsumer.java
@@ -1,0 +1,122 @@
+/**
+ * Copyright 2015-2016 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin.storage.elasticsearch;
+
+import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.Moshi;
+import java.io.IOException;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import okhttp3.Call;
+import okhttp3.HttpUrl;
+import okhttp3.MediaType;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+import okio.Buffer;
+import zipkin.Span;
+import zipkin.storage.AsyncSpanConsumer;
+import zipkin.storage.Callback;
+
+import static zipkin.internal.ApplyTimestampAndDuration.guessTimestamp;
+
+final class ElasticsearchRestSpanConsumer implements AsyncSpanConsumer {
+  static final MediaType APPLICATION_JSON = MediaType.parse("application/json");
+
+  final OkHttpClient client;
+  final HttpUrl baseUrl;
+  final JsonAdapter<Span> spanAdapter;
+  final JsonAdapter<IndexActionAndMetadata> indexActionAndMetadataAdapter;
+  final IndexNameFormatter indexNameFormatter;
+  final Request flushRequest;
+
+  ElasticsearchRestSpanConsumer(OkHttpClient client, HttpUrl baseUrl,
+      Moshi moshi, IndexNameFormatter indexNameFormatter) {
+    this.client = client;
+    this.baseUrl = baseUrl;
+    this.spanAdapter = moshi.adapter(Span.class);
+    this.indexActionAndMetadataAdapter = moshi.adapter(IndexActionAndMetadata.class);
+    this.indexNameFormatter = indexNameFormatter;
+    this.flushRequest = new Request.Builder()
+        .url(baseUrl.resolve("/" + indexNameFormatter.catchAll() + "/_flush"))
+        .post(RequestBody.create(APPLICATION_JSON, "")).build();
+  }
+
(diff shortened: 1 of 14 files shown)


zipkin-storage/elasticsearch-rest/src/main/java/zipkin/storage/elasticsearch/ElasticsearchRestSpanConsumer.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  @Override public void accept(List<Span> spans, final Callback<Void> callback) {
+    StringBuilder request = new StringBuilder();
+    for (Span s : spans) {
+      createSpanIndexRequest(s, request);
+    }
+
+    HttpEntity entity = new NStringEntity(request.toString(), ContentType.APPLICATION_JSON);
+    client.performRequest("POST", "/_bulk", Collections.<String, String>emptyMap(), entity,
```
codefromthecrypt: not terribly thrilled at what the library seems to be giving us. Looks like a combination of lock-into apache http client types plus all the work you usually need when manually writing http apis!

It might be more straightforward (less duplication etc) to hack a http adapter for the transport client. At least then we can choose any http library.

codefromthecrypt: cc @devinsba @anuraaga @shakuzen

zipkin-storage/elasticsearch-rest/src/main/java/zipkin/storage/elasticsearch/ElasticsearchRestStorage.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.io.IOException;
+import org.apache.http.HttpHost;
+import org.elasticsearch.client.RestClient;
+import zipkin.storage.AsyncSpanConsumer;
+import zipkin.storage.AsyncSpanStore;
+import zipkin.storage.SpanStore;
+import zipkin.storage.StorageComponent;
+
+public final class ElasticsearchRestStorage implements StorageComponent {
```
codefromthecrypt, line 31: temporarily forwards to a transport-client so we can test things.

zipkin-storage/elasticsearch-rest/src/test/java/zipkin/storage/elasticsearch/ElasticsearchRestSpanConsumerTest.java
```diff
@@ -0,0 +1,127 @@
...
+import static java.util.Arrays.asList;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.elasticsearch.index.query.QueryBuilders.termQuery;
+import static zipkin.Constants.SERVER_RECV;
+import static zipkin.Constants.SERVER_SEND;
+import static zipkin.TestObjects.DAY;
+import static zipkin.TestObjects.TODAY;
+import static zipkin.TestObjects.WEB_ENDPOINT;
+
+public class ElasticsearchRestSpanConsumerTest {
```
codefromthecrypt, line 32: copy pasted test which actually passes

zipkin-storage/elasticsearch-rest/src/main/java/zipkin/storage/elasticsearch/ElasticsearchRestSpanConsumer.java
codefromthecrypt: basically I'm not thinking about implementing a standalone library for TransportClient, rather intercept the places where we use Client and make a zipkin-specific interface which either forwards to TransportClient or invokes http. That reduces the scope a lot I think, as it only requires us to support things we actually use.

devinsba: In order to authenticate requests to amazon when locking down the security on our domain, we will need to be able to add on an authentication header which is a signature based on the request body.

http://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/what-is-amazon-elasticsearch-service.html#signing-requests

http://docs.aws.amazon.com/general/latest/gr/signature-v4-examples.html#signature-v4-examples-java

codefromthecrypt: luckily you are talking to the original author of jclouds.. I think I've
written that signature code several times :P

codefromthecrypt: wonder if we shouldn't use jest.. it seems to be more than just an apache http configuration project https://github.com/searchbox-io/Jest

codefromthecrypt: jest seems to also have a AWS thing in limbo https://github.com/searchbox-io/Jest/pull/368

codefromthecrypt: jest depends on guava 19, which will also throw things out of whack.. reverting to either using the apache substrate from elastic or similar

codefromthecrypt:
implemented storage flush and wired in a delegate for span store.

Next step: incrementally implement SpanStore, probably easiest ones first :)

zipkin-storage/elasticsearch-rest/src/main/java/zipkin/storage/elasticsearch/ElasticsearchRestSpanConsumer.java
```diff
@@ -0,0 +1,96 @@
...
+          }
+
+          @Override public void onFailure(Exception e) {
+            callback.onError(e);
+          }
+        });
+  }
+
+  // mostly copy/pasted from ElasticSearchSpanConsumer
+  void createSpanIndexRequest(Span span, StringBuilder builder) {
```
codefromthecrypt: @rogeralsing fyi, so we are custom making a bulk request, here, but here are notes in general:

We insert an extra "timestamp_millis" field into each span, which allows kibana to work. When we index, we index as type "span" to the index "span-yyyy-MM-dd"

codefromthecrypt:
using a lightly wrapped apachehc to implement a json api is too much work. I'm switching to okhttp.

codefromthecrypt:
switched to OkHttp+Moshi and implemented the first read api: getRawTrace

codefromthecrypt:
obviated by #1302

PR #1295, opened 2016-09-14
Lowers scope of Elasticsearch to code used in prod

codefromthecrypt:
writeDependencyLinks is package protected and was only used in tests. the spark job has its own implementation


codefromthecrypt:
cc @anuraaga

Issue #1262, opened 2016-08-31
RFC: Transition to 128bit trace ids (still 64bit span ids)


codefromthecrypt:
[Many times](https://groups.google.com/forum/#!msg/zipkin-user/7BUNqrtG5oo/Isf_UyMQ2FMJ;context-place=forum/zipkin-user) over the last years, we've had folks ask about 128 bit trace ids.

There have been a few use-cases mentioned...
- archiving traces longer than a couple days (concern with collisions when retrieving by id)
- using externally generated ids as trace identifiers (ex AWS lambda UUID)
- joining traces that have 128bit identifiers (ex Dapper/Cloud Trace now use 128bit trace ids)

Doing this comes at a cost of figuring out how to do interop, but that's well-charted territory:
- Zipkin already decouples trace ids from span/parent ids; This reduces scope to only trace ids.
- Zipkin trace identifier encoding is hex, which is easy to widen and detect.
- There exists practice of people padding 64bit ids to 128 bit, and bit choosing to reduce to 64 bit.
- Cassandra and MySQL support blobs in key clauses, and elasticsearch can as easily index a hex string of length 16 as it can one of length 32.

Practically speaking we'd need to do the following to make this possible:
- B3 - Update both http and binary forms to optionally accept wider trace ids
- Thrift - add an optional binary field to hold the wider id
- Model - Add Span.Builder.traceId(blob) and pad .traceId(long) to fit 128bits
- Storage
  - overload getTraceById to accept a 128bit blob.
  - add 128bit column to mysql
  - add 128bit index to cassandra
  - write both 64 and 128bit ids until Model v2

Provided we don't creep scope to also change span, parent id encoding, I think we can actually do this, and end up with an incremental win on the model that can help us bridge into systems that use wider ids increasingly often.

codefromthecrypt:
spamming some folks of interest, of course not exhaustive @nicmunroe @felixbarny @shakuzen @yurishkuro @jcarres-mdsol @abesto @eirslett @kristofa @michaelsembwever @kevinoliver @mosesn @anuraaga @marcingrzejszczak @prat0318 @devinsba @basvanbeek @schlosna @ewhauser @klingerf @bogdandrutu @clehene

codefromthecrypt:
One idea to bridge this in the current model is to add another u64 field: traceIdHigh

[...]

PR #1296, opened 2016-09-14
Tolerate 128bit ids in json and http api by throwing out high bits

codefromthecrypt:
This bridges support for 128bit trace ids by not failing on their
receipt. Basically, this throws out any hex characters to the left
of the 16 needed for the 64bit trace id.

See #1262


codefromthecrypt:
cc @yurishkuro fyi @dominikh

Issue #1256, opened 2016-08-29
Make zipkin easier to import into eclipse and intellij


lijunyong:
I git clone this repository and import my eclipse IDE,but it has many errors,how should I do.

codefromthecrypt:
This issue isn't descriptive enough to take action on. We had a fix recently for @bedag-moo who was using eclipse. If you can be more specific on what zipkin needs to do in order to support eclipse, let us know.

lijunyong:
@adriancole I run ./mvnw -DskipTests --also-make -pl zipkin-server clean install  show exception:
[ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.0:npm (npm install) on project zipkin-ui: Failed to run task: 'npm install' failed. (error code -1073741795) -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
[ERROR]
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <goals> -rf :zipkin-ui

and which you use IDE?

codefromthecrypt:
@lijunyong the error you pasted has nothing to do with the IDE. it means you haven't installed npm https://www.npmjs.com/package/npm

I use intellij

codefromthecrypt:
OK, here's the deal on eclipse..

If we import with this, language level isn't set properly and so you have to go through and reset everything to java 8. 

``` bash
$ ./mvnw eclipse:eclipse
```

There's probably other things, too, as eclipse is more strict by default than intellij.

lijunyong:
@adriancole thank you very much

codefromthecrypt:
@lijunyong no prob

eirslett:
You don't need to install node or npm to open the project in Eclipse.

lijunyong:
@eirslett How to do?

codefromthecrypt:
PS recent versions of intellij are now harder to use as they have module-specific compiler settings, but not the ability to override them globally. In a sense, this is similar to the problems in Eclipse.

https://youtrack.jetbrains.com/issue/IDEA-161060

[...]

PR #1302, opened 2016-09-15
Support for AWS managed elasticsearch

sethp-jive:
In order to get a proof of concept working against AWS managed elasticsearch (press button get cluster) I found it necessary to refactor some of the existing storage code to allow it to operate over the native java API or the HTTP API (AWS elasticsearch only supports the latter).

With these changes, a simple `STORAGE_TYPE=elasticsearch ES_HOSTS=https://my-cool-endpoint.us-west-2.es.amazonaws.com/ java -jar ./zipkin-server/target/zipkin-server-*exec.jar` is all it takes to get a reasonably well managed ES backend for a Zipkin deployment.

I'm passing this along in the hope someone else would find this useful – I look forward to your feedback!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -228,6 +228,12 @@
 
       <dependency>
         <groupId>${project.groupId}</groupId>
+        <artifactId>zipkin-autoconfigure-storage-elasticsearch-aws</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>${project.groupId}</groupId>
         <artifactId>zipkin-autoconfigure-metrics-prometheus</artifactId>
         <version>${project.version}</version>
       </dependency>
--- a/zipkin-autoconfigure/storage-elasticsearch-aws/src/main/java/zipkin/autoconfigure/storage/elasticsearch/aws/AwsSignatureInterceptor.java
+++ b/zipkin-autoconfigure/storage-elasticsearch-aws/src/main/java/zipkin/autoconfigure/storage/elasticsearch/aws/AwsSignatureInterceptor.java
@@ -1,0 +1,164 @@
+/**
+ * Copyright 2015-2016 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin.autoconfigure.storage.elasticsearch.aws;
+
+import com.amazonaws.AmazonClientException;
+import com.amazonaws.ReadLimitInfo;
+import com.amazonaws.RequestClientOptions;
+import com.amazonaws.SignableRequest;
+import com.amazonaws.auth.AWS4Signer;
+import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.http.HttpMethodName;
+import com.google.common.base.Charsets;
+import com.google.common.collect.Lists;
+import org.apache.http.*;
+import org.apache.http.client.methods.HttpRequestWrapper;
+import org.apache.http.client.protocol.HttpClientContext;
+import org.apache.http.client.utils.URLEncodedUtils;
+import org.apache.http.protocol.HttpContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Performs aws v4 signing on an apache http request
+ */
+final class AwsSignatureInterceptor implements HttpRequestInterceptor {
+
+  private static final Logger log = LoggerFactory.getLogger(AwsSignatureInterceptor.class);
+
+  private final String serviceName;
+  private final String region;
+  private final AWSCredentialsProvider credentialsProvider;
+
+  public AwsSignatureInterceptor(String serviceName, String region, AWSCredentialsProvider credentialsProvider) {
+    this.serviceName = serviceName;
+    this.region = region;
+    this.credentialsProvider = credentialsProvider;
+  }
+
+  @Override
+  public void process(HttpRequest hr, HttpContext hc) throws HttpException, IOException {
+    AWSCredentials creds;
+    try {
+      creds = credentialsProvider.getCredentials();
+    } catch (AmazonClientException ace) {
+      log.debug("Unable to load AWS credentials", ace);
+      return;
+    }
+
+    AWS4Signer signer = new AWS4Signer();
+    signer.setServiceName(serviceName);
+    signer.setRegionName(region);
+
+    signer.sign(new SignableHttpRequest(hr, hc), creds);
+  }
+
+  private static final class SignableHttpRequest implements SignableRequest<Object> {
+    private final HttpRequestWrapper hr;
+    private final HttpClientContext hc;
+    // read your writes lol
+    private final Map<String, String> signingHeaders = new HashMap<>();
+
+    private SignableHttpRequest(HttpRequest hr, HttpContext hc) {
+      this.hr = hr instanceof HttpRequestWrapper ? (HttpRequestWrapper) hr : HttpRequestWrapper.wrap(hr);
+      this.hc = HttpClientContext.adapt(hc);
+    }
+
+    @Override
+    public void addHeader(String name, String value) {
+      hr.addHeader(name, value);
+      signingHeaders.put(name, value);
+    }
+
+    @Override public Map<String, String> getHeaders() { return signingHeaders; }
+
+    @Override public String getResourcePath() { return hr.getURI().getRawPath(); }
+
+    @Override public void addParameter(String name, String value) { throw new UnsupportedOperationException(); }
+
+    @Override public Map<String, List<String>> getParameters() { return parseQueryParams(hr.getURI()); }
+
+    @Override
+    public URI getEndpoint() {
+      try {
+        return new URI(HttpClientContext.adapt(hc).getTargetHost().toURI());
+      } catch (URISyntaxException ex) {
+        throw new RuntimeException(ex);
+      }
+    }
+
+    @Override public HttpMethodName getHttpMethod() { return HttpMethodName.valueOf(hr.getMethod()); }
+
+    @Override public int getTimeOffset() { return 0; }
+
+    @Override
+    public InputStream getContent() {
+      try {
+        return extractBody(hr);
+      } catch (IOException ex) {
+        throw new RuntimeException(ex);
+      }
+    }
+
+    @Override
+    public InputStream getContentUnwrapped() {
+      try {
+        return extractBody(hr);
+      } catch (IOException ex) {
+        throw new RuntimeException(ex);
+      }
+    }
+
+    @Override
+    public ReadLimitInfo getReadLimitInfo() {
+      return new ReadLimitInfo() {
+        @Override
+        public int getReadLimit() {
+          return RequestClientOptions.DEFAULT_STREAM_BUFFER_SIZE;
+        }
+      };
+    }
+
+    @Override public Object getOriginalRequestObject() { throw new UnsupportedOperationException(); }
+
+    @Override public void setContent(InputStream in) { throw new UnsupportedOperationException(); }
+
+    private static Map<String, List<String>> parseQueryParams(URI uri) {
+      Map<String, List<String>> params = new HashMap<>();
+      for (NameValuePair pair : URLEncodedUtils.parse(uri, Charsets.UTF_8.name())) {
+        params.put(pair.getName(), Lists.newArrayList(pair.getValue()));
+      }
+
+      return params;
+    }
+
+    private static InputStream extractBody(HttpRequest request) throws IOException {
+      if (request instanceof HttpEntityEnclosingRequest) {
+        HttpEntity entity = ((HttpEntityEnclosingRequest) request).getEntity();
+        return entity != null ? entity.getContent() : null;
+      } else {
+        return null;
+      }
+    }
+  }
+}
--- a/zipkin-autoconfigure/storage-elasticsearch-aws/src/main/java/zipkin/storage/elasticsearch/LazyDomainLookup.java
+++ b/zipkin-autoconfigure/storage-elasticsearch-aws/src/main/java/zipkin/storage/elasticsearch/LazyDomainLookup.java
@@ -1,0 +1,68 @@
+/**
+ * Copyright 2015-2016 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin.storage.elasticsearch;
+
+import com.amazonaws.services.elasticsearch.AWSElasticsearch;
+import com.amazonaws.services.elasticsearch.AWSElasticsearchClientBuilder;
+import com.amazonaws.services.elasticsearch.model.DescribeElasticsearchDomainRequest;
+import com.amazonaws.services.elasticsearch.model.DescribeElasticsearchDomainResult;
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+
+public class LazyDomainLookup implements InternalElasticsearchClient.Builder {
+
+  private final String domain;
+  private final String region;
+  private final InternalElasticsearchClient.Builder delegate;
+
+  public LazyDomainLookup(String domain, String region, InternalElasticsearchClient.Builder delegate) {
+    this.domain = domain;
+    this.region = region;
+    this.delegate = delegate;
+  }
+
+  @Override public InternalElasticsearchClient.Builder cluster(String cluster) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override public InternalElasticsearchClient.Builder hosts(List<String> hosts) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override public InternalElasticsearchClient.Builder flushOnWrites(boolean flushOnWrites) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override public InternalElasticsearchClient.Factory buildFactory() {
+    return new InternalElasticsearchClient.Factory() {
+      @Override public InternalElasticsearchClient create(String allIndices) {
+        AWSElasticsearch es = AWSElasticsearchClientBuilder.standard()
+            .withRegion(region)
+            .build();
+
+        DescribeElasticsearchDomainResult result =
+            es.describeElasticsearchDomain(new DescribeElasticsearchDomainRequest()
+                .withDomainName(domain));
+
+        String endpoint = result.getDomainStatus().getEndpoint();
+        if (!endpoint.startsWith("https://")) {
+          endpoint = "https://" + endpoint;
+        }
+        delegate.hosts(ImmutableList.of(endpoint));
+        return delegate.buildFactory().create(allIndices);
+      }
+    };
+
+  }
+}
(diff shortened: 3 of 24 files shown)


codefromthecrypt:
holy moly! this is quite a bit of progress!

codefromthecrypt:
@sethp-jive thanks for this work! I'll stop my work and help yours.

I think the next step is to modularize. Right now, there are a number of classpath problems we have already, so adding two new dep trees (especially with guava etc), is something we should avoid forcing. Here's the idea.

create elasticsearch-jest (which has the jest dependency)
create elasticsearch-aws (which depends on elasticsearch-jest and aws sdk)

This implies exposing Client, probably as InternalElasticsearchClient extending LazyCloseable. The Internal thing implies that it is internal eventhough it is exposed public. place the jest and aws impls in a separate package like zipkin.elasticsearch.jest and zipkin.elasticsearch.aws. These classes that extend InternalElasticsearchClient should have builders that look like other builders, and be an arg to ElasticSearchStorage.builder without any prior knowledge (ex  ElasticSearchStorage has no registry)

In those modules, we would have any unit tests they need, but at the very least copies of the DependenciesTest and SpanStoreTest so that they can be tested independently. The AWS variant would look for AWS access key or similar. If these args are present, the AWS variant runs or silently ignores (because its TestGraph is written to). Avoid creating a elasticsearch test jar unless it really really really helps :)

Once that's together we can work on auto-configuration (via spring). No magic until then. ex please don't detect anything :) magic is for auto-configuration, storage components should have smart defaults, but no magic.

codefromthecrypt:
pps please limit use of guava in InternalElasticsearchClient, and if it is really required, limit it to signatures present in 18.0

codefromthecrypt:
incidentally, aws could probably be done as an autoconfigure module. It doesn't seem to need to affect anything else, so I think we'd have..

### modify: zipkin-storage/elasticsearch

refactored to expose zipkin.elasticsearch.InternalElasticsearchClient and accept it

### add: zipkin-storage/elasticsearch-jest

depends on ^^ and jest, implements  zipkin.elasticsearch.jest.JestElasticsearchClient

### add: zipkin-autoconfigure/storage-elasticsearch-jest

maps spring properties for http endpoints and instantiates JestElasticsearchClient via classes in package zipkin.autoconfigure.storage.elasticsearch.jest

### add: zipkin-autoconfigure/storage-elasticsearch-aws

uses AWS sdk to provide overrideable defaults for http endpoint and authentication and instantiates JestElasticsearchClient via classes in package zipkin.autoconfigure.storage.elasticsearch.aws

codefromthecrypt:
ps wrt commits, it is probably nicest to merge the refactor one first, as it is less code.

Then a PR for:
zipkin-storage/elasticsearch-jest, zipkin-autoconfigure/storage-elasticsearch-jest (and changes to zipkin-server including property/env variable documentation)

Then, a PR for:
zipkin-autoconfigure/storage-elasticsearch-aws (and changes to zipkin-server including property/env variable documentation)

I'm ok reviewing everything as one pull request, just this ^^ is one way to break it up.

sethp-jive:
Oh, cool, I had totally missed that this was already in progress, I'm sorry for dropping in unannounced. Thanks for the feedback! 

I'll get started teasing this stuff out into modules as you described, but I have a few questions for you:
1. Jest's dependency on Elasticsearch core (and its drag) is optional. I've just kept it around it for the query builders really – if you're comfortable with embedding those in the source as JSON instead of objects, we could have a zipkin-storage/elasticsearch, zipkin-storage/elasticsearch-transport, zipkin-storage/elasticsearch-http (or Jest) tree?
2. As far as I know (which is not very far, admittedly), the difference between the Transport client and Jest is largely of Java API (+ discovery entrypoint, naturally) – do we want to consider simplifying things by just replacing the elasticsearch implementation with one over HTTP instead of the binary transport? So just zipkin-storage/elasticsearch and zipkin-storage/elasticsearch-http? I didn't go that route initially because it would be a breaking change for existing configurations, but I thought I should mention it.
3. I'm not very familiar with spring-boot or the autoconfiguration stuff, but I think I can figure it out if you can help me understand the final vision for how this stuff gets driven. Is the current (user-facing) magic regarding URL parsing and host-pattern-matching acceptable? e.g. one `STORAGE_TYPE=elasticsearch` and switching on the value in `ES_HOSTS`, or are we aiming for a suite of storage options like `STORAGE_TYPE=elasticsearch`, `STORAGE_TYPE=elasticsearch-http`, `STORAGE_TYPE=elasticsearch-aws`?
4. Regarding testing the AWS integration – I don't know if it was clear, but the AWS code makes identical requests to the normal HTTP client with the addition of a pair of magic AWS headers. Regrettably, the contents of those headers are tough to validate except against a running AWS ES instance (I have a gnarly story about just how many URL encoding rounds it accepts for commas, if you want to hear it). I think that means just detecting on the presence of credentials is not quite enough – we also need a "domain" (what AWS calls an ES cluster) to connect to. How would you imagine this working in an environment that's not just my machine (with access to my AWS account)? i.e. should I be looking in an environment variable for that domain? I could code the tests to try creating a domain with the credentials at hand, but regrettably that takes quite a few minutes.

I think if you're alright with it, I'd like to keep this PR as a kind of "tracking" issue and for the final zipkin-storage/elasticsearch-aws modules when we get there, and I can send you other PRs for the steps along the way.

codefromthecrypt:
coolio. this will need to be rebased on top of #1304 once merged

codefromthecrypt:
This is a summarising reply :)

I can help with the magic stuff and also some of the test setup issues.

Let's go with jest for now (vs ground up http or es 5 client) as we can
always optimize later. If you run into trouble with modularisation or
magic, just note in the pull request and I can take some efforts to tidy
things.

I suspect that we can handle most of this in spring by looking at the value
of the elasticsearth hosts. Ex if they are amazon, assume that. If the are
not port 9300 assume http. We can have something like ES_CLIENT_TYPE=http
or native or aws to force a decision if need be.

sethp-jive:
I put up the alternative implementation as you described (I think) in #1306 – I wasn't quite sure what to do with some parts (especially the autoconfigure conditional stuff), but it seems to approximate the right behavior for me.

If you've got anything you want me to take care of, let me know. Otherwise, I'll focus on getting the AWS request signing stuff wedged in via a new autoconfigure package.

codefromthecrypt:
looks like we are on final lap. I'd like to cut a release from master sometime in the next day or two as there's a method I'd like to deprecate and get a replacement out for. do you think you'll have time for the aws autoconfigure in the next day or two? If so, it would be nice to make a general announcement for elasticsearch http+aws vs separately.

sethp-jive:
Yup – I'm in the final stages of testing here on my end before wrapping up the last change and sending it along for review.

codefromthecrypt:
@sethp-jive great. once you have this is, I'll take a look at it and possibly modify it to use https://github.com/spring-cloud/spring-cloud-aws

Reason being is there's a standard way to get AWS credentials in spring boot, I can port the approach if you don't have cycles to look into it.

cc @dsyer

zipkin-autoconfigure/storage-elasticsearch-http/src/main/java/zipkin/autoconfigure/storage/elasticsearch/http/ZipkinHttpElasticsearchStorageAutoConfiguration.java
```diff
@@ -25,21 +30,61 @@
 import zipkin.autoconfigure.storage.elasticsearch.ZipkinElasticsearchStorageAutoConfiguration;
 import zipkin.storage.StorageComponent;
 
+import static com.google.common.base.Preconditions.checkArgument;
+
 @Configuration
-@EnableConfigurationProperties(ZipkinElasticsearchHttpStorageProperties.class)
+@EnableConfigurationProperties(ZipkinHttpElasticsearchStorageProperties.class)
```
sethp-jive: I think this is more conventional this way? I changed it to mirror your change to the HTTP autoconfiguration.

zipkin-autoconfigure/storage-elasticsearch-aws/src/main/java/zipkin/storage/elasticsearch/LazyDomainLookup.java
```diff
@@ -0,0 +1,68 @@
...
+  @Override public InternalElasticsearchClient.Builder hosts(List<String> hosts) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override public InternalElasticsearchClient.Builder flushOnWrites(boolean flushOnWrites) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override public InternalElasticsearchClient.Factory buildFactory() {
+    return new InternalElasticsearchClient.Factory() {
```
sethp-jive, line 48: Since the endpoint URLs that get dumped out of the AWS ES service aren't particularly friendly (e.g. `https://search-name-wvkjy7wgk6j7iihjqcqemhhfju.us-west-2.es.amazonaws.com` ) I found it useful to have Zipkin do the lookup for me to map from the name of the domain to the endpoint for a few different testing purposes. I also thought it might be nice to have a second more explicit path in just in case the AWS "magic" parsing of the URLs fails.

zipkin-server/src/main/java/zipkin/server/ZipkinServer.java
```diff
@@ -14,10 +14,11 @@
 package zipkin.server;
 
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.autoconfigure.elasticsearch.jest.JestAutoConfiguration;
 import org.springframework.boot.builder.SpringApplicationBuilder;
 import zipkin.server.brave.BootstrapTrace;
 
-@SpringBootApplication
+@SpringBootApplication(exclude = JestAutoConfiguration.class)
```
sethp-jive, line 21: I'm sorry about this, but it was the only way I could find to prevent Spring from (eagerly) spinning up an unused JestClient in the application context (with pursuant confusing log messages about its initialization). I suppose we could have found a way to use that configuration, but it doesn't quite meet our needs – there's no hook for adding the signing interceptor, for example. I'd also suggest that makes more sense to have the HttpClient own its own connectivity rather than having it injected, similarly to how the NativeClient handles things.

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/HttpClient.java
```diff
@@ -289,23 +322,18 @@ public void clear(String index) {
     checkState(!"RED".equalsIgnoreCase(status), "Health status is RED");
   }
 
-  private static final class IndicesHealth extends GenericResultAbstractAction {
+  private static final class IndicesHealth extends Health {
```
sethp-jive: I had this change locally because, as I'm sure you noted, my previous attempt to move the method-local class out didn't quite work (the constructor needed the additional setURI call).

The worst part about this is that I think it means the ES tests didn't run for my last PR – all the tests are passing consistently for me locally, but maybe having some trouble in CI. I might need your help to dig into that if it persists.

sethp-jive:
Hmm, my only concern about that spring-aws module is that it looks like it might establish a different set of conventions from the pre-existing AWS client stuff (a rich set of system properties, environment variables, and/or config files that parameterize a bunch of different entrypoints into things like AWS credentials).

For example, in local development it's really convenient to use `AWS_PROFILE=zipkin` and have a role defined that I can switch into that emulates a restricted permission user comparable to what'd be running in production. Do you think that module will match pre-existing expectations?

codefromthecrypt:
> Hmm, my only concern about that spring-aws module is that it looks like it
> might establish a different set of conventions from the pre-existing AWS
> client stuff (a rich set of system properties, environment variables,
> and/or config files that parameterize a bunch of different entrypoints into
> things like AWS credentials).
> 
> For example, in local development it's really convenient to use
> AWS_PROFILE=zipkin and have a role defined that I can switch into that
> emulates a restricted permission user comparable to what'd be running in
> production. Do you think that module will match pre-existing expectations?
> 
> Not sure if it will be a problem or not. Sounds like a valid concern.

pom.xml
```diff
@@ -40,7 +40,8 @@
     <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
 
     <!-- default bytecode version for src/main -->
-    <main.java.version>1.7</main.java.version>
+    <!-- IntelliJ isn't clever enough to have different language levels for tests vs. main code -->
+    <main.java.version>1.8</main.java.version>
```
codefromthecrypt: Sorry, but this isn't a good enough reason. This will break api as it will make change the bytecode level for components that don't define it. Also, it changes the default, which was specifically there.

I do have a workaround, which I paste into the terminal any time dependencies change. That isn't terribly often, though certainly has been in the work you are doing.

``` bash
find . -name '*.iml' -print0 | xargs -0 sed -i '' 's/JDK_1_6/JDK_1_8/g'
find . -name '*.iml' -print0 | xargs -0 sed -i '' 's/JDK_1_7/JDK_1_8/g'
sed -i '' 's/target="1.7"/target="1.8"/g' .idea/compiler.xml
sed -i '' 's/target="1.6"/target="1.8"/g' .idea/compiler.xml
```

ps please upvote https://youtrack.jetbrains.com/issue/IDEA-161060

https://github.com/openzipkin/zipkin/issues/1256#issuecomment-246564227

zipkin-autoconfigure/storage-elasticsearch-aws/src/main/java/zipkin/autoconfigure/storage/elasticsearch/aws/AwsSignatureInterceptor.java
```diff
@@ -0,0 +1,164 @@
...
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Performs aws v4 signing on an apache http request
+ */
+public class AwsSignatureInterceptor implements HttpRequestInterceptor {
```
codefromthecrypt: unless this needs to be public, I wouldn't make it public (rather final and package private). That way we don't have to worry about changing ctor etc.

zipkin-autoconfigure/storage-elasticsearch-http/src/main/java/zipkin/autoconfigure/storage/elasticsearch/http/ZipkinHttpElasticsearchStorageAutoConfiguration.java
codefromthecrypt: thx yeah opposite order of the maven artifact

zipkin-server/src/main/java/zipkin/server/ZipkinServer.java
codefromthecrypt: I'll look into it. I think we need to improve this

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/HttpClient.java
codefromthecrypt: yeah I switched as GenericResultAbstractAction as the Health constructor had a race condition on setting its url. That was scary to me and very dodgy code. I preferred using the base type that wasn't dodgy

zipkin/pom.xml
```diff
@@ -28,7 +28,8 @@
 
   <properties>
     <main.basedir>${project.basedir}/..</main.basedir>
-    <main.java.version>1.6</main.java.version>
+    <!-- IntelliJ isn't clever enough to have different language levels for tests vs. main code -->
+    <main.java.version>1.8</main.java.version>
```
codefromthecrypt: obviously I will revert this. we can't willy nilly change zipkin's core library to java 8! It will break a few libraries

codefromthecrypt (commented):
Thanks for getting this started. Presuming it works, I think the most efficient way is for me to try and unravel some of the forward referencing offline. So, I'll test it and see it pass, and see if I can help with that.

Appreciate all your efforts! Nearly there!

zipkin/pom.xml
sethp-jive: ah, sorry! I thought I was careful to get that out of the way before committing...

pom.xml
sethp-jive: yeah, I'm sorry, I'd been good about keeping this out of the commit (I know it's not going in), but an errant `git commit -a` seems to have undone that effort.

Thanks for the JIRA and the workaround!

codefromthecrypt:
no worries. I'm only sensitive as I've broken JRE compat in two related
zipkin libraries accidentally in the last year :P

codefromthecrypt:
Oh wait.. I forgot :) the intent of the naming convention here is to relate
to the path of properties  what you had was right

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/HttpClient.java
sethp-jive: Was there a race condition? I only saw a janky ordering thing (with the constructor calling setURI out of order with its descendant) but it was limited to the constructing single thread (no leaking `this` outside the class that I saw, at least).

I'll swap the base class out and fix up LenientSearch (which has sorta the same problem, except it has no constructor parameters for now). In that case, though, it seems I just misunderstood the Jest API and that class can go away entirely.

codefromthecrypt:
> Was there a race condition? I only saw a janky ordering thing (with the constructor calling setURI out of order with its descendant) but it was limited to the constructing single thread (no leaking this outside the class that I saw, at least).
> 
>  sorry I meant an ordering problem, which robbed me of time when
> innocently attempting to make the class static. Stuff like that is dodgy to
> me. If a class isn't designed for extension (because of method calls in the
> constructor), it should be marked final.
> 
> I'll swap the base class out and fix up LenientSearch (which has sorta the
> same problem, except it has no constructor parameters for now). In that
> case, though, it seems I just misunderstood the Jest API and that class can
> go away entirely.
> 
> k. lemme know when you are finished polishing I probably won't pull this
> down for an hour or so anyway. Cheers

zipkin-autoconfigure/storage-elasticsearch-http/src/main/java/zipkin/autoconfigure/storage/elasticsearch/http/ZipkinHttpElasticsearchStorageAutoConfiguration.java
```diff
@@ -25,21 +30,61 @@
...
+  /**
+   * Only here for consumption by {@link HostsArentAwsUrls}; ideally we could find a way to express
+   * a hierarchy of autoconfigurations in spring without each parent needing to turn themselves off
+   * for their children, but there seems to be no reliable way to order autoconfigure'd classes
+   * (neither @AutoConfigureBefore nor @AutoConfigureAfter seemd to have the desired effect)
+   */
+  private static final Pattern AWS_URL =
+      Pattern.compile("^https://[^.]+\\.([^.]+)\\.es\\.amazonaws\\.com", Pattern.CASE_INSENSITIVE);
+
+  public static Optional<String> regionFromAwsUrls(List<String> hosts) {
```
codefromthecrypt: I'm having doubts about maintaining code like this. this sort of stuff has breeds bugs. I'd much rather point people to the existing docs and code like use "cloud.aws.region.auto" or "cloud.aws.region.static", which of course we can map to env variables like AWS_REGION.

sethp-jive: I don't disagree, but this seemed the only way to semi-reliably make a really easy path to get started – setting ES_HOSTS=https://my-aws-domain.us-west-2.amazonaws.com was the first thing I tried, and I think there's value in making that sort of thing work. And since the region is encoded in that URL (and is unfortunately required for the AWS v4 signature stuff), it seemed much more straightforward to peel it out of the semi-standard URL format.

That said, it's definitely the kind of magic that breeds complications. Wether it's worth the tradeoff for ease of getting-started, I suppose, is up to you.

sethp-jive:
I have good news and bad news about the test failures – the good news is I can reproduce them locally now that I figured out Circle CI is running elasticsearch 1.7.2! Huzzah!

The bad news is that there seems to be a fundamental incompatibility between Jest for the 2.x elasticsearches and for the 1.x stuff – we might be able to paper over the difference with enough effort, but what do you think about just declaring Zipkin as only supporting elasticsearch >= 2.x?

codefromthecrypt:
zipkin already doesn't support elasticsearch 1! what you've found is likely
a glitch in the circleci config. thanks for finding it

sethp-jive:
Hooray – in that case, I think I'm going to take my green checkmark and declare victory for now. Thank you very much for all your help getting this in!

codefromthecrypt:
cherry-picked 92df918f7ca567b7afa2decf8ed40439ffe9052a independently into master

zipkin-server/src/main/java/zipkin/server/ZipkinServer.java
codefromthecrypt: ah I see what you mean. this isn't a class we defined.

codefromthecrypt:
lol "AwsMagic"

codefromthecrypt:
update: just trying to unravel the tension between the http client and the AWS config. I think I can do it.

codefromthecrypt:
pushed some tidying here https://github.com/openzipkin/zipkin/tree/es-aws
https://travis-ci.org/openzipkin/zipkin/builds/162150722

I provisioned an AWS elasticsearch domain and waiting for it to come up. once it does, I'll confirm things.

codefromthecrypt:
closed via 738b51f

PR #1304, opened 2016-09-17
Unbreaks ElasticsearchStorage.Builder; reduces scope of internal client

codefromthecrypt:
This does a few things to make the recently refactored Elasticsearch
component more maintainable:
- Unbreaks api changes to ElasticsearchStorage.Builder
  - This now delegates to InternalElasticsearchClient.Builder
- Reduces scope of InternalElasticsearchClient
  - Removes dependency writes as this is only used in tests
- Various small cleanups.

See #1303

### DIFF
--- a/zipkin-autoconfigure/storage-elasticsearch/src/main/java/zipkin/autoconfigure/storage/elasticsearch/ZipkinElasticsearchStorageProperties.java
+++ b/zipkin-autoconfigure/storage-elasticsearch/src/main/java/zipkin/autoconfigure/storage/elasticsearch/ZipkinElasticsearchStorageProperties.java
@@ -17,13 +17,12 @@ import java.util.Collections;
 import java.util.List;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import zipkin.storage.elasticsearch.ElasticsearchStorage;
-import zipkin.storage.elasticsearch.NativeClient;
 
 @ConfigurationProperties("zipkin.storage.elasticsearch")
 public class ZipkinElasticsearchStorageProperties {
-  /** @see NativeClient.Builder#cluster(String) */
+  /** @see ElasticsearchStorage.Builder#cluster(String) */
   private String cluster = "elasticsearch";
-  /** @see NativeClient.Builder#hosts(List) */
+  /** @see ElasticsearchStorage.Builder#hosts(List) */
   private List<String> hosts = Collections.singletonList("localhost:9300");
   /** @see ElasticsearchStorage.Builder#index(String) */
   private String index = "zipkin";
@@ -77,9 +76,8 @@ public class ZipkinElasticsearchStorageProperties {
 
   public ElasticsearchStorage.Builder toBuilder() {
     return ElasticsearchStorage.builder()
-        .client(NativeClient.builder()
-            .cluster(cluster)
-            .hosts(hosts).build())
+        .cluster(cluster)
+        .hosts(hosts)
         .index(index)
         .indexShards(indexShards)
         .indexReplicas(indexReplicas);
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/ElasticsearchStorage.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/ElasticsearchStorage.java
@@ -14,36 +14,53 @@
 package zipkin.storage.elasticsearch;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Supplier;
-import com.google.common.collect.Iterables;
-import java.util.Arrays;
-import zipkin.storage.elasticsearch.InternalElasticsearchClient.ClientFactory;
-import zipkin.storage.elasticsearch.InternalElasticsearchClient.HealthStatus;
+import java.util.List;
 import zipkin.storage.guava.LazyGuavaStorageComponent;
 
-import static com.google.common.base.Preconditions.checkState;
-import static com.google.common.collect.Iterables.getOnlyElement;
 import static zipkin.internal.Util.checkNotNull;
 
 public final class ElasticsearchStorage
     extends LazyGuavaStorageComponent<ElasticsearchSpanStore, ElasticsearchSpanConsumer> {
 
-  /**
-   * Internal flag that allows you read-your-writes consistency during tests. With Elasticsearch, it
-   * is not sufficient to block on futures since the index also needs to be flushed.
-   */
-  @VisibleForTesting
-  static boolean FLUSH_ON_WRITES;
-
   public static Builder builder() {
-    return new Builder();
+    return new Builder(new NativeClient.Builder());
+  }
+
+  /**
+   * The client supplier to consume for connectivity to elasticsearch. Defaults to using the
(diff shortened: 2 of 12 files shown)


codefromthecrypt:
cc @sethp-jive

zipkin-autoconfigure/storage-elasticsearch/src/main/java/zipkin/autoconfigure/storage/elasticsearch/ZipkinElasticsearchStorageProperties.java
```diff
@@ -77,9 +76,8 @@ public void setIndexReplicas(int indexReplicas) {
 
   public ElasticsearchStorage.Builder toBuilder() {
     return ElasticsearchStorage.builder()
-        .client(NativeClient.builder()
-            .cluster(cluster)
-            .hosts(hosts).build())
+        .cluster(cluster)
```
codefromthecrypt, line 79: this unbreaks the api, which is helpful :)

zipkin-autoconfigure/storage-elasticsearch/src/main/java/zipkin/autoconfigure/storage/elasticsearch/ZipkinElasticsearchStorageProperties.java
```diff
@@ -17,13 +17,12 @@
 import java.util.List;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import zipkin.storage.elasticsearch.ElasticsearchStorage;
-import zipkin.storage.elasticsearch.NativeClient;
```
codefromthecrypt, line 20: this doesn't need to know about NativeClient

zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/ElasticsearchStorage.java
```diff
@@ -14,38 +14,55 @@
...
+
+    final InternalElasticsearchClient.Builder clientBuilder;
     String index = "zipkin";
     int indexShards = 5;
     int indexReplicas = 1;
 
     /**
+     * The elasticsearch cluster to connect to, defaults to "elasticsearch".
+     */
+    public Builder cluster(String cluster) {
```
codefromthecrypt, line 51: resurrects these so no api break

zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/ElasticsearchStorage.java
```diff
@@ -80,22 +97,15 @@ public Builder indexReplicas(int indexReplicas) {
...
 
-    /**
-     * The client supplier to consume for connectivity to elasticsearch. Defaults to using the
-     * transport-client based {@link NativeClient}, but can be overriden with a HTTP-speaking client
-     * (e.g. for use in cloud environments where the transport protocol is not exposed).
-     */
-    public Builder client(ClientFactory clientFactory) {
-      this.clientFactory = clientFactory;
+    // punch a hole so that tests don't need to share a static variable
+    @VisibleForTesting  Builder flushOnWrites(boolean flushOnWrites) {
```
codefromthecrypt, line 101: better than sharing a static across multiple implementations

codefromthecrypt:
merging to keep trunk shippable and @sethp-jive unblocked

@sethp-jive feel free to change the internal interface as you need to in your next PR

zipkin-storage/elasticsearch/src/test/java/zipkin/storage/elasticsearch/ElasticsearchTestGraph.java
```diff
@@ -13,27 +13,25 @@
...
     @Override protected ElasticsearchStorage compute() {
       if (ex != null) throw ex;
-      ElasticsearchStorage result =
-          new ElasticsearchStorage.Builder().index("test_zipkin_native").build();
+      ElasticsearchStorage result = ElasticsearchStorage.builder()
+          .index("test_zipkin_native").flushOnWrites(true).build();
       CheckResult check = result.check();
       if (check.ok) return result;
-      throw ex = new AssumptionViolatedException("cluster was not healthy", check.exception);
+      throw ex = new AssumptionViolatedException(check.exception.getMessage());
```
sethp-jive, line 34: Would you help me understand the logic behind this line? I found it helpful to have the full exception + stacktrace when my problem was an NPE (e.g. accessing the wrong field in a JSON response) – I'm happy to keep it just the message, but if this change shows up again it means I hit another NPE and forgot to back it out before committing :)

codefromthecrypt: copy/paste error!

codefromthecrypt: here's a fix (for all providers) https://github.com/openzipkin/zipkin/pull/1305

Note I didn't presume a summary message as the cause for skip could be anything. Nice catch

zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/InternalElasticsearchClient.java
```diff
@@ -55,17 +86,18 @@
...
-  interface Buckets {
+  protected interface Buckets {
     /**
      * Retrieves the aggregation bucket keys from the aggregation at name[>nestedPath]
      *
      * The elasticsearch store only uses this one attribute of any aggregations, and so this
      * interface wraps the ugliness of dealing with different aggregation APIs while leaving control
      * of the aggregation keys in the hands of the caller.
      */
+    // TODO(adrian): revisit once the other implementation is in: can we avoid varags?
```
sethp-jive, line 100: We totally can – the varargs is just there to implement one-or-many semantics. Either this is a first-level aggregation, or it's nested "arbitrarily" deep (in practice, we only have two levels). This seemed clearer from both the implementation and user side to me than two methods (e.g. `getKeys` and `getNestedKeys`) .

zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/InternalElasticsearchClient.java
```diff
@@ -85,32 +117,18 @@
...
-    final String id;
-    final Map<String, ?> data;
-
-    public IndexableLink(String id, Map<String, ?> data) {
-      this.id = id;
-      this.data = data;
-    }
-  }
+  // TODO(adrian): revisit once the other implementation is in: why we are using arrays vs lists?
+  protected abstract ListenableFuture<Collection<DependencyLink>> findDependencies(String[] indices);
```
sethp-jive, line 121: The java elasticsearch client API uses varargs for its index lists; this allows a direct pass-through on that stuff and a zero-copy single-allocation wrapper when we need a list (i.e. `Arrays#asList`).

Alternatively, we can take a list here, but then we'd have to do the `indices.toArray(new String[indices.size()])` dance before handing them off to the elasticsearch client.

PR #1306, opened 2016-09-20
storage: provide http-based elasticsearch client

sethp-jive:
Taking advantage of & expanding the previous commit's shim, this chnage
introduces an elasticsearch client that operates over the HTTP(S)
protocol. This commit's functional changes are minimal, limited to
restricting flushes on a few operations to the desired indices
instead of globally.


zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/InternalElasticsearchClient.java
```diff
@@ -48,7 +48,7 @@
   protected static final String DEPENDENCY_LINK = "dependencylink";
 
   protected interface Factory {
-    InternalElasticsearchClient create();
+    InternalElasticsearchClient create(String allIndices);
```
sethp-jive, line 51: I had to bring this back to support the index-query-rewriting thing I'm doing in the RestClient (see the javadoc attached to `MAX_INDICES`). It's definitely ugly, but IMO less so than making the client check the # of indices and do the query rewriting themselves.

zipkin-storage/elasticsearch-jest/src/test/java/zipkin/storage/elasticsearch/jest/ElasticsearchJestDependencyWriter.java
```diff
@@ -0,0 +1,47 @@
...
+import io.searchbox.indices.Flush;
+import java.io.IOException;
+import java.util.List;
+import zipkin.DependencyLink;
+import zipkin.storage.elasticsearch.InternalElasticsearchClient;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+public class ElasticsearchJestDependencyWriter {
+  public static void writeDependencyLinks(InternalElasticsearchClient genericClient,
```
codefromthecrypt: I'd prefer this moved to be inside the class of its only caller

zipkin-storage/elasticsearch-jest/src/test/java/zipkin/storage/elasticsearch/jest/ElasticsearchJestTestGraph.java
```diff
@@ -0,0 +1,40 @@
...
+
+        @Override protected ElasticsearchStorage compute() {
+          if (ex != null) throw ex;
+          ElasticsearchStorage result =
+              ElasticsearchStorage.builder(new RestClient.Builder()
+                  .hosts(ImmutableList.of("http://localhost:9200")).flushOnWrites(true))
+                  .index("test_zipkin_rest").build();
+          CheckResult check = result.check();
+          if (check.ok) return result;
+          throw ex = new AssumptionViolatedException("cluster was not healthy", check.exception);
```
codefromthecrypt: check.exception.getMessage() instead of "cluster was not healthy" as the failure could be unrelated to the cluster

zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/NativeClient.java
```diff
@@ -149,9 +149,9 @@
   }
 
   final TransportClient client;
-  final boolean flushOnWrites;
+  private final boolean flushOnWrites;
```
codefromthecrypt: not required to change, but just so you know why I didn't mark these private. This is a package protected final internal class anyway, so we have no contract to uphold wrt changing thing. each private thing has potential to bloat class sizes and mixed visibility is distracting.

If you really want these things to be invisible to the package, a quicker and less bloaty way to do it is just to make the type private. At that point, you can't reference the fields and constructors anyway, so default visibility is fine. You'll notice this practice used in grpc I think..

zipkin-storage/elasticsearch/src/test/java/zipkin/storage/elasticsearch/ElasticsearchSpanConsumerTest.java
```diff
@@ -14,6 +14,7 @@
 package zipkin.storage.elasticsearch;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ObjectArrays;
```
codefromthecrypt, line 17: imports here are likely a mistake as I don't see any other changes

zipkin-autoconfigure/storage-elasticsearch/src/main/java/zipkin/autoconfigure/storage/elasticsearch/ZipkinElasticsearchStorageAutoConfiguration.java
```diff
@@ -13,19 +13,69 @@
  */
 package zipkin.autoconfigure.storage.elasticsearch;
 
+import com.google.common.base.MoreObjects;
```
codefromthecrypt: are all these imports used?

zipkin-storage/elasticsearch-jest/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+Implements a client provider for Elasticsearch that uses Elasticsearch's HTTP APIs by way of the [Jest client](https://github.com/searchbox-io/Jest) for flexibility.
+
+See the main elasticsearch module for more details.
+
+## Testing this component
+This module conditionally runs integration tests against a local Elasticsearch instance.
+
+Tests are configured to automatically access Elasticsearch started with its defaults.
+To ensure tests execute, download an Elasticsearch 2.x distribution, extract it, and run `bin/elasticsearch`.
```
codefromthecrypt: does this work on elasticsearch 5?

zipkin-storage/elasticsearch-jest/src/main/java/zipkin/storage/elasticsearch/jest/RestClient.java
```diff
@@ -0,0 +1,398 @@
...
+  private <T extends JestResult> T executeUnchecked(Action<T> action) {
+    return getUnchecked(toGuava(action));
+  }
+
+  private enum SpanDeserializer implements JsonDeserializer<Span> {
+    INSTANCE;
+
+    @Override
+    public Span deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context)
+        throws JsonParseException {
```
codefromthecrypt: probably don't need throws here and below

codefromthecrypt (approved):
Looks very good. only some cursory changes.

That said, I'm thinking about re-evaluating the choice I suggested to name this "elasticsearch-jest" This will end up being in server configuration, and I don't expect us to support multiple libraries for elasticsearch http. Similar to how we don't call zipkin-storage-mysql zipkin-storage-jooq, I think we should name this with an "http" suffix afterall.

If I have edit access to this branch, I can change this, if it is a pain..

@anuraaga @shakuzen wdyt?

zipkin-autoconfigure/storage-elasticsearch/src/main/java/zipkin/autoconfigure/storage/elasticsearch/ZipkinElasticsearchStorageAutoConfiguration.java
sethp-jive: nope, good catch!

zipkin-autoconfigure/storage-elasticsearch-jest/src/main/java/zipkin/autoconfigure/storage/elasticsearch/jest/ZipkinElasticsearchJestStorageProperties.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.Collections;
+import java.util.List;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import zipkin.storage.elasticsearch.ElasticsearchStorage;
+import zipkin.storage.elasticsearch.jest.RestClient;
+
+@ConfigurationProperties("zipkin.storage.elasticsearch")
+public class ZipkinElasticsearchJestStorageProperties {
+  /** @see ElasticsearchStorage.Builder#hosts(List) */
```
shakuzen: Per convention, JavaDoc comments on `@ConfigurationProperties` fields are plain text (don't use JavaDoc keywords like `@see`) because they are used by the configuration processor and then by IDEs as descriptions of the fields during auto-completion, for instance.

zipkin-storage/elasticsearch-jest/README.md
sethp-jive: Fairly well – they got rid of the `string` data type (I think we want `keyword`, since we're treating ES as a big `String.equals` engine and not doing "full text" search), but with a few changes I can get many of the tests to pass. The rest fail because ES goes into overdrive and pegs my CPU without returning any results within 10 seconds. The errors it does log suggest I've missed some important fields (i.e. "traceId" is still "text" and not "keyword"), but I stopped there for now.

zipkin-storage/elasticsearch-jest/src/main/java/zipkin/storage/elasticsearch/jest/RestClient.java
sethp-jive: Removed!

zipkin-storage/elasticsearch-jest/src/test/java/zipkin/storage/elasticsearch/jest/ElasticsearchJestTestGraph.java
sethp-jive: Yup, good catch – I've fixed it up to look like the others in #1305

zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/NativeClient.java
sethp-jive: I changed it back – as you say, not a big deal, IntelliJ just happened to mention that it could be a weaker visibility & so I made the warning go away :)

zipkin-storage/elasticsearch/src/test/java/zipkin/storage/elasticsearch/ElasticsearchSpanConsumerTest.java
sethp-jive: Yup, cleaned up now!

zipkin-storage/elasticsearch-jest/src/main/java/zipkin/storage/elasticsearch/jest/RestClient.java
```diff
@@ -0,0 +1,398 @@
...
+    /**
+     * Unnecessary, but we'll check it's not null for consistency's sake.
+     */
+    @Override public Builder cluster(String cluster) {
+      checkNotNull(cluster, "cluster");
+      return this;
+    }
+
+    /**
+     * A comma separated list of elasticsearch hostnodes to connect to, in http://host:port or
```
shakuzen: Not a comma-separated list but an actual `List`

sethp-jive (commented):
I made the changes you suggested, as well as fumbled in the dark a bit on the test failures (you'll be unsurprised to hear they pass locally just fine). 

I did go ahead and rename  elasticsearch-jest to elasticsearch-http, but kept it in a separate commit that should be easy to cut out of the history if we want to go back.

codefromthecrypt:
This one is my fault I think. So is the best behaviour to copy/paste a
one-liner?

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/jest/RestClient.java
```diff
@@ -0,0 +1,397 @@
...
+        });
+  }
+
+  @Override
+  public ListenableFuture<Void> indexSpans(List<IndexableSpan> spans) {
+    if (spans.isEmpty()) return Futures.immediateFuture(null);
+
+    // Create a bulk request when there is more than one span to store
+    ListenableFuture<?> future;
+    final Set<String> indices = new HashSet<>();
```
codefromthecrypt: nit: please use LinkedHashSet/Map always :) order bugs in tests aren't worth tracking down, and such things often happen well after the first commit.

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/jest/RestClient.java
```diff
@@ -0,0 +1,397 @@
...
+
+  private RestClient(Factory f, String allIndices) {
+    JestClientFactory factory = new JestClientFactory();
+    factory.setHttpClientConfig(new HttpClientConfig.Builder(f.hosts)
+        .defaultMaxTotalConnectionPerRoute(6) // matches "regular" TransportClient node conns
+        .maxTotalConnection(6 * 10) // would be 20 otherwise, or ~3 routes
+        .connTimeout(Ints.checkedCast(TimeUnit.MILLISECONDS.convert(10, TimeUnit.SECONDS)))
+        .readTimeout(Ints.checkedCast(TimeUnit.MILLISECONDS.convert(10, TimeUnit.SECONDS)))
+        .multiThreaded(true)
+        .gson(
```
codefromthecrypt: heh, so jest not only binds to apache hc, but also gson?

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/jest/RestClient.java
```diff
@@ -0,0 +1,397 @@
...
+  @VisibleForTesting final JestClient client;
+  private final String[] allIndices;
+  private final boolean flushOnWrites;
+
+  private RestClient(Factory f, String allIndices) {
+    JestClientFactory factory = new JestClientFactory();
+    factory.setHttpClientConfig(new HttpClientConfig.Builder(f.hosts)
+        .defaultMaxTotalConnectionPerRoute(6) // matches "regular" TransportClient node conns
+        .maxTotalConnection(6 * 10) // would be 20 otherwise, or ~3 routes
+        .connTimeout(Ints.checkedCast(TimeUnit.MILLISECONDS.convert(10, TimeUnit.SECONDS)))
```
codefromthecrypt: pretty heavy imho. why not 10 \* 1000? this is fairly obvious milliseconds and the compiler will fold the constant

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/jest/RestClient.java
```diff
@@ -0,0 +1,397 @@
...
+          return toGuava(new Flush.Builder().addIndex(indices).build());
+        }
+      });
+    }
+
+    return transform(future, Functions.<Void>constant(null));
+  }
+
+  @Override protected void ensureClusterReady(final String catchAll) {
+    class IndicesHealth extends Health {
```
codefromthecrypt: method scoped classes creep me out. they can easily accidentally leak references to the enclosing object. mind making this a static inner class?

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/jest/RestClient.java
```diff
@@ -0,0 +1,397 @@
...
+        return super.buildURI() + catchAll;
+      }
+    }
+
+    String status = executeUnchecked(new IndicesHealth(new Health.Builder()))
+        .getJsonObject().get("status").getAsString();
+    checkState(!"RED".equalsIgnoreCase(status), "Health status is RED");
+  }
+
+  private Index toIndexRequest(IndexableSpan span) {
```
codefromthecrypt: static utility methods unless they require access to the enclosing object. makes it easier to see impact.

zipkin-storage/elasticsearch/src/test/java/zipkin/storage/elasticsearch/ElasticsearchTestGraph.java
```diff
@@ -28,7 +26,7 @@
     @Override protected ElasticsearchStorage compute() {
       if (ex != null) throw ex;
       ElasticsearchStorage result = ElasticsearchStorage.builder()
-          .index("test_zipkin_native").flushOnWrites(true).build();
+          .index("test_zipkin_" + System.currentTimeMillis()).flushOnWrites(true).build();
```
codefromthecrypt: please revert this. I don't want my ES polluted

codefromthecrypt (commented):
found a few other things on second look

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/jest/RestClient.java
sethp-jive: I'll change it, but as to the "why" – I would've preferred the java 8 `Duration.ofSeconds(10).toMillis()` (and the checked cast is there because apparently Jest uses ints for time instead of longs), but this is the closest I could come in Java 7.

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/jest/RestClient.java
sethp-jive: Yeah, but due to repackaging it's a different gson than zipkin core uses.

zipkin-autoconfigure/storage-elasticsearch-http/src/main/java/zipkin/autoconfigure/storage/elasticsearch/http/ZipkinElasticsearchHttpStorageProperties.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.List;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import zipkin.storage.elasticsearch.ElasticsearchStorage;
+import zipkin.storage.elasticsearch.http.RestClient;
+
+@ConfigurationProperties("zipkin.storage.elasticsearch")
+public class ZipkinElasticsearchHttpStorageProperties {
+  /** The elasticsearch cluster to connect to, defaults to "elasticsearch". */
+  private String cluster = "elasticsearch";
+  /** A List of hosts to connect to, e.g.  */
```
codefromthecrypt, line 26: "A List of base http urls to connect to, e.g. http://localhost:9200" right?

zipkin-autoconfigure/storage-elasticsearch-http/src/main/java/zipkin/autoconfigure/storage/elasticsearch/http/ZipkinElasticsearchHttpStorageProperties.java
```diff
@@ -0,0 +1,76 @@
...
+  private List<String> hosts = Collections.singletonList("localhost:9300");
```
codefromthecrypt, line 27: http://localhost:9200, right?

zipkin-autoconfigure/storage-elasticsearch/src/main/java/zipkin/autoconfigure/storage/elasticsearch/ZipkinElasticsearchStorageAutoConfiguration.java
```diff
@@ -13,19 +13,65 @@
...
+        String hosts =
+            conditionContext.getEnvironment().getProperty("zipkin.storage.elasticsearch.hosts");
+        return !hostsAreUrls(firstNonNull(hosts, ""));
+      }
+    }
+
+    private static boolean hostsAreUrls(String hostsProperty) {
+      List<String> hosts = Arrays.asList(hostsProperty.split(","));
+      try {
+        URL url = new URL(Iterables.getFirst(hosts, ""));
```
codefromthecrypt, line 68: :) I used to over-use guava like this.. it is hard to realize when it is happening.

zipkin-storage/elasticsearch/src/test/java/zipkin/storage/elasticsearch/ElasticsearchSpanConsumerTest.java
codefromthecrypt: ehh.. still here :P

codefromthecrypt (approved):
I'll do some last tidying and squash the interim commits on the way in (leaving the separate one for http as you mentioned)

great work

codefromthecrypt:
cherry-picked in. take note of a few conventional file renames (also it is now plumbed to zipkin-server), so the following works:

``` bash
$ STORAGE_TYPE=elasticsearch ES_HOSTS=http://localhost:9200 java -jar zipkin-server/target/zipkin-server-1.11.2-SNAPSHOT-exec.jar
```

PR #1308, opened 2016-09-20
WIP -- Add a storage backend that works for DSE-5.0

michaelsembwever:
Add a storage backend that works for DSE-5.0
  This would be helpful for those that can't upgrade to cassandra-3.9

### DIFF
--- a/zipkin-storage/dse5/README.md
+++ b/zipkin-storage/dse5/README.md
@@ -1,0 +1,51 @@
+# storage-dse5
+
+*This module is experimental. Please help test this, but do not use it in production.*
+
+This CQL-based DSE 5.0+ storage component includes a `GuavaSpanStore` and `GuavaSpanConsumer`.
+`GuavaSpanStore.getDependencies()` returns pre-aggregated dependency links (ex via [zipkin-dependencies](https://github.com/openzipkin/zipkin-dependencies)).
+
+The implementation uses the [Datastax Java Driver 3.x](https://github.com/datastax/java-driver).
+
+`zipkin.storage.cassandra3.Cassandra3Storage.Builder` includes defaults that will operate against a local Cassandra installation.
+
+## Logging
+Queries are logged to the category "com.datastax.driver.core.QueryLogger" when debug or trace is
+enabled via SLF4J. Trace level includes bound values.
+
+See [Logging Query Latencies](http://docs.datastax.com/en/developer/java-driver/3.0/supplemental/manual/logging/#logging-query-latencies) for more details.
+
+## Testing
+This module conditionally runs integration tests against a local Cassandra instance.
+
+Tests are configured to automatically access Cassandra started with its defaults.
+To ensure tests execute, download a DSE 5.0+ distribution, extract it, and run `dse start`.
+
+If you run tests via Maven or otherwise when Cassandra is not running,
+you'll notice tests are silently skipped.
+```
+Results :
+
+Tests run: 62, Failures: 0, Errors: 0, Skipped: 48
+```
+
+This behaviour is intentional: We don't want to burden developers with
+installing and running all storage options to test unrelated change.
+That said, all integration tests run on pull request via Travis.
+
+### Trace indexing
+
+Indexing in CQL is simplified by using both MATERIALIZED VIEWs and DSE Solr Cores.
+
+### Time-To_live
+Time-To-Live is default now at the table level. It can not be overridden in write requests.
+
+### Compaction
+Time-series data is compacted using TimeWindowCompactionStrategy, a known improved over DateTieredCompactionStrategy. Data is
+optimised for queries with a single day. The penalty of reading multiple days is small, a few disk seeks, compared to the
+otherwise overhead of reading a significantly larger amount of data.
+
+### Benchmarking
+Benchmarking the new datamodel demonstrates a significant performance improvement on reads. How much of this translates to te
+Zipkin UI is hard to tell due to the complexity of CassandraSpanConsumer and how searches are possible. Benchmarking stress
+profiles are found in traces-stress.yaml and trace_by_service_span-stress.yaml.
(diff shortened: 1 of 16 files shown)


michaelsembwever:
Basic thoughts and a skeleton for using Zipkin with the new style of Cassandra datamodel, but replacing the SASI with Solr Cores. This allows the data model to run against Cassandra-3.0.8+ but does require a DSE installation.

A fair bit of work still has to happen here, wiring it together, etc.
But the basic idea is in place.

codefromthecrypt:
interested, as I'm interested in decommissioning the "cassandra" module, but I don't think this gets us there, as not everyone uses DSE. Basically, we end up with 3 codebases for cassandra that don't share code.

I know @schrepfler asked about DSE, but not sure if he uses zipkin at the moment or not. Do we have a good gauge of how many use DSE?

I'm concerned about the maintenance problem of bringing this into the core repo, particularly as it is getting bloated and it doesn't seem obvious how to test commercial variants like this. There's also bus factor.. we're lucky we have you, but there's few others working on cassandra, almost none. Can we afford to have a third codebase? Finally, we already blow up on memory often as it is and our builds are approaching 20minutes.. Feels like it should be a separate repo.

@openzipkin/cassandra @abesto What are your thoughts?

michaelsembwever:
Personally I would rather not include the code into master, for the same reasons you describe @adriancole 

But I did want to at least throw the code out here, so that others could take it (if only the idea) and run if they wanted to.

codefromthecrypt:
@michaelsembwever I noticed this has a dep on cassandra3. seems `CassandraSpanConsumer` is the same type, but it has a different front-end (SpanStore) and schema.

Am I right to think that this is "effectively" the same schema? Else, I'd assume you wouldn't be able to use cassandra3's `CassandraSpanConsumer`

While this PR is >2K lines, it seems some of it is benign (ex UDT related cruft) or some things that have to be copy/pasted. The heaviest class seems to be the SpanStore variant. Is that a fair assessment?

michaelsembwever:
Yes it is the same schema as far as writes go. Just that the SASI are not used. DSE Solr Cores replace them.

codefromthecrypt:
> Personally I would rather not include the code into master, for the same
> reasons you describe @adriancole https://github.com/adriancole
> 
> But I did want to at least throw the code out here, so that others could
> take it (if only the idea) and run if they wanted to.
> 
> Historically, most of the repos we spun out from concerns about interest
> didn't turn into anything. Ex mongo and redis went to separate repos to
> die. However, I think it is more reasonable that this could live. I'd offer
> effort to carve out a repo for this.

We could attract attention to it, saying that once cassandra3 is done,
we'll deprecate "cassandra" for removal at some point. Those who can't
upgrade on account of DSE should look to the zipkin-storage-dse5 repo. If
no-one is interested enough even to comment there, then there's not much
else we can do... worst case, it dies.

codefromthecrypt:
Thanks for answering my questions, @michaelsembwever! I'll shut up and see what others think

schrepfler:
I would go with external repo and some sort of contrib model and zipkin-storage-dse5 sounds fine (ideally Datastax themselves would be maintainers as they know their features best and it would advocate optimal use of their product.)
Is the DSE Solr interface so much different when starting Cassandra with the parameters to turn on string indexing using ES or Solr (I was under the impression Cassandra ships with such an example config but I can't find it in my instance, perhaps I've seen it in use on some modified Cassandra ex. Titan+Cassandra+ES)?

michaelsembwever:
> Is the DSE Solr interface so much different when starting Cassandra with the parameters to turn on string indexing using ES or Solr …

@schrepfler i'm only familiar with the rather basic approach of using `dsetool create_core …` from the command line to create the Solr Core off a C\* table.

schrepfler:
I think that's the way to go too, haven't dealt with it directly though.

danchia:
AFAIK there's the command line client, and a deprecrated? HTTP endpoint. Sadly can't create the indexes from CQL directly since you need the solr config.

zipkin-storage/dse5/README.md
```diff
@@ -0,0 +1,51 @@
...
+### Time-To_live
+Time-To-Live is default now at the table level. It can not be overridden in write requests.
+
+### Compaction
+Time-series data is compacted using TimeWindowCompactionStrategy, a known improved over DateTieredCompactionStrategy. Data is
+optimised for queries with a single day. The penalty of reading multiple days is small, a few disk seeks, compared to the
+otherwise overhead of reading a significantly larger amount of data.
+
+### Benchmarking
+Benchmarking the new datamodel demonstrates a significant performance improvement on reads. How much of this translates to te
```
danchia, line 49: there's a typo here. `translates to the`

danchia: Oops sorry, hit finish review too early. Meant to finish my pass later..

zipkin-storage/dse5/src/main/java/zipkin/storage/cassandra3/Dse5Schema.java
```diff
@@ -0,0 +1,332 @@
...
+      this.ts = annotation.timestamp;
+      this.v = annotation.value;
+      this.ep = annotation.endpoint != null ? new EndpointUDT(annotation.endpoint) : null;
+    }
+
+    public long getTs() {
+      return ts;
+    }
+
+    public String getV() {
```
danchia, line 208: Minor readability suggestion (@adriancole probably can comment more) but I prefer to not have abbreviations that are not generally well known in public APIs. e.g `ep` probably don't mean much to most people

zipkin-storage/dse5/src/main/resources/dse5-schema.cql
```diff
@@ -0,0 +1,70 @@
...
+);
+
+CREATE TYPE IF NOT EXISTS zipkin3.annotation (
+    ts bigint,
+    v  text,
+    ep frozen<endpoint>
+);
+
+CREATE TYPE IF NOT EXISTS zipkin3.binary_annotation (
+    k  text,
```
danchia, line 17: I don't recall if in Cassandra 3.x whether long column names inflates storage use, but I would encourage not using abbreviations.

danchia (commented):
Maybe I missed it, but perhaps include the solr schema and / or the dsetool command to create the solr core?

PR #1309, opened 2016-09-21
Deprecates Endpoint.create(String,int,int) as it leads to null pointers

codefromthecrypt:
Multiple people have been bitten by `Endpoint.create(String,int,int)` as
`Endpoint.port` is nullable. This deprecates that method and advises to
use a better documented builder method or simply `Endpoint.toBuilder()`
when copying an endpoint.


codefromthecrypt:
cc @devinsba

I'll make a similar change to brave, too

codefromthecrypt:
PS in hindsight I wouldn't have defined port as a short, rather left that to codec.

This class used to be a literal struct, and latter changed to decouple from field lengths. In zipkin 2 we could break api and change port accordingly, if we felt the need to. For now, this change should help mitigate the concern.

Issue #1312, opened 2016-09-25
Support Elasticsearch 5


codefromthecrypt:
Lets work towards elasticsearch 5, from both a functionality perspective and also a classpath-dodging one. For example, our server is used as a base for others. We need to be careful about class conflicts with ES libraries, and as always guava.

Here's a note about status quo on elasticsearch-http and ES5 from @sethp-jive

> Fairly well – they got rid of the string data type (I think we want keyword, since we're treating ES as a big String.equals engine and not doing "full text" search), but with a few changes I can get many of the tests to pass. The rest fail because ES goes into overdrive and pegs my CPU without returning any results within 10 seconds. The errors it does log suggest I've missed some important fields (i.e. "traceId" is still "text" and not "keyword"), but I stopped there for now.

codefromthecrypt:
cc @openzipkin/elasticsearch ps I've started resurrecting my old okhttp work towards this. The client design that @sethp-jive made makes it much lighter effort.

I'm personally interested in an okhttp option for a number of reasons:
- We want a single library for ES http 2-5 if possible. Each dependency makes life difficult, and I'm pretty sure ES 2 will be a problem in the same classpath as ES 5.
- Similar, but focused on Jest. Library compatibility bites me in every project used as platform code and jest uses many of the usual suspects of api compat problems: guava, gson and apachehc. OkHttp (and similar like moshi) have the same dependency policy: switch package names when you break api. 
- I really don't like how inefficient the apachehc stuff is, especially when using the AWS interceptor, it literally copies the data when we don't need to. While less concerning for query, we should be careful about performance on the write side.
- We want self-tracing for ES performance and behavior debugging. OkHttp already has a brave integration, as well a way to tag calls.
- And should be said.. I like to enjoy coding and OkHttp is much more fun than worrying about AWS SDK library gotchas, terrible apachehc interceptors, things like constructor order in Jest, or other indirection.

codefromthecrypt:
I've completed the okhttp storage layer for elasticsearch 2.x (haven't looked into 5.x portability, yet). I've also written a tracing interceptor so that we can debug the performance. 
https://github.com/openzipkin/brave/pull/242

After I get an okhttp aws signature v4 interceptor together, I'll raise a pull request. should be tomorrow or next day. here's a peek for now.

[...]

PR #1313, opened 2016-09-25
Cleans up InternalElasticsearchClient and fixes aggregation bug

codefromthecrypt:
This cleans up InternalElasticsearchClient in preparation of
Elasticsearch 5 work.

Formerly, the client implementation was required to search the response
of a request for specific paths of aggregation bucket keys. This
required an intermediary to allow the same response to be read for two
paths. Since implementations are trusted, the change is that we search
for "any" bucket keys. This is simpler to test and requires less
coordination.

On that note, there was a bug in the aggregation logic, where annotation
aggregations weren't treated as nested. This fixes the bug and rewrites
the test to expose it in the future.

In order to plug-in a different span indexer, I needed to re-organize
that code. Now, there's a BulkSpanIndexer type which can be exteded as
needed.

See #1312

### DIFF
--- a/zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/BucketKeysTest.java
+++ b/zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/BucketKeysTest.java
@@ -1,0 +1,160 @@
+/**
+ * Copyright 2015-2016 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin.storage.elasticsearch.http;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import io.searchbox.core.SearchResult;
+import org.junit.Test;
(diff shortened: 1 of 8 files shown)


codefromthecrypt:
cc @openzipkin/elasticsearch ps this is net less main code, the increase in lines is due to backfilling tests on the aggregation bucket code which had a bug

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/BucketKeysTest.java
```diff
@@ -0,0 +1,160 @@
...
+        + "        \"sum_other_doc_count\": 0,\n"
+        + "        \"buckets\": [\n"
+        + "          {\n"
+        + "            \"key\": \"yak\",\n"
+        + "            \"doc_count\": 1\n"
+        + "          }\n"
+        + "        ]\n"
+        + "      }\n"
+        + "    },\n"
+        + "    \"annotations_agg\": {\n"
```
codefromthecrypt, line 104: before, we weren't returning the nested annotations agg.. now we are, so it properly merges

zipkin/src/test/java/zipkin/storage/SpanStoreTest.java
```diff
@@ -205,9 +205,10 @@ public void getSpanNames_allReturned() {
   }
 
   @Test
-  public void getAllServiceNames() {
+  public void getAllServiceNames_mergesAnnotation_andBinaryAnnotation() {
+    // creates a span with mutual exclusive endpoints in binary annotations and annotations
     BinaryAnnotation yak = BinaryAnnotation.address("sa", Endpoint.create("yak", 127 << 24 | 1));
-    accept(span1.toBuilder().addBinaryAnnotation(yak).build(), span4);
+    accept(span1.toBuilder().binaryAnnotations(asList(yak)).build());
```
codefromthecrypt, line 211: before, a binary annotation had the same service name as a regular annotation, masking the bug where we weren't querying the latter

PR #1320, opened 2016-09-30
Rewrites elasticsearch-http as an OkHttp client

codefromthecrypt:
This does the following:
- reduces dependency lockup by relying on okhttp and moshi vs specific versions of apachehc, guava and gson
- allows self-tracing to work by using brave instrumented interceptors


codefromthecrypt:
most of the additional lines (~750) here are from resurrecting the dependency link and span moshi adapters and tests.

codefromthecrypt:
@openzipkin/elasticsearch I need to backfill some unit tests, but I've tested this manually against local and AWS elasticsearch and 👍 

Note: some of the new code here is self-tracing, which we never had before.

screen shot 2016-09-30 at 3 59 22 pm

codefromthecrypt:
didn't get much progress on this today. maybe tomorrow

codefromthecrypt:
ok ready to go now. once this is in, we can start doing more unwinding, such as ES 5 interop.

PR #1323, opened 2016-09-30
README polish

shakuzen:
Use relative links instead of absolute links where possible.
Fix some broken links in the READMEs.


codefromthecrypt (approved):
thanks!

PR #1324, opened 2016-10-01
Adds Endpoint.toString in json form

codefromthecrypt:
This is helpful for debugging instrumentation, particularly their local
endpoints.

See https://github.com/openzipkin/brave/pull/245


codefromthecrypt:
cc @schlosna this is a nicer way to fix brave, imho.. just convert brave's endpoint to this, then toString it.

PR #1326, opened 2016-10-04
Defaults UI to expand all

### LINKED ISSUE #1046: [ui] Make "Expand All" the default view for a trace
Transferring one of our internal tickets. Many users have asked for trace to be fully expanded by default, unless it exceeds a certain N of spans. Especially new users often get confused by seeing incomplete traces by default.

codefromthecrypt:
Routinely, we ask users to click "Expand All" when they are having
trouble with the UI. This sometimes is the only thing they need to do.

The default to "Collapse All" optimizes for very large traces, but we've
no history of what a good number would be (or even a nice place to code
that number). Also, we've never heard anyone defend the "Collapse All"
default despite multiple complaints asking to change it.

This changes the default trace view to expand all traces, optimizing for
those who are starting with Zipkin or have traces that aren't huge. The
impact is that those who have huge traces might want to click "Collapse
All" and/or report back or raise a pull request to have an appropriate
behavior with a default backed up by some user experience.

Fixes #1046

### DIFF
--- a/zipkin-ui/templates/trace.mustache
+++ b/zipkin-ui/templates/trace.mustache
@@ -9,8 +9,8 @@
 
   <form class='form-inline' role='form'>
     <div class='btn-group btn-group-sm' id='filterAllServices'>
-      <button type='button' class='btn btn-default' value='uiExpandAllSpans'>Expand All</button>
-      <button type='button' class='btn btn-default active' value='uiCollapseAllSpans'>Collapse All</button>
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
tested locally and has the intended behavior

screen shot 2016-10-04 at 9 53 00 am

jcarres-mdsol:
+1 
For newcomers, this has always been confusing

mansu:
+1 for fixing this fast! This is confusing people who are new to the UI.

PR #1328, opened 2016-10-04
Adds collector_timestamp_millis to spans stored in elasticsearch

### LINKED ISSUE #1318: Can we help uncover or explain span reporting lag?
from @dan-tr

> @dan-tr Sep 28 23:09
> Is timestamp_millis created based on the timestamp field or is created based on the current time when the record is inserted into ES? I'm looking for away to discern if there is any lag time between the message being put on the Kafka queue (client-side) and the record being stored in ES (server-side).

We should add the description to the README here, but no. this timestamp isn't used for lag, it is based on the best timestamp.
https://github.com/openzipkin/zipkin/tree/master/zipkin-storage/elasticsearch

timestamp + duration would be the time a span was recorded and eligible to be sent. In https://github.com/openzipkin/zipkin-reporter-java, and other libraries, there would be up to a second before the span goes on the wire.

We could look at our local endpoint vs one of the endpoints in the span and try to determine roughly the lag, based on limitations of clock skew guessing. If we did, we'd likely record that as a stat vs put it into the span... unless we created a "fake span" which we then recorded against zipkin-server.

Another option would be to accept traces originated by reporters. Ex start a trace when the span was sent. In kafka, we'd have to hijack the message key, which isn't a problem in zipkin topics. https://github.com/openzipkin/brave/pull/212

Thoughts?

codefromthecrypt:
Zipkin's timestamps are in epoch microseconds, which is not a supported date type in Elasticsearch.
In consideration of tools like like Kibana, this component adds "timestamp_millis" when writing
spans. This is mapped to the Elasticsearch date type, so can be used to any date-based queries.

If you are investigating a span, you'll also notice the "collector_timestamp_millis" field. This
is the epoch timestamp that the collector wrote before sending to storage. A gap between
"timestamp_millis" and "collector_timestamp_millis" is normal. For example, a span cannot be reported until it is complete (span.timestamp + span.duration). Next, reporters often buffer for a second or so before sending a message onto the transport. Also, there is transport delay prior to a collector receiving that message. Finally, there may be clock skew between the reporter and the collector.

Fixes #1318

### DIFF
--- a/zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/ElasticsearchSpanConsumerTest.java
+++ b/zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/ElasticsearchSpanConsumerTest.java
@@ -21,7 +21,6 @@ import java.util.concurrent.ExecutionException;
 import org.junit.Before;
 import org.junit.Test;
 import zipkin.Annotation;
-import zipkin.Codec;
 import zipkin.Span;
 import zipkin.storage.elasticsearch.http.HttpElasticsearchTestGraph;
 
@@ -119,22 +118,6 @@ public class ElasticsearchSpanConsumerTest {
 
     assertThat(indexFromToday.size())
         .isEqualTo(1);
-  }
-
-  @Test
-  public void prefixWithTimestampMillis() {
-    Span span = Span.builder().traceId(20L).id(20L).name("get")
-        .timestamp(TODAY * 1000).build();
-
-    byte[] result =
-        ElasticsearchSpanConsumer.prefixWithTimestampMillis(Codec.JSON.writeSpan(span), TODAY);
-
-    String json = new String(result);
-    assertThat(json)
-        .startsWith("{\"timestamp_millis\":" + Long.toString(TODAY) + ",\"traceId\":");
-
-    assertThat(Codec.JSON.readSpan(json.getBytes()))
-        .isEqualTo(span); // ignores timestamp_millis field
   }
 
   void accept(Span span) {
--- a/zipkin-storage/elasticsearch/README.md
+++ b/zipkin-storage/elasticsearch/README.md
@@ -14,6 +14,13 @@ Zipkin's timestamps are in epoch microseconds, which is not a supported date typ
 In consideration of tools like like Kibana, this component adds "timestamp_millis" when writing
 spans. This is mapped to the Elasticsearch date type, so can be used to any date-based queries.
 
+If you are investigating a span, you'll also notice the "collector_timestamp_millis" field. This
+is the epoch timestamp that the collector wrote before sending to storage. A gap between
+"timestamp_millis" and "collector_timestamp_millis" is normal. For example, a span cannot be reported
+until it is complete (span.timestamp + span.duration). Next, reporters often buffer for a second
+or so before sending a message onto the transport. Also, there is transport delay prior to a collector
+receiving that message. Finally, there may be clock skew between the reporter and the collector.
+
 `zipkin.storage.elasticsearch.ElasticsearchStorage.Builder` includes defaults
 that will operate against a local Elasticsearch installation.
 
--- a/zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/HttpBulkSpanIndexer.java
+++ b/zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/HttpBulkSpanIndexer.java
@@ -29,10 +29,6 @@ final class HttpBulkSpanIndexer extends HttpBulkIndexer<Span> implements
 
   @Override public void add(String index, Span span, Long timestampMillis) throws IOException {
     String id = null; // Allow ES to choose an ID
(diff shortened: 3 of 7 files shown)


zipkin-storage/elasticsearch/README.md
```diff
@@ -14,6 +14,13 @@ Zipkin's timestamps are in epoch microseconds, which is not a supported date typ
 In consideration of tools like like Kibana, this component adds "timestamp_millis" when writing
 spans. This is mapped to the Elasticsearch date type, so can be used to any date-based queries.
 
+If you are investigating a span, you'll also notice the "collector_timestamp_millis" field. This
+is the epoch timestamp that the collector wrote before sending to storage. A gap between
+"timestamp_millis" and "collector_timestamp_millis" is normal. For example, a span cannot be reported
+until it is complete (span.timestamp + span.duration). Next, reporters often buffer for a second
+or so before sending a message onto the transport. Also, there is transport delay prior to a collector
+receiving that message. Finally, there may be clock skew between the reporter and the collector.
```
shakuzen, line 22: Good, thorough explanation 💯

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/ElasticsearchSpanConsumerTest.java
```diff
@@ -121,22 +120,6 @@ public void searchByTimestampMillis() throws ExecutionException, InterruptedExce
         .isEqualTo(1);
   }
 
-  @Test
-  public void prefixWithTimestampMillis() {
```
codefromthecrypt, line 125: this was a copy/pasted test of a static method (so it is redundant)

codefromthecrypt:
fwiw I still don't like this feature even more so after adding it. fields aren't like code, they are very hard to remove. I only raised this bc a couple people are interested in it. So, I'd like both to say they really still want it before merging, and no-one say "please don't merge this"

cc @openzipkin/elasticsearch

dan-tr:
I would really like this feature. As comprise though, I wouldn't mind if this was a configurable options specifically for the Elasticsearch storage option.

codefromthecrypt:
> I would really like this feature. As comprise though, I wouldn't mind if
> this was a configurable options specifically for the Elasticsearch storage
> option.
> 
> a configuration option allows folks who don't want this to leave it off
> (likely default at this point)
> a configuration option makes the code more spaghetti and also has the same
> or higher support burden, as we'd not be able to remove it later.

things like this in general need multiple people to buy into it because of
permanence. Let me see if I can find some folks who might be sympathetic or
know of an alternative that doesn't require us to maintain this.

codefromthecrypt:
@lukeotterblad thoughts?

codefromthecrypt:
ex if elasticsearch could be configured to add a timestamp at write time, we'd not need to add a third timestamp in zipkin code.

codefromthecrypt:
ps still thinking about this and logistics. @dan-tr how do you assemble zipkin? do you use docker, or can you make your own image given artifacts? What if we made an extension jar which if you pop into a directory (or otherwise add to the classpath), it would enable this feature?

lukeotterblad:
You can add a value in your index mapping where elasticsearch will record the timestamp of the documents last ingestion time (even if you provide it nothing). This might be a better solution for you, though it appears they are also deprecating the feature. 

See the following for examples:
http://stackoverflow.com/questions/27255596/elasticsearch-index-last-update-time
https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-timestamp-field.html

codefromthecrypt:
@lukeotterblad yeah I think the deprecation part was the initial motivation for this feature. I wonder when it will be removed...

dan-tr:
@adriancole Yes, we are using Docker. I like the idea of implementing a plugin for this. Maybe it could be even more generic where by I could build a jar (or something similar) that could intercept the data as it written to ES and that would allow me to add the collection timestamp or do anything else at the time of ingestion.

Just a thought.

dan-tr:
I was just made aware of the new feature in ES 5:
https://www.elastic.co/blog/new-way-to-ingest-part-1

With the new processors we may be able to have ES add an ingest date for us.

codefromthecrypt:
> I was just made aware of the new feature in ES 5:
> https://www.elastic.co/blog/new-way-to-ingest-part-1
> 
> With the new processors we may be able to have ES add an ingest date for
> us.
> 
> would you be ok running ES 5, then? We've an open issue to support it
> anyway (doesn't work now). We can test through that.

dan-tr:
We're not running ES 5 yet, but hope to be soon (Next 6 months :smile: ), but I am willing to table this issue until after we get a chance to see if ES 5 will solve the problem.

codefromthecrypt:
deal!

codefromthecrypt:
verified that the below works with ES 5, so we can close this in favor

```
PUT _ingest/pipeline/my-pipeline-id
{
  "description" : "add collector_timestamp_millis",
  "processors" : [
    {
      "set" : {
        "field": "collector_timestamp_millis",
        "value": "{{_ingest.timestamp}}"
      }
    }
  ]
}
```

PR #1331, opened 2016-10-06
Fixes missing CORS config

codefromthecrypt:
We forgot one of the endpoints when adding CORS config. With this change
zipkin-js can send spans directly to Zipkin via HTTP.

See #1234


codefromthecrypt:
cc @shakuzen @hyleung

With this in, the http transport doesn't die on CORS error https://github.com/openzipkin/zipkin-js/tree/master/packages/zipkin-transport-http

PR #1340, opened 2016-10-13
Simplifies AWSSignatureVersion4 by using Okio 1.11

codefromthecrypt:


codefromthecrypt:
cc @llinder fyi

llinder (commented):
Nice! I should read up on how signing works with STS and cross account authorization. Not having to include the AWS sdk would be nice some day.

PR #1341, opened 2016-10-13
Unwinds elasticsearch-http configuration by using List<Interceptor>

codefromthecrypt:
Before, we had to pass around more state in order to set OkHttp
interceptors used in elasticsearch (ex AWS authentication). This
injects `List<Interceptor>` so that the process is a bit easier.

### DIFF
--- a/zipkin-autoconfigure/storage-elasticsearch-aws/src/test/java/zipkin/autoconfigure/storage/elasticsearch/aws/ZipkinElasticsearchAwsStorageAutoConfigurationTest.java
+++ b/zipkin-autoconfigure/storage-elasticsearch-aws/src/test/java/zipkin/autoconfigure/storage/elasticsearch/aws/ZipkinElasticsearchAwsStorageAutoConfigurationTest.java
@@ -13,7 +13,7 @@
  */
 package zipkin.autoconfigure.storage.elasticsearch.aws;
 
-import okhttp3.OkHttpClient;
+import okhttp3.Interceptor;
 import org.junit.After;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
cc folks who might be curious @llinder @sethp-jive and @shakuzen who mentioned to me the list approach

codefromthecrypt:
whoohoo no test flake

zipkin-autoconfigure/storage-elasticsearch-aws/src/test/java/zipkin/autoconfigure/storage/elasticsearch/aws/ZipkinElasticsearchAwsStorageAutoConfigurationTest.java
```diff
@@ -62,9 +62,8 @@ public void providesAWSSignatureVersion4_whenStorageTypeElasticsearchAndHostsAre
         ZipkinElasticsearchAwsStorageAutoConfiguration.class);
     context.refresh();
 
-    assertThat(context.getBean(OkHttpClient.class).networkInterceptors())
-        .extracting(i -> i.getClass())
-        .contains((Class) AWSSignatureVersion4.class);
+    assertThat(context.getBean(Interceptor.class))
```
shakuzen, line 65: Can you get a `List<Interceptor>` from the context? Otherwise this looks like an assertion that _may_ break in the future when there is more than one `Interceptor` bean, which I thought was the point of switching to the `List`.

codefromthecrypt: I actually started googling how to get a generic type from a bean factory.. probably best is to make a holder class that has an auto-wired list?

codefromthecrypt: https://github.com/openzipkin/zipkin/pull/1342 thanks for noticing!

PR #1344, opened 2016-10-17
Reference current class for UI properties

shakuzen:
The UI properties are mapped to ZipkinUiProperties now, and this will help people find where they are used when searching the code.


codefromthecrypt:
thanks!

PR #1351, opened 2016-10-25
Adds an example to Endpoint.port note

### LINKED ISSUE #1350: Annotations with endpoint port larger than 32767 fails
Endpoint ports are cast to a short leading to query errors for ports larger than 32767.

This is a bug I guess? Or am I missing something?

```
result.port((short) reader.nextInt());
```

https://github.com/openzipkin/zipkin/blob/master/zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ZipkinAdapters.java#L154

codefromthecrypt:
This adds a code example to Endpoint.port javadoc to help explain what
the watch for negatives comment was about.

Fixes #1350


codefromthecrypt:
added an example to the javadoc that mentioned this https://github.com/openzipkin/zipkin/pull/1351

Issue #1298, opened 2016-09-15
Transition to 128bit trace ids (still 64bit span ids)


codefromthecrypt:
This tracks work needed to transition to 128bit trace ids (still 64bit span ids). The overall plan was decided in #1262
### Wave 1: tolerate 128-bit trace ids

First step is to make it possible to receive 128-bit ids, even if the high bits are thrown out.
This decouples the rest of the project.
- [x] [Tolerate 128bit trace ids when reading json and api requests](https://github.com/openzipkin/zipkin/pull/1296) - Fixed in Zipkin 1.11.1
- [x] [Tolerate 128bit trace ids when reading X-B3-TraceId](https://github.com/openzipkin/b3-propagation/issues/6)
### Wave 2: store 128-bit trace ids, but retrieve by lower 64-bits

Next step is to support storage of 128-bit ids. Once this is in, instrumentation can report them to zipkin. Understanding deployments will be mixed and not support 128-bit for a while, query remains 64-bit.
- [x] Add trace_id_high (default 0) to the thrift definition
- [x] Add traceIdHigh (default 0) to zipkin.Span and implement it in codecs
  - in json, it is simply a longer traceId field to accommodate the extra 16 hex characters
- [x] Ensure supported storage components are not lossy with regards to 128-bit ids
- [x] Ensure supported storage can query by 64-bit trace ids
### Wave 3: Propagate and report 128-bit ids

Now that ids are storable, we can propagate and report them safely.
- [x] publish B3 library with 128-bit support (including binary representation)
- [x] Update instrumentation libraries to optionally start 128-bit traces See https://github.com/openzipkin/b3-propagation/issues/6 for status

### Wave 4: Support query by 128-bit id

While conflict is unlikely when querying trace ids by the lower 64-bits, we should be able to support queries and uniqueness by 128-bit.
- [x] add SpanStore.getTrace and getRawTrace with a 128-bit id
- [x] add tests that show traces that share the same lower 64 bits are not returned together
- [x] change zipkin-UI/api to accept the longer 128-bit (32char) hex id

codefromthecrypt:
Most libraries now accommodate 128bit trace ids. Next step is to make it possible to store them

codefromthecrypt:
[...]

PR #1353, opened 2016-10-25
128-bit trace ID part 2: store 128-bit trace ids, but retrieve by lower 64-bits

codefromthecrypt:
This supports storage of 128-bit ids. Once this in, instrumentation can report 128-bit trace ids to zipkin. Understanding deployments will be mixed and not support 128-bit for a while, query remains 64-bit.
- Adds traceIdHigh (default 0) to zipkin.Span and implement it in codecs
  - in json, it is simply a longer traceId field to accommodate the extra 16 hex characters
- Ensures supported storage components are not lossy with regards to 128-bit ids
- Ensures supported storage can query by 64-bit trace ids
#### Notes on storage of 128-bit trace id:
- mysql needs the following schema update:

``` mysql
mysql> alter table zipkin_spans add `trace_id_high` BIGINT NOT NULL DEFAULT 0;
```

Notes on retrieval by lower-64 bit:
- cassandra3 doesn't support retrieval by lower-64 bit
  - This is likely solvable but too much for this PR.
- elasticsearch requires an index template update
  - This automatically applies when the day rolls over.

See #1298

### DIFF
--- a/zipkin-storage/cassandra3/src/test/java/zipkin/storage/cassandra3/CassandraSpanStoreTest.java
+++ b/zipkin-storage/cassandra3/src/test/java/zipkin/storage/cassandra3/CassandraSpanStoreTest.java
@@ -15,6 +15,7 @@ package zipkin.storage.cassandra3;
 
 import java.util.ArrayList;
 import java.util.List;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
@@ -95,6 +96,17 @@ public class CassandraSpanStoreTest extends SpanStoreTest {
     assertThat(
         store().getTraces(QueryRequest.builder().lookback(86400000L).limit(traceCount).build()))
         .hasSize(traceCount);
+  }
+
+  /**
+   * Currently, trace ids are stored as a variable-length number (as opposed to a 2 part hi/lo
+   * 128bit id). In order to support query by low 64 bits, we need to split this up somehow or add a
+   * separate index.
+   */
+  @Override
+  @Test
+  @Ignore("fetch by lower 64-bit isn't supported in cassandra3")
+  public void getTrace_retrieves128bitTraceIdByLower64Bits() {
   }
 
   long rowCount(String table) {
--- a/zipkin-storage/elasticsearch/src/main/resources/zipkin/storage/elasticsearch/zipkin_template.json
+++ b/zipkin-storage/elasticsearch/src/main/resources/zipkin/storage/elasticsearch/zipkin_template.json
@@ -12,6 +12,18 @@
           "filter": [
             "lowercase"
           ]
+        },
+        "traceId_analyzer": {
+          "type": "custom",
+          "tokenizer": "keyword",
+          "filter": "traceId_filter"
+        }
+      },
+      "filter": {
+        "traceId_filter": {
+          "type": "pattern_capture",
+          "patterns": ["([0-9a-f]{1,16})$"],
+          "preserve_original": true
         }
       }
     }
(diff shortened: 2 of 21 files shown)


codefromthecrypt:
cc @openzipkin/core @openzipkin/elasticsearch @jcarres-mdsol

zipkin-storage/cassandra3/src/test/java/zipkin/storage/cassandra3/CassandraSpanStoreTest.java
```diff
@@ -97,6 +98,17 @@ public void overFetchesToCompensateForDuplicateIndexData() {
...
 
+  /**
+   * Currently, trace ids are stored as a variable-length number (as opposed to a 2 part hi/lo
+   * 128bit id). In order to support query by low 64 bits, we need to split this up somehow or add a
+   * separate index.
+   */
+  @Override
+  @Test
+  @Ignore("fetch by lower 64-bit isn't supported in cassandra3")
+  public void getTrace_retrieves128bitTraceIdByLower64Bits() {
```
codefromthecrypt, line 109: @michaelsembwever will need your help post-merge on this I think. I started playing around with a TraceIdUDT (with hi/lo fields).. that worked, but then I didn't get how to query by half of the traceId UDT considering it is a part of the partition key. Maybe it is a SASI concern..

zipkin-storage/elasticsearch/src/main/resources/zipkin/storage/elasticsearch/zipkin_template.json
```diff
@@ -12,6 +12,16 @@
           "filter": [
             "lowercase"
           ]
+        },
+        "traceId_analyzer": {
```
codefromthecrypt, line 16: @openzipkin/elasticsearch My assumption is that this is not a costly step, and it also doesn't seem to increase the size of index needed (since this doesn't create multiple tokens). If these assumptions are invalid, report back!

zipkin-storage/mysql/src/main/java/zipkin/storage/mysql/HasTraceIdHigh.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.sql.DataSource;
+import org.jooq.DSLContext;
+import org.jooq.exception.DataAccessException;
+import zipkin.internal.Lazy;
+
+import static zipkin.storage.mysql.internal.generated.tables.ZipkinSpans.ZIPKIN_SPANS;
+
+final class HasTraceIdHigh extends Lazy<Boolean> {
```
codefromthecrypt, line 27: As much as I don't love creating things like this, it allows us to work even when the field isn't present, yet. This is how we addressed other minor schema changes such as pre-aggregated deps and ipv6.

codefromthecrypt:
revised elasticsearch so that it tokenizes both 64 and 128bit lengths. This will allow future apis to support exact retrieval on trace id without preventing retrieval on 64-bit (interop mode).

zipkin-storage/cassandra3/src/test/java/zipkin/storage/cassandra3/CassandraSpanStoreTest.java
michaelsembwever: thanks for opening https://github.com/openzipkin/zipkin/issues/1364
will let the discussion flow from there…

zipkin-storage/elasticsearch/src/main/resources/zipkin/storage/elasticsearch/zipkin_template.json
codefromthecrypt: this might need to be reconsidered if we want to support elasticsearch 5 https://github.com/openzipkin/zipkin/issues/1379

PR #1359, opened 2016-10-26
Writes Span.traceIdHigh last so that it is exactly the same as libthrift

codefromthecrypt:
Consistency on field order isn't required by thrift, but this writes the
fields in ID order to make exact binary comparison pass vs libthrift.


codefromthecrypt:
cc @basvanbeek this uses code generated from the latest merge in zipkin-api

PR #1367, opened 2016-10-28
Fix annotation query parsing

### LINKED ISSUE #1366: Search by complex values
@mansu raised a request on gitter to search by arbitrarily complex values (i.e. ones that have equals tokens in them)

> @mansu 05:05
> I think the problem is more subtler here: for example I can search for X-Real-Ip=66.87.120.xx, but X-Pinterest-Referrer=https://mail.google.com/mail/mu/mp/608/?source=nap&hr=1&hl=en-US doesn't work. Can it be that special characters are causing issues?

Realistically speaking, such a search isn't a great example as full urls aren't likely to work well with exact matching. that's why we call out which keys are good for filtering, vs others which are just context https://github.com/openzipkin/zipkin-api/blob/master/thrift/zipkinCore.thrift#L128

I suspect that equals characters interfere with the parser, as equals is a token in the encoding of a search expression

``` yaml
        - name: annotationQuery
          in: query
          type: string
          required: false
          description: |
             Ex. `http.uri=/foo and retried` - If key/value (has an `=`),
             constrains against Span.binaryAnnotations of time string. If just
             a word, constrains against Span.annotations. Any values are AND
             against eachother. This means a span in the trace must match all of
             these
```

Suman mentioned that the unpacking of the search expression could be at fault, for example it could look for the first = vs any of them. This might work, but needs to be tested well.

mansu:
The current annotation query parser only returns value if a value contains an '=' sign. Fixed the parser to return the complete value even if they are '=' sign in the value.

Fixes #1366

### DIFF
--- a/zipkin/src/main/java/zipkin/storage/QueryRequest.java
+++ b/zipkin/src/main/java/zipkin/storage/QueryRequest.java
@@ -225,7 +225,7 @@ public final class QueryRequest {
             addAnnotation(ann);
           } else {
             String[] keyValue = ann.split("=");
-            addBinaryAnnotation(ann.substring(0, idx - 1), keyValue.length < 2 ? "" : ann.substring(idx+1));
+            addBinaryAnnotation(ann.substring(0, idx), keyValue.length < 2 ? "" : ann.substring(idx+1));
           }
         }
       }
--- a/zipkin/src/main/java/zipkin/storage/QueryRequest.java
+++ b/zipkin/src/main/java/zipkin/storage/QueryRequest.java
@@ -220,11 +220,12 @@ public final class QueryRequest {
     public Builder parseAnnotationQuery(String annotationQuery) {
       if (annotationQuery != null && !annotationQuery.isEmpty()) {
         for (String ann : annotationQuery.split(" and ")) {
-          if (ann.indexOf('=') == -1) {
+          int idx = ann.indexOf('=');
+          if (idx == -1) {
             addAnnotation(ann);
           } else {
             String[] keyValue = ann.split("=");
-            addBinaryAnnotation(keyValue[0], keyValue.length < 2 ? "" :keyValue[1]);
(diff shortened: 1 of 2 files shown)


zipkin/src/test/java/zipkin/storage/QueryRequestTest.java
```diff
@@ -105,7 +105,7 @@ public void annotationQuery_roundTrip() {
 
     QueryRequest request =
         QueryRequest.builder().serviceName("security-service").parseAnnotationQuery(annotationQuery).build();
-
```
codefromthecrypt, line 108: nit: fuzz

zipkin/src/main/java/zipkin/storage/QueryRequest.java
```diff
@@ -220,11 +220,12 @@ public Builder spanName(@Nullable String spanName) {
...
       if (annotationQuery != null && !annotationQuery.isEmpty()) {
         for (String ann : annotationQuery.split(" and ")) {
-          if (ann.indexOf('=') == -1) {
+          int idx = ann.indexOf('=');
+          if (idx == -1) {
             addAnnotation(ann);
           } else {
             String[] keyValue = ann.split("=");
-            addBinaryAnnotation(keyValue[0], keyValue.length < 2 ? "" :keyValue[1]);
+            addBinaryAnnotation(ann.substring(0, idx), keyValue.length < 2 ? "" : ann.substring(idx+1));
```
codefromthecrypt, line 228: you could probably remove keyValue and just compare idx vs ann.length (ex I think the goal here is to ensure it works with a value like `foo=`

codefromthecrypt: formatting nit s/idx+1/idx + 1/

codefromthecrypt (commented):
Thanks! slight cleanup then good to go!

codefromthecrypt:
releasing now in 1.14.2

PR #1373, opened 2016-10-31
Serializes I64 numbers greater than 53bits as strings in json

### LINKED ISSUE #1372: 64 bit annotations only shown with 53 bit precision in UI
We have a few binary annotations in our spans (emitted by Finagle using `Trace.binaryAnnotation`)

```
      {
        "key": "XXXXXXId",
        "value": 793024884742246336,
        "type": "I64",
        "endpoint": {
          "serviceName": "XXXXXXX",
          "ipv4": "10.19.XXX.XXX",
          "port": 10579
        }
      },
```

In the UI, however, the value shows as `793024884742246400`, due to the 53 bit javascript limitation. Perhaps `I64` should be sent as strings from the back-end?

codefromthecrypt:
While usage of binary annotation type I64 is discouraged, Finagle lets you do it

Fixes #1372


codefromthecrypt:
cc @nollbit @eirslett

PR #1377, opened 2016-11-01
Compresses indexes to search based on date expression

### LINKED ISSUE #1327: Dynamically bound lookback for elasticsearch queries
Currently, we have a heck of a long disclaimer in our code to defend against too many indexes being queried. When the lookback hits over 100 index entries (days), we reroute the search to a post request.

``` java
  /**
   * Because the {@link HttpClient} is forced to send the index names in the URL, too many indices
   * (or long names) can quickly exceed the 4096-byte limit commonly imposed on a single HTTP line.
   * This limit is not entirely within the control of Elasticsearch, either, as it may be imposed by
   * arbitrary HTTP-speaking middleware along the request's route. Some more details can be found on
   * this "wontfix" issue: https://github.com/elastic/elasticsearch/issues/7298
   */
  private static final int MAX_INDICES = 100;
```

Apart from dependency links, Zipkin data is usually only retained for orders of days, certainly not hundred days. The reason why this hack was needed is that our unit tests tests by default lookback to epoch. The server looks back to `zipkin.query.lookback` which is 7 days by default, but due to modularity, this property isn't readable by tests.

Except for the test cases, the only way we'd ever need 100 days of data returned is if someone adjusted `zipkin.query.lookback` to an order of magnitude higher than its default (or if someone was using the library directly).

One way or another, we should put a hard-limit on how far back we should look as certainly getting data from 1970 will return null. The storage schema wasn't even written until mid-march of 2016.

My suggestion is the following:
- Hard limit the begin timestamp to no later than when zipkin's elasticsearch was written, March 2016.
- Lookup the indexes dynamically and hard-limit queries to not attempt to read before them

Anyone have feelings about this? @openzipkin/elasticsearch 

```
$ curl -s 'localhost:9200/test_zipkin_http-*/_settings?pretty'
{
  "test_zipkin_http-2016-10-04" : {
    "settings" : {
      "index" : {
        "number_of_shards" : "5",
        "creation_date" : "1475553054979",
        "requests" : {
          "cache" : {
            "enable" : "true"
          }
        },
        "analysis" : {
          "analyzer" : {
            "lowercase" : {
              "filter" : [ "lowercase" ],
              "type" : "custom",
              "tokenizer" : "keyword"
            }
          }
        },
        "number_of_replicas" : "1",
        "uuid" : "_RCZiBM2TUiLbyLge5Yl_Q",
        "version" : {
          "created" : "2040099"
        }
      }
    }
  }
}
```

codefromthecrypt:
This adds `IndexNameFormatter.indexNamePatternsForRange` which 
compresses api requests by using wildcard expressions for months or 
years.

For example, if begin is 2016-11-30 and end is 2017-01-02, the result 
will be 2016-11-30, 2016-12-*, 2017-01-01 and 2017-01-02 as opposed to
a set of >30 individual index names.

As a side effect, we no longer need special http configuration as we
won't create request lines larger than 4096 on account of index names.

Fixes #1327
Obviates #1371

### DIFF
--- a/zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/HttpClient.java
+++ b/zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/HttpClient.java
@@ -55,15 +55,6 @@ import static zipkin.moshi.JsonReaders.enterPath;
 final class HttpClient extends InternalElasticsearchClient {
   static final MediaType APPLICATION_JSON = MediaType.parse("application/json");
 
-  /**
-   * Because the Http api requires sending index names in the URL, too many indices
-   * (or long names) can quickly exceed the 4096-byte limit commonly imposed on a single HTTP line.
-   * This limit is not entirely within the control of Elasticsearch, either, as it may be imposed by
-   * arbitrary HTTP-speaking middleware along the request's route. Some more details can be found on
-   * this "wontfix" issue: https://github.com/elastic/elasticsearch/issues/7298
-   */
-  private static final int MAX_INDICES = 100;
-
   static final class Factory implements InternalElasticsearchClient.Factory {
     final Lazy<List<String>> hosts;
     final OkHttpClient client;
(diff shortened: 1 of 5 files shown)


codefromthecrypt:
cc @openzipkin/elasticsearch

codefromthecrypt:
Here's an example of the compression in action

screen shot 2016-11-01 at 3 46 10 pm

PR #1381, opened 2016-11-02
Makes charset explicit in MySQL tables

codefromthecrypt:
Travis just started failing due to unicode characters not returning.
This makes character set explicit in our defaults in case others have
the same problem.


codefromthecrypt:
note: not sure exactly what caused this in travis, but it appears to have recently picked up mysql 5.6.33

codefromthecrypt:
merging as needed to unbreak travis

PR #1385, opened 2016-11-03
Supports querying traces with mixed-length trace IDs

codefromthecrypt:
Unless you only issue 128-bit traces when all applications support them,
the process of updating applications from 64 to 128-bit trace IDs results
in a mixed state. This mixed state is mitigated by the setting
`STRICT_TRACE_ID=false`, explained below. Once a migration is complete,
remove the setting `STRICT_TRACE_ID=false` or set it to true.

Supporting changes are:

`SpanStore.getTrace(long traceIdHigh, long traceIdLow)`
`StorageComponent.Builder.strictTraceId(boolean strictTraceId)`


codefromthecrypt:
There's an unlikely scenario where the conditions searched returned results with traces that include exactly the same low bits, but have different and non-zero high bits.

It is possible that one of the traces didn't actually match the predicate. Accordingly, we have to filter them client-side to ensure we don't return false-positives.

This particular scenario affects both cassandra and elasticsearch, so extracting common code to deal with it. This code should be up tomorrow.

codefromthecrypt:
Still quite a lot of work left, particularly around testing this and making it configurable. Probably won't get a revision up until Monday as I'm flying most of tomorrow.

codefromthecrypt:
At the moment, this is what I'm going with for the storage component parameter. This would be mapped to an environment variable in the server/docker...

```

  interface Builder {

    /**
     * Zipkin historically had 64-bit {@link Span#traceId trace IDs}, but it now supports 128-bit
     * trace IDs via {@link Span#traceIdHigh}, or its 32-character hex representation.
     *
     * <p>This setting allows you to look up traces who have 128-bit trace ids by the lower-64 bits
     * using {@link SpanStore#getTrace(long)}. In most implementations, this implies a second index
     * which can be more expensive in terms of storage or memory than only supporting 128-bit
     * lookups via {@link SpanStore#getTrace(long, long)}
     *
     * <p>This should be enabled until all instrumentation report 128-bit trace IDs consistently,
     * and {@link SpanStore#getTrace(long)} is no longer in use.
     */
    Builder mixedTraceIdLength(boolean mixedTraceIdLength);

    StorageComponent build();
  }
```

Particularly, this should allow sites who have transitioned to 128-bit ids to opt-out of tokenizing the traceId in elasticsearch or possibly another column in cassandra3.

codefromthecrypt:
pushed a code update which needs cleanup (particularly in MySQL which is the nastiest).

codefromthecrypt:
changed flag to strictTraceId

codefromthecrypt:
wow.. actually green, huh

codefromthecrypt:
pretty happy with docs and config. test backfilling and some code polishing needed, readying for merge

codefromthecrypt:
added config integration tests.

NEXT STEP:
- see if I can make ES template opt-out of tokenization.
- mysql cleanup
- more test backfills

ETA sometime tomorrow

codefromthecrypt:
got ES to conditionally tokenize and cleaned up some of the code. want to do one last cleanup and double-check the test cases are named and documented well. Will merge tomorrow

codefromthecrypt:
ok only thing I'm doing between now and merge is backfilling tests for GroupByTraceId... so expect this in by tomorrow. I'll cut a version following

PR #1389, opened 2016-11-04
Blocking import for gson as we embed it

cschneider:


zipkin/pom.xml
```diff
@@ -63,6 +63,7 @@
 
   <build>
     <plugins>
+      <!-- Need to block import of shaded packages as maven bundle plugin analyzes the unshaded jar -->
```
codefromthecrypt: maybe remind people the file that blocked imports go is `bnd.bnd`? I think many don't know this

cschneider: Makes sense .. I also found another issue

cschneider:
I am now at the stage where I can install zipkin and zipkin reporter bundles. I will now try to first get everything installed and started in OSGi and then write a first actual example that uses it.

codefromthecrypt:
cool beans.. all of the snapshot repos should be working, so please ping back how it goes.

codefromthecrypt:
fyi this (and other related change) is going out as 1.14.4

Issue #1387, opened 2016-11-04
Zipkin server does not work with MySQL 5.7


ajacques:
### Technical Details
MySQL Server Version ```5.7.14-8 - Percona Server (GPL), Release '8', Revision '1f84ccd'```
Zipkin Server: ```1.14.3``` 

MySQL SQL Mode: ```ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION``` (Default for Percona distributed by Docker containers)

### Summary
Running Zipkin with ```STORAGE_TYPE=mysql``` causes certain requests to fail with exceptions due to semantically invalid SQL queries. This is caused by SQL_MODE=ONLY_FULL_GROUP_BY.

As per the [MySQL 5.7 release notes](http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_only_full_group_by), "[...] the default SQL mode includes ONLY_FULL_GROUP_BY. [...]"

#### Exhibit A
```
11/3/2016 10:00:56 PMorg.mariadb.jdbc.internal.util.dao.QueryException: Expression #1 of ORDER BY clause is not in SELECT list, references column 'Zipkin.zipkin_spans.start_ts' which is not in SELECT list; this is incompatible with DISTINCT
11/3/2016 10:00:56 PMQuery is : select distinct `zipkin_spans`.`trace_id` from `zipkin_spans` join `zipkin_annotations` on (`zipkin_spans`.`trace_id` = `zipkin_annotations`.`trace_id` and `zipkin_spans`.`id` = `zipkin_annotations`.`span_id`) where (`zipkin_spans`.`start_ts` between 1478232055406000 and 1478235655406000 and `zipkin_annotations`.`endpoint_service_name` = 'front-end-ui') order by `zipkin_spans`.`start_ts` desc limit 10
11/3/2016 10:00:56 PM	at org.mariadb.jdbc.internal.protocol.AbstractQueryProtocol.getResult(AbstractQueryProtocol.java:939) ~[mariadb-java-client-1.4.6.jar!/:na]
[...]
11/3/2016 10:00:56 PM	at org.jooq.impl.AbstractResultQuery.fetch(AbstractResultQuery.java:312) ~[jooq-3.8.4.jar!/:na]
[...]

PR #1390, opened 2016-11-05
Zipkin #1387 - Support for MySQL 5.7 with SQL_MODE=ONLY_FULL_GROUP_BY

ajacques:
### Summary
As documented in #1387, this pull request will correct the ambiguous SQL query for MySQL backends. 

### Change
This changes the SQL query used to get a list of the traces to one that is now MySQL 5.7 compatible.
Previously the query was:
```
select distinct `zipkin_spans`.`trace_id` from `zipkin_spans` [...] order by `zipkin_spans`.`start_ts`
```
This query is ambiguous because it does not specify how to order along the start_ts when there's multiple records with the same trace_id.

It's now:
```
select `zipkin_spans`.`trace_id`, MAX(`zipkin_spans`.`start_ts`) [...] group by `zipkin_spans`.`trace_id` order by MAX(`zipkin_spans`.`start_ts`) desc limit 10
```

This forces MySQL to order along the last start_ts.

### Testing Done
Verified unit tests still succeed [Travis CI](https://travis-ci.org/ajacques/zipkin/builds/173357222)
I've also uploaded this change to my own testing server and verified that I can now view spans. Previously the index page would fail to load with a 500 with the previously mentioned failure.

### Notes
We may also want to remove the comment in zipkin\zipkin-storage\mysql\README.md that references the SQL_MODE required change, but I haven't done enough testing to verify that it's completely unnecessary everywhere. In my use case, it's not needed anymore in this repository. The aforementioned issue does note there might be an issue in the zipkin-dependencies repository.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -56,7 +56,6 @@ dependencies:
 database:
   override:
     - mysql -u ubuntu -e 'SET GLOBAL innodb_file_format=Barracuda'
-    - mysql -u ubuntu -e "SET GLOBAL sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''))"
     - mysql -u ubuntu circle_test < zipkin-storage/mysql/src/main/resources/mysql.sql
 
 test:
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
Great stuff. do you mind removing the workaround from the README and circle.yml? Once the latter is done, you can assume the README is no longer needed (as circleci uses mysql 5.7 and travis 5.6)

cc @iocanel

codefromthecrypt:
Nice work!

codefromthecrypt:
fyi this (and other related change) is going out as 1.14.4

PR #1391, opened 2016-11-05
Adds missing column: zipkin_annotations.trace_id_high

codefromthecrypt:
The previous schema change left out `zipkin_annotations.trace_id_high`
which is needed for future work. This adds it and defaults to 0.

Add the `trace_id_high` field
```mysql
alter table zipkin_annotations add `trace_id_high` BIGINT NOT NULL DEFAULT 0 COMMENT 'If non zero, this means the trace uses 128 bit traceIds instead of 64 bit';
```

Update unique constraints to consider `trace_id_high`
```
ALTER TABLE zipkin_spans
   DROP INDEX trace_id,
   ADD UNIQUE KEY(`trace_id_high`, `trace_id`, `id`) COMMENT 'ignore insert on duplicate';

ALTER TABLE zipkin_annotations
   DROP INDEX trace_id,
   ADD UNIQUE KEY(`trace_id_high`, `trace_id`, `span_id`, `a_key`, `a_timestamp`) COMMENT 'Ignore insert on duplicate';
```

See #1353


codefromthecrypt:
@ajacques @jcarres-mdsol not sure if you've updated to zipkin_spans.trace_id_high, yet, but if not you'll also need to add zipkin_annotations.trace_id_high

PR #1374, opened 2016-10-31
Use span name table instead of materialized view

llinder:
- This switches span name and service name lookups to use a table instead of a materialized view.  This change will fix #1360
- Change limit logic to limit on a set of trace IDs instead of limiting on the provided collection first. In pracitce this didn't make a noticiable difference in the results but it seems like the intended logic.

PR #1392, opened 2016-11-06
Use span name table instead of materialized view

### LINKED ISSUE #1360: Empty span name list in UI with Cassandra3 storage component
While testing the Cassandra3 storage component I noticed that the span names dropdown in the UI went from showing 10+ entries to 0.   Running the same CQL statement through cqlsh shows that there are over 1000 records with blank span names.  This appears to be causing the UI to no longer show any span names after a period of time.

/cc @michaelsembwever

michaelsembwever:
- This switches span name and service name lookups to use a table instead of a materialized view.  This change will fix #1360
- Change limit logic to limit on a set of trace IDs instead of limiting on the provided collection first. In pracitce this didn't make a noticiable difference in the results but it seems like the intended logic.
- update stress profiles for new span_name_by_service tables

references:
 - https://github.com/openzipkin/zipkin/pull/1392
 - https://github.com/openzipkin/zipkin/issues/1360
 - https://github.com/openzipkin/zipkin/pull/1374

### DIFF
--- a/zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/Cassandra3Storage.java
+++ b/zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/Cassandra3Storage.java
@@ -216,6 +216,7 @@ public final class Cassandra3Storage
     for (String cf : ImmutableList.of(
         Schema.TABLE_TRACES,
         Schema.TABLE_TRACE_BY_SERVICE_SPAN,
+        Schema.TABLE_SERVICE_SPANS,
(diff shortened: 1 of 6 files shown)


michaelsembwever:
@llinder the commit is still authored by you… so please check it and say if you're ok with it.

codefromthecrypt:
tested locally.. all good!

codefromthecrypt:
thanks for resurrecting this, @michaelsembwever

llinder:
@michaelsembwever thanks for jumping on this so quickly!

Issue #1364, opened 2016-10-27
Support lookup of 128-bit traces by lower 64-bits in cassandra3


codefromthecrypt:
Currently, trace ids are stored as a variable-length number (as opposed to a 2 part hi/lo 128bit id). In order to support query by low 64 bits, we need to split this up somehow or add a separate index.

You'll notice `SpanStoreTest.getTrace_retrieves128bitTraceIdByLower64Bits` is disabled due to this.

cc @openzipkin/cassandra

codefromthecrypt:
I started playing around with replacing varint (BigInteger) with a new TraceIdUDT (with hi/lo fields). 

That worked, except I didn't get how to query by half of the traceId UDT considering traceId is a part of the partition key. Maybe this could be solved with SASI w/this or a different supporting 64-bit traceIdLo lookup field.

michaelsembwever:
Possible approaches are 
1) SASI based

```
CREATE TABLE IF NOT EXISTS zipkin3.traces (
    trace_id            varint,
    trace_id_low        bigint    STATIC,
    ts_uuid             timeuuid,
    id                  bigint,
    ts                  bigint,
    span_name           text,
    parent_id           bigint,
    duration            bigint,
    annotations         list<frozen<annotation>>,
    binary_annotations  list<frozen<binary_annotation>>,
    all_annotations     text, //-- can't do SASI on set<text>: comma-joined until CASSANDRA-11182
    PRIMARY KEY (trace_id, ts_uuid, id)
)
    WITH CLUSTERING ORDER BY (ts_uuid DESC)
    AND compaction = {'class': 'org.apache.cassandra.db.compaction.TimeWindowCompactionStrategy'}
    AND default_time_to_live =  604800;

CREATE CUSTOM INDEX ON zipkin3.traces (trace_id_low) USING 'org.apache.cassandra.index.sasi.SASIIndex'
   WITH OPTIONS = {'mode': 'SPARSE'};
```

or
2) clustering key based

```
CREATE TABLE IF NOT EXISTS zipkin3.traces (
    trace_id_low        bigint,
    trace_id            varint,
    ts_uuid             timeuuid,
    id                  bigint,
    ts                  bigint,
    span_name           text,
    parent_id           bigint,
    duration            bigint,
    annotations         list<frozen<annotation>>,
    binary_annotations  list<frozen<binary_annotation>>,
    all_annotations     text, //-- can't do SASI on set<text>: comma-joined until CASSANDRA-11182
[...]

PR #1394, opened 2016-11-06
Moves cassandra3 traceId from BigInteger to TraceIdUDT(high,low)

codefromthecrypt:
This moves cassandra3's schema to high/low which is more precise and
allows us more options on how to index by only the low 64-bits of the
trace ID.

See #1364


codefromthecrypt:
cc @michaelsembwever so I removed the indexing change from this commit just to separate the changes. I've not thought through the double-store thing, but I do believe that regardless of how we handle get by lower 64, this PR part is good change.

codefromthecrypt:
@openzipkin/cassandra fyi, this primarily changes the experimental cassandra3 impl to use fixed-length trace id in storage, particularly with the same high/low approach used in thrift and mysql.

codefromthecrypt:
tested locally

codefromthecrypt:
Let's layer any further work on #1364 over this change

PR #1395, opened 2016-11-07
Fixes tests not run due to bad naming convention

codefromthecrypt:


shakuzen (approved):
Alternatively, we could configure the build to pick up classes ending in `Test` and `Tests`. Like this: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-starters/spring-boot-starter-parent/pom.xml#L88-L100

codefromthecrypt:
@shakuzen good point, though I think normally we don't create tests ending in Tests.. I think there was some C&P to blame here :P

PR #1396, opened 2016-11-07
Add support for specifying a trusted cert via env

jroes:
This adds support for specifying a trusted cert via an environment variable.

This may not be quite up to standard, but wanted to start a conversation and confirm that there was interest in merging this support into the mainline before pursuing hygienic improvements.

### DIFF
--- a/zipkin-storage/cassandra3/pom.xml
+++ b/zipkin-storage/cassandra3/pom.xml
@@ -52,6 +52,12 @@
     </dependency>
 
     <dependency>
+      <groupId>com.github.jkutner</groupId>
+      <artifactId>env-keystore</artifactId>
+      <version>0.1.0</version>
+    </dependency>
+
+    <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>zipkin</artifactId>
       <type>test-jar</type>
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
thanks for opening up the topic for conversation. good start and helpful.

So, at the component abstraction, we don't use anything that binds environment variables, as configuration can vary wildly depending on the framework used. Firstly, I'm curious if things "just work" if the java system provided trust store is present.. ex. `-Djavax.net.ssl.trustStore`

If not, we'd want to do the following:
- identify the surface area we need for SSL configuration needed
- figure out how to make it possible to configure
  - ex via a direct option, or exposing or accepting a Cluster.Builder

cc @openzipkin/cassandra

codefromthecrypt:
@michaelsembwever any thoughts on this? do you know if the default trust store works, or if configuration here is inevitable for TLS client auth

ys:
I have the feeling that looking at the driver, the `withSSL` part is inevitable for cassandra. 
With that part, I think that using the default keystore by be enough... but would be rad to have a basic support with just the `withSSL` merged already so that we can try it.

Nobody ever thought about using encrypted communication between server and Cassandra?

ys:
According to the docs https://datastax.github.io/java-driver/manual/ssl/#jsse-property-based it should work

michaelsembwever:
Can we close this in favour of https://github.com/openzipkin/zipkin/pull/1401 ?

codefromthecrypt:
I think most are interested in the other PR. Let's go with that for now! Thanks for starting the gears turning, @jroes

PR #1400, opened 2016-11-14
Limit some ES indexing overhead and document C* limits

codefromthecrypt:
This removes the explicit mixed-case support in the ES index because it
adds overhead. For example, service and span names are downcased prior
to storage already. This also limits string indexing to the same as C*:
256 characters or less.

This also clarifies the cassandra3 README and backfills a test to make
sure we don't accidentally drop large annotations when we choose to not
index them.


codefromthecrypt:
@openzipkin/elasticsearch @openzipkin/cassandra PTAL: this is documenting and enforcing things that hopefully are not surprising

PR #1401, opened 2016-11-14
Use default withSSL() settings for cassandra

ys:
This is an optin to use `withSSL` for the cassandra driver.

This will require the java keystore and so on to be configured correctly on the server so that the cassandra certificate is present.

I have not tested this yet but wanted more feedback.

This is a simplification of #1396

### DIFF
--- a/zipkin-server/src/main/resources/zipkin-server.yml
+++ b/zipkin-server/src/main/resources/zipkin-server.yml
@@ -75,6 +75,8 @@ zipkin:
       ensure-schema: ${CASSANDRA3_ENSURE_SCHEMA:true}
       # how many more index rows to fetch than the user-supplied query limit
       index-fetch-multiplier: ${CASSANDRA3_INDEX_FETCH_MULTIPLIER:3}
+      # Using ssl for connection, rely on Keystore
+      use-ssl: ${CASSANDRA3_USE_SSL:false}
     elasticsearch:
       cluster: ${ES_CLUSTER:elasticsearch}
       # host is left unset intentionally, to defer the decision
--- a/zipkin-server/src/main/resources/zipkin-server.yml
+++ b/zipkin-server/src/main/resources/zipkin-server.yml
@@ -58,6 +58,8 @@ zipkin:
       index-cache-ttl: ${CASSANDRA_INDEX_CACHE_TTL:60}
       # how many more index rows to fetch than the user-supplied query limit
       index-fetch-multiplier: ${CASSANDRA_INDEX_FETCH_MULTIPLIER:3}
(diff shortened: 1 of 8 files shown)


zipkin-server/src/main/resources/zipkin-server.yml
```diff
@@ -58,6 +58,8 @@ zipkin:
       index-cache-ttl: ${CASSANDRA_INDEX_CACHE_TTL:60}
       # how many more index rows to fetch than the user-supplied query limit
       index-fetch-multiplier: ${CASSANDRA_INDEX_FETCH_MULTIPLIER:3}
+      # Using ssl for connection, rely on Keystore
+      use-ssl: ${CASSANDRA_USE_SSL:false}
```
codefromthecrypt, line 62: I think you need to affect another file ZipkinCassandraStorageProperties for this to take.

ys: Did it in https://github.com/openzipkin/zipkin/pull/1401/commits/75d19a6b95cae017207d8a767418e78bef1c34c7

codefromthecrypt:
try testing with the build instructions at the bottom of zipkin-server/README if it works, add the property to the others in that file. I presume this is only about encryption, not auth? or is there truststore (via sys props or otherwise) that would handle this?

llinder:
I like this better than the approach in #1396.  Further configuration of the driver SSL context beyond the defaults feels like it should supplied as Spring Boot auto configuration extension.  That is obviously a bigger change and doesn't make sense if the default driver SSL context works.

ys:
It works. I confirmed locally.
Could we get this landed with a version bump?

codefromthecrypt:
looks good. still needs to be added to zipkin-server/README (with the other cassandra settings), then merge time!

codefromthecrypt:
ps we'll likely need a cassandra3 port of this in the future. if you feel up to it, go ahead and put it in. Otherwise, someone else will need to replicate this work there

michaelsembwever:
LGTM, including 
- @adriancole 's comments on README
- need to add this also to cassandra3
- squash the commits before merge

ys:
@michaelsembwever @adriancole  I did the port to CASSANDRA3, Added the info to the readme and the squash can be done here via the merge button.

codefromthecrypt:
@ys hate to be a nag, but there's still no update to zipkin-server/README here.. do you mind adding the new env variables to there (for both cassandra and cassandra3)? that finishes the job!

ys:
Forgot to push it -_- @adriancole Nagging is fine:D

codefromthecrypt:
👍

PR #1402, opened 2016-11-14
change for /zipkin context root

jaredstehler:


jaredstehler:
sorry, only meant this for my fork

PR #1403, opened 2016-11-15
Adds Elasticsearch 5.x support via storage type: elasticsearch-http

### LINKED ISSUE #1312: Support Elasticsearch 5

### LINKED ISSUE #1318: Can we help uncover or explain span reporting lag?

codefromthecrypt:
This allows use of Elasticsearch 5.x and its notable ingest pipeline
feature when using `zipkin-storage-elasticsearch-http`.

It is important to note that `zipkin-storage-elasticsearch` remains
pinned to ES 2.x libraries as they are compile incompatible with 5.x.
In other words, you must use http if you want to use ES 5 (for now).

Version detection is implemented in order to choose the correct index
template format for the major version number. The implementation of
such is string manip, as it was less work than making a new type.

This adds a new parameter `ES_PIPELINE` which allows you to manipulate
the json sent by Zipkin collector before it is indexed. This could be
used for many things including cleaning service names or adding ingest
timestamps.

Integration tests run version 2.x on CircleCI and 5.x on Travis

Fixes #1312
Fixes #1318

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,8 +40,8 @@ before_install:
   - git config credential.helper "store --file=.git/credentials"
   - echo "https://$GH_TOKEN:@github.com" > .git/credentials
 
-  # Manually install elasticsearch until https://github.com/travis-ci/apt-source-whitelist/issues/190
-  - curl -SL https://download.elasticsearch.org/elasticsearch/release/org/elasticsearch/distribution/tar/elasticsearch/2.2.1/elasticsearch-2.2.1.tar.gz | tar xz
+  # Manually install elasticsearch 5 (since 2.x is default in travis)
+  - curl -SL https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-5.0.0.tar.gz | tar xz
   - elasticsearch-*/bin/elasticsearch -d > /dev/null
 
   # Manually install and run zk+kafka as it isn't an available service
--- a/zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/CallbackListenableFuture.java
+++ b/zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/CallbackListenableFuture.java
@@ -40,13 +40,28 @@ class CallbackListenableFuture<V> extends AbstractFuture<V> implements okhttp3.C
     call.cancel();
   }
 
-  @Override public void onResponse(Call call, Response response) throws IOException {
+  /** Note: this runs on the {@link okhttp3.OkHttpClient#dispatcher() dispatcher} thread! */
+  @Override public void onResponse(Call call, Response response) {
     try (ResponseBody responseBody = response.body()) {
       if (response.isSuccessful()) {
         set(convert(responseBody));
       } else {
         setException(new IllegalStateException("response failed: " + response));
       }
+    } catch (Throwable t) {
+      propagateIfFatal(t);
+      setException(t);
+    }
+  }
+
+  // Taken from RxJava, which was taken from scala
+  static void propagateIfFatal(Throwable t) {
+    if (t instanceof VirtualMachineError) {
+      throw (VirtualMachineError) t;
+    } else if (t instanceof ThreadDeath) {
+      throw (ThreadDeath) t;
+    } else if (t instanceof LinkageError) {
+      throw (LinkageError) t;
     }
   }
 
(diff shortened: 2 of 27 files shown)


codefromthecrypt:
cc @openzipkin/elasticsearch and @dragontree101 @dan-tr @shakuzen and @mansu particularly (as they had a lot of interest here)

codefromthecrypt:
circleci died trying to start elasticsearch 5, so swapped and have travis running it (which has more resources)

codefromthecrypt:
here's the playbook I used to test this.

Install elasticsearch 5 by downloading and running it

``` bash
$ curl -SL https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-5.0.0.tar.gz | tar xz
$ elasticsearch-*/bin/elasticsearch
```

Add a custom pipeline to it

``` bash
$ curl -X PUT -s localhost:9200/_ingest/pipeline/zipkin -d '{
  "description" : "add collector_timestamp_millis",
  "processors" : [
    {
      "set" : {
        "field": "collector_timestamp_millis",
        "value": "{{_ingest.timestamp}}"
      }
    }
  ]
}'
```

Start zipkin, pointed at that pipeline

``` bash
$ STORAGE_TYPE=elasticsearch ES_HOSTS=http://localhost:9200 ES_PIPELINE=zipkin java -jar zipkin-server/target/zipkin-*exec.jar
```

Add some traces (ex using one of the examples), then check the resulting index

``` bash
$ curl -s localhost:9200/zipkin-2016-11-15/_search|jq '.hits.hits[]._source.collector_timestamp_millis'
```

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/HttpClientBuilder.java
```diff
@@ -53,9 +53,14 @@ public static HttpClientBuilder create(OkHttpClient client) {
     return this;
   }
 
-  /** Default true. true implies that spans will be gzipped before transport. */
```
codefromthecrypt, line 56: this was dead code!

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/CallbackListenableFuture.java
```diff
@@ -40,13 +40,28 @@
...
-  @Override public void onResponse(Call call, Response response) throws IOException {
+  /** Note: this runs on the {@link okhttp3.OkHttpClient#dispatcher() dispatcher} thread! */
+  @Override public void onResponse(Call call, Response response) {
     try (ResponseBody responseBody = response.body()) {
       if (response.isSuccessful()) {
         set(convert(responseBody));
       } else {
         setException(new IllegalStateException("response failed: " + response));
       }
+    } catch (Throwable t) {
```
codefromthecrypt, line 51: running against mixed versions increases the sorts of exception we can get, so tightened this and backfilled test cases.

.travis.yml
```diff
@@ -41,7 +41,7 @@ before_install:
   - echo "https://$GH_TOKEN:@github.com" > .git/credentials
 
   # Manually install elasticsearch until https://github.com/travis-ci/apt-source-whitelist/issues/190
```
shakuzen: I was curious and looked at that issue. It looks like ES 2.x is whitelisted now.

shakuzen (approved):

.travis.yml
codefromthecrypt: I'll change that comment, since we've a different reason for using curl (to get v5)

PR #1404, opened 2016-11-15
be more forgiving of IPv4-mapped addresses in json

cburroughs:
Currently Zipkin checks that all IPv6 addresses are 16 bytes long.
However, if an IPv4-mapped addresses such as `::ffff:192.168.1.128` is
given then InetAddress.getByName will return an Inet4Address and the
assertion will fail.

Instead accept the address, but store it as ipv4 to maintain all
existing assertions.  This absolves clients of reconciling the nuances
of IPv4-mapped addresses in their environments (for example, in node
`net.isIPv6('::ffff:192.168.1.128')` is true) with Java's.

### DIFF
--- a/zipkin/src/main/java/zipkin/internal/JsonCodec.java
+++ b/zipkin/src/main/java/zipkin/internal/JsonCodec.java
@@ -19,6 +19,7 @@ import com.google.gson.stream.MalformedJsonException;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStreamReader;
+import java.net.InetAddress;
 import java.net.Inet6Address;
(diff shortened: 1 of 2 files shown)


cburroughs:
I'm having trouble understanding the build infrastructure enough to know what failed with circleci.  When I click on the details tab I see "Your build ran 256 tests in junit with 0 failures" :-/

If I click on container 1 (does circleci run two instances of the test script?) I see a failure in `postSpans_disconnectDuringBody` but in container 2  there is a failure in `usesInterceptorsQualifiedWith_zipkinElasticsearchHttp`.

codefromthecrypt:
no worries. nice work.

PR #1410, opened 2016-11-18
integrate TestContainers

bsideup:
to simplify Elasticsearch HTTP testing by starting a Docker container with Elasticsearch


zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/HttpElasticsearchTestGraph.java
```diff
@@ -16,13 +16,35 @@
...
+    new Lazy<String>() {
+
+      @Override
+      protected String compute() {
+        try {
+          // TODO test different ES versions by using other images from Docker Hub
+          GenericContainer container = new GenericContainer("elasticsearch:2")
+              .withExposedPorts(9200)
+              .waitingFor(new HttpWaitStrategy().forPath("/"));
+          container.start();
```
bsideup: wouldn't be started until `.start()`. 
Will be automatically destroyed on JVM shutdown. 
Can be manually destroyed with `container.stop()`.

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/HttpElasticsearchTestGraph.java
```diff
@@ -16,13 +16,35 @@
...
 
+  public Lazy<String> endpoint =
+    new Lazy<String>() {
+
+      @Override
+      protected String compute() {
+        try {
+          // TODO test different ES versions by using other images from Docker Hub
+          GenericContainer container = new GenericContainer("elasticsearch:2")
+              .withExposedPorts(9200)
```
bsideup: read as "map random port on my host to 9200 inside a container". No port conflicts.

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/HttpElasticsearchTestGraph.java
```diff
@@ -16,13 +16,35 @@
...
   INSTANCE;
 
+  public Lazy<String> endpoint =
+    new Lazy<String>() {
+
+      @Override
+      protected String compute() {
+        try {
+          // TODO test different ES versions by using other images from Docker Hub
+          GenericContainer container = new GenericContainer("elasticsearch:2")
```
bsideup: "elasticsearch:5" works as well, wasn't sure which one to use. Also, can be set dynamically, of course. Maybe two separate test suites to test different ES versions (even locally)

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/HttpElasticsearchTestGraph.java
```diff
@@ -16,13 +16,35 @@
...
+              .withExposedPorts(9200)
+              .waitingFor(new HttpWaitStrategy().forPath("/"));
+          container.start();
+          return String.format("http://%s:%d", container.getContainerIpAddress(), container.getMappedPort(9200));
+        } catch (Exception e) {
```
bsideup: will throw if Docker is not available (but will try to find it with different strategies like Docker Machine, Docker for Mac or native Docker on linux

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/ElasticsearchStrictTraceIdFalseTest.java
```diff
@@ -31,7 +31,7 @@ public ElasticsearchStrictTraceIdFalseTest() throws IOException {
     storage = ElasticsearchStorage.builder(
         HttpClientBuilder.create(new OkHttpClient())
             .flushOnWrites(true)
-            .hosts(ImmutableList.of("http://localhost:9200")))
+            .hosts(ImmutableList.of(HttpElasticsearchTestGraph.INSTANCE.endpoint.get())))
```
bsideup: wasn't sure how to fix it, decided to use endpoint's getter. Is it fine?

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/HttpElasticsearchTestGraph.java
codefromthecrypt: ideally we'd want to test both versions without changing configuration, so that the module fails if 2.x or 5.x fail. That way people don't have to remember to run the module's tests twice

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/ElasticsearchStrictTraceIdFalseTest.java
codefromthecrypt: probably :)

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/HttpElasticsearchTestGraph.java
```diff
@@ -16,13 +16,35 @@
...
+          // Use localhost if we failed to start a container (i.e. Docker is not available)
+          return "http://localhost:9200";
```
codefromthecrypt: ah interesting idea!

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/HttpElasticsearchTestGraph.java
bsideup: Ok, it's just a string after all. Do you have any good ideas how to make it use both quickly? I Mean, in standard JUnit I would use Parameterized runner, for instance, but Zipkin's testing infrastructure is a bit more advanced

codefromthecrypt (commented):
clever. can you check the travis/circleci logs to see if it actually worked? not sure if our CI config runs docker or not.

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/HttpElasticsearchTestGraph.java
codefromthecrypt: just copy/paste the classes and rename them to like ElasticsearchHttp2xSpanStoreTest? (for now.. we can do more fancy refactoring later, but having a class name like this is very simple to deal with, and we won't have more than 2 versions to check)

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/HttpElasticsearchTestGraph.java
codefromthecrypt: neat. once we port-in changes like this, we could probably do something to stop as necessary. For example, the zookeeper tests kill a ZK node to test resilience. we'll eventually need things like this I think.

bsideup:
@adriancole if you think that it's fine to change your testing infrastructure right now, I can make sure that Travis will run your tests with TestContainers already :)

codefromthecrypt:
we run 2 CI jobs right now, travis and also circle ci. If you pick one of them in this PR and instead of starting ES, start docker, we can get experience. We'd still be covered as the other is running ES normally. sg?

bsideup:
@adriancole added Travis, I see debug message `Will use TestContainers Elasticsearch instance running at http://localhost:32769` and storage/elasticsearch-http tests are green:

```
Results :
Tests run: 100, Failures: 0, Errors: 0, Skipped: 0
```

But I got one failure:

```
usesInterceptorsQualifiedWith_zipkinElasticsearchHttp(zipkin.storage.elasticsearch.http.ZipkinElasticsearchHttpStorageAutoConfigurationTest)  Time elapsed: 0.147 sec  <<< FAILURE!
java.lang.AssertionError: 
Actual and expected have the same elements but not in the same order, at index 0 actual element was:
  <zipkin.storage.elasticsearch.http.ZipkinElasticsearchHttpStorageAutoConfigurationTest$InterceptorConfiguration$$Lambda$3/257608605@21282ed8>
whereas expected element was:
  <zipkin.storage.elasticsearch.http.ZipkinElasticsearchHttpStorageAutoConfigurationTest$InterceptorConfiguration$$Lambda$2/1636050357@36916eb0>
```

Will investigate and do 2.x.x/5.x.x a bit later

codefromthecrypt:
@bsideup that's a flake that's unrelated. I'll see about fixing it

.travis.yml
```diff
@@ -24,6 +24,7 @@ jdk:
   - oraclejdk8
 
 services:
+  - docker
```
codefromthecrypt, line 27: so I think the real real test is removing the manual installation of elasticsearch further down this file and seeing if the tests still run

bsideup:
Hey @adriancole,

I did some major refactorings to ES-related tests, also switched Travis to run Docker+TestContainers-based mode. 

Could you please take a look? You might not agree with some of them, so let me know if you want me to change something or duplicate some classes to remove the dependency :)

.travis.yml
bsideup: done :+1:

codefromthecrypt: does caching of docker image downloads happen incidentally, or would we need separate setup for that?

pom.xml
```diff
@@ -210,6 +211,13 @@
 
       <dependency>
         <groupId>${project.groupId}</groupId>
+        <artifactId>zipkin-storage-elasticsearch</artifactId>
```
codefromthecrypt: yeah I was hoping to not publish another jar in support of this..

codefromthecrypt:
so looks like progress for sure. I particularly like the enclosed runner and think regardless of testcontainers we should switch to that. Really excited about this.

As you guessed I'm sure, I'm not keen on publishing a new test jar to avoid copy/pasting a 100 line text fixture. Copy/paste is fine during experimentation. Can you please reverse that part?
Details of why: -http and not already have setup due to the clients involved (ex sometimes I add a logging interceptor to the http one). Also, if this experiment proves itself, we'd put the testcontainers fixture in the existing zipkin test jar, not in the elasticsearch one. That would obviate a lot of the duplication. TL;DR; I really don't want to taint our public maven namespace for tactical reasons. Let's wait it out and see.

Also, I noticed the test duration went up by 2+ minutes, which is a significant bump. Wondering if this is mostly downloading the elasticsearch image or not. 2 minutes is a big enough issue to dig into as if we are to use testcontainers generically, we might be significantly increasing our build time as we port this to more things. Let me know what the source of the lag is (if you notice where it is). For example, a travis cache config might help (or might not)

bsideup:
@adriancole ok, I will remove the dependency by duplicating the code

> Wondering if this is mostly downloading the Elasticsearch image or not

I spotted this as well, but... 
TestContainers does a good logging, so it's pretty easy to verify:

```
10:07:19.260 [main] INFO  o.t.d.DockerClientProviderStrategy - Found docker client settings from environment
10:07:19.263 [main] INFO  o.t.d.DockerClientProviderStrategy - Docker host IP address is localhost
10:07:19.287 [main] INFO  o.t.d.DockerClientProviderStrategy - Looking for Docker environment. <...>
10:07:19.610 [main] INFO  o.testcontainers.DockerClientFactory - Connected to docker: <...>
10:07:21.414 [main] INFO  o.testcontainers.DockerClientFactory - Disk utilization in Docker environment is 64% (10261 MB available )
```

Docker environment detection took 2 seconds in total.

Images:

```
10:07:21.442 [main] INFO  🐳 [elasticsearch:5.0.1] - Pulling docker image: elasticsearch:5.0.1. Please be patient; this may take some time but only needs to be done once.
10:07:35.774 [main] INFO  🐳 [elasticsearch:5.0.1] - Creating container for image: elasticsearch:5.0.1
10:07:35.864 [main] INFO  🐳 [elasticsearch:5.0.1] - Starting container with ID: 6ec98192cd60c9540d4a8c1838d53d2c1a19339c75e4804adb89f03ea00fb315
10:07:36.071 [main] INFO  🐳 [elasticsearch:5.0.1] - Container elasticsearch:5.0.1 is starting: 6ec98192cd60c9540d4a8c1838d53d2c1a19339c75e4804adb89f03ea00fb315
10:07:36.320 [main] INFO  🐳 [elasticsearch:5.0.1] - Waiting for 60 seconds for URL: http://localhost:32769/
10:07:46.462 [main] INFO  🐳 [elasticsearch:5.0.1] - Container elasticsearch:5.0.1 started
```

5.x took 14 seconds to download. Once we migrate to Zipkin's image, it should be lower.

```
10:07:48.114 [main] INFO  ?.16.1] - Pulling docker image: openzipkin/zipkin-elasticsearch:1.16.1. Please be patient; this may take some time but only needs to be done once.
10:07:54.639 [main] INFO  ?.16.1] - Creating container for image: openzipkin/zipkin-elasticsearch:1.16.1
10:07:54.718 [main] INFO  ?.16.1] - Starting container with ID: 6afcf6613742fd05883ec82b27017473d42d8882bcb0a03c90fcc7a52f2bd46c
10:07:54.898 [main] INFO  ?.16.1] - Container openzipkin/zipkin-elasticsearch:1.16.1 is starting: 6afcf6613742fd05883ec82b27017473d42d8882bcb0a03c90fcc7a52f2bd46c
10:07:54.909 [main] INFO  ?.16.1] - Waiting for 60 seconds for URL: http://localhost:32771/
10:08:01.968 [main] INFO  ?.16.1] - Container openzipkin/zipkin-elasticsearch:1.16.1 started
```

Only 6 second for Zipkin's 2.x. Startup took 7 seconds (I think we can tune Elasticsearch a bit to make it start faster)
`SpanConsumerTest` is slow itself. Right now we test both 2.x and 5.x, it means that the time was increased by test executing mostly. 

Here is a screenshot from IntelliJ IDEA with test execution times:
screen shot 2016-11-20 at 10 19 25 am
So I would assume that most of the time is not an overhead from TestContainers, but the second test suite for ES 5.x

We can configure both 2.x and 5.x on Travis, start them, and then run the tests again.Numbers will be pretty the same :)

pom.xml
```diff
@@ -73,6 +73,7 @@
     <assertj.version>3.5.1</assertj.version>
     <hamcrest.version>1.3</hamcrest.version>
     <okhttp.version>3.4.1</okhttp.version>
+    <testcontainers.version>1.1.6</testcontainers.version>
```
rnorth: 1.1.7 is out now 😁

codefromthecrypt:
ahh right. of course. we are running both now. geez we need to instrument
maven and surefire with zipkin, right?

Thanks for the analysis!

pom.xml
bsideup: :D changed :D

bsideup:
@adriancole well, you can always migrate to Gradle and use their build scans ;) Gradle can also share some code without having to publish an artifact. Just saying =) 

FYI I pushed the changes. Now there is no dependency between es and es-http.

.travis.yml
bsideup: Docker image caching is a hot topic in Travis world :) It's not supported by Travis, however you can cache it manually with "docker export", "docker import". But! Travis devs were saying that it's actually slower than pulling from Docker Hub :D

codefromthecrypt:
good to know. thanks!

PR #1411, opened 2016-11-20
Fix javadoc publish even more. This shouldn't be that hard.

abesto:


codefromthecrypt:
Thanks, for whack-a-mole'ing!

PR #1413, opened 2016-11-21
Moves ES 5 docker to new base image; deletes broken transport 5 test

codefromthecrypt:
This moves the ES 5 docker tests to our new base image, which
significantly reduces first-time download size (particularly helpful for
those on slow networks).

It also deletes the `ElasticsearchV5TransportTest` as it will never work
since the 2.x protocol is incompatible with 5.x and we can only use one.

Finally, it reformats imports so they match the others.


zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/ElasticsearchV5HttpTest.java
```diff
@@ -13,19 +13,18 @@
...
 
-import java.io.IOException;
-
 @RunWith(Enclosed.class)
 public class ElasticsearchV5HttpTest {
 
   @ClassRule
   public static LazyElasticsearchHttpStorage storage =
-      new LazyElasticsearchHttpStorage("elasticsearch:5.0.1");
+      new LazyElasticsearchHttpStorage("openzipkin/zipkin-elasticsearch5:1.16.1");
```
codefromthecrypt, line 27: We could consider removing bookkeeping at the slight chance of error by using the "latest" tag here instead of a specific version. This works fine on master, since it is always ahead of docker and shouldn't break. If/when we start doing a long-term release branch (like 1.x) we might consider changing that. Food for thought.

bsideup: Problems with `latest`:
1) inconsistency - test might work locally (cache), but fail on CI
2) Minor ES release (i.e. 5.1) might break your app. Happened to us with 2.x branch :)

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/ElasticsearchDependenciesTest.java
```diff
@@ -14,6 +14,8 @@
 package zipkin.storage.elasticsearch;
 
 import com.google.common.base.Throwables;
+import java.io.IOException;
+import java.util.List;
```
bsideup, line 18: Sorry :)

Would be really helpful to integrate something like Eclipse Code Formatter, so it will fail if the code style is broken (the same as eslint fails on the frontend)

I use IntelliJ IDEA and their vision of imports is a bit different :)

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/ElasticsearchV5HttpTest.java
codefromthecrypt: ok maybe something like a DockerZipkin class in zipkin's core test jar? That way people can run the tests in an IDE with no config, and can also override manually (as needed sometimes).

ex.
```java
  new LazyElasticsearchHttpStorage("openzipkin/zipkin-elasticsearch5", DockerZipkin.VERSION);
```

which would allow someone to easily temporarily adjust:
ex.
```java
  new LazyElasticsearchHttpStorage("openzipkin/zipkin-elasticsearch5", "1.6.2");
```

Main thing I'm looking for is to keep the layers consistent (to reduce downloading for folks like me), but yeah without making it hard to start the tests in the IDE.

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/ElasticsearchDependenciesTest.java
codefromthecrypt: right now we use square code style (noted in the CONTRIBUTING section), and there's a PR for eclipse as yet unmerged https://github.com/square/java-code-styles/pull/32

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/ElasticsearchV5HttpTest.java
bsideup: If the version is consistent across all images under openzipkin then we can extract some base class for all TestContainers-based storages and just pass an image name to it. Base class will append version stored as a private constant inside it.

WDYT?

codefromthecrypt:
ps it will be non-storage, too. ex we ought to do the same thing for Kafka.

Issue #1300, opened 2016-09-15
UI: Show trace and span IDs


C-Otto:
The UI currently only shows the trace ID as part of the URL, and does not show the spans' IDs at all. In order to find related log messages in, for example, Kibana (ELK stack), I'd like to copy-paste the span's ID. However, the UI does not provide this information in a user-friendly way.

codefromthecrypt:
Span detail view could show this I guess

PR #1414, opened 2016-11-21
Add logs button on the trace page.

fedj:
When troubleshooting our system, we often need to check
if there were logs. Since a good pratice is to add the
trace id to your logs, it seems really natural to have
a button to open a link to query your logs with this id.
Also, people generally have kibana to search for specific
logs (or at least have a query-able log service).

The button is deactivated by default (if you don't have a log
query service or the trace id in the logs) and it's simply
activated by adding LOGS_URL="http://company.com/logs?query={traceId}".
The traceId will be contextually replaced.

Here's a preview:
logs button

### DIFF
--- a/zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiProperties.java
+++ b/zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiProperties.java
@@ -15,6 +15,7 @@ package zipkin.autoconfigure.ui;
 
 import java.util.concurrent.TimeUnit;
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.util.StringUtils;
 
 @ConfigurationProperties("zipkin.ui")
 public class ZipkinUiProperties {
@@ -22,6 +23,7 @@ public class ZipkinUiProperties {
   private int queryLimit = 10;
   private int defaultLookback = (int) TimeUnit.DAYS.toMillis(7);
   private String instrumented = ".*";
+  private String logsUrl = null;
 
   public int getDefaultLookback() {
     return defaultLookback;
@@ -53,5 +55,15 @@ public class ZipkinUiProperties {
 
   public void setInstrumented(String instrumented) {
     this.instrumented = instrumented;
+  }
+
+  public String getLogsUrl() {
+    return logsUrl;
+  }
+
+  public void setLogsUrl(String logsUrl) {
+    if (!StringUtils.isEmpty(logsUrl)) {
+      this.logsUrl = logsUrl;
+    }
   }
 }
--- a/zipkin-server/src/main/resources/zipkin-server.yml
+++ b/zipkin-server/src/main/resources/zipkin-server.yml
@@ -111,7 +111,9 @@ zipkin:
     # - .*example2.com
     # Default is "match all websites"
     instrumented: .*
-
+    # Url of the logs associated with a given request.
+    # It will be formatted with the trace id if your url contains {traceId} (e.g. http://kibana/trace{traceId})
+    logs-url: ${LOGS_URL:}
 server:
   port: ${QUERY_PORT:9411}
   compression:
--- a/zipkin-ui/js/component_data/trace.js
+++ b/zipkin-ui/js/component_data/trace.js
@@ -3,13 +3,22 @@ import $ from 'jquery';
 import {getError} from '../../js/component_ui/error';
 import traceToMustache from '../../js/component_ui/traceToMustache';
 
+export function toContextualLogsUrl(logsUrl, traceId) {
+  if (logsUrl) {
+    return logsUrl.replace('{traceId}', traceId);
+  }
+  return logsUrl;
+}
+
 export default component(function TraceData() {
   this.after('initialize', function() {
-    $.ajax(`/api/v1/trace/${this.attr.traceId}`, {
+    const traceId = this.attr.traceId;
+    const logsUrl = toContextualLogsUrl(this.attr.logsUrl, traceId);
+    $.ajax(`/api/v1/trace/${traceId}`, {
       type: 'GET',
       dataType: 'json'
     }).done(trace => {
-      const modelview = traceToMustache(trace);
+      const modelview = traceToMustache(trace, logsUrl);
       this.trigger('tracePageModelView', {modelview, trace});
     }).fail(e => {
       this.trigger('uiServerError',
--- a/zipkin-ui/templates/trace.mustache
+++ b/zipkin-ui/templates/trace.mustache
@@ -4,6 +4,9 @@
     <li class=''><a href='#'><strong>Services:</strong> <span class='badge'>{{services}}</span></a></li>
     <li class=''><a href='#'><strong>Depth:</strong> <span class='badge'>{{depth}}</span></a></li>
     <li class=''><a href='#'><strong>Total Spans:</strong> <span class='badge'>{{totalSpans}}</span></a></li>
+    {{#logsUrl}}
+    <li class='navbar-right'><a href='{{ logsUrl }}' id='traceLogsLink'><span class='btn btn-primary btn-xs badge'>Logs</span></a></li>
+    {{/logsUrl}}
     <li class='navbar-right'><a href='/api/v1/trace/{{ traceId }}' id='traceJsonLink'><span class='btn btn-primary btn-xs badge'>JSON</span></a></li>
   </ul>
 
--- a/zipkin-ui/test/component_ui/traceToMustache.test.js
+++ b/zipkin-ui/test/component_ui/traceToMustache.test.js
@@ -52,6 +52,12 @@ describe('traceToMustache', () => {
     modelview.services.should.equal(3);
   });
 
+  it('should show logsUrl', () => {
+    const logsUrl = 'http/url.com';
+    const modelview = traceToMustache(trace, logsUrl);
+    modelview.logsUrl.should.equal(logsUrl);
+  });
+
   it('should show service counts', () => {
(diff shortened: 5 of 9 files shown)


codefromthecrypt:
cool start! wanna add a screen shot?

cc @openzipkin/elasticsearch @rogeralsing @eirslett @basvanbeek

codefromthecrypt:
cc'ing other folks who have interest in zipkin + kibana @marcingrzejszczak @dragontree101 @fedj @dan-tr @C-Otto

C-Otto:
This seems to solve a similar issue as described in #1300

zipkin-server/src/main/resources/zipkin-server.yml
```diff
@@ -111,7 +111,9 @@ zipkin:
     # - .*example2.com
     # Default is "match all websites"
     instrumented: .*
-
+    # Url of the logs associated with a given request.
+    # It will be formatted with the trace id if your url contains ${traceId}} (e.g. http://kibana/trace${traceId})
```
C-Otto: There are two } instead of one

zipkin-ui/js/component_data/trace.js
```diff
@@ -5,11 +5,14 @@ import traceToMustache from '../../js/component_ui/traceToMustache';
 
 export default component(function TraceData() {
   this.after('initialize', function() {
-    $.ajax(`/api/v1/trace/${this.attr.traceId}`, {
+    const traceId = this.attr.traceId;
+    const logsUrlNonFormatted = this.attr.logsUrl;
+    const logsUrl = logsUrlNonFormatted.replace('{traceId}', traceId);
```
C-Otto: I don't know JavaScript enough to be sure, but I assume that `${traceId}` and `{traceId}` are different when replacing?

C-Otto:
I like the idea! A screenshot would help a bit.

zipkin-server/src/main/resources/zipkin-server.yml
```diff
@@ -111,7 +111,9 @@ zipkin:
...
+    logsUrl: ${LOGS_URL:}
```
shakuzen: Small nit-pick: other variables are kebab-case in this file, so it might match the style better if this were too. Thanks to Spring Boot's relaxed property binding, it shouldn't effectively matter either way, though.

shakuzen (changes requested):
This sounds like a nice enhancement. Thanks for contributing. Adding a couple of tests for this new behavior would be extra nice.

1) Verify the absence of the button when the property is not set
2) Verify the presence of the button with the correctly replaced URL when the property is set.

fedj:
@C-Otto done
@shakuzen I'm not really familiar with javascript (actually not at all), would you have a pointer on how to do these kind of unit tests ?

zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiProperties.java
```diff
@@ -22,6 +22,7 @@
   private int queryLimit = 10;
   private int defaultLookback = (int) TimeUnit.DAYS.toMillis(7);
   private String instrumented = ".*";
+  private String logsUrl = "";
```
codefromthecrypt: I'd initialize this to null, so that you don't end up with an empty string in the json

zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiProperties.java
```diff
@@ -54,4 +55,12 @@ public String getInstrumented() {
   public void setInstrumented(String instrumented) {
     this.instrumented = instrumented;
   }
+
+  public String getLogsUrl() {
+    return logsUrl;
+  }
+
+  public void setLogsUrl(String logsUrl) {
```
codefromthecrypt, line 64: set if not null or empty

codefromthecrypt: I'd test this in ZipkinUiAutoConfigurationTest basically that when `zipkin.ui.logs-url:` (empty string), `getLogsUrl()` is null, and when it is set to a url, that's readable as `getLogsUrl()`. This helps make sure there's no typos hard to see.

zipkin-ui/js/component_data/trace.js
```diff
@@ -3,13 +3,22 @@ import $ from 'jquery';
 import {getError} from '../../js/component_ui/error';
 import traceToMustache from '../../js/component_ui/traceToMustache';
 
+function toContextualLogsUrl(logsUrl, traceId) {
+  if (logsUrl !== undefined) {
```
codefromthecrypt: I think you can shorten this to `if (logsUrl)` ?

codefromthecrypt: if you export this function, you should be able to test it by making a class test/component_data/trace.test.js 

something like... (please use the exact format of kibana search url you use as that helps folks understand how it is used)
```javascript
import {toContextualLogsUrl} from '../../js/component_data/trace';

describe('toContextualLogsUrl', () => {
  it('replaces token in logsUrl when set', () => {
    toContextualLogsUrl('http://some-realistic-logs-url/{traceId}', 'abcd').should.equal(''http://some-realistic-logs-url/abcd'');
  });
});
```

zipkin-ui/test/component_ui/traceToMustache.test.js
```diff
@@ -40,20 +40,21 @@ const span3 = span(12345, 'methodcall2', span3Id, span2Id, 300, 50, annotations3
 const span4 = span(12345, 'methodcall2', span4Id, span3Id, 400, 100, annotations4);
 
 const trace = [span1, span2, span3, span4];
+const logsUrl = '';
```
codefromthecrypt: I would make this method scope, and set it to undefined as I think that's what it would be in practice.
then, add a test for what happens when it isn't undefined. For example, traceToMustache result includes the logsUrl

zipkin-ui/templates/trace.mustache
```diff
@@ -4,6 +4,9 @@
     <li class=''><a href='#'><strong>Services:</strong> <span class='badge'>{{services}}</span></a></li>
     <li class=''><a href='#'><strong>Depth:</strong> <span class='badge'>{{depth}}</span></a></li>
     <li class=''><a href='#'><strong>Total Spans:</strong> <span class='badge'>{{totalSpans}}</span></a></li>
+    {{#logsUrl}}
```
codefromthecrypt, line 7: not sure there's an existing class to cover this, so one would need to be made.. I can help think of one, or maybe the notes I've made so far can get you there?

codefromthecrypt (commented):
I'm fairly novice at javascript, too, but hopefully the notes help. There's instructions on how to run tests in the zipkin-ui README iirc.

It would be nice to see how it looks when someone clicks the link, too. Ex does it pop out, etc? feel free to use something like OSX preview, tools, annotate to black-out anything sensitive.

zipkin-ui/templates/trace.mustache
fedj: Done for all but for the mustache test. I'm not sure to see how to do it properly here.

codefromthecrypt:
@haf what do you think about something like this and logary?

codefromthecrypt:
@fedj mind sharing a screen shot of the logs detail? is it a pop-out or a new tab?

haf:
@adriancole I think it would be excellent. We have a Kibana dashboard at qvitoo and frequently need to correlate logs. Logary does all of the above collecting too. I'm working through the https://github.com/logary/logary/issues and have been for a while now; the next two parts is creating a complete C# facade like the existing Logary.Facade, improving how derived metrics are configured and then create examples including ones with Zipkin in them. This would enable e.g. sending a snapshot 1 minute rolling average together with the span, together with the annotations, that are sent to Zipkin.

However, Logary's API is now well geared towards the logging structure needed for Zipkin; hierarchical passing-down of loggers, applying middleware on the fly (with TimeScope), or alternatively correlating span annotations in the Logary Zipkin target. So the extra work needed to fully support Zipkin now, is very small.

fedj:
@adriancole Currently it's just a link so it depends on your browser settings (I guess). I think it will open the link in the same tab by default (at least this is what it does on Safari without previewing it)

codefromthecrypt:
thanks for the help on this. It will give folks a way to experiment with log correlation without visible impact to those who aren't interested.

codefromthecrypt:
if you don't mind, please verify the snapshot exec jar acts as expected!
http://oss.jfrog.org/oss-snapshot-local/io/zipkin/java/zipkin-server/1.16.3-SNAPSHOT/

oh and probably we need to update zipkin-server/README.md with details on this feature.

jcarres-mdsol:
I think this PR lacked documentation on how to use this new feature.

codefromthecrypt:
yeah noticed that post-merge, too. we need to add the description of this PR or similar to zipkin-server/README.md

codefromthecrypt:
actually, the detailed stuff from the description should go to zipkin-ui/README.md and only description of the env variable to zipkin-server/README.md that way the latter doesn't grow endlessly

fedj:
@jcarres-mdsol, @adriancole You're both right, I'll do that tomorrow if you don't mind

fedj:
@adriancole It's working as expected for the last jar (-6-exec.jar).
I didn't forget about doc, will do another PR today.

codefromthecrypt:
:thumbsup: thanks for the follow-up

codefromthecrypt:
shortly, brave will include SLF4J support which could integrate with this https://github.com/openzipkin/brave/pull/389

PR #1419, opened 2016-11-23
Use Spring Boot config profiles

llinder:
This change enables custom Zipkin Server builds the ability to add extra config without copying the entire upstream config to the new project.  The main use case is to avoid copy/paste config duplication in the Zipkin AWS Server where additional AWS specific configuration is required.

/cc @adriancole @anuraaga


codefromthecrypt:
I've been meaning to move the component based config to their corresponding autoconfigure modules, but didn't know how. would profiles be the answer to that as well? cc @shakuzen

codefromthecrypt:
that said, if we moved them, it might be harder to figure out where to look for settings.. ps I think we'd need to look for a README link to change as it is..

codefromthecrypt:
I think the minimum change needed here is to change the README links.. I'll do that on the way in.

codefromthecrypt:
closed via 890bf402c01bc72197f295c4760f124317b9f52c

zhanghba:
@adriancole 
if we use springboot profiles to differentiate environment with a demo application. but not found on zipkin with elasticsearch. support it on zipkin?

thanks

PR #1421, opened 2016-11-29
Adds better error message when a binary annotation key or value is unset

codefromthecrypt:
When people are making instrumentation, they could accidentally leave
data missing. This gives a contextual error message of which binary
annotation was malformed.


codefromthecrypt:
thanks @nicolai86 for reporting this

nicolai86:
Thanks @adriancole for improving this so fast!

PR #1422, opened 2016-11-29
Allows elasticsearch-http to accept multiple hosts

### LINKED ISSUE #1420: only one es hosts?
i config es hosts is ```ES_HOSTS=http://10.215.242.15:9200, http://10.215.242.16:9200, http://10.215.242.17:9200, http://10.215.242.18:9200```
and zipkin'ui say ```ERROR: cannot load span names: Only a single hostname is supported [http://10.215.242.15:9200, http://10.215.242.16:9200, http://10.215.242.17:9200, http://10.215.242.18:9200]```
so i can only set one host?

codefromthecrypt:
Our first version of elasticsearch-http had a TODO to support multiple
base urls. This implements that feature using round-robin DNS.

Fixes #1420 

## Multiple hosts

Most users will supply a DNS name that's mapped to multiple A or AAAA
records. For example, `http://elasticsearch:9200` will use normal host
lookups to get the list of IP addresses.

You can alternatively supply a list of http base urls. This list is used
to recover from failures. Note that all ports must be the same, and the
scheme must be http, not https.

Here are some examples:

* http://1.1.1.1:9200,http://2.2.2.2:9200
* http://1.1.1.1:9200,http://[2001:db8::c001]:9200
* http://elasticsearch:9200,http://1.2.3.4:9200
* http://elasticsearch-1:9200,http://elasticsearch-2:9200

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,7 @@
     <mockito.version>1.10.19</mockito.version>
     <assertj.version>3.5.1</assertj.version>
     <hamcrest.version>1.3</hamcrest.version>
-    <okhttp.version>3.4.1</okhttp.version>
+    <okhttp.version>3.4.2</okhttp.version>
     <testcontainers.version>1.1.7</testcontainers.version>
 
     <animal-sniffer-maven-plugin.version>1.15</animal-sniffer-maven-plugin.version>
--- a/zipkin-server/README.md
+++ b/zipkin-server/README.md
@@ -155,6 +155,13 @@ $ STORAGE_TYPE=mysql MYSQL_USER=root java -jar zipkin.jar
 ```
 
 ### Elasticsearch Storage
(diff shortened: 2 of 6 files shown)


codefromthecrypt:
cc @openzipkin/elasticsearch

codefromthecrypt:
tested locally, by setting `ES_HOSTS=http://1.1.1.1:9200,http://localhost:9200`

Then, I did `curl -s localhost:9411/api/v1/services` which succeeded.

Here's what self-tracing showed (note.. it doesn't show the connection failure, but that's a Brave/OkHttp tracing issue)

screen shot 2016-11-29 at 4 00 40 pm

codefromthecrypt:
revised example which shows a slow trace where the connection auto-recovered

codefromthecrypt:
will be in zipkin 1.17.1

PR #1423, opened 2016-11-29
Add documentation on the logsUrl feature.

fedj:
With https://github.com/openzipkin/zipkin/pull/1414, we can add a button
to link directly the trace and its logs. This review adds the documentation
of this feature.

### DIFF
--- a/zipkin-server/README.md
+++ b/zipkin-server/README.md
@@ -100,6 +100,7 @@ environment | zipkin.ui.environment | The value here becomes a label in the top-
 defaultLookback | zipkin.ui.default-lookback | Default duration in millis to look back when finding traces. Affects the "Start time" element in the UI. Defaults to 3600000 (1 hour in millis).
 queryLimit | zipkin.ui.query-limit | Default limit for Find Traces. Defaults to 10.
 instrumented | zipkin.ui.instrumented | Which sites this Zipkin UI covers. Regex syntax. e.g. `http:\/\/example.com\/.*` Defaults to match all websites (`.*`).
+logsUrl | zipkin.ui.logs-url | Logs query service url pattern. If specified, a button will appear on the trace page and will replace {traceId} in the url by the traceId. Not required.
 
 For example, if using docker you can set `ZIPKIN_UI_QUERY_LIMIT=100` to affect `$.queryLimit` in `/config.json`.
 
--- a/zipkin-ui/README.md
+++ b/zipkin-ui/README.md
@@ -70,3 +70,21 @@ and then you can submit the span(s) via curl:
 ```bash
 $ curl -H "Content-Type: application/json" --data @span.json http://localhost:9411/api/v1/spans 
 ```
+
+#### How do I find logs associated with a particular trace
+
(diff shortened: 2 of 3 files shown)


zipkin-server/README.md
```diff
@@ -114,6 +115,7 @@ zipkin-server is a drop-in replacement for the [scala query service](https://git
     * `STORAGE_TYPE`: SpanStore implementation: one of `mem`, `mysql`, `cassandra`, `elasticsearch`
     * `COLLECTOR_PORT`: Listen port for the scribe thrift api; Defaults to 9410
     * `COLLECTOR_SAMPLE_RATE`: Percentage of traces to retain, defaults to always sample (1.0).
+    * `LOGS_URL`: Logs query service url pattern
```
codefromthecrypt: actually, looking at this here, it seems we probably shouldn't have given this a special env variable (since none of the other UI ones do). Can you remove the mapping from zipkin-server-shared.yml? Instead, it would be set normally with `ZIPKIN_UI_LOGS_URL` (which eliminates the need to duplicate some of the docs)

zipkin-ui/README.md
```diff
@@ -70,3 +70,18 @@ and then you can submit the span(s) via curl:
...
+
+One convenient way to switch from a trace to its logs is to get a button which directly links a trace to its logs.
+
+This feature can be activated adding this environment variable:
+
+`LOGS_URL=http://kibana.company.com/query={traceId}`
+
+where `{traceId}` will be contextually replaced by the trace id.
+
+If this feature is activated, you'll see on the trace detail page an additional button named `logs`
```
codefromthecrypt: add a link below to an example screenshot logs button

zipkin-ui/README.md
```diff
@@ -70,3 +70,18 @@ and then you can submit the span(s) via curl:
...
 ```
+
+#### How do I find logs associated with a particular trace
+
+Since zipkin provides a correlation id (the trace id), it's a good pattern to add it in your logs.
+If you use zipkin, it's very likely that you have distributed logging sytem also and a way to query your logs (e.g. ELK stack).
+
+One convenient way to switch from a trace to its logs is to get a button which directly links a trace to its logs.
+
+This feature can be activated adding this environment variable:
```
codefromthecrypt: this can be activated by setting the property `zipkin.ui.logs-url` or its corresponding environment variable:

then use the environment syntax

zipkin-ui/README.md
```diff
@@ -70,3 +70,18 @@ and then you can submit the span(s) via curl:
...
+
+`LOGS_URL=http://kibana.company.com/query={traceId}`
```
codefromthecrypt: @marcingrzejszczak do you have an example kibana link that looks up from the id pattern sleuth logs with slf4j?

codefromthecrypt (commented):
Thanks for putting this together! we can do a release shortly. made some comments.

fedj:
@adriancole Done except for the kibana url pattern that I miss

codefromthecrypt:
thanks tons!

PR #1424, opened 2016-11-29
use TestContainers for MySQL tests

bsideup:
With this PR we can test schema changes, also later we probably want to test different MySQL versions as well - also possible thanks to TestContainers.

It uses an official Zipkin MySQL Docker image by dropping an old schema before we apply a new one.


codefromthecrypt:
I'll try to help with this, or you can raise a PR on the zipkin-docker repo to make something more configurable. It is a step backwards to rely on a big image even if travis can pull it down quickly enough. I'm often in slow networks and am usually the one maintaining this code. 

Weird that testcontainers can't use mariadb.. can this be addressed? I'd like to not introduce any more dependencies that we'd have to track separately.

codefromthecrypt:
for example, if you need a more powerful user or something in order to drop schema, you could add it here: https://github.com/openzipkin/docker-zipkin/blob/master/mysql/install.sh#L35

In other words, let's try to make this all coherent. We use the same test images in our tests and ad-hoc docker testing. One docker image version... one base layer.

Zipkin's complex enough without having to remember a matrix of driver and docker images :P

-- all this said, I appreciate that if this were a project that didn't publish its own images for reasons described, I'd also be tempted to use whatever's the standard image was.

bsideup:
> Weird that testcontainers can't use mariadb

It can :) There is `mariadb` container as well. Just if we use `MySQLContainer` it will attempt to use MySQL driver to check if it's started. I can remove that, for sure

> One docker image version... one base layer.

Ok, I'll attempt to use Zipkin's image one more time, shouldn't be that hard after all :)

codefromthecrypt:
thanks for the tenacity sergei!

codefromthecrypt:
great! looks like it worked fine. 2 small things then merge time!
* change description of the PR to reflect today :)
* unhook circleci as I think that's still running the old mysql manually

bsideup:
@adriancole 
> change description of the PR to reflect today :)

done

> unhook circleci as I think that's still running the old mysql manually

I didn't touch CircleCI config in both Elasticsearch and MySQL PRs because I was thinking that we want to keep (for now) one of the CI services to use old testing approach without Docker, no?

codefromthecrypt:
thanks for checking me. yeah, let's keep them, though we should probably add a comment to the circle ci reminding us that we aren't using docker for services intentionally (to ensure docker-free tests work)

bsideup:
@adriancole well, lack of `- docker` service record is a good reminder I guess :D But ok, I can make it explicit :)

codefromthecrypt:
yeah just a reminder of why we aren't adding docker in case we forget.

IIRC we are not yet relying completely on it, nor ask those developing local to always have docker installed. By running on circle w/o docker we have a backup (in case we goofed in our new tests) and also implicitly test the non-docker way.

At some point we may yank the non-docker way of course.

bsideup:
@adriancole added :)

codefromthecrypt:
thanks again!

bsideup:
:+1:

PR #1425, opened 2016-12-02
Uses RetroLambda to get rid of IDE setup and cruft

### LINKED ISSUE #1256: Make zipkin easier to import into eclipse and intellij

codefromthecrypt:
This uses RetroLambda which allows you to use Java 8 source level even
if your bytecode level is Java 6. This makes coding more fun and also
obviates IDE setup needed to manage different source levels between main
and test code trees.

This also moves modules who don't use Java 8 types back to language
level 7.

The same has been in use by zipkin-reporter-java for some time.

See https://github.com/orfjackal/retrolambda
Closes #1256

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,10 +43,6 @@
     <main.java.version>1.7</main.java.version>
     <main.signature.artifact>java17</main.signature.artifact>
 
-    <!-- default bytecode version for src/test -->
-    <maven.compiler.source>1.8</maven.compiler.source>
-    <maven.compiler.target>1.8</maven.compiler.target>
-
     <main.basedir>${project.basedir}</main.basedir>
 
     <brave.version>3.13.0</brave.version>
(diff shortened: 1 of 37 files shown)


pom.xml
```diff
@@ -467,21 +467,10 @@
...
-              <goal>compile</goal>
-            </goals>
-            <configuration>
-              <source>${main.java.version}</source>
-              <target>${main.java.version}</target>
-            </configuration>
-          </execution>
-        </executions>
         <configuration>
+          <!-- Retrolambda will rewrite lambdas as Java 6 bytecode -->
```
shakuzen: Looks like `main.java.version` is 1.7.

shakuzen (commented):
How does this affect, for instance, the zipkin-server module that is already Java 8 source? Will it rewrite its lambdas into Java 7-compatible bytecode?

bsideup:
@adriancole great change! IDEA setup is painful indeed

codefromthecrypt:
@shakuzen good catch. I opted out of retrolambda for modules that actually need java 8 (as opposed to lazily using it just for lambdas)

pom.xml
```diff
@@ -433,55 +429,62 @@
...
+                    </configurator>
+                  </action>
+                </pluginExecution>
+              </pluginExecutions>
+            </lifecycleMappingMetadata>
+          </configuration>
+        </plugin>
+
         <plugin>
-                <groupId>org.eclipse.m2e</groupId>
```
codefromthecrypt, line 437: fixed formatting here as it was annoying me..

PR #1429, opened 2016-12-08
Allows arbitrary overriding of Kafka consumer properties

### LINKED ISSUE #1427: i think should change auto.offset.reset value by property.
i read zipkin-server's code 
```
props.put("auto.offset.reset", "smallest");
```

i think this property can be set `latest` or `smallest` by customer?

because of some time , kafka have a lot of message in topci, but i don't want to get old message , i should set `auto.offset.reset` to `lastest`.

codefromthecrypt:
You may need to override other consumer properties than what zipkin
explicitly defines. In such case, you need to prefix that property name
with "zipkin.collector.kafka.overrides" and pass it as a CLI argument or
system property.

For example, to override "overrides.auto.offset.reset", you can set a
prefixed system property:

```bash
```

Fixes #1427

### DIFF
--- a/zipkin-autoconfigure/collector-kafka/src/main/java/zipkin/autoconfigure/collector/kafka/ZipkinKafkaCollectorProperties.java
+++ b/zipkin-autoconfigure/collector-kafka/src/main/java/zipkin/autoconfigure/collector/kafka/ZipkinKafkaCollectorProperties.java
@@ -13,6 +13,9 @@
  */
 package zipkin.autoconfigure.collector.kafka;
 
+import java.util.LinkedHashMap;
+import java.util.Map;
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
cc @prat0318 @dgarson @SimenB @sveisvei @eirslett @kristofa @NegatioN @shakuzen

NegatioN:
Nice. :+1:  This is almost identical to the way zipkin-reporter does it, am I right? Double-plus for consistency.

codefromthecrypt:
>
> Nice. 👍 This is almost identical to the way zipkin-reporter does it, am
> I right? Double-plus for consistency.
>
yes exactly the same. keen eye!

PR #1433, opened 2016-12-14
really make "Expand All" the default view for a trace

cburroughs:
Previously the default expansion was either:
 * The `serviceName` query parameter, or
 * A heuristic roughly along the lines of "the immediate children of
   the root"

PR #1326 set the active button to "Expand All".  However, to mimic the
behavior of actually pressing "Expand All" we also need to:
 * Fire the UI event
 * Set the filters to active.

The interaction with `serviceName` and filters (should a filter be
toggled by the query parameter?) is not improved.

ref #1046 #1326


codefromthecrypt:
thank you so much!

PR #1434, opened 2016-12-14
show {trace,span}-id in span details pane

cburroughs:
Placed at the bottom since it is primarily intended as a debugging
convenience.

ref #1300

 * I'm not attached to the placement at the bottom, but it was convenient and I think it looks okay.
 * Passing traceId through the html to get it hack to the js feels a bit wonky, but I couldn't find a cleaner approach and it does happen to mirror the current json data format.

### DIFF
--- a/zipkin-ui/css/trace.css
+++ b/zipkin-ui/css/trace.css
@@ -194,6 +194,11 @@
   max-width: 75%;
 }
 
+#spanPanel #debugInfo td.value {
+  word-break: break-all;
+  max-width: 75%;
+}
+
 .modal-header .save {
   float: right;
   margin-right: 5px;
--- a/zipkin-ui/js/component_ui/spanPanel.js
+++ b/zipkin-ui/js/component_ui/spanPanel.js
@@ -5,6 +5,7 @@ import {Constants} from './traceConstants';
 export default component(function spanPanel() {
   this.$annotationTemplate = null;
   this.$binaryAnnotationTemplate = null;
+  this.$moreInfoTemplate = null;
 
   this.show = function(e, span) {
     const self = this;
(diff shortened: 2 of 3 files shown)


codefromthecrypt:
thanks for the help. screen shot?

cburroughs:
trace-id-screenshot

codefromthecrypt:
thx. I might right-orient it so that it distracts less from the tags.

any one have feelings against merging something like this?

yurishkuro:
@adriancole in our UI we make it collapsed by default, so as not to distract casual users.

image

cburroughs:
@yurishkuro What else do you include in Debug Info?

yurishkuro:
At the moment just the span ID ('cause trace ID is in the URL anyway). We also have the capability to show the JSON (pretty-printed) of that particular span, but didn't feel we need it yet (we have a separate JSON view for the full trace).

yurishkuro:
image

cburroughs:
I kind of like the collapsible thingy approach.  I can take a stab at that if there are no other UI comments.

cburroughs:
debug-info-collapse


debug-info

codefromthecrypt:
I don't like the label debug info as debug is an overloaded term, and the key here is trace identifiers (aka the trace context). Ex I might randomly figure out to click on that to find the trace identifiers, but why not make it explicit? ex "trace identifiers" or "trace IDs" PS still think it should be right justified since it is a secondary concern and less distracting.

cc @openzipkin/zipkin-ui @jcarres-mdsol etc anyone else have thoughts?

jcarres-mdsol:
Are you planning more info there in the future? If not I would just show the ids there , no collapsable thing, else the naming will depend on what is coming next IMHO

codefromthecrypt:
think I've finally an idea for what to do next..

we can either say "More info" instead of "Debug info", since the ids aren't necessarily debug in nature, or.. make this a debug toggle.

For example, occasionally we have overlap in tags (since client+server spans are the norm). The debug toggle could add the host information which could clarify these.

blob

just thoughts. @cburroughs @jcarres-mdsol wdyt?

cburroughs:
Here is a shot with right aligned 'More Info'.

more-info

If I understand it right, with v2/simplified formats, having data from multiple addresses on a single span is going to become even less common, and I'm wary of dynamically mucking with the number of table columns.

codefromthecrypt:
nice! thanks for the help

> If I understand it right, with v2/simplified formats, having data from
> multiple addresses on a single span is going to become even less common,
> and I'm wary of dynamically mucking with the number of table columns.
>
fair point. on conflict, there's always the json

Issue #1170, opened 2016-07-06
Zipkin-UI is not optimized at all. Delivers a bloated zipkin-ui.jar


brockovercash:
Working with zipkin, I've discovered that the webpack config for zipkin is crazy unoptimized vs what it could be. Even just with a simple --optimize-minimize --optimize-dedupe added to the npm run build script reduces file size to nearly 1/2th of the original size (1026KB vs 2.1MB). It should also be possible to chunk the dagre-d3 dependency to just load with the dependencies page so that it and it's dependencies don't bloat loading the trace view. I'm not a js/webpack expert or I would have coupled a PR with this, but I think this is definitely something that should be improved on.

PR #1436, opened 2016-12-15
apply basic size optimizations to js output

cburroughs:
See:
 * https://github.com/webpack/docs/wiki/optimization
 * http://webpack.github.io/docs/list-of-plugins.html#uglifyjsplugin

for background.  All default settings should be safe and
license headers preserved.  Only applied if `NODE_ENV=production` is
set (an apparently common webpack idiom) so the dev server workflow is
unaffected.

This reduces the size of the `min.js` file from about 2.2 MiB to 821
KiB.

ref #1170

### DIFF
--- a/zipkin-ui/package.json
+++ b/zipkin-ui/package.json
@@ -8,7 +8,7 @@
     "test": "node node_modules/karma/bin/karma start --single-run",
     "test-watch": "node node_modules/karma/bin/karma start",
     "lint": "node node_modules/eslint/bin/eslint js test",
-    "build": "node node_modules/webpack/bin/webpack.js --bail"
+    "build": "node node_modules/webpack/bin/webpack.js --bail --optimize-minimize --optimize-occurrence-order --optimize-dedupe"
   },
   "author": "",
   "license": "Apache-2.0",
--- a/zipkin-ui/webpack.config.js
+++ b/zipkin-ui/webpack.config.js
@@ -6,7 +6,7 @@ var CopyWebpackPlugin = require('copy-webpack-plugin');
 var proxyURL = process.env.proxy || "http://localhost:8080/";
 console.log("API requests are forwarded to " + proxyURL);
 
-module.exports = {
+var webpackConfig = {
     entry: [
         __dirname + '/js/main.js',
         __dirname + '/css/style-loader.js'
@@ -56,3 +56,5 @@ module.exports = {
         }
     }
 };
+
+module.exports = webpackConfig;


zipkin-ui/webpack.config.js
```diff
@@ -56,3 +56,11 @@ module.exports = {
         }
     }
 };
+
+if (process.env.NODE_ENV === 'production') {
```
codefromthecrypt: Maybe add your comment here that this reduces the resulting size to ~NNNKiB

codefromthecrypt (approved):
assuming you've tested this work :)

zipkin-ui/package.json
```diff
@@ -5,10 +5,10 @@
   "main": "index.js",
   "scripts": {
     "dev": "node node_modules/webpack-dev-server/bin/webpack-dev-server.js --hot --inline",
-    "test": "node node_modules/karma/bin/karma start --single-run",
+    "test": "NODE_ENV=production node node_modules/karma/bin/karma start --single-run",
     "test-watch": "node node_modules/karma/bin/karma start",
     "lint": "node node_modules/eslint/bin/eslint js test",
-    "build": "node node_modules/webpack/bin/webpack.js --bail"
+    "build": "NODE_ENV=production node node_modules/webpack/bin/webpack.js --bail"
```
eirslett: It's maybe easier to just add `--optimize-minimize --optimize-dedupe` here, to the CLI arguments, then we don't need to tweak the webpack config or set environment variables etc.

Passing environment variables to the script like this will also not work on Windows, so we'd have to use crossenv or something like that.

cburroughs:
If we ever did need to customize the behavior of these plugins I don't think we can do that from the cli args easily.  However, dealing with environment variables on Windows sounds like way less fun -- and the upstream defaults are sane -- so I think using the cli args is the best course.

zipkin-ui/package.json
```diff
@@ -5,10 +5,10 @@
   "main": "index.js",
   "scripts": {
     "dev": "node node_modules/webpack-dev-server/bin/webpack-dev-server.js --hot --inline",
-    "test": "node node_modules/karma/bin/karma start --single-run",
+    "test": "node node_modules/karma/bin/karma start --single-run --optimize-minimize --optimize-occurrence-order --optimize-dedupe",
```
eirslett: I don't think karma understands these flags... anyways for unit tests we prefer fast runs over minimized code. (After all, unit test code is not shipped to the user)

PR #1437, opened 2016-12-15
do not use serviceName when linking to trace results

cburroughs:
38275c75229bf1941af09d5afacda8fe9ef72073 "really" made "Expand All"
the default way to view a span.  However, in practice search results
are a popular way to view a span, and they used the `serviceName`
query parameter to confusingly show only a single service.

ref #1046

### DIFF
--- a/zipkin-ui/templates/index.mustache
+++ b/zipkin-ui/templates/index.mustache
@@ -87,7 +87,7 @@
   <ul id="traces">
   {{#traces}}
     <li class="trace {{infoClass}}" data-traceId="{{traceId}}" data-duration="{{duration}}" data-timestamp="{{timestamp}}" data-service-percentage="{{servicePercentage}}">
-      <a href="/traces/{{traceId}}?serviceName={{serviceName}}">
+      <a href="/traces/{{traceId}}">
         <div class="bar-block">
           <span class="bar-graphic" style="width:{{width}}%;"></span>
           <span class="bar-label">{{durationStr}}</span>


cburroughs:
Sorry about missing this.  It is an obvious test path in hindsight.

codefromthecrypt:
I think it was to focus on the service you searched for, right? Ex given 300 spans, you'd see the spans expanded that included your search term.

cburroughs:
My understanding of #1046 was that showing partial spans was consistently confusing to users, so just show them the whole span was hoped to be a better default.  For reference this is how it all looks back in 122bb0193ccae30353fd9683d33031836b025871

I can search for spans by service name but not with any name (that I can tell, I wish I could!)  

ex-0

If I click on a span I get taken to `traces/{{traceId}}?serviceName=foo` which highlights spans with that service name, but hides all other spans

ex-1

This is a different view from "Collapse All"

ex-2

"Expand All"

ex-3

And from navigating to the trace URL directly (no query parameter) and clicking on the service name

ex-4

I *think* the best default view would be with up to 100 (or whatever limit) spans expanded. The service you searched for would be *highlighted* but other services dimmed, not hidden nor collapsed.  But that's not an option that exists today and I agree with #1046 (or at least what I thought the sentiment there was) that "Expand All" is the least confusing of the current options.  Maybe this is a subtle different in use cases between "I am trying to understand the emergent properties of my whole system" and "as service owner X, I wan to know how my thing is used and can be made faster?

codefromthecrypt:
appreciate the explanation, and I agree with your rationale. Thanks for taking time to make it easy to understand!

Issue #1277, opened 2016-09-08
UI: Escape key should close dialogs


rogeralsing:
There is no way to close UI dialogs w/o mouse clicks.
Pressing Esc would be the natural thing to use.

Some dialogs even lack the close X button. and the only way to close is by clicking outside the dialog.

PR #1438, opened 2016-12-15
allow escape key to close modal dialogs

cburroughs:
An "x" button is also added to the dependency modals which
inconsistently lacked one.

On this "tabindex" thing: http://stackoverflow.com/a/32912224

ref #1277

### DIFF
--- a/zipkin-ui/templates/dependency.mustache
+++ b/zipkin-ui/templates/dependency.mustache
@@ -19,10 +19,11 @@
 </div>
 
 <div id="service-data-modal-container">
-  <div class="modal" id="serviceModal">
+  <div class="modal" id="serviceModal" tabindex="-1">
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
for those who don't speak modal.. can you mention what you can do now that you couldn't before? Ex. If you click on a span, or click the json button, you can now hit escape to return to the trace detail view

cburroughs:
Sure, so if you click on a span you can now hit `esc` to close the dialog with the span details:
esc-0
And the same is true of the help message...
esc-1

The two dependency related dialogs boxes can also be closed with `esc` or an `x` button.  (Before you had to click somewhere else to close them which was called out in #1277.)
esc-2
esc-3

codefromthecrypt:
lovely. thanks for the help!

Issue #1440, opened 2016-12-16
span names not updated on first time view


cburroughs:
On the first view of of the search page the zipkin UI loads the service names.  But does not load the span names for that service.

![snap-name-0](https://cloud.githubusercontent.com/assets/24433/21278683/8b7d6392-c3aa-11e6-8472-128ddfc30bf3.png)

However, we save `last-serviceName` as a cookie so this won't occur after a service name is selected, or if a service name query parameter is present.

![snap-name-1](https://cloud.githubusercontent.com/assets/24433/21278689/92b9706a-c3aa-11e6-91df-55227189c088.png)

Issue #1072, opened 2016-03-29
First (index) page is hitting /api/v1/spans?serviceName=undefined 


gena01:
When first loading the Zipkin UI (tested with zipkin-query 1.39 docker container) it seems that the Javascript is trying to load spans with `undefined` serviceName.

codefromthecrypt:
@eirslett mind helping on this one? I'm not quite sure the best way to early exit on this..

eirslett:
It's not really a big problem. But if somebody wants to fix it, we could first get the list of services, then pick the first service in the list and ask for spans for that service, instead of reading service name from the query string.

gena01:
Actually i thought it was even simpler than that. Maybe I should have rephrased things a little better.

When loading zipkin ui homepage (/ url) it shouldn't be loading any spans since nothing has been selected. This is also where the `undefined` comes from for the serviceName. Is it possible to make this a conditional of sorts just for the homepage case?

eirslett:
Yes, the event that triggers the API call should only be triggered when service name is set in the query string. We use query-string for that already, in other places.

PR #1441, opened 2016-12-16
cleanup first load events for service/span names

cburroughs:
During initialization an unconditional `uiChangeServiceName` is fired
to populate the lists of service and corresponding span names.  This
had two problematic consequence:
 * Since there was no service name yet, an http request for spans in
   the `undefined` service was made (looking "bad" in dev tools).
 * On first load we never went back and loaded the span names if no
   service was selected (via query parameters or cookies).  This meant
   the list of span names would just be "all".

These can both be resolved by:
 * Guarding against span search queries when no service name is set.
 * Loading the spans names for the first service in the list when no
   service name is set.

ref #1440 #1072

### DIFF
--- a/zipkin-ui/js/component_data/spanNames.js
+++ b/zipkin-ui/js/component_data/spanNames.js
@@ -4,6 +4,10 @@ import $ from 'jquery';
 
 export default component(function spanNames() {
   this.updateSpanNames = function(ev, serviceName) {
+    if (!serviceName) {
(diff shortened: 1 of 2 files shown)


cburroughs:
On first load:
service-names

jcarres-mdsol:
Awesome! This has always bothered me!

codefromthecrypt:
fantastic. I'll cut another release today sometime

PR #1443, opened 2016-12-20
Update Span merge

naoman:
In the current implementation, when we merge spans from client and server, we compute timestamp and duration for the Span by looking at the min and max timestamps. This breaks if we have a clock skew. Once timestamp and duration is set here, we don't recompute it in [ApplyTimestampAndDuration](https://github.com/openzipkin/zipkin/blob/1d0e657e6f89b929a0be27db843a848611242b92/zipkin/src/main/java/zipkin/internal/ApplyTimestampAndDuration.java#L34). 

In this change all we are doing is set timestamp and duration to null, when we merge two spans and and both have direct timestamp set.

### DIFF
--- a/zipkin/src/main/java/zipkin/Span.java
+++ b/zipkin/src/main/java/zipkin/Span.java
@@ -21,7 +21,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
-import zipkin.internal.ApplyTimestampAndDuration;
 import zipkin.internal.Nullable;
 import zipkin.storage.StorageComponent;
 
@@ -245,16 +244,6 @@ public final class Span implements Comparable<Span>, Serializable {
         this.parentId = that.parentId;
       }
 
-      for (Annotation a : that.annotations) {
-        addAnnotation(a);
-      }
-      for (BinaryAnnotation a : that.binaryAnnotations) {
-        addBinaryAnnotation(a);
-      }
-      if (this.debug == null) {
-        this.debug = that.debug;
-      }
-
       // Single timestamp makes duration easy: just choose max
       if (this.timestamp == null || that.timestamp == null || this.timestamp.equals(
           that.timestamp)) {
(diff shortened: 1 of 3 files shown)


zipkin/src/test/java/zipkin/storage/SpanStoreTest.java
```diff
@@ -806,6 +806,43 @@ public void clientTimestampAndDurationWinInSharedSpan() {
     }
   }
 
+  /**
```
codefromthecrypt: I'd move this test to SpanTest as it doesn't require live servers to verify

zipkin/src/test/java/zipkin/storage/SpanStoreTest.java
```diff
@@ -789,7 +789,7 @@ public void clientTimestampAndDurationWinInSharedSpan() {
     // neither client, nor server set span.timestamp, duration
     Span clientViewDerived = Span.builder().traceId(1).name("derived").id(666)
         .addAnnotation(Annotation.create(clientTimestamp, CLIENT_SEND, client))
-        .addAnnotation(Annotation.create(clientTimestamp + clientDuration, CLIENT_SEND, client))
```
codefromthecrypt, line 792: keep this fix though!

zipkin/src/main/java/zipkin/Span.java
```diff
@@ -253,11 +253,11 @@ public Builder merge(Span that) {
         } else if (that.duration != null) {
           this.duration = Math.max(this.duration, that.duration);
         }
-      } else { // duration might need to be recalculated, since we have 2 different timestamps
-        long thisEndTs = this.duration != null ? this.timestamp + this.duration : this.timestamp;
-        long thatEndTs = that.duration != null ? that.timestamp + that.duration : that.timestamp;
-        this.timestamp = Math.min(this.timestamp, that.timestamp);
-        this.duration = Math.max(thisEndTs, thatEndTs) - this.timestamp;
+      } else {
+        // We have 2 different timestamps and we don't know which one is authoritative.
```
codefromthecrypt: we actually do. the client is authoritative in a shared span. Let's set it to the client one (span with "cr" annotation), or null if cr is absent

codefromthecrypt: so a suggested fix (which considers client span) would be to move this block below where we iterate through annotations. When we iterate through annotations decide which side is the client span and prefer that somehow..

codefromthecrypt:
thanks for the help!

naoman:
Thanks for reviewing the code @adriancole. I've made following changes:
1. Moved the test code from SpanStoreTest to SpanTest
2. Updated Span.Builder.merge() code to return timestamp and duration of client after merge. However, instead of computing it by going through annotations, I'm calling ApplyTimestampAndDuration.apply(). Since the code is so similar, I didn't want to copy paste. And later if we want to change the logic of computing ApplyTimestampAndDuration, there will be just one place to make the change. Let me know what you think.

zipkin/src/main/java/zipkin/Span.java
```diff
@@ -253,22 +264,13 @@ public Builder merge(Span that) {
...
-      } else { // duration might need to be recalculated, since we have 2 different timestamps
-        long thisEndTs = this.duration != null ? this.timestamp + this.duration : this.timestamp;
-        long thatEndTs = that.duration != null ? that.timestamp + that.duration : that.timestamp;
-        this.timestamp = Math.min(this.timestamp, that.timestamp);
-        this.duration = Math.max(thisEndTs, thatEndTs) - this.timestamp;
+      } else {
+        // We have 2 different timestamps. Use ApplyTimestampAndDuration to compute final timestamp.
+        this.timestamp = null;
+        this.duration = null;
+        return ApplyTimestampAndDuration.apply(this.build()).toBuilder();
```
codefromthecrypt: we can't do this as we only include authoritative data reported to us (as opposed to deriving span's timestamp from annotations). The prior data incorrectly tried to merge authoritatively set timestamps (which was the bug). The better way is to prefer the client as mentioned before.

codefromthecrypt:
Thanks for the help on this. I appreciate you've spent some time on this, so if you don't have time to budget to loop through the annotations once, I can do it as a part of merging. lemme know!

naoman:
It would be great if you can do it as part of the merge. I couldn't think of a cleaner way of doing this, since we were only looping through "that" annotations before the change.
Thanks!

codefromthecrypt:
@naoman thanks for the help. I added a little temporary debt because what you are trying to sort out will be straightforward in the next model. https://github.com/openzipkin/zipkin/commit/80e99c482e06da7fa0fdc8cbf1055d07bb690853

Issue #1143, opened 2016-06-24
Should we add the "json button" to the main search view?


codefromthecrypt:
Sometimes folks report issues about inconsistency of list responses, but delay in responding with json that would help us validate or fix the issue.

For example, #1142 claims limit isn't working right in cassandra eventhough it seems to in mysql. It might well be a problem, but attempting to guess is very inefficient timewise and rarely effective vs json or a test case.

Should we place a JSON button somewhere subtle in the search pane so that troubleshooting doesn't require issue submitters to hand enter api paths?

My assumption is that if we placed a button there, it might reduce the overhead of solving an issue. Typically, I can validate an issue in a minute or two given json. Back and forth without any data often displaces an hour or more. We need to be able to help support users without displacing so much time.

eirslett:
That's fine with me!

PR #1448, opened 2016-12-20
add a button to view the raw json of a search query

cburroughs:
Easy access to the raw json can aid in debugging.  Uses the same panel
as the per trace json display.

ref #1143

### DIFF
--- a/zipkin-ui/js/component_data/default.js
+++ b/zipkin-ui/js/component_data/default.js
@@ -21,12 +21,15 @@ export default component(function DefaultData() {
     const query = convertToApiQuery(window.location.search);
     const serviceName = query.serviceName;
     if (serviceName) {
-      $.ajax(`/api/v1/traces?${queryString.stringify(query)}`, {
+      const apiURL = `/api/v1/traces?${queryString.stringify(query)}`;
+      $.ajax(apiURL, {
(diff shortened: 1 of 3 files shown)


cburroughs:
q-json-0
q-json-1

codefromthecrypt:
oh interesting.. this is the raw json for all the results. the single span in the result list threw me off :)

I can see this as valuable, but wonder if for first timers if this isn't distracting? maybe others can comment cc @openzipkin/core

cburroughs:
> oh interesting.. this is the raw json for all the results.

Wasn't that your proposal in #1143?

codefromthecrypt:
lol wow you are actually going through the backlog! thanks. sorry I forgot
about my own issue :P

codefromthecrypt:
so TBH I don't have a better suggestion for where/how to present the json button. with that context in mind, I think it will help with debugging as currently we ask people to use curl.

will merge tomorrow unless feedback suggests against. Thanks again!

nit: it isn't really "raw" since the list api doesn't have that option. in the get trace api, raw indicates no adjustments to the data. http://zipkin.io/zipkin-api/#/paths/%252Ftrace%252F%7BtraceId%7D/get/parameters/raw

PR #1450, opened 2016-12-21
Adds ES_MAX_REQUESTS to control maximum in-flight Elasticsearch requests

### LINKED ISSUE #1449: can't collect and query at the same time
I use the **elasticsearch** to store and **kafka** to collect
while the zipkin is collecting kafka msgs, and it did realy store in the cluster of elasticsearch, and search something is also quickly with elasticsearch's http-query-api, but it can't be query with zipkin UI or API at the same time.
once i stop collecting kafka msgs, zipkin query can work.
i think is it because of the collecting too busy cause zipkin can't handle the query request?
because when i run on my desktop, with all the same config, but just a little data to put in it,and it work normal, that can collect and query at the same time 

**here is my config**:
`
STORAGE_TYPE=elasticsearch,ES_HOSTS="http://host1:9200,http://host2:9200,http://host3:9200,http://host4:9200",KAFKA_STREAMS=6,KAFKA_ZOOKEEPER="host1:port,host2:port,host3,host4:port:port/kafkacluster",KAFKA_TOPIC="halo.msg.trace",KAFKA_MAX_MESSAGE_SIZE="200000000",KAFKA_GROUP_ID="new-zipkin",SELF_TRACING_ENABLED="false"
`

codefromthecrypt:
The http client used for Elasticsearch is exclusively for Elasticsearch.
However, it was using defaults for web sites, which constrained it to 5
in-flight requests per host. This raises the default to the same as max
requests, which is 64. It also opens `ES_MAX_REQUESTS` to adjust that
count.

The intended behavior is that there will be less contension in a server
that is both storing spans and serving api requests. It should also
allow higher throughput to the Elasticsearch cluster.

Fixes #1449

### DIFF
--- a/zipkin-autoconfigure/storage-elasticsearch-http/src/main/java/zipkin/autoconfigure/storage/elasticsearch/http/ZipkinElasticsearchHttpStorageAutoConfiguration.java
+++ b/zipkin-autoconfigure/storage-elasticsearch-http/src/main/java/zipkin/autoconfigure/storage/elasticsearch/http/ZipkinElasticsearchHttpStorageAutoConfiguration.java
@@ -36,8 +36,11 @@ public class ZipkinElasticsearchHttpStorageAutoConfiguration {
   @ConditionalOnMissingBean
   InternalElasticsearchClient.Builder clientBuilder(
       @Qualifier("zipkinElasticsearchHttp") OkHttpClient client,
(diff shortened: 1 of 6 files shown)


codefromthecrypt:
merging and will revert if it doesn't solve the issue

PR #1451, opened 2016-12-22
Small fix for ArrayIndexOutOfBoundsException in Buffer class

naoman:
Looks like there is typo, and a char is used instead of an int. This throws ArrayIndexOutOfBoundsException if binary annotation length is greater than 65,535 (range to char)

### DIFF
--- a/zipkin/src/main/java/zipkin/internal/Buffer.java
+++ b/zipkin/src/main/java/zipkin/internal/Buffer.java
@@ -103,7 +103,7 @@ final class Buffer {
 
   Buffer writeAscii(String v) {
     int length = v.length();
-    for (char i = 0; i < length; i++) {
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
eek :) thanks. will merge

aside: that's a pretty big annotation! usually spans are all-in <1KiB. systems aren't designed for very large annotation even though there are some safe guards to help stop huge annotations from blowing up indexes. Curious.. what sort of data ended up as 64KiB?

http://zipkin.io/pages/instrumenting.html

naoman:
We have added checks in our pipeline, since we ran into this issue, to make sure we keep the annotation size in control. The data that triggered this was a value from a key value store, that was exceptionally large.

codefromthecrypt:
@naoman cool. was hoping it was an exception!

PR #1458, opened 2016-12-28
Beautify JSON in case of binary annotation value

jakubhava:
In case JSON is passed as value of binary annotation, the JSON is displayed as one liner. This small fix will ensure that JSON is displayed with correct spacing and indentation level of 4 spaces.

The use-case for this is for example passing current thread's stack trace as JSON value in binary annotation where each stack trace element is element of JSON array.

btw: Thanks guys for great work!

### DIFF
--- a/zipkin-ui/js/component_ui/spanPanel.js
+++ b/zipkin-ui/js/component_ui/spanPanel.js
@@ -47,7 +47,12 @@ export default component(function spanPanel() {
         const maybeObject = anno[$this.data('key')];
         // In case someone is storing escaped json as binary annotation values
         // TODO: this class is not testable at the moment
-        $this.text($.type(maybeObject) === 'object' ? JSON.stringify(maybeObject) : maybeObject);
+        const type = $.type(maybeObject);
+        if (type === 'object' || type === 'array') {
+          $this.append(`<pre>${JSON.stringify(maybeObject, null, 2)}</pre>`);
+        } else {
+          $this.text(maybeObject);
+        }
       });
       $binAnnoBody.append($row);
     });


codefromthecrypt:
mind making a before and after screen shot? I suspect 4 spaces is ok vs 2... @eirslett any strong opinion?

jakubhava:
Sure! Will share screenshots with both 4 and 2 spaces so we can visually compare

jakubhava:
Oki, it took me more time than I expected, but works as it should at this moment.
<h3>Array 2 spaces</h3>
array_2_spaces
<h3>Object 2 spaces</h3>
object_2_spaces
<h3>Array 4 spaces</h3>
array_4_spaces
<h3>Object 4 spaces</h3>
object_4_spaces

2 spaces might actually be sufficient. Screenshots before the change will be here in a bit

jakubhava:
Screens before the change ( on current master ) 
<b>Object is displayed as json, but no indentation</b>
screen shot 2016-12-28 at 14 01 20
<b>Json array is treated as normal string</b>
screen shot 2016-12-28 at 14 02 06

codefromthecrypt:
looks better! do you mind having a go at a unit test?

on spacing, I don't care as they both look fine.

jakubhava:
Sure 👍 , I just need to make myself a bit familiar with tests structure

jakubhava:
So I've been trying to code the tests, but it seems like I'll pass on it, at least for now because:

My time to work on this is limited and I couldn't find any test doing anything similar - ie. simulate receiving JSON at zipkin-ui, look at the created span at html level and the do bunch of asserts. All tests test some basic span attributes, but I couldn't find any tests which are actually testing if `spanPanel` is showing correct data. It seems more like need for integration test ? This is definitely doable, but it's more time consuming then I initially thought.

Maybe if you can point me to some basics how to test this scenario that would be great and I could give it a second try!

zipkin-ui/js/component_ui/spanPanel.js
```diff
@@ -25,7 +25,12 @@ export default component(function spanPanel() {
         const maybeObject = anno[$this.data('key')];
         // In case someone is storing escaped json as an annotation value
         // TODO: this class is not testable at the moment
-        $this.text($.type(maybeObject) === 'object' ? JSON.stringify(maybeObject) : maybeObject);
+        const type = $.type(maybeObject);
```
codefromthecrypt: ps this part is for annotations (not binary annotion/tags). This change would make the UI timeline look weird, as single-line is ideal. I would revert this part.

codefromthecrypt:
ok I'd remove the first part (which looks the same, but is not), and then I'll merge. Scout's honor I will create a test class afterwords sometime and ping you for review. cheers!

zipkin-ui/js/component_ui/spanPanel.js
jakubhava: oki, reverting this part

jakubhava:
Awesome, thanks! I applied the change only to binary annotation and squashed all the commits into one for nicer git history.

Thanks for support with the test, I'll be happy to review or help with it.

codefromthecrypt:
thanks @jakubhava!

PR #1459, opened 2016-12-28
Don't spam logs when RPC spans are in use.

fedj:
As mentionned in openzipkin/openzipkin.github.io#49, timestamp should be set on span originators.
In the case of a span sent from a mid-tier server (e.g. receiving a request from a instrumented service), it is not supposed to set the timestamp, leading to a log message when using cassandra as storage.
This log can be only wrote when the span is an originator. It is still trying to guess the correct timestamp to cope with non-compliant tracers.

### DIFF
--- a/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
@@ -31,7 +31,9 @@ import java.nio.ByteBuffer;
 import java.util.List;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import zipkin.Annotation;
 import zipkin.Codec;
+import zipkin.Constants;
 import zipkin.Span;
 import zipkin.internal.Nullable;
 import zipkin.internal.Pair;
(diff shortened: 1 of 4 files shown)


zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
```diff
@@ -136,12 +141,24 @@ private RegularStatement maybeUseTtl(Insert value) {
...
+    return false;
+  }
+
   /**
    * Store the span in the underlying storage for later retrieval.
    */
-  ListenableFuture<?> storeSpan(long traceId, long timestamp, String key, ByteBuffer span) {
+  ListenableFuture<?> storeSpan(long traceId, long timestamp, boolean isServerRecvSpan, String key, ByteBuffer span) {
     try {
-      if (0 == timestamp && metadata.compactionClass.contains("DateTieredCompactionStrategy")) {
```
codefromthecrypt, line 144: nit: instead of isServerRecvSpan as a parameter, I'd just invoke the method here and save as a local

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
```diff
@@ -136,12 +141,24 @@ private RegularStatement maybeUseTtl(Insert value) {
     return transform(Futures.allAsList(futures.build()), TO_VOID);
   }
 
+  private static boolean isServerRecvSpan(Span span) {
```
codefromthecrypt, line 144: nit move this below storeSpan if you choose to not pass as a param

codefromthecrypt (approved):
lovely. only a couple nits

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
fedj: It would need to pass the span though and from what I see in the code, this is not what was intended (we extract everything needed from the span before)

codefromthecrypt: works for me :P

PR #1461, opened 2016-12-29
Annotation query can be error prone.

fedj:
With cassandra storage, results from cassandra are intersected with unfiltered traces. The result is that the number of result may vary depending on the limit but would likely to answer no result by default.


codefromthecrypt:
cc @openzipkin/cassandra funny we didn't notice this before! (ps I've verified this is a problem and the fix)

codefromthecrypt (approved):

llinder:
good catch! 👍

Issue #1462, opened 2016-12-30
Elasticsearch http slow on getting span names when there's a lot of data


codefromthecrypt:
from @dragontree101

because of we have lots of data, service name  and span name all time out,  i remember store data in cassandra, have table only store service_name and span_name, but store in es, service_name and span_name get by es, so time is very long. 

![image](https://cloud.githubusercontent.com/assets/1763935/21561603/a7a8d63c-cea9-11e6-93b0-0ef800b1f7df.png)

and chrome console have

![image](https://cloud.githubusercontent.com/assets/1763935/21561714/5254d35a-ceab-11e6-8fa2-caf5158fc269.png)

i found i set SELF_TRACING_ENABLED="true" in environment

![image](https://cloud.githubusercontent.com/assets/1763935/21562108/a01478c0-ceb0-11e6-83fb-b25d98af937d.png)

and yestoday's es index is 
```
health status index                           uuid                   pri rep docs.count docs.deleted store.size pri.store.size
green  open   zipkin-2016-12-29               2_uXTThATbyoJ7khnXbo4g   5   1  609199726            0      140gb           70gb
```

codefromthecrypt:
Currently, the query for service names looks at all indexes (it also sorts which is redundant). So, it probably affects more than just today's data.

codefromthecrypt:
@openzipkin/elasticsearch I'm suspicious of using catch-all for service and span names. On one hand we want to make sure data is readable, but on the other hand do you think we need to search all indexes for service and span names? what if we cut-off at a particular date?

mansu:
@adriancole While cutting off after a certain date is a good idea I think a better alternative is to add a mechanism to cache the service and span names in the persistent layer. That way we can run an expensive query periodically and cache accurate data instead of returning incomplete data every time.  The current model of relying on HTTP caching is too expensive when a lot of users use the zipkin UI and is not scalable(for us) in the long run. Loading span and service names is the slowest part in our UI and currently it takes 10-15 seconds before the UI is usable. We are currently sampling 1/3 the expected rate but once we sample at the expected rate, the UI will take even longer to load.

codefromthecrypt:
elasticsearch caching is enabled by default by our schema. Have you looked
into this and perhaps what it would take to better tune the service and
span name cache (or at least see if it is or isn't working)?

https://www.elastic.co/guide/en/elasticsearch/reference/current/shard-request-cache.html

Ideally, we can find a solution that works well at ES level since not
[...]

PR #1463, opened 2016-12-30
Don't sort span names in elasticsearch

codefromthecrypt:
While probably not a big contributor to performance, we don't need to
sort span names in elasticsearch as we already copy out into a sorted
list.

See #1462


codefromthecrypt:
fyi @openzipkin/elasticsearch probably nothing

dragontree101:
could you give me an es query, i want to compare have sort query with no sort query use time. thanks

codefromthecrypt:
you'd have to use charles proxy or something to capture the post traffic,
or try adjusting your server logs as I can't dig into this at the moment.
https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules-slowlog.html

Regardless of performance, we don't need to ask server to do this.

PR #1464, opened 2016-12-30
ICM-634. Added prefix "zipkin" for all request in UI.

knikitin:
[ICM-634. Service requests tracing.](https://intapp.atlassian.net/browse/ICM-634)

- Added prefix "zipkin".

### DIFF
--- a/zipkin-ui/js/component_data/default.js
+++ b/zipkin-ui/js/component_data/default.js
@@ -21,7 +21,7 @@ export default component(function DefaultData() {
     const query = convertToApiQuery(window.location.search);
     const serviceName = query.serviceName;
     if (serviceName) {
-      const apiURL = `/api/v1/traces?${queryString.stringify(query)}`;
+      const apiURL = `/zipkin/api/v1/traces?${queryString.stringify(query)}`;
       $.ajax(apiURL, {
         type: 'GET',
         dataType: 'json'
--- a/zipkin-ui/js/component_data/dependency.js
+++ b/zipkin-ui/js/component_data/dependency.js
@@ -7,7 +7,7 @@ export default component(function dependency() {
   let dependencies = {};
 
   this.getDependency = function(endTs, lookback) {
-    let url = `/api/v1/dependencies?endTs=${endTs}`;
+    let url = `/zipkin/api/v1/dependencies?endTs=${endTs}`;
     if (lookback) {
       url += `&lookback=${lookback}`;
     }
--- a/zipkin-ui/js/component_data/serviceNames.js
+++ b/zipkin-ui/js/component_data/serviceNames.js
@@ -4,7 +4,7 @@ import $ from 'jquery';
 
 export default component(function serviceNames() {
   this.updateServiceNames = function(ev, lastServiceName) {
-    $.ajax('/api/v1/services', {
+    $.ajax('/zipkin/api/v1/services', {
       type: 'GET',
       dataType: 'json'
     }).done(names => {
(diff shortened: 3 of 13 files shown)


zipkin-ui/js/component_data/default.js
```diff
@@ -21,7 +21,7 @@ export default component(function DefaultData() {
     const query = convertToApiQuery(window.location.search);
     const serviceName = query.serviceName;
     if (serviceName) {
-      const apiURL = `/api/v1/traces?${queryString.stringify(query)}`;
+      const apiURL = `/zipkin/api/v1/traces?${queryString.stringify(query)}`;
```
codefromthecrypt, line 24: is this only affecting crossroads? I don't expect us to mount the api under the subpath /zipkin at the server level by default.

codefromthecrypt: well at least not without significantly broadening the discussion beyond those interested in this UI (as it has deployment and other doc implications)

codefromthecrypt (commented):
Thanks for the help!

:) good sign that there's no unit tests is that this passed without changing them. This means we have a bit more burden to figure out if this will work or not.

Also worth raising an issue or otherwise trying here as it is expected that this will finish at some point https://github.com/openzipkin/zipkin-ui

Finally, if you don't mind, there have been several issues here to link to on the same topic. mind pinging them on your work?

zipkin-ui/js/component_data/default.js
eirslett: This would affect the API as well.

codefromthecrypt:
is it possible to read this prefix from config.json (and parameterize all
other paths based on that value)? It would be nice to not have to play
games for the base case where the url space is root.

eirslett:
> is it possible to read this prefix from config.json

There's a chicken-and-egg problem there, as the app doesn't know the location of the config.json file (today we assume it's located at root)

We could try to heuristically detect whether we're mounted under root or not:

```javascript
// inside the web UI code:

// check if pathname starts with /zipkin/, and if it does,
// assume that we're mounted under a /zipkin/ context path
const isMountedAtRoot = window.location.pathname.indexOf('/zipkin/') !== 0;
const basePath = isMountedAtRoot ? '/' : '/zipkin/';
```

That is assuming we don't have any request paths in our application that start with /zipkin/, but I don't think we have.

We would then support two use cases: 1) mounting zipkin on root, as we do today, and 2) mounting zipkin under /zipkin/.
The case we would not support is 3), to mount zipkin under an arbitrary context path, but that would be an edge case. I don't think we should take on the burden of supporting it. (Why would somebody want to mount an app under /zipkin/ that is not zipkin?)

The context path mapping itself can either be done on the zipkin server side if we choose to implement it (MOUNT_CONTEXT_PATH=true) or in a proxy setup (apache 2 or nginx) running in front of the server. In the latter case, this issue can be solved purely on the UI side, with no server side changes necessary. People who mount apps under a context path, with multiple apps on the same domain, are already using a frontend proxy anyways.

codefromthecrypt:
I like this idea

zipkin-ui/js/component_data/default.js
knikitin: Sorry. I made a mistake with the creation of the pull request. I made a mistake with the choice of destination.  
If you do not mind, I will close this pull request.

tschiman:
Sorry to drag this thread back to life, but did this get added to the project? 

I see many threads in the issues about this topic, and I am another user that would like to see this feature added. More specifically the one where I can set the prefix myself. If this isn't the most active thread on this topic can you point me to it?

PR #1465, opened 2017-01-02
Clock skew correction is now applied.

fedj:
When traces begin by SR-SS, clock skew correction wasn't applied.

### DIFF
--- a/zipkin/src/main/java/zipkin/internal/CorrectForClockSkew.java
+++ b/zipkin/src/main/java/zipkin/internal/CorrectForClockSkew.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2016 The OpenZipkin Authors
+ * Copyright 2015-2017 The OpenZipkin Authors
  *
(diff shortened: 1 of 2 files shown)


zipkin/src/main/java/zipkin/internal/CorrectForClockSkew.java
```diff
@@ -70,11 +70,10 @@ static void adjust(Node<Span> node, @Nullable ClockSkew skewFromParent) {
...
       // the current span's skew may be a different endpoint than skewFromParent, adjust again.
       node.value(adjustTimestamps(node.value(), skew));
-
-      // propagate skew to any children
-      for (Node<Span> child : node.children()) {
-        adjust(child, skew);
-      }
+    }
+    // propagate skew to any children
+    for (Node<Span> child : node.children()) {
```
codefromthecrypt, line 75: wow simple fix turns out. bet it wasn't easy to find.

codefromthecrypt (approved):
thanks tons for this!

PR #1468, opened 2017-01-03
Span timestamp is not correctly correct in case of clock skew.

fedj:
The timestamp of the span is now set to the value of the timestamp
of the CS annotation.

### DIFF
--- a/zipkin/src/test/java/zipkin/internal/CorrectForClockSkewTest.java
+++ b/zipkin/src/test/java/zipkin/internal/CorrectForClockSkewTest.java
@@ -21,9 +21,8 @@ import zipkin.Span;
 
 import java.util.Arrays;
 import java.util.List;
-import java.util.Map;
 
-import static org.junit.Assert.assertEquals;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import static zipkin.Constants.CLIENT_RECV;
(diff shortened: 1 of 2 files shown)


zipkin/src/test/java/zipkin/internal/CorrectForClockSkewTest.java
```diff
@@ -91,38 +90,73 @@ public void ipsMatch_whenIpv4Match() {
...
+  }
+
+  private static long addMsToMicroseconds(long microseconds, long ms) {
+    return microseconds + TimeUnit.MICROSECONDS.convert(ms, TimeUnit.MILLISECONDS);
   }
 
-  private long getSpanDuration(Span span, String beginAnnotation, String endAnnotation) {
-    Map<String, Annotation> annotationsAsMap = asMap(span.annotations);
-    return getTimestamp(annotationsAsMap, endAnnotation) - getTimestamp(annotationsAsMap, beginAnnotation);
+  private static void assertSpanTimestampEqualdsClientSendTimestamp(Span span) {
```
codefromthecrypt: spelling Equalds

zipkin/src/test/java/zipkin/internal/CorrectForClockSkewTest.java
```diff
@@ -91,38 +90,73 @@ public void ipsMatch_whenIpv4Match() {
...
+    assertClockSkewIsCorrectlyApplied(50000);
+  }
+
+  @Test
+  public void clockSkewIsCorrectedIfRpcSpanSendsBeforeClientSend() {
+    assertClockSkewIsCorrectlyApplied(-50000);
+  }
+
+  private static void assertClockSkewIsCorrectlyApplied(long skew) {
+    long now = System.currentTimeMillis() * 1000;
```
codefromthecrypt: since all units are microseconds. I'd take out things that make us have to think back and forth, especially in this test as it doesn't actually use a clock.

For example, the skew doesn't need to be in a unit different than microseconds. If everything is the same unit, we don't need to multiply by 1000 or increase the navigation needed in this test.

codefromthecrypt (commented):
good work. I like the robust test (though a little hard to read). If there's any polishing you feel up to, please do. I'll merge whatever you end up with in the morning.

fedj:
I didn't have time to finish it. Can you please wait until I polish it to merge it ?

codefromthecrypt:
sure thing

codefromthecrypt:
looks good. I think this is more straightforward. thanks!

PR #1469, opened 2017-01-04
Fix NullPointerException when Span.Builder merge other span without i…

kojilin:
…nitializing traceIdHigh


codefromthecrypt:
thanks for the help @kojilin

PR #1470, opened 2017-01-04
Clock skew correction is not propagated to enclosed local spans

fedj:
Since local spans share the same clock skew as the server they are running on, we need to propagate to children.

### DIFF
--- a/zipkin/src/main/java/zipkin/internal/CorrectForClockSkew.java
+++ b/zipkin/src/main/java/zipkin/internal/CorrectForClockSkew.java
@@ -70,11 +70,39 @@ public final class CorrectForClockSkew {
     if (skew != null) {
       // the current span's skew may be a different endpoint than skewFromParent, adjust again.
       node.value(adjustTimestamps(node.value(), skew));
+    } else {
+      if (skewFromParent != null && isLocalSpan(node.value())) {
+        //Propagate skewFromParent to local spans
+         skew = skewFromParent;
+      }
     }
     // propagate skew to any children
     for (Node<Span> child : node.children()) {
       adjust(child, skew);
     }
+  }
+
+  static boolean isLocalSpan(Span span) {
+    Endpoint endPoint = null;
+    for (int i = 0, length = span.annotations.size(); i < length; i++) {
+      Annotation annotation = span.annotations.get(i);
+      if (endPoint == null) {
+        endPoint = annotation.endpoint;
+      }
+      if (endPoint != null && !endPoint.equals(annotation.endpoint)) {
+        return false;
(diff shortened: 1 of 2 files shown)


zipkin/src/main/java/zipkin/internal/CorrectForClockSkew.java
```diff
@@ -70,6 +70,18 @@ static void adjust(Node<Span> node, @Nullable ClockSkew skewFromParent) {
     if (skew != null) {
       // the current span's skew may be a different endpoint than skewFromParent, adjust again.
       node.value(adjustTimestamps(node.value(), skew));
+    } else {
+      if (skewFromParent != null) {
+        //Propagate skewFromParent to local spans
+        for (int i = 0, length = node.value().binaryAnnotations.size(); i < length; i++) {
+          BinaryAnnotation b = node.value().binaryAnnotations.get(i);
+          if (b.endpoint == null) continue;
+          if (b.key.equals(Constants.LOCAL_COMPONENT) && ipsMatch(skewFromParent.endpoint, b.endpoint)) {
```
codefromthecrypt: just an fyi, but this could result in further maintenance as local spans are not required to have the tag "lc". The local component tag is primarily a tool to use when you don't have any other annotation or binary annotations (as it allows you to have some endpoint attached that you can search with.

So, while this will work for many single-host spans, it won't for all of them. It might be best to change the heuristic or note in a TODO that this won't work for single-host spans that don't contain "lc"

fedj: Done

zipkin/src/main/java/zipkin/internal/CorrectForClockSkew.java
```diff
@@ -70,13 +70,43 @@ static void adjust(Node<Span> node, @Nullable ClockSkew skewFromParent) {
...
       adjust(child, skew);
     }
   }
 
+  static boolean isLocalSpan(Span span) {
+    Endpoint endPoint = null;
+    for (int i = 0, length = span.annotations.size(); i < length; i++) {
+      Annotation annotation = span.annotations.get(i);
+      if (endPoint == null)
+      {
```
codefromthecrypt: nit formatting here and below

codefromthecrypt (approved):
LG will merge in the morning.

nice work

zipkin/src/main/java/zipkin/internal/CorrectForClockSkew.java
fedj: Good catch, done

PR #1474, opened 2017-01-07
Clarifies why we format data in spanPanel.js

codefromthecrypt:
I found it confusing that we format data both in traceToMustache and
spanPanel. This hopefully explains a bit better, and also cleans up
some javascript along the way.


codefromthecrypt:
cc @jakubhava @cburroughs @eirslett

PR #1475, opened 2017-01-07
Flags yellow for annotations that match case insensitive error

### LINKED ISSUE #1472: UI: flag yellow for annotations that match case insensitive error
Currently, we flag yellow a span that includes annotations that exact match error. In finagle and spring cloud sleuth, error is often a part of a format string, and might not be lowercase. I suggest we more leniently accept annotation highlighting to case insensitive contains.

@HancaiLi requested this in spring-cloud-sleuth
See https://github.com/openzipkin/openzipkin.github.io/issues/52 for more complete description

Note, it is still best to strictly treat the tag (binary annotation) key "error" as that is used in search.
https://github.com/openzipkin/zipkin-api/blob/master/thrift/zipkinCore.thrift#L218

codefromthecrypt:
Particularly, this flags existing finagle formatted errors, like so:
"Server Send Error: TimeoutException: socket timed out"

Fixes #1472
See https://github.com/openzipkin/openzipkin.github.io/issues/52

### DIFF
--- a/zipkin-ui/js/component_ui/spanPanel.js
+++ b/zipkin-ui/js/component_ui/spanPanel.js
@@ -2,6 +2,14 @@ import {component} from 'flightjs';
 import $ from 'jquery';
 import {Constants} from './traceConstants';
 
+// Annotation values that contain the word "error" hint of a transient error.
+// This adds a class when that's the case.
+export function maybeMarkTransientError(row, anno) {
+  if (/error/i.test(anno.value)) {
+    row.addClass('anno-error-transient');
+  }
+}
+
 // annotations are named events which shouldn't hold json. If someone passed
 // json, format as a single line. That way the rows corresponding to timestamps
 // aren't disrupted.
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
screen shot 2017-01-07 at 12 30 44 pm


Using this data:
```bash
curl -H "Content-Type: application/json" -s localhost:9411/api/v1/spans -d '[
  {
    "traceId": "4e441824ec2b6a44",
    "id": "4e441824ec2b6a44",
    "name": "get",
    "timestamp": 1478624611519000,
    "annotations": [
    {
      "timestamp": 1478624611519000,
      "value": "Server Send Error: TimeoutException: socket timed out",
      "endpoint": {
        "ipv4": "192.168.99.1",
        "port": 9411
      }
    },
    {
      "timestamp": 1478624611519001,
      "value": "{\"foo\": \"bar\"}",
      "endpoint": {
        "ipv4": "192.168.99.1",
        "port": 9411
      }
    }],
    "binaryAnnotations": [
    {
      "key": "bar",
      "value": "bar",
      "endpoint": {
        "ipv4": "192.168.99.1",
        "port": 9411
      }
    },
      {
        "key": "foo",
        "value": "{\"foo\": \"bar\"}",
        "endpoint": {
          "ipv4": "192.168.99.1",
          "port": 9411
        }
      }
    ]
  }
  ]'
```

codefromthecrypt:
@kevinoliver @mosesn better late than never :P

codefromthecrypt:
(ps ignore the escaped json in the spans.. that's unrelated)

PR #1476, opened 2017-01-09
Switches to release commit user instead of amplifying my commits

codefromthecrypt:
This switches to a dummy user instead of my email when doing release
commits.


codefromthecrypt:
cc @openzipkin/devops-tooling

PR #1477, opened 2017-01-09
Update developers section in pom.xml

codefromthecrypt:
Developers section content is outdated. It is required
for publishing to Maven Central so update with Zipkin dev
google group info.


codefromthecrypt:
cc @openzipkin/devops-tooling we did the same in brave via https://github.com/openzipkin/brave/pull/326

Issue #1484, opened 2017-01-11
UI common problems detection and highlight


fedj:
When integrating Zipkin, it can happen that the instrumentation acts weirdly.
When there is a problem in the trace (example: SS is after SR), the trace is displayed badly and that's the only way to detect that there is a problem.

I think that there are common patterns that could be detected at runtime (and we could add more and more rules with more experience).

The idea would be that the UI could display warnings about wrong patterns to alert integrators that there might be a problem.

It can also be logged in the collector for spans problems.

codefromthecrypt:
cc @cburroughs

cburroughs:
Would the actual "badness" detection happen on the server, or in javascript?  My initial inclination leans towards server side, but I"m not sure if adding more fields to the API response has complications or not.

codefromthecrypt:
I think the problem is easier to take action on in the client side, not
only because there are non-UI users of the API, but also purely from a
"what action to take" standpoint. I suspect there are more ways to help
visually vs trying to encode in a message format. That said, there's
probably prior art somewhere that does a good job with either way.

Issue #324, opened 2014-01-10
Add Query / UI support for traces with multiple parent spans


hoxworth:
While instrumenting a number of distributed services recently to enable end to end tracing, I ran into a snag: zipkin's UI seems to only support displaying a single span that lacks a parent ID in a trace. 

My services are all guaranteed to receive a unique ID which the trace ID is based off of, but services that read from a queue are not dependent on an invocation from a parent service, and thus start their own new spans. This introduces multiple spans in a trace that do not have parent IDs, which the zipkin UI (and possibly query engine) cannot handle currently.

Yes, I could possibly pass a parent ID along in the message queue, but the services are logically separated; this just seems like a hack.

bmdhacks:
Yeah, this is related to other issues where some work unit will get enqueued to be executed later, such as a web submission that spawns (or contributes to) a hadoop job an hour later.  It would be nice to link individual traces together if they are related.

jorgheymans:
This is an older issue relating to zipkin classic. Nevertheless i verified that Lens, the new UI, is able to handle such spans and it does. I took the yelp test trace https://github.com/openzipkin/zipkin/blob/master/zipkin-lens/testdata/yelp.json and modified both mysql spans to have no parent. 

![image](https://user-images.githubusercontent.com/193792/79500161-4f906180-802c-11ea-850f-3ae96d931fbe.png)

In any case, tracing over messaging systems is very well supported in brave nowadays so it would be more natural to let the trace continue over the broker into the consumer.

PR #1487, opened 2017-01-13
do not silently drop subsequent spans missing a parent

cburroughs:
Currently zipkin models traces as a tree of spans.  During the
conversation from a list (or whatever the underlying storage engine
uses) to a tree, spans are dropped if more than one of them is missing
a parentId (it is a "root" node).  This appears to be an unintentional
regression (possibly around 9d8261a33c3adaad9aecd290744b651290c72e24),
since the comment still indicate a workaround for this problem:
attribute missing parents to root.

This is still imperfect, but exposes it to the the user so they can
debug the instrumentation problem instead of silently dropping data.
Note that the workaround only applies when constructing a tree
internally the actual spans returned by the API still have no
parentID, and this is what is displayed in the web ui.

This problem is alluded to in #324, but the thrust of that ticket is
modeling asynchronous actions.

Highlighting suspect traces is discussed in #1484 and this condition
would be a prime candidate.

### DIFF
--- a/zipkin/src/test/java/zipkin/internal/NodeTest.java
+++ b/zipkin/src/test/java/zipkin/internal/NodeTest.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2016 The OpenZipkin Authors
+ * Copyright 2015-2017 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -14,7 +14,9 @@
 package zipkin.internal;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 import org.junit.Test;
 import zipkin.Span;
@@ -77,5 +79,24 @@ public class NodeTest {
     Node<Span> child = root.children().iterator().next();
     assertThat(child.children()).extracting(Node::value)
         .containsExactly(TestObjects.TRACE.get(2));
+  }
+
+  @Test
+  public void constructsTraceTree_noChildLeftBehind() {
+    List<Span> spans = Arrays.
+      asList(
+             Span.builder().traceId(137L).id(1L).name("root-0").build(),
+             Span.builder().traceId(137L).parentId(1L).id(2L).name("child-0").build(),
+             Span.builder().traceId(137L).parentId(1L).id(3L).name("child-1").build(),
+             Span.builder().traceId(137L).id(4L).name("lost-0").build(),
+             Span.builder().traceId(137L).id(5L).name("lost-1").build());
+    int treeSize = 0;
+    Node<Span> tree = Node.constructTree(spans);
+    Iterator<Node<Span>> iter = tree.traverse();
+    while (iter.hasNext()) {
+      iter.next();
+      treeSize++;
+    }
+    assertThat(treeSize).isEqualTo(spans.size());
   }
 }
--- a/zipkin-ui/js/component_ui/spanPanel.js
+++ b/zipkin-ui/js/component_ui/spanPanel.js
@@ -91,7 +91,8 @@ export default component(function spanPanel() {
 
     const $moreInfoBody = this.$node.find('#moreInfo tbody').text('');
     const moreInfo = [['traceId', span.traceId],
-                       ['spanId', span.id]];
+                      ['spanId', span.id],
+                      ['parentId', span.parentId]];
     $.each(moreInfo, (i, pair) => {
       const $row = self.$moreInfoTemplate.clone();
       $row.find('.key').text(pair[0]);
--- a/zipkin-ui/templates/trace.mustache
+++ b/zipkin-ui/templates/trace.mustache
@@ -44,7 +44,7 @@
   <div
     id='{{spanId}}'
     class='span service-span depth-{{depthClass}}'
-    data-keys='id,traceId,spanName,serviceNames,serviceName,durationStr,duration'
+    data-keys='id,traceId,parentId,spanName,serviceNames,serviceName,durationStr,duration'
     data-id='{{spanId}}'
     data-trace-id='{{traceId}}'
     data-parent-id='{{parentId}}'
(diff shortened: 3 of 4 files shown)


zipkin/src/test/java/zipkin/internal/NodeTest.java
```diff
@@ -78,4 +80,23 @@ public void constructsTraceTree() {
     assertThat(child.children()).extracting(Node::value)
         .containsExactly(TestObjects.TRACE.get(2));
   }
+
+  @Test
+  public void constructsTraceTree_noChildLeftBehind() {
```
codefromthecrypt, line 85: :heart:

codefromthecrypt:
Thanks for digging into this.

so before, people used `/api/v1/trace/my-id?raw` to debug data (as that skips adjustment). The concept of "root-most-span" has been around a while in various forms, seemed to be a usability thing, for the UI. Can you show a screen shot of how the UI behaves before and after.. with a corrupt span like this?

Regardless, I like making it obvious that a trace is malformed, and allowing multi-roots through should be a signal of that! (maybe we can use it to yell at a user one day so they don't get confused)

codefromthecrypt:
looking at this deeper with sample data (ex delete parentId from a span in the trace)

codefromthecrypt:
I ran this locally and I think it is better than before. We should follow-up with #1484 to see if there's a way to visualize problems better.

Before:
screen shot 2017-01-16 at 6 15 41 pm

After:
screen shot 2017-01-16 at 6 13 47 pm

Data:
```bash
curl -s localhost:9411/api/v1/spans -H'Content-Type: application/json' -d '[
  {
    "traceId": "d91d3ff6ac8d3ee0",
    "id": "d91d3ff6ac8d3ee0",
    "name": "get",
    "timestamp": 1484560599062664,
    "duration": 9880,
    "annotations": [
      {
        "timestamp": 1484560599062664,
        "value": "sr",
        "endpoint": {
          "ipv4": "192.168.99.1",
          "port": 9411
        }
      },
      {
        "timestamp": 1484560599072544,
        "value": "ss",
        "endpoint": {
          "ipv4": "192.168.99.1",
          "port": 9411
        }
      }
    ],
    "binaryAnnotations": [
      {
        "key": "ca",
        "value": true,
        "endpoint": {
          "serviceName": "",
          "port": 50278,
          "ipv6": "::1"
        }
      },
      {
        "key": "http.status_code",
        "value": "200",
        "endpoint": {
          "ipv4": "192.168.99.1",
          "port": 9411
        }
      },
      {
        "key": "http.url",
        "value": "/api/v1/traces",
        "endpoint": {
          "ipv4": "192.168.99.1",
          "port": 9411
        }
      }
    ]
  },
  {
    "traceId": "d91d3ff6ac8d3ee0",
    "id": "7e88135971d84c28",
    "name": "get-traces",
    "timestamp": 1484560599063323,
    "duration": 8261,
    "binaryAnnotations": [
      {
        "key": "lc",
        "value": "CassandraStorage",
        "endpoint": {
          "ipv4": "192.168.99.1",
          "port": 9411
        }
      },
      {
        "key": "request",
        "value": "QueryRequest{serviceName=zipkin-server, spanName=get-span-names, annotations=[], binaryAnnotations={foo=bar, baz=qux}, minDuration=null, maxDuration=null, endTs=1484560580968, lookback=3600000, limit=10}",
        "endpoint": {
          "ipv4": "192.168.99.1",
          "port": 9411
        }
      }
    ]
  },
  {
    "traceId": "d91d3ff6ac8d3ee0",
    "id": "918ae5aa60724d7b",
    "name": "select-trace-ids-by-span-name",
    "parentId": "7e88135971d84c28",
    "timestamp": 1484560599063688,
    "duration": 5128,
    "annotations": [
      {
        "timestamp": 1484560599063688,
        "value": "cs",
        "endpoint": {
          "ipv4": "192.168.99.1",
          "port": 9411
        }
      },
      {
        "timestamp": 1484560599068816,
        "value": "cr",
        "endpoint": {
          "ipv4": "192.168.99.1",
          "port": 9411
        }
      }
    ],
    "binaryAnnotations": [
      {
        "key": "cql.query",
        "value": "SELECT ts,trace_id FROM service_span_name_index WHERE service_span_name=:service_span_name AND ts>=:start_ts AND ts<=:end_ts ORDER BY ts DESC LIMIT :limit_;",
        "endpoint": {
          "ipv4": "192.168.99.1",
          "port": 9411
        }
      }
    ]
  },
  {
    "traceId": "d91d3ff6ac8d3ee0",
    "id": "b05692b601e64647",
    "name": "select-trace-ids-by-annotation",
    "parentId": "7e88135971d84c28",
    "timestamp": 1484560599064213,
    "duration": 7313,
    "annotations": [
      {
        "timestamp": 1484560599064213,
        "value": "cs",
        "endpoint": {
          "ipv4": "192.168.99.1",
          "port": 9411
        }
      },
      {
        "timestamp": 1484560599071526,
        "value": "cr",
        "endpoint": {
          "ipv4": "192.168.99.1",
          "port": 9411
        }
      }
    ],
    "binaryAnnotations": [
      {
        "key": "cql.query",
        "value": "SELECT ts,trace_id FROM annotations_index WHERE annotation=:annotation AND bucket IN :bucket AND ts>=:start_ts AND ts<=:end_ts ORDER BY ts DESC LIMIT :limit_;",
        "endpoint": {
          "ipv4": "192.168.99.1",
          "port": 9411
        }
      }
    ]
  },
  {
    "traceId": "d91d3ff6ac8d3ee0",
    "id": "618c29891541be7f",
    "name": "select-trace-ids-by-annotation",
    "timestamp": 1484560599064714,
    "duration": 5475,
    "annotations": [
      {
        "timestamp": 1484560599064714,
        "value": "cs",
        "endpoint": {
          "ipv4": "192.168.99.1",
          "port": 9411
        }
      },
      {
        "timestamp": 1484560599070189,
        "value": "cr",
        "endpoint": {
          "ipv4": "192.168.99.1",
          "port": 9411
        }
      }
    ],
    "binaryAnnotations": [
      {
        "key": "cql.query",
        "value": "SELECT ts,trace_id FROM annotations_index WHERE annotation=:annotation AND bucket IN :bucket AND ts>=:start_ts AND ts<=:end_ts ORDER BY ts DESC LIMIT :limit_;",
        "endpoint": {
          "ipv4": "192.168.99.1",
          "port": 9411
        }
      }
    ]
  }
]'
```

codefromthecrypt (approved):

codefromthecrypt:
At first, I was a bit hesitant, but after seeing this in action, I think it would be silly not to merge. Thanks, @cburroughs!

PR #1489, opened 2017-01-16
Fixes bad self-tracing configuration resulting in unknown serviceName

### LINKED ISSUE #1486: Unknown service in Zipkin UI when self tracing is enabled
Version 1.19.1 reports "unknown" as the service name when self tracing is enabled.  I haven't dug too deep but rolling back to 1.19.0 fixes the problem.

image

codefromthecrypt:
Fixes #1486

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,7 +45,7 @@
 
     <main.basedir>${project.basedir}</main.basedir>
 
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
cc @llinder ps something relating to self-tracing, but not related to this noticed .. another fix coming up!

PR #1490, opened 2017-01-16
Consolidates logic that deals with embedded IPv4 addresses

codefromthecrypt:
This consolidates logic around mapped or embedded IPv4 addresses, as
previously logic was duplicated here and in Brave.

See #1404
See https://github.com/openzipkin/brave/pull/337

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,7 +45,7 @@
 
     <main.basedir>${project.basedir}</main.basedir>
 
-    <brave.version>4.0.1</brave.version>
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
fyi @cburroughs

codefromthecrypt:
This fixes a situation where self-tracing mistook localhost for an embedded ipv4 address "0.0.0.1"

here's a screenshot of the fix:

screen shot 2017-01-16 at 5 19 51 pm

PR #1491, opened 2017-01-16
Skips creating dependency links when the serviceName is empty string

codefromthecrypt:
Sometimes, empty string is sent for endpoint.serviceName. This prevents
empty string from showing up in the dependency graph.


codefromthecrypt:
here's the before and after:

screen shot 2017-01-16 at 5 51 04 pm

screen shot 2017-01-16 at 5 45 36 pm

PR #1496, opened 2017-01-18
Adds issue template to hopefully reduce the Q/A issues opened

codefromthecrypt:
Adding an issue template will make what's appropriate more explicit. The
goal here is to set expectations.

The text here was modified from OkHttp


codefromthecrypt:
cc @openzipkin/core note I duplicated this as https://github.com/openzipkin/brave/pull/345 and will propagate it to other repos once merged.

shakuzen (approved):

Issue #1243, opened 2016-08-17
Support async spans


codefromthecrypt:
We've had many issues about representing one-way activity. This is an attempt to define the most semantic means to address this concern in zipkin v1.

In zipkin, we have an RPC span, where two hosts participate. They participate by adding bookend annotations to a single span. This allows for many conveniences including hosts able to report incomplete work, and report in any order.

For example, a typical RPC span is ["cs", "sr", "ss", "cr"], for a dual host (client+server) span, or either of its partial ones ["sr", "ss"](uninstrumented client) and ["cs", "cr"](uninstrumented server).

The path of least resistance for async (half-duplex, or messaging), is to follow this model. This prevents us from needing to wait for a big-bang zipkin 2 release of the entire ecosystem.

My proposal is to add a messaging span: "ms" and "mr". 

The message sender adds a "ms" when a one-way message is sent. They are free to add "ws" to indicate it going on the wire.

The message receiver adds a "mr" when a one-way message is received. They are free to precede that with "wr" to indicate one being processed from the wire.

This works like the RPC spans, but to pin it down, here are example spans.

["mr"] <- message received from an uninstrumented sender. The host of "mr" is the processor.
["ms"] <- message sent to an uninstrumented recipient, or the message was dropped. The host of "mr" is the sender.
["ms", "mr"] <- message sent and received from instrumented hosts. Similarly, the host of "ms" is the sender and the host of "mr" is the receiver.

The duration of a complete messaging span can be a lot longer than an RPC span, potentially orders of magnitude longer than the processing time. Once we have a means to indicate this (via "ms" "mr"), we can teach the UI to collapse the timeline or do other tricks.

Messaging spans need to be propagated, just like RPC ones. For example, storing the binary version of a trace id as metadata on the message is appropriate.

This design adds pressure for "linked traces" especially when messages are needed to create a new outbound activity. This is out of scope of this design, but not precluded by it. In other words, we need to take on "linked traces" separately.

Fixes #1189
Fixes #925

adriancole:
let's see if we can nail a design down for zipkin v1 model here: https://github.com/openzipkin/zipkin/issues/1243

codefromthecrypt:
related issue: multiple parents aka linked traces https://github.com/openzipkin/zipkin/issues/1244

codefromthecrypt:
had a chat with @mbogoevici from spring-cloud-stream.

[...]

PR #1497, opened 2017-01-18
Supports backfilling timestamp for one-way spans

codefromthecrypt:
Before, we only supported backfilling timestamp and duration for spans
that have at least two annotations. This special cases an incomplete
one-way span (ex only cs), by adding its timestamp. This prevents the
UI from crashing.

See #1243

### DIFF
--- a/zipkin/src/main/java/zipkin/internal/ApplyTimestampAndDuration.java
+++ b/zipkin/src/main/java/zipkin/internal/ApplyTimestampAndDuration.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2016 The OpenZipkin Authors
+ * Copyright 2015-2017 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
cc @dragontree101

codefromthecrypt:
Sending one side separately from the other..
```bash
$ curl -s localhost:9411/api/v1/spans -H'Content-type: application/json' -d '[
  {
    "traceId": "77047bf9cf04ed5d",
    "id": "77047bf9cf04ed5d",
    "name": "get",
    "timestamp": 1484722951945904,
    "duration": 156470,
    "annotations": [
      {
        "timestamp": 1484722951945904,
        "value": "sr",
        "endpoint": {
          "serviceName": "client"
        }
      },
      {
        "timestamp": 1484722952102374,
        "value": "ss",
        "endpoint": {
          "serviceName": "client"
        }
      }
    ]
  },
  {
    "traceId": "77047bf9cf04ed5d",
    "id": "40ef4e93ab3b1507",
    "name": "get",
    "parentId": "77047bf9cf04ed5d",
    "annotations": [
      {
        "timestamp": 1484722952100852,
        "value": "cs",
        "endpoint": {
          "serviceName": "client"
        }
      }
    ]
  }
]'
$ curl -s localhost:9411/api/v1/spans -H'Content-type: application/json' -d '[
  {
    "traceId": "77047bf9cf04ed5d",
    "id": "40ef4e93ab3b1507",
    "parentId": "77047bf9cf04ed5d",
    "annotations": [
      {
        "timestamp": 1484722952121651,
        "value": "sr",
        "endpoint": {
          "serviceName": "server"
        }
      }
    ]
  },
  {
    "traceId": "77047bf9cf04ed5d",
    "id": "c8a3701afd819d33",
    "name": "process",
    "parentId": "40ef4e93ab3b1507",
    "timestamp": 1484722952121748,
    "duration": 656,
    "binaryAnnotations": [
      {
        "key": "lc",
        "value": "message-processor",
        "endpoint": {
          "serviceName": "server"
        }
      }
    ]
  }
]'
```

we get this due to backfilling of timestamp and duration of the middle one-way span
screen shot 2017-01-18 at 3 05 53 pm

dragontree101:
now in zipkin-ui one-way is NaN?
image

codefromthecrypt:
I can't see your example, but this change will likely fix the UI. That's
because the old duration logic did not expect a span with only "cs" and "sr"

codefromthecrypt:
simplified code and updated description. key change here is making sure there's a timestamp added (which stops the UI from crashing)

codefromthecrypt:
PS this is a screenshot of the in-flight span. duration absent appearing as NaN is a separate issue, which I'll raise separately.

screen shot 2017-01-18 at 8 03 46 pm

PR #1503, opened 2017-01-23
Set result size when fetching dependency links from ElasticSearch

mbarbon:
Otherwise it only returns 10.

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/InternalElasticsearchClient.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/InternalElasticsearchClient.java
@@ -49,7 +49,7 @@ public abstract class InternalElasticsearchClient implements Closeable {
    *
    * <p> See https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-from-size.html
    */
-  protected static final int MAX_RAW_SPANS = 10000; // the default elasticsearch allowed limit
+  protected static final int MAX_RAW_ITEMS = 10000; // the default elasticsearch allowed limit
   protected static final String SPAN = "span";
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
thanks.

your test fails due to license header failure. you can run this to update it.

```
./mvnw com.mycila:license-maven-plugin:format -pl -:zipkin-ui
```

We should also make sure a test is present to prove that more than 10 links are returned (as it might be the case other span stores are silently failing.

Let me know, if you want to take this on or if I should!

zipkin-storage/elasticsearch/src/main/java/zipkin/storage/elasticsearch/InternalElasticsearchClient.java
```diff
@@ -49,7 +49,7 @@
    *
    * <p> See https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-from-size.html
    */
-  protected static final int MAX_RAW_SPANS = 10000; // the default elasticsearch allowed limit
+  protected static final int MAX_RAW_ITEMS = 10000; // the default elasticsearch allowed limit
```
codefromthecrypt, line 52: I think this is used in a subquery of spans, too, so it wouldn't be accurate to say max raw items (I can verify). Would be nice if there was a way to say "give me all"

codefromthecrypt: maybe `MAX_RESULT_WINDOW` instead, as that matches `index.max_result_window` which is what this is about?

codefromthecrypt:
I'll tidy this up on the way in. thanks!

codefromthecrypt:
replaced by #1504

codefromthecrypt:
going out now as 1.19.4

PR #1507, opened 2017-01-30
Add a date separator for elasticsearch index names

### LINKED ISSUE #1506: Add the ability to change Elasticsearch index date formatting
Hello,
By default, elasticsearch indexes are formatted the following way `
indexname-yyyy.MM.dd`by logstash for example ([doc here](https://www.elastic.co/guide/en/logstash/5.1/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-index)).
I recently had a request from a client to match their elasticsearch index pattern (which is the logstash one) and I couldn't because the current elasticsearch one is 'yyyy-MM-dd'. So I was wondering if I could add the support for a `date separator` for the elasticsearch storage ?

PRs are incoming (for the zipkin code, docker image and the dependencies job).
Thanks.

ImFlog:
This PR fixes #1506 
The code could be better I guess so every comment is appreciated :)


codefromthecrypt:
I'll clean this up on the way in, notably adding docs and a couple tests

PR #1515, opened 2017-02-06
Do not add incorrect dependencies when processing partial/broken traces

mbarbon:
The current dependency building logic uses the first spen as root for span
trees that are missing a span with null parentId.

So for example given this span tree (where all spans are server spans):

    1: {traceId: 1, spanId: 1, serviceName: A}
    2: {traceId: 1, spanId: 2, serviceName: B, parentId: 1}
    3: {traceId: 1, spanId: 3, serviceName: C, parentId: 1}

if the first span is not received, the dependency graph will contain a B -> C
dependency, which is obviously wrong.

This commit uses an empty node as the tree root for this case instead of using
the first server span as the parent.

### DIFF
--- a/zipkin/src/main/java/zipkin/internal/DependencyLinker.java
+++ b/zipkin/src/main/java/zipkin/internal/DependencyLinker.java
@@ -68,20 +68,20 @@ public final class DependencyLinker {
     if (logger.isLoggable(FINE)) logger.fine("traversing trace tree, breadth-first");
     for (Iterator<Node<DependencyLinkSpan>> i = tree.traverse(); i.hasNext(); ) {
       Node<DependencyLinkSpan> current = i.next();
-      DependencyLinkSpan value = current.value();
+      DependencyLinkSpan currentSpan = current.value();
       if (logger.isLoggable(FINE)) {
-        logger.fine("processing " + value);
+        logger.fine("processing " + currentSpan);
       }
-      if (value == null) {
+      if (currentSpan == null) {
         logger.fine("skipping synthetic node for broken span tree");
         continue;
       }
       String child;
       String parent;
-      switch (value.kind) {
+      switch (currentSpan.kind) {
         case SERVER:
-          child = current.value().service;
-          parent = current.value().peerService;
+          child = currentSpan.service;
+          parent = currentSpan.peerService;
           if (current == tree) { // we are the root-most span.
             if (parent == null) {
               logger.fine("root's peer is unknown; skipping");
@@ -90,8 +90,8 @@ public final class DependencyLinker {
           }
           break;
         case CLIENT:
-          child = current.value().peerService;
-          parent = current.value().service;
+          child = currentSpan.peerService;
+          parent = currentSpan.service;
           break;
         default:
           logger.fine("non-rpc span; skipping");
--- a/zipkin/src/main/java/zipkin/internal/DependencyLinker.java
+++ b/zipkin/src/main/java/zipkin/internal/DependencyLinker.java
@@ -72,7 +72,7 @@ public final class DependencyLinker {
       if (logger.isLoggable(FINE)) {
         logger.fine("processing " + currentSpan);
       }
-      if (currentSpan == null) {
+      if (current.isSyntheticRootForPartialTree()) {
         logger.fine("skipping synthetic node for broken span tree");
         continue;
       }
@@ -110,7 +110,7 @@ public final class DependencyLinker {
           logger.fine("processing ancestor " + ancestor.value());
         }
         DependencyLinkSpan ancestorLink = ancestor.value();
-        if (ancestorLink != null &&
+        if (!ancestor.isSyntheticRootForPartialTree() &&
               ancestorLink.kind == DependencyLinkSpan.Kind.SERVER) {
           parent = ancestorLink.service;
           break;
--- a/zipkin/src/main/java/zipkin/internal/DependencyLinker.java
+++ b/zipkin/src/main/java/zipkin/internal/DependencyLinker.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2016 The OpenZipkin Authors
+ * Copyright 2015-2017 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


zipkin/src/test/java/zipkin/internal/DependencyLinkerTest.java
```diff
@@ -182,6 +182,18 @@ public void cannotLinkSingleSpanWithoutBothServiceNames() {
   }
 
   @Test
+  public void doesntLinkUnrelatedSpansWhenMissingRootSpan() {
+    List<DependencyLinkSpan> trace = asList(
```
codefromthecrypt: in tests, most of the time, we use incrementing numbers to make obvious the order in the tree. This confused me a little as 3 is the parent of 1 and 2.

Maybe rework a little?

long parentId = 1L; // by declaring a local, it is obvious that the parent is missing. It is more obvious because the traceId low is the same as the parent id.
// then use (parentId, 2L) and (parentId, 3L)

zipkin/src/test/java/zipkin/internal/DependencyLinkerTest.java
```diff
@@ -182,6 +182,18 @@ public void cannotLinkSingleSpanWithoutBothServiceNames() {
...
+        new DependencyLinkSpan(new TraceId(0L, 1L), 3L, 1L, Kind.SERVER, "service1", null),
+        new DependencyLinkSpan(new TraceId(0L, 1L), 3L, 2L, Kind.SERVER, "service2", null)
+    );
+
+    assertThat(new DependencyLinker()
+        .putTrace(trace.iterator()).link())
+        .isEmpty();
+  }
+
```
codefromthecrypt, line 196: maybe also add a test here for headless trace that has obvious links.

For example, 
(rootId, 2L, "service1") and (rootId, 2L, 4L, "service2")
(rootId, 3L, "service1") and (rootId, 3L, 5L, "service3")

This will make sure this didn't accidentally break the headless function (without resorting to integration tests)

zipkin/src/main/java/zipkin/internal/DependencyLinker.java
```diff
@@ -68,12 +68,17 @@ public DependencyLinker putTrace(Iterator<DependencyLinkSpan> spans) {
     if (logger.isLoggable(FINE)) logger.fine("traversing trace tree, breadth-first");
     for (Iterator<Node<DependencyLinkSpan>> i = tree.traverse(); i.hasNext(); ) {
       Node<DependencyLinkSpan> current = i.next();
+      DependencyLinkSpan value = current.value();
```
codefromthecrypt: rename this variable to currentSpan

zipkin/src/main/java/zipkin/internal/DependencyLinker.java
```diff
@@ -68,12 +68,17 @@ public DependencyLinker putTrace(Iterator<DependencyLinkSpan> spans) {
...
       if (logger.isLoggable(FINE)) {
-        logger.fine("processing " + current.value());
+        logger.fine("processing " + value);
+      }
+      if (value == null) {
+        logger.fine("skipping synthetic node for broken span tree");
```
codefromthecrypt: If I'm not mistaken, the only case where value is null is the case where we've added a synthetic (sentinel).
I think the code will be more clear if we show this a different way. ex. a package-private static constant `Node.MISSING_ROOT`

Then, you can compare against this reference, vs checking null.

ex. `if (current == Node.MISSING_ROOT) continue` (ps do this at the very top of this block before the processing statement)

zipkin/src/main/java/zipkin/internal/DependencyLinker.java
```diff
@@ -104,8 +109,11 @@ public DependencyLinker putTrace(Iterator<DependencyLinkSpan> spans) {
         if (logger.isLoggable(FINE)) {
           logger.fine("processing ancestor " + ancestor.value());
         }
-        if (ancestor.value().kind == DependencyLinkSpan.Kind.SERVER) {
-          parent = ancestor.value().service;
+        DependencyLinkSpan ancestorLink = ancestor.value();
```
codefromthecrypt, line 112: similarly here, if we use a sentinel instead of null, I think the code is more self-describing..

```
if (ancestor == Node.MISSING_ROOT) {
  // explain
} else ... existing code
```

zipkin/src/main/java/zipkin/internal/Node.java
```diff
@@ -145,9 +145,9 @@ public void addNode(Long parentId, long id, @Nullable V value) {
       for (Map.Entry<Long, Long> entry : idToParent.entrySet()) {
         Node<V> node = idToNode.get(entry.getKey());
         Node<V> parent = idToNode.get(entry.getValue());
-        if (parent == null && rootNode == null) { // handle headless trace
-          rootNode = node;
-        } else if (parent == null) { // attribute missing parents to root
+        if (parent == null) { // handle headless trace
+          if (rootNode == null)
```
codefromthecrypt: rootNode = Node.MISSING_ROOT

^^ something like this makes the code more obvious, plus you can search references for Node.MISSING_ROOT, where you can't search references for null

codefromthecrypt (commented):
Thanks tons for this.. probably was hard to figure out. I made suggestions focusing on using a sentinel of Node.MISSING_ROOT to reduce logic and make more clear what we're handling. Let me know if that doesn't make sense.

mbarbon:
I added the extra test as suggested (not pushed yet).

WRT Node.MISSING_ROOT: it can't be a statically-allocated node instance (because childs are added to it); I can make it a sentinel value passed to the TreeBuilder, and then user code can check with node.value() == MISSING_ROOT.

Specifically regarding == MISSING_ROOT: I'm getting a warning that maybe I don't mean reference equality. Given I do, how do you prefer for me to suppress the warning?

Or I can add an isFakeRoot() accessor to the node.

Let me know which solution you prefer.

codefromthecrypt:
good catch on the sentinel.

I think reference equality is actually good here.. I think the
SuppressWarnings("ReferenceEquality") annotation might work in
error-prone. If not, then I'd punt and use
MISSING_ROOT.equals(node.value()) as eventhough it looks not great, it
should work since equals doesn't bind the type of the input.

ex static final MISSING_ROOT = new Object();

shouldn't matter that this isn't of type DependencyLinkSpan since
equals accepts object

codefromthecrypt:
might be nice for logging to override the toString actually.. like..

static final Object MISSING_ROOT = new Object() {
    return "MISSING_ROOT";
  }
};

codefromthecrypt:
need a hand on this? fyi I'll be flying out on tuesday for a couple days though someone else can also review

mbarbon:
Sorry it took a while, I had other stuff to do at work.

In the end I did not go with the sentinel (using an Object subclass as the sentinel works, but it would make code like Span foo = node.value() throw a ClassCast exception, unless you special-case the sentinel in value() to be returned as null).

codefromthecrypt (approved):
sensible. thanks for the help and thinking things through (out loud!)

PR #1520, opened 2017-02-11
Makes storage auto-configuration properties serializable for Spark

codefromthecrypt:
zipkin-sparkstreaming is best being able to re-use properties here.
As jobs have to be serialized across a cluster, the easiest way is
to make storage properties objects serializable.


codefromthecrypt:
needed for https://github.com/openzipkin/zipkin-sparkstreaming/issues/8 cc @devinsba @naoman

codefromthecrypt:
see https://github.com/openzipkin/zipkin-sparkstreaming/pull/14

PR #1521, opened 2017-02-11
Exposes property derived MySQL datasource so sparkstreaming can use it

codefromthecrypt:
The only thing ZipkinMySQLStorageProperties is used for is creating a
datasource. This exposes that so that the logic doesn't have to be copy
pasted into zipkin-sparkstreaming.


codefromthecrypt:
see https://github.com/openzipkin/zipkin-sparkstreaming/pull/14

PR #1533, opened 2017-03-06
Revises info about Cassandra TTL

codefromthecrypt:
per chat with @pauldraper

### DIFF
--- a/zipkin-storage/cassandra3/README.md
+++ b/zipkin-storage/cassandra3/README.md
@@ -61,6 +61,9 @@ for use in user queries.
 ### Time-To_live
 Time-To-Live is default now at the table level. It can not be overridden in write requests.
 
+There's a different default TTL for trace data and indexes, 7 days vs 3 days respectively. The impact is that you can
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
@openzipkin/cassandra PS does anyone know how to adjust the TTL after the schema has been installed?

@pdraper tried this..

```bash
cqlsh> alter table zipkin.traces with default_time_to_live = 60480;
ServerError: java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.AssertionError
```

codefromthecrypt:
@michaelsembwever ideas on changing the default ttl on an existing schema?

codefromthecrypt:
opened another issue for how to modify ttl after the fact

yurishkuro:
Curious as to the reasoning behind different TTLs

codefromthecrypt:
This has been in place in some fashion since zipkin was around iirc. A
different ttl for data vs indexes supports retrieval by id for a longer
period without impacting the size or performance of search requests
(indexes). Runbooks in twitter had info go get trace by id when they were
in log files for example. This wouldnt need a search to work. Not sure if
this split is popular anymore though.

Issue #782, opened 2015-10-20
AuthN Support


srijs:
In our deployment of Zipkin, some internal customers are asking for authentication support, such that we can restrict the access on a per-service level. The reason for this is mostly that binary annotations may contain security-sensitive information.

What I would be interested in is securing the query service via some form of auth, and then modifying web to prompt and forward that information.

Thoughts?

adriancole:
So, there are two query servers, the one in this project would be limited
to what's supported in Finatra, and the 'zipkin-java' would be limited to
what's supported in spring boot. Neither of these are called directly by js
code.

Js code terminates on zipkin-web, which is actually worse, as it uses
ostrich, an ancient framework.

The best long term plan would be to refactor the js so it can call the
query server directly. This means you could use oauth etc (in spring boot)
or whatever Finatra supports.

I don't have a good answer tactically, but others more familiar with
ostrich might.

adriancole:
Actually I might be wrong about zipkin-web.. I think it is TwitterServer

adriancole:
I'll take some time to summarize and then clean up my prior comments.

codefromthecrypt:
The `zipkin-web` terminates all traffic from the UI. One way to reduce surface area is to only expose `zipkin-web`. There's work to do in order to lock that down, and bear in mind I've not seen all things possible in finagle-land.

`zipkin-web` is a [TwitterServer](https://github.com/twitter/twitter-server), and very similar to [Finatra](https://github.com/twitter/finatra). Most patterns of Finatra could be applied to `zipkin-web`. To do authentication, you'd likely end up writing a filter or explicit code. I've found two examples:  [finatra-angular-example](https://github.com/pvoznenko/finatra-angular-example/blob/master/src/main/scala/com/acme/ShoppingCart/controllers/Api/UsersApi.scala#L19) and [finatra-tweet-example](https://github.com/twitter/finatra/blob/master/http/src/test/scala/com/twitter/finatra/http/integration/tweetexample/main/filters/AuthFilter.scala)

The subject of authorization would be slightly more complex. There's a mix of responsibilities inside `zipkin-web`, presumably folks optimizing for how much data is going to the browser, but honestly, I've no idea. This mix presents itself as some javascript components calling json endpoints, and mustache template rendering doing other things. This scatters the authorization concern a bit, as if there was only a "normal" json query api, you could just secure that. For example, to authorize by service, code would likely need to wrap the underlying SpanStore object as that's the choke-point for data.

[...]

PR #1537, opened 2017-03-10
Add basic auth security configuration

devinsba:
I would like to add security onto our zipkin server so we can do away with our IP whitelists that we currently use. This is a small first step toward that. (Ideally I will have an OAuth type solution at some point)

This leverages only existing code that was already in the project so it should only grow the artifact by a few properties files and a pom.

Running the docker images with these changes:

```
$ docker run -it -p 9411:9411 -e ZIPKIN_BASIC_SECURITY_USERNAME=zipkin \
    -e ZIPKIN_BASIC_SECURITY_PASSWORD=password \
    -e JAVA_OPTS='-Dspring.profiles.active=security' zipkin:local
```

### DIFF
--- a/zipkin-server/src/main/resources/zipkin-server-security.yml
+++ b/zipkin-server/src/main/resources/zipkin-server-security.yml
@@ -1,0 +1,14 @@
+security:
+  basic:
+    enabled: true
+  user:
+    name: ${ZIPKIN_BASIC_SECURITY_USERNAME:}
+    password: ${ZIPKIN_BASIC_SECURITY_PASSWORD:}
+
+management:
+  security:
+    basic:
+      enabled: true
+    user:
+      name: ${ZIPKIN_BASIC_SECURITY_USERNAME:}
+      password: ${ZIPKIN_BASIC_SECURITY_PASSWORD:}
--- a/zipkin-server/src/main/resources/zipkin-server-shared.yml
+++ b/zipkin-server/src/main/resources/zipkin-server-shared.yml
@@ -136,17 +136,17 @@ spring:
 security:
   basic:
     enabled: ${ZIPKIN_BASIC_SECURITY:false}
-    user:
-      name: ${ZIPKIN_BASIC_SECURITY_USERNAME:}
-      password: ${ZIPKIN_BASIC_SECURITY_PASSWORD:}
+  user:
+    name: ${ZIPKIN_BASIC_SECURITY_USERNAME:}
+    password: ${ZIPKIN_BASIC_SECURITY_PASSWORD:}
 
 management:
   security:
     basic:
       enabled: ${ZIPKIN_BASIC_SECURITY:false}
-      user:
-        name: ${ZIPKIN_BASIC_SECURITY_USERNAME:}
-        password: ${ZIPKIN_BASIC_SECURITY_PASSWORD:}
+    user:
+      name: ${ZIPKIN_BASIC_SECURITY_USERNAME:}
+      password: ${ZIPKIN_BASIC_SECURITY_PASSWORD:}
 
 info:
   zipkin:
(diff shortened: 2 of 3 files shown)


zipkin-server/src/main/resources/zipkin-server-shared.yml
```diff
@@ -133,6 +133,21 @@ spring:
...
+      password: ${ZIPKIN_BASIC_SECURITY_PASSWORD:}
+
+management:
+  security:
+    basic:
+      enabled: ${ZIPKIN_BASIC_SECURITY:false}
+      user:
+        name: ${ZIPKIN_BASIC_SECURITY_USERNAME:}
+        password: ${ZIPKIN_BASIC_SECURITY_PASSWORD:}
+
```
devinsba: By default there will be no change for users other than seeing the randomly generated UUID password that spring-security-starter creates no matter what

devinsba:
EDIT: turns out the tests wanted me to not have the `security.user.*` properties even if they are empty, so I ended up switching to a profile anyway

codefromthecrypt:
I like the idea of finding a lightweight manner to progress #782 (which I think has been discussed offline as well, including using either custom auto-config or nginx proxy for the same).

Until reading further, it seems that even shared password would have drawbacks, notably that no instrumentation projects support it. That means if it was turned on, you could no longer use any instrumentation projects who use http transport, right? There's also the matter of how reusable this is to others.. I can't pretend to know that, so we should just ping folks.

I don't know the dependencies implied by spring-boot-starter-security, so it is probably more than just adding one existing jar and a config file layer.. have you tried this way?

codefromthecrypt:
ps I don't see any tests, but we'd check the impact to see if this affects both the api and the UI (presumably it affects both)

devinsba:
For the question of it affecting both the API and UI, yes it does affect both.

On the question of the reporters working with this: Sleuth does support this out of the box with a single extra bean being required to add the `Authorization` header.

Honestly I'd prefer a pluggable interface for this that would allow us to create a custom authentication module that we could drop into the docker image and would just work. I just wanted to put something out there quick in order to get a conversation started. I can try to find time to prototype an authentication provider component and create a PR with that? Or I can put together a proposal issue and try to get feedback from those in the gitter room.

codefromthecrypt:
as a conversation starter, I think it succeeded :) To continue it, I'd
probably ping folks on issues who have formerly mentioned auth
requirements. Sleuth is one of many libraries, so if it were me, I'd try to
see what folks from other languages are interested in, or how other similar
projects secure their http transports for telemetry data.

I'm glad you started this.. if you can keep it up, please summarize updates
to #782 <https://github.com/openzipkin/zipkin/issues/782> I'll make some
notes there, which might help.

Coalla:
When do you suppose to merge this?

codefromthecrypt:
I dont think anything has progressed. It is important we dont add
speculative features. For example, none of the open source zipkin reporters
would work out of the box as they dont expect to authenticate.

Besides comments above and others made in the past, we need to make sure
this isnt a "tease feature" (looks neat but no one in the ecosystem
supports it)

zipkin-server/src/main/resources/zipkin-server-security.yml
```diff
@@ -0,0 +1,14 @@
+security:
+  basic:
+    enabled: true
+  user:
+    name: ${ZIPKIN_BASIC_SECURITY_USERNAME:}
```
maryoush, line 5: would that support multiple credentials ?

devinsba:
Closing as my requirements have changed and a solution that works for everyone is unlikely to come from this work

codefromthecrypt:
Are you in a position to share your new requirements? In back of mind was
thinking we can re-eval some things about modularity when we go spring boot
2

devinsba:
We no longer need to use the http collector so an auth proxy just works
without having to worry about the reporters

codefromthecrypt:
also obviated by this workaround https://github.com/openzipkin/zipkin/issues/782#issuecomment-433306013

PR #1538, opened 2017-03-11
Limits names lookback in Elasticsearch

### LINKED ISSUE #1462: Elasticsearch http slow on getting span names when there's a lot of data

codefromthecrypt:
This reuses `QUERY_LOOKBACK` to limit names queries (service and span
names) from searching all indexes. This is particularly important for
high-volume sites (ex 100% sampling or high traffic), and those who do
not use curator or otherwise to prune data.

Fixes #1462 (at least helps a lot with no schema change)

### DIFF
--- a/zipkin-autoconfigure/storage-elasticsearch-http/src/main/java/zipkin/autoconfigure/storage/elasticsearch/http/ZipkinElasticsearchHttpStorageAutoConfiguration.java
+++ b/zipkin-autoconfigure/storage-elasticsearch-http/src/main/java/zipkin/autoconfigure/storage/elasticsearch/http/ZipkinElasticsearchHttpStorageAutoConfiguration.java
@@ -45,8 +45,11 @@ public class ZipkinElasticsearchHttpStorageAutoConfiguration {
   ElasticsearchHttpStorage.Builder esHttpBuilder(
       ZipkinElasticsearchHttpStorageProperties elasticsearch,
       @Qualifier("zipkinElasticsearchHttp") OkHttpClient client,
-      @Value("${zipkin.storage.strict-trace-id:true}") boolean strictTraceId) {
-    return elasticsearch.toBuilder(client).strictTraceId(strictTraceId);
+      @Value("${zipkin.storage.strict-trace-id:true}") boolean strictTraceId,
+      @Value("${zipkin.query.lookback:86400000}") int namesLookback) {
+    return elasticsearch.toBuilder(client)
+        .strictTraceId(strictTraceId)
+        .namesLookback(namesLookback);
   }
 
   /** cheap check to see if we are likely to include urls */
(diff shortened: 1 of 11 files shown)


codefromthecrypt:
circleci is freaking out on license header dates (which was a problem we had in travis prior)

zipkin-autoconfigure/storage-elasticsearch-http/src/main/java/zipkin/autoconfigure/storage/elasticsearch/http/ZipkinElasticsearchHttpStorageAutoConfiguration.java
```diff
@@ -45,8 +45,11 @@ StorageComponent storage(ElasticsearchHttpStorage.Builder esHttpBuilder) {
   ElasticsearchHttpStorage.Builder esHttpBuilder(
       ZipkinElasticsearchHttpStorageProperties elasticsearch,
       @Qualifier("zipkinElasticsearchHttp") OkHttpClient client,
-      @Value("${zipkin.storage.strict-trace-id:true}") boolean strictTraceId) {
-    return elasticsearch.toBuilder(client).strictTraceId(strictTraceId);
+      @Value("${zipkin.storage.strict-trace-id:true}") boolean strictTraceId,
+      @Value("${zipkin.query.lookback:86400000}") int namesLookback) {
```
codefromthecrypt, line 49: ps I'm ok to add this as a elasticsearch property instead. ex zipkin.storage.elasticsearch.names-lookback (we have similar things in cassandra)

leegggg: Add this as a elasticsearch property may be a better idea. `zipkin.query.lookback` is a 32bit int variable which has a maximum of 2_147_483_647(ms) that is about 59.6621 hours. It is really not enough. 
A `zipkin.storage.elasticsearch.names-lookback` property in secs or even days can be great. Or may be you can use a 64bit variable for `zipkin.query.lookback`.
ps. Should this be a new issue?

codefromthecrypt:
sounds like a bug. mind sending a PR to fix this (along with a unit test to
show it working)?

nice find!

leegggg:
@adriancole Thank you for your conform, I will check if there are still similar issues and try fixing them.

PR #1540, opened 2017-03-22
Disables scribe by default

### LINKED ISSUE #1192: Disable scribe by default
We routinely get user questions about TooLongFrameException. This is almost always someone accidentally using the scribe port 9410 as an http port by accident.

I think we should disable scribe by default and eliminate this problem.

Here are some issues logged against this, but it also comes up on gitter
https://github.com/openzipkin/docker-zipkin/issues/106
https://github.com/openzipkin/zipkin/issues/1186

codefromthecrypt:
Scribe has been listening on the legacy port 9410 to help with migration
off the old zipkin codebase. A year and a half has passed, and all major
instrumentation packages support HTTP. By disabling scribe by default,
we only listen on one port, which helps in configuration management and
also eliminate the common confusion where people accidentally send http
traffic to the scribe port.

Fixes #1192


codefromthecrypt:
leaving this open for a day or two in case someone isn't looking

shakuzen (approved):

PR #1543, opened 2017-03-23
UI: Preserve sort order when going back to start screen

joel-airspring:


codefromthecrypt:
Thanks for the start.

I'm not sure how this is supposed to work. Can you mention the steps that would preserve order?

I go and find traces, set the order to oldest, then click on a trace. Regardless of how I navigate to the search screen (back button, clicking on find trace etc), the order isn't kept.

this is using chrome, in case this helps.

joel-airspring:
1. Create two or more traces so you can see changes in sort order.
2. Open up the zipkin web page.
3. Click on 'Find Traces'.  You should see your traces listed below in the default sort order.
4. Change the sort order to Newest First.  You should see the traces re-sorted to match.
5. Click on 'Find Traces' again.

In the release version I was using (1.20.1), step 5 results in the traces being show with the default sort order.  With this PR, the traces will be shown with the Newest First sort order.  You'll notice that after clicking on 'Find Traces' for the second time, the sortOrder= query parameter shows up in the URL.

Clicking 'Find Traces' after changing the sort order is the only path that works, however, because the sort order HTML elements are not present in the default page but only after finding traces.  Since the default page doesn't have the sort order elements present, they can't be remembered.  So, this works:
a. Load page
b. Find traces
c. Change sort order
d. Find traces
e. Click on a trace.
f. Navigate back (sort order preserved)
but this does not work:
a. Load page
b. Find traces
c. Change sort order
d. ---- skip this step ------
e. Click on a trace.
f. Navigate back (sort order NOT preserved)

The {{#queryWasPerformed}} section of index.mustache is conditionally included.  If it were always included, the sort order would be a part of every page load and could always be remembered, so both scenarios above would work.  This PR provides a path that works with minimal changes.  Perhaps a slightly larger change would be less confusing.

Would do you think is the right approach here?

Thanks!
Joel



Notes:
I'm on OSX, using Chrome 56.0.2924.87.
I built my zipkin server locally using this command:
  ./mvnw -DskipTests --also-make -pl zipkin-server clean install
And ran it using this command:
  java -jar zipkin-server/target/zipkin-server-1.21.1-SNAPSHOT-exec.jar

joel-airspring:
I just noticed that this PR accidentally changed the default sort order.  I'll push a fix for that.

codefromthecrypt:
The trouble I have here is that I don't know how natural it is for someone to change the sort order then click search, and the problem is only sorted (no pun intended) when that's the case. It sounds fragile, because it requires someone to know the way to not make it break :)

@eirslett other ideas?

joel-airspring:
If the {{#queryWasPerformed}} section of index.mustache was always included, then the problem goes away.  The only thing that would change would be the default view from 0 traces showing.

From: 

current

To:

always

codefromthecrypt:
sounds ok to me..

joel-airspring:
I've made the change, plush squashed down to two commits:

* Mustache-ize UI sort order.
* Pass sort order as a query parameter.

codefromthecrypt:
I will take a look tomorrow.. if works and needs no changes, and no one objects, it will get in the next cut this weekend. thx for the help!

joel-airspring:
No problem.  We're beginning to use zipkin here at my job, and it's a great tool!

codefromthecrypt:
is your job air spring mattress or air spring software? :D

joel-airspring:
The latter.

If we hit it big, we're going to buy the mattress company though.  Ha ha!

codefromthecrypt:
works for me. if anyone complains, we'll revisit

PR #1549, opened 2017-04-04
Preformats multi-line tag/binary annotations in the UI

codefromthecrypt:
Before, we didn't preformat text that included newlines. This made them
run off the screen. Now we do.

Before:
screen shot 2017-04-04 at 1 14 16 pm

After:
screen shot 2017-04-04 at 1 54 20 pm


codefromthecrypt:
cc @felixbarny

felixbarny:
Nice!

Issue #1526, opened 2017-02-24
Linear response time on services query in elasticsearch


codefromthecrypt:
Getting some reports of eventual timeouts on service and span name queries on elasticsearch. Ex at 80GB/day you eventually will time out (reported by @pauldraper)

Ultimately one big issue is the shape of the json, which requires nested queries to access the service name of a span. The new reporting format (#1499) will allow us to eventually fix this part, but we can't pin to a span format-based implementation to fix a performance problem!

In cassandra3, we started with a materialized view, but had to switch to manual indexing for similar performance reasons (#1374). The downside is that it amplifies writes and also is less straightforward for those who might be simply writing json to ES (as opposed to via a collector).

Regardless of what we do, it would be nice to have a perf test which can showcase evidence of linear latency patterns on service or span name queries. This would catch any system having indexes which aren't ideal. 

We have some next steps to consider...

* firstly, properly diagnose the issue. As ES is primarily a cache, maybe there's some tuning of the index template to help with nested queries on service and span names
* next, see what needs to be done in the collector directly vs somewhere else. For example, if we have to do manual indexing we could consider an ES pipeline stage to compose that.

cc @openzipkin/elasticsearch

codefromthecrypt:
another thing I thought of is that we should collect notes about what sort of data they are putting into zipkin. For example, historically spans had very few tags and people used sampling. Particularly OpenTracing tend to put a lot of tags (each of which ends up in the subquery for service names). It will be interesting to know how many spans and how big they tend to be when users experience problems.

semyonslepov:
There are some statistics about our "bad" experience:

- ~6 Gb of traces in each of daily indices (keeping 5 last indices)
- ~1 Kb per span
- 3-4 tags per span 

( - running on m3.medium.elasticsearch in AWS, if it matters)

The main workload is writing and some users accidentally come to a web-interface for looking their traces.

codefromthecrypt:
Thanks very interesting. You have very few but large tags. Is it safe to do
math to extrapolate span count from this?

codefromthecrypt:
I have been thinking about this and believe adding a new index for service and span name is the best option. This would be just like cassandra where we keep a separate document for service/span lookup:

```
CREATE TABLE IF NOT EXISTS zipkin3.span_name_by_service (
    service_name text,
    span_name    text,
    PRIMARY KEY (service_name, span_name)
)
```

[...]

PR #1554, opened 2017-04-10
add internal cache for /api/v1/services with scheduled update

semyonslepov:
We use Zipkin with ElasticSearch in AWS and have the following problem: on a big amount of data request to /api/v1/services takes too long time and often we have timeouts in UI.
There is a try to implement opt-in internal caching of getServiceNames() result with scheduled updates. Now we are using this patch in our external environment, it improves a bit our users' experience. Hope it can help for upstream (maybe in another way of implementation).

### DIFF
--- a/zipkin-server/src/main/java/zipkin/server/ZipkinQueryApiV1.java
+++ b/zipkin-server/src/main/java/zipkin/server/ZipkinQueryApiV1.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2016 The OpenZipkin Authors
+ * Copyright 2015-2017 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -14,12 +14,16 @@
 package zipkin.server;
 
 import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.CacheControl;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
+import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.web.bind.annotation.CrossOrigin;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.bind.annotation.PathVariable;
@@ -57,7 +61,14 @@ public class ZipkinQueryApiV1 {
   int namesMaxAge = 300; // 5 minutes
   volatile int serviceCount; // used as a threshold to start returning cache-control headers
 
+  /** Internal cache for /api/v1/services, updated with fixed interval */
+  @Value("${zipkin.query.internal-service-names-cache-enabled:false}")
+  boolean internalServiceNamesCacheEnabled;
+  
+  private List<String> internalServiceNamesCache = new CopyOnWriteArrayList<String>();
+
   private final StorageComponent storage;
+  private static final Logger log = LoggerFactory.getLogger(ZipkinQueryApiV1.class);
 
   @Autowired
   public ZipkinQueryApiV1(StorageComponent storage) {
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
Here are two related issues on this topic to review before we talk about
new code and where:

cache control:
https://github.com/openzipkin/zipkin/issues/718#issuecomment-264052551
elasticsearch performance: https://github.com/openzipkin/zipkin/issues/1526

do you mind commenting inside each, particularly in #1526 note how much you
are ingesting and if you are doing any sampling.

semyonslepov:
Sure, I've seen these issues before, but I'll look for them again and put comments.

codefromthecrypt:
one last thing to verify. we recently compensated for this in #1538 
are you running latest?

semyonslepov:
Yes, we're already using it, it's a great thing and helps a lot! But I suppose if we increase our workload significantly (and yes, we will do it), we will again face the same problem (and we won't be able to decrease QUERY_LOOPBACK because it's not affordable for our users).

codefromthecrypt:
>
> Yes, we're already using it, it's a great thing and helps a lot! But I
> suppose if we increase our workload significantly (and yes, we will do it),
> we will again face the same problem (and we won't be able to decrease
> QUERY_LOOPBACK because it's not affordable for our users).
>
cool. just covering the bases..

I think we'll end up with a decision about where to introduce a cache if we
do (in UI code vs in the ES impl).

Also, if that cache needs to be managed like it is here (the response is so
slow that you can't rely on users to ever succeed). I kindof prefer not
having machinery in-process, as it would be handy regardless of the cache
impl for it to be user scheduled. maybe there's something we can do in the
UI to not block, but not make repeated calls, when a user first calls for
service and span names?

then there's also the chance we toy with data format.. I forget which issue
is tracking that one, but flattening service+span names inside ES might end
up being a way out.

adriancole:
I have been thinking about this and believe adding a new index for service and span name is the best option. This would be just like cassandra where we keep a separate document for service/span lookup:

```
CREATE TABLE IF NOT EXISTS zipkin3.span_name_by_service (
    service_name text,
    span_name    text,
    PRIMARY KEY (service_name, span_name)
)
```

Backfilling this index would be relatively simple and could be done a myriad of ways. Unlike adding new data to the span index, this approach doesn't risk people accidentally clobbering their spans to support it. It also doesn't depend on any specific span representation (current or future).

In implementation, easiest would be a daily index like we use for dependency links. We could set the document ID to service and span name so that the data doesn't grow unbounded.  If daily is too course granularity, we could make the ID naming convention include the hour of the day or minute of the hour and still significantly bound the data size.

We could make the read side look at both, and only execute the slow nested query when the ideal index contains no data. This means users don't have to do anything to start using it.

Operationally, I like this approach as being similar to cassandra means less cognitive overhead for cross-storage maintainers.

@openzipkin/elasticsearch @semyonslepov wdyt?

codefromthecrypt:
moved my last comment to #1526 which was the issue I thought I was on!

codefromthecrypt:
@semyonslepov assuming w/current perf we might be able to close this?

semyonslepov:
So, results of the feature with new indexes are really good at response times. For our current setup and number of users, it's good enough. 
But if there are tens/hundreds of users simultaneously opening Zipkin UI, it will be a disaster (because these requests are very expensive for ES). I wish it won't be the nearest future (:
If you see this type of machinery ugly for current architecture, we can close this PR now.

We will try to live with upstream version anyway (obviously it's much better than keep our local patches and synchronize them on new releases). In the bad case (I don't expect it now, but our setup is very young and we don't really know yet amount of our future users), we'll have to return to these patches or some similar machinery, because it doesn't depend on user count/RPS. (Or maybe there will be another better way to avoid such problems. For example, we thought about trying Cassandra instead of ES, haven't ever used it with Zipkin yet)

codefromthecrypt:
I would love to hear if you end up with hundreds of users who all refresh
browser cache/expire these headers :) would be a sign of a very effective
deployment.

I do think we will get better and probably simplest might be an caching
intermediary as it could be applied to all and invalidation isnt terribly
important with service names. Simpler than changing storage (plus not
positive of perf differences here or cassandra either). We could also
consider an optional caching decorator storagecomponent.

I do think we will change the format stored in ES by year end though this
will likely only help with trace queries as service span are now simple as
possible.

Meanwhile thanks for all the feedback. You have been very helpful in this!

PR #1555, opened 2017-04-10
remove traces with null timestamps on rendering

semyonslepov:
I began receiving "blank pages" (with header) when enabled self-tracing and tried to search for them with following JS error:

Uncaught TypeError: Cannot read property 'duration' of null (traceSummary.js)

Then figured out that there are some traces with absent timestamp field in root spans (actually, there are cases with some contiguous spans with null timestamps, starting from root spans).
As far as I remember, there are some existing valid cases (aren't they?) when timestamps may be null.
So, there is a straightforward (maybe not very elegant) solution with removing useless null-timestamped traces. Actually, I'm not sure that it's the right way to solve the problem, but it allows to see at least valid traces and not blank pages.

### DIFF
--- a/zipkin-ui/js/component_ui/traceSummary.js
+++ b/zipkin-ui/js/component_ui/traceSummary.js
@@ -220,13 +220,24 @@ export function mkDurationStr(duration) {
   }
 }
 
+function removeEmptyFromArray(array) {
+  const newArray = [];
+  for (let i = 0; i < array.length; i++) {
+    if (array[i]) {
+      newArray.push(array[i]);
+    }
+  }
+  return newArray;
+}
+
 export function traceSummariesToMustache(serviceName = null, traceSummaries, utc = false) {
   if (traceSummaries.length === 0) {
     return [];
   } else {
-    const maxDuration = Math.max(...traceSummaries.map((s) => s.duration)) / 1000;
+    const traceSummariesCleaned = removeEmptyFromArray(traceSummaries);
+    const maxDuration = Math.max(...traceSummariesCleaned.map((s) => s.duration)) / 1000;
 
-    return traceSummaries.map((t) => {
+    return traceSummariesCleaned.map((t) => {
       const duration = t.duration / 1000;
       const groupedTimestamps = getGroupedTimestamps(t);
       const serviceDurations = getServiceDurations(groupedTimestamps);


codefromthecrypt:
would be interesting to see what the trace had when it had no timestamp. possibly could be a one-way span which hasn't seen its caller yet (in-flight). if you have a trace that results in a blank page, possibly anonymizing that trace could lead to test data to put in the unit test for this.

semyonslepov:
There is an example of such trace (notice again that it is self-trace of Zipkin, haven't found such traces in other places yet):

    {
      "traceId": "03ac7fee421c76a0",
      "id": "6c353b6ed9baabf1",
      "name": "get-service-names",
      "parentId": "2cb3257011291dfe",
      "annotations": [
        {
          "timestamp": 1491454307215328,
          "value": "brave.flush",
          "endpoint": {
            "ipv4": "X.X.X.X",
            "port": 9411
          }
        }
      ],
      "binaryAnnotations": [
        {
          "key": "lc",
          "value": "AutoValue_ElasticsearchHttpStorage",
          "endpoint": {
            "ipv4": "X.X.X.X",
            "port": 9411
          }
        }
      ]
    },

    {
      "traceId": "03ac7fee421c76a0",
      "id": "21e299b050cd0a85",
      "name": "get-service-names",
      "parentId": "5ecae66e040c3a35",
      "timestamp": 1491436907215516,
      "duration": 13820,
      "binaryAnnotations": [
        {
          "key": "lc",
          "value": "AutoValue_ElasticsearchHttpStorage",
          "endpoint": {
            "ipv4": "X.X.X.X",
            "port": 9411
          }
        }
      ]
    },

    {
      "traceId": "03ac7fee421c76a0",
      "id": "01944d480bd78a91",
      "name": "get-service-names",
      "parentId": "21e299b050cd0a85",
      "timestamp": 1491437207215343,
      "duration": 17292,
      "binaryAnnotations": [
        {
          "key": "lc",
          "value": "AutoValue_ElasticsearchHttpStorage",
          "endpoint": {
            "ipv4": "X.X.X.X",
            "port": 9411
          }
        }
      ]
    },

// other spans with timestamps \\

What really surprises me is the fact that we have a timestamp in annotation of the first span, but not in the span.

codefromthecrypt:
heh looks like self-tracing is broke :) there's a brave.flush annotation!

codefromthecrypt:
on my honor, I will add backfill a test later :)

PR #1560, opened 2017-04-12
Refactors elasticsearch service and span name query to be like c*

### LINKED ISSUE #1526: Linear response time on services query in elasticsearch

codefromthecrypt:
Service and span names are written to the same daily indexes as spans 
and dependency links as the document type "servicespan".

The document is a simple structure, like:
```json
{
  "serviceName": "MyServiceName",
  "spanName": "MySpanName",
}
```

When the "servicespan" type is missing from an index, or there's no 
results returned, a fallback nested query is invoked.

Fixes #1526


codefromthecrypt:
ok code now sends service/span indexes in the same batch command. This should help with AWS transfer charges, especially since we gzip requests. cc @devinsba @semyonslepov  

still need some code cleanup and other notes in the TODO

codefromthecrypt:
backport logic in place. going to give a stab at a backport script as particularly serviceNames being current helps.

codefromthecrypt:
I think the best way to backport is to sneak hidden feature into zipkin-dependencies (as it is already a spark job and can do everything)

codefromthecrypt:
If someone wants to migrate old data, they'd need to do the following.

For each span in an index (ex zipkin-2017-02-14/span)
If span.name isn't empty, create a document like this for each service name in span.annotations[].endpoint.serviceName, span.binaryAnnotations[].endpoint.serviceName, 

id = myServiceName|mySpanName
{ "serviceName":"myServiceName",  "spanName":"mySpanName", 

save the unique set of documents to the "servicespan" type  in the same index (ex zipkin-2017-02-14/servicespan)

codefromthecrypt:
depends on #1562

codefromthecrypt:
fixed a very silly test setup bug which consumed a lot of my last night.. will be ready to go on green

PS ignoring circleci because it thinks the date is different than it is (a routine problem unrelated to this code)

PR #1565, opened 2017-04-16
trigger cassandra page fetch failure

llinder:


codefromthecrypt:
If this turns out to work without dogging the test time, lets comment on
6000 that it intends to trigger pagination

codefromthecrypt:
Ps might end up wanting to chop this into slices ex maybe 100 accept
commands with 60 spans each (guessing at cardinality that is best). This
implies mild join on async work, but could considerably speed up test
execution.

llinder:
Was hoping that this would have triggered it.  Setting the fetch size to something lower immediately triggers it so I'm a bit puzzled why this didn't.

I'm going to close this for now until I can trigger this in a local test.  Then I will tackle the accept async join bit.

llinder:
Locally I've been able to trigger this error with as few as 2000 traces.  After some investigation I see that the Cassandra tests are skipped in the Travis PR builds which is why this PR didn't fail.

Also doubling the trace count didn't have a noticeable impact on test time so I'm wondering if its worth the effort to fan out the accept calls.  I can still add that if it seems valuable though.

Right now I'm not sure how to trigger the failure in a pull request without merging it which will surely break the build :)

codefromthecrypt:
reopened as circleci was turned off for forks!

codefromthecrypt:
so my advice would be to use the least amount to trip the test (document the count).

Then, you can cherry-pick this commit into the one that fixes it (essentially this broken test doesn't merge until there's a fix)

sg?

codefromthecrypt:
uh-oh.. passed circleci!!

codefromthecrypt:
circle is running c* 3.10 iirc.. wonder if is version-specific, the
parameter?

llinder:
Looks like CircleCI is skipping the C* tests as well.  My suspicion is that C* is failing to fully start.  When I get a chance I will SSH into the build machine and look at the logs.

```
Running zipkin.storage.cassandra3.CassandraSpanStoreTest
21:10:37.900 [main] INFO  c.datastax.driver.core.ClockFactory - Using native clock to generate timestamps.
21:10:37.952 [main] INFO  com.datastax.driver.core.NettyUtil - Did not find Netty's native epoll transport in the classpath, defaulting to NIO.
Tests run: 41, Failures: 0, Errors: 0, Skipped: 41, Time elapsed: 3.571 sec

Results :

Tests run: 60, Failures: 0, Errors: 0, Skipped: 41
```

codefromthecrypt:
Thx for the tip. Makes sense

codefromthecrypt:
trying this after noticing the process crash locally

https://circleci.com/gh/openzipkin/zipkin/761

codefromthecrypt:
ok I think this failure is legit. we probably need to dockerize
cassandra tests and quit trying to run them in circleci (because
circle starts cassandra even if it is never used and it is really
fragile)

https://circleci.com/gh/openzipkin/zipkin/tree/cassandra-tests

Issue #1567, opened 2017-04-18
UI does not show latest spans


strizhov:
Hello,

We have a setup of zipkin jar (v1.22) with MySQL RDS backend. 

After about 20 mins collecting spans from our services, the UI does not show latest spans. 

A simple mysql query `select count(*) from zipkin_spans;` shows that database is being constantly updated with new spans, but UI search does not show them (in fact it shows that the latest span was 20 mins ago). 

We have run tcpdump that confirmed that there are new spans coming on port 9411 from our services.

If we switch back to memory based backend, there is no such issue, however we can't run memory backend due to out-of-memory problem discussed in https://github.com/openzipkin/zipkin/issues/1528

"/health" endpoint shows "UP" in all values.

Let me know if you need any additional details.
Thanks

codefromthecrypt:
troubleshooting is a bit easier on gitter https://gitter.im/openzipkin/zipkin

that said, it is interesting this pattern only exists in mysql so far (you could also try aws elasticsearch, but still we should figure out what's up).

try setting `SELF_TRACING_ENABLED=true` and see if zipkin traces make it even if yours don't? This might help narrow things down

codefromthecrypt:
here's an example query which might help, too https://github.com/openzipkin/zipkin/pull/1568

strizhov:
thanks for response. 

I have tried `SELF_TRACING_ENABLED=true` and received internal traces, but UI stopped reporting any new traces after about 20 mins run.

I also tried sql command from README file (after about 30 min zipkin collector run), but it didn't return anything. 

I am going to try cassandra storage at this point.

codefromthecrypt:
Sorry about timezones making this troubleshooting more difficult. So are
you saying that you saw internal traces but not app ones after 30m or that
you saw no traces? Seems like none if the sql query returned none. You
could check the /metrics endpoint to ensure you are receiving messages.

Yep es or c* are better long term anyway although i would like to
understand what you are bumping into.. guessing theres an error receiving
or writing data which is somehow not in you logs. Seeing metrics will at
least show you spans incoming and if these are plotted with something you
might see a correlation between the incoming messages counted and the count
in storage. If you look at the docker-zipkin repo theres a pull request
regarding metrics and prometheus if you dont have a different tool (or just
calling /metrics isnt enough)

feliksik:
Do make sure your browser is using the right begin and end time; that is, check whether you're not bitten by #1587 .

[...]

PR #1568, opened 2017-04-19
Adds a query which might help users troubleshoot their data in MySQL

codefromthecrypt:
Especially with custom or new instrumentation, pinpointing data problems
is important. This is an example query which can format things in a way
that is easier to work with in SQL.

See #1567
See #1566

### DIFF
--- a/zipkin-storage/mysql/README.md
+++ b/zipkin-storage/mysql/README.md
@@ -41,6 +41,38 @@ For example, all the below query the same trace using different tools:
 * zipkin-api: `http://1.2.3.4:9411/api/v1/trace/27960dafb1ea7454?raw`
 * mysql: `select * from zipkin_spans where trace_id = x'27960dafb1ea7454';`
 
+If you are trying to debug from data in the database, it is helpful to
+format IDs as hex, and timestamps as dates. The following is an example
+query which will return one line for each update to a span in the last
+5 minutes.
+
+```sql
+SELECT lower(concat(CASE trace_id_high
+                        WHEN '0' THEN ''
+                        ELSE hex(trace_id_high)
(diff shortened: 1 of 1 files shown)


zipkin-storage/mysql/README.md
```diff
@@ -41,6 +41,38 @@ For example, all the below query the same trace using different tools:
...
+
+```sql
+SELECT lower(concat(CASE trace_id_high
+                        WHEN '0' THEN ''
+                        ELSE hex(trace_id_high)
+                    END,hex(trace_id))) AS trace_id,
+       lower(hex(parent_id)) as parent_id,
+       lower(hex(id)) as span_id,
+       name,
+       from_unixtime(start_ts/1000000) as timestamp
```
codefromthecrypt, line 57: by reformatting into a date, it will be obvious when someone accidentally stores span.timestamp as seconds or millis (instead of micros). It will be obvious as the date will come out as 1970 (and in this query not return at all)

zipkin-storage/mysql/README.md
```diff
@@ -41,6 +41,38 @@ For example, all the below query the same trace using different tools:
 * zipkin-api: `http://1.2.3.4:9411/api/v1/trace/27960dafb1ea7454?raw`
 * mysql: `select * from zipkin_spans where trace_id = x'27960dafb1ea7454';`
 
+If you are trying to debug from data in the database, it is helpful to
+format IDs as hex, and timestamps as dates. The following is an example
+query which will return one line for each update to a span in the last
+5 minutes.
+
+```sql
+SELECT lower(concat(CASE trace_id_high
```
codefromthecrypt, line 50: while many ppl use 64bit trace ids, this part of the query will match what's in the json or B3 headers

PR #1571, opened 2017-04-21
Fix comments in docs about QUERY_LOOKBACK

semyonslepov:
```
3600*24*1000 = 86400000 (1day)
3600*24*1000*7 = 604800000 (7 days)
```

Just to not confuse users (: (It's correct in some places and wrong in another)


codefromthecrypt:
thank you!

PR #1573, opened 2017-04-24
use docker for C* tests

llinder:
Refactoring to match elasticsearch tests.

This has exposed a few failures that may have went unnoticed. Either that or I just broke things but I feel its probably a mixture so it might take some time to finish this up.

closes #1565


llinder:
Looks like the tests are still being skipped on the CI server.  This has been working locally for me so my best guess is that the CI server is a bit slower and the container isn't completely ready so the health check fails and subsequently the tests are skipped.  Might need to work on a custom CassandraContainer implementation that executes a command as the container wait condition.

llinder:
Success!  We now have 5 C* test failures in the CI build :-)

codefromthecrypt:
Congrats on the new failures :) knowing is half the battle

codefromthecrypt:
ok I rejigged a little on the tests. One thing I had to do was throttle the tests not to write so much at a time to the docker cassandra, as it makes it all crash and burn with BusyPoolException. There are two failing tests now, which we can take care of afterwards.

zipkin-storage/cassandra3/src/test/java/zipkin/storage/cassandra3/integration/CassandraDependenciesTest.java
```diff
@@ -0,0 +1,50 @@
...
+import static zipkin.internal.Util.midnightUTC;
+
+abstract class CassandraDependenciesTest extends DependenciesTest {
+  abstract protected Cassandra3Storage storage();
+
+  /**
+   * The current implementation does not include dependency aggregation. It includes retrieval of
+   * pre-aggregated links.
+   *
+   * <p>Note: The zipkin-dependencies-spark doesn't use any of these classes: it reads and writes to
```
codefromthecrypt, line 37: hehe well it would if there was a zipkin-dependencies-spark for cassandra3 :P

llinder: :D https://github.com/SmartThingsOSS/zipkin-dependencies-cassandra3

Tried to make that work in the Java project but couldn't figure out how to write TypeConverter classes for the UDT objects.  To be honest I didn't try hard to figure it out since I knew I could get it working in Scala fairly quickly.

PR #1578, opened 2017-04-28
Fixes ES service/span indexing for mid-tier server spans

codefromthecrypt:
Before, the service/span indexing skipped mid-tier server spans. This
corrects the associated logic so that the UI drop-down includes all
valid service names.


codefromthecrypt:
cc @openzipkin/elasticsearch this is a pretty important bugfix, so will roll out immediately

Thanks to @sirtyro for noticing this!

codefromthecrypt:
also cc @semyonslepov

Issue #1570, opened 2017-04-21
Consider caching name writes in elasticsearch


codefromthecrypt:
Right now, the ES span consumer does stateless writes to the "servicespan" type. For example, each message that contains a service and span name will try to write that name to the corresponding index (usually today).

This implies very redundant writes, which adds load to the ES cluster which isn't necessary. One way to avoid this is to keep a cache of service-span name points already written. This could be bounded so that faulty instrumentation don't end up blowing up memory on the collectors. for example it could be limited to 2 days (today and yesterday) and 1000 names each.

Implementation would be trimming the `Map<String, Set<Pair<String>>> indexToServiceSpans` which are used to write to the "servicespan" type, if they exist in that cache. We'd be careful that if adding to that cache, we remove the entries we added if a callback failed.

The easiest way to make such a cache without adding dependencies is using a hacked LinkedHashMap for the outer map and the keyset of a hacked linkedhash map to hold the pairs.

http://stackoverflow.com/questions/27475797/use-linkedhashmap-to-implement-lru-cache

Note, very likely normal synchronization would be fine.

PR #1579, opened 2017-04-30
Add cache for servicespans, reduce unnecessary writes

semyonslepov:
Trying to implement cache for servicespans to reduce excessive writes ( https://github.com/openzipkin/zipkin/issues/1570 ).
Cache limited with 2 days and 1000 entries for each day (non-configurable for now).

### DIFF
--- a/zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
+++ b/zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.Collections;
 import okio.Buffer;
 import zipkin.Codec;
 import zipkin.Span;
@@ -37,6 +38,15 @@ class ElasticsearchHttpSpanConsumer implements AsyncSpanConsumer { // not final 
 
   final ElasticsearchHttpStorage es;
   final IndexNameFormatter indexNameFormatter;
+  private final static int MAX_CACHE_DAYS = 2;
+  private final static int MAX_CACHE_ENTRIES = 1000;
+
+  private final static Map<String, Set<Pair<String>>> indexToServiceSpansCache = new LinkedHashMap<String, Set<Pair<String>>>() {
+    @Override
+    protected boolean removeEldestEntry(Map.Entry<String, Set<Pair<String>>> eldest) {
+      return size() > MAX_CACHE_DAYS;
+    }
+  };
 
   ElasticsearchHttpSpanConsumer(ElasticsearchHttpStorage es) {
     this.es = es;
@@ -48,9 +58,34 @@ class ElasticsearchHttpSpanConsumer implements AsyncSpanConsumer { // not final 
       callback.onSuccess(null);
       return;
     }
+    Map<String, Set<Pair<String>>> indexToServiceSpans = new LinkedHashMap<>();
     try {
       HttpBulkIndexer indexer = new HttpBulkIndexer("index-span", es);
-      Map<String, Set<Pair<String>>> indexToServiceSpans = indexSpans(indexer, spans);
+      indexToServiceSpans = indexSpans(indexer, spans);
+      
+      // remove already cached items from indexToServiceSpans, add new items to cache
+      for (Map.Entry<String, Set<Pair<String>>> entry : indexToServiceSpans.entrySet()) {
+        Set<Pair<String>> serviceSpansEntryCached = null;
+        synchronized(indexToServiceSpansCache) {
+          serviceSpansEntryCached = indexToServiceSpansCache.get(entry.getKey());
+        }
+        if (serviceSpansEntryCached == null) {
+          synchronized(indexToServiceSpansCache) {
+            indexToServiceSpansCache.put(entry.getKey(), serviceSpansEntryCached = Collections.newSetFromMap(new LinkedHashMap<Pair<String>, Boolean>() {
+              @Override
+              protected boolean removeEldestEntry(Map.Entry<Pair<String>, Boolean> eldest) {
+                return size() > MAX_CACHE_ENTRIES;
+              }
+            }));
+          }
+        }
+        Set<Pair<String>> serviceSpansEntry = entry.getValue();
+        synchronized(serviceSpansEntryCached) {
+          serviceSpansEntry.removeAll(serviceSpansEntryCached);
+          serviceSpansEntryCached.addAll(serviceSpansEntry);
+        }
+      }
+
       if (!indexToServiceSpans.isEmpty()) {
         indexNames(indexer, indexToServiceSpans);
       }
@@ -58,6 +93,19 @@ class ElasticsearchHttpSpanConsumer implements AsyncSpanConsumer { // not final 
     } catch (Throwable t) {
       propagateIfFatal(t);
       callback.onError(t);
+    
+      // remove recently added items from cache
+      for (Map.Entry<String, Set<Pair<String>>> entry : indexToServiceSpans.entrySet()) {
+        Set<Pair<String>> serviceSpansEntryCached = null;
+        synchronized (indexToServiceSpansCache) {
+          serviceSpansEntryCached = indexToServiceSpansCache.get(entry.getKey());
+        }
+        if (serviceSpansEntryCached == null || serviceSpansEntryCached.isEmpty()) continue;
+
+        synchronized (serviceSpansEntryCached) {
+          serviceSpansEntryCached.removeAll(entry.getValue());
+        }
+      }
     }
   }
 
--- a/zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
+++ b/zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
@@ -41,7 +41,7 @@ class ElasticsearchHttpSpanConsumer implements AsyncSpanConsumer { // not final 
   private final static int MAX_CACHE_DAYS = 2;
   private final static int MAX_CACHE_ENTRIES = 1000;
 
-  private final static Map<String, Set<Pair<String>>> indexToServiceSpansCache = new LinkedHashMap<String, Set<Pair<String>>>() {
+  final static Map<String, Set<Pair<String>>> indexToServiceSpansCache = new LinkedHashMap<String, Set<Pair<String>>>() {
     @Override
     protected boolean removeEldestEntry(Map.Entry<String, Set<Pair<String>>> eldest) {
       return size() > MAX_CACHE_DAYS;
--- a/zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
+++ b/zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
@@ -75,7 +75,7 @@ class ElasticsearchHttpSpanConsumer implements AsyncSpanConsumer { // not final 
     try {
       HttpBulkIndexer indexer = new HttpBulkIndexer("index-span", es);
       indexToServiceSpans = indexSpans(indexer, spans);
-      
+
       // remove already cached items from indexToServiceSpans, add new items to cache
       for (Map.Entry<String, Set<Pair<String>>> entry : indexToServiceSpans.entrySet()) {
         Set<Pair<String>> serviceSpansEntryCached = null;
@@ -108,7 +108,7 @@ class ElasticsearchHttpSpanConsumer implements AsyncSpanConsumer { // not final 
       callbackWrapper.onError(t);
     }
   }
-  /** Remove recently added items from cache */
+  /** Remove recently added items from servicespan cache */
   void clearIndexToServiceSpansCache(Map<String, Set<Pair<String>>> index, Map<String, Set<Pair<String>>> cache) {
     for (Map.Entry<String, Set<Pair<String>>> entry : index.entrySet()) {
       Set<Pair<String>> serviceSpansEntryCached = null;
@@ -121,6 +121,11 @@ class ElasticsearchHttpSpanConsumer implements AsyncSpanConsumer { // not final 
         serviceSpansEntryCached.removeAll(entry.getValue());
       }
     }
+  }
+
+  /** Clear servicespan cache (for testing purposes) */
+  void resetIndexToServiceSpansCache() {
+    indexToServiceSpansCache.clear();
   }
 
   /** Indexes spans and returns a mapping of indexes that may need a names update */
--- a/zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
+++ b/zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
@@ -47,6 +47,7 @@ class ElasticsearchHttpSpanConsumer implements AsyncSpanConsumer { // not final 
       return size() > MAX_CACHE_DAYS;
     }
   };
+  private Map<String, Set<Pair<String>>> indexToServiceSpans = new LinkedHashMap<>();
 
   ElasticsearchHttpSpanConsumer(ElasticsearchHttpStorage es) {
     this.es = es;
@@ -54,11 +55,23 @@ class ElasticsearchHttpSpanConsumer implements AsyncSpanConsumer { // not final 
   }
 
   @Override public void accept(List<Span> spans, Callback<Void> callback) {
+    Callback<Void> callbackWrapper = new Callback<Void>() {
+        @Override
+        public void onSuccess(Void value) {
+            callback.onSuccess(value);
+        }
+
+        @Override
+        public void onError(Throwable t) {
+            clearIndexToServiceSpansCache(indexToServiceSpans, indexToServiceSpansCache);
+            callback.onError(t);
+        }
+    };
+
     if (spans.isEmpty()) {
-      callback.onSuccess(null);
+      callbackWrapper.onSuccess(null);
       return;
     }
-    Map<String, Set<Pair<String>>> indexToServiceSpans = new LinkedHashMap<>();
     try {
       HttpBulkIndexer indexer = new HttpBulkIndexer("index-span", es);
       indexToServiceSpans = indexSpans(indexer, spans);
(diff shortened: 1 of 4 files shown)


zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
```diff
@@ -37,27 +38,45 @@
...
   final IndexNameFormatter indexNameFormatter;
+  private final static int MAX_CACHE_DAYS = 2;
+  private final static int MAX_CACHE_ENTRIES = 1000;
+
+  static Map<String, Set<Pair<String>>> indexToServiceSpansCache;
 
   ElasticsearchHttpSpanConsumer(ElasticsearchHttpStorage es) {
     this.es = es;
     this.indexNameFormatter = es.indexNameFormatter();
+    this.indexToServiceSpansCache = Collections.synchronizedMap(new LinkedHashMap<String, Set<Pair<String>>>() {
```
codefromthecrypt: if static, should be initialized outside the ctor.. maybe make it final static.

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
```diff
@@ -93,9 +112,22 @@
 
   void putServiceSpans(Map<String, Set<Pair<String>>> indexToServiceSpans, String index, Span s) {
     Set<Pair<String>> serviceSpans = indexToServiceSpans.get(index);
-    if (serviceSpans == null) indexToServiceSpans.put(index, serviceSpans = new LinkedHashSet<>());
-    for (String serviceName : s.serviceNames()) {
-      serviceSpans.add(Pair.create(serviceName, s.name));
+    synchronized(indexToServiceSpansCache) {
```
codefromthecrypt: you may end up needing to tune this later.. maybe for now put in some unit tests, then later some contended benchmarks

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
```diff
@@ -37,27 +38,45 @@
 
   final ElasticsearchHttpStorage es;
   final IndexNameFormatter indexNameFormatter;
+  private final static int MAX_CACHE_DAYS = 2;
```
codefromthecrypt, line 41: can make a comment that usually spans will be in the same day, except at midnight. below, that entries correspond to unique service/span combinations. it is probably find to say 1000, but even 10000 should be reasonable as there's not much ram involved (ex if you have 1000 services, you will need a count greater than that for the diversity of span names w/in a service).

codefromthecrypt:
thx for the start! I think it doesn't need to be configurable at the moment. we have many things that aren't configurable.

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
```diff
@@ -37,6 +38,15 @@
...
+  private final static int MAX_CACHE_ENTRIES = 1000;
+
+  private final static Map<String, Set<Pair<String>>> indexToServiceSpansCache = Collections.synchronizedMap(new LinkedHashMap<String, Set<Pair<String>>>() {
```
codefromthecrypt: I wouldn't do synchronized map here, because you have a good bit of places which need atomic ops (put if absent). more straightforward to just synchronize when you need to imho

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
```diff
@@ -48,16 +58,24 @@
...
         indexNames(indexer, indexToServiceSpans);
       }
       indexer.execute(callback);
     } catch (Throwable t) {
       propagateIfFatal(t);
       callback.onError(t);
+      synchronized(indexToServiceSpansCache) {
+          for (String index : indexToServiceSpans.keySet()) {
+              Set<Pair<String>> serviceSpansEntry = indexToServiceSpans.get(index);
+              Set<Pair<String>> serviceSpansEntryCached = indexToServiceSpansCache.get(index);
```
codefromthecrypt: here's how I think I'd write this..

no outer synchronized ..

```java
for (Map.Entry<String, Set<Pair<String>>> entry : indexToServiceSpans.entrySet()) {
  Set<Pair<String>> serviceSpansEntryCached = null;
  synchronized (indexToServiceSpansCache) { // lock the whole cache 
    serviceSpansEntryCached = indexToServiceSpansCache.get(entry.getKey());
  }
  if (serviceSpansEntryCached == null || serviceSpansEntryCached.isEmpty()) continue;

  synchronized (serviceSpansEntryCached) { // lock the cache for the index
    serviceSpansEntryCached.removeAll(entry.getValue());
  }
}
```

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
codefromthecrypt: also note.. this code probably needs to be extracted to a method, because you will need to do this in the catch block, but more importantly in response to Callback.onError. You'll need to decorate the input callback with one like this...

```java
new Callback() {
  void onSuccess(){
    realCallback.onSuccess();
  }
  void onError(Throwable t){
    clearCache(indexToServiceSpans);
    realCallback.onError(t);
  }
}

```

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
codefromthecrypt: looking at this, the logic is in a tight loop (for each span). I'd move the cache stuff to a higher scope, where you see `if (!indexToServiceSpans.isEmpty()) {`

semyonslepov: How can we move this stuff to "try" block if we do this removing only on fail (maybe I'm missing something)?

semyonslepov: Or you are talking about moving filling cache outside of putServiceSpans()? Messed a bit by comments (:
In this case I agree, don't see any reason for doing this job for each span too...

adriancole:
Sorry yep!  You got it ;)

codefromthecrypt:
I rebased and squashed this for you. need to do another pass on review as not sure where we left off

semyonslepov:
I'm not sure as well, hope I'll find time finally and work on it

zipkin/src/test/java/zipkin/TestObjects.java
```diff
@@ -65,6 +65,8 @@
           .addAnnotation(Annotation.create((TODAY + 100) * 1000, SERVER_RECV, APP_ENDPOINT))
           .addAnnotation(Annotation.create((TODAY + 250) * 1000, SERVER_SEND, APP_ENDPOINT))
           .addAnnotation(Annotation.create((TODAY + 300) * 1000, CLIENT_RECV, WEB_ENDPOINT))
+          .addAnnotation(Annotation.create((TODAY + 350) * 1000, SERVER_SEND, DB_ENDPOINT))
```
codefromthecrypt: If you are looking to create multiple annotations, I might do so in a different file as this one is used for all storage tests (and also some non-storage ones). Particularly, this creates an invalid "party-line" span, which has multiple servers for the same RPC.

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumerTest.java
```diff
@@ -110,83 +110,60 @@ public void writesSpanNaturallyWhenNoTimestamp() throws Exception {
...
+      "{\"index\":{\"_index\":\"zipkin-1971-01-01\",\"_type\":\"servicespan\",\"_id\":\"app|s\"}}\n"
+    );
   }
 
   @Test
   public void indexesServiceSpan_basedOnGuessTimestamp() throws Exception {
     es.enqueue(new MockResponse());
 
     Annotation cs = Annotation.create(
-        TimeUnit.DAYS.toMicros(365), // 1971-01-01
```
codefromthecrypt: please don't change the behavior of existing tests (if that's what's going on here.. can't tell if this is changing the test or if the file has been re-ordered). Main thing we need to know is if this is changing behavior or not.. usually we add tests to show new things happening and keeping the old tests help us know we didn't break anything.

semyonslepov: This particular one was reordering.
We need to make some changes in the tests here, because servicespan cache changes behaviour: some tests expect adding of some servicespan pairs in request to ES, but the purpose of this cache is not to create existing pairs twice and old tests can't work anymore

codefromthecrypt:
ah ok.. well anyway when you are ready for it, make github line comments or
otherwise to explain tests removed or repurposed.

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumerTest.java
```diff
@@ -110,83 +110,60 @@ public void writesSpanNaturallyWhenNoTimestamp() throws Exception {
   }
 
   @Test
-  public void indexesServiceSpan_explicitTimestamp() throws Exception {
```
semyonslepov: Some existing tests have been changed for several reasons:

- related functionality ("servicespan" caching) eliminates repetitive writing of the same "servicespan" pairs to ES, therefore breaking some tests expectations
- tests order is important in a case of "servicespan" caching, so reordered existing tests in the file for better readability

codefromthecrypt:
test order can bite us later perhaps. maybe we should expose a way to clean
the cache (package private access) and make sure it is invoked in a Before
block?

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumerTest.java
```diff
@@ -57,6 +57,8 @@ public void ensureIndexTemplate() throws IOException, InterruptedException {
     storage.ensureIndexTemplate();
     es.takeRequest(); // get version
     es.takeRequest(); // get template
+    // clear servicespan cache in order to prevent tests breaking
+    ((ElasticsearchHttpSpanConsumer)storage.asyncSpanConsumer()).resetIndexToServiceSpansCache();
```
codefromthecrypt, line 63: nice

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
```diff
@@ -37,29 +38,95 @@
...
+      indexToServiceSpans = indexSpans(indexer, spans);
+
+      // remove already cached items from indexToServiceSpans, add new items to cache
+      for (Map.Entry<String, Set<Pair<String>>> entry : indexToServiceSpans.entrySet()) {
+        Set<Pair<String>> serviceSpansEntryCached = null;
+        synchronized(indexToServiceSpansCache) {
+          serviceSpansEntryCached = indexToServiceSpansCache.get(entry.getKey());
+        }
+        if (serviceSpansEntryCached == null) {
+          synchronized(indexToServiceSpansCache) {
```
codefromthecrypt, line 86: next step is to make some concurrency tests... something like https://github.com/openzipkin/brave/blob/7fd9bae7ee3ba8abaa536ad1d29634099a337d83/brave/src/test/java/brave/CurrentTracingTest.java#L53

you'll want to make sure that they break when you remove the synchronized block

semyonslepov:
Still don't understand why these ES integration tests are failing... Executing the same thing with the same parameters on my notebook and it passes...

codefromthecrypt:
Do you have docker running on your laptop? These tests conditionally run if
docker (like docker machine or such) is installed

semyonslepov:
Yeah, it was docker issue because I ran tests on my MacBook and it's a bit tricky sometimes to run Docker properly there. 
Enabled Travis for my forked repo and finally fixed the cause.

codefromthecrypt:
Glad you got to the bottom of it and 100thx for putting in effort to fix it
on your own. Does this mean you think this is ready for a "final review"?

semyonslepov:
I'm not 100% sure about "final" (: But yes, anyway, I think now it's on some step which is ready for review.

codefromthecrypt:
all things in quotes aren't final :) but cool.. will do a pass in a bit

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
```diff
@@ -37,29 +38,95 @@
...
+  private final static int MAX_CACHE_DAYS = 2;
+  private final static int MAX_CACHE_ENTRIES = 1000;
+
+  final static Map<String, Set<Pair<String>>> indexToServiceSpansCache = new LinkedHashMap<String, Set<Pair<String>>>() {
+    @Override
+    protected boolean removeEldestEntry(Map.Entry<String, Set<Pair<String>>> eldest) {
+      return size() > MAX_CACHE_DAYS;
+    }
+  };
+  private Map<String, Set<Pair<String>>> indexToServiceSpans = new LinkedHashMap<>();
```
codefromthecrypt, line 50: nit final

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
```diff
@@ -37,29 +38,95 @@
...
+
+      // remove already cached items from indexToServiceSpans, add new items to cache
+      for (Map.Entry<String, Set<Pair<String>>> entry : indexToServiceSpans.entrySet()) {
+        Set<Pair<String>> serviceSpansEntryCached = null;
+        synchronized(indexToServiceSpansCache) {
+          serviceSpansEntryCached = indexToServiceSpansCache.get(entry.getKey());
+        }
+        if (serviceSpansEntryCached == null) {
+          synchronized(indexToServiceSpansCache) {
+            indexToServiceSpansCache.put(entry.getKey(), serviceSpansEntryCached = Collections.newSetFromMap(new LinkedHashMap<Pair<String>, Boolean>() {
```
codefromthecrypt, line 87: I think there's still a race condition here. you should double-check lost race I think. (ex do indexToServiceSpansCache.get here in the synchronized block before put)

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
```diff
@@ -37,29 +38,95 @@
...
-      indexer.execute(callback);
+      indexer.execute(callbackWrapper);
     } catch (Throwable t) {
       propagateIfFatal(t);
-      callback.onError(t);
+      callbackWrapper.onError(t);
     }
   }
+  /** Remove recently added items from servicespan cache */
+  void clearIndexToServiceSpansCache(Map<String, Set<Pair<String>>> index, Map<String, Set<Pair<String>>> cache) {
```
codefromthecrypt, line 112: eventhough this makes a line break please rename "cache" to "indexToServiceSpansCache" or visa versa, so it is easier to tell what's going on from a lock pov.

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
```diff
@@ -37,29 +38,95 @@
...
+    for (Map.Entry<String, Set<Pair<String>>> entry : index.entrySet()) {
+      Set<Pair<String>> serviceSpansEntryCached = null;
+      synchronized (cache) {
+        serviceSpansEntryCached = cache.get(entry.getKey());
+      }
+      if (serviceSpansEntryCached == null || serviceSpansEntryCached.isEmpty()) continue;
```
codefromthecrypt, line 118: I suspect the serviceSpansEntryCached.isEmpty() needs to be in the synchronized block for visibility concerns.

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
```diff
@@ -37,29 +38,95 @@
 
   final ElasticsearchHttpStorage es;
   final IndexNameFormatter indexNameFormatter;
+  private final static int MAX_CACHE_DAYS = 2;
+  private final static int MAX_CACHE_ENTRIES = 1000;
+
+  final static Map<String, Set<Pair<String>>> indexToServiceSpansCache = new LinkedHashMap<String, Set<Pair<String>>>() {
```
codefromthecrypt, line 44: I think you want to hide this private to ensure it is read and modified only via synchronized methods

zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumer.java
```diff
@@ -37,29 +38,95 @@
...
+
+      synchronized (serviceSpansEntryCached) {
+        serviceSpansEntryCached.removeAll(entry.getValue());
+      }
+    }
+  }
+
+  /** Clear servicespan cache (for testing purposes) */
+  void resetIndexToServiceSpansCache() {
+    indexToServiceSpansCache.clear();
```
codefromthecrypt, line 128: this should be synchronized on indexToServiceSpansCache just to make it obvious. I'd probably make a package-private type that encapsulates access to this map, with the map itself private. All things that affect it would be package protected synchronized methods to make it easier to reason with.

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumerTest.java
```diff
@@ -14,6 +14,7 @@
 package zipkin.storage.elasticsearch.http;
 
 import java.io.IOException;
+import java.util.*;
```
codefromthecrypt, line 17: nit: no wildcards

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpSpanConsumerTest.java
```diff
@@ -266,6 +270,121 @@ public void addsPipelineId() throws Exception {
...
+  public void indexesServiceSpanCache_failedConcurrency() throws Exception {
+    final Integer threadCount = 10;
+    List<String> testElemsForCache = Arrays.asList("a", "b", "c", "d", "e", "f", "g", "h", "k", "l");
+    Boolean broken = false;
+    for (int j = 0; j < 10; j++) {
+      List<Thread> getOrSet = new ArrayList<>(threadCount * 2);
+      for (int i = 0; i < threadCount; i++) {
+        getOrSet.add(new Thread(
+          () -> testElemsForCache.forEach(elem ->
+            ElasticsearchHttpSpanConsumer.indexToServiceSpansCache.put(elem,
```
codefromthecrypt, line 312: I like the idea here, but I think you are testing map, not your code. If you encapsulate access to this indexToServiceSpansCache, you can test high-level methods you need (ex similar to how we have Multimap in InMemorySpanStore)

zipkin-storage/elasticsearch-http/src/test/java/zipkin/storage/elasticsearch/http/InternalForTests.java
```diff
@@ -43,6 +43,8 @@ public static void writeDependencyLinks(ElasticsearchHttpStorage es, List<Depend
 
   public static void clear(ElasticsearchHttpStorage es) throws IOException {
     es.clear();
+    // clear servicespan cache in order to prevent tests breaking
+    ((ElasticsearchHttpSpanConsumer)es.asyncSpanConsumer()).resetIndexToServiceSpansCache();
```
codefromthecrypt, line 47: es.clear should call this

codefromthecrypt (commented):
Thanks for the continued effort on this. I'm hoping we'll soon not need this (or need it less) due to the simplified json work. Nevertheless if you are still on board with continuing, I can continue to give feedback on it.

codefromthecrypt:
gonna close this as we should re-evaluate with the new single-type index. If needed, we can re-open

PR #1580, opened 2017-05-01
Helps tracers be more safe about IP Addresses

codefromthecrypt:
Before, tracers including Brave and Finagle blindly assumed addresses
and strings were IPv4. While that's usually the case, it can lead to
very late problems, such as runtime exceptions.

This adds a utility to encourage safe parsing practice of potentially
null inputs. By re-using code from guava, we can safely parse IP
literals as well, avoiding troublesome IP from name service lookups.

Ex. if your input is an `HttpServletRequest`, the following is safe:
```java
if (!builder.parseIp(input.getHeader("X-Forwarded-For"))) {
  builder.parseIp(input.getRemoteAddr());
}
```

### DIFF
--- a/zipkin/src/main/java/zipkin/internal/InetAddresses.java
+++ b/zipkin/src/main/java/zipkin/internal/InetAddresses.java
@@ -1,0 +1,191 @@
+/**
+ * Copyright 2015-2017 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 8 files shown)


zipkin/src/main/java/zipkin/internal/InetAddresses.java
```diff
@@ -0,0 +1,191 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin.internal;
+
+import java.net.InetAddress;
+import java.nio.ByteBuffer;
+
+/** Utilities for working with IP Addresses. */
+public final class InetAddresses {
```
codefromthecrypt, line 20: note an identical copy of this patched class was in Brave for quite a while now.

PR #1581, opened 2017-05-02
Breaking test for trace search in ES.

garyd203:
The Elastic Search implementation should be able to search traces
by serviceName of the server, when the server is not also a client
of another service.

For @adriancole

### DIFF
--- a/zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraSpanConsumer.java
+++ b/zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraSpanConsumer.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2016 The OpenZipkin Authors
+ * Copyright 2015-2017 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -32,7 +32,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import zipkin.Annotation;
 import zipkin.BinaryAnnotation;
-import zipkin.Constants;
 import zipkin.Span;
 import zipkin.storage.cassandra3.Schema.AnnotationUDT;
 import zipkin.storage.cassandra3.Schema.BinaryAnnotationUDT;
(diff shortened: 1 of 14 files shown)


codefromthecrypt:
Thanks Gary. Of course this change is welcome for anyone I dont mind
hogging it ;)

codefromthecrypt:
ok this actually breaks for everything except mysql and mem

codefromthecrypt:
in cassandra, this seems to be related to the server not setting timestamp (which it shouldn't)

codefromthecrypt:
nice find.. so far seems legit. I'll probably make sure this works for all or is understood why not.

codefromthecrypt:
@garyd203 @sirtyro I added a commit to fix it

zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/CassandraSpanConsumer.java
```diff
@@ -103,43 +102,29 @@
...
-        if (timestamp != null) {
-          // Contract for Repository.storeTraceServiceSpanName is to store the span twice, once with
-          // the span name and another with empty string.
-          futures.add(storeTraceServiceSpanName(serviceName, span.name, timestamp, span.duration,
-              traceId));
-          if (!span.name.isEmpty()) { // If span.name == "", this would be redundant
-            futures.add(storeTraceServiceSpanName(serviceName, "", timestamp, span.duration, traceId));
-          }
-          futures.add(storeServiceSpanName(serviceName, span.name));
+        if (timestamp == null) continue; // QueryRequest.min/maxDuration needs timestamp
```
codefromthecrypt, line 108: cc @llinder fyi this was just for sanity as indent level was blinding. no logic change

zipkin/src/test/java/zipkin/storage/SpanStoreTest.java
```diff
@@ -933,6 +935,34 @@ public void rawTrace_doesntPerformQueryTimeAdjustment() {
         .containsExactly(earlyTraces);
   }
 
+  /**
+   * Shared server-spans are not supposed to report timestamp, as that interferes with the
+   * authoritative timestamp of the caller. This makes sure that server spans can still be looked up
+   * when they didn't start a span.
+   */
+  @Test
+  public void traceIsSearchableBySRServiceName() throws Exception {
```
codefromthecrypt, line 944: here's the new test

PR #1582, opened 2017-05-03
Enables search by binary annotation aka tag key

codefromthecrypt:
I believe we once were able to search by presence of a tag key. Ex
"http.method" would return any traces that included any http request.
Somewhere along the way this stopped working. This resurrects the
feature along with tests. Here's the impact:

* Cassandra will index a bit more: once per unique service/tag key
* Elasticsearch now does two nested queries when looking for a key
* MySQL now considers all annotation rows when looking for a key

I think the value is greater than the impact as this is intuitive and
answers a functionality many expect to "just work".


codefromthecrypt:
cc @kellabyte @basvanbeek @fedj @garyd203 @sirtyro @semyonslepov

codefromthecrypt:
merging based on thumbs present!

codefromthecrypt:
going out as 1.24

Issue #1219, opened 2016-08-10
Document how to make a custom standalone collector


virtuald:
This involves some spring magic and isn't particularly obvious on how to do it. I did it earlier today, but I'll have to look at it tomorrow to remember what exactly is needed. Here's what I remember so far:

POM modifications:
- Lots of copy/paste from zipkin-server POM
  - Need to add the correct zipkin autoconfigure dependencies to your POM (so for example, if you're writing to cassandra, add zipkin-autoconfigure-storage-cassandra... or just add all of them)
  - Add spring boot configuration parser to POM 
  - Add sprint boot starter to POM

Creating the collector:
- Lots of copy/paste from the kafka collector
  - Two pieces: configuration and stream collector, it's fairly straightforward copy/pasting
  - Need to add a spring autoconfigure thing to your resources

Creating a main function to run:
- Create a main class, see the main zipkin-server class
  - This class should have the sampler `@Bean` in it
- If you want to use environment variables to configure the app like everything else in zipkin, then you need to copy/paste a bunch of stuff from zipkin-server.yml, and add that to your own app's resources

Finally, make sure that you don't mess up the annotation compilation, otherwise you'll get really weird errors and it won't work.

At some point, it might be useful to make a sample project with all of these. 

Execution is however one would normally execute a spring boot application.

codefromthecrypt:
fodder for examples, here's the only custom collector I'm aware of on github https://github.com/spring-cloud/spring-cloud-sleuth/tree/master/spring-cloud-sleuth-zipkin-stream

codefromthecrypt:
another example, is a standalone server with no collector or UI components: https://github.com/openzipkin/zipkin/tree/master/zipkin-server/src/it/minimal-dependencies

codefromthecrypt:
#1233 would be a great example for this. an example of how to make an FQ collector kills two birds with one stone.

codefromthecrypt:
Since last time we discussed how to deal with bootstrapping third parties and now, there's something brewing from @dsyer called https://github.com/dsyer/spring-boot-thin-launcher

The idea is that you could mix and match components this way, for example, adding the core zipkin server with elasticsearch storage and azure collector autoconfiguration deps.

A stable docker image could still be created by doing a "dry run" which just downloads the dependencies listed in the manifest during the build.

What do you all think about trying this? @aliostad @shakuzen @marcingrzejszczak @llinder @devinsba @anuraaga @kevinmdavis @denyska

[...]

PR #1586, opened 2017-05-09
Add a Kafka 0.10+ compatible collector and attempt to integrate into zipkin-server

### LINKED ISSUE #1405: Update Kafka (to 0.10)
Last year, we attempted to update Kafka to 0.9, but had to revert it due to the ordering problem implied.

https://github.com/openzipkin/zipkin/pull/904

Unfortunately no one responded to the email asking about kafka update plans: https://groups.google.com/forum/#!topic/zipkin-user/rQpecbvc6Y0

We shouldn't be pinned to 0.8 forever, and we are now getting requests for 0.10 support. I'm wondering how we should address Kafka updates, particularly as the server includes an all-jar. Unless packaging is different, it might imply weird tricks to be able support multiple versions in the same binary, or a separate build.

Can anyone share their current status wrt kafka and/or have ideas on how to support multiple versions simply?

cc @prat0318 @dgarson @SimenB @sveisvei @eirslett @kristofa @NegatioN @shakuzen

dgrabows:
These changes add a new collector module supporting Kafka 0.10+ (zipkin-collector-kafka10) and the associated auto-configure module (zipkin-autoconfigure-collector-kafka10). They also make the auto-configure module a dependency of zipkin-server, with the collector activated by setting the `KAFKA_BOOSTRAP_SERVERS` environment variable. See https://github.com/dgrabows/zipkin/tree/kafka10-collector/zipkin-server#kafka-010-collector for more details on the supported configuration.

Further changes are needed to how the new collector is integrated with zipkin-server. As currently implemented, the 0.10.2.0 version of the org.apache.kafka:kafka-clients dependency takes precendence over the 0.8.x.x one when zipkin-server is built. This will prevent the existing 0.8 Kafka collector from working. From https://github.com/openzipkin/zipkin/issues/1405#issuecomment-300023053:

> wrt how we address integration, I think the best way is to act like sqs,
where we package a module jar (and this includes a zipkin-server-kafka file
like
https://github.com/openzipkin/zipkin-aws/blob/master/autoconfigure/collector-sqs/src/main/resources/zipkin-server-sqs.yml
)
>
> The pom would include a module file as well, like this, except different
excludes:
https://github.com/openzipkin/zipkin-aws/blob/master/autoconfigure/collector-sqs/pom.xml
>
> Running will be a little messier than this because of classpath
https://github.com/openzipkin/zipkin-aws/tree/master/autoconfigure/collector-sqs#running
>
> For example, we'll probably have to customize the docker image based on
presence of the KAFKA_BOOTSTRAP_SERVERS variable, and move old jars out of
the way.. we may end up lucky enough to not need to do that. Basically,
once there's a module dir the next step is seeing how lucky we are.. in
best case classpath order works and we don't need to hack the all jar to
take out the kafka08 stuff.
>
> In the future, we could make a thin launcher
https://github.com/dsyer/spring-boot-thin-launcher as the default image

Outstanding tasks I'm aware of:
- [x] Add unit test to https://github.com/dgrabows/zipkin/blob/kafka10-collector/zipkin-collector/kafka10/src/test/java/zipkin/collector/kafka10/KafkaCollectorTest.java#L204 for collector running multiple threads.
- [x] Review/update closing of Kafka consumers when consumer thread is interrupted. See https://github.com/dgrabows/zipkin/blob/kafka10-collector/zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaConsumerProcessor.java#L83
- [x] Add zipkin-server-kafka.yml configuration file to https://github.com/dgrabows/zipkin/tree/kafka10-collector/zipkin-autoconfigure/collector-kafka10/src/main/resources
- [x] Update https://github.com/dgrabows/zipkin/blob/kafka10-collector/zipkin-autoconfigure/collector-kafka10/pom.xml to produce a module jar
- [x] Remove zipkin-autoconfigure-collector-kafka10 dependency from https://github.com/dgrabows/zipkin/blob/kafka10-collector/zipkin-server/pom.xml
- [x] Move usage documentation from https://github.com/dgrabows/zipkin/blob/kafka10-collector/zipkin-server/README.md#kafka-010-collector to new README.md in https://github.com/dgrabows/zipkin/tree/kafka10-collector/zipkin-autoconfigure/collector-kafka10 and update to reflect approach of loading the kafka10 collector from an external module jar
- [x] Uncomment https://github.com/dgrabows/zipkin/blob/kafka10-collector/zipkin-autoconfigure/collector-kafka10/src/test/java/zipkin/collector/kafka10/ZipkinKafkaCollectorAutoConfigurationTest.java#L88 and get test passing
- [x] Clean up logging APIs, implementations, and configuration in https://github.com/dgrabows/zipkin/tree/kafka10-collector/zipkin-collector/kafka10

There are probably also changes to https://github.com/openzipkin/docker-zipkin that are not yet clear to me, although that's probably a separate pull request given that it's a different repository.

Attempts to resolve #1405.


dgrabows:
The zipkin-autoconfigure/collector-kafka10 module now builds a module jar, e.g. `zipkin-autoconfigure-collector-kafka10-1.24.1-SNAPSHOT-module.jar`.

Next step is getting the collector to run in zipkin-server again as an external module, based off of the SQS collector example and instructions on https://github.com/openzipkin/zipkin/issues/1219#issuecomment-274451253.

codefromthecrypt:
Thanks for getting this started!

Here are some notes as most looks picture perfect!

* toss the extra logback config in test sources. Add a test dep of a slf4j
to log4j2 bridge if somehow missing
* if slf4j is the implied logging api of kafka, make a comment in the pom
of why we are binding to it. Otherwise switch to whatever kafka10 uses and
similarly make a comment
* you have a commented out test. You should be able to test this by
sneakily using package access to the kafkaConsumers field. Try it!
* move/rework the kafka properties description from zipkin-server to the
autoconfig README. you can look at zipkin-sparkstreaming or zipkin-aws for
exampkes.. mainly we should link in nice bold in the zipkin-server readme
to the right place. (This is different than othe4s but a good time to break
from sprawling)

All above are small things.. good work!

dgrabows:
I made some changes to clean up logging for the unit tests, but I'm not sure if I went in the direction you intended. 

I took your comments to mean there is a preference for binding the logging to a log4j implementation, although that seems to differ from the preference in elsewhere in the project. Modules that reference a logging implementation all seem to depend on logback and https://github.com/openzipkin/zipkin/blob/master/pom.xml defines a logback.version. I can't find any references to log4j other than some dependency exclusions and the unnecessary log4j.properties in test/resources for zipkin-collector-kafka, which I carried over when I created the kafka10 collector.

org.apache.kafka:kafka-clients, at least for 0.10+, implement logging against slf4j-api.

I opted to delete the log4j.properties and do some cleanup on the logback.xml. I can easily go the other direction and swap logback for slf4j, but this seemed more consistent with other modules. Just let me know.

codefromthecrypt:
>
> I made some changes to clean up logging for the unit tests, but I'm not
> sure if I went in the direction you intended.
>
this is fine. I forgot we hadn't switched tests in the main repo to log4j2,
yet. my mistake!

dgrabows:
The kafka10 collector works with zipkin-server as an external module jar. There was no classpath manipulation required. I pretty much just worked once I figured out how to get all the right stuff into `loader.path`. See https://github.com/dgrabows/zipkin/tree/kafka10-collector/zipkin-autoconfigure/collector-kafka10 for instructions on how to run it.

codefromthecrypt:
great. do you want to try and get conditional loading working in the docker
image? There might be a chicken/egg problem as the docker image pulls
release images..

dgrabows:
I spent some time working to update the docker image today. Most of that was spent getting familiar with docker, docker-compose, etc., as I've not done much more than copy/paste execution of containers previously. I'm hoping tomorrow I can get something hacked together and into a PR in docker-zipkin so I can get some early feedback.

StephenWithPH:
@dgrabows let me know if I can do anything to help with the Docker work.

dgrabows:
It took a lot longer than I was hoping, but I've got something working with the docker images.

See https://github.com/dgrabows/docker-zipkin/tree/kafka10-collector and the comments on the commit https://github.com/dgrabows/docker-zipkin/commit/02470624be97b2b5108b91c2b138be96c1a7d160.

I don't expect any of that to be in a final state, but it's a (more or less) working starting point for discussion on how this should work, what to do about upgrading the kafka container image to 0.10, and how to deal with differences between (at least on Macs) Docker Toolbox and Docker for Mac.

I'll try to write up a cohert PR tonight for those changes in docker-zipkin to allow for more detailed feedback.

@StephenWithPH Any feedback on the direction those docker changes are going would be greatly appreciated.

StephenWithPH:
@dgrabows looking forward to your PR in docker-zipkin. In the interest of keeping the discussion contained, I'll add comments there. I think everything is very close. Thank you for taking this on!

dgrabows:
Pull request for the initial attempt a docker changes is now created.

https://github.com/openzipkin/docker-zipkin/pull/142

codefromthecrypt:
Thanks. How do you feel about this PR btw. Anything left?

dgrabows:
That last commit address my final outstanding todo. Assuming the tests pass (they did for me locally) and there are no concerns from anyone else, I think this is good to go.

zipkin-autoconfigure/collector-kafka10/README.md
```diff
@@ -0,0 +1,130 @@
+# Kafka 0.10+ Collector
+
+This collector is implemented as a Kafka consumer supporting Kafka brokers running version 0.10.0.0 
+or later. To activate this collector, reference the module jar when running the Zipkin server 
+and configure one or more bootstrap brokers via the `KAFKA_BOOTSTRAP_SERVERS` environment
+variable or `zipkin.collector.kafka.bootstrap-servers` property.
+
+## Quick start
+
+The minimum version of Java required to run Zipkin server and the Kafka 0.10+ collector is JRE 8.
```
codefromthecrypt: is JRE8 a Kafka 0.10 constraint? If not, I might drop it back to JRE 7 as some use the components directly, and we usually don't use a JRE level higher than the underlying driver.

zipkin-autoconfigure/collector-kafka10/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+Fetch the latest released
```
codefromthecrypt: remember some use this code w/o autoconfigure (as a library). Notes about autoconfigure and modules etc ideally go into the auto-configure module README, since it is produced there. You can look at this for an example:

https://github.com/openzipkin/zipkin-aws/tree/master/collector-sqs
https://github.com/openzipkin/zipkin-aws/tree/master/autoconfigure/collector-sqs

zipkin-autoconfigure/collector-kafka10/pom.xml
```diff
@@ -0,0 +1,64 @@
...
+    the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>io.zipkin.java</groupId>
+    <artifactId>zipkin-autoconfigure</artifactId>
+    <version>1.24.1-SNAPSHOT</version>
```
codefromthecrypt: note: rebase and bump versions as we are now at 1.25.1-SNAPSHOT

zipkin-autoconfigure/collector-kafka10/src/main/java/zipkin/autoconfigure/collector/kafka10/KafkaBootstrapServersSetCondition.java
```diff
@@ -0,0 +1,42 @@
...
+ * non-empty.
+ *
+ * <p>This is here because the yaml defaults this property to empty like this, and spring-boot
+ * doesn't have an option to treat empty properties as unset.
+ *
+ * <pre>{@code
+ * bootstrap-servers: ${KAFKA_BOOTSTRAP_SERVERS:}
+ * }</pre>
+ */
+final class KafkaBootstrapServersSetCondition extends SpringBootCondition {
```
codefromthecrypt: I've started inlining little classes like this as static inner classes on the autoconfigure type. It keeps the header-weight down, and they are pretty insignificant.

Also, for single-property conditions, it is unambiguous what failed, so you can use the slightly less complex normal condition as `@Conditional(KafkaBootstrapServersSet.class)`:

ex.
```java
  static final class HttpLoggingSet implements Condition {
    @Override public boolean matches(ConditionContext condition, AnnotatedTypeMetadata ignored) {
      return !isEmpty(condition.getEnvironment()
          .getProperty("zipkin.storage.elasticsearch.http-logging"));
    }
  }
```

zipkin-autoconfigure/collector-kafka10/src/main/java/zipkin/autoconfigure/collector/kafka10/ZipkinKafkaCollectorAutoConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+@Configuration
+@EnableConfigurationProperties(ZipkinKafkaCollectorProperties.class)
+@Conditional(KafkaBootstrapServersSetCondition.class)
+public class ZipkinKafkaCollectorAutoConfiguration {
+
+  @Bean(initMethod = "start") KafkaCollector kafka(ZipkinKafkaCollectorProperties properties,
+      CollectorSampler sampler, CollectorMetrics metrics, StorageComponent storage) {
+    return  properties.toBuilder().sampler(sampler).metrics(metrics).storage(storage).build();
+  }
```
codefromthecrypt, line 40: ex this is a tiny file, you can add condition type here (noting you'll need to qualify it )

zipkin-autoconfigure/collector-kafka10/src/main/java/zipkin/autoconfigure/collector/kafka10/ZipkinKafkaCollectorProperties.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import zipkin.collector.kafka10.KafkaCollector;
+
+@ConfigurationProperties("zipkin.collector.kafka")
+public class ZipkinKafkaCollectorProperties {
+  private String bootstrapServers;
+  private String groupId = "zipkin";
```
codefromthecrypt: Latest fashion is to provide one-line docs here `/** password used for basic auth. Needed when Shield or X-Pack security is enabled */` and leave all values at null (imply change streams to Integer).

The one-lines shouldn't include javadoc tags like `@see` as this is looked at by spring boot utilities that add a tool tip.

zipkin-autoconfigure/collector-kafka10/src/main/java/zipkin/autoconfigure/collector/kafka10/ZipkinKafkaCollectorProperties.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public Map<String, String> getOverrides() {
+    return overrides;
+  }
+
+  public void setOverrides(Map<String, String> overrides) {
+    this.overrides = overrides;
+  }
+
+  public KafkaCollector.Builder toBuilder() {
```
codefromthecrypt, line 74: this doesn't need to be public as nothing else uses it.

zipkin-autoconfigure/collector-kafka10/src/main/java/zipkin/autoconfigure/collector/kafka10/ZipkinKafkaCollectorProperties.java
```diff
@@ -0,0 +1,77 @@
...
+    return KafkaCollector.builder()
```
codefromthecrypt: here, you can use this form to avoid accidentally overriding defaults (which avoids coordinating defaults in two places)

```
    KafkaCollector.Builder result = KafkaCollector.builder();
    if (topic != null) result.topic(topic);
    if (groupId != null) result.groupId(groupId);
    if (bootstrapServers != null) result.bootstrapServers(bootstrapServers);
   ...
```

zipkin-autoconfigure/collector-kafka10/src/main/java/zipkin/autoconfigure/collector/kafka10/ZipkinKafkaCollectorProperties.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public void setGroupId(String groupId) {
+    this.groupId = groupId;
+  }
+
+  public String getTopic() {
+    return topic;
+  }
+
+  public void setTopic(String topic) {
```
codefromthecrypt, line 54: note: you'll want to coerce empty to null as sometimes people can't set null with properties. Usually I add a static method for this called emptyToNull

zipkin-autoconfigure/collector-kafka10/src/test/java/zipkin/collector/kafka10/ZipkinKafkaCollectorAutoConfigurationTest.java
```diff
@@ -0,0 +1,127 @@
...
+import zipkin.autoconfigure.collector.kafka10.ZipkinKafkaCollectorProperties;
+import zipkin.collector.CollectorMetrics;
+import zipkin.collector.CollectorSampler;
+import zipkin.storage.InMemoryStorage;
+import zipkin.storage.StorageComponent;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.boot.test.util.EnvironmentTestUtils.addEnvironment;
+
+public class ZipkinKafkaCollectorAutoConfigurationTest {
```
codefromthecrypt, line 34: here's a neat way to test properties set properly. you'll still need a test to ensure the object only loads when its signaling property is set https://github.com/openzipkin/zipkin-sparkstreaming/blob/master/autoconfigure/stream-kafka/src/test/java/zipkin/sparkstreaming/autoconfigure/stream/kafka/ZipkinKafkaStreamFactoryPropertiesTest.java

zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaCollector.java
```diff
@@ -0,0 +1,223 @@
...
+     * <p>For example: Only consume spans since you connected by setting the below.
+     * <pre>{@code
+     * Map<String, String> overrides = new LinkedHashMap<>();
+     * overrides.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "latest");
+     * builder.overrides(overrides);
+     * }</pre>
+     *
+     * @see org.apache.kafka.clients.consumer.ConsumerConfig
+     */
+    public final Builder overrides(Map<String, String> overrides) {
```
codefromthecrypt, line 117: while not required, you could consider using autovalue to make your builder https://github.com/openzipkin/zipkin-sparkstreaming/blob/master/stream/kafka/src/main/java/zipkin/sparkstreaming/stream/kafka/KafkaStreamFactory.java

zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaCollector.java
```diff
@@ -0,0 +1,223 @@
...
+      };
+    }
+
+    @Override
+    public void close() {
+      ExecutorService maybeNull = maybeNull();
+      if (maybeNull != null) {
+        maybeNull.shutdownNow();
+        try {
+          maybeNull.awaitTermination(5, TimeUnit.SECONDS);
```
codefromthecrypt: I think we usually wait 1 second by default now.. though it is arbitrary anyway

zipkin-collector/kafka10/src/test/java/zipkin/collector/kafka10/KafkaCollectorTest.java
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    assertThat(kafkaMetrics.spansDropped()).isEqualTo(TRACE.size());
+  }
+
+  @Test
+  public void messagesDistributedAcrossMultipleThreadsSuccessfully() throws Exception {
+    Builder builder = builder("multi_thread", 2);
+
+    // Producing this empty message triggers auto-creation of the topic and gets things "warmed up"
```
codefromthecrypt: maybe pull this into a method named warmUp with this as its javadoc?

zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaCollectorWorker.java
```diff
@@ -0,0 +1,114 @@
...
+final class KafkaCollectorWorker implements Runnable {
+  private static final Logger LOG = LoggerFactory.getLogger(KafkaCollectorWorker.class);
+
+  final Consumer<byte[], byte[]> kafkaConsumer;
+  final Collector collector;
+  final CollectorMetrics metrics;
+  final AtomicReference<List<TopicPartition>> assignedPartitions =
+      new AtomicReference<>(Collections.emptyList());
+
+  KafkaCollectorWorker(Properties kafkaConsumerConfig, String topic, Collector collector,
```
codefromthecrypt: nit: probably looks nicer to just accept the Builder and take values from it (using package protected)

zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaCollectorWorker.java
```diff
@@ -0,0 +1,114 @@
...
+    });
+    this.collector = collector;
+    this.metrics = metrics;
+  }
+
+  /**
+   * @return Kafka topic partitions currently assigned to this worker for processing. Returned
+   * list is not modifiable.
+   */
+  public List<TopicPartition> assignedPartitions() {
```
codefromthecrypt: doesn't need to be public. you can use the package protected field for testing. that lets you move this javadoc to the field itself.

zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaCollectorWorker.java
```diff
@@ -0,0 +1,114 @@
...
+              }
+            }
+          }
+        });
+      }
+    } finally {
+      LOG.info("Kafka consumer polling loop stopped.");
+      LOG.info("Closing Kafka consumer...");
+      // Clear interrupted status of the thread so the Kafka consumer can at least try to shutdown.
+      Thread.interrupted();
```
codefromthecrypt: hmm ok :)

zipkin-server/README.md
```diff
@@ -271,6 +271,12 @@ prefixed system property:
 $ KAFKA_ZOOKEEPER=127.0.0.1:2181 java -Dzipkin.collector.kafka.overrides.auto.offset.reset=largest -jar zipkin.jar
 ```
 
+### Kafka 0.10+ Collector
```
codefromthecrypt: move this to a note under "This collector remains a Kafka 0.8.x consumer, while Zipkin systems update to 0.9+."? I think more people will notice it this way.

codefromthecrypt (approved):
I made mostly style comments, in efforts to make this collector have all the latest practice. That way, the next to copy/paste will have a better starting point.

The code itself is great. Really appreciate the tests. Please coordinate with @fedj and he can merge this when ready. Then, this will be published into the jfrog snapshots repository. Once that's in place, you can adjust your pending docker PR to download from there, which makes things more realistic. Once that's all together, coordinate w/ @fedj for a release of zipkin. After the zipkin release, you can adjust the docker PR to work against the release repository. Once that's in, it can be merged and released (or the version can be re-released). Please let me know any questions on this

zipkin-autoconfigure/collector-kafka10/README.md
dgrabows: This is the autoconfigure readme. :)

I can change the title of the page to make that clearer.

zipkin-autoconfigure/collector-kafka10/README.md
dgrabows: JRE7 is the minimum requirement for using kafka-clients, including version 0.10. I copied the JRE8 requirement from the zipkin-server readme.

There's no reason collector-kafka10 can't be JRE 7 compatible. I'll check the code to see if there's anything Java 8 specific (other than labmdas, which I'm assuming retrolambda will take care of).

zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaCollectorWorker.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  @Override
+  public void run() {
+    try {
+      LOG.info("Kafka consumer starting polling loop.");
+      while (true) {
+        final ConsumerRecords<byte[], byte[]> consumerRecords = kafkaConsumer.poll(1000);
+        LOG.debug("Kafka polling returned batch of {} messages.", consumerRecords.count());
+        consumerRecords.forEach((cr) -> {
```
dgrabows: I think I'll need to replace the forEach() call here to maintain Java 7 compatibility.

zipkin-autoconfigure/collector-kafka10/README.md
dgrabows: There is one use of forEach() in KafkaCollectorWorker that needs to be replaced (see comment further down).

codefromthecrypt:
>
> I think I'll need to replace the forEach() call here to maintain Java 7
> compatibility.
>
cool. I think one for loop is reasonable. If it had higher impact, I'd
recommend retrolambda

zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaCollectorWorker.java
dgrabows: I went back and forth on this a few times. It seems awkard. But, without it, the Kafka consumers will rarely get closed gracefully, because a thread interrupt is the primary reason the polling loop will be terminated and the kafka consumer will throw an exception if the thread has the interrupted status set.

Closing the consumer isn't strictly necessary, although it can reduce the delay before the consumer group rebalances to account for the consumer being gone.

Your reaction is enough to tip me towards taking it out, since it was a 50/50 decision to begin with. I will do that.

zipkin-autoconfigure/collector-kafka10/src/main/java/zipkin/autoconfigure/collector/kafka10/ZipkinKafkaCollectorAutoConfiguration.java
dgrabows: Done.

zipkin-autoconfigure/collector-kafka10/src/main/java/zipkin/autoconfigure/collector/kafka10/ZipkinKafkaCollectorProperties.java
dgrabows: Done.

zipkin-autoconfigure/collector-kafka10/src/main/java/zipkin/autoconfigure/collector/kafka10/ZipkinKafkaCollectorProperties.java
dgrabows: I ended up leaving this public, because ZipkinKafkaCollectorPropertiesOverrideTest needed access to both this method and the package-private fields on KafkaCollector.Builder. This seemed better than adding public getters for each of those fields.

zipkin-autoconfigure/collector-kafka10/src/test/java/zipkin/collector/kafka10/ZipkinKafkaCollectorAutoConfigurationTest.java
dgrabows: Done in newly added ZipkinKafkaCollectorPropertiesOverrideTest.

zipkin-autoconfigure/collector-kafka10/README.md
dgrabows: I updated this to clarify that zipkin-server and the kafka10 collector have difference java/JRE requirements.

zipkin-autoconfigure/collector-kafka10/README.md
dgrabows: I did change the page title and make some other edits to this readme and the readme for the collector module to try and clarify this.

zipkin-autoconfigure/collector-kafka10/src/main/java/zipkin/autoconfigure/collector/kafka10/KafkaBootstrapServersSetCondition.java
dgrabows: Done.

zipkin-autoconfigure/collector-kafka10/src/main/java/zipkin/autoconfigure/collector/kafka10/ZipkinKafkaCollectorProperties.java
dgrabows: Done.

zipkin-autoconfigure/collector-kafka10/src/main/java/zipkin/autoconfigure/collector/kafka10/ZipkinKafkaCollectorProperties.java
dgrabows: Done.

zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaCollector.java
dgrabows: I switched this to wait 1 second.

zipkin-collector/kafka10/src/test/java/zipkin/collector/kafka10/KafkaCollectorTest.java
dgrabows: Done.

zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaCollectorWorker.java
dgrabows: Done.

zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaCollectorWorker.java
dgrabows: Moved the comment and removed this method.

zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaCollectorWorker.java
dgrabows: The call to Thread.interrupted() is gone now.

zipkin-server/README.md
dgrabows: Done.

zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaCollectorWorker.java
dgrabows: Done.

zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaCollector.java
dgrabows: I ran out of steam before getting to this. If anybody feels strongly about using autovalue here, let me know and I will circle back around to it.

codefromthecrypt:
>
> I ran out of steam before getting to this. If anybody feels strongly about
> using autovalue here, let me know and I will circle back around to it.
>
not important!

zipkin-autoconfigure/collector-kafka10/pom.xml
dgrabows: I've rebased the kafka10-collector branch in my local repo and committed the version updates to these two pom files. It won't push cleanly, because my local branch and the one on github have diverged. I suspect I can force the push to resolve this (`git push -f origin +kafka10-collector'), but I'm wary of doing something that rewrites history and screws up the PR.

Am I heading in the right direction, or did I not execute the rebase properly?

Current state of affairs on my local clone of the repo:
```dgrabowski:zipkin$ git status
On branch kafka10-collector
Your branch and 'origin/kafka10-collector' have diverged,
and have 37 and 25 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)
nothing to commit, working tree clean
dgrabowski:zipkin$ git push
To github.com:dgrabows/zipkin.git
 ! [rejected]        kafka10-collector -> kafka10-collector (non-fast-forward)
error: failed to push some refs to 'git@github.com:dgrabows/zipkin.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
dgrabowski:zipkin$```

codefromthecrypt:
I don't think anyone else is using your branch, so you can force push it.
The commits will be squashed on merge anyway.

dgrabows:
Alright. Rebase and version catch-up are complete. Assuming the tests pass like they did locally, I think that's all the outstanding items.

codefromthecrypt:
ok will merge on green, which would then put it into the snapshot repo

codefromthecrypt:
Grand work, btw 👍

dgrabows:
Thank you.

PR #1589, opened 2017-05-11
Updates zipkin self-tracing to Brave 4

codefromthecrypt:


codefromthecrypt:
there might be a propagation bug in brave-cassandra, as this trace looks like it has spans for other traces inside..

screen shot 2017-06-29 at 2 58 10 pm

started via `STORAGE_TYPE=cassandra3 SELF_TRACING_ENABLED=true java -jar ./zipkin-server/target/zipkin-server-*exec.jar` cc @llinder

flipdiving:
The information you share is very useful. It is closely related to my work and has helped me grow. Thank you!
[lucky patcher](http://luckypatcher-apks.com) , 
[color switch 2](http://colorswitch2.com)

PR #1593, opened 2017-05-15
Allows collector-only servers to disable the UI and query apis

### LINKED ISSUE #1591: Make it possible to disable the query api for collector-only servers
We are deploying a zipkin server to a high-security production environment, and wish to disable the query API (ZipkinQueryApiV1) in order to minimise access to details of our system architecture.

codefromthecrypt:
This adds the ability to set `QUERY_ENABLED=false` for reasons including
security policy.

Fixes #1591

### DIFF
--- a/zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiAutoConfiguration.java
+++ b/zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiAutoConfiguration.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2016 The OpenZipkin Authors
+ * Copyright 2015-2017 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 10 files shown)


zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiAutoConfiguration.java
```diff
@@ -59,6 +60,7 @@
  */
 @Configuration
 @EnableConfigurationProperties(ZipkinUiProperties.class)
+@ConditionalOnProperty(value = "zipkin.ui.enabled", havingValue = "true", matchIfMissing = true)
```
shakuzen: Nits: `name` might be more clear than `value` (which is an alias for `name`). Maybe it is just me but `value = x, havingValue = y` reads kind of funny. Also, the default of `havingValue` should be sufficient for our purposes. Specifying `"true"` might mean something like `"TRUE"` doesn't work.

zipkin-server/README.md
```diff
@@ -110,6 +110,7 @@ zipkin-server is a drop-in replacement for the [scala query service](https://git
 [yaml configuration](src/main/resources/zipkin-server-shared.yml) binds the following environment variables from zipkin-scala:
 
     * `QUERY_PORT`: Listen port for the http api and web ui; Defaults to 9411
+    * `QUERY_ENABLED`: False disables the query api and UI assets; Defaults to true
```
shakuzen: If we do remove the `havingValue` parameter on the `@ConditionalOnProperty` sites, perhaps writing `false` (formatted) makes it clearer we're looking for that specifically to disable querying. With the `havingValue` anything but `true` will disable, I believe.

shakuzen (commented):
Looks good as is; just mentioned a couple nits.

codefromthecrypt:
polished up thx to @shakuzen for the advice

shakuzen (approved):

ghost:
Nice guys! That's pretty much how I would have done it :dancer:

PR #1603, opened 2017-06-05
Add .editorconfig and make sure we use LF (not CRLF) for line endings

eirslett:
IntelliJ complained a lot about CRLF in the logback.xml file. And besides, it's nice to have a common convention. `.editorconfig` will enforce it (at least in IDEs)


codefromthecrypt:
Thanks @eirslett!

PR #1609, opened 2017-06-10
Use upstream Prometheus client and Spring integration

abesto:
This change replaces our custom logic for exposing metrics with the implementation in the upstream Prometheus client, as noted in https://github.com/openzipkin/zipkin/pull/1144#issuecomment-288762140. The notable improvement (other than less code) is the addition of histogram data of response times, using a servlet filter (the data looks something like this: https://gist.github.com/abesto/46aeb18ab45e5126ccebc515eb3fc99f)

Metrics comparison: [before](https://gist.github.com/abesto/25594dc217e70ad9b9bddd656eb663d4) / [after](https://gist.github.com/abesto/0e9afff23daacb0f4301c9b977012ad5)

There were two issues that blocked us from merging this for a while:

 * [x] We're hit by https://github.com/prometheus/client_java/issues/252 (pops up only when Prometheus tries to query the data, you can still check out `/prometheus` from a browser just fine)
 * [x] The upstream integration with Spring Boot Actuator metrics doesn't use the `counter.` and `gauge.` prefixes to figure out the right metric type, always defaults to `gauge`. https://github.com/prometheus/client_java/pull/254 aims to fix this.


abesto:
The Spring counter metrics don't become, and shouldn't become Prometheus counter metrics, since in Spring Boot, counters can decrease. Current practice is to just make metrics like this `gauge`s in Prometheus (https://github.com/prometheus/client_java/pull/254#issuecomment-307566369).

We should also see more JVM metrics exposed, in a way more  Prometheus friendly way, letting up improve the dashboard further.

zipkin-autoconfigure/metrics-prometheus/pom.xml
```diff
@@ -52,11 +52,29 @@
...
 
-        <!-- Test dependencies -->
+        <!-- Prometheus client -->
         <dependency>
-            <groupId>org.springframework</groupId>
-            <artifactId>spring-test</artifactId>
-            <scope>test</scope>
+          <groupId>io.prometheus</groupId>
+          <artifactId>simpleclient</artifactId>
+          <version>0.0.23</version>
```
codefromthecrypt: seems ok to have an unstable dep here because the server-side dependency is optional (which makes it a part of the all-jar, but not a strict dep for customizers)

codefromthecrypt:
might be worth making some sample traffic against this to ensure that collector metrics show up (they are probably the most important metric). ex run one of the example projects like https://github.com/openzipkin/pyramid_zipkin-example

zipkin-autoconfigure/metrics-prometheus/pom.xml
abesto: Sorry, I don't speak fluent Maven. By unstable dep, do you mean we shouldn't pin the version of io.prometheus:simpleclient? Or that I shouldn't remove org.springframework:spring-test?

abesto:
The upstream client exposes all the Spring Boot Actuator metrics, in almost exactly the same way as the previous Zipkin implementation. Still, due diligence is due, I'll check.

zipkin-autoconfigure/metrics-prometheus/pom.xml
codefromthecrypt: unstable I meant the patch-only version number :)

nit: maybe use a property to coordinate these? ex.

```
in the properties section
<simpleclient.version>0.0.23</simpleclient.version>

then in places like here:
${simpleclient.version}

```

abesto: Ah, I see.

Yep, will set up a property for the version number.

abesto:
Metrics after sending a few spans using `pyramid_zipkin-example`: https://gist.github.com/abesto/93bd70bf907d16d3b7ef4e408408f69f

Methinks we're fine:

```
# HELP counter_zipkin_collector_messages_http counter_zipkin_collector_messages_http
# TYPE counter_zipkin_collector_messages_http gauge
counter_zipkin_collector_messages_http 13.0
```

> they are probably the most important metric

Oh! Then I guess I should add them to the Grafana dashboard.

codefromthecrypt:
>
> they are probably the most important metric
>
> Oh! Then I guess I should add them to the Grafana dashboard!
>
yes I had a nag in the back of my head to nag you on the other issue. By
naming convention substitute http for kafka (sqs scribe) etc as not
everyone sends spans via http transport.

messages, spans and bytes are interesting metrics, especially if
coordinated client-side. drop metrics are something that one would alert on
at some point.

abesto:
Important note, just realized: this change does actually break the currently used metric names. The Zipkin implementation trims the `counter_` and `gauge_` prefixes (and sets the Prometheus metric type accordingly), while the upstream one does _not_ trim those prefixes, and makes everything a gauge in Prometheus (see https://github.com/openzipkin/zipkin/pull/1609#issuecomment-307608094).

Assumption: not a lot of people are currently using these metrics. Proposal: let's drop as much custom logic as we can, and break the metric names now, before publishing the dashboard using the metric names.

codefromthecrypt:
>
> Assumption: not a lot of people are currently using these metrics.
> Proposal: let's drop as much custom logic as we can, and break the metric
> names now, before publishing the dashboard using the metric names.
>
I'd look at changelog for the directory and ping any implicated. After that
go for it

abesto:
@klette Having implemented the original Zipkin Prometheus exporter, what are your thoughts on this PR? Especially https://github.com/openzipkin/zipkin/pull/1609#issuecomment-307740186?

(That exhausts the list of people who contributed to the Prometheus exporter :smile: )

zipkin-autoconfigure/metrics-prometheus/pom.xml
```diff
@@ -27,6 +27,7 @@
 
     <properties>
         <main.basedir>${project.basedir}/../..</main.basedir>
+        <prometheus_simpleclient.version>0.0.23</prometheus_simpleclient.version>
```
codefromthecrypt: the fix to the aforementioned bug is in master now, so guessing 0.0.24 will sort it out

codefromthecrypt: 0.0.26 is out, so probably time to see if we're sorted?

abesto: 😨 Sad response time is sad. Yep, I'll check it out today.

abesto: Confirmed, Prometheus can successfully scrape data with `0.0.26`. Pushed a commit; I'll start updating the Grafana dashboard for the new metric names, upload once this is merged, then we can hopefully also merge https://github.com/openzipkin/docker-zipkin/pull/135

codefromthecrypt:
@klette nagtime!

codefromthecrypt (approved):
Will merge after testing against dashboard

codefromthecrypt:
gracias!

Issue #1512, opened 2017-02-02
Maven license plugin fails in travis


codefromthecrypt:
After investigation, it appears that the maven license plugin is failing due to travis having incorrect dates on files.

Ex I added this into the .travis.yml
```
install:
  - date
  - git --version
  - git log -n 1 ./zipkin/src/test/java/zipkin/internal/DependencyLinkerTest.java
```

then the output for a file was this..

```
$ date
Thu Feb  2 04:13:46 UTC 2017
install.2
0.03s$ git --version
git version 2.11.0
install.3
0.02s$ git log -n 1 ./zipkin/src/test/java/zipkin/internal/DependencyLinkerTest.java
commit ffac4ff4eac53b82dea2f66179e3be920bfa6276
Author: Adrian Cole <adriancole@users.noreply.github.com>
Date:   Tue Jan 3 09:28:48 2017 +0800
```

you can see that it says the file was modified on Jan 3 this year..

however, it wasn't.. it was modified last year.

```bash
$ git log -n 1 ./zipkin/src/test/java/zipkin/internal/DependencyLinkerTest.java
commit 21fc32b965e70f0985a164f34ed89ae55e115a11
Author: Adrian Cole <adriancole@users.noreply.github.com>
Date:   Thu Oct 27 09:38:34 2016 +0800
```

There is a current travis issue which might be related https://www.traviscistatus.com/incidents/k79mjcv403c4

In the mean time, I'll try to disable the license plugin temporarily so that we can get a green build (and release code)

codefromthecrypt:
thanks @shakuzen for the help thinking this through. We may want to consider moving off travis at some point, eventhough there's a bottleneck in circleci (max workers per org)

cc @openzipkin/devops-tooling @bsideup

codefromthecrypt:
ps in case not clear, this does not impact our circleci pull request checkers in any way. license check works fine

codefromthecrypt:
0a4ef9566459c06c9c6f705f076461462378693d 6a0d6e7a00ddcd0ff4cb73458adc708838f995c5 and a5a84ede544004d02546a5535f48c589d05ac280 can be reverted once whatever's going on in travis slaves is over

bsideup:
@adriancole Travis is really nice when it comes to Docker testing; I wouldn't recommend to drop it entirely. 

I can't find any reported issue, so Travis team might not be aware of it.

codefromthecrypt:
[...]

PR #1610, opened 2017-06-10
./mvnw license:format

abesto:
@guozhaoyun failed to build from source locally:

```
[WARNING] Missing header in: /home/usr/zk/circleci/test.sh
[WARNING] Missing header in: /home/usr/zk/travis/publish.sh
[WARNING] Missing header in: /home/usr/zk/circleci/pom-no-crossmodule-dependencies.xsl
```
I can't seem to reproduce that specific failure, not to mention the license check should be failing on all CI builds. Still, it's a good opportunity to update the headers. I'm not sure I understand how things have not been failing with the headers missing from most of `zipkin-ui`, but adding them seems like a good idea to me.


codefromthecrypt:
the mystery of why CI is skipping license check is here: #1512

abesto:
The thick plottens! Closing this then, it's not solving the right problem.

PR #1612, opened 2017-06-13
Add zipkin banner

making:
It would be nice to show Zipkin banner instead of Spring Boot

image


codefromthecrypt:
satisfies the rule-of-three. merge time!

PR #1615, opened 2017-06-13
Allow query-only servers to disable the HTTP collector

dgrabows:
We have a Zipkin deployment where there is one instance of zipkin-server providing the query API and a second instance running the (kafka in our case) collector. The zipkin-server instance providing the query API will be exposed to HTTP traffic from our internal network and it is preferable if the exposed HTTP API is read-only to prevent users from creating span data. Disabling the HTTP collector accomplishes this.

The motivation for this change is similar to the rationale described on #1591, and the approach taken to implement this is very similar to the one from the PR for that issue (#1593).

These changes add a configuration point for enabling or disabling the HTTP collector in zipkin-server. This can be controlled through the property zipkin.collector.http.enabled or the corresponding environment variable HTTP_COLLECTOR_ENABLED. The collector is enabled by default.

### DIFF
--- a/zipkin-server/README.md
+++ b/zipkin-server/README.md
@@ -115,6 +115,7 @@ zipkin-server is a drop-in replacement for the [scala query service](https://git
     * `QUERY_LOOKBACK`: How many milliseconds queries can look back from endTs; Defaults to 24 hours (two daily buckets: one for today and one for yesterday)
     * `STORAGE_TYPE`: SpanStore implementation: one of `mem`, `mysql`, `cassandra`, `elasticsearch`
     * `COLLECTOR_SAMPLE_RATE`: Percentage of traces to retain, defaults to always sample (1.0).
+    * `HTTP_COLLECTOR_ENABLED`: `false` disables the http collector; Defaults to true
 
 ### Cassandra Storage
 Zipkin's [Cassandra storage component](../zipkin-storage/cassandra)
--- a/zipkin-server/README.md
+++ b/zipkin-server/README.md
@@ -115,7 +115,6 @@ zipkin-server is a drop-in replacement for the [scala query service](https://git
     * `QUERY_LOOKBACK`: How many milliseconds queries can look back from endTs; Defaults to 24 hours (two daily buckets: one for today and one for yesterday)
     * `STORAGE_TYPE`: SpanStore implementation: one of `mem`, `mysql`, `cassandra`, `elasticsearch`
     * `COLLECTOR_SAMPLE_RATE`: Percentage of traces to retain, defaults to always sample (1.0).
-    * `HTTP_COLLECTOR_ENABLED`: `false` disables the http collector; Defaults to true
 
 ### Cassandra Storage
 Zipkin's [Cassandra storage component](../zipkin-storage/cassandra)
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
Sounds reasonable. cc @sityro

codefromthecrypt:
sorry meant to cc @sirtyro

zipkin-server/README.md
```diff
@@ -115,6 +115,7 @@ zipkin-server is a drop-in replacement for the [scala query service](https://git
     * `QUERY_LOOKBACK`: How many milliseconds queries can look back from endTs; Defaults to 24 hours (two daily buckets: one for today and one for yesterday)
     * `STORAGE_TYPE`: SpanStore implementation: one of `mem`, `mysql`, `cassandra`, `elasticsearch`
     * `COLLECTOR_SAMPLE_RATE`: Percentage of traces to retain, defaults to always sample (1.0).
+    * `HTTP_COLLECTOR_ENABLED`: `false` disables the http collector; Defaults to true
```
codefromthecrypt: adding QUERY_ENABLED to this section was probably a bit of a mistake as that env never existed before. How about adding a new section based on `Scribe Collector`? In the new section, I'd explain this accepts spans via `POST /api/v1/spans`

I feel awkward about `HTTP_COLLECTOR_ENABLED` as much as `HTTP_ENABLED` almost to the point of punting re-mapping of the spring property. That said, it isn't a strong feeling.

codefromthecrypt (commented):
just some thoughts on readme otherwise LG

zipkin-server/README.md
dgrabows: I made some changes to the readme to create a section for the HTTP collector.

zipkin-server/README.md
```diff
@@ -224,6 +224,14 @@ to prevent excessive load, service and span name queries are limited by
 `QUERY_LOOKBACK`, which defaults to 24hrs (two daily buckets: one for
 today and one for yesterday)
 
+### HTTP Collector
+The HTTP collector is enabled by default. It accepts spans via `POST /api/v1/spans`. The HTTP 
+collector supports the following configuration:
+
+Property | Environment Variable | Description
+--- | --- | ---
+`zipkin.collector.http.enabled` | `HTTP_COLLECTOR_ENABLED` | `false` disables the HTTP collector. Defaults to `true`.
```
dgrabows, line 233: As far as the naming or existence of the environment variable, I'm happy to leave it in or take it out (of the docs and [zipkin-server-shared.yml](https://github.com/dgrabows/zipkin/blob/disable-http-collector/zipkin-server/src/main/resources/zipkin-server-shared.yml)). The one argument I can think of in favor of keeping it is support for disabling the HTTP collector for users of the openzipkin/zipkin docker image. Either way won't affect my particular situation, because I am already configuring other aspects of zipkin-server via Java system properties.

codefromthecrypt:
Looks good. On the env part, spring boot already avails upper underscore
env for correspinding properties. Here and elsewhere we are just making
shorter ones available. At any rate, I think this is good. Thanks!

codefromthecrypt:
thank you!

ghost:
Sorry @adriancole , late to the party, but we like this and will def use it!

PR #1616, opened 2017-06-15
Accepts json span names (previously broke ElasticSearch)

### LINKED ISSUE #1613: ES indexing breaks when a span name is json
After testing I found only ES5.1.2 can, other high version simply can not be integrated. In other high-profile environment error.

codefromthecrypt:
While not good practice, if we break on json span names, users can't
easily troubleshoot. This works around the problem by escaping json.

Fixes #1613


codefromthecrypt:
here's a follow-up on the overall problem of span data https://github.com/openzipkin/zipkin/issues/1617

PR #1618, opened 2017-06-15
Updates versions notably Spring Boot, MariaDB and ErrorProne

codefromthecrypt:
ErrorProne caught a small bug in cassandra3. MariaDB bump caught a
missing dialect.


pom.xml
```diff
@@ -45,46 +45,46 @@
...
     <main.basedir>${project.basedir}</main.basedir>
 
-    <brave.version>4.1.1</brave.version>
+    <brave.version>4.3.3</brave.version>
     <cassandra-driver-core.version>3.2.0</cassandra-driver-core.version>
     <okio.version>1.13.0</okio.version>
-    <!-- important to keep this with spring-boot -->
-    <jooq.version>3.9.1</jooq.version>
-    <spring-boot.version>1.5.3.RELEASE</spring-boot.version>
+    <!-- important to keep this in sync with spring-boot-dependencies -->
```
shakuzen, line 51: Is there a reason we don't use the `spring-boot-dependencies` BOM (or Spring IO Platform BOM) for dependency management if we want to keep versions in sync?

codefromthecrypt: I suppose trying to be aware, but not limited by what spring-boot-dependencies says. For example, we'll soon get to a point where people may be using us for multiple versions of boot, so not sure I'd want to implicitly take what's in one version.

codefromthecrypt:
gonna merge and see if this helps with aws aurora

PR #1622, opened 2017-06-19
Adding ability to listen to multiple kafka10 topics via comma separated list

danielkwinsor:
Useful for cases like span info from different environments coming from different topics, etc.

### DIFF
--- a/zipkin-autoconfigure/collector-kafka10/README.md
+++ b/zipkin-autoconfigure/collector-kafka10/README.md
@@ -71,7 +71,7 @@ Environment Variable | Property | New Consumer Config | Description
 --- | --- | --- | ---
 `KAFKA_BOOTSTRAP_SERVERS` | `zipkin.collector.kafka.bootstrap-servers` | bootstrap.servers | Comma-separated list of brokers, ex. 127.0.0.1:9092. No default
 `KAFKA_GROUP_ID` | `zipkin.collector.kafka.group-id` | group.id | The consumer group this process is consuming on behalf of. Defaults to `zipkin`
-`KAFKA_TOPIC` | `zipkin.collector.kafka.topic` | N/A | Topic zipkin spans will be consumed from. Defaults to `zipkin`
+`KAFKA_TOPIC` | `zipkin.collector.kafka.topic` | N/A | Comma-separated list of topics that zipkin spans will be consumed from. Defaults to `zipkin`
 `KAFKA_STREAMS` | `zipkin.collector.kafka.streams` | N/A | Count of threads consuming the topic. Defaults to `1`
 
 ### Other Kafka consumer properties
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
seems handy, but would need a test and updates on the config docs. Also, we should make sure that this is something others agree with as I don't know the support implications of always using ConsumerRebalanceListener.

Sorry we don't have a kafka mailer at the moment. pinging @dgrabows as he touched this recently

dgrabows:
This change make sense to me.

I don't think the ConsumerRebalanceListener is a concern. That was already there and not introduced with this change. It gets used for the worker to maintain a list of currently assigned partitions. That list is only used in the tests for multiple consumer threads, where the test needs to wait until each consumer has some assignment before proceeding.

The listener will be invoked each time partitions are reassigned in the consumer group, which will only happen once per restart if running a single instance of the collector. If running multiple collectors, invocations will occur whenever a collector is started, stopped, or looses connectivity to the kafka brokers. Those should be relatively infrequent events.

codefromthecrypt:
ok great. @danielkwinsor mind adding a test? (can be as simple as checking that split worked)

lemme know if you need help with this

danielkwinsor:
I was figuring a request for tests would come :)  I definitely will do it when I can fit it between my heavy work queue.  My only concern is that area didn't look easy to mock, so I don't know when I'll be able to get to it.

Manual testing did show this worked out with single or multiple topics, but I didn't test things like whitespace, etc.

codefromthecrypt:
looks like you might be able to write a simple test that uses package access to check kafkaConsumer.subscription()

basically read-back that to show a simple split works. wanna try?

codefromthecrypt:
k very least this needs an update to the docs KafkaCollector.Builder and in zipkin-server/README like..

```
    * `KAFKA_TOPIC`: Topic zipkin spans will be consumed from. Defaults to "zipkin"
      * When Kafka 0.10 is in use, multiple topics may be specified if comma delimited
```

codefromthecrypt:
All good or did I scare you off? :)

danielkwinsor:
All good... in the process of adding docs and tests, but rebased from
upstream and it wiped out my commit, then auto closed.  Expect by EoD.

danielkwinsor:
I spent too long on the unit test and gave up.  It's much too hard to mock something coming from a constructor without powermockito.  Here's the code plus docs.

zipkin-autoconfigure/collector-kafka10/README.md
```diff
@@ -107,14 +107,14 @@ $ KAFKA_BOOTSTRAP_SERVERS=broker1:9092.local,broker2.local:9092 \
...
-Alternate topic name:
+Alternate topic name(s):
 
 ```bash
 $ KAFKA_BOOTSTRAP_SERVERS=127.0.0.1:9092 \
     java \
     -Dloader.path='zipkin-autoconfigure-collector-kafka10-module.jar,zipkin-autoconfigure-collector-kafka10-module.jar!/lib' \
     -Dspring.profiles.active=kafka \
-    -Dzipkin.collector.kafka.topic=zapkin \
+    -Dzipkin.collector.kafka.topic=zapkin,zipken \
```
codefromthecrypt, line 117: hehe

codefromthecrypt:
Thanks very much @danielkwinsor! (and thanks for the timeboxed attempt on tests.. I'll copy you on a test commit)

PR #1624, opened 2017-06-20
InMemorySpanStore.clear() should clear all its data

joel-airspring:
This is a minor change, but it seems like InMemorySpanStore.clear() should clear all the data it has accumulated.

This affects no tests because all tests get a new instantiation of InMemorySpanStore.  I'd be happy to add a test for the clear if it seems worthwhile.

All tests passed as reported by:  ./mvnw --also-make -pl zipkin-server clean install


codefromthecrypt:
Thanks for this. umm we currently only use this in tests, so we can go with as-is

codefromthecrypt:
going out as 1.26.2

PR #1625, opened 2017-06-20
Escape HTML input

### LINKED ISSUE #1623: XSS vulnerabilities
There exist some XSS vulnerabilities within zipkin and its UI. Either through an injection attack via tags/annotations or by appending injection attack data to the end of the URL, we were able to successfully get the attack to fire when viewing the details of a span. We were only able to get it to happen in the span detail modal window, so it looks as though the data needs sanitising before rendering.

The tag/annotation injection was done using brave, as per the below example:

`    span.annotate("<script>alert(123)</script>");`
`    span.annotate("Annotation 2</td><script>alert(321)</script><td>");`
`    span.annotate("");`
`    span.tag("Tag key 1</td><script>alert('tag key 1')</script><td>", "<script>alert(\"tag value 1\")</script>");`
`    span.tag("Tag key 2</td><!--", "\"quotes are great two\"");`
`    span.tag("", "\"quotes are great two\"");`

Some examples of how this effects the UI is as follows. Please excuse the sanitised screenshots.

Clicking on a span that has been injected by tag/annotation

How the tag/annotation injected span detail looks:

How the tag/annotation injected span script is in the DOM:

Injection attack by URL

Clicking on a span that has had injection attack by URL:

How injection attack by URL is rendered in DOM:

eirslett:
This should fix #1623.

Actually I haven't tried running this patch myself, I just did it in GitHub's in-browser code editor. But I think it works.


ghost:
Improved but not totally fixed. When I click on a span I get two pop ups. Its less than I was getting before. The detail screen looks better now and you can tell that a few more things are sanitised:
improved

codefromthecrypt:
ok there was one miss which was about binary annotation key. no more popups

screen shot 2017-06-21 at 1 24 57 pm

codefromthecrypt:
@sirtyro can you give the OK on this? if works, I'll monkey in some tests

eirslett:
@adriancole can you pick up this PR from here?

codefromthecrypt:
ayup!

ghost:
@adriancole @eirslett , nice work. No more pop ups! :D :+1:

ghost:
:shipit: , straight to prod ;)

codefromthecrypt:
fixed tests and made sure json literals still work

codefromthecrypt:
going out as 1.26.2

PR #1626, opened 2017-06-22
Add migration documentation from kafka 0.8 to 0.10

fedj:
When migrating Kafka from 0.8.1 to 0.10, users should pay attention to the change in offset management.


codefromthecrypt:
thanks!

Issue #1452, opened 2016-12-22
Hosting zipkin UI through a proxy


raiRaiyan:
The zipkin ui that is hosted when a java zipkin server is launched is accessible only through the hostname of the machine where the ui is hosted.  
Like `http://localhost:9411/` 

If the request to the ui is proxied through a gateway, the UI becomes inaccessible.  
Like `http://abcd.com/zipkin/`  

The problem seems to be the `app.js` bundle file for the ui is not requested relatively. The request is something like `http://abcd.com/app.js/`, whereas this request should also be proxied `http://abcd.com/zipkin/app.js`

Looking into it, I figured that `publicPath` in `webpack.config` file is specified as `/`, which makes the requests non relative.  

Even after changing that, all the jquery requests that the ui makes to retreive the data is not proxied.
Can this be changed, maybe accept a `basePath` property in `config.json` of the ui, which will allow the ui to be proxied? Or is there already a way to do this?

codefromthecrypt:
assumption here is that you cannot deploy the UI assets to the proxy? ex
https://github.com/openzipkin/docker-zipkin/tree/master/zipkin-ui

raiRaiyan:
@adriancole The ui itself is not accessible when proxied.  The index.html requests for a file `/app.js` non relatively.

codefromthecrypt:
oh right.. you are talking about proxying, and also hosting under a different path.

can you please find the existing issue on this and close this as a dupe?

eirslett:
Ok, here is the solution we can implement (it was discussed before) Always serve the zipkin ui under a context path, like /zipkin/. Any requests that don't start with that URL will automatically be forwarded. That means there will be no special handling when using a proxy.

codefromthecrypt:
@eirslett seems right, since iirc most are asking for that specific path anyway :) We'd have to whitelist quite a few things, like the api config health checks, etc. Maybe we forward the root itself or is a whitelist ok?

eirslett:
/api/ can be duplicate-served under /zipkin/api/ for a while, I guess not all http clients people use handle 302? But all browsers will.

codefromthecrypt:
@eirslett what if we did it the other way.. we change the content to use paths of /zipkin, and mount the UI assets both under /zipkin and also under root. Might be naive, but could that solve the migration problem as well? cc @abesto

eirslett:
The problem is with crossroads, which doesn't know what to serve (which route) if it doesn't know about the context path.

[...]

PR #1627, opened 2017-06-23
Moves the zipkin UI under the path /zipkin

### LINKED ISSUE #1229: Investigate custom context root
_From @adriancole on May 21, 2016 13:47_

https://github.com/openzipkin/docker-zipkin/issues/97 was opened about moving all zipkin things under the context root /zipkin  (directly vs via a reverse proxy).

This might be possible with OOB spring boot. If so, we could consider this. Otherwise we can close it with explanation in case the question comes up again.

_Copied from original issue: openzipkin/zipkin-java#244_

codefromthecrypt:
This moves the zipkin UI under the path /zipkin, adding a redirect for
/ as well. As UI assets look for the api to be relative to it, requests
to /zipkin/api are forwarded to /api, too.

Fixes #1229 #1452

### DIFF
--- a/zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiAutoConfiguration.java
+++ b/zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiAutoConfiguration.java
@@ -18,7 +18,6 @@ import javax.servlet.http.HttpServletRequest;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnResource;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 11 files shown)


eirslett:
Great!

codefromthecrypt:
those who want to experiment can use snapshot from http://oss.jfrog.org/artifactory/oss-snapshot-local

neilstevenson:
@adriancole `zipkin-ui/js/config.js` tries to load from the old URL

codefromthecrypt:
thx for the fix!

naoman:
This change breaks the UI if we run it behind ELB with port mapping. Our ELB listens on port 80 and backend server listens on port 9411. http://zipkin_domain.com/ now redirects to http://zipkin_domain.com:9411/zipkin/

codefromthecrypt:
Thanks for raising. will take a look.

Is it possible for you to update your ELB to forward the path / to /zipkin?

http://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-listeners.html#listener-rules

codefromthecrypt:
failing test https://github.com/openzipkin/zipkin/pull/1723

codefromthecrypt:
@naoman next release will be default, but anyway you can set `--server.use-forward-headers=true` in the mean time, if you can't change the ELB config

naoman:
@adriancole thanks for making the change. Also, `--server.use-forward-headers=true` fixed the problem. 
Adding a bit more information about our issue, in case someone else runs into same. We had two layers of proxy, one for SSL termination (NGINX) and one for load balancing (ELB). Both layers were doing port mapping, and adding X-Forwarded headers. On top of using `--server.use-forward-headers=true`, we had to update the ELB to stop updating X-Forwarded headers ([by converting it to TCP from HTTP](https://forums.aws.amazon.com/thread.jspa?threadID=85725)).

codefromthecrypt:
Thx for the feedback. Nice details!

PR #1629, opened 2017-06-26
Fix to PR 1627

neilstevenson:
From https://github.com/openzipkin/zipkin/pull/1627, looks like `config.js` was missed.

Without it, you get "Error loading config.json: No message available"


codefromthecrypt:
thank you very much for noticing and fixing!

Issue #1528, opened 2017-03-01
Bound in-memory storage


codefromthecrypt:
Right now, the in-memory storage option is unbounded, which means that it can lead to out-of-memory errors which can kill the server. This has come up in the kubernetes repo and again recently when asked by @kellabyte_twitter

> I could see some cases where historical data isn't required, more just tracing "recently" or "whats going wrong right now"

It is possible to make a storage option which can evict based on last timestamp, separately tracking size of index data from encoded spans. There's some related code in zipkin-reporter-java for making a fixed-size for the pending spans queue. Here, codecs support estimating the encoded size (ex that a span is 298bytes or so). Such code could be reused or at least looked at.

By doing something like this, people could have a cheap single-node zipkin which won't blow up on memory, even if it is persistent and not replicated. The tradeoff is that this sort of code, while fun, is even less straightforward than the existing in-memory code.

tajdars:
+1 for bounded InMemoryStorage.

harsharastogi:
+1

joel-airspring:
This would be useful for us too.  I'd like to take a crack at it.

I think it makes sense to take the approach of letting users either set a maximum number of spans or a maximum number of span bytes (using a rough estimate 500 bytes per span), so
  --max-spans=10000
and
  --max-span-bytes=5mb
would be the same.  Or perhaps since this applies to the in-memory storage only, the options should be --max-inmemory-spans and --max-inmemory-span-bytes. . .something like that.

In the process of familiarizing myself with the code, I've made a few cleanups that I'll submit via separate PRs.  I haven't actually written the code to purge the old spans yet, but I hope to tackle that today.

Does that sound OK?

Joel

codefromthecrypt:
Go for simplest auto-purge based on oldest data. I suspect span or trace id
is likely easier to implement, but happy to review whatever proves easiest.

One knob is ideal because we dont want to start encouraging use of mem
storage by adding too many features to it.

joel-airspring:
OK.  I'll just go with a maximum # spans.

I've got a purge() method added to the InMemorySpanStore class.  This method deletes the oldest traces (and their spans) when invoked.  Do you have an opinion on who should do the invoking?  And how?  I'm don't quite know most most idiomatic in Java.  For example:

1. Perhaps InMemoryStorage should have a worker thread that just invokes it every minute or so (cron-like).  In this case, purge() would be being called from outside InMemorySpanStore.
[...]

PR #1631, opened 2017-06-27
Bound in-memory

joel-airspring:
Here's a first whack at fixing #1528 .

A few comments:

1.  Here's a before and after of memory usage:
before
after
If I recall correctly, this is a run with 200,000 spans stored.  You can see in the non-bounded case, that the JVM hits the heap limit, then thrashes desperately (the flatline) before dying.  In the second picture, you can see the heap usage increase until it hits the max-span limit and old spans start getting purged.

2. I set the default number of spans to be 500,000.  This requires about 600MB.  I don't know if this is the right default or not.

3. I'm open to refactoring the code to create better interfaces, but here's what I did for now.  I did three commits for ease of understanding.  The first adds a StorageMetrics interface so that the number of actual spans retained can be shown (as opposed to the CollectorMetrics, which shows the number of spans received).  The second adds the InMemorySpanConsumer.purge() method, along with tests and metrics for it, but doesn't actually enforce a max span size.  The third and final commit wires all that up.

4. I added a max(long, long) method temporarily until I can figure out how to make Long.max(long, long) acceptable to animal-sniffer.

5. I had to add two purgeTraceIdTimeStamps() methods since I haven't figured out how to get access to the values of serviceToTraceIdTimeStamp as an ordered set, like I was able to for traceIdTimeStamps.  Because I can't access the values in order, I can't break out early when looping like for data to purge.  This causes the performance show in the next point:

6. I've tested up to 1 million spans.  When adding spans, the response time is 2-3 ms until we hit the limit and need to start purging.  At this point, the response time becomes ~150 ms.  This is not unbearably slow but I think will improve once 5) gets fixed.

### DIFF
--- a/zipkin-server/src/main/resources/zipkin-server-shared.yml
+++ b/zipkin-server/src/main/resources/zipkin-server-shared.yml
@@ -39,6 +39,13 @@ zipkin:
   storage:
     strict-trace-id: ${STRICT_TRACE_ID:true}
     type: ${STORAGE_TYPE:mem}
+    mem:
+      # Maximum number of spans to keep in memory.  When exceeded, oldest traces (and their spans) will be purged.
+      # A safe estimate is 1K of memory per span (each span with 2 annotations + 1 binary annotation), plus
+      # 100 MB for a safety buffer.  You'll need to verify in your own environment.
+      # Experimentally, it works with: max-spans of 500,000 with JRE argument -Xmx600m.
+      # Set to -1 for unlimited spans
+      max-spans: 500000
     cassandra:
       # Comma separated list of host addresses part of Cassandra cluster. Ports default to 9042 but you can also specify a custom port with 'host:port'.
       contact-points: ${CASSANDRA_CONTACT_POINTS:localhost}
--- a/zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
+++ b/zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
@@ -89,8 +89,8 @@ public final class InMemorySpanStore implements SpanStore {
    *
    * <p>QueryRequest.limit needs trace ids are returned in timestamp descending order.
    */
-  private final SortedMultimap<String, Pair<Long>> serviceToTraceIdTimestamp =
-      new TreeSetSortedMultimap<>(String::compareTo, VALUE_2_DESCENDING);
+  private final ServiceNameToTraceIdTimeStamp serviceToTraceIdTimestamp =
+      new ServiceNameToTraceIdTimeStamp();
   /** This is an index of {@link Span#name} by {@link zipkin.Endpoint#serviceName service name} */
   private final SortedMultimap<String, String> serviceToSpanNames =
       new LinkedHashSetSortedMultimap<>(String::compareTo);
@@ -179,25 +179,13 @@ public final class InMemorySpanStore implements SpanStore {
         serviceNames.addAll(span.serviceNames());
       }
     }
-    deleteServiceTraceIdTimeStamps(serviceNames, traceId);
-    return spansEvicted;
-  }
-
-  private synchronized void deleteServiceTraceIdTimeStamps(Collection<String> serviceNames, Long targetTraceId) {
     for (Iterator<String> serviceNameIter = serviceNames.iterator(); serviceNameIter.hasNext(); ) {
       String serviceName = serviceNameIter.next();
-      Collection<Pair<Long>> traceIdTimeStamps = serviceToTraceIdTimestamp.get(serviceName);
-      for (Iterator<Pair<Long>> traceIdTimeStampIter = traceIdTimeStamps.iterator();
-          traceIdTimeStampIter.hasNext(); ) {
-        if (traceIdTimeStampIter.next()._1.equals(targetTraceId)) {
-          traceIdTimeStampIter.remove();
-        }
-      }
-      if (traceIdTimeStamps.isEmpty()) {
-        serviceToTraceIdTimestamp.remove(serviceName);
+      if (serviceToTraceIdTimestamp.removeServiceIfTraceId(serviceName, traceId)) {
         serviceToSpanNames.remove(serviceName);
       }
     }
+    return spansEvicted;
   }
 
   /**
@@ -323,7 +311,37 @@ public final class InMemorySpanStore implements SpanStore {
     return right._1.compareTo(left._1);
   };
 
-  static final class TreeSetSortedMultimap<K, V> extends SortedMultimap<K, V> {
+  static final class ServiceNameToTraceIdTimeStamp extends TreeSetSortedMultimap<String, Pair<Long>> {
+    ServiceNameToTraceIdTimeStamp() {
+      super(String::compareTo, VALUE_2_DESCENDING);
+    }
+
+    @Override Set<Pair<Long>> valueContainer() {
+      return new TreeSet<>(valueComparator);
+    }
+
+    /**
+     * Deletes traceIdTimeStamps matching traceId for serviceName
+     * @return true if all traceIdTimeStamps have been removed for serviceName
+     */
+    boolean removeServiceIfTraceId(String serviceName, Long traceId) {
+      Collection<Pair<Long>> traceIdTimeStamps = get(serviceName);
+      for (Iterator<Pair<Long>> traceIdTimeStampIter = traceIdTimeStamps.iterator();
+        traceIdTimeStampIter.hasNext(); ) {
+        if (traceIdTimeStampIter.next()._1.equals(traceId)) {
+          traceIdTimeStampIter.remove();
+          size--;
+        }
+      }
+      if (traceIdTimeStamps.isEmpty()) {
+        remove(serviceName);
+        return true;
+      }
+      return false;
+    }
+  }
+
+  static class TreeSetSortedMultimap<K, V> extends SortedMultimap<K, V> {
     final Comparator<V> valueComparator;
 
     TreeSetSortedMultimap(Comparator<K> keyComparator, Comparator<V> valueComparator) {
@@ -348,7 +366,7 @@ public final class InMemorySpanStore implements SpanStore {
 
   static abstract class SortedMultimap<K, V> {
     private final TreeMap<K, Collection<V>> delegate;
-    private int size = 0;
+    int size = 0;
 
     SortedMultimap(Comparator<K> comparator) {
       delegate = new TreeMap<>(comparator);
--- a/zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
+++ b/zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
@@ -37,7 +37,35 @@ import static zipkin.internal.ApplyTimestampAndDuration.guessTimestamp;
 import static zipkin.internal.GroupByTraceId.TRACE_DESCENDING;
 import static zipkin.internal.Util.sortedList;
 
-/** Internally, spans are indexed on 64-bit trace ID */
+/**
+ * Internally, spans are indexed on 64-bit trace ID
+ *
+ * <p>Here's an example of some traces in memory:
+ *
+ * <pre>{@code
+ * spansByTraceIdTimeStamp:
+ *    <aaaa,July 4> --> ( spanA(time:July 4, traceId:aaaa, service:foo, name:GET),
+ *                        spanB(time:July 4, traceId:aaaa, service:bar, name:GET) )
+ *    <cccc,July 4> --> ( spanC(time:July 4, traceId:aaaa, service:foo, name:GET) )
+ *    <bbbb,July 5> --> ( spanD(time:July 5, traceId:bbbb, service:biz, name:GET) )
+ *    <bbbb,July 6> --> ( spanE(time:July 6, traceId:bbbb) service:foo, name:POST )
+ *
+ * traceIdToTraceIdTimeStamps:
+ *    aaaa --> [ <aaaa,July 4> ]
+ *    bbbb --> [ <bbbb,July 5>, <bbbb,July 6> ]
+ *    cccc --> [ <cccc,July 4> ]
+ *
+ * serviceToTraceIdTimeStamp:
+ *    foo --> [ <aaaa,July 4>, <cccc,July 4>, <bbbb,July 6> ]
+ *    bar --> [ <aaaa,July 4> ]
+ *    biz --> [ <bbbb,July 5> ]
+ *
+ * serviceToSpanNames:
+ *    bar --> ( GET )
+ *    biz --> ( GET )
+ *    foo --> ( GET, POST )
+ * }</pre>
+ */
 public final class InMemorySpanStore implements SpanStore {
   /**
    * Primary source of data is this map, which includes spans ordered descending by timestamp. All
--- a/zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
+++ b/zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
@@ -283,13 +283,15 @@ public final class InMemorySpanStore implements SpanStore {
       Collection<V> valueContainer = delegate.get(key);
       if (valueContainer == null) {
         synchronized (delegate) {
-          if (!delegate.containsKey(key)) {
-            valueContainer = valueContainer();
-            delegate.put(key, valueContainer);
+          Collection<V> old = delegate.put(key, valueContainer = valueContainer());
+          if (old != null) { // lost race
+            valueContainer = old;
           }
         }
       }
-      valueContainer.add(value);
+      synchronized (delegate) {
+        valueContainer.add(value);
+      }
     }
 
     // not synchronized as only used for for testing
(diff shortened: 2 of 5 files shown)


joel-airspring:
Just made a change to speed up the purge.  With commit e0b6af5, a purge of old spans with 1 million spans in-memory goes from ~150 ms to 5 ms.

joel-airspring:
I fixed the issue with Long.max() not being available in Java 1.7, and fixed a warning about a missing "@Override".

zipkin-server/src/main/java/zipkin/server/ActuateStorageMetrics.java
```diff
@@ -0,0 +1,103 @@
...
+ *     <li>counter.storage.spans.retained.$type - Number of spans available for query.  Each span
+ *     will belong to a trace</li>
+ *     <li>counter.storage.traces.purged.$type - Number of traces that have been deleted</li>
+ *     <li>counter.storage.spans.purged.$type - Number of spans that have been deleted</li>
+ * </ul>
+ * </pre>
+ *
+ * See https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-metrics.html
+ */
+public final class ActuateStorageMetrics implements StorageMetrics, PublicMetrics
```
codefromthecrypt: I like the idea of what you are doing, but I don't want us to add a top-level type that only supports in-memory storage. It adds a significant cognitive and maintenance impact in support of a non-production configuration.

If this is needed, I would probably flatten the hierarchy and expose as an explicit hook for InMemoryStorage (without exposing a new public type). If not needed beyond tests, I would remove it entirely. 

Keep in mind we don't even have counters like this for production code such as cassandra cache counters!

zipkin-server/src/main/resources/zipkin-server-shared.yml
```diff
@@ -39,6 +39,13 @@ zipkin:
   storage:
     strict-trace-id: ${STRICT_TRACE_ID:true}
     type: ${STORAGE_TYPE:mem}
+    mem:
+      # Maximum number of spans to keep in memory.  When exceeded, oldest traces (and their spans) will be purged.
+      # A safe estimate is 1K of memory per span (each span with 2 annotations + 1 binary annotation), plus
+      # 100 MB for a safety buffer.  You'll need to verify in your own environment.
+      # Experimentally, it works with: max-spans of 500,000 with JRE argument -Xmx=600m.
+      # Set to -1 for unlimited spans
+      max-spans: 500000
```
codefromthecrypt: this is good.

zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
```diff
@@ -17,14 +17,17 @@
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashSet;
```
codefromthecrypt: always linkedhashset :) helps to avoid breaking tests or other things that look at order. less distractions

zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
```diff
@@ -38,16 +41,45 @@
 import static zipkin.internal.GroupByTraceId.TRACE_DESCENDING;
 import static zipkin.internal.Util.sortedList;
 
-/** Internally, spans are indexed on 64-bit trace ID */
+/**
+ * Internally, spans are indexed on 64-bit trace ID
+ *
+ * Here's an example of some traces in memory:
```
codefromthecrypt: for each page break, prefix with `<p>` so the javadoc looks like what you mean

zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
```diff
@@ -38,16 +41,45 @@
...
+ *
+ * traceIdToSpans:
```
codefromthecrypt: wrap with
```
<pre>{@code

your stuff here

}</pre>
```

zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
```diff
@@ -38,16 +41,45 @@
...
   private final Multimap<Long, Span> traceIdToSpans = new LinkedListMultimap<>();
-  private final Set<Pair<Long>> traceIdTimeStamps = new TreeSet<>(VALUE_2_DESCENDING);
-  private final Multimap<String, Pair<Long>> serviceToTraceIdTimeStamp =
+  private final TreeSet<Pair<Long>> traceIdTimeStamps = new TreeSet<>(VALUE_2_DESCENDING);
+  private final SortedByValue2Descending serviceToTraceIdTimeStamp =
       new SortedByValue2Descending<>();
   private final Multimap<String, String> serviceToSpanNames =
       new LinkedHashSetMultimap<>();
   private final boolean strictTraceId;
+  private final StorageMetrics metrics;
```
codefromthecrypt: mentioned earlier, squash by exposing one or two key counters you need in tests. Avoid new public types.

zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
```diff
@@ -38,16 +41,45 @@
...
+  private final int maxSpans;
   volatile int acceptedSpanCount;
+  volatile int retainedSpanCount;
```
codefromthecrypt: this feels like a fair amount of bookkeeping which may make this class even less attractive to maintain. See how much of this you can get rid of and have it still work

zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
```diff
@@ -56,9 +88,11 @@ public InMemorySpanStore() {
 
   InMemorySpanStore(InMemoryStorage.Builder builder) {
     this.strictTraceId = builder.strictTraceId;
+    this.metrics = builder.metrics == null ? StorageMetrics.NOOP_METRICS : builder.metrics;
```
codefromthecrypt: nix this

zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
```diff
@@ -69,18 +103,91 @@ public InMemorySpanStore() {
...
 
     @Override public String toString() {
       return "InMemorySpanConsumer";
     }
+
+    /*
+     * Accepts number of spans to remain after purging.  Will delete a trace and all
+     * its spans until this goal is achieved.
+     *
+     * Returns number of spans purged.
```
codefromthecrypt: since the description is written like this, maybe `@return` is better that a newline and the same.

codefromthecrypt (commented):
I like that you have things working, while at the same time troubled by how difficult it must have been to achieve. I'm wondering if we can't rejig the internals of the in memory storage to make it easier to purge...

will think on this a bit on my side, too.

Regardless, thanks for greatly progressing this with working code.

codefromthecrypt:
playing a little here, I think a couple things might help simplify or at least strangle the nasty parts to one place.

Consider doing the eviction up-front. This is a little nicer I think that afterwards, especially if the input is impossible (ex input spans are larger than max!).
```
  final StorageAdapters.SpanConsumer spanConsumer = new StorageAdapters.SpanConsumer() {
     @Override public void accept(List<Span> spans) {
+      int delta = spans.size();
+      int spansToRecover = (traceIdToSpans.size() + delta) - maxSpanCount;
+      evictToRecoverSpans(spansToRecover);
+
       for (Span span : spans) {
         Long timestamp = guessTimestamp(span);
         Pair<Long> traceIdTimeStamp =
@@ -83,6 +88,17 @@ public final class InMemorySpanStore implements SpanStore {
     }
   };
```

Adding size to multimap looks something like this... (one distracting bugfix in there..)
```
+    int size() {
+      return size;
+    }
+
     void put(K key, V value) {
       Collection<V> valueContainer = delegate.get(key);
       if (valueContainer == null) {
         synchronized (delegate) {
-          if (!delegate.containsKey(key)) {
-            valueContainer = valueContainer();
-            delegate.put(key, valueContainer);
+          Collection<V> old = delegate.put(key, valueContainer = valueContainer());
+          if (old != null) { // lost race
+            valueContainer = old;
           }
         }
       }
-      valueContainer.add(value);
+      synchronized (delegate) {
+        if (valueContainer.add(value)) size++;
+      }
     }
```

Implement the eviction method, quite similar in nature to yours, except as enclosed directly by InMemorySpanStore for easier testing.
```java 
+  /** Returns the count of spans evicted. */
+  int evictToRecoverSpans(int spansToRecover) {
+    int spansEvicted = 0;
+    while (spansToRecover > 0) {
+      int spansInOldestTrace = deleteOldestTrace();
+      spansToRecover -= spansInOldestTrace;
+      spansEvicted += spansInOldestTrace;
+    }
+    return spansEvicted;
+  }
```

Deleting the oldest trace has one major edge case, which is where a span came with no guessable timestamp. I think that's ok, just skip those if you can, and choose the first valid timestamped trace and go from there. This has a bunch of scanning involved at the moment, but shouldn't require you to peer at any span data. For example, `traceIdTimeStamps` is already ordered, scan until you find anything with a valid timestamp and delete all you can associated with it. that would be `spansInOldestTrace` above. Let the loop continue even if expensive (for now).

This advice isn't looking as deeply as you have, just I'd like to find a way to avoid re-calculating things added on input.

codefromthecrypt:
@joel-airspring not sure this makes things very clear or not, but it is an attempt to rejig so that implementing this could be more straightforward. lemme know if you think it doesn't help :) https://github.com/openzipkin/zipkin/pull/1636

joel-airspring:
Adrian, thanks for all the detailed feedback!  I'll go through it today and check out #1636 too.

Joel

joel-airspring:
@adriancole Here's a second pass.  I think I've incorporated all the feedback.

1. It's based on your latest refactor.
2. All metrics/bookkeeping stuff is gone.
3. I used your evictToRecoverSpans() and generally changed all terminology from 'purge' to 'evict'.  Much better nomenclature!
4. It does have the same performance profile as my original checkin.  Before it starts evicting, adding new spans takes 2-5 ms, but after it starts evicting, it takes 200 - 400 ms (@ 1M spans).  I'd like to improve that.
5. I added a size() method to SortedMultimap, but it doesn't handle properly the case of using an iterator to remove items.  The current code doesn't depend on size() in these cases, but it's kind of ugly, and I'd like to fix that.

codefromthecrypt:
thx. mind rebasing over master (having merged). It will make it easier to
see what's new.

joel-airspring:
Rebased.

joel-airspring:
One other comment:  I made the documentation describing the in-memory data layout a separate commit so it can be axed if you want.  It's helpful to to me to see things visually like that, but you may not love it since it's not super-precise.

codefromthecrypt:
one note on the internals of sortedmultimap:

it might be easiest to reason with synchronization as coarse grained. ex
synchronize every method, then as performance requires, tidy the locks. I
know I started the suggestion towards finer policy, but realized I missed a
few spots :)

For ex copy-out in the methods which returns collections, which ensures
read ops don't get hit when eviction is in progress. Then, just need to
more carefully check null on gets resulting from an iteration of those
collections (ex maybe there's a race condition when eviction is in
progress, where you iterate a key that was there when you read the list,
but not when you went to get the spans associated with it).

once this naive stuff is in place, then we can go back and tune for
concurrency as needed and as doesn't complicate things too much.

sound like a plan?

joel-airspring:
> it might be easiest to reason with synchronization as coarse grained. ex
synchronize every method, then as performance requires, tidy the locks. I
know I started the suggestion towards finer policy, but realized I missed a
few spots :)

Does that mean to add 'synchronized' to every mutating method?  Those would be put() and remove().  (And, of course, take out any fine-grained synchronize blocks within the methods.)

> For ex copy-out in the methods which returns collections, which ensures
read ops don't get hit when eviction is in progress. Then, just need to
more carefully check null on gets resulting from an iteration of those
collections (ex maybe there's a race condition when eviction is in
progress, where you iterate a key that was there when you read the list,
but not when you went to get the spans associated with it).

I'm not quite sure what you mean by 'copy-out'.  Do you mean that every time we do a get(), we should make a copy of the data and return that?  I guess that means we either couldn't mutate that copy or if we did, it would have to be merged back into the real data, which sounds painful.

Perhaps it would simplify things if we only made changes directly via the API of SortedMultimap (and not by getting a linked list, for example, and deleting its contents directly.  Maybe add a synchronized removeIf(key, <condition>) method to SortedMultimap?  And use it like:

serviceToTraceIdTimeStamp.removeIf("foo", "id:aaaaa");  --> Go through the list for service name "foo" and remove any entries that have an traceId of aaaaa.

Am I headed in the right direction here?

codefromthecrypt:
>
> Does that mean to add 'synchronized' to every mutating method? Those would
> be put() and remove(). (And, of course, take out any fine-grained
> synchronize blocks within the methods.)
>
> if we don't synchronize the reads, then they wouldn't see writes made by
other threads, right? Def yeah take out synchronized inside them. Think of
it the same as Collections.synchronizedX for our multimap

> I'm not quite sure what you mean by 'copy-out'. Do you mean that every
> time we do a get(), we should make a copy of the data and return that? I
> guess that means we either couldn't mutate that copy or if we did, it would
> have to be merged back into the real data, which sounds painful.
>
> I was thinking of a copy-out on read because I didn't expect any mutation
except add (covered by the synchronized op on multimap) and delete by key
(because we are pruning traces, all spans with a trace id, that match an
expression) ex spansByTraceIdTimeStamp and traceIdToTraceIdTimeStamps can
delete everything with a specific key on eviction.

eviction on traceId won't clean up serviceToTraceIdTimestamp (as trace ids
are in the values, not the keys). This might be something solvable with a
special-case method on multimap, or you can punt to GC by changing the
values to weak references.

> Perhaps it would simplify things if we only made changes directly via the
> API of SortedMultimap (and not by getting a linked list, for example, and
> deleting its contents directly. Maybe add a synchronized removeIf(key, )
> method to SortedMultimap? And use it like:
>
> serviceToTraceIdTimeStamp.removeIf("foo", "id:aaaaa"); --> Go through the
> list for service name "foo" and remove any entries that have an traceId of
> aaaaa.
>
yep, we'd only need something like removeIf on serviceToTraceIdTimestamp,
and this could be a very special cased method (ex a subtype of
SortedMultimap for serviceToTraceIdTimestamp with a typesafe
removeServiceIfTraceId(...)

Definitely on the right track. On later work for perf, we should put in
JMH. You can look at zipkin-reporter-java for some examples of things if
you need more, but yeah you are on the right track, so keep up the good
work!

joel-airspring:
Third pass:

1. All methods on InMemorySpanStore synchronized.  This is the simplest possible way to ensure that every read sees every write.
With 4 clients writing simultaneously, it's 2-10ms for each span submission (~30 at a time).  Once the evicting starts @1M spans, it drops to 250-2000 ms.
2. I added a subclass of SortedMultimap for the serviceToTraceIdTimeStamp type with a removeServiceIfTraceId() method.
With 4 clients writing simultaneously, it's 2-10ms for each span submission (~30 at a time).  Once the evicting starts @1M spans, it drops to 400-4000 ms.  (I don't know why this is slower than case 1, above)
3. In another commit, I enhanced removeServiceIfTraceId() to terminate early based on latest time stamp.  This cut the time when evicting to double-digit ms when evicting starts @1M spans.

Is there anything else you'd like for this PR?  It seems to be fast enough with the coarse locking with 4 clients sending data simultaneously while evicting @ 1 million spans.

zipkin-server/src/main/resources/zipkin-server-shared.yml
```diff
@@ -39,6 +39,13 @@ zipkin:
   storage:
     strict-trace-id: ${STRICT_TRACE_ID:true}
     type: ${STORAGE_TYPE:mem}
+    mem:
+      # Maximum number of spans to keep in memory.  When exceeded, oldest traces (and their spans) will be purged.
+      # A safe estimate is 1K of memory per span (each span with 2 annotations + 1 binary annotation), plus
```
codefromthecrypt, line 44: hopefully, this estimate is high :)

zipkin-server/src/main/resources/zipkin-server-shared.yml
```diff
@@ -39,6 +39,13 @@ zipkin:
...
+      # 100 MB for a safety buffer.  You'll need to verify in your own environment.
+      # Experimentally, it works with: max-spans of 500,000 with JRE argument -Xmx600m.
+      # Set to -1 for unlimited spans
```
codefromthecrypt, line 47: I will remove this note on the way in, as -1 won't work :P

zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
```diff
@@ -120,6 +135,65 @@ synchronized void clear() {
     serviceToSpanNames.clear();
   }
 
+  synchronized void addSpans(List<Span> spans) {
+    int delta = spans.size();
+    int spansToRecover = (spansByTraceIdTimeStamp.size() + delta) - maxSpanCount;
+    evictToRecoverSpans(spansToRecover);      for (Span span : spans) {
```
codefromthecrypt, line 141: I'll fix formatting on the way in.

codefromthecrypt (approved):
well done

PR #1633, opened 2017-06-27
Make all project files consistently require Java 1.8.

joel-airspring:
One pom.xml required Java 1.6, and another required 1.7.

I discovered this when animal-sniffer flagged Long.max() as invalid.  It was confusing to figure out because most sub-projects assume 1.8.  These two pom.xml's were the only ones that did not.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,8 +40,8 @@
     <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
 
     <!-- default bytecode version for src/main -->
-    <main.java.version>1.7</main.java.version>
-    <main.signature.artifact>java17</main.signature.artifact>
+    <main.java.version>1.8</main.java.version>
+    <main.signature.artifact>java18</main.signature.artifact>
 
     <main.basedir>${project.basedir}</main.basedir>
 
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
The default is intentional (not java 8) for a couple reasons, some are just
background.

We dont want to enforce more constraints than the libraries we use. Ex jooq
requires JRE8, so no problem. Though cassandra does not. Not escalating
code level prevents accidents that limit library reuse.

Some are using "local reporters" for example reporting to zipkin without a
separate transport tier. Taking the least dep on JRE level interferes least.

Another non server user is the dependency linker spark job. As recent as
last year the job needed to be pinned to JRE 7 due to some hadoop reasons
which was another reason why we had to take a second look at versions here.

The core library is not only used in spark, but also application code ...
and also java agents. Java agents are used to instrument legacy systems and
sometimes terribly so, but sometimes these use JRE6. Moreover, it was
intended to be used in android and depending on level this is kindof like
java 6.5. Long story short.. the core library is most important to remain
JRE 6.

There are some annoyances, such as copy pasting long compare and lack of
things like threadlocalrandom, but in core, this is a limited impact. The
animal sniffer did its job in this case, preventing us from breaking our
contract.

codefromthecrypt:
There's a longer explanation held hostage until my phone reconnects to the internet :) suffice to say this is very intentional, especially the core version at JRE 6. Feel free to ping on gitter if you have any q's about approaches considering the limitation

PR #1635, opened 2017-06-27
Fix test broken by commit for #1625

joel-airspring:
This has been silently failing since it was committed.


codefromthecrypt:
Hmm wonder why it is silently failing (as opposed to loudly failing).

codefromthecrypt:
thanks!

PR #1636, opened 2017-06-28
Refactors in-memory storage so that there's a common type across data sets

codefromthecrypt:
This rejigs things so that the pair of (traceID, timestamp) is used consistently
across all data sets that grow unbounded. This should make it easier to support
purging later. For example one implementation could be to iterate
traceIdToTraceIdTimeStamps in reverse (using delegate.descendingKeySet()),
which should give the eldest trace ID and delete that. When weak references are
used elsewhere, GC should clean them up. Other mechanisms besides weak
references could be used as well.

See #1528

### DIFF
--- a/zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
+++ b/zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
@@ -18,12 +18,11 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Iterator;
-import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.TreeSet;
 import zipkin.DependencyLink;
 import zipkin.Span;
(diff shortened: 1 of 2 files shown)


zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
```diff
@@ -40,12 +39,29 @@
...
+   */
+  private final SortedMultimap<Pair<Long>, Span> spansByTraceIdTimeStamp =
+      new LinkedHashSetSortedMultimap<>(VALUE_2_DESCENDING);
+
+  /** This supports span lookup by {@link zipkin.Span#traceId lower 64-bits of the trace ID} */
+  private final SortedMultimap<Long, Pair<Long>> traceIdToTraceIdTimeStamps =
+      new LinkedListSortedMultimap<>(Long::compareTo);
+  /** This supports span lookup by {@link zipkin.Endpoint#serviceName service name} */
+  private final SortedMultimap<String, Pair<Long>> serviceToTraceIdTimestamp =
+      new SortedByValue2Descending<>(String::compareTo, VALUE_2_DESCENDING);
```
codefromthecrypt: ps noticing the names of these types are getting a bit terrible. might clean them up later if this goes anywhere

joel-airspring (commented):
I like this refactor, and will base my changes on them.

Quick question:  What's the use case for this being a list:

```Java/** This supports span lookup by {@link zipkin.Span#traceId lower 64-bits of the trace ID} */
 +  private final SortedMultimap<Long, Pair<Long>> traceIdToTraceIdTimeStamps =
 +      new LinkedListSortedMultimap<>(Long::compareTo);
```

instead of a LinkedHashSetMultiMap?

Are you anticipating cases where you want to know that

aaaa --> [<aaaa,July4>, <aaaa,July4>, <aaaa,July5>]

as opposed to just

aaaa --> (<aaaa,July4>, <aaaa,July5>)

(Notice that the first example can and does contain duplicates because it's a linked list.)

codefromthecrypt:
I suspect there is no value in a list here. I will switch and if tests
still pass, merge it as a set. Nice catch.

codefromthecrypt:
@joel-airspring keen eye: LinkedListSortedMultimap would not have helped before or after this change

joel-airspring:
Thanks.  I only noticed it because I hit a problem with that data structure when trying to evict spans!

zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
```diff
@@ -40,12 +38,33 @@
...
+  /** This supports span lookup by {@link zipkin.Span#traceId lower 64-bits of the trace ID} */
+  private final SortedMultimap<Long, Pair<Long>> traceIdToTraceIdTimeStamps =
+      new LinkedHashSetSortedMultimap<>(Long::compareTo);
+  /**
+   * This supports span lookup by {@link zipkin.Endpoint#serviceName service name}.
+   *
+   * <p>QueryRequest.limit needs trace ids are returned in timestamp descending order.
+   */
+  private final SortedMultimap<String, Pair<Long>> serviceToTraceIdTimestamp =
+      new TreeSetSortedMultimap<>(String::compareTo, VALUE_2_DESCENDING);
```
codefromthecrypt, line 65: @joel-airspring FYI: I renamed this type to more generic as it keeps the comparators in one place.

codefromthecrypt:
oops killing linkedlist broke a test :) It supported reading back raw spans (which helps show when instrumentation sent the same spans twice)

zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
```diff
@@ -40,12 +39,34 @@
...
-  private final Multimap<Long, Span> traceIdToSpans = new LinkedListMultimap<>();
-  private final Set<Pair<Long>> traceIdTimeStamps = new TreeSet<>(VALUE_2_DESCENDING);
-  private final Multimap<String, Pair<Long>> serviceToTraceIdTimeStamp =
-      new SortedByValue2Descending<>();
-  private final Multimap<String, String> serviceToSpanNames =
-      new LinkedHashSetMultimap<>();
+  /**
+   * Primary source of data is this map, which includes spans ordered descending by timestamp. All
+   * other maps are derived from the span values here. This uses a list for the spans, so that it is
+   * visible (via /api/v1/trace/id?raw) when instrumentation report the same spans multiple times.
```
codefromthecrypt, line 45: @joel-airspring added this note, as clearly I forgot why when you asked earlier. ZipkinRuleTest reminded me :)

joel-airspring:
Bummer!  I had run the zipkin/zipkin tests, but not those zipkin/zipkin-junit tests that failed.  Thanks for checking.  I'll keep rolling with the existing linked list implementation.

codefromthecrypt:
>
> Bummer! I had run the zipkin/zipkin tests, but not those
> zipkin/zipkin-junit tests that failed. Thanks for checking. I'll keep
> rolling with the existing linked list implementation.
>
Note: I moved the linkedlist to the primary data source (as opposed to the
index where it was in the last diff you saw)

>

zipkin/src/test/java/zipkin/storage/InMemorySpanStoreTest.java
```diff
@@ -13,13 +13,26 @@
...
 public class InMemorySpanStoreTest extends SpanStoreTest {
   final InMemoryStorage storage = new InMemoryStorage();
 
   @Override protected StorageComponent storage() {
     return storage;
   }
 
+  /** This shows when spans are sent multiple times. Doing so can reveal instrumentation bugs. */
+  @Test public void getRawTrace_sameSpanTwice(){
+    accept(span1);
```
codefromthecrypt, line 29: added a test closer to the source for the redundant write issue

PR #1637, opened 2017-06-28
Fix UI issue

fgcui1204:
We found zipkin UI can't display the friendly if there are lots of spans. Looks like the following picture:
image

I just remove the width of the span. It works now. 

please review it. Thanks

### DIFF
--- a/zipkin-ui/css/trace.css
+++ b/zipkin-ui/css/trace.css
@@ -38,7 +38,6 @@
 }
 
 #trace-container .span .handle {
-  width: 150px;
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
can you paste before and after of the entire trace detail? Ex does removing this end up affecting hiding spans that start early?

codefromthecrypt:
tell you what... I'm going to take a chance and merge this. If it was a bad choice, we will hear about it  and revert. Thanks!

PR #1641, opened 2017-06-30
Deletes empty test.

joel-airspring:
It looks like the behavior to be tested is covered in the test above:
  'should order traces by duration and tie-break using trace id'


codefromthecrypt:
ty again!

PR #1642, opened 2017-06-30
In UI, show nothing instead of 'NaN' for null duration

### LINKED ISSUE #1575: NaN on duration <1us
Some instrumentation special case when duration is <1 (by rounding up to 1). We should document this as it prevents problems like NaN in the UI. We should also fix the UI :P

joel-airspring:
This fixes #1575 for me.

The case where I was seeing NaN was when I had a span with two annotations with the same timestamp, like SR/SS bot at the exact same time.  In this case, the server was returning a span with no duration field, so Javascript sees it as undefined.


codefromthecrypt:
thanks so much for this.. NaN is not a great experience.

PR #1645, opened 2017-07-04
Refactors in-memory storage eviction logic to remove lockups

codefromthecrypt:
I ran the server with self-tracing-enabled and did get services in a
for loop. The previous variant wasn't happy due to some race conditions
and in attempting to fix it, I noticed some places to simplify things.

### DIFF
--- a/zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
+++ b/zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
@@ -21,9 +21,9 @@ import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
-import java.util.TreeSet;
 import zipkin.DependencyLink;
 import zipkin.Span;
 import zipkin.internal.CorrectForClockSkew;
@@ -55,10 +55,10 @@ import static zipkin.internal.Util.sortedList;
  *    bbbb --> [ <bbbb,July 5>, <bbbb,July 6> ]
  *    cccc --> [ <cccc,July 4> ]
  *
- * serviceToTraceIdTimeStamp:
- *    foo --> [ <aaaa,July 4>, <cccc,July 4>, <bbbb,July 6> ]
- *    bar --> [ <aaaa,July 4> ]
- *    biz --> [ <bbbb,July 5> ]
+ * serviceToTraceIds:
+ *    foo --> [ <aaaa>, <cccc>, <bbbb> ]
+ *    bar --> [ <aaaa> ]
+ *    biz --> [ <bbbb> ]
  *
  * serviceToSpanNames:
  *    bar --> ( GET )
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
cc @joel-airspring so this fixes my curl-in-a-for loop. Mind checking before we cut a release?

zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
```diff
@@ -84,13 +84,8 @@
...
       new LinkedHashSetSortedMultimap<>(Long::compareTo);
-  /**
-   * This supports span lookup by {@link zipkin.Endpoint#serviceName service name}.
-   *
-   * <p>QueryRequest.limit needs trace ids are returned in timestamp descending order.
-   */
-  private final ServiceNameToTraceIdTimeStamp serviceToTraceIdTimestamp =
-      new ServiceNameToTraceIdTimeStamp();
+  /** This is an index of {@link Span#traceId} by {@link zipkin.Endpoint#serviceName service name} */
+  private final ServiceNameToTraceIds serviceToTraceIds = new ServiceNameToTraceIds();
```
codefromthecrypt, line 88: big change here is reduced cardinality on the values to service -> traceId vs service -> (traceId, timestamp)

The impact is easier cleanup and a bit less scanning during writes (due to less cleanup there). Should be measurable, but we don't have benchmarks in for this at the moment.

codefromthecrypt:
here's a quick test case:

```bash
# run with self-tracing enabled, and max-spans set lower
$ SELF_TRACING_ENABLED=true java -Dzipkin.storage.mem.max-spans=500 -jar ./zipkin-server/target/zipkin-server-*exec.jar
# in another window, do this for a while
$ while true; do curl -s localhost:9411/api/v1/services;done
# then, check to see the count is less than or equal to what you set it to (default is really high)
$ curl -s localhost:9411/api/v1/traces?limit=1000000|jq '.[]|.[]|.id'|wc -l
```

zipkin/src/main/java/zipkin/storage/InMemorySpanStore.java
```diff
@@ -167,31 +166,18 @@ synchronized int evictToRecoverSpans(int spansToRecover) {
...
     for (Iterator<Pair<Long>> traceIdTimeStampIter = traceIdTimeStamps.iterator();
       traceIdTimeStampIter.hasNext(); ) {
       Pair<Long> traceIdTimeStamp = traceIdTimeStampIter.next();
-      latestTimeStamp = Math.max(latestTimeStamp, traceIdTimeStamp._2);
       Collection<Span> spans = spansByTraceIdTimeStamp.remove(traceIdTimeStamp);
       spansEvicted += spans.size();
-      for (Iterator<Span> spansIter = spans.iterator(); spansIter.hasNext(); ) {
-        Span span = spansIter.next();
-        if (span.timestamp != null) {
-          latestTimeStamp = Math.max(latestTimeStamp, span.timestamp);
```
codefromthecrypt, line 185: note with this change, we no longer need to derive data from spans when cleaning up. Instead we use the values pinned at insert time (because we are deleting by trace ID, this all works)

joel-airspring:
I tested with commit eabaebe, ran it up to 1 M datapoints.  Looks good to me!

I like this cleanup.  It alway feels good to get rid of a loop!

Joel

codefromthecrypt:
Thanks for testing! Will release your work shortly

PR #1649, opened 2017-07-08
Fix UI code to reference api under base path /zipkin

codefromthecrypt:
The last change "missed a spot" which wasn't noticable due to a false
positive in ZipkinServerIntegrationTest. We should consider rewriting
this test class if MockMvc isn't reliable.


codefromthecrypt:
thanks @a86c6f7964 for reporting cc @shakuzen

shakuzen (approved):

PR #1650, opened 2017-07-10
Refactors codec to reduce redundant list creation

codefromthecrypt:
This includes cleanups and performance improvements when reading spans.
Notably, this stops recreating lists by re-using Span.Builder

To test this, we use the same client-span json from zipkin-reporter.

Benchmark                                          Mode  Cnt   Score   Error  Units
CodecBenchmarks.readClientSpan_json_zipkin         avgt   15  10.070 ± 0.431  us/op
CodecBenchmarks.readClientSpan_thrift_libthrift    avgt   15   2.633 ± 0.047  us/op
CodecBenchmarks.readClientSpan_thrift_zipkin       avgt   15   2.846 ± 0.059  us/op
CodecBenchmarks.readTenClientSpans_json_zipkin     avgt   15  78.381 ± 2.028  us/op
CodecBenchmarks.readTenClientSpans_thrift_zipkin   avgt   15  28.216 ± 0.365  us/op

After:

Benchmark                                          Mode  Cnt   Score   Error  Units
CodecBenchmarks.readClientSpan_json_zipkin         avgt   15  9.700 ± 0.261  us/op
CodecBenchmarks.readClientSpan_thrift_libthrift    avgt   15   2.367 ± 0.022  us/op
CodecBenchmarks.readClientSpan_thrift_zipkin       avgt   15   2.396 ± 0.034  us/op
CodecBenchmarks.readTenClientSpans_json_zipkin     avgt   15  76.188 ± 1.164  us/op
CodecBenchmarks.readTenClientSpans_thrift_zipkin   avgt   15  23.903 ± 0.404  us/op

### DIFF
--- a/benchmarks/src/main/java/zipkin/benchmarks/CodecBenchmarks.java
+++ b/benchmarks/src/main/java/zipkin/benchmarks/CodecBenchmarks.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2016 The OpenZipkin Authors
+ * Copyright 2015-2017 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -17,6 +17,8 @@ import com.google.common.io.ByteStreams;
 import com.twitter.zipkin.thriftjava.Annotation;
 import com.twitter.zipkin.thriftjava.BinaryAnnotation;
 import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 import org.apache.thrift.TDeserializer;
 import org.apache.thrift.TException;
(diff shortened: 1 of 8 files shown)


codefromthecrypt:
added a commit for thrift doing the same. will later run benchmarks there

benchmarks/src/main/java/zipkin/benchmarks/CodecBenchmarks.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2016 The OpenZipkin Authors
+ * Copyright 2015-2017 The OpenZipkin Authors
```
adriancole: most new lines of code here are actually increase in the benchmarks module. Otherwise, this change is pretty much line of code neutral

codefromthecrypt:
the "readTen" benchmarks are important to show optimizations when decoding lists which nowadays are often much larger than 10 spans. For example, most new tracers batch and can send quite a lot of spans per message (easily hundreds in some scenarios)

Issue #1644, opened 2017-07-04
Implementation of span2 format


codefromthecrypt:
This is the implementation of #1499 Not all steps are sequential

## Overview
This work supports a new data type which represents a single-host view of an operation. It will be represented in json w/ optional proto3 mapping.

Full description is in #1499, but the format is below:
```
{
  "traceId": 32lowerHexCharacters,
  "parentId": 16lowerHexCharactersOrAbsentForRoot,
  "id": 16lowerHexCharacters,
  "kind": enum(CLIENT|SERVER|PRODUCER|CONSUMER|Absent),
  "name": stringOrAbsent,
  "timestamp": uint53EpochMicrosOrAbsentIfIncomplete,
  "duration": uint53MicrosOrAbsentIfIncomplete,
  "localEndpoint": existingEndpointTypeOrAbsent,
  "remoteEndpoint": existingEndpointTypeOrAbsent,
  "annotations": [
    {"timestamp": uint53EpochMicros, "value": string},
    ...
  ],
  "tags": {
    string: string,
    ...
  },
  "debug": trueOrAbsent,
  "shared": trueOrAbsent
}
```

## Implementation overview
As we are working in the brown field, various phases can work at their own pace. Tracer libraries should be able to opt-into this without a breaking change to user apis. Any storage or transport work should anticipate mixed formats, but allow the ability to strictly accept the new one. The UI is completely decoupled from this, as all read apis still use the old format. We may revisit such later.

### Basic library support
Since the server code is written in java, a first step could be to write the java library
- [x] Create a working java Span2 type #1669 
  * This allows it to coexist in the classpath with existing model
- [x] Make a conversion utility between the original and new model from Span2 (splitting as necessary) #1670
- [x] Create a Span2 json codec #1671
- [x] Make a detecting json decoder which can tell the difference between a "simple span" list and a normal one #1684
  * This allows us to decode "simple spans" from existing transports
  * the localEndpoint field is the best signal you are in a "simple span" as it is required and not present formerly.
- [x] create zipkin2 package which formalizes the Span2 library (so that other java libraries like reporter can use it directly)

### Upload Api Definition
While new servers can detect based on content, we should make it possible for new sites to strictly use the new type.
- [x] Decide whether to use a new path (/api/v2/spans) or a media type header used when not auto-detecting. Ex. `application/vnd.zipkin.span2s+json` #1684
[...]

PR #1658, opened 2017-07-21
Test cleanup including being less sensitive to single-host span storage

codefromthecrypt:
There were tests accidentally assuming merge semantics when testing
unrelated things. This scrubs some of the tests to focus on what they
are testing, and in doing so help pave forward single-host span storage.

See #1644


zipkin/src/test/java/zipkin/TestObjects.java
```diff
@@ -72,7 +72,6 @@
           .addAnnotation(Annotation.create((TODAY + 150) * 1000, CLIENT_SEND, APP_ENDPOINT))
           .addAnnotation(Annotation.create((TODAY + 200) * 1000, CLIENT_RECV, APP_ENDPOINT))
           .addAnnotation(Annotation.create((TODAY + 190) * 1000, "⻩", NO_IP_ENDPOINT))
-          .addBinaryAnnotation(BinaryAnnotation.address(CLIENT_ADDR, APP_ENDPOINT))
```
codefromthecrypt, line 75: It is invalid to put a CLIENT_ADDR on a client span. Scrubbed this as it confuses other things

zipkin/src/test/java/zipkin/storage/InMemorySpanStoreEvictionTest.java
```diff
@@ -144,14 +144,12 @@ public void evict_detailed() {
 
   @Test
   public void evict_oneTraceMultipleSpans() {
-    Span testSpan1 = span1.toBuilder().traceIdHigh(1L).traceId(123).
```
codefromthecrypt, line 147: This test was accidentally testing defaults related to ignoring traceIdHigh.

zipkin/src/test/java/zipkin/storage/SpanStoreTest.java
```diff
@@ -519,11 +520,11 @@ public void getTraces_differentiateOnServiceName() {
 
     Span trace2 = Span.builder().traceId(2).name("get").id(2)
         .timestamp((today + 2) * 1000)
-        .addAnnotation(Annotation.create((today + 1) * 1000, CLIENT_SEND, APP_ENDPOINT))
```
codefromthecrypt, line 522: this data had incorrectly aligned timestamps (which wasn't the point of the test)

zipkin/src/test/java/zipkin/storage/SpanStoreTest.java
```diff
@@ -816,17 +817,16 @@ public void clientTimestampAndDurationWinInSharedSpan() {
...
     IntStream.range(0, 100).forEach(i -> {
-      Span s = TestObjects.TRACE.get(1);
-      trace[i + 1] = s.toBuilder()
-          .id(s.id + i)
-          .timestamp(s.timestamp + i)
-          .annotations(s.annotations.stream()
-              .map(a -> Annotation.create(a.timestamp + i, a.value, a.endpoint))
-              .collect(toList()))
-          .build();
+      trace[i + 1] = Span.builder().traceId(trace[0].traceId).parentId(trace[0].id).id(i).name("")
```
codefromthecrypt: this test is also testing "raw spans". To make sure this passes, we should make a span which has no authoritative timestamp (client spans do have an authoritative timestamp and that was the prior test data)

codefromthecrypt (commented):
made some notes explaining how these changes help focus tests on what they are doing. Before, they just made troubleshooting mysterious

PR #1659, opened 2017-07-21
Fixes bug where Span.Builder.duration wasn't cleared

codefromthecrypt:


codefromthecrypt:
FYI circleci is busted due to https://discuss.circleci.com/t/connection-resets-to-maven-central/14563 cc @openzipkin/devops-tooling

Issue #1619, opened 2017-06-17
Add debug logging to CorrectForClockSkew


codefromthecrypt:
We added a lot of useful things to the DependencyLinker, as it helps answer or rule out problems quickly. When poking around a negative timestamp issue with @trtg, seems debug logging in CorrectForClockSkew could do the same (ex notify missing ip address or other cause for not affecting a trace).

cc @fedj

trtg:
the clock skew correction requires that all 4 of cs,sr,ss,cr be set or it kicks out early. I have a service written in python that calls a service written in golang. The first service has all four annotations set because py_zipkin does that by default, apparently. The first span in the second service has only sr and ss set but that means clock skew correction does not run and I'm getting negative or massively offset positive  times for this second(golang) service. Someone suggested that mergebyid in zipkin-server would have combined the client span and the server span, but they do not have the same span-id, so that's not happening either. I'm really beginning to grasp at straws as far getting one service to call a remote service and not have their spans be terribly mis-aligned in time.

codefromthecrypt:
In between now and then you should be able to configure go to use the same
span ID. Iirc it is something like ClienServerSameId or similary named (in
zipkin-go-opentracing)

codefromthecrypt:
started on this in #1663. if you enable debug, you'll see some validation warnings like so:
```
skipping redundant root span: traceId=0000000000000001, rootSpanId=0000000000000001, spanId=ffffffffffffffff
skipping clock skew adjustment due to data errors: traceId=0000000000000001
```

More work to do and pull requests welcome

jorgheymans:
`CorrectForClockSkew` was ported to javascript for Zipkin classic here https://github.com/openzipkin/zipkin/commit/d950bd8dc6ab7cc26da85050b976eefb7789f549 . Any enhancements needed to detect invalid trace data in the view layer should be reported directly against Lens.

PR #1663, opened 2017-07-25
Drops invalid span data vs counting them in dependency links

codefromthecrypt:
Previously, I over-fixed an NPE on dependency linking when bad data is
present. Instead of trying to fix cyclical references, this just drops
them. In doing so, the code is easier to maintain.

See #1640

### DIFF
--- a/zipkin/src/main/java/zipkin/internal/Node.java
+++ b/zipkin/src/main/java/zipkin/internal/Node.java
@@ -22,10 +22,13 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Queue;
-import zipkin.Span;
+import java.util.logging.Logger;
 
+import static java.lang.String.format;
+import static java.util.logging.Level.FINE;
 import static zipkin.internal.Util.checkArgument;
 import static zipkin.internal.Util.checkNotNull;
+import static zipkin.internal.Util.toLowerHex;
 
 /**
  * Convenience type representing a tree. This is here because multiple facets in zipkin require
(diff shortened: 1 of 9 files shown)


codefromthecrypt:
PS rationale to drop here is that this particular pattern we are avoiding is only possible on buggy instrumentation (ex that dole out incorrect IDs).

Common usage errors, like putting the remote service name on client annotations.. code for this is kept at the moment.

zipkin/src/test/java/zipkin/internal/DependencyLinkerTest.java
```diff
@@ -42,15 +42,12 @@ public void linksSpans() {
 
   /** This ensures a NPE isn't raised when children point to themselves as a parent. */
```
fedj: OPT: we could update this comment

zipkin/src/main/java/zipkin/internal/Node.java
```diff
@@ -132,6 +132,8 @@ public void remove() {
     Map<Long, Long> idToParent = new LinkedHashMap<>(idToNode.size());
 
     public void addNode(@Nullable Long parentId, long id, V value) {
+      // drop circular deps vs NPE later
+      if (parentId != null && parentId.equals(id)) return;
```
fedj: Since we're dropping datas, shouldn't we log that there's something wrong ? IMO, dropping data is a good way to deal with this problem but people need to know that their datas have a problem

codefromthecrypt: ok will do in a bit.

codefromthecrypt:
thx for the feedback @fedj will sort this on the way in.

zipkin/src/main/java/zipkin/internal/Node.java
codefromthecrypt: I'll change this code to throw, and the code that calls it to skip and log (since that code has the logger people turn on)

codefromthecrypt:
Added a bunch of logging and logging tests. The logging is incidentally a start of #1619

PR #1665, opened 2017-07-26
Don't use strict ID when processing dependency links (in-memory)

codefromthecrypt:
The existing code inadvertently used strict trace IDs when looking up
dependency link spans. This caused me a lot of head scratching when
trying to figure out why later refactoring caused multiple links on a
mixed trace.

### DIFF
--- a/zipkin/src/test/java/zipkin/storage/DependenciesTest.java
+++ b/zipkin/src/test/java/zipkin/storage/DependenciesTest.java
@@ -95,12 +95,29 @@ public abstract class DependenciesTest {
    */
   @Test
   public void getDependencies_strictTraceId() {
-    List<Span> mixedTrace = new ArrayList<>(TRACE);
-    mixedTrace.set(1, TRACE.get(1).toBuilder().traceIdHigh(2).build());
+    long traceIdHigh = Long.MAX_VALUE;
+    long traceId = Long.MIN_VALUE;
+    List<Span> mixedTrace = asList(
+      Span.builder().traceIdHigh(traceIdHigh).traceId(traceId).id(1L).name("get")
+        .addAnnotation(Annotation.create(TODAY * 1000, SERVER_RECV, WEB_ENDPOINT))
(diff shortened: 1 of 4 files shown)


zipkin/src/test/java/zipkin/storage/DependenciesTest.java
```diff
@@ -95,12 +95,27 @@ public void getDependencies() {
    */
   @Test
   public void getDependencies_strictTraceId() {
-    List<Span> mixedTrace = new ArrayList<>(TRACE);
-    mixedTrace.set(1, TRACE.get(1).toBuilder().traceIdHigh(2).build());
+    List<Span> mixedTrace = asList(
+      Span.builder().traceIdHigh(1L).traceId(1L).id(1L).name("get")
```
fedj: OPT: By extracting the traceId in a constant, it would emphasize that the trace is supposed to be considered the same

fedj (commented):
FAR: Shouldn't we put into the README the [strictTraceId feature](https://github.com/openzipkin/zipkin/blob/7cd7b687aebe52db2923fdd0a7d827df1940ca8d/zipkin/src/main/java/zipkin/storage/StorageComponent.java#L35) ?

codefromthecrypt:
*@fedj* commented on this pull request.

FAR: Shouldn't we put into the README the strictTraceId feature
<https://github.com/openzipkin/zipkin/blob/7cd7b687aebe52db2923fdd0a7d827df1940ca8d/zipkin/src/main/java/zipkin/storage/StorageComponent.java#L35>
?

Yeah.. surprised it isnt there. Will add it.

OPT: By extracting the traceId in a constant, it would emphasize that the
trace is supposed to be considered the same

Good idea. Thx

fedj:
LGTM

codefromthecrypt:
circleci is fussy for reasons unrelated. there are troubles downloading dependencies (happened last week, too)

PR #1667, opened 2017-07-27
Simplifies loop on MySQL dependency linking

codefromthecrypt:
fyi @hfgbarrigas


codefromthecrypt:
rationale for this is I'm trying to squeeze out the internal DependencyLinkSpan type (for #1499), and this simplifies the code that depends on it.

PR #1668, opened 2017-07-27
Moves dependency link test to MySQL package

codefromthecrypt:
Before, the MySQL code for dependency linking was not unit tested,
rather relying on integration tests. This moves tests that were for
MySQL, but stored in core to where they are used.


codefromthecrypt:
cc @shakuzen @hfgbarrigas

PR #1675, opened 2017-07-28
Adds errorCount to the dependency graph and colors accordingly

### LINKED ISSUE #1662: Add error count in api/v1/dependencies response
It would be helpful to add more metadata to the current api/v1/dependencies response.
Error count should be simple to add as @adriancole mentioned.

codefromthecrypt:
Thanks to @hfgbarrigas  for requesting this. This adds errorCount to
dependency links, indicating how many of callCount were in error.
Once merged, an update to the zipkin-api project will follow, documenting
the new field.

MySQL users who want this need to add the `error_count` column:
```sql
alter table zipkin_dependencies add `error_count` BIGINT
```

For the UI, I reused the existing color codes for transient and critical status, coloring yellow when 50% or more calls are in error, and red when 75%

I tested this by making one of the sample apps "backend" throw an error.

Example link detail screen
screen shot 2017-07-28 at 6 25 44 pm

Example of when >50% of calls are in error
screen shot 2017-07-28 at 6 25 35 pm

Example of when >75% of calls are in error
screen shot 2017-07-28 at 6 25 08 pm

Fixes #1662

### DIFF
--- a/zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiProperties.java
+++ b/zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiProperties.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2016 The OpenZipkin Authors
+ * Copyright 2015-2017 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -24,6 +24,7 @@ public class ZipkinUiProperties {
   private int defaultLookback = (int) TimeUnit.DAYS.toMillis(7);
   private String instrumented = ".*";
   private String logsUrl = null;
+  private Dependency dependency = new Dependency();
 
   public int getDefaultLookback() {
     return defaultLookback;
@@ -64,6 +65,35 @@ public class ZipkinUiProperties {
   public void setLogsUrl(String logsUrl) {
     if (!StringUtils.isEmpty(logsUrl)) {
       this.logsUrl = logsUrl;
+    }
+  }
+
+  public Dependency getDependency() {
+    return dependency;
+  }
+
+  public void setDependency(Dependency dependency) {
+    this.dependency = dependency;
+  }
+
+  public static class Dependency {
+    private float lowErrorRate = 0.5f; // 50% of calls in error turns line yellow
+    private float highErrorRate = 0.75f; // 75% of calls in error turns line red
+
+    public float getLowErrorRate() {
+      return lowErrorRate;
+    }
+
+    public void setLowErrorRate(float lowErrorRate) {
+      this.lowErrorRate = lowErrorRate;
+    }
+
+    public float getHighErrorRate() {
+      return highErrorRate;
+    }
+
+    public void setHighErrorRate(float highErrorRate) {
+      this.highErrorRate = highErrorRate;
     }
   }
 }
--- a/zipkin-ui/README.md
+++ b/zipkin-ui/README.md
@@ -91,3 +91,15 @@ where `{traceId}` will be contextually replaced by the trace id.
 If this feature is activated, you'll see on the trace detail page an additional button named `logs`.
 
 ![Logs Button](https://cloud.githubusercontent.com/assets/9842366/20482538/6e35ca66-afed-11e6-90e9-1e28f66d985e.png)
+
+#### How do I adjust the error rates in the dependency graph
+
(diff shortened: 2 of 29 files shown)


codefromthecrypt:
@hfgbarrigas can you take this for a spin?

codefromthecrypt:
ok last tests are in. I'm not planning to add unit tests for the javascript as there aren't similar ones in for the classes effected. If folks have strong feelings, I can backfill some in a week or two (takes me a while to do these, but I don't want to hold up #1499 over it).

shakuzen:
Could we make the red and yellow thresholds configurable?
I guess maybe something like highErrorRate (red) and lowErrorRate (yellow) instead of the hard-coded values?

codefromthecrypt:
>
> Could we make the red and yellow thresholds configurable?
> I guess maybe something like highErrorRate (red) and lowErrorRate (yellow)
> instead of the hard-coded values?
>
yes. we can read them from /config.json I think

codefromthecrypt:
@shakuzen added a commit to parameterize error rates

zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiProperties.java
```diff
@@ -24,6 +24,8 @@
   private int defaultLookback = (int) TimeUnit.DAYS.toMillis(7);
   private String instrumented = ".*";
   private String logsUrl = null;
+  private float dependencyLowErrorRate = 0.5f; // 50% of calls in error turns line yellow
```
codefromthecrypt: ps dependencyWarnThreshold and dependencyCriticalThreshold might also work?

codefromthecrypt:
@sirtyro does this look of interest to you?

zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiProperties.java
shakuzen: I like those too. Just one thing: it is nice that rate (hopefully) conveys that the value should be between 0 and 1. Threshold could be (mis)taken to mean an absolute number of errors, perhaps.

zipkin-ui/README.md
```diff
@@ -91,3 +91,13 @@ where `{traceId}` will be contextually replaced by the trace id.
...
+
+#### How do I adjust the error rates in the dependency graph
+
+By default, the /dependency endpoint colors a link yellow when the error
+rate is 50% or higher, or red when it 75% or higher. You can control
+these rates via the `dependency-low-error-rate` and `dependency-high-error-rate`
+properties:
+
+Ex. To make lines yellow when there's a 10% error rate, set:
+`ZIPKIN_UI_DEPENDENCY_LOW_ERROR_RATE=0.1`
```
shakuzen: Looking forward to our eventual Spring Boot 2 upgrade, this would become the hard-to-read `ZIPKIN_UI_DEPENDENCYLOWERRORRATE` per [the new relaxed binding for environment variables](https://github.com/spring-projects/spring-boot/wiki/Relaxed-Binding-2.0#environment-variables). If we wanted to make it a bit easier to read, we could nest `dependency` properties (even though the two introduced in this PR are the only properties currently...) and it would become `ZIPKIN_UI_DEPENDENCY_LOWERRORRATE`.

zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiProperties.java
shakuzen: Along those lines, should we have some validation that values are within bounds?

codefromthecrypt:
Thanks for thinking this through, Tommy. I think the original names are
best and probably best (dot)nested names

codefromthecrypt:
@shakuzen (and others). Should we say a rate of zero disables the color? For example, if "low" is zero, no yellow lines occur. If both low and high are zero, red lines don't show up either.

shakuzen:
@adriancole Wouldn't a rate of zero (or less) make all the lines yellow or red? A value above 1 I would expect to effectively disable the coloring (since you can't have more than a 100% error rate).

codefromthecrypt:
@shakuzen I meant using zero as a signal and change code to check for zero.
We could say use greater than 1 and not add code.. didnt consider it,
sounds fine to me!

hfgbarrigas:
@adriancole looks good. Thank you

ghost:
@adriancole, sorry for the lag. It certainly looks interesting. If we bump to latest zipkin, do we get it for free?

codefromthecrypt:
You do!

PR #1677, opened 2017-08-01
Adds messaging span

### LINKED ISSUE #1654: Broadcast (SPMC) messaging tracing
In #1243, we addressed one-way SPSC tracing. Multiple consumers are another thing (for example, broadcast messaging). Turns out that while linked traces (#1244) could be helpful for things like batches , it isn't the more straightforward way to deal with messages that have multiple consumers (SPMC). The most important thing in multiple consumers is to not share a span ID between the producer and the consumer.

In a recent tracing workshop, we were able to elaborate a bit on async tracing. In many cases, new traces are not needed. Ex if propagation does not imply sharing spans, you can do single producer multiple consumer as normal spans in the same trace. Here's a diagram describing most concerns, notably SPMC being the topic of this issue:

```
    ┌──────────────────────────────┐             ┌───────────────────────────────────────────┐
    │   RPC with Async Callbacks   │             │                 Messaging                 │
    ├──────────────────────────────┤             ├───────────────────────────────────────────┤
    │                              │             │                                           │
    │ request-response             │      ┌──────▶ Single Producer Single Consumer           │
    │                              │      │      │                                           │
    │   deferring completion       │      │      │ No guarantee of only one consumer from    │
    │    onto a different thread   │      │      │ the POV of the producer. No guarantee     │
    │                              │      │      │ of the scale of latency between producer  │
    │                              │      │      │ and consumer. Can consider collapsing     │
    │                              │      │      │ part of the timeline.                     │
    │ one-way                      │      │      │                                           │
    │                              │      │      │                                           │
    │   breaks assumptions of      │      │      │ Single Producer Multiple Consumer         │
    │   caller knowing when the    │──────▶      │                                           │
    │   request finished           │             │ Breaks shared span model, ex each consumer│
    │                              │             │ may accidentally use the same span ID     │
    │   depending on POV knowing   │             │ Harder to visualize if there are varied   │
    │   a request finished may     │             │ time scales per consumer. That said, each │
    │   be less important than if  │             │ consumer can still work as a child span   │
    │   a caller was blocked       │             │                                           │
    │                              │             │                                           │
    │                              │             │                                           │
    └──────────────────────────────┘             └───────────────────────────────────────────┘

    ┌─────────────────────────────────┐           ┌───────────────────────────────────────────┐
    │            Batching             │           │        Distributed Flow Processing        │
    ├─────────────────────────────────┤           ├───────────────────────────────────────────┤
    │                                 │           │                                           │
    │ Examples are coallesced writes  │           │ Inter-process Streams (like Spark)        │
    │ or flushing. A batch can be a   │           │                                           │
    │ separate trace with links added │           │ The user can define a join/zip/etc which  │
    │ for each enclosed element. The  │           │ may combine work started in multiple      │
    │ root is the batch. Each inserted│           │ traces. Phases tend to be coarse grained  │
    │ span is a direct child.         │           │ which maps to fewer spans. Challenge is   │
    │                                 │           │ if we can detect or assist in how to      │
    │ Representing the batch explains │           │ represent this (ex like we do in batching)│
    │ worst case latency like if there│           │                                           │
    │ was a long write ahead of you   │           │ How do we address latency of a data flow? │
    │                                 │           │                                           │
    │ Also could use correlation id   │           └──────────────────┬────────────────────────┘
    │ for the batch or another parent │                              │
    │                                 │                              │
    │                                 │                              │
    │                                 │                              │
    │                                 │                              │
    └─────────────────────────────────┘                              │
                                                                     │
                                                        ┌────────────┴───────────────────────┐
┌───────────────────────────────────────────┐           │                                    │
│   Intra-process Streams (like Reactor)    │           │Apis look the same, but challenges  │
├───────────────────────────────────────────┤           │can be different. For example,      │
│                                           │           │reactive streams define threading   │
│ Intra-process Streams (like Reactor)      │           │semantics which can be helpful in   │
│ Automatic instrumentation challenges:     │           │understanding how an application    │
│ Granularity challenges include small as   │     ┌─────┤behaves. Data flows have coarser    │
│ clock ticks. Infinite sequences exist.    ├─────┘     │granularity which maps easier to    │
│ No way to tell by type if a flux will end │           │separate spans.                     │
│ and many operators dont have metadata to  │           │                                    │
│ indicate if it is a grouping command or   │           │Both break models or visualizations │
│ not. Notable exception is parallel flux.  │           │that assume a tree shape (ex joins  │
│                                           │           │are possible in either). Both can be│
│ In cases where you create the source, ex  │           │infinite, though each process w/in  │
│ web flux, you can know for sure it is     │           │the flow has a beginning and end.   │
│ finite and how model spans on behalf of   │           │                                    │
│ requests.                                 │           └────────────────────────────────────┘
│                                           │
│ Manual instrumentation is possible where  │
│ you can materialize a context which will  │
│ be available at any point in the flux.    │
└───────────────────────────────────────────┘
```

So for example, using B3 approach, it would be immediately forking a child on receipt (ignoring X-B3-ParentId and using X-B3-SpanId as the parent). By doing this, the receiver doesn't need to know if there are any peers or not. Other propagation approaches such as [trace-context](https://github.com/TraceContext/tracecontext-spec) imply this, to such degree we [documented this behavior](https://github.com/openzipkin/b3-propagation#why-is-parentspanid-propagated) in the past.

Besides propagation, there are other concerns such as the difference between data delivery and processing. For example, we want to ack a message received as soon as we consume it, similar to how http servers work. However, @ImFlog noticed at least in Kafka, there are different apis for consuming messages and processing them. In fact you may not ever know what api is used to process a message. This means instrumentation may need to carry context on the data (ex in headers) and hold it until such time a processor acts upon it. One way it to complete the span as soon as the message is off the wire, but keep the span context on the message for a potential later processor to read as its parent. This stuff has some exploration needed.

Hence this issue forking off the subtopic of single-producer multiple-consumer tracing and how folks are able to work with it.

codefromthecrypt:
See https://github.com/openzipkin/openzipkin.github.io/pull/82 for docs
See https://github.com/openzipkin/zipkin-api/pull/29 for thrift constants

Here is what work in progress looks like for a single producer, single consumer POV where the remote endpoint (broker) is set to kafka:

```bash
$ curl -X POST -s localhost:9411/api/v1/spans -H 'Content-Type: application/json' -d '[{"traceId":"040c464a2b70a87a","id":"040c464a2b70a87a","name":"","timestamp":1501654414312417,"annotations":[{"timestamp":1501654414312417,"value":"ms","endpoint":{"serviceName":"producer","ipv4":"192.168.1.10"}}],"binaryAnnotations":[{"key":"kafka.key","value":"foo","endpoint":{"serviceName":"producer","ipv4":"192.168.1.10"}},{"key":"kafka.topic","value":"myTopic","endpoint":{"serviceName":"producer","ipv4":"192.168.1.10"}},{"key":"ma","value":true,"endpoint":{"serviceName":"kafka"}}]}]'
$ curl -X POST -s localhost:9411/api/v1/spans -H 'Content-Type: application/json' -d '[{"traceId":"040c464a2b70a87a","id":"b169190ffec10531","name":"","parentId":"040c464a2b70a87a","timestamp":1501654414816112,"annotations":[{"timestamp":1501654414816112,"value":"mr","endpoint":{"serviceName":"consumer","ipv4":"192.168.1.10"}}],"binaryAnnotations":[{"key":"ma","value":true,"endpoint":{"serviceName":"kafka"}}]}]'
```

screen shot 2017-08-02 at 2 14 41 pm

screen shot 2017-08-02 at 2 15 12 pm

Fixes #1654

### DIFF
--- a/zipkin-ui/js/component_ui/traceConstants.js
+++ b/zipkin-ui/js/component_ui/traceConstants.js
@@ -2,19 +2,23 @@ const CLIENT_SEND = 'cs';
 const CLIENT_SEND_FRAGMENT = 'csf';
 const CLIENT_RECEIVE = 'cr';
 const CLIENT_RECEIVE_FRAGMENT = 'crf';
+const MESSAGE_SEND = 'ms';
+const MESSAGE_RECEIVE = 'mr';
 const SERVER_SEND = 'ss';
 const SERVER_SEND_FRAGMENT = 'ssf';
 const SERVER_RECEIVE = 'sr';
 const SERVER_RECEIVE_FRAGMENT = 'srf';
 const SERVER_ADDR = 'sa';
 const CLIENT_ADDR = 'ca';
+const MESSAGE_ADDR = 'ma';
 const WIRE_SEND = 'ws';
 const WIRE_RECEIVE = 'wr';
 const ERROR = 'error';
 const LOCAL_COMPONENT = 'lc';
 const CORE_CLIENT = [CLIENT_RECEIVE, CLIENT_RECEIVE_FRAGMENT, CLIENT_SEND, CLIENT_SEND_FRAGMENT];
+const CORE_MESSAGE = [MESSAGE_SEND, MESSAGE_RECEIVE];
 const CORE_SERVER = [SERVER_RECEIVE, SERVER_RECEIVE_FRAGMENT, SERVER_SEND, SERVER_SEND_FRAGMENT];
-const CORE_ADDRESS = [CLIENT_ADDR, SERVER_ADDR];
+const CORE_ADDRESS = [CLIENT_ADDR, SERVER_ADDR, MESSAGE_ADDR];
 const CORE_WIRE = [WIRE_SEND, WIRE_RECEIVE];
 const CORE_LOCAL = [LOCAL_COMPONENT];
 const CORE_ANNOTATIONS = [...CORE_CLIENT, ...CORE_SERVER, ...CORE_WIRE, ...CORE_LOCAL];
@@ -23,13 +27,17 @@ export const Constants = {
   CLIENT_SEND_FRAGMENT,
   CLIENT_RECEIVE,
   CLIENT_RECEIVE_FRAGMENT,
+  MESSAGE_SEND,
+  MESSAGE_RECEIVE,
   SERVER_SEND,
   SERVER_SEND_FRAGMENT,
   SERVER_RECEIVE,
   SERVER_RECEIVE_FRAGMENT,
   SERVER_ADDR,
   CLIENT_ADDR,
+  MESSAGE_ADDR,
   CORE_CLIENT,
+  CORE_MESSAGE,
   CORE_SERVER,
   ERROR,
   LOCAL_COMPONENT,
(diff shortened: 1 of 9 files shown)


codefromthecrypt:
ok dependency linking is done. I'll look at UI next

PR #1684, opened 2017-08-10
Accepts Zipkin v2 Span format in all current transports

codefromthecrypt:
This accepts the json format from #1499 on current transports. It does
so by generalizing format detection from the two Kafka libraries, and
a new `SpanDecoder` interface. Types are still internal, but this allows
us to proceed with other work in #1644, including implementing reporters
in any language.

Concretely, you can send a json list of span2 format as a Kafka or Http
message. If using http, use the /api/v2/spans endpoint like so:

```bash
$ curl -X POST -s localhost:9411/api/v2/spans -H'Content-Type: application/json' -d'[{
  "id": "9032b04972e475c5",
  "kind": "SERVER",
  "name": "get",
  "timestamp": 1502101460678880,
  "duration": 612898,
  "localEndpoint": {
    "ipv4": "192.168.1.113"
  },
  "remoteEndpoint": {
    "serviceName": "",
    "ipv4": "127.0.0.1",
    "port": 60149
  },
  "tags": {
    "http.path": "/a"
  }
}]'
```

### DIFF
--- a/zipkin-collector/kafka/src/main/java/zipkin/collector/kafka/KafkaStreamProcessor.java
+++ b/zipkin-collector/kafka/src/main/java/zipkin/collector/kafka/KafkaStreamProcessor.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2016 The OpenZipkin Authors
+ * Copyright 2015-2017 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -13,10 +13,8 @@
  */
 package zipkin.collector.kafka;
 
-import java.util.Collections;
 import kafka.consumer.ConsumerIterator;
 import kafka.consumer.KafkaStream;
-import zipkin.Codec;
 import zipkin.collector.Collector;
 import zipkin.collector.CollectorMetrics;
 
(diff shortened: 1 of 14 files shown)


codefromthecrypt:
@openzipkin/core @bplotnick @llinder @aliostad @devinsba @flier @ellispritchard @mosesn @pavolloffay @clguimanMSFT @neilstevenson @goller @SergeyKanzhelev

So this change is pretty significant as it means folks can start using the new simplified json format defined in #1499 as soon as this is released (likely zipkin 1.30). In reality a lot of sites won't automatically upgrade to this, but it is important to know anyway.

Concretely, this means tracers can start offering the ability to send in the new format, ex via kafka (sqs azure hub etc) or POST /api/v2/spans. This format is the same as what we're using internally in Elasticsearch 6.x (#1674), and will be the default choice for document stores.

I'm cc'ing you for heads up in case you'd like to try using this, or update your libraries to support it. In the next few weeks a lot of stuff around this will be landing tracked in #1644. Watch or comment on that issue if you want others to know your status.

pavolloffay:
What is the format for annotations?

codefromthecrypt:
>
> What is the format for annotations?
>
We will publish an open api spec and it is noted in #1499, but annotations
are simple. It is the same as before except no endpoint.

{
  "traceId": "86154a4ba6e91385",
  "parentId": "86154a4ba6e91385",
  "id": "4d1e00c0db9010db",
  "kind": "CLIENT",
  "name": "get",
  "timestamp": 1472470996199000,
  "duration": 207000,
  "localEndpoint": {
    "serviceName": "frontend",
    "ipv4": "127.0.0.1"
  },
  "remoteEndpoint": {
    "serviceName": "backend",
    "ipv4": "192.168.99.101",
    "port": 9000
  },
  "annotations": [
    {
      "timestamp": 1472470996238000,
      "value": "ws"
    },
    {
      "timestamp": 1472470996403000,
      "value": "wr"
    }
  ],
  "tags": {
    "http.path": "/api",
    "clnt/finagle.version": "6.45.0"
  }
}

codefromthecrypt:
not merging this, yet due to https://github.com/openzipkin/zipkin/issues/1499#issuecomment-321719394

codefromthecrypt:
ok ES indexing problem has been resolved. will merge on green

pvprsd:
Hi,

I am new to the zipkin. I have started using zipkin from docker image from docker hub i.e. openzipkin/zipkin:latest .
I found that the /spans POST API is not supported by default. Do we need to enable it explicitly?
Is this supported zipkin v2?

Thanks,
Prasad

shakuzen:
@pvprsd please join us on Gitter (https://gitter.im/openzipkin/zipkin) if you have questions.

PR #1686, opened 2017-08-12
Allows span2 json format to omit serviceName instead of setting empty

codefromthecrypt:
A snag in the old model was that we needed to pass an empty string for
serviceName because thrift requires the field. In span2 format, we can
act more idiomatic and either leave it out or set it to null.


codefromthecrypt:
this gets rid of empty service names in ES 6.x

PR #1696, opened 2017-08-21
Starts preparing v2 Codec package

codefromthecrypt:
All current Zipkin codec operations include the following:
* encoding spans one-by-one as reported by instrumentation
* encoding a list of spans ready to send into a bytes message
* decoding a list of spans off a transport message

Zipkin v1 includes legacy encoding, which is a bit different. For
example, old variants of Kafka accepted single-element messages instead
of a list. We don't need to carry-forward this anymore. Moreover,
interfaces of reporter vs collector were split across jars in Zipkin v1
due to the former being defined late.

This introduces the following key interfaces and cleans up some code
around codec.

* `byte[] Encoder.encode(S span)`
* `M MessageEncoder.encode(List<byte[]> encodedSpans)`
* `List<S> Decoder.decodeList(byte[] message)`

All of these are temporarily placed in an internal package until other
code around it stabilizes.

### DIFF
--- a/zipkin/src/main/java/zipkin/internal/v2/codec/Encoder.java
+++ b/zipkin/src/main/java/zipkin/internal/v2/codec/Encoder.java
@@ -1,0 +1,44 @@
+/**
+ * Copyright 2015-2017 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin.internal.v2.codec;
+
+import zipkin.Span;
+import zipkin.internal.JsonCodec;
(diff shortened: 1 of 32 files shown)


zipkin/src/main/java/zipkin/internal/v2/codec/Encoder.java
```diff
@@ -0,0 +1,44 @@
...
+import zipkin.Span;
+import zipkin.internal.JsonCodec;
+import zipkin.internal.Span2;
+
+import static zipkin.internal.v2.codec.Span2JsonAdapters.SPAN_WRITER;
+
+/**
+ * @param <S> type of the span, usually {@link Span}
+ */
+public interface Encoder<S> {
```
codefromthecrypt, line 25: This, Encoding and MessageEncoder were taken from zipkin-reporter-java, which afterwards can be a place to host just the implementations

zipkin/src/test/java/zipkin/internal/Span2Test.java
```diff
@@ -119,19 +146,17 @@
...
-    new ObjectOutputStream(buffer.outputStream()).writeObject(base);
 
-    assertThat(new ObjectInputStream(buffer.inputStream()).readObject())
-      .isEqualTo(base);
-  }
+    Span2 span = base.toBuilder()
+      .addAnnotation(1L, "foo")
+      .build();
 
-  @Test public void serializationUsesJson() throws Exception {
```
codefromthecrypt, line 130: removed requirement to use json as serialization as it isn't important and adds code. people can use whatever they want, like kryo

PR #1697, opened 2017-08-21
Switches CollectorSampler to use (trace ID, debug) vs Span v1 object

codefromthecrypt:
Currently, transports can decode zipkin2 format, but it has to be
converted to v1 objects prior to storage. Part of the reason is
CollectorSampler (used for load shedding to protect storage), accepts
an object type as opposed to the parameters trace ID and debug. This
deprecates the former in favor of the latter, as trace ID and debug
exist in both formats.
We are switching code to permit use of zipkin2 format in transports.

### DIFF
--- a/zipkin/src/test/java/zipkin/collector/CollectorTest.java
+++ b/zipkin/src/test/java/zipkin/collector/CollectorTest.java
@@ -13,127 +13,130 @@
  */
 package zipkin.collector;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.logging.Level;
-import java.util.logging.Logger;
+import org.junit.Before;
 import org.junit.Test;
 import zipkin.Span;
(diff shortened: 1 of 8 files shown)


zipkin/src/test/java/zipkin/collector/CollectorTest.java
```diff
@@ -13,127 +13,130 @@
...
 import org.junit.Test;
 import zipkin.Span;
 import zipkin.internal.ApplyTimestampAndDuration;
 import zipkin.storage.Callback;
-import zipkin.storage.InMemoryStorage;
+import zipkin.storage.StorageComponent;
 
 import static java.util.Arrays.asList;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
```
codefromthecrypt, line 25: converted this to mock because a near future change will make this type abstract

PR #1698, opened 2017-08-22
Switches Elasticsearch to only write single-type indexes

codefromthecrypt:
Multi-type indexes are no longer supported starting with Elasticsearch
6. We added support for single-type writes in Zipkin 1.29 as an opt-in
functionality via `ES_EXPERIMENTAL_SPAN2=true`. This makes single-
type writes the only means to ingest spans, eventhough we still read
both models.

To turn off dual-reads, set `ES_LEGACY_READS_ENABLED=false` once your
collectors have been updated to Zipkin 1.31 and old indexes are no longer
valuable.


codefromthecrypt:
cc @openzipkin/elasticsearch

codefromthecrypt:
here's the double-conversion which is easier to obviate when we know how storage happens by type https://github.com/openzipkin/zipkin/pull/1700

PR #1705, opened 2017-08-24
Adds v2 Call interface based on okhttp and retrofit

codefromthecrypt:
In Zipkin v1, we had a pair of SPI interfaces: `SpanConsumer`
and `AsyncSpanConsumer`.

For example, this would define the same thing twice:

```java
void SpanConsumer.accept(List<Span> spans);

// and

void AsyncSpanConsumer.accept(List<Span> spans, Callback<Void> spans);
```

in v2, it looks like this (inspired by Retrofit and OkHttp's call):

```java
Call<Void> accept(List<Span> spans);

// then make a call like this..
call = spanConsumer.accept(spans);

// to do a blocking call..
call.execute();

// to do an async call..
call.enqueue(callback);

// to cancel
call.cancel();
```

The important part of this design is that it reduces the redundant
interfaces. We end up making less work for users and implementors
at the cost of an intermediary type: Call.

This design is near identical to [Retrofit's Call](https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit2/Call.java), which is itself similar to
[OkHttp's call](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Call.java). The mild difference here is that the underlying remote
work may not always be http even if the first impl (elasticsearch)
happens to be.

Many thanks to @JakeWharton and @swankjesse for the prior art.

### DIFF
--- a/zipkin/src/test/java/zipkin/internal/v2/CallTest.java
+++ b/zipkin/src/test/java/zipkin/internal/v2/CallTest.java
@@ -1,0 +1,154 @@
+/**
+ * Copyright 2015-2017 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin.internal.v2;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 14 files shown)


codefromthecrypt:
This change still needs polish but raising for early feedback cc @anuraaga @shakuzen @llinder

codefromthecrypt:
Did significant polishing on flight, but needs unit tests

swankjesse (approved):
Neat.

zipkin/src/test/java/zipkin/internal/v2/CallTest.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }
+
+  @Test public void enqueuesOnce() throws Exception {
+    Call<Void> call = Call.create(null);
+    call.enqueue(callback);
+
+    try {
+      call.enqueue(callback);
+      failBecauseExceptionWasNotThrown(IllegalStateException.class);
```
anuraaga, line 107: Since you're using assertj, just wondering any reason not to use 

`assertThatThrownBy(() -> call.enqueue(callback)).isInstanceOf(IllegalStateException.class);`

codefromthecrypt:
Will play with that in next change to the file. Thx for the tip!

adriancole:
here's follow-ups needed to compete the read side of elasticsearch storage

codefromthecrypt:
here's follow-ups needed to compete the read side of elasticsearch storage https://github.com/openzipkin/zipkin/pull/1708

PR #1708, opened 2017-08-27
Adds call utilities commonly needed

codefromthecrypt:
This adds `Call.emptyList()` needed for almost all span store operations
and `Call.flatMap()` needed to transform an ID list into a get by ID
call.

### DIFF
--- a/zipkin/src/main/java/zipkin/internal/v2/Call.java
+++ b/zipkin/src/main/java/zipkin/internal/v2/Call.java
@@ -14,7 +14,8 @@
 package zipkin.internal.v2;
 
 import java.io.IOException;
-import javax.annotation.Nullable;
+import java.util.Collections;
+import java.util.List;
 
 /**
  * This captures a (usually remote) request and can be used once, either {@link #execute()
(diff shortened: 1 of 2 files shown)


zipkin/src/test/java/zipkin/internal/v2/CallTest.java
```diff
@@ -151,4 +136,83 @@ public void concurrent_executesOrSubmitsOnce() throws InterruptedException {
...
+
+  @Test public void emptyList_independentInstances() throws Exception {
+    assertThat(Call.emptyList())
+      .isNotSameAs(Call.emptyList());
+  }
+
+  @Test public void flatMap_execute() throws Exception {
+    Call<String> fooCall = Call.create("foo");
+    Call<String> barCall = Call.create("bar");
+    Call<String> fooBarCall = fooCall.flatMap(foo -> barCall);
```
anuraaga: Assert foo is "foo" (or could do something like `foo.equals("foo") ? barCall : badCall`)

zipkin/src/main/java/zipkin/internal/v2/Call.java
```diff
@@ -47,6 +49,37 @@
  * @param <V> the success type, only null when {@code V} is {@linkplain Void}.
  */
 public abstract class Call<V> implements Cloneable {
+  static final Call CALL_EMPTY_LIST = Call.create(Collections.emptyList());
+
+  @SuppressWarnings("unchecked")
+  public static <T> Call<List<T>> emptyList() {
+    return (Call<List<T>>) CALL_EMPTY_LIST.clone();
```
anuraaga: I believe it would be the same exact logic if you use `Call.create(Collections.emptyList())` here instead of `clone`, but would let you avoid having the extra static constant.

adriancole:
Good advice

On 27 Aug 2017 16:38, "Anuraag Agrawal" <notifications@github.com> wrote:

*@anuraaga* commented on this pull request.
------------------------------

In zipkin/src/test/java/zipkin/internal/v2/CallTest.java
<https://github.com/openzipkin/zipkin/pull/1708#discussion_r135405822>:

> +
+  @Test public void emptyList() throws Exception {
+    Call<List<String>> call = Call.emptyList();
+
+    assertThat(call.execute()).isEmpty();
+  }
+
+  @Test public void emptyList_independentInstances() throws Exception {
+    assertThat(Call.emptyList())
+      .isNotSameAs(Call.emptyList());
+  }
+
+  @Test public void flatMap_execute() throws Exception {
+    Call<String> fooCall = Call.create("foo");
+    Call<String> barCall = Call.create("bar");
+    Call<String> fooBarCall = fooCall.flatMap(foo -> barCall);

Assert foo is "foo" (or could do something like foo.equals("foo") ? barCall
: badCall)
------------------------------

In zipkin/src/main/java/zipkin/internal/v2/Call.java
<https://github.com/openzipkin/zipkin/pull/1708#discussion_r135405888>:

> @@ -47,6 +49,37 @@
  * @param <V> the success type, only null when {@code V} is {@linkplain Void}.
  */
 public abstract class Call<V> implements Cloneable {
+  static final Call CALL_EMPTY_LIST = Call.create(Collections.emptyList());
+
+  @SuppressWarnings("unchecked")
+  public static <T> Call<List<T>> emptyList() {
+    return (Call<List<T>>) CALL_EMPTY_LIST.clone();

I believe it would be the same exact logic if you use
Call.create(Collections.emptyList()) here instead of clone, but would let
you avoid having the extra static constant.

—
You are receiving this because you authored the thread.
<https://github.com/openzipkin/zipkin/pull/1708#pullrequestreview-58817026>,
or mute the thread

zipkin/src/test/java/zipkin/internal/v2/CallTest.java
codefromthecrypt: I added this to the first test that invokes a map or flatMap (as doing for all tests would be distracting)

codefromthecrypt:
thx @anuraaga

PR #1709, opened 2017-08-27
Adds v2 SpanStore and a bridging adapter to v1

codefromthecrypt:
This represents the minimal read apis needed for Zipkin

As discussed earlier on gitter w/ @basvanbeek, we no longer do
adjustment at storage layer. Essentially, all operations are "raw",
which reduces the responsibility of implementors. This also midly
changes the `getTrace` call, disallowing it from returning null (as
empty will suffice).

This adds SpanStore which returns Call objects as opposed to replicating
signatures for synchronous and asynchronous invocations. Unlike before,
this does not support "adjustments" rather assumes they happen at a
layer above storage. This decision refines the storage query api to the
following:

```java
Call<List<List<Span>>> getTraces(QueryRequest request);

Call<List<Span>> getTrace(long traceIdHigh, long traceIdLow);

Call<List<String>> getServiceNames();

Call<List<String>> getSpanNames(String serviceName);

Call<List<DependencyLink>> getDependencies(long endTs, long lookback);
```

A later change might replace the getTrace call with a hex string 
parameter.

### DIFF
--- a/zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpStorage.java
+++ b/zipkin-storage/elasticsearch-http/src/main/java/zipkin/storage/elasticsearch/http/ElasticsearchHttpStorage.java
@@ -28,12 +28,12 @@ import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.RequestBody;
 import okio.Buffer;
+import zipkin.internal.LenientDoubleCallbackAsyncSpanStore;
 import zipkin.internal.V2StorageComponent;
 import zipkin.internal.v2.storage.SpanConsumer;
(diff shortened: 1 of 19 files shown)


codefromthecrypt:
PS once this is in, the v2 http api (openapi/swagger etc) can be implemented. Later, we can formalize and make the java libraries not-internal.

codefromthecrypt:
ok all apis are here, and I'll raise two other PRs: one to add the http api and another to implement with elasticsearch. There's still a solid amount of work on this PR, yet, which is to backfill tests and make a v2 replacement for SpanStoreTest

One thing you'll notice is I made a copy of the bridging read api from elasticsearch to here. This is to decouple change. The elasticsearch copy will be deleted later.

cc @anuraaga

codefromthecrypt:
OK backfilled almost all tests. nearly there

PR #1710, opened 2017-08-28
Adds v2 http query api

codefromthecrypt:
This adds the v2 read endpoints. Swagger will come later.


adriancole:
this should be just about ready now. I'll make openapi files to verify

PR #1717, opened 2017-08-30
Adds Call.handleError to address remote resource not found

codefromthecrypt:
To complete an http span store adapter, we need to be able to coerce
404 to empty. This does that like this:

```java
call.handleError((error, callback) -> {
  if (error instanceof HttpException && ((HttpException) error).code == 404) {
    callback.onSuccess(Collections.emptyList());
  } else {
    callback.onError(error);
  }
});
```


codefromthecrypt:
This is the last function needed to complete the v2 http api (the junit rule tests the api impl with okhttp) cc @anuraaga @shakuzen

codefromthecrypt:
will take feedback post-commit, even renames and such. won't release this immediately, but need the function to proceed.

PR #1722, opened 2017-09-01
Redirect /config.json to /zipkin/config.json in order to work with the chrome extension

SimenB:
For some reason the path is cut off when added to the chrome extension, meaning I can't specify`/zipkin` there.

Hopefully the extension follows redirects...

https://github.com/openzipkin/zipkin-browser-extension


eirslett:
The extension could be updated to support both... check one and fallback to the other

SimenB:
Probably a better idea, I just noticed that the URLs in the table generated by the plugin are missing `/zipkin` as well

codefromthecrypt:
@SimenB up for updating the plugin?

SimenB:
I can take a whack at it, sure. How should we check if `/zipkin` path is needed? Make a request to root, and if we're redirected, use that?

codefromthecrypt:
you could check /zipkin/config.json first and if 404 try /config.json

SimenB:
Cool, I'll get a PR out soonish. I'll try to fix the issue I reported as well

eirslett:
A bit tricky if /zipkin/config.json returns index.html though... Maybe it's possible to check the content-type header, or check if the payload is valid json...

PR #1723, opened 2017-09-02
Adds property for X-Forwarded handling on redirect

codefromthecrypt:
See https://github.com/openzipkin/zipkin/pull/1627#issuecomment-326702774
See https://github.com/spring-projects/spring-boot/issues/6232

### DIFF
--- a/zipkin-server/src/main/resources/zipkin-server-shared.yml
+++ b/zipkin-server/src/main/resources/zipkin-server-shared.yml
@@ -130,6 +130,7 @@ zipkin:
 
 server:
   port: ${QUERY_PORT:9411}
+  use-forward-headers: true
   compression:
     enabled: true
     # compresses any response over min-response-size (default is 2KiB)
--- a/zipkin-server/src/test/java/zipkin/server/ZipkinServerIntegrationTest.java
+++ b/zipkin-server/src/test/java/zipkin/server/ZipkinServerIntegrationTest.java
@@ -14,17 +14,20 @@
 package zipkin.server;
 
 import java.util.List;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
 import okio.Buffer;
 import okio.GzipSink;
 import org.junit.Before;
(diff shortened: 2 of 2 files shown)


codefromthecrypt:
cc @naoman who raised the issue. I'm not sure if it is a simple workaround to have ELB do the path forwarding or not.

Any @shakuzen @wilkinsona do you know latest on X-Forwarded-For on spring boot? we are using default container on 1.5.6.RELEASE

codefromthecrypt:
fwiw I tried adding the following to no avail:

```
server:
  tomcat:
    protocol-header: X-Forwarded-Proto
    remote-ip-header: X-Forwarded-For
    port-header: X-Forwarded-Port
    protocol-header-https-value: https
```

Here's the redirect controller:
```java
  /** Make sure users who aren't familiar with /zipkin get to the right path */
  @RequestMapping(value = "/", method = GET)
  public ModelAndView redirectRoot() {
    return new ModelAndView("redirect:/zipkin/");
  }
```

wilkinsona:
Looks like you're [missing `server.use-forward-headers: true`](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#howto-use-tomcat-behind-a-proxy-server)

codefromthecrypt:
@wilkinsona thanks.. almost there.. just `X-Forwarded-Host` isn't being picked up.. not really sure which setting would accomadate that.

```
server:
  use-forward-headers: true
  tomcat:
    protocol-header: X-Forwarded-Proto
    remote-ip-header: X-Forwarded-For
    port-header: X-Forwarded-Port
    protocol-header-https-value: https
```

```
org.junit.ComparisonFailure: 
Expected :"https://zipkin.com:443/zipkin/"
Actual   :"https://localhost/zipkin/"
```

codefromthecrypt:
I've removed `X-Forwarded-Host` from the test as `Host` is more typical anyway.

codefromthecrypt:
Thanks @wilkinsona!

PR #1728, opened 2017-09-08
Teaches the v2 tree how to encode v1 format

codefromthecrypt:
The zipkin v2 library is intended to be used by itself in tracers. Since
tracers cannot control the span transport, it likely will need to write
the old (historical) json format for at least months. To avoid requiring
dual versions of jars to accomplish this, we add an encoder that writes
json v1 format given a v2 span.

Note: we don't decode as the library problem is easier to manage on the
server side. For example, zipkin-server will need both versions of the
jars for things like collection for a while.


codefromthecrypt:
here are the benchmarks. While slower than zipkin2's native format, the v2 encoder is faster writing v1 format than the v1 encoder because the v2 span model has only one endpoint, and so we can reliably cache it when we need to write annotations and binary annotations.

```
Benchmark                                                Mode  Cnt   Score   Error  Units
CodecBenchmarks.readClientSpan_json_zipkin               avgt   15   8.497 ± 0.255  us/op
CodecBenchmarks.readClientSpan_json_zipkin2              avgt   15   4.449 ± 0.249  us/op
CodecBenchmarks.readClientSpan_thrift_libthrift          avgt   15   2.145 ± 0.069  us/op
CodecBenchmarks.readClientSpan_thrift_zipkin             avgt   15   2.220 ± 0.087  us/op
CodecBenchmarks.readTenClientSpans_json_zipkin           avgt   15  68.368 ± 3.120  us/op
CodecBenchmarks.readTenClientSpans_json_zipkin2          avgt   15  31.352 ± 1.215  us/op
CodecBenchmarks.readTenClientSpans_thrift_zipkin         avgt   15  21.185 ± 0.243  us/op
CodecBenchmarks.writeClientSpan_json_zipkin              avgt   15   3.024 ± 0.106  us/op
CodecBenchmarks.writeClientSpan_json_zipkin2             avgt   15   1.079 ± 0.073  us/op
CodecBenchmarks.writeClientSpan_json_zipkin2_legacy      avgt   15   1.487 ± 0.043  us/op
CodecBenchmarks.writeClientSpan_thrift_libthrift         avgt   15   2.824 ± 0.080  us/op
CodecBenchmarks.writeClientSpan_thrift_zipkin            avgt   15   0.630 ± 0.013  us/op
CodecBenchmarks.writeTenClientSpans_json_zipkin          avgt   15  29.569 ± 0.968  us/op
CodecBenchmarks.writeTenClientSpans_json_zipkin2         avgt   15  10.343 ± 0.419  us/op
CodecBenchmarks.writeTenClientSpans_json_zipkin2_legacy  avgt   15  14.023 ± 0.732  us/op
CodecBenchmarks.writeTenClientSpans_thrift_zipkin        avgt   15   6.317 ± 0.126  us/op
```

Issue #1731, opened 2017-09-12
Allow customization of UI base tag for running behind proxies


stepanv:
This is about introducing `ZIPKIN_UI_BASEPATH` so that proxies can naturally serve the correct base path for zipkin UI static content.

Original desc below. If more is needed follow-up issue should be raised.
---
It would be neat if Zipkin didn't enforce `/zipkin/` absolute path for the UI, but was able to run in variable context root such as `/api/v1/namespaces/default/services/zipkin-service/proxy/zipkin/` (Kubernetes apiserver proxy feature).

This is related to #1627 but it doesn't enforce the `zipkin` folder to be in a _root_.

I'm submitting a pull request.

eirslett:
Unfortunately this is more complicated than you would think, because of the web UI is a single page webapp, and that comes with some limitations. We chose to hardcode it to /zipkin/ as a compromise between usability and code complexity.

stepanv:
You're right it adds complexity. As you can see from the linked merge request, `/` (as a root) has to be substituted with the _context path_ where absolute paths are used.

Maybe it would be worth replacing the `__webpack_public_path__` with something else but in general I think the idea itself doesn't result with too complex code. Let me know what you think.

wdittmer:
Was typing the stuff below and then found it was probably the same as this issue.
Below are my findings, which might help someone else to see if there can be done something to make this work.

This issue is related to the work already done for #1229 and #1627 .
What we are trying to do is the following:
We have build up a Kubernetes cluster where we run the openzipkin/zipkin service and an ingress controller to reach various services like Kibana, Grafana, ElasticSearch and also zipkin.

The ingress controller looks a follows:
```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: cbo-ingress-mgmt-elb
  namespace: monitoring
  labels:
    app: ingressservice-mgmt-elb
    dns: route53
  annotations:
    # kubernetes.io/tls-acme: "true"
    kubernetes.io/ingress.class: "nginx"
    ingress.kubernetes.io/rewrite-target: /
    ingress.kubernetes.io/ssl-redirect: "false"
    domainName: "mgmt.$DOMAIN"
spec:
  rules:
  - host: management.${CNF_ENV}.cbo
    http:
      paths:
      - backend:
          serviceName: alertmanager
          servicePort: 9093
        path: /alertmanager
      - backend:
          serviceName: elasticsearch
          servicePort: 9200
[...]

PR #1732, opened 2017-09-12
Support for running Zipkin behind a reverse proxy

stepanv:
- subfolder 'zipkin' is enforced
 - rewrite of the 'base' tag in the 'index.html' file is required

This implements the #1731

### DIFF
--- a/zipkin-ui/js/component_data/default.js
+++ b/zipkin-ui/js/component_data/default.js
@@ -21,7 +21,8 @@ export default component(function DefaultData() {
     const query = convertToApiQuery(window.location.search);
     const serviceName = query.serviceName;
     if (serviceName) {
-      const apiURL = `/zipkin/api/v1/traces?${queryString.stringify(query)}`;
+      // eslint-disable-next-line camelcase, no-undef
+      const apiURL = `${__webpack_public_path__}api/v1/traces?${queryString.stringify(query)}`;
       $.ajax(apiURL, {
         type: 'GET',
         dataType: 'json'
--- a/zipkin-ui/js/component_data/default.js
+++ b/zipkin-ui/js/component_data/default.js
@@ -3,6 +3,7 @@ import {errToStr} from '../../js/component_ui/error';
 import $ from 'jquery';
 import queryString from 'query-string';
 import {traceSummary, traceSummariesToMustache} from '../component_ui/traceSummary';
+import {contextRoot} from '../publicPath';
 
 export function convertToApiQuery(windowLocationSearch) {
   const query = queryString.parse(windowLocationSearch);
@@ -21,8 +22,7 @@ export default component(function DefaultData() {
     const query = convertToApiQuery(window.location.search);
     const serviceName = query.serviceName;
     if (serviceName) {
-      // eslint-disable-next-line camelcase, no-undef
-      const apiURL = `${__webpack_public_path__}api/v1/traces?${queryString.stringify(query)}`;
+      const apiURL = `${contextRoot}api/v1/traces?${queryString.stringify(query)}`;
       $.ajax(apiURL, {
         type: 'GET',
         dataType: 'json'
--- a/zipkin-ui/js/component_data/default.js
+++ b/zipkin-ui/js/component_data/default.js
@@ -3,7 +3,6 @@ import {errToStr} from '../../js/component_ui/error';
 import $ from 'jquery';
 import queryString from 'query-string';
 import {traceSummary, traceSummariesToMustache} from '../component_ui/traceSummary';
-import {contextRoot} from '../publicPath';
 
 export function convertToApiQuery(windowLocationSearch) {
   const query = queryString.parse(windowLocationSearch);
@@ -22,7 +21,7 @@ export default component(function DefaultData() {
     const query = convertToApiQuery(window.location.search);
     const serviceName = query.serviceName;
     if (serviceName) {
-      const apiURL = `${contextRoot}api/v1/traces?${queryString.stringify(query)}`;
+      const apiURL = `api/v1/traces?${queryString.stringify(query)}`;
       $.ajax(apiURL, {
         type: 'GET',
         dataType: 'json'
--- a/zipkin-ui/js/publicPath.js
+++ b/zipkin-ui/js/publicPath.js
@@ -3,7 +3,8 @@ import $ from 'jquery';
 // read the public path from the <base> tag where it has to be set anyway because of
 // html-webpack-plugin limitations: https://github.com/jantimon/html-webpack-plugin/issues/119
 // otherwise it could be: window.location.pathname.replace(/(.*)\/zipkin\/.*/, '$1/zipkin/')
-const contextRoot = $('base').attr('href');
+let contextRoot = $('base').attr('href') || '/zipkin/';
+contextRoot += contextRoot.endsWith('/') ? '' : '/';
 
 // set dynamically 'output.publicPath' as per https://webpack.github.io/docs/configuration.html#output-publicpath
 __webpack_public_path__ = contextRoot; // eslint-disable-line camelcase, no-undef
--- a/zipkin-ui/js/publicPath.js
+++ b/zipkin-ui/js/publicPath.js
@@ -1,0 +1,11 @@
+import $ from 'jquery';
+
+// read the public path from the <base> tag where it has to be set anyway because of
+// html-webpack-plugin limitations: https://github.com/jantimon/html-webpack-plugin/issues/119
+// otherwise it could be: window.location.pathname.replace(/(.*)\/zipkin\/.*/, '$1/zipkin/')
+const contextRoot = $('base').attr('href');
+
+// set dynamically 'output.publicPath' as per https://webpack.github.io/docs/configuration.html#output-publicpath
+__webpack_public_path__ = contextRoot; // eslint-disable-line camelcase, no-undef
+
+export {contextRoot};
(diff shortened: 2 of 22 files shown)


codefromthecrypt:
Thanks for taking a stab. Can you elaborate a little on how this leads to reverse proxy support, how it is tested, and if there are any README updates or otherwise needed? (ex in zipkin-ui for running locally)

stepanv:
My use-case is quite simple, I need to run Zipkin in a different _path_ than just `/zipkin/`.

The best solution would be to have all static links relative and let Javascript calculate the current url and not rely on `/zipkin/`.
I was able to implement the latter but the former was not that easy due to limitations of the `html-webpack-plugin` (see comments in the code).

With the changes I'm proposing, the good news is that if the reverse proxy supports rewriting of the static content (`base` tag and its `href` attribute) (e.g., Apache Httpd with enabled `ProxyHTMLURLMap` for the `base` tag, see https://httpd.apache.org/docs/2.4/mod/mod_proxy_html.html#proxyhtmlurlmap), the Zipkin works in any path such as `/foo/bar/zipkin/`.

It is of course possible to run Zipkin the same as before at the path `/zipkin/`, so this is not a breaking change.

Regarding tests and readme, I understand that both are important but I wanted to know how you feel about this change before updating the Docs. There are no tests ensuring that this functionality is preserved in future development. I propose to implement a single sanity integration test that Zipkin loads and that most common links work under a different path.

codefromthecrypt:
cc @naoman @devinsba I think you are also running under a proxy. feedback?

eirslett:
How will this work with the browser plugin for Zipkin?

codefromthecrypt:
cc + @SimenB for testing relating to the browser plugin

stepanv:
By the way, as for the test error reported by _CircleCI_:
```
[ERROR] Tests run: 14, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 6.503 s <<< FAILURE! - in zipkin.junit.ZipkinRuleTest
[ERROR] postSpans_disconnectDuringBody(zipkin.junit.ZipkinRuleTest)  Time elapsed: 5.028 s  <<< ERROR!
java.io.IOException: unexpected end of stream on Connection{localhost:46744, proxy=DIRECT hostAddress=localhost/127.0.0.1:46744 cipherSuite=none protocol=http/1.1}
	at zipkin.junit.ZipkinRuleTest.postSpans(ZipkinRuleTest.java:276)
	at zipkin.junit.ZipkinRuleTest.postSpans_disconnectDuringBody(ZipkinRuleTest.java:197)
Caused by: java.io.EOFException: \n not found: limit=0 content=…
	at zipkin.junit.ZipkinRuleTest.postSpans(ZipkinRuleTest.java:276)
	at zipkin.junit.ZipkinRuleTest.postSpans_disconnectDuringBody(ZipkinRuleTest.java:197)
```

I'm unable to reproduce that locally. Isn't it possible it's an intermittently failing test? If so, are you able to instruct _CircleCI_ to re-run the test/build? In the worst case I could add a dummy commit to re-run it.

codefromthecrypt:
sorry about the flake. kicked

devinsba:
Our reverse proxy setup is direct, we aren't modifying the paths at all so this won't affect me at the moment. But..

Can you put a blurb in the readme about the technical requirements for this to work? Also a sample apache config would probably be good. Thanks

eirslett:
I would say the implementation itself isn't too bad, I imagined it would be more complex. Perhaps we should follow the rule-of-three?

naoman:
@adriancole similar to @devinsba, we don't modify the path and are not planning to any time soon.

stepanv:
To answer the question regarding the _Apache Httpd_ configuration for the reverse proxying. 

(For now I'm not adding it to a readme, let me know if you have any suggestions where to put the information regarding reverse proxying.)

Lets assume we want to access Zipkin (running on `localhost:9411`) through Apache Httpd server (running on `localhost:80` as a reverse proxy) in a context root `/test/foo/zipkin/`. No extra Zipkin configuration is required.

Apache Httpd configuration:
```
LoadModule proxy_module libexec/apache2/mod_proxy.so
LoadModule proxy_html_module libexec/apache2/mod_proxy_html.so
LoadModule proxy_http_module libexec/apache2/mod_proxy_http.so

ProxyPass /test/foo/ http://localhost:9411/
SetOutputFilter proxy-html
ProxyHTMLURLMap /zipkin/ /test/foo/zipkin/
ProxyHTMLLinks  base        href
```
To access Zipkin UI behind the reverse proxy,  execute:
```
$ curl http://localhost/test/foo/zipkin/
<html><head><!--
      add 'base' tag to work around the fact that 'html-webpack-plugin' does not work
      with '__webpack_public_path__' being set as reported at https://github.com/jantimon/html-webpack-plugin/issues/119
    --><base href="/test/foo/zipkin/"><link rel="icon" type="image/x-icon" href="favicon.ico"><meta charset="UTF-8"><title>Webpack App</title><link href="app-94a6ee84dc608c5f9e66.min.css" rel="stylesheet"></head><body>
  <script type="text/javascript" src="app-94a6ee84dc608c5f9e66.min.js"></script></body></html>
```
As you can see, the attribute `href` of the `base` tag is rewritten which is the way to get around the `html-webpack-plugin` limitations.

Uploading the span is easy as
```
curl -H "Content-Type: application/json" --data-binary "[$(cat /path/to/zipkin/benchmarks/src/main/resources/span-local.json)]" http://localhost/test/foo/api/v1/spans
```

And then it's of course observable in the UI:
```
open http://localhost/test/foo/zipkin/?serviceName=zipkin-server&startTs=1378193040000&endTs=1505463856013
```

stepanv:
By the way, in the latest commit I got rid of the `__webpack_public_path__ ` everywhere. Instead, a `publicPath` module is used to import `contextRoot` variable that is referenced in all the URIs.

eirslett:
How much effort would it take to fix the issue in html-webpack-plugin?

> By the way, in the latest commit I got rid of the __webpack_public_path__ everywhere. Instead, a publicPath module is used to import contextRoot variable that is referenced in all the URIs.

Sounds like a good idea, so we don't leak webpack-specific stuff everywhere in the code base

zipkin-ui/js/publicPath.js
```diff
@@ -0,0 +1,11 @@
+import $ from 'jquery';
+
+// read the public path from the <base> tag where it has to be set anyway because of
+// html-webpack-plugin limitations: https://github.com/jantimon/html-webpack-plugin/issues/119
+// otherwise it could be: window.location.pathname.replace(/(.*)\/zipkin\/.*/, '$1/zipkin/')
+const contextRoot = $('base').attr('href');
```
eirslett: Should we sanity check that it ends with `/`? Looks like the rest of the code assumes that. We could either show an error to the user, or just append `/` here?

zipkin-ui/js/component_data/default.js
```diff
@@ -21,7 +22,7 @@ export default component(function DefaultData() {
     const query = convertToApiQuery(window.location.search);
     const serviceName = query.serviceName;
     if (serviceName) {
-      const apiURL = `/zipkin/api/v1/traces?${queryString.stringify(query)}`;
+      const apiURL = `${contextRoot}api/v1/traces?${queryString.stringify(query)}`;
```
eirslett: There's a lot of code duplication going on here... maybe we should make a `apiClient.js` which basically just accepts relative paths and wraps `$.ajax`?

zipkin-ui/js/publicPath.js
stepanv: You're right, I'll append `/` if missing.

zipkin-ui/js/component_data/default.js
stepanv: Seems reasonable. I'll take a look at it.

stepanv:
@eirslett, to be honest, I wanted to avoid even looking at the `html-webpack-plugin` codebase. Even if it was easy, it would take weeks to have the feature I need in Zipkin working. 
The `base` tag is not an ugly work around, it's a way to go, clean and simple, that happens to work well even for the `html-webpack-plugin`. It puts a constraint on the reverse proxy, to rewrite the tag, but that's such a common behavior. For me it's a reasonable trade-off.

stepanv:
And the fact that `html-webpack-plugin` devs don't bother that it doesn't conform with the webpack spec (https://github.com/jantimon/html-webpack-plugin/issues/119) is quite discouraging for me, someone who's not from the JS world..

zipkin-ui/js/publicPath.js
stepanv: Done

zipkin-ui/js/component_data/default.js
stepanv: Done. As a matter of fact, the `$.ajax` call with relative paths use the `<base>` tag and as such there was no need to wrap it as you suggested.

stepanv:
I pushed new changes. 
If you think it's ok, please let me know what kind of tests and documentation you require in order to accept the pull request.

codefromthecrypt:
sorry about the delay. trying now. I really like this, actually. a lot tighter than I expected it to be.

codefromthecrypt:
tested and looks very good. I'd add your comments about how to run in apache as an README.md section.

A favor.. we broke local dev mode at some point. Can you help fix ` proxy=http://localhost:9411 ./npm.sh run dev` or similar?

stepanv:
@adriancole, I pushed another commit with extended README.md and fixed _devServer_.
Just note that I had to add `/zipkin/` context root to the `proxy` env var to make it work (I hope it's ok):
```
proxy=http://localhost:9411/zipkin/ ./npm.sh run dev
```

codefromthecrypt:
very nice work.

PR #1734, opened 2017-09-13
Ensures each span has a service name unless there was none

codefromthecrypt:
Before, we only displayed "prioritized names", like the service name
of the "lc" binary annotation. This allows non-standard spans to also
display the service name.


codefromthecrypt:
@jorgheymans brought this up

before

screen shot 2017-09-13 at 7 18 04 pm


after

screen shot 2017-09-13 at 7 12 31 pm

PR #1742, opened 2017-09-20
RabbitMQ collector

### LINKED ISSUE #1614: RabbitMQ collector support
As [mentioned by Adrian on Gitter](https://gitter.im/openzipkin/zipkin?at=5934c1c1cf9c13503c53ccc6), support for RabbitMQ has been requested multiple times, so I am making this issue to track it. Those who would like to see this implemented can vote for it.
(As it were, there seemed to be some interest in it at the Zipkin Meetup tonight)

> @byeswanth @shakuzen rabbitmq has been incidentally requested many times, and custom implemented in many zipkin systems. However, it has never been requested as a standard transport. You could be the first to actually make an issue requesting it 

Deliverables:

- [x] Defining the transport data format (Span2?)
- [x] Implementing a collector that accepts that format #1742

---
Out of scope for this issue (should be implemented in other repositories), but can be listed here for convenience when searching this topic.

- Implementing reporters that speak defined format
  - openzipkin/zipkin-reporter-java#58

Feel free to comment/edit this if anything is missing.

shakuzen:
This adds a RabbitMQ collector module along with its corresponding auto-configuration.

The implementation is largely inspired by the existing collector implementations. In fact, I think some of the logic and tests could be shared between them, but I did not try to tackle that in this PR.

Resolves #1614


zipkin-autoconfigure/collector-rabbitmq/src/test/java/zipkin/collector/rabbitmq/ZipkinRabbitMqCollectorAutoConfigurationTest.java
```diff
@@ -0,0 +1,99 @@
...
+    context.register(PropertyPlaceholderAutoConfiguration.class,
+        ZipkinRabbitMqCollectorAutoConfiguration.class, InMemoryConfiguration.class);
+    context.refresh();
+
+    thrown.expect(NoSuchBeanDefinitionException.class);
+    context.getBean(RabbitMqCollector.class);
+  }
+
+  @Test
+  @Ignore
```
shakuzen: This is ignored because it can't actually connect to a RabbitMQ server at localhost and the application context fails to refresh. This test was taken from the Kafka collector tests, but Kafka doesn't fail to start if it cannot connect to Kafka, I believe. It would still be nice to test that the conditional and rest of the auto-configuration is working. Let me know what you want to do about this.

zipkin-autoconfigure/collector-rabbitmq/src/test/java/zipkin/collector/rabbitmq/ZipkinRabbitMqCollectorPropertiesOverrideTest.java
```diff
@@ -0,0 +1,119 @@
...
+  @Parameterized.Parameter(1) public Object value;
+  @Parameterized.Parameter(2) public Function<ZipkinRabbitMqCollectorProperties, Object>
+      propertiesExtractor;
+  @Parameterized.Parameter(3) public Function<RabbitMqCollector.Builder, Object> builderExtractor;
+
+  @Parameterized.Parameters(name = "{0}")
+  public static Iterable<Object[]> data() {
+    return Arrays.asList(
+        parameters("addresses", "localhost:5671, localhost:5673",
+          properties -> properties.getAddresses().toString().substring(1, properties.getAddresses().toString().length() - 1),
```
shakuzen: This is a bit wonky due to the mismatch of types: the properties uses a `List<String>`, the builder `Address[]`, and of course we are passing just a String via the environment/YAML but Spring converts to a list for us. We could add a converter for `Address` but this felt like more work than it is worth right now.

zipkin-collector/rabbitmq/src/test/java/zipkin/collector/rabbitmq/RabbitMqCollectorTest.java
```diff
@@ -0,0 +1,301 @@
...
+import zipkin2.codec.SpanBytesEncoder;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static zipkin.TestObjects.LOTS_OF_SPANS;
+import static zipkin.TestObjects.TRACE;
+
+public class RabbitMqCollectorTest {
+
+  private static final int RABBIT_PORT = 5672;
+  private static final String RABBIT_DOCKER_IMAGE = "rabbitmq:3.6-alpine";
```
shakuzen: This tests against the latest patch version of RabbitMQ 3.6. We could probably fairly easily parameterize this if we wanted to test against multiple versions. A 3.7 release candidate was published recently, for instance.

shakuzen: I chose the `alpine` varient just because it is a smaller image

zipkin-collector/rabbitmq/src/main/java/zipkin/collector/rabbitmq/RabbitMqWorker.java
```diff
@@ -0,0 +1,47 @@
...
+
+  @Override
+  public void run() {
+    try {
+      Channel channel = this.connection.createChannel();
+      channel.queueDeclare(this.builder.queue, true, false, false, null);
+      final RabbitMqSpanConsumer consumer = new RabbitMqSpanConsumer(this.builder, channel);
+      channel.basicConsume(this.builder.queue, true, this.name, consumer);
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to start RabbitMQ consumer " + this.name, e);
```
codefromthecrypt: IllegalStateException?

zipkin-collector/rabbitmq/src/test/java/zipkin/collector/rabbitmq/RabbitMqCollectorTest.java
codefromthecrypt: sure.. we might make a test image at some point to reduce layers needed to be downloaded

zipkin-server/README.md
```diff
@@ -293,6 +293,25 @@ prefixed system property:
 $ KAFKA_ZOOKEEPER=127.0.0.1:2181 java -Dzipkin.collector.kafka.overrides.auto.offset.reset=largest -jar zipkin.jar
 ```
 
+### RabbitMQ collector
+The [RabbitMQ collector](../zipkin-collector/rabbitmq) will be enabled when the RabbitMQ server(s) `addresses` are set.
+
```
codefromthecrypt: I might suggest following conventions in the file and/or restructuring them generally

zipkin-collector/rabbitmq/src/main/java/zipkin/collector/rabbitmq/RabbitMqCollector.java
```diff
@@ -0,0 +1,197 @@
...
+      return this;
+    }
+
+    @Override public Builder metrics(CollectorMetrics metrics) {
+      this.metrics = checkNotNull(metrics, "metrics").forTransport("rabbitmq");
+      this.delegate.metrics(this.metrics);
+      return this;
+    }
+
+    public Builder addresses(Address[] addresses) {
```
codefromthecrypt: I'd accept a list, so that we can later use things like AutoValue (which don't allow arrays), and just convert to array at the site you need.

codefromthecrypt (approved):
Looks good. nothing major pops out at me. great work!

zipkin-server/README.md
shakuzen: I'm not sure I understand. I copied this convention from the corresponding table under "HTTP Collector". There seems to be a lack of consistency throughout the README. The "Self-Tracing" section reverses two of the columns, for the order of: "Variable", "Property", "Description". All the Storage sections and the "Scribe Collector" only give the environment variable and a description. Is this perhaps the convention you meant?

codefromthecrypt:
I can see the inconsistency. The reason why http collector stuff is only
listed in the zipkin-server readme is because the code only exists here
(there is no separate module even if maybe we should have one). You'll
notice a difference in how kafka etc are noted.. listing the env variables
here, but linking to a table in the kafka module's readme.

Make sense?

zipkin-collector/rabbitmq/README.md
```diff
@@ -0,0 +1,57 @@
...
+one-off spans to RabbitMQ to be collected by this collector.
+
+1. Start RabbitMQ server
+2. Start Zipkin server
+```bash
+$ RABBIT_ADDRESSES=localhost java -jar zipkin.jar
+```
+3. Save an array of spans to a file like `sample-spans.json`
+```json
+[{"traceId":"73321ecb02595c39","id":"73321ecb02595c39","name":"get","annotations":[{"timestamp":1505990621526000,"value":"sr","endpoint":{"serviceName":"service","ipv4":"127.0.0.1","port":8080}},{"timestamp":1505990621527000,"value":"ss","endpoint":{"serviceName":"service","ipv4":"127.0.0.1","port":8080}}],"binaryAnnotations":[{"key":"http.method","value":"GET","endpoint":{"serviceName":"service","ipv4":"127.0.0.1","port":8080}}]}]
```
codefromthecrypt: how about a v2 one!

```
[{
  "id": "9032b04972e475c5",
  "kind": "SERVER",
  "name": "get",
  "timestamp": 1505990621526000,
  "duration": 612898,
  "localEndpoint": {
    "ipv4": "192.168.1.113"
  },
  "remoteEndpoint": {
    "serviceName": "",
    "ipv4": "127.0.0.1",
    "port": 60149
  },
  "tags": {
    "http.path": "/a"
  }
}]
```

zipkin-collector/rabbitmq/README.md
```diff
@@ -0,0 +1,57 @@
...
+```
+4. Publish them using the CLI 
+```bash
+$ rabbitmqadmin publish exchange=amq.default routing_key=zipkin.spans < sample-spans.json
```
codefromthecrypt: I think the topic-like things have always defaulted to "zipkin" If that becomes ambiguous, we can append for whatever else we do. I'd prefer consistency and something obvious to guess.

(you might suggest this consistency is violated in ES, but that's a bit different as we have no dependency link transport, even if we have dependency link storage. if we did, we'd have to append to all the others, too.

zipkin-collector/rabbitmq/src/main/java/zipkin/collector/rabbitmq/RabbitMqCollector.java
```diff
@@ -0,0 +1,213 @@
...
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  /** Configuration including defaults needed to consume spans from a RabbitMQ queue. */
+  public static final class Builder implements CollectorComponent.Builder {
+    Collector.Builder delegate = Collector.builder(RabbitMqCollector.class);
+    CollectorMetrics metrics = CollectorMetrics.NOOP_METRICS;
+    String queue = "zipkin.spans";
```
codefromthecrypt: I'd prefer plain old zipkin here, as otherwise folks might accidentally suggest it in support and have that be a lie!

codefromthecrypt:
tested locally. good!

zipkin-collector/rabbitmq/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+  </properties>
+
+	<dependencies>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>zipkin</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.slf4j</groupId>
```
codefromthecrypt, line 44: is this an indirect dep of rabbit? if not, let's not use it (as it is unnecessary). use JUL instead.

zipkin-collector/rabbitmq/src/main/java/zipkin/collector/rabbitmq/RabbitMqCollector.java
```diff
@@ -0,0 +1,213 @@
...
+  public void close() throws IOException {
+    this.rabbitWorkers.close();
+  }
+
+  static final class LazyRabbitWorkers extends LazyCloseable<ExecutorService> {
+
+    final Builder builder;
+    final int concurrency;
+    final AtomicReference<CheckResult> failure = new AtomicReference<>();
+    private Connection connection;
```
codefromthecrypt: volatile and start and close are different threads

zipkin-collector/rabbitmq/src/main/java/zipkin/collector/rabbitmq/RabbitMqCollector.java
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  @Override
+  public void close() throws IOException {
+    this.rabbitWorkers.close();
+  }
+
+  static final class LazyRabbitWorkers extends LazyCloseable<ExecutorService> {
+
+    final Builder builder;
```
codefromthecrypt: careful about saving builder here as it is mutable and start is different thread than .build(). one way is to make your builder auto-value and make a copy of the builder (or just save off the params). otherwise add a TODO mentioning the edge-case mutability concern

zipkin-collector/rabbitmq/src/main/java/zipkin/collector/rabbitmq/RabbitMqSpanConsumer.java
```diff
@@ -0,0 +1,67 @@
...
+    byte[] body) throws IOException {
+    try {
+      this.metrics.incrementMessages();
+      // If we received legacy single-span encoding, decode it into a singleton list
+      if (body[0] <= 16 && body[0] != 12 /* thrift, but not a list */) {
+        this.metrics.incrementBytes(body.length);
+        Span span = SpanDecoder.THRIFT_DECODER.readSpan(body);
+        this.collector.accept(Collections.singletonList(span), NOOP);
+      } else {
+        this.collector.acceptSpans(body, DETECTING_DECODER, NOOP);
```
codefromthecrypt: kill all this for collector.acceptSpans as single-span encoding was a legacy of kafka specifically (we don't need to carry that here)

zipkin-collector/rabbitmq/src/test/java/zipkin/collector/rabbitmq/RabbitMqCollectorTest.java
```diff
@@ -0,0 +1,303 @@
...
+import zipkin.storage.SpanStore;
+import zipkin.storage.StorageComponent;
+import zipkin2.codec.SpanBytesEncoder;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static zipkin.TestObjects.LOTS_OF_SPANS;
+import static zipkin.TestObjects.TRACE;
+import static zipkin.collector.rabbitmq.RabbitMqCollector.convertAddresses;
+
+public class RabbitMqCollectorTest {
```
codefromthecrypt: there is insulation and also the ability to test locally in helpers like LazyCassandra3Storage might want to do that here

codefromthecrypt (commented):
spent an hour or so last night and found some things..

zipkin-collector/rabbitmq/pom.xml
shakuzen: `slf4j-api` is the RabbitMQ client library's [only required dependency](https://github.com/rabbitmq/rabbitmq-java-client/blob/master/pom.xml#L625-L629) but we don't need to use it ourselves. I'm not sure why I did... probably just habit.

zipkin-collector/rabbitmq/src/main/java/zipkin/collector/rabbitmq/RabbitMqSpanConsumer.java
shakuzen: Done in latest force pushed commit.

zipkin-collector/rabbitmq/src/main/java/zipkin/collector/rabbitmq/RabbitMQCollector.java
```diff
@@ -0,0 +1,202 @@
...
+      return this;
+    }
+
+    public Builder connectionFactory(ConnectionFactory connectionFactory) {
+      this.connectionFactory = checkNotNull(connectionFactory, "connectionFactory");
+      return this;
+    }
+
+    /**
+     * Queue zipkin spans will be consumed from. Defaults to "zipkin-spans".
```
codefromthecrypt, line 86: nit doc drift here perhaps elsewhere (default)

zipkin-collector/rabbitmq/src/main/java/zipkin/collector/rabbitmq/RabbitMQCollector.java
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  @Override public void close() throws IOException {
+    connection.close();
+  }
+
+  /** Lazy creates a connection and a queue before starting consumers */
+  static final class LazyInit extends LazyCloseable<Connection> {
+    final Builder builder;
+    final AtomicReference<CheckResult> failure = new AtomicReference<>();
```
codefromthecrypt, line 130: oops this isn't used anymore (or needed)

zipkin-collector/rabbitmq/src/main/java/zipkin/collector/rabbitmq/RabbitMQCollector.java
```diff
@@ -0,0 +1,202 @@
...
+  @Override
+  public RabbitMQCollector start() {
+    connection.get();
+    return this;
+  }
+
+  @Override
+  public CheckResult check() {
+    try {
+      CheckResult failure = connection.failure.get();
```
codefromthecrypt, line 115: this should check the connection, if there's a way to

zipkin-collector/rabbitmq/src/test/java/zipkin/collector/rabbitmq/RabbitMQCollectorTest.java
```diff
@@ -0,0 +1,136 @@
...
+import zipkin.Span;
+import zipkin.internal.ApplyTimestampAndDuration;
+import zipkin.internal.V2SpanConverter;
+import zipkin2.codec.SpanBytesEncoder;
+
+import static org.assertj.core.api.Java6Assertions.assertThat;
+import static zipkin.TestObjects.LOTS_OF_SPANS;
+import static zipkin.collector.rabbitmq.RabbitMQCollector.builder;
+
+public class RabbitMQCollectorTest {
```
codefromthecrypt: nit should be renamed to ITRabbitMQCollector (as it uses docker)

zipkin-collector/rabbitmq/src/test/java/zipkin/collector/rabbitmq/RabbitMQCollectorTest.java
```diff
@@ -0,0 +1,136 @@
...
+    collector.storage.clear();
+  }
+
+  @Rule public ExpectedException thrown = ExpectedException.none();
+
+  @Test public void checkPasses() throws Exception {
+    assertThat(collector.get().check().ok).isTrue();
+  }
+
+  @Test public void startFailsWithInvalidRabbitMqServer() throws Exception {
```
codefromthecrypt: this could possibly be moved to a normal unit test or left.. no matter

codefromthecrypt:
@bsideup we are stumped as to why this fails in travis.. do you have some time to help?

bsideup:
Hi Adrian! I'm happy to help, but can take a look only tomorrow, I hope
it's fine :)

codefromthecrypt:
sounds good to me! thanks, sergei

zipkin-collector/rabbitmq/src/main/java/zipkin/collector/rabbitmq/RabbitMQCollector.java
```diff
@@ -0,0 +1,202 @@
...
+    LazyInit(Builder builder) {
+      this.builder = builder;
+    }
+
+    @Override
+    protected Connection compute() {
+      Connection connection;
+      try {
+        connection = builder.connectionFactory.newConnection(builder.addresses);
+        connection.createChannel().queueDeclare(builder.queue, true, false, false, null);
```
shakuzen, line 141: The `Channel` created here will remain open until the connection is closed, which is not a big deal but it isn't necessary and it may be asked about by operators looking at open channels on RabbitMQ.

zipkin-collector/rabbitmq/src/main/java/zipkin/collector/rabbitmq/RabbitMQCollector.java
```diff
@@ -0,0 +1,202 @@
...
+      } catch (IOException | TimeoutException e) {
+        throw new IllegalStateException("Unable to establish connection to RabbitMQ server", e);
+      }
+      Collector collector = builder.delegate.build();
+      CollectorMetrics metrics = builder.metrics;
+
+      for (int i = 0; i < builder.concurrency; i++) {
+        String name = RabbitMQSpanConsumer.class.getName() + i;
+        try {
+          // TODO: Is channel 1-1 with consumer? if not re-use the same channel for each consumer
```
shakuzen: Generally speaking, yes. Channels should not be shared among threads. From the [Channel JavaDoc](https://www.rabbitmq.com/releases/rabbitmq-java-client/current-javadoc/com/rabbitmq/client/Channel.html):

> Channel instances must not be shared between threads. Applications should prefer using a Channel per thread instead of sharing the same Channel across multiple threads. While some operations on channels are safe to invoke concurrently, some are not and will result in incorrect frame interleaving on the wire. Sharing channels between threads will also interfere with Publisher Confirms. As such, applications need to use a Channel per thread.

zipkin-collector/rabbitmq/src/test/java/zipkin/collector/rabbitmq/LazyRabbitMQCollector.java
```diff
@@ -0,0 +1,122 @@
...
+
+  LazyRabbitMQCollector(String image) {
+    this.image = image;
+  }
+
+  @Override protected RabbitMQCollector compute() {
+    try {
+      container = new GenericContainer(image)
+        .withExposedPorts(RABBIT_PORT)
+        .waitingFor(new HostPortWaitStrategy());
```
bsideup: no need for it, it's a default value

zipkin-collector/rabbitmq/src/test/java/zipkin/collector/rabbitmq/LazyRabbitMQCollector.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  @Override public void close() throws IOException {
+    try {
+      RabbitMQCollector collector = maybeNull();
+      if (collector != null) collector.close();
+    } finally {
+      if (container != null) {
+        System.out.println("Stopping docker image " + image);
+        container.stop();
```
bsideup: `compute` runs once, lazily, but you call `close` many times (after every JUnit statement). This is why the tests are failing

bsideup: @adriancole

codefromthecrypt: hmm why would this pass locally!

zipkin-collector/rabbitmq/src/test/java/zipkin/collector/rabbitmq/LazyRabbitMQCollector.java
codefromthecrypt: k

zipkin-collector/rabbitmq/src/main/java/zipkin/collector/rabbitmq/RabbitMQCollector.java
codefromthecrypt: thx for the ref

zipkin-collector/rabbitmq/src/main/java/zipkin/collector/rabbitmq/RabbitMQCollector.java
codefromthecrypt: I guess this is implicitly closed via the connection, ya?

zipkin-collector/rabbitmq/src/test/java/zipkin/collector/rabbitmq/LazyRabbitMQCollector.java
codefromthecrypt: since this is a class rule, it shouldn't run for each test, right?

The error looks like it is failing even before tests are run...

```
Running zipkin.collector.rabbitmq.ITRabbitMQCollector
        ℹ︎ Checking the system...
        ✔ Docker version is newer than 1.6.0
        ✔ Docker environment has more than 2GB free
        ✔ File should be mountable
        ✔ Exposed port is accessible
Starting docker image rabbitmq:3.6-alpine
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 5.925 s <<< FAILURE! - in zipkin.collector.rabbitmq.ITRabbitMQCollector
[ERROR] zipkin.collector.rabbitmq.ITRabbitMQCollector  Time elapsed: 5.925 s  <<< ERROR!
java.lang.IllegalStateException: Unable to establish connection to RabbitMQ server
Caused by: java.io.IOException
Caused by: com.rabbitmq.client.ShutdownSignalException: connection error
Caused by: java.io.EOFException
```

codefromthecrypt:
added a commit to undo the lazy stuff (still using a class rule)

bsideup:
@adriancole does it work locally if you stop your local RabbitMQ and run tests with Docker? I have a feeling that some tests use local RabbitMQ (it explains why it passes locally)

codefromthecrypt:
>
> @adriancole <https://github.com/adriancole> does it work locally if you
> stop your local RabbitMQ and run tests with Docker? I have a feeling that
> some tests use local RabbitMQ (it explains why it passes locally)
>
very good guess. sadly after killing my rabbit still passes. I added a log
statement to verify

zipkin-collector/rabbitmq/src/main/java/zipkin/collector/rabbitmq/RabbitMQCollector.java
shakuzen: Yes, closing the connection will close any Channels that it opened. Ref. https://www.rabbitmq.com/releases/rabbitmq-java-client/current-javadoc/com/rabbitmq/client/Connection.html#close--

zipkin-collector/rabbitmq/src/test/java/zipkin/collector/rabbitmq/LazyRabbitMQCollector.java
shakuzen: @bsideup 
For our other tests run with test containers, I see the following output before the `Checking the system...` sequence:

```
[INFO] Running zipkin.storage.mysql.ITMySQLStorage
07:07:47.617 [main] INFO  o.t.d.EnvironmentAndSystemPropertyClientProviderStrategy - Found docker client settings from environment
07:07:47.633 [main] INFO  o.t.d.DockerClientProviderStrategy - Found Docker environment with Environment variables, system properties and defaults. Resolved: 
    dockerHost=unix:///var/run/docker.sock
    apiVersion='{UNKNOWN_VERSION}'
    registryUrl='https://index.docker.io/v1/'
    registryUsername='travis'
    registryPassword='null'
    registryEmail='null'
    dockerConfig='DefaultDockerClientConfig[dockerHost=unix:///var/run/docker.sock,registryUsername=travis,registryPassword=<null>,registryEmail=<null>,registryUrl=https://index.docker.io/v1/,dockerConfig=/home/travis/.docker,sslConfig=<null>,apiVersion={UNKNOWN_VERSION}]'
07:07:47.638 [main] INFO  o.testcontainers.DockerClientFactory - Docker host IP address is localhost
07:07:47.780 [main] INFO  o.testcontainers.DockerClientFactory - Connected to docker: 
  Server Version: 17.03.1-ce
  API Version: 1.27
  Operating System: Ubuntu 14.04.5 LTS
  Total Memory: 7479 MB
        ℹ︎ Checking the system...
        ✔ Docker version is newer than 1.6.0
        ✔ Docker environment has more than 2GB free
        ✔ File should be mountable
        ✔ Exposed port is accessible
```

Note it is entirely missing for the RabbitMQ tests case (as you can see from the output Adrian posted above), which leads me to believe that a different Docker discovery strategy is being used, but looking at the code for the tests I can't figure out why this would be different.

codefromthecrypt:
pushed a commit to use travis rabbitmq service (to see if local passes)

codefromthecrypt:
nice work, @shakuzen!

PR #1745, opened 2017-09-24
Backfills tests of late or lost parent ID

codefromthecrypt:
This starts to investigate impact of a missing or late parentId in a shared span. This could happen if a client doesn't propagate the `X-B3-ParentId` by mistake or by some signal loss.

cc @jcarres-mdsol @bogdandrutu @basvanbeek

### DIFF
--- a/zipkin-lens/src/zipkin/dependency-linker.test.js
+++ b/zipkin-lens/src/zipkin/dependency-linker.test.js
@@ -15,18 +15,19 @@ import { DependencyLinker } from './dependency-linker';
 import { SpanNodeBuilder } from './span-node';
 
 const debug = false; // switch to enable console output during tests
+// in reverse order as reporting is more likely to occur this way
 const trace = [
   {
-    traceId: 'a', id: 'a', kind: 'SERVER', localEndpoint: { serviceName: 'web' },
-  },
-  {
-    traceId: 'a', parentId: 'a', id: 'b', kind: 'CLIENT', localEndpoint: { serviceName: 'web' }, remoteEndpoint: { serviceName: 'app' },
+    traceId: 'a', parentId: 'b', id: 'c', kind: 'CLIENT', localEndpoint: { serviceName: 'app' }, remoteEndpoint: { serviceName: 'db' }, tags: { error: true },
   },
   {
     traceId: 'a', parentId: 'a', id: 'b', kind: 'SERVER', localEndpoint: { serviceName: 'app' }, remoteEndpoint: { serviceName: 'web' }, shared: true,
   },
   {
-    traceId: 'a', parentId: 'b', id: 'c', kind: 'CLIENT', localEndpoint: { serviceName: 'app' }, remoteEndpoint: { serviceName: 'db' }, tags: { error: true },
+    traceId: 'a', parentId: 'a', id: 'b', kind: 'CLIENT', localEndpoint: { serviceName: 'web' }, remoteEndpoint: { serviceName: 'app' },
+  },
+  {
+    traceId: 'a', id: 'a', kind: 'SERVER', localEndpoint: { serviceName: 'web' },
   },
 ];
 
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
Here's an example of this.. in the census library, the SpanContext type does not include the parent-id, so it can't propagate it (not without changing to somehow capture it).

Today, in Brave, we could make a function like `supportsJoin = false` similar to `ClientServerSameSpan ` in zipkin-go, which we set in any environment which can contain a node which might not send the parent ID. This issue is more about figuring out the "what if" ex knowing what we can resolve server-side from what we can't.

For example, dependency graph will be the worst off in a case like this, but we could at least differentiate a root span from one missing parent conventionally. For example, if the span ID is the same as the trace ID, conventionally we know this one is more "root like" even if it isn't 100% accurate.

cc also @jcchavezs since you are now supporting a zipkin tracer :)

codefromthecrypt:
added a commit to make mysql backfill this (might be the only storage impacted)
dependency linker will need more smarts to detect this.. not yet sorted.

codefromthecrypt:
note to folks who might use this to say "use my tracing system instead!"

B3 is a format in wide use, and so we have some shared stake investigating how this works. Obviously zipkin could drop support for shared spans, but that would orphan the ecosystem. That's precisely why I take time on issues like this.

codefromthecrypt:
PS I just looked deeply into Amazon X-Ray. It is mostly compatible with zipkin from a trace ID storage POV.

The trace ID includes a version bit, 32bits of epoch seconds and 96 bits of random. This means it is 128bits. Most zipkin systems only sample on lower 64-bits anyway, so supporting X-Ray compatible trace IDs mean changing generation to be the same approach. Span IDs are 64bit, so exact match zipkin.

Like `trace-context` only the calling ID is propagated (they call it Parent, we call it X-B3-SpanId), so fixing this issue will also also align with X-Ray integration.

cc @devinsba

bogdandrutu:
@adriancole nice discovery. You still need to fix to not share the span between client and server I think, correct?

codefromthecrypt:
>
> @adriancole <https://github.com/adriancole> nice discovery. You still
> need to fix to not share the span between client and server I think,
> correct?
>
well "fix" is relative :) we have to be able to express this is some way
because B3 once upon a time implied sharing. Here's how we do this in brave
https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/Propagation.java#L38

sokac:
Hit issue with `zipkin-dependency` when server didn't receive parentId due bug in client propagation. Depending on span/node order, the end result of the dependency graph calculation may be incorrect. Zipkin UI does render graph correctly, though.

This is where it happens:
https://github.com/openzipkin/zipkin/blob/master/zipkin2/src/main/java/zipkin2/internal/Node.java#L160 -> sets first node as rootId

Attached trace render correctly in zipkin UI:
dependency_graph_error
visualization

zipkin-dependency output. Note that rootSpanId is incorrectly set (sets graph span as root; expected: website)
```
18/03/10 19:28:51 DEBUG DependencyLinker: linking trace 1fd6cc4d8ddf461f3c7f1e93e9a12a2d
18/03/10 19:28:51 DEBUG DependencyLinker: attributing span missing parent to root: traceId=1fd6cc4d8ddf461f3c7f1e93e9a12a2d, rootSpanId=bcfd66c462760348, spanId=91b2eb9bb450ca94
18/03/10 19:28:51 DEBUG DependencyLinker: attributing span missing parent to root: traceId=1fd6cc4d8ddf461f3c7f1e93e9a12a2d, rootSpanId=bcfd66c462760348, spanId=0e5223f847261a90
18/03/10 19:28:51 DEBUG DependencyLinker: attributing span missing parent to root: traceId=1fd6cc4d8ddf461f3c7f1e93e9a12a2d, rootSpanId=bcfd66c462760348, spanId=ec08eb6a206e4a6e
18/03/10 19:28:51 DEBUG DependencyLinker: traversing trace tree, breadth-first
18/03/10 19:28:51 DEBUG DependencyLinker: processing {"traceId":"1fd6cc4d8ddf461f3c7f1e93e9a12a2d","id":"bcfd66c462760348","kind":"SERVER","name":"thrift:graph","timestamp":1520701883768137,"duration":1784,"localEndpoint":{"serviceName":"graph","ipv4":"127.0.0.1","port":17289},"tags":{"span.kind":"server"},"shared":true}
18/03/10 19:28:51 DEBUG DependencyLinker: root's peer is unknown; skipping
18/03/10 19:28:51 DEBUG DependencyLinker: processing {"traceId":"1fd6cc4d8ddf461f3c7f1e93e9a12a2d","id":"0e5223f847261a90","kind":"SERVER","name":"thrift:graph","timestamp":1520701884190118,"duration":6857,"localEndpoint":{"serviceName":"graph","ipv4":"127.0.0.1","port":17289},"remoteEndpoint":{"serviceName":"website","ipv4":"127.0.0.1","port":80},"tags":{"span.kind":"server"},"shared":true}
18/03/10 19:28:51 DEBUG DependencyLinker: processing ancestor {"traceId":"1fd6cc4d8ddf461f3c7f1e93e9a12a2d","id":"bcfd66c462760348","kind":"SERVER","name":"thrift:graph","timestamp":1520701883768137,"duration":1784,"localEndpoint":{"serviceName":"graph","ipv4":"127.0.0.1","port":17289},"tags":{"span.kind":"server"},"shared":true}
18/03/10 19:28:51 DEBUG DependencyLinker: incrementing link website -> graph
18/03/10 19:28:51 DEBUG DependencyLinker: processing {"traceId":"1fd6cc4d8ddf461f3c7f1e93e9a12a2d","id":"ec08eb6a206e4a6e","kind":"SERVER","name":"website","timestamp":1520701883318534,"duration":988343,"localEndpoint":{"serviceName":"website","ipv4":"127.0.0.1","port":17644}}
18/03/10 19:28:51 DEBUG DependencyLinker: cannot determine parent, looking for first server ancestor
18/03/10 19:28:51 DEBUG DependencyLinker: processing ancestor {"traceId":"1fd6cc4d8ddf461f3c7f1e93e9a12a2d","id":"bcfd66c462760348","kind":"SERVER","name":"thrift:graph","timestamp":1520701883768137,"duration":1784,"localEndpoint":{"serviceName":"graph","ipv4":"127.0.0.1","port":17289},"tags":{"span.kind":"server"},"shared":true}
18/03/10 19:28:51 DEBUG DependencyLinker: incrementing link graph -> website
```


A few solutions I can think of:
* check if there's another node with the same spanId
* sort nodes by timestamp
* maybe utilize property `shared`

codefromthecrypt:
@sokac https://github.com/openzipkin/zipkin/pull/1945 on yours

codefromthecrypt:
self-note: merge or close

zipkin/src/test/java/zipkin/storage/SpanStoreTest.java
```diff
@@ -828,6 +828,35 @@ public void clientTimestampAndDurationWinInSharedSpan() {
     }
   }
 
+  /**
+   * This test shows that if a parent ID is stored late (ex because it wasn't propagated), the span
+   * can resolve once it is.
+   */
+  @Test
+  public void lateParentIdInSharedSpan() {
```
codefromthecrypt: we don't merge client+server in java anymore. logic moved here https://github.com/openzipkin/zipkin/blob/master/zipkin-lens/src/zipkin/span-cleaner.test.js#L389

PR #1747, opened 2017-09-27
Save trace to permanent store

naoman:
Ref: #1093 
Quick (and dirty) implementation of `save trace`. Sending out for review to get some feedback.

Following @adriancole suggestion of deploying Zipkin Server twice, one backed by regular datastore and one with `permanent` datastore. 

Adding `Save Trace` button on the UI. Clicking this butting will make an AJAX request to backend api `/api/v1/save/trace/<trace_id>`, and on receiving this request, zipkin server will pull spans for this trace, and POST them to permanent store using api `/api/v1/spans`

Address of permanent datastore server can be configured by property `zipkin.permanent.store`

### DIFF
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -52,10 +52,6 @@
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-web</artifactId>
     </dependency>
-    <dependency>
-      <groupId>org.springframework</groupId>
-      <artifactId>spring-web</artifactId>
-    </dependency>
 
     <!-- zipkin requires exporting /health endpoint -->
     <dependency>
@@ -162,6 +158,11 @@
       <artifactId>brave-okhttp</artifactId>
       <version>${brave.version}</version>
       <optional>true</optional>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-web</artifactId>
     </dependency>
 
     <dependency>
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -124,6 +124,13 @@
       <optional>true</optional>
     </dependency>
 
+    <!-- RabbitMQ Collector -->
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>zipkin-autoconfigure-collector-rabbitmq</artifactId>
+      <optional>true</optional>
+    </dependency>
+
     <!-- Scribe Collector -->
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -20,7 +20,7 @@
   <parent>
     <groupId>io.zipkin.java</groupId>
     <artifactId>parent</artifactId>
-    <version>2.0.2-SNAPSHOT</version>
+    <version>2.1.0</version>
   </parent>
 
   <artifactId>zipkin-server</artifactId>
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -20,7 +20,7 @@
   <parent>
     <groupId>io.zipkin.java</groupId>
     <artifactId>parent</artifactId>
-    <version>2.1.0</version>
+    <version>2.1.1-SNAPSHOT</version>
   </parent>
 
   <artifactId>zipkin-server</artifactId>
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -168,11 +168,6 @@
     </dependency>
 
     <dependency>
-      <groupId>org.springframework</groupId>
-      <artifactId>spring-web</artifactId>
-    </dependency>
-
-    <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-test</artifactId>
       <scope>test</scope>
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -52,6 +52,10 @@
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-web</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-web</artifactId>
+    </dependency>
 
     <!-- zipkin requires exporting /health endpoint -->
     <dependency>
--- a/zipkin-ui/js/page/trace.js
+++ b/zipkin-ui/js/page/trace.js
@@ -42,15 +42,33 @@ const TracePageComponent = component(function TracePage() {
                                             link: `/api/v1/trace/${this.attr.traceId}`});
       });
 
-      this.$node.find('#saveTraceLink').click(e => {
+      this.$node.find('#archiveTraceLink').click(e => {
         e.preventDefault();
         const traceId = this.attr.traceId;
+        const archiveEndpoint = this.attr.config('archiveEndpoint');
+        const archiveReadEndpoint = this.attr.config('archiveReadEndpoint');
 
-        $.ajax(`/api/v1/save/trace/${traceId}`).done(result => {
-          alert(`Trace saved : ${result}`);
-        }).fail(error => {
+        $.ajax(`/zipkin/api/v1/trace/${traceId}`, {
+          type: 'GET',
+          dataType: 'json'
+        }).done(trace => {
+          console.log(trace);
+          $.ajax(`${archiveEndpoint}`, {
+            type: 'POST',
+            dataType: 'json',
+            crossDomain: true,
+            data: JSON.stringify(trace),
+            contentType: 'application/json; charset=utf-8'
+          }).done(result => {
+            console.log(result);
+            alert(`Trace Archived : ${archiveReadEndpoint}/${traceId}`);
+          }).fail(error => {
+            console.log(error);
+            alert(`Unable to save trace ${this.attr.traceId}`);
+          });
+        }).fail(error2 => {
+          console.log(error2);
           alert(`Unable to save trace ${this.attr.traceId}`);
-          console.log(error);
         });
       });
 
--- a/zipkin-ui/js/page/trace.js
+++ b/zipkin-ui/js/page/trace.js
@@ -17,7 +17,9 @@ const TracePageComponent = component(function TracePage() {
 
     TraceData.attachTo(document, {
       traceId: this.attr.traceId,
-      logsUrl: this.attr.config('logsUrl')
+      logsUrl: this.attr.config('logsUrl'),
+      archiveEndpoint: this.attr.config('archiveEndpoint'),
+      archiveReadEndpoint: this.attr.config('archiveReadEndpoint')
     });
     this.on(document, 'tracePageModelView', function(ev, data) {
       this.$node.html(traceTemplate(data.modelview));
--- a/zipkin-ui/js/page/trace.js
+++ b/zipkin-ui/js/page/trace.js
@@ -61,7 +61,8 @@ const TracePageComponent = component(function TracePage() {
             contentType: 'application/json; charset=utf-8'
           }).done(result => {
             console.log(result);
-            alert(`Trace Archived : ${archiveReadEndpoint}/${traceId}`);
+            window.prompt('Trace archived. Copy link to clipboard: Cmd+C, Enter',
+              `${archiveReadEndpoint}/${traceId}`);
           }).fail(error => {
             console.log(error);
             alert(`Unable to save trace ${this.attr.traceId}`);
--- a/zipkin-ui/js/page/trace.js
+++ b/zipkin-ui/js/page/trace.js
@@ -48,7 +48,7 @@ const TracePageComponent = component(function TracePage() {
         const archiveEndpoint = this.attr.config('archiveEndpoint');
         const archiveReadEndpoint = this.attr.config('archiveReadEndpoint');
 
-        $.ajax(`/zipkin/api/v1/trace/${traceId}`, {
+        $.ajax(`/api/v2/trace/${traceId}`, {
           type: 'GET',
           dataType: 'json'
         }).done(trace => {
--- a/zipkin-ui/js/page/trace.js
+++ b/zipkin-ui/js/page/trace.js
@@ -74,11 +74,11 @@ const TracePageComponent = component(function TracePage() {
             }
           }).fail(error => {
             console.log(error);
-            alert(`Unable to save trace ${this.attr.traceId}`);
+            alert(`Unable to archive trace ${this.attr.traceId}`);
           });
         }).fail(error2 => {
           console.log(error2);
-          alert(`Unable to save trace ${this.attr.traceId}`);
+          alert(`Unable to archive trace ${this.attr.traceId}`);
         });
       });
 
--- a/zipkin-ui/js/page/trace.js
+++ b/zipkin-ui/js/page/trace.js
@@ -10,6 +10,7 @@ import TraceUI from '../component_ui/trace';
 import FilterLabelUI from '../component_ui/filterLabel';
 import ZoomOut from '../component_ui/zoomOutSpans';
 import {traceTemplate} from '../templates';
+import {contextRoot} from '../publicPath';
 
 const TracePageComponent = component(function TracePage() {
   this.after('initialize', function() {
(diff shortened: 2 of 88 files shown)


codefromthecrypt:
Add screen shot?

PS: use v2 api as this is a new feature? Easier than having to do a migration of this to v2 later

codefromthecrypt (commented):
interesting.. I was thinking this was going to happen in javascript. maybe it should? that way it doesn't need to expose a new server-side endpoint.

naoman:
screen shot 2017-09-28 at 2 57 55 pm
screen shot 2017-09-28 at 2 58 13 pm
screen shot 2017-09-28 at 3 01 58 pm

> use v2 api as this is a new feature? Easier than having to do a migration of this to v2 later

We have not migrated to V2 yet. Our datastore is in V1 format, so we have to read via V1 api. For writing to permanent store, either V1 or V2 is fine. Looks like zipkin server writes in V2 format even if we call V1 api, which seems a bit strange.
Ideally we should be able to run the server in V1 mode or V2 mode, and all APIs can then work accordingly. Do we have a config property that can do this?

>interesting.. I was thinking this was going to happen in javascript. maybe it should? that way it doesn't need to expose a new server-side endpoint.

I wanted to keep the client light. Also, if its on the server side, other clients/scripts can also use it. We can move it to javascript if needed.

codefromthecrypt:
I don't want to make an api like this as a part of the server api, especially not under v1 namespace. it doesn't impact the UI to do the blocking here or there.

In order to deploy this UI change, you will inherit the v2 api anyway, unless you are somehow hosting static assets independently.. is that the case?

naoman:
>I don't want to make an api like this as a part of the server api, especially not under v1 namespace. it doesn't impact the UI to do the blocking here or there.

So you want to go for javascript solution, where client (UI) reads form one zipkin-server and saves to `permanent` zipkin-server? I'll make the change.

 > In order to deploy this UI change, you will inherit the v2 api anyway, unless you are somehow hosting static assets independently.. is that the case?

Not sure what you mean. What I'm thinking is that UI/Javascript will read raw spans for a trace using `/api/v1/trace/<trace_id>` and post them to `http://<permanent_zipkin_address>/api/v1/spans`

zipkin-ui/js/page/trace.js
```diff
@@ -40,6 +40,18 @@ const TracePageComponent = component(function TracePage() {
                                             link: `/api/v1/trace/${this.attr.traceId}`});
       });
 
+      this.$node.find('#saveTraceLink').click(e => {
+        e.preventDefault();
+        const traceId = this.attr.traceId;
+
+        $.ajax(`/api/v1/save/trace/${traceId}`).done(result => {
```
codefromthecrypt: so basically I'd suggest having an alternative that reads from config.json and if a key like "permanentZipkinUrl" is present, show the link, and on clicking the link copy the result of /api/v2/trace/ID to a POST ${permanentZipkinEndpoint} (ex http://foo/bar/api/v2/spans) 

you don't need to process the body in any way or even know about the v2 model. The server hosting the permanent repo doesn't need to be using v2 native storage, though obviously in the case of elasticsearch it should as v2 is a better investment for a concept like permanent.

For the handling, just copy the bytes from the first response to the request body of the second. the code will be easy and the results easy to reason with. The url associated with this task could be defined in crossroads.

If this isn't clear etc, lemme know!

codefromthecrypt: one other consideration  is to add a tag to the first span (easy in v2 model), just insert one in the tags dict.
ex tags.outage='disaster happened in JIRA FOO-1234' then in your permanent store, you can search for "outage" and get this trace. Such a feature could be worth a little javascript to make it work.

This will allow you to label the trace such that in your long-term storage you can look it up somehow. Otherwise, eventually, these favorite traces will become annoying to browse.

codefromthecrypt:
> Not sure what you mean. What I'm thinking is that UI/Javascript will read raw spans for a trace using /api/v1/trace/<trace_id> and post them to http://<permanent_zipkin_address>/api/v1/spans

I don't want to introduce new features that depend on v1 api. That's
why I mentioned if your concern is about not having v2 api, yet, you
would as a side-effect of deploying this update! So basically we
shouldn't make a long term decision based on a short term upgrade
concern.

Ex save /api/v2/trace/<trace_id> to  POST
<permanent_zipkinv2_endpoint> not using /api/v2/spans explicitly in
the config entry as some will mount it via proxy maybe to the same
origin under a different path. Make sense?

codefromthecrypt:
Sorry to not think of this earlier, but would this experience be ok? Ex we
have a saved trace, but this UI wont be able to find saved traces.. Is that
ok? If so, why? If not, why not?

naoman:
>Ex we have a saved trace, but this UI wont be able to find saved traces.. Is that ok? If so, why? If not, why not?

Ideally same UI should provide access to saved traces, but I think its okay to have a separate UI, if it helps keeping the system simple and maintainable. It could be considered an `archiving` feature, and in most systems, retrieving archived data has a separate process than active data.

Also, if a user is trying to access a trace thats not available in the regular datastore, the UI can share a link to archive/permanent store, or even pull trace from there. Search will be a bit tricker, but we can handle that too by picking the correct datastore based on the TTL for the regular store.

codefromthecrypt:
Thanks for the info. I kindof like the archive term, as I don't want
us necessarily to try to make the UI a fan-out system at the moment.
Maybe we align the property name as such? like archiveEndpoint? Also,
folks deploying an archive instance of zipkin can set the property
`zipkin.ui.environment` to archive to know that's what they are
looking at.

I basically would like this feature to be easy to explain with some
coherent suggestions on setup... and so far I think archive is it. do
you agree?

naoman:
>... and so far I think archive is it. do you agree?

Yes.

screen shot 2017-10-03 at 4 10 25 pm

Error message
screen shot 2017-10-03 at 4 10 38 pm

Success message, when `zipkin.ui.archive-read-endpoint` is not specified.
screen shot 2017-10-03 at 4 11 20 pm

Success message, when `zipkin.ui.archive-read-endpoint` is specified.
screen shot 2017-10-03 at 4 12 04 pm

zipkin-server/pom.xml
```diff
@@ -161,6 +161,11 @@
     </dependency>
 
     <dependency>
+      <groupId>org.springframework</groupId>
```
codefromthecrypt: not needed right?

codefromthecrypt (commented):
I like it. great docs, too!

Can you poke around the zipkin-js tests section to see if there's any low-hanging fruit (ex test case that already exists that you can pop a related unit test into)?

codefromthecrypt:
you'll want to rebase this on master, as the thing that allows proxies to work was merged

zipkin-server/pom.xml
naoman: I was getting this exception while posting spans to `/api/v2/spans`
```
2017-10-03 11:08:28.605 ERROR 46979 --- [nio-9412-exec-2] o.a.c.c.C.[.[.[/].[dispatcherServlet]    : Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Handler dispatch failed; nested exception is java.lang.NoClassDefFoundError: org/springframework/util/concurrent/SettableListenableFuture] with root cause

java.lang.NoClassDefFoundError: org/springframework/util/concurrent/SettableListenableFuture
	at zipkin.server.ZipkinHttpCollector.validateAndStoreSpans(ZipkinHttpCollector.java:88) ~[classes!/:na]
	at zipkin.server.ZipkinHttpCollector.uploadSpansJson2(ZipkinHttpCollector.java:67) ~[classes!/:na]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_101]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_101]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_101]
```
This [stack overflow](https://stackoverflow.com/questions/41727213/nested-exception-is-java-lang-noclassdeffounderror-org-springframework-web-cont) suggested this solution. I couldn't get it working without adding this dependency.

naoman:
hum.. I messed up rebase

codefromthecrypt:
maybe add




as an xml comment above the dependency? That way, we can figure out
later what's up

codefromthecrypt:
I can fix the rebase stuff on the way in, no worries

naoman:
Added unit test and removed unnecessary dependency. The exception I was seeing earlier must be because of some setup issue.

codefromthecrypt:
looks like the test is failing:

```
[INFO] JS 2.1.1 (Linux 0.0.0) traceToMustache should show archiveReadEndpoint FAILED
[INFO] 	undefined is not an object (evaluating 'modelview.archiveEndpoint.should')
[INFO] 	/home/travis/build/openzipkin/zipkin/zipkin-ui/test/component_ui/traceToMustache.test.js:110:31 <- webpack:///test/component_ui/traceToMustache.test.js:70:30
[INFO] PhantomJS 2.1.1 (Linux 0.0.0): Executed 82 of 93 (1 FAILED) (0 secs / 0.041 secs)
```

naoman:
@adriancole can you help with this `ci/circleci` failure? Looks like its a setup issue. 

screen shot 2017-10-04 at 10 26 02 am

codefromthecrypt:
@naoman circle has been hell lately. just ignore it for now as I can't dedicate time towards it. I do spend probably 10hrs a month on various build issues... just need to get some work done.

codefromthecrypt:
@naoman can you confirm that the archive button is absent when there's no url present?

jcarres-mdsol:
Seems a hacky solution, most people will not have this setup. Would the button appear in the UI by default? It may set expectations it would work when in most cases would not

naoman:
@adriancole Yes, `Archive` button shows only when `zipkin.ui.archive-endpoint` is present. Similar to `Logs` button.

codefromthecrypt:
@jcarres-mdsol so we have had a number of requests for how to achieve permanent storage w/o complicating other things. I've not found a simpler way than this..

The notes would be that if you want permanent storage, setup zipkin somewhere (or a zipkin proxy like stackdriver), then point env variables to that. If using a normal zipkin server, the archival one should probably set the environment name to "archive" (something you can configure on the UI).

Does this clarify the setup? indeed it would be new for a lot of folks, but fairly easy to provision (add a permanent namespace to storage, keep a server running which needn't deal with high volume as it is only archival)

jorgheymans:
will the UI have an href somewhere then to the archive endpoint ?

naoman:
@jorgheymans when a trace is archived, the UI will show a link to archived trace in the confirmation dialog. Screenshot attached above.

>@naoman just curious how many traces do you save to permanent store ? Is it a regularly occuring thing ?

Not many, maybe just a few traces per day. Because of our high trace volume, regular datastore has TTL of 10 days. This feature will let users save "interesting" traces, specially something related to a bug, for longer duration.

jorgheymans:
@naoman what i meant was just a top level link on the ui somewhere that points to the archiving endpoint if it is configured. That way ppl only have to remember (or bookmark) the "main" zipkin instance.

naoman:
@jorgheymans Yes, this can be useful. But at this point I want to keep the code simple and clean, since this feature will not be used by many users. In my personal experience, when archiving a trace, users will save the link to the trace itself, since they are archiving the trace to save or share it.

I would say we should go ahead with the current code, and revisit based on user feedback.

codefromthecrypt:
I'm not sure how much impact there is to code hygiene to present a link to
the archive repo. It does feel like it is more a feature if one has the
ability to know where the archive repo is without saving a trace to figure
it out.

There's currently code that conditionally shows the environment label when
present in configuration. How bad would the code look if a similar approach
was used to place a link to the archives where people can see it?

naoman:
Like I mentioned in my last message, this change is ready and fully functional. Adding a link would be an improvement on top of this change that can be made in a separate pull request. Lets merge this change, get the feature out there so that people can start using it, get user feedback, and then improve iteratively.

codefromthecrypt:
Naoman I dont think you understand. This change is for the community and a
valid request was made. We dont merge single user stuff, so consider it
lucky someone else is interested!

naoman:
>We don't merge single user stuff, so consider it lucky someone else is interested!

Not sure what you mean. Do you think I'm trying to push something thats Pinterest specific, and not needed by other users? 

Let me try to explain one more time. While the request to add an additional url link is valid, its an add-on feature for this change, and does not block it. 

Can you please explain why these two changes can not be made in two separate pull requests?

jorgheymans:
For sure this can be done in a separate PR, after all it was just a
suggestion. We don't really have the need to save traces right now but once
the feature is in I'll definitely configure an archive endpoint because it
does pop up every now and then. I'll send a PR myself if I still see the
need for it then.

Op za 7 okt. 2017 08:06 schreef Naoman Abbas <notifications@github.com>:

codefromthecrypt:
>
> Let me try to explain one more time. While the request to add an
> additional url link is valid, its an add-on feature for this change, and
> does not block it.

This is a community project. One of the user/core team members felt this
was hacky. Another user asked for an adjustment to the change. No one so
far has actually said they want the change actively. We have to be careful
to make sure code is actually useful as often non-core people disappear
after they land something.

Historically, UI code is write once, very few go back and maintain changes
later. I don't have enough experience with you personally to know if you
would actually go back later and do a pull request.

The time you have taken to intentionally not answer the question has rid me
of effort I can spend elsewhere and also yourself as looking at the
suggestion wouldn't have been costly. Please be more considerate especially
if you want others to help merge an unpopular change.

codefromthecrypt:
@openzipkin/core I personally don't want to go around in explanation circles on this. I actually am ok with the change in general, but not motivated to continue this line of inquiry. If someone else is ok driving this, please do. I'm going to unsubscribe for the time being

naoman:
@jorgheymans I've added a link to archive server on the menu bar. Let me know what you think. Below is the screenshot.

screen shot 2017-10-07 at 1 03 43 pm

>Seems a hacky solution, most people will not have this setup. 

@jcarres-mdsol can you please elaborate on why you feel this solution is hacky and how it can be improved. 

>Would the button appear in the UI by default? It may set expectations it would work when in most cases would not

@jcarres-mdsol No it would not. I've added this information in Readme as well. Let me know if you think it needs more explanation there.

jorgheymans:
Spot on that's exactly where I had imagined the link to appear :-)

Op za 7 okt. 2017 22:15 schreef Naoman Abbas <notifications@github.com>:

eirslett:
Zipkin already has a feature to download a trace as a JSON file (and I think also one to upload a JSON file and render it in the UI). Wouldn't that cover the use case? You could just store the JSON file wherever you want, on a hard drive, dropbox etc...

I do agree with @adriancole's comments in general.

eirslett:
Another solution is to run a cache proxy like Varnish in front of Zipkin, and make it cache results for 30 days. (Assuming that the amount of Zipkin data developers look at will be very small relative to the total amount of Zipkin data, so it could comfortably fit on a cache instance)

naoman:
@eirslett Yes, thats do able, however it wont be very scaleable and convenient. Consider the use case where we want to save a trace related to a bug. Without this feature, we'll first have to download the trace, attach it to the bug ticket, and upload if back to zipkin server for looking at the trace. This feature provides a convenient way to archive traces with a single click.

eirslett:
> Without this feature, we'll first have to download the trace, attach it to the bug ticket, and upload if back to zipkin server for looking at the trace. This feature provides a convenient way to archive traces with a single click.

With the Varnish solution, I guess you could just copy/paste a link to the trace in the bug report?

naoman:
@eirslett varnish proxy based cache will have its own limitations. Below are few:
- what if we want to keep data for longer than cache TTL? 
- how do we enable data replication?  
- What if other tools (service dependency graph, trace analyzer scripts, etc) want to pull this data?

jcarres-mdsol:
@naoman  I think it is hacky because to add this new functionality, I need to double my infrastructure.

naoman:
> I think it is hacky because to add this new functionality, I need to double my infrastructure.

@jcarres-mdsol Can you please elaborate on this? This solution does require some additional infrastructure setup, but by no means it would double the infrastructure. For example, below is how Elasticsearch based solution look like:

- A new index for archived traces in the same Elasticsearch cluster. 
- A new Zipkin Server instance for archived traces. This instance can run on the same hosts running regular Zipkin Server
- A load-balancer in front of Zipkin archive server. This is the only additional infrastructure component needed for this solution.

If you still feel that this is too much hassle for archiving traces, can you propose a better solution?

flipdiving:
Thanks for your sharing. Hope you can contribute more quality posts to this page. Thank you!
[laravel quiz](http://quizbucket.org/quiz/nodejs)

PR #1752, opened 2017-10-01
Keep a separate stress schema cql file, as stress can't use use UDTs

michaelsembwever:
For safety, create the stress schema in a different keyspace.


codefromthecrypt:
circleci is busted for a different reason (it is very bustable lately)

codefromthecrypt (approved):
maybe make a note in the src/main file to update this file when updating

PR #1753, opened 2017-10-01
In Cassandra3 storage, only use consistency level LOCAL_ONE if a local_dc has been defined, otherwise use ONE

michaelsembwever:
Also…

 - Mark all writes as idempotent.
 - Update the retry policy, honouring idempotence, consistency level, and applying appropriate back-pressure.

### DIFF
--- a/zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/DefaultSessionFactory.java
+++ b/zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/DefaultSessionFactory.java
@@ -127,10 +127,10 @@ final class DefaultSessionFactory implements Cassandra3Storage.SessionFactory {
     ));
 
     builder.withQueryOptions(
-                new QueryOptions()
-                        .setConsistencyLevel(
-                                null != cassandra.localDc ? ConsistencyLevel.LOCAL_ONE : ConsistencyLevel.ONE)
-                        .setDefaultIdempotence(true));
+      new QueryOptions()
+        .setConsistencyLevel(
+          null != cassandra.localDc ? ConsistencyLevel.LOCAL_ONE : ConsistencyLevel.ONE)
+        .setDefaultIdempotence(true));
 
     if (cassandra.useSsl) {
       builder = builder.withSSL();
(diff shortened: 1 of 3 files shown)


zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/DefaultSessionFactory.java
```diff
@@ -123,6 +125,13 @@ static Cluster buildCluster(Cassandra3Storage cassandra) {
     builder.withPoolingOptions(new PoolingOptions().setMaxConnectionsPerHost(
         HostDistance.LOCAL, cassandra.maxConnections
     ));
+
+    builder.withQueryOptions(
```
codefromthecrypt, line 129: this is the only one I don't know just by reading it. Can you put a comment line of why we do this, or tell me and I'll add it? Thx!

codefromthecrypt (approved):
reformatted w/ https://github.com/square/java-code-styles vs nagging you :)

just one thing on my end, just to understand this. @llinder how do you feel?

llinder (approved):

zipkin-storage/cassandra3/src/main/java/zipkin/storage/cassandra3/DefaultSessionFactory.java
michaelsembwever: if `local_dc` isn't defined then LOCAL_ONE sticks requests to the dc of the first seed host that the driver connects to.
ONE will at least in this situation connect to any dc, and then any latency awareness will bring it back to (eventually) using the local dc.

furthermore `LOCAL_ONE` can fail easily if you're not using 'NetworkTopologyStrategy` but that's a fix for another day.

michaelsembwever: added comment. (can you please squash the 4 commits!)

codefromthecrypt: great info. ty

codefromthecrypt:
thanks tons!

PR #1757, opened 2017-10-05
Add a profile to the root pom.xml that contains the basic code style rules that NetBeans will honour

michaelsembwever:


codefromthecrypt:
thanks, mick!

Issue #1695, opened 2017-08-18
Consider switching cassandra3 to span2 model


codefromthecrypt:
The cassandra3 type isn't finalized, so we have an opportunity to drop support for complexity around binary annotations and nested endpoints. This might result in the ability of using SASI to index service names (as they are fields in the span now as opposed to nested parts of binary annotation or annotation).

I don't know how well this would go with indexing, but here are the salient changes.
```
CREATE TYPE IF NOT EXISTS zipkin3.annotation (
    timestamp bigint,
    value  text
);

CREATE TABLE IF NOT EXISTS zipkin3.traces (
    trace_id            frozen<trace_id>,
    ts_uuid             timeuuid,
    parent_id           bigint,
    id                  bigint,
    timestamp           bigint,
    name                text,
    duration            bigint,
    local_endpoint      frozen<Endpoint>,
    remote_endpoint     frozen<Endpoint>,
    annotations         list<frozen<annotation>>,
    tags                frozen<map<string,string>>,
    shared              boolean,
    annotation_query    text, //-- can't do SASI on set<text>: comma-joined until CASSANDRA-11182
    PRIMARY KEY (trace_id, ts_uuid, id)
)
```

cc @openzipkin/cassandra

codefromthecrypt:
note annotation_query would be shorter per span as it only needs to concat annotation and tag pairs (as opposed to also putting in the service name, because the local_endpoint covers this part)

michaelsembwever:
> note annotation_query would be shorter per span as it only needs to concat annotation and tag pairs (as opposed to also putting in the service name, because the local_endpoint covers this part)

i don't get this. when we search against that index we still are searching for `local_serviceName:annotation` or `local_serviceName;tag_key;tag_value`.

codefromthecrypt:
> i don't get this. when we search against that index we still are searching for local_serviceName:annotation or local_serviceName;tag_key;tag_value.

I mean to say that the serviceName can be refined independent of the
ball of tags. for example, span.local_serviceName = foo filtered
before annotationQuery. So instead of redundantly encoding the service
name in the annotationQuery concatenation, access it as a field.

My assumption is that this could be more efficient, but might be
wrong. At any rate, it would be easier to read.

Make sense?

michaelsembwever:
> Make sense?

Yup. I've put the extra column in and the extra SASI on it, so to achieve this.
[...]

PR #1758, opened 2017-10-05
Switching cassandra3 to span2 model #1695

michaelsembwever:
https://github.com/openzipkin/zipkin/issues/1695

(a lot here is from @adriancole !)


llinder:
Looks great so far!  Thanks a ton for working on this @michaelsembwever

zipkin-storage/cassandra3/src/main/resources/cassandra3-schema.cql
```diff
@@ -1,88 +1,79 @@
-CREATE KEYSPACE IF NOT EXISTS zipkin3 WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'};
+CREATE KEYSPACE IF NOT EXISTS zipkin2_cassandra3 WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'};
```
michaelsembwever: in this file i still have to add some comments, and i want to duplicate all comments into the schema's table-level ` WITH COMMENT = '…';` so that operators get that information too…

michaelsembwever: done.

michaelsembwever:
Some experimental benchmarking, aiming only to give rough latency spread using a safe throughput rate.

 - macbook pro (macOS 10.13 , 3.5GHz dual-core Intel Core i7 processor, 16GB 2133MHz, SSD)
 - CCM (4 nodes, Cassandra-3.11.0)
 - `ccm stress` run for each of the three profiles in write mode beforehand
 - stress testing ran for 6 hours (overnight), but throttled to very low throughput.

Results:

`ccm stress  user profile=trace-stress.yaml ops\(insert=1,by_trace=1,by_trace_ts_id=1,by_annotation=1\) no-warmup  duration=6h  -rate threads=4 throttle=50/s`

```
Results:
Op rate                   :       50 op/s  [by_annotation: 12 op/s, by_trace: 13 op/s, by_trace_ts_id: 13 op/s, insert: 13 op/s]
Partition rate            :       50 pk/s  [by_annotation: 12 pk/s, by_trace: 13 pk/s, by_trace_ts_id: 13 pk/s, insert: 13 pk/s]
Row rate                  :      174 row/s [by_annotation: 137 row/s, by_trace: 13 row/s, by_trace_ts_id: 13 row/s, insert: 13 row/s]
Latency mean              :   10.1 ms [by_annotation: 31.9 ms, by_trace: 1.8 ms, by_trace_ts_id: 1.8 ms, insert: 5.1 ms]
Latency median            :    2.5 ms [by_annotation: 23.7 ms, by_trace: 0.6 ms, by_trace_ts_id: 0.7 ms, insert: 3.1 ms]
Latency 95th percentile   :   43.6 ms [by_annotation: 77.7 ms, by_trace: 6.3 ms, by_trace_ts_id: 6.3 ms, insert: 15.4 ms]
Latency 99th percentile   :   84.7 ms [by_annotation: 125.1 ms, by_trace: 20.6 ms, by_trace_ts_id: 20.5 ms, insert: 37.5 ms]
Latency 99.9th percentile :  159.6 ms [by_annotation: 221.4 ms, by_trace: 65.2 ms, by_trace_ts_id: 65.3 ms, insert: 88.3 ms]
```

`ccm stress  user profile=trace_by_service_span-stress.yaml ops\(insert=1,select=1,by_duration=1\) no-warmup  duration=6h  -rate threads=4 throttle=50/s`

```
Results:
Op rate                   :       50 op/s  [by_duration: 17 op/s, insert: 17 op/s, select: 17 op/s]
Partition rate            :       33 pk/s  [by_duration: 0 pk/s, insert: 17 pk/s, select: 17 pk/s]
Row rate                  :       33 row/s [by_duration: 0 row/s, insert: 17 row/s, select: 17 row/s]
Latency mean              :    2.6 ms [by_duration: 4.0 ms, insert: 1.8 ms, select: 1.9 ms]
Latency median            :    0.8 ms [by_duration: 1.8 ms, insert: 0.6 ms, select: 0.6 ms]
Latency 95th percentile   :   11.1 ms [by_duration: 15.9 ms, insert: 7.5 ms, select: 8.1 ms]
Latency 99th percentile   :   27.9 ms [by_duration: 35.0 ms, insert: 22.6 ms, select: 22.8 ms]
Latency 99.9th percentile :   73.3 ms [by_duration: 82.4 ms, insert: 64.2 ms, select: 69.9 ms]
```

`ccm stress  user profile=span_by_service-stress.yaml ops\(insert=1,select=1,select_spans=1\) no-warmup  duration=6h  -rate threads=4 throttle=50/s`

```
Results:
Op rate                   :       50 op/s  [insert: 17 op/s, select: 17 op/s, select_spans: 17 op/s]
Partition rate            :       50 pk/s  [insert: 17 pk/s, select: 17 pk/s, select_spans: 17 pk/s]
Row rate                  :    1,707 row/s [insert: 17 row/s, select: 1,673 row/s, select_spans: 17 row/s]
Latency mean              :    7.5 ms [insert: 1.7 ms, select: 18.9 ms, select_spans: 1.8 ms]
Latency median            :    0.8 ms [insert: 0.5 ms, select: 11.2 ms, select_spans: 0.5 ms]
Latency 95th percentile   :   34.3 ms [insert: 7.4 ms, select: 57.1 ms, select_spans: 8.1 ms]
Latency 99th percentile   :   69.5 ms [insert: 21.5 ms, select: 93.3 ms, select_spans: 22.5 ms]
Latency 99.9th percentile :  136.4 ms [insert: 63.6 ms, select: 176.8 ms, select_spans: 68.5 ms]
```

As is to be expected the writes are reasonably on-par for the different tables (bc bound by the one piece of hardware).
Again as expected reads are fast when selecting traces by their IDs, when finding trace ids by the service+span, and when selecting span names by service name.
The three SASI indexes here all perform well, roughly 5x times the latency compared to the fast (select by partition key) reads.
Of the three SASI indexes reads, searching "by_annotation" in the traces table is the slowest, as expected as this actually uses two SASI indexes, one of which is the only SASI index that uses the `CONTAINS` mode permitting full text searching.

codefromthecrypt:
Thanks for the benching. Will take a review sweep today

michaelsembwever:
@adriancole ta.
Two commits added:
 - changing dependency table to store the individual fields to DependencyLink, rather than the whole list of links as a blob,
 - refactoring `zipkin.storage.cassandra3` to `zipkin2.storage.cassandra3`

codefromthecrypt:
- changing dependency table to store the individual fields to
   DependencyLink, rather than the whole list of links as a blob,
   - refactoring zipkin.storage.cassandra3 to zipkin2.storage.cassandra3

Ah ok will wait on that

michaelsembwever:
@adriancole oh, i edited my comment above. both items have been pushed already (and exist as separate commits).

codefromthecrypt:
Perfect

michaelsembwever:
LGTM. thanks for a lot of the clean up in the last commit: a few silly bugs and typos there you caught!

codefromthecrypt:
just added a commit to scrub the v1 compile dep (also preventing re-preparing)
one package change I made was autoconfiguration.. right now, the zipkin-server is still based on the v1 dep (basically it has v2 endpoints bolted on)

sometime in the near future we'll add a v2 zipkin-server which then removes the whole dep tree on v1 types (and spring boot 1.5). so for now, the autoconfiguration for cassandra3 makes most sense in the v1 package (as its only user is the v1 zipkin server)

zipkin-storage/cassandra3/src/main/resources/cassandra3-schema.cql
```diff
@@ -1,88 +1,100 @@
...
-    service_name  text,             //-- service name
-    span_name     text,             //-- span name, or blank for queries without span name
+CREATE TABLE IF NOT EXISTS zipkin2_cassandra3.trace_by_service_span (
+    service       text,             //-- service name
+    span          text,             //-- span name, or blank for queries without span name
     bucket        int,              //-- time bucket, calculated as ts/interval (in microseconds), for some pre-configured interval like 1 day.
     ts            timeuuid,         //-- start timestamp of the span, truncated to millisecond precision
-    trace_id      frozen<trace_id>, //-- trace ID
+    trace_id      text,             //-- trace ID
     duration      bigint,           //-- span duration, in microseconds
```
michaelsembwever: @llinder @fedj 
i want to change duration to be stored as tens of milliseconds (or in hundredths of seconds).
this will make the SASI on this column contains fewer distinct values, improving write performance.
and i can't see users being so discerning about searches they perform in the UI.

duration would be rounded up, so the result list would not skip any results. and duration in the span table would remain accurate to microseconds.

any objections?

llinder: Sounds like a good optimization to me.

codefromthecrypt: cc @openzipkin/cassandra

michaelsembwever: added as an extra commit. can be undone if any objections/concerns.

codefromthecrypt:
fyi refactored to reduce commits, but intentionally left the important ones (relevant for performance)

zipkin-storage/zipkin2_cassandra/src/main/resources/zipkin2_cassandra-schema.cql
```diff
@@ -15,8 +15,9 @@ CREATE TYPE IF NOT EXISTS zipkin2_cassandra.annotation (
 );
 
 CREATE TABLE IF NOT EXISTS zipkin2_cassandra.span (
-    trace_id            text,
+    trace_id            text, // when strictTraceId=false, only contains right-most 16 chars
     ts_uuid             timeuuid,
+    trace_id_high       text, // when strictTraceId=false, contains right-most 16 chars if present
```
michaelsembwever: did you mean here "left-most 16 chars"?

codefromthecrypt:
ok I have 23 failures locally.. trying to move that number below 20 soon :)

codefromthecrypt:
>
> >      ts_uuid             timeuuid,
> +    trace_id_high       text, // when strictTraceId=false, contains right-most 16 chars if present
>
> did you mean here "left-most 16 chars"?
>

yep!

codefromthecrypt:
16 failures locally

codefromthecrypt:
Down to 10 failures

codefromthecrypt:
about 5 failures now..

codefromthecrypt:
there's still something in the test setup that doesn't fail cleanly as 10 will fail, but if you re-run failures 4 or 5 will fail

codefromthecrypt:
According to local docker, there are 4 tests failing. I ran the first in isolation just to verify

ITCassandraStorage$SpanStoreTest#getTraces_duration
ITCassandraStorage$SpanStoreTest#getTraces_endTsAndLookback
ITCassandraStorage$SpanStoreTest#whenSpanTimestampIsMissingClientSendIsPreferred
ITCassandraStorage$SpanStoreTest#getAllServiceNames__allReturned

./mvnw -Dit.test='ITCassandraStorage$SpanStoreTest#getTraces_duration' -pl zipkin-storage/zipkin2_cassandra clean verify

codefromthecrypt:
actually getTraces_duration is the only legit fail. the others fail when getTraces_duration fails, but not when run independently (must be some lack of cleanup somewhere..)

codefromthecrypt:
I have some local failures after latest fix.. checking to see if they are a test hygiene thing

codefromthecrypt:
It appears the test failures are something I can fix without changing the code as they pass when run individually. Thanks @michaelsembwever for getting us this far!

More later, hopefully green.

codefromthecrypt:
fyi simplified default schema name to "zipkin2" as "zipkin2_cassandra" was a bit redundant once inside cassandra

codefromthecrypt:
ps an earlier change of mine caused a regression. I'll look into it tomorrow morning before working on the spark job

codefromthecrypt:
looks green!

codefromthecrypt:
ran into guava compatibility issue in spark due to transformAsync. Since the next version of cassandra driver won't have guava anyway, I yanked async composition.  A little polish is needed, but at any rate.. progress. I will try again spark after things prove green or not

codefromthecrypt:
ok spark job now works, but I need to fix tests here and there before we can merge either :)

help-wanted on cleaning up the spark job https://github.com/openzipkin/zipkin-dependencies/pull/95

codefromthecrypt:
ok I think tests are passing again, but I have some tech debt to cleanup. also zipkin-dependencies still fails 3 tests.. will swing again later

codefromthecrypt:
whoot travis is happy again

codefromthecrypt:
zipkin-dependencies now works.. will cleanup tech debt but on final lap for real

codefromthecrypt:
just refactored tests as one failing would kill the rest. each use their own keyspace now

codefromthecrypt:
fyi added a commit which pulls out call objects on the read side. The work isn't 100% complete, but I hope the overall flow is easier to read now.

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/SelectFromSpan.java
```diff
@@ -0,0 +1,209 @@
...
+      }
+      return new AutoValue_SelectFromSpan_ByTraceId(
+        session,
+        byTraceIds,
+        maxTraceCols, // amount of spans per trace is almost always larger than trace IDs
+        traceIds
+      ).flatMap(accumulateSpans);
+    }
+
+    FlatMapper<Set<String>, List<List<Span>>> newFlatMapper(int limit) {
```
codefromthecrypt: this part needs buffing

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/SelectSpanNames.java
```diff
@@ -0,0 +1,81 @@
...
+  abstract String service();
+
+  @Override public Statement newStatement() {
+    return preparedStatement().bind()
+      .setString("service", service())
+      .setInt("limit_", 1000); // no one is ever going to browse so many span names
+  }
+
+  @Override public SelectSpanNames clone() {
+    return new AutoValue_SelectSpanNames(session(), preparedStatement(), service());
```
codefromthecrypt: I will probably un-autovalue some or most of these as I need to override toString anyway as preparedStatement's toString is crap

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/internal/call/AccumulateAllResults.java
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  @AutoValue
+  static abstract class AccumulateNextResults<T> implements FlatMapper<ResultSet, T> {
+    abstract T pendingResults();
+
+    abstract BiConsumer<Row, T> accumulator();
+
+    @Override public Call<T> map(ResultSet rs) {
+      if (!rs.isFullyFetched()) rs.fetchMoreResults(); // TODO: dropped future
```
codefromthecrypt: @llinder if you know why we can drop futures in the places I made TODO, or why 2000 is special, please note and I'll backfill some comments.

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/internal/call/AccumulateAllResults.java
```diff
@@ -0,0 +1,70 @@
...
+      for (Row row : rs) {
+        accumulator().accept(row, pendingResults());
+        // TODO magic number
+        if (2000 == rs.getAvailableWithoutFetching() && !rs.isFullyFetched()) {
+          rs.fetchMoreResults(); // TODO: dropped future
+        }
+        if (0 == rs.getAvailableWithoutFetching()) break;
+      }
+      if (rs.getExecutionInfo().getPagingState() == null) return Call.create(pendingResults());
```
codefromthecrypt: I presume this means we are out of pages..

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/internal/call/AccumulateAllResults.java
michaelsembwever: `rs.fetchMoreResult()` is both a hint as well as a way to wait on the subsequent page.
here it is only being used as a hint: that is to initiate in the background the fetch of the subsequent page while we process the current page. otherwise the fetch for the next page is not initiated until this page becomes exhausted. during the processing of any specific page, multiple calls to `rs.fetchMoreResult()` will always return the same future. even if we do nothing with the returned future, when the current page is exhausted it will seamlessly move onto the next page as it has been fetched for us in the background (asynchronously).

`2000` is just an appropriate place (through the default of 5000 rows per page) to pre-emptively kick-off the asynchronous fetch of the next page.

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/internal/call/AccumulateAllResults.java
michaelsembwever: the current page is still valid. but there are no subsequent pages.

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/internal/call/AccumulateAllResults.java
codefromthecrypt: thanks. will refactor slightly to ensure call modeling is coherent with this

codefromthecrypt:
TODO: I think we are missing a spot for all storage providers as our base test classes do not query both by tag and duration.

codefromthecrypt:
did a couple things.. found an interesting bug in a test which had my head scratching. Fixing that made me realize we need to revisit how we handle limit when scanning the spans table.

Also, I moved logic around such that async fetching is explicit and only when needed. before, we were fetching in anticipation of needing to do so later. I don't think that's needed anymore as decoding the result set isn't likely to be time consuming. If this proves wrong, happy to re-introduce the early fetch, but let's document and unit test that if we do.

codefromthecrypt:
opened  as https://github.com/openzipkin/zipkin/pull/1780 as I noticed our base tests missed a spot

zipkin-storage/zipkin2_cassandra/src/main/resources/zipkin2-schema.cql
```diff
@@ -0,0 +1,108 @@
...
+);
+
+CREATE TABLE IF NOT EXISTS zipkin2.span (
+    trace_id            text, // when strictTraceId=false, only contains right-most 16 chars
+    ts_uuid             timeuuid,
+    id                  text,
+    trace_id_high       text, // when strictTraceId=false, contains left-most 16 chars if present
+    parent_id           text,
+    kind                text,
+    span                text, // span.name
```
codefromthecrypt, line 24: @michaelsembwever @llinder I noticed we have no SASI on this, which afaict means that we can't do a search for span name when service isn't present.. what should we do about this?

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/internal/call/AccumulateAllResults.java
```diff
@@ -53,18 +57,15 @@
...
+    /** Iterates through the rows in each page, flatmapping on more results until exhausted */
     @Override public Call<T> map(ResultSet rs) {
-      if (!rs.isFullyFetched()) rs.fetchMoreResults(); // TODO: dropped future
-      for (Row row : rs) {
-        accumulator().accept(row, pendingResults());
-        // TODO magic number
-        if (2000 == rs.getAvailableWithoutFetching() && !rs.isFullyFetched()) {
-          rs.fetchMoreResults(); // TODO: dropped future
-        }
-        if (0 == rs.getAvailableWithoutFetching()) break;
```
michaelsembwever: So FTR: without the pre-emptive asynchronous fetching of the next page, the parsing will stop and block every 5k rows while it fetches the next page. Queries that process more than 5k rows are going to be significantly slower (the network fetch is the slowest part).

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/CassandraSpanConsumer.java
```diff
@@ -0,0 +1,283 @@
...
+    BoundStatement bound = bind(insertTraceServiceSpanName)
+      .setString("trace_id", traceId)
+      .setString("service", serviceName)
+      .setString("span", spanName)
+      .setInt("bucket", bucket)
+      .setUUID("ts", ts_uuid);
+
+    if (null != duration) {
+      // stored as milliseconds, not microseconds
+      // round up to tens of milliseconds (or hundredths of seconds)
```
michaelsembwever: line 229 needs to be removed. rounding to centiseconds no longer happens.
(after further benchmarking it was shown to be a bad idea)

zipkin-storage/zipkin2_cassandra/README.md
```diff
@@ -66,6 +66,13 @@ Note: annotations with values longer than 256 characters
 are not written to the `annotation_query` SASI, as they aren't intended
 for use in user queries.
 
+The `trace_by_service_span` index is only used by query apis, and notably supports centisecond
+resolution duration. In other words, query inputs are rounded up to the next 10 milliseconds. For
+example, a call to GET /api/v2/traces?minDuration=12345 will returns traces who include a span that
+has at least 20 milliseconds (20000 microsecond) duration. This resolution only affects the query:
+original duration data remains at microsecond granularity. Meanwhile, write performance is
+dramatically better than writing discrete values, via fewer distinct writes.
+
```
michaelsembwever: this isn't true anymore.
(it was added and then later removed)

after further benchmarking it was seen that rounding from millisecond to centisecond was not a good idea.
(it is still rounded from microsecond to millisecond)

(this whole commit e268af9 can just be removed)

zipkin-storage/zipkin2_cassandra/src/main/resources/zipkin2-schema.cql
michaelsembwever: if you need to search by span name without service name, you'd do this via the `trace_by_service_span` table.
you'd do this via writing `(("", span_name, bucket), timestamp_uuid)` rows…

but… take care of hot partitions: if there's not a high enough cardinality on span_name then there's going to be hot partitions on those `("",span_name)` combinations. hot partitions affect write throughput and latency. it can be addressed by putting in a write cache like the `DeduplicatingExecutor`.

zipkin-storage/zipkin2_cassandra/src/main/resources/zipkin2-schema.cql
```diff
@@ -0,0 +1,108 @@
...
+    AND speculative_retry = '95percentile'
+    AND comment = 'Secondary table for looking up span names by a service name.';
+
+CREATE CUSTOM INDEX IF NOT EXISTS ON zipkin2.span (annotation_query) USING 'org.apache.cassandra.index.sasi.SASIIndex'
+   WITH OPTIONS = {
+    'mode': 'CONTAINS',
+    'analyzed': 'true',
+    'analyzer_class':'org.apache.cassandra.index.sasi.analyzer.NonTokenizingAnalyzer',
+    'case_sensitive': 'false'
+   };
```
michaelsembwever, line 85: tests that CassandraSpanStore still works smoothly after dropping the SASI index would be a fantastic addition.

this SASI index is the limiting factor in being able to truly scale this backend to Zipkin.

this, and to a lesser degree the hot partitions, are the hurdles to making this a million+ writes per second backend (even though i can't see anymore provisioning hundreds of virtual server c* nodes just for the sake of zipkin visibility…)

in benchmarking i was able to write comfortably 150 spans per second per cpu, over many hours, keeping latency generally in double-digit milliseconds. (cpu isn't really the best scaling base, but for typical cpu-ram-diskIO, it's good enough. this benchmarking was on a laptop). 
when i dropped just the one SASI on annotation_query, I could then write comfortably 6k spans per second per cpu, over many hours, keeping latency in single-digit milliseconds.

codefromthecrypt:
>
> So FTR: without the pre-emptive asynchronous fetching of the next page,
> the parsing will stop and block every 5k rows while it fetches the next
> page. Queries that process more than 5k rows are going to be significantly
> slower (the network fetch is the slowest part).
>
Good food for thought. since parsing is currently sequential regardless,
you are suggesting that "significantly slower" is a factor of parsing time
for those 5k rows, right? (or specifically however many rows are left
before we made the decision to pipeline)

ex. currently, if a request takes 3ms and parsing takes 100us, then if we
have 10k rows we take 6200us vs 6100us (or whatever offset). On the other
hand, we could do pipelining differently which doesn't sequentially parse...

I'll add a TODO on this topic. I'm not against it, just want to be able to
measure it (this stuff is fun code anyway)

michaelsembwever:
>  if a request takes 3ms and parsing takes 100us, then if we have 10k rows we take 6200us vs 6100us (or whatever offset)

that's correct :-)

codefromthecrypt:
added a WIP commit to the end, porting all the cassandra write commands to "call" objects. Only raised for the curious as I will replace the commit tomorrow with cleaner code.

Tomorrow, I need to add tests for the call deduper (done internally in InsertServiceSpan). I also need to go back and cleanup (ex make aggregate call do work in parallel, cleanup toStrings etc).

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/InsertTraceByServiceSpan.java
```diff
@@ -0,0 +1,113 @@
...
+      .setUUID("ts", input.ts())
+      .setString("trace_id", input.trace_id());
+
+    if (null != input.duration()) {
+      bound.setLong("duration", input.duration());
+    }
+    return factory.session.executeAsync(bound);
+  }
+
+  @Override public String toString() {
```
codefromthecrypt, line 106: @llinder @michaelsembwever ok I've rewritten everything at this point. everything has pretty toStrings, ex if you put a breakpoint, you can tell exactly what it is doing.

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/internal/call/AggregateCall.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import zipkin2.Call;
+import zipkin2.Callback;
+import zipkin2.internal.Nullable;
+
+public abstract class AggregateCall<I, O> extends Call.Base<O> {
```
codefromthecrypt, line 27: This type is pretty much done now, except docs. This works in parallel and when there are errors in a fan-out the call that caused the async failure is logged along with the problem.

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/internal/call/AggregateCall.java
```diff
@@ -0,0 +1,129 @@
...
+          if (remaining.decrementAndGet() == 0) {
+            callback.onSuccess(result);
+          }
+        }
+      }
+    }
+
+    @Override public synchronized void onError(Throwable throwable) {
+      if (log.isLoggable(Level.INFO)) {
+        log.log(Level.INFO, "error from " + call, throwable);
```
codefromthecrypt, line 117: ex. the "error from " + call already saved my butt today.. Some reads and writes have a fairly significant fan-out. Now you know which exact command resulted in a particular cassandra response error

codefromthecrypt:
I'd like to merge and release this tomorrow. If that's scary, scream :)

codefromthecrypt:
@openzipkin/cassandra ok I'm ready to merge this. You might want to check about how pre-computed queries are delimited in the annotation_query column. They are braced with an unlikely character to stop us from accidentally matching on substring in the API. ex bracing like `░error░` ensures we don't accidentally match on a tag key named not_error.

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/InsertSpan.java
```diff
@@ -0,0 +1,184 @@
...
+        span.timestamp(),
+        span.duration(),
+        span.localEndpoint() != null ? new EndpointUDT(span.localEndpoint()) : null,
+        span.remoteEndpoint() != null ? new EndpointUDT(span.remoteEndpoint()) : null,
+        annotations,
+        span.tags(),
+        annotation_query,
+        Boolean.TRUE.equals(span.debug()),
+        Boolean.TRUE.equals(span.shared())
+      );
```
michaelsembwever, line 127: Retouching on the null binding resulting in tombstone issue.

Here there's 8 columns here that can be null. (maybe more?)

If there's consistently 8 tombstones (nulls) per row, then we'll only need 125 spans in a trace (rows in a partition) to trigger the `tombstone_warn_threshold warnings being logged in the C* nodes. And if we go to 12500 spans in a trace then that whole trace partition would become unreadable. Cassandra warns at a 1000 tombstones in any query, and fails on 100000 tombstones.

There's also a small question about disk usage efficiency. Each tombstone is a cell name and basically empty cell value entry stored on disk. Given that the cells are, apart from tags and annotations, generally very small then this could be proportionally an unnecessary waste of disk.

To avoid this relying upon a number of variant prepared statements for inserting a span is the normal practice.

Another popular practice is to insert those potentially null columns as separate statements (and optionally put them together into UNLOGGED batches). This works as multiple writes to the same partition has little overhead, and here we're not worried about lack of isolation between those writes, as the write is asynchronous anyway. An example of this approach is in the cassandra-reaper project here: https://github.com/thelastpickle/cassandra-reaper/blob/master/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java#L622-L642

UPDATE: a few lines down, under `protected ResultSetFuture newFuture()` the nulls are in fact not written (not bound).

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/InsertSpan.java
```diff
@@ -0,0 +1,184 @@
...
+    if (null != input.l_ep()) bound.set("l_ep", input.l_ep(), EndpointUDT.class);
+    if (null != input.l_ep()) bound.setString("l_service", input.l_ep().getService());
+    if (null != input.r_ep()) bound.set("r_ep", input.r_ep(), EndpointUDT.class);
+    if (!input.annotations().isEmpty()) bound.setList("annotations", input.annotations());
+    if (!input.tags().isEmpty()) bound.setMap("tags", input.tags());
+    if (null != input.annotation_query()) {
+      bound.setString("annotation_query", input.annotation_query());
+    }
+    if (input.shared()) bound.setBool("shared", true);
+    if (input.debug()) bound.setBool("debug", true);
```
michaelsembwever, line 189: the code above avoids writing the nulls (tombstones) :-)

zipkin-storage/zipkin2_cassandra/src/test/resources/trace_by_service_span-stress.yaml
```diff
@@ -71,8 +71,8 @@ insert:
...
    select:
-    cql: SELECT * FROM trace_by_service_span WHERE service_name = ? AND span_name = ? AND bucket = ? LIMIT 1
+    cql: SELECT * FROM trace_by_service_span WHERE service = ? AND span = ? AND bucket = ? LIMIT 1
     fields: samerow
    by_duration:
-    cql: SELECT * FROM trace_by_service_span WHERE service_name = ? AND span_name = ? AND bucket = ? AND duration < ? LIMIT 1
-    fields: samerow
\ No newline at end of file
+    cql: SELECT * FROM trace_by_service_span WHERE service = ? AND span = ? AND bucket = ? AND duration < ? LIMIT 1
+    fields: samerow
```
michaelsembwever, line 78: it looks like we're missing two queries here:
 - search by timestamp range,
 - search by timestamp range and duration

michaelsembwever (commented):
all LGTM! approved.
(i added a small comment about adding two extra queries to the stress tests…)

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/InsertSpan.java
codefromthecrypt: thanks captured

zipkin-storage/zipkin2_cassandra/src/test/resources/trace_by_service_span-stress.yaml
codefromthecrypt: I couldn't quickly find out how to do a range query in a stress test, so left it out for now with a TODO

codefromthecrypt:
ok I've got the dependencies job working.. just waiting for green

codefromthecrypt:
Will release 2.3 on master green. Epic help @michaelsembwever and @llinder thanks for all the work on the experimental driver preceding this, including the spark code

PR #1764, opened 2017-10-10
Bounds the amount of data dumped to the WARN category

codefromthecrypt:
Before, when there were problems, the entire input json and any trace
IDs ended up in the server logs. This by itself can take out the server.

This takes a conservative step forward, leaving warning level, but
significantly reducing the output.


codefromthecrypt:
now it is only one line per malformed message

PR #1765, opened 2017-10-10
Prevents backlog when writing to Elasticsearch

### LINKED ISSUE #1760: Investigate how to limit backlog on Elasticsearch collector
Our elasticsearch span consumer uses okhttp under the scenes, which is subject to a default of 64 in-flight connections. We should think about a cancelation policy/algo so that we don't OOM when the elasticsearch cluster gets backed up for whatever reason. I don't see a control to limit the backlog in the okhttp dispatcher, so we might have to do something here..

here are notes from @logic-32 https://github.com/openzipkin/zipkin/issues/1332#issuecomment-334795332
> We are using Brave with a URLConnectionSender to send the data. I'm surprised that no tool could handle more than 10k spans in your workshop; we handle it with minimal problems on only one zipkin-server, no cluster. The only problem is that the ES_READ_TIMEOUT is set to 60s and it sometimes takes more than 60s to write the data out. So, the data does get written but zipkin erroneously reports them as dropped. But our specific situation that got us into that state was having a 10s messageTimeout and a tight loop that hit the database; both of which have been remedied.
> 
> Quick review of the documentation indicates that the max connections only affects the size of the pool they use; which probably means everything else would be queued and backlogged.
> 
> Again, not knowing OkHttp that well, I wonder if you can inspect the Dispatcher queue and cancel anything in the queue if you exceed a certain threshold?
Ref: https://gist.github.com/RyanRamchandar/64c5863838940ec67f03

codefromthecrypt:
In the past, delayed or otherwise unhealthy elasticsearch clusters could
create a backlog leading to a OOM arising from the http dispatcher ready
queue. This chooses to prevent a ready queue instead. This means we drop
spans when the backend isn't responding instead of crashing the server.

Fixes #1760

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
@@ -302,7 +302,7 @@ public abstract class ElasticsearchStorage extends zipkin2.storage.StorageCompon
         .build();
     ok.dispatcher().setMaxRequests(maxRequests());
     ok.dispatcher().setMaxRequestsPerHost(maxRequests());
-    return new HttpCall.Factory(ok, maxRequests(), HttpUrl.parse(hosts.get(0)));
+    return new HttpCall.Factory(ok, HttpUrl.parse(hosts.get(0)));
   }
 
   @Override public void close() {
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
@@ -107,7 +107,13 @@ public abstract class ElasticsearchStorage extends zipkin2.storage.StorageCompon
      */
     public abstract Builder hostsSupplier(HostsSupplier hosts);
 
-    /** Sets maximum in-flight requests from this process to any Elasticsearch host. Defaults to 64 */
+    /**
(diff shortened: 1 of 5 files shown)


codefromthecrypt:
cc @openzipkin/elasticsearch

codefromthecrypt:
actually I was able to crash the server even with this when setting `ES_MAX_REQUESTS=2` and 5 concurrent senders of 10 spans. Took about 5m to kill the server.

```bash
zipkin                      | 2017-10-10 12:42:26.056  INFO 5 --- [nio-9411-exec-1] o.s.web.servlet.DispatcherServlet        : FrameworkServlet 'dispatcherServlet': initialization completed in 138 ms
zipkin                      | OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x00000000f8f2f000, 107704320, 0) failed; error='Cannot allocate memory' (errno=12)
zipkin                      | #
zipkin                      | # There is insufficient memory for the Java Runtime Environment to continue.
zipkin                      | # Native memory allocation (mmap) failed to map 107704320 bytes for committing reserved memory.
zipkin                      | # An error report file with more information is saved as:
zipkin                      | # /zipkin/hs_err_pid5.log
zipkin exited with code 1
```

codefromthecrypt:
ran with less logging per https://github.com/openzipkin/zipkin/pull/1766
I simulated a surge, and definitely dropped spans works. 

I watched things recover via prometheus https://github.com/openzipkin/docker-zipkin/pull/135 and a grafana dashboard https://grafana.com/dashboards/1598/

@Logic-32 I'm going to merge this and cut 2.2.0 (which has the prometheus setup I used). Please setup a dashboard and alerts.. also you probably want to add the elasticsearch queue length to whatever that is. There are probably many places to improve, but I hope this gets things better.

PR #1766, opened 2017-10-10
Moves log messages behind drop messages to debug level

codefromthecrypt:
Otherwise, we get at least one line in the server log per message, which can fill up disks during a surge. The alternative is to watch server metrics and enable logging level as needed.

See https://github.com/openzipkin/docker-zipkin/pull/135


codefromthecrypt:
closed via 4c939617a92dada9e25bff5e11f793129c9f4b3a

PR #1781, opened 2017-11-07
InfluxDB backend

gianarb:
Hello

This PR is a WIP to add InfluxDB HTTP as zipkin backend.
Yesterday I created zipkin-storage-influxdb-http, zipkin-autoconfigure-influxdb-http modules. They are using  https://github.com/influxdata/influxdb-java to instantiate the influxdb-client.

I called this storage `influxdb-http` because InfluxDB supports UDP and I would like to keep the path open for an eventual future `influxdb-udp` implementation.


zipkin-autoconfigure/storage-influxdb-http/src/main/java/zipkin/autoconfigure/storage/influxdb/http/ZipkinInfluxDBHTTPStorageProperties.java
```diff
@@ -0,0 +1,62 @@
...
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+public class ZipkinInfluxDBHTTPStorageProperties {
+  private String address = "http://localhost:8086";
+  private String password = "";
+  private String username = "root";
+  private String retention_policy = "default";
+  private String database = "zipkin2";
```
codefromthecrypt: is this to dodge something? If not, use "zipkin" as when there's no existing storage we don't have to dodge.

zipkin-storage/influxdb-http/src/main/java/zipkin/storage/influxdb/http/InfluxDBHttpStorage.java
```diff
@@ -0,0 +1,76 @@
...
+public class InfluxDBHttpStorage extends StorageComponent implements SpanStore, SpanConsumer {
+  private InfluxDB influxDBClient;
+
+  public InfluxDBHttpStorage(InfluxDB influxDBClient) {
+    this.influxDBClient = influxDBClient;
+  }
+
+  @Override
+  public Call<Void> accept(List<Span> spans) {
+    return null;
```
codefromthecrypt: probably better to throw UnsupportedOperationException() here and elsewhere

zipkin-storage/influxdb-http/src/main/java/zipkin/storage/influxdb/http/InfluxDBHttpStorage.java
```diff
@@ -0,0 +1,76 @@
...
+import zipkin2.DependencyLink;
+import zipkin2.Span;
+import zipkin2.storage.QueryRequest;
+import zipkin2.storage.SpanConsumer;
+import zipkin2.storage.SpanStore;
+import zipkin2.storage.StorageComponent;
+
+import java.util.List;
+
+public class InfluxDBHttpStorage extends StorageComponent implements SpanStore, SpanConsumer {
```
codefromthecrypt: long term, you'll probably need to decouple these (ex not implement SpanStore or SpanConsumer rather provide instances of them)

zipkin-storage/influxdb-http/src/main/java/zipkin/storage/influxdb/http/InfluxDBHttpStorage.java
```diff
@@ -0,0 +1,76 @@
...
+  private InfluxDB influxDBClient;
```
codefromthecrypt: I think the client makes a network connection, so it should be lazy initialized and closed. you can look for similar code where `@Memoized` is used

zipkin-storage/influxdb-http/src/main/java/zipkin/storage/influxdb/http/InfluxDBHttpStorage.java
```diff
@@ -0,0 +1,76 @@
...
+
+  public InfluxDBHttpStorage(InfluxDB influxDBClient) {
```
codefromthecrypt: Make a builder even if it only calls this.. it will help you later. we don't do public constructors as defensive programming

zipkin-server/pom.xml
```diff
@@ -117,6 +117,13 @@
       <optional>true</optional>
     </dependency>
 
+    <!-- InfluxDB HTTP backend -->
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>zipkin-autoconfigure-storage-influxdb-http</artifactId>
+      <optional>true</optional>
+    </dependency>
```
codefromthecrypt: you are missing a META-INF directory in src/main/resources as well a yaml file. you can look at others for this. The META-INF contents will let the server automagically find this

codefromthecrypt (commented):
good start. made some notes to move forward (ps I would nix the -http suffix (and Http suffix on files) as unlikely we'll have multiple impls

codefromthecrypt:
ps counter-rationale is that I suspect http is the only impl that will fully support reads and writes. udp sounds like something that would only work as a span collector.

zipkin-storage/influxdb-http/src/main/java/zipkin/storage/influxdb/http/InfluxDBHttpStorage.java
gianarb: InfluxDB client supports batching

```
// Flush every 2000 Points, at least every 100ms
influxDB.enableBatch(2000, 100, TimeUnit.MILLISECONDS);
```
Do you think is it something that we can add to zipkin?

codefromthecrypt: hmm maybe, but all the more reason to lazy initialize (so that the batch processor can be closed). Can you link to the code behind this? In general it is not great to have hidden buffers, as it makes what's in flight more difficult to see.

codefromthecrypt: fwiw a collector2 module would likely duplicate this batching, for example. It matters a bit less now (provided the batching is smart enough to not write empty data)

codefromthecrypt: some other thoughts.. in tests, you'll need to disable batching or else they will fail on readback. also, probably best to hide the influxdb type (ex not accept it as a builder). this will give us flexibility later, for example if we end up with an okhttp or moshi conflict in the future, we can write native influxdb driver. for that matter, maybe best to not start with batch mode :)

gianarb: Perfect, let's keep this thinks simple and we can add more complexity if we need it

zipkin-autoconfigure/storage-influxdb/src/main/java/zipkin/autoconfigure/storage/influxdb/ZipkinInfluxDBStorageAutoConfiguration.java
```diff
@@ -0,0 +1,40 @@
...
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import zipkin.internal.V2StorageComponent;
+import zipkin.storage.StorageComponent;
+import zipkin.storage.influxdb.InfluxDBStorage;
+
+@Configuration
+@EnableConfigurationProperties(ZipkinInfluxDBStorageProperties.class)
+@ConditionalOnProperty(name = "zipkin.storage.type", havingValue = "influxdb-http")
```
codefromthecrypt: nix the -http

zipkin-server/pom.xml
```diff
@@ -117,6 +117,13 @@
       <optional>true</optional>
     </dependency>
 
+    <!-- InfluxDB HTTP backend -->
```
codefromthecrypt: nit: InfluxDB Storage

zipkin-storage/influxdb/src/main/java/zipkin/storage/influxdb/InfluxDBStorage.java
```diff
@@ -0,0 +1,130 @@
...
+
+public class InfluxDBStorage extends StorageComponent implements SpanStore, SpanConsumer {
+
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+
+  public static final class Builder extends StorageComponent.Builder {
+    boolean strictTraceId = true;
+    String address;
```
codefromthecrypt: set defaults here instead of in in the properties type. for the properties type https://github.com/openzipkin/zipkin/blob/master/zipkin-autoconfigure/collector-rabbitmq/src/main/java/zipkin/autoconfigure/collector/rabbitmq/ZipkinRabbitMQCollectorProperties.java is a better example as it ensures you only have to change properties once (here)

codefromthecrypt (commented):
left some notes.. might go ahead and add a "hint commit" to get you through more scaffolding

codefromthecrypt:
I rebased this on latest master and added a commit showing how to lazy initiate the connection, also for example, how to use it in the check() function which is used for health checks. Some of the earlier comments still apply (ex need the META-INF directory and properties could be better based on another), but hopefully this will get you further along

gianarb:
Thanks @adriancole ! I merged what @goller did around InfluxDB queries. I think now the next step is to hydrate what we get from influxdb with what zipkin requires.

zipkin-storage/influxdb/src/main/java/zipkin/storage/influxdb/InfluxDBSpanStore.java
```diff
@@ -0,0 +1,114 @@
...
+
+    throw new UnsupportedOperationException();
+  }
+
+  @Override public Call<List<Span>> getTrace(String traceId) {
+    // make sure we have a 16 or 32 character trace ID
+    traceId = Span.normalizeTraceId(traceId);
+
+    // Unless we are strict, truncate the trace ID to 64bit (encoded as 16 characters)
+    if (!strictTraceId && traceId.length() == 32) traceId = traceId.substring(16);
```
codefromthecrypt: ps this part only works if you can match on a substring of the trace ID. For example, in cassandra, we store the trace_id as either whatever was passed or right-16 characters when strictId=false. when strictId = false, we store the other part in a field trace_id_high. This field is ignored and never set unless strictId=false

In elasticsearch, we can change indexing to actually store traceID twice when strictId=false (via tokenizing). How to handle this in influxdb will depend on features.

codefromthecrypt:
probably next step is to carve out a test that uses influxdb.. start by verifying .check() works (i.e. it works by noticing the server isn't available (by pointing to a wrong url) and also works when it is available. We can use docker(testcontainers) to start influx.. you can look at RabbitMQCollectorRule or similar for an example. lemme know if you need a hand

zipkin-server/src/main/resources/zipkin-server-shared.yml
```diff
@@ -121,6 +121,13 @@ zipkin:
       db: ${MYSQL_DB:zipkin}
       max-active: ${MYSQL_MAX_CONNECTIONS:10}
       use-ssl: ${MYSQL_USE_SSL:false}
+    influxdb:
+      url: ${INFLUXDB_URL:http://localhost:8086}
+      database: ${INFLUXDB_DB:zipkin}
+      measurement: ${INFLUXDB_MEASUREMENT:zipkin}
+      retentionPolicy: ${INFLUXDB_RETETION_POLICY:default}
+      username: ${INFLUXDB_USERNAME:root}
+      password: ${INFLUXDB_PASSWORD}
```
codefromthecrypt: I think you want INFLUXDB_PASSWORD: (trailing colon which makes it default to empty string)

flipdiving:
Your site has a lot of useful information for myself. I visit regularly. Hope to have more quality items.
[flip diving](http://flipdiving.online)

codefromthecrypt:
@gianarb I kindof rejigged things to get the most test things ready to go.

One thing I noticed is that the DB driver is synchronous eventhough underneath it uses retrofit which allows asynchronous calls. 

ex you can see here that it wraps an async call and invokes it synchronously
https://github.com/influxdata/influxdb-java/blob/47e768930ba6144dfef8abd9c737471c5ee154b7/src/main/java/org/influxdb/impl/InfluxDBImpl.java#L238

The problem is that this limits us as for example, there's response mapping code ^^ that we would need, but it would be a limitation to artificially have to invoke things synchronously especially since there are asynchronous mechanisms available. Can you check upstream if it is possible to expose an asynchronous api we can use instead?

codefromthecrypt:
ps if it isn't possible to customize the upstream client, it could possibly be less work to just write the http requests directly like we do with okhttp. that's because our zipkin2.Call object has mechanisms such as map and flatMap we'd need to implement the api asynchronously

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanStore.java
```diff
@@ -0,0 +1,123 @@
...
+    Query query = new Query(q, this.storage.database());
+    QueryResult result = this.storage.get().query(query);
+    throw new UnsupportedOperationException();
+  }
+
+  @Override public Call<List<String>> getServiceNames() {
+    String queryStr = String.format("SHOW TAG VALUES FROM \"%s\" WITH KEY = \"service_name\"",
+      this.storage.measurement());
+    Query query = new Query(queryStr, this.storage.measurement());
+    QueryResult result = this.storage.get().query(query);
```
codefromthecrypt: ps to proceed, and get practice, you can continue this blocking transform. Ex get the service names from the query result into a list. then return Call.create(thatList). This helps get most of the logic where it needs to be, and allows you to start implementing tests. A future move to async invocation can build on progress like this independently

zipkin-autoconfigure/storage-influxdb/src/main/java/zipkin/autoconfigure/storage/influxdb/ZipkinInfluxDBStorageAutoConfiguration.java
```diff
@@ -0,0 +1,40 @@
...
+public class ZipkinInfluxDBStorageAutoConfiguration {
+
+  @Bean
+  @ConditionalOnMissingBean
+  V2StorageComponent storage(
+    ZipkinInfluxDBStorageProperties properties,
+    @Value("${zipkin.storage.strict-trace-id:true}") boolean strictTraceId) {
+    InfluxDBStorage storage = properties.toBuilder().strictTraceId(strictTraceId).build();
+    return V2StorageComponent.create(storage);
+  }
```
codefromthecrypt, line 39: add another bean like below as some components specifically look for v2 variants. V2StorageComponent is an adapter that is actually an instance of the v1 api.

```java
  @Bean InfluxDBStorage v2Storage(V2StorageComponent component) {
    return (InfluxDBStorage) component.delegate();
  }
```

gianarb: done

zipkin-storage/influxdb/README.md
```diff
@@ -0,0 +1,31 @@
...
+This behaviour is intentional: We don't want to burden developers with
+installing and running all storage options to test unrelated change.
+That said, all integration tests run on pull request via Travis.
+
+### Running a single test
+
+To run a single integration test, use the following syntax:
+
+```bash
+$ ./mvnw -Dit.test='ITInfluxDBStorage$SpanStoreTest#getTraces_duration' -pl zipkin-storage/influxdb clean verify
```
codefromthecrypt, line 30: @gianarb this is how to run a storage integration test from the CLI, though you can via the IDE as well

zipkin-storage/influxdb/src/test/java/zipkin2/storage/influxdb/integration/ITInfluxDBStorage.java
```diff
@@ -0,0 +1,110 @@
...
+@RunWith(Enclosed.class)
+public class ITInfluxDBStorage {
+
+  /** Written intentionally to allow you to run a single nested method via the CLI. See README */
+  static InfluxDBStorageRule classRule() {
+    return new InfluxDBStorageRule("influxdb:1.4.1-alpine", "test_zipkin3");
+  }
+
+  // This test is temporary playground to be removed later once things are implemented
+  public static class BabyStepsTest {
```
codefromthecrypt, line 39: ex you can make sure the test setup works by running this..

zipkin-storage/influxdb/src/test/java/zipkin2/storage/influxdb/integration/ITInfluxDBStorage.java
```diff
@@ -0,0 +1,110 @@
...
+    @Before @Override public void clear() {
+      InternalForTests.dropDatabase(v2Storage(), database());
+    }
+
+    @Override protected InfluxDBStorage.Builder storageBuilder() {
+      return influxdb.computeStorageBuilder();
+    }
+  }
+
+  @Ignore("Ignored as we know all tests will fail")
```
codefromthecrypt: then take off the ignore tag to see what works and what doesnt

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanConsumer.java
```diff
@@ -0,0 +1,124 @@
...
+          .time(span.timestamp(), TimeUnit.MICROSECONDS)
+          .build();
+        batch.point(annoPoint);
+      }
+
+      if (span.tags() != null) {
+        for (Map.Entry<String, String> tag : span.tags().entrySet()) {
+          Point taggedPoint = Point
+            .measurement(storage.measurement())
+            .tag("trace_id", span.traceId())
```
fmachado: Just a heads up:

> "Tags containing highly variable information like UUIDs, hashes, and random strings will lead to a large number of series in the database, known colloquially as high series cardinality. High series cardinality is a primary driver of high memory usage for many database workloads."

https://docs.influxdata.com/influxdb/v1.3/concepts/schema_and_data_layout/#don-t-have-too-many-series

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanConsumer.java
```diff
@@ -0,0 +1,124 @@
...
+            .tag("annotation", tag.getValue())
+            .tag("endpoint_host", host(span))
+            .addField("duration_ns", span.duration() * 1000)
+            .time(span.timestamp(), TimeUnit.MICROSECONDS)
+            .build();
+          batch.point(taggedPoint);
+        }
+      }
+    }
+    storage.get().write(batch);
```
fmachado: This will synchronously write to InfluxDB (https://github.com/influxdata/influxdb-java/issues/386).
If you want to do it async, enable the batch mode when you create the client:
```Java
// Flush every 2000 Points, at least every 1000ms
influxDBClient.enableBatch(2000, 1000, TimeUnit.MILLISECONDS);
```
and use the `write` method for `Point` and not `BatchPoint`.

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanStore.java
```diff
@@ -255,10 +255,15 @@ private String trimLeadingZeroes(String str){
...
-      for (QueryResult.Series series : result.getSeries()){
-        for (List<Object> values: series.getValues()) {
-          services.add(values.get(1).toString());
+    List<QueryResult.Result> results = response.getResults();
+    if (results != null) {
+      for (QueryResult.Result result : results) {
+        if (result != null && result.getSeries() != null) {
+          for (QueryResult.Series series : result.getSeries()) {
+            for (List<Object> values : series.getValues()) {
+              services.add(values.get(1).toString());
```
fmachado, line 312: This may cause a NPE if there is a null value returned by InfluxDB. Would be possible to use `String.valueOf(...)`?

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanStore.java
```diff
@@ -277,10 +282,15 @@ private String trimLeadingZeroes(String str){
...
-      for (QueryResult.Series series : result.getSeries()){
-        for (List<Object> values: series.getValues()) {
-          spans.add(values.get(1).toString());
+    List<QueryResult.Result> results = response.getResults();
+    if (results != null) {
+      for (QueryResult.Result result : results) {
+        if (result != null && result.getSeries() != null) {
+          for (QueryResult.Series series : result.getSeries()) {
+            for (List<Object> values : series.getValues()) {
+              spans.add(values.get(1).toString());
```
fmachado, line 340: This may cause a NPE if there is a null value returned by InfluxDB. Would be possible to use `String.valueOf(...)`?

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanConsumer.java
gianarb: Hello @fmachado thanks for your feedback. Cardinality is something that we are aware of. With the new TSI we will handle a high cardinality and `trace_id` needs to be indexed because it is used as lookup value during a lot of operation. Tracing from a storage point of view is a complicated and big uses cases :+1:

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanConsumer.java
gianarb: We discussed a long time ago with @adriancole about this and I think it's good as it is atm. Thanks @fmachado

gianarb:
Hello! 

We are all busy these days I would like to write down what we did and what we need to do on this PR to allow other people to work a bit on it if they have time.

All the setup for `influxdb-store` and `influxdb-autoconfigure` is done. There are modules and they are taking configurations and so on. If you start the `zipkin-server.jar` with `STORAGE_TYPE=influxdb` it will look for an influxdb running on localhost:8086.

You can start one using Docker:
```
docker run -p 8086:8086 -i influxdb
```
The storage module has queries to support:

* writes
* get spans/traces
* get span/trace
* dependencies

It should be almost everything there. We tried to remove the `@Ignore` from integration tests but they are not working. Probably some queries are not right or something else.

That's what it's still missed. Why integration tests are failing. As soon as we know it we should be able to fix what is wrong.

codefromthecrypt:
Thanks for the update on all your hard work! I'll swing the axe on this a little this weekend and report back.

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanConsumer.java
```diff
@@ -0,0 +1,124 @@
...
+      .build();
+    for (Span span : spans) {
+      Point point = Point
+        .measurement(storage.measurement())
+        .tag("trace_id", span.traceId())
+        .tag("id", span.id())
+        .tag("parent_id", span.parentId() == null ? span.id() : span.parentId())
+        .tag("name", span.name())
+        .tag("service_name", serviceName(span))
+        .addField("duration_ns", span.duration() * 1000)
```
codefromthecrypt: considering this is for zipkin, why are we changing the unit to nanoseconds?

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanConsumer.java
```diff
@@ -0,0 +1,124 @@
...
+    BatchPoints batch = BatchPoints
+      .database(storage.database())
+      .retentionPolicy(storage.retentionPolicy())
+      .build();
+    for (Span span : spans) {
+      Point point = Point
+        .measurement(storage.measurement())
+        .tag("trace_id", span.traceId())
+        .tag("id", span.id())
+        .tag("parent_id", span.parentId() == null ? span.id() : span.parentId())
```
codefromthecrypt: this is invalid.. why put the span id as parent id? If there's some special reason why, it should be in a comment.

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanConsumer.java
```diff
@@ -0,0 +1,124 @@
...
+          Point taggedPoint = Point
+            .measurement(storage.measurement())
+            .tag("trace_id", span.traceId())
+            .tag("id", span.id())
+            .tag("parent_id", span.parentId() == null ? span.id() : span.parentId())
+            .tag("name", span.name())
+            .tag("service_name", serviceName(span))
+            .tag("annotation_key", tag.getKey())
+            .tag("annotation", tag.getValue())
+            .tag("endpoint_host", host(span))
```
codefromthecrypt: why the word host? (this represents a ip:port string it looks like). please comment when doing something not zipkin conventional

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanConsumer.java
```diff
@@ -0,0 +1,124 @@
...
+      if (span.tags() != null) {
+        for (Map.Entry<String, String> tag : span.tags().entrySet()) {
+          Point taggedPoint = Point
+            .measurement(storage.measurement())
+            .tag("trace_id", span.traceId())
+            .tag("id", span.id())
+            .tag("parent_id", span.parentId() == null ? span.id() : span.parentId())
+            .tag("name", span.name())
+            .tag("service_name", serviceName(span))
+            .tag("annotation_key", tag.getKey())
```
codefromthecrypt: this is confusing.. why are we using the word annotation to describe tags? If this is about special-casing the annotationQuery API, it should be mentioned.. however, this seems like it is not for such a reason..

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanConsumer.java
```diff
@@ -0,0 +1,124 @@
...
+            .build();
+          batch.point(taggedPoint);
+        }
+      }
+    }
+    storage.get().write(batch);
+    return Call.create(null);
+  }
+
+  private String serviceName(Span span) {
```
codefromthecrypt: this looks not good :) let's not conflate the local and remote service name!

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanStore.java
```diff
@@ -0,0 +1,355 @@
...
+      if (i.hasNext()) {
+        result.append(" and ");
+      }
+    }
+    if (result.length() > 0) {
+      q += result.toString();
+    }
+
+    if (request.minDuration() != null) {
+      q += String.format(" AND \"duration_ns\" >= %d ", request.minDuration());
```
codefromthecrypt: this is incorrect because of the mapping of the table to ns which should be us. The minDuration is in us

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanStore.java
```diff
@@ -0,0 +1,355 @@
...
+    return Call.create(results);
+  }
+
+  private Span spanResults(QueryResult.Series series) {
+    List<List<Object>> values = series.getValues();
+
+    if (values == null) {
+      return null;
+    }
+    Map<String, String> tags = series.getTags();
```
codefromthecrypt, line 125: rename to seriesTags as confusing otherwise as zipkin also has a term tags

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanStore.java
codefromthecrypt: also comment on magic numbers, in this case 1

codefromthecrypt (commented):
made some notes and fixed some of them in the latest commit. There's a lot of little work to do.. It would probably be easier to not rely on integration tests and instead use MockWebServer tests so that you can easily see the queries produced by this code.

Thanks for all the work so far. Realistically speaking, I think this will take several more rounds to get passing, as there are a lot of data issues left to knock out.

goller:
@adriancole awesome ! thanks for the review!  I'll try to fix up on these this weekend.

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanConsumer.java
goller: that's the go dev in me... I'll change that to milliseconds.

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanConsumer.java
goller: good point.  I'll change this to tag.  This was from our previous notions of binary annotations.

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanConsumer.java
goller: I'll change this as well.  I was trying to decrease cardinality a small bit by combining the ip and port together into one string.

codefromthecrypt:
Micros! (But anyway there are two issues.. retrieving the unit sent, and
querying. In the case of duration, both units are micros. For span
timestamp, the query resolution is millis, but still need to retrieve in
original form (micros)

On 17 Dec 2017 8:04 am, "Chris Goller" <notifications@github.com> wrote:

*@goller* commented on this pull request.
------------------------------

In zipkin-storage/influxdb/src/main/java/zipkin2/storage/
influxdb/InfluxDBSpanConsumer.java
<https://github.com/openzipkin/zipkin/pull/1781#discussion_r157354645>:

> +  public Call<Void> accept(List<Span> spans) {
+    if (spans.isEmpty()) return Call.create(null);
+
+    BatchPoints batch = BatchPoints
+      .database(storage.database())
+      .retentionPolicy(storage.retentionPolicy())
+      .build();
+    for (Span span : spans) {
+      Point point = Point
+        .measurement(storage.measurement())
+        .tag("trace_id", span.traceId())
+        .tag("id", span.id())
+        .tag("parent_id", span.parentId() == null ? span.id() :
span.parentId())
+        .tag("name", span.name())
+        .tag("service_name", serviceName(span))
+        .addField("duration_ns", span.duration() * 1000)

that's the go dev in me... I'll change that to milliseconds.

—
You are receiving this because you were mentioned.
<https://github.com/openzipkin/zipkin/pull/1781#discussion_r157354645>, or mute
the thread

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanConsumer.java
goller: Do you think we need to have different tags for both local and remote service names?  That'll add more cardinality to the database.  Do the queries need to specify a difference between local and remote?

codefromthecrypt:
> Do you think we need to have different tags for both local and remote
service names? That'll add more cardinality to the database. Do the queries
need to specify a difference between local and remote?

At the very least, we need to reconstruct the input that was sent.
Otherwise other tools such as dependency linking wont work. Usually the raw
endpoint fields are stored regardless as we need users to see what they
wrote.

Remote service is being rediscussed here
https://github.com/openzipkin/zipkin/issues/1794

codefromthecrypt:
Ps have a look at ES code. One thing we do is a separate field for
index/search timestamp, which always has a value in millis vs the actual
timestamp. With two fields you can eliminate problems like how to handle
null when exists in raw data

zipkin-storage/influxdb/src/main/java/zipkin2/storage/influxdb/InfluxDBSpanConsumer.java
goller: Having parent set to id allows efficient querying for relationships for dependencies.

The query to find all relationships (and aggregate them server-side) is:

```sql
SELECT count(duration_us) FROM zipkin.autogen.zipkin WHERE annotation = '' GROUP BY id, parent_id, service_name, time(1d) fill(none)
```


If we set `parent_id` to `id` I can get all the relationships in one query.  If `parent_id` is not set then I'm not able to get back the root span in the count.

I'll add a comment about this for sure.

codefromthecrypt:
>
> If we set parent_id to id I can get all the relationships in one query.
> If parent_id is not set then I'm not able to get back the root span in
> the count.
>
Is there another way to accomplish this? parentid == id has caused a fair
amount of bugs, and when people look at the data they will be tempted to
assume that a root span is where parentid == id (buggy mindset). What would
happen if parentid = 0 (as opposed to unset)?

codefromthecrypt:
PS the most common problem with people doing parent ID = span ID is them accidentally causing a cycle when they assume that's how zipkin expects the data. While easy enough to resolve, this mistaken assumption has caused troubleshooting issues in dependency graph parsing and clock skew correction. We've so far avoided having to compromise by sealing this practice in a storage layer before, so I really want to make sure there's no other way as almost guaranteed this will cause problems myself or others will have to answer later.

zipkin-storage/influxdb/src/test/java/zipkin2/storage/influxdb/integration/ITInfluxDBStorage.java
codefromthecrypt: ps still valid. once tests are running here, we've the framework in place to objectively squash issues with or without unit tests (some things should be unit tested, though)

codefromthecrypt:
PS if you rebase on latest zipkin, we allow parentID=0L to effectively mean null (root span). Hope this can help

zipkin-storage/influxdb/src/test/java/zipkin2/storage/influxdb/integration/ITInfluxDBStorage.java
gianarb: That test is working 
```
./mvnw -Dlicense.skip=true -Dit.test='ITInfluxDBStorage$BabyStepsTest' -pl zipkin-storage/influxdb clean verify
```

codefromthecrypt:
a year has passed since this opened. Feel free to re-open later. we have some ASF stuff to do, so easier to inventory current work

PR #1784, opened 2017-11-10
Fixes encoding bug where core annotations not written in JSON_V1

codefromthecrypt:
Fixes https://github.com/openzipkin/zipkin-reporter-java/issues/79


codefromthecrypt:
fixed in latest zipkin/zipkin-reporter/brave

PR #1788, opened 2017-11-11
Uses shaded netty in server jar to avoid downstream conflicts.

codefromthecrypt:
gRPC uses Netty 4.1 and `netty-tcnative-boringssl-static`, which is
sensitive to Netty versions. DataStax Cassandra driver 4.0 uses Netty
4.1, but the driver isn't out, yet. By using the published shaded jar,
we can dodge the conflict for now.

I've verified that we can connect to google cloud services after this
change. I've also verified that there's no significant jar size change.

See https://github.com/grpc/grpc-java/blob/master/SECURITY.md#netty
See https://github.com/GoogleCloudPlatform/stackdriver-zipkin/issues/45

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -312,17 +312,6 @@
         <version>${project.version}</version>
       </dependency>
 
-      <!-- Needs to be defined here or else spring-boot-dependencies will override it -->
-      <dependency>
-        <groupId>com.datastax.cassandra</groupId>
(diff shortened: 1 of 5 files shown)


codefromthecrypt:
fyi @michaelsembwever @llinder

codefromthecrypt:
pps of course I also verified cassandra still works by manually starting a zipkin server with storage type set to both versions

PR #1793, opened 2017-11-14
Fixes accidental hot partition with empty service name

codefromthecrypt:
In porting over the all-services query, I accidentally created a piping
hot partition. This re-instates code similar to what we had in the old
schema, except that we can avoid an all-service fan out in more cases.


codefromthecrypt:
PS this is a pretty big bug I caused (in the hot off the press new cassandra impl).

Until this goes out a patch release, any first to fire that redid their schema yesterday will have very hot partitions (pun very intended). There's no schema change here, just we aren't slamming it anymore.

Thanks to @michaelsembwever for talking me through this.

cc @openzipkin/cassandra

codefromthecrypt:
out in 2.3.1

michaelsembwever:
thanks @adriancole 
mea culpa for not catching this.

PR #1795, opened 2017-11-14
Fix traces rendering with numeric-like `serviceNames`

dos65:
In case when we have `serviceNames` that looks like numbers, `data` function from jQuery returns it as `number` that leading to error, because `number` doesn't have method `split`.

### DIFF
--- a/zipkin-ui/js/component_ui/trace.js
+++ b/zipkin-ui/js/component_ui/trace.js
@@ -70,45 +70,77 @@ export function hideSpans(spans, parents, children, selectedSpans, childrenOnly)
   family.forEach(id => hideSpan(spans[id]));
 }
 
+function spanChildren($span) {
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
Thanks! Can you attempt a unit test to ensure this stays fixed?

dos65:
Ok, I will do it.

codefromthecrypt:
ping (aka nag :) )

dos65:
@adriancole oh, sorry for waiting. I started investigating how to test `TraceUI` component, but it's not so obvious for me.

codefromthecrypt:
test probably means extracting an exported function for `setupSpan` like we do for `showSpans`. then you can test it in trace.test.js

lemme know if you want more clues

dos65:
@adriancole it was good idea to test it, fix was incomplete )

codefromthecrypt:
Thanks and good work!

PR #1801, opened 2017-11-17
add i18n support

gsta-cloudins:
add i18n support,according to browser language

### DIFF
--- a/zipkin-ui/templates/trace.mustache
+++ b/zipkin-ui/templates/trace.mustache
@@ -1,9 +1,9 @@
 <div id='trace-controls' class='row well well-sm'>
   <ul class='nav nav-pills'>
-    <li class=''><a href='#'><strong>Duration:</strong> <span class='badge'>{{duration}}</span></a></li>
-    <li class=''><a href='#'><strong>Services:</strong> <span class='badge'>{{services}}</span></a></li>
-    <li class=''><a href='#'><strong>Depth:</strong> <span class='badge'>{{depth}}</span></a></li>
-    <li class=''><a href='#'><strong>Total Spans:</strong> <span class='badge'>{{totalSpans}}</span></a></li>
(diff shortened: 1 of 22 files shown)


zipkin-ui/templates/trace.mustache
```diff
@@ -137,7 +137,7 @@
         </table>
 
         <button class="btn btn-info btn-xs pull-right" type="button" data-toggle="collapse"
-                data-target="#moreInfo" aria-expanded="false" aria-controls="moreInfo">More Info</button>
+                data-target="#moreInfo" aria-expanded="false" aria-controls="moreInfo" data-i18n="trace.more">More Info</button>
```
codefromthecrypt, line 140: so is it normal to have the english repeated here and in the resource file? Just asking as I don't know

codefromthecrypt (commented):
Looks sensible.. can you add screen shots?

codefromthecrypt:
Thanks for getting this started!

zipkin-ui/templates/trace.mustache
shakuzen: I believe this is so if you opened this file, it could (mostly?) render as normal HTML and it would have values in it. Those values will be replaced when the whole application is running.

gzchenyong:
This pull request is done by our UI team, sorry for the wire account name :)
Please refer to the attached screen shot, including English and Chinese UI
picture1
picture2

codefromthecrypt:
there are some linting problems here https://travis-ci.org/openzipkin/zipkin/builds/303415206?utm_source=github_status&utm_medium=notification

you can change to the zipkin-ui directory and run `npm run lint` to see them locally

codefromthecrypt:
Thanks for the help!

codefromthecrypt:
@gianarb noticed this crashes when browser language is italian. There's probably an easy fix for that but.. wonder if we shouldn't crowd-source some more languages?

@shakuzen japanese?
@gianarb italian?
@marcingrzejszczak polish?
@dsyer english? (hehe)

PR #1809, opened 2017-11-21
Guard against long strings and byte arrays when encoding spans to thr…

mikewrighton:
The intention here is to apply some of the bounds checking logic that we do on thrift decode, to the encode phase. This will prevent malformed spans (spans containing very long strings) from ever being written to storage in the first place.


codefromthecrypt (commented):
Only comment is to switch to illegalargument as encoding is processing an argument (ex usually this can bubble back to the caller)

Otherwise looks great!

mikewrighton:
Good point, changed.

codefromthecrypt:
thx mike!

Issue #1811, opened 2017-11-22
Consider what to do about the prometheus library


codefromthecrypt:
Until recently, we had our own prometheus exporter. In efforts to reduce code, we used the up-and-coming standard library. Unfortunately, this library is intentionally written to not cleanup after itself, noted for architecture reasons etc.

This leads to reloading problems because a metrics filter assumes it is a JVM singleton eventhough it lives in the scope of an application context. There's currently no way to unload it (without a terrible hack which clears the whole registry)
https://github.com/prometheus/client_java/issues/279

Here's an example error which displaced a fair amount of time to understand, as one wouldn't assume that a servlet filter initializes something it doesn't destroy!
```
2017-11-22 14:04:20.770 ERROR 26628 --- [ost-startStop-1] o.a.c.c.C.[Tomcat-1].[localhost].[/]     : Exception starting filter [prometheusMetricsFilter]

java.lang.IllegalArgumentException: Collector already registered that provides name: http_request_duration_seconds_count
        at io.prometheus.client.CollectorRegistry.register(CollectorRegistry.java:54) ~[simpleclient-0.1.0.jar:na]
        at io.prometheus.client.SimpleCollector$Builder.register(SimpleCollector.java:245) ~[simpleclient-0.1.0.jar:na]
        at io.prometheus.client.SimpleCollector$Builder.register(SimpleCollector.java:237) ~[simpleclient-0.1.0.jar:na]
        at io.prometheus.client.filter.MetricsFilter.init(MetricsFilter.java:151) ~[simpleclient_servlet-0.1.0.jar:na]
        at org.apache.catalina.core.ApplicationFilterConfig.initFilter(ApplicationFilterConfig.java:285) ~[tomcat-embed-core-8.5.23.jar:8.5.23]
        at org.apache.catalina.core.ApplicationFilterConfig.<init>(ApplicationFilterConfig.java:112) ~[tomcat-embed-core-8.5.23.jar:8.5.23]
        at org.apache.catalina.core.StandardContext.filterStart(StandardContext.java:4591) [tomcat-embed-core-8.5.23.jar:8.5.23]
        at org.apache.catalina.core.StandardContext.startInternal(StandardContext.java:5233) [tomcat-embed-core-8.5.23.jar:8.5.23]
        at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:150) [tomcat-embed-core-8.5.23.jar:8.5.23]
[...]

PR #1812, opened 2017-11-22
Prevent's "Brian's bomb" by manually clearing metrics on each test class

codefromthecrypt:
As discussed in https://github.com/openzipkin/zipkin/issues/1811, the
current prometheus client isn't fit for usage in tests until
@brian-brazil changes his mind. This copy/pastes bomb prevention where
bomb is the application failing to start due to prometheus state error
on duplicate metric registration (the dupe is because a servlet filter
creates and hides a metric).


codefromthecrypt:
closed via 8973e1a6627afacdf6238cf94f3e9fffc6f3648f

PR #1825, opened 2017-11-27
Fixes IE which can't use String.endsWith

### LINKED ISSUE #1824: .endsWith javascript error in IE11
`java -jar zipkin-server-2.3.2-20171127.080232-13-exec.jar`

when accessing http://localhost:9411/zipkin/ this error pops up in the debugger:

image

this is using IE11 directly, not edge-compatibility mode

codefromthecrypt:
Fixes #1824


codefromthecrypt:
merged after circleci passed (as javascript only)

PR #1828, opened 2017-11-29
fix some chinese translation mistakes

MrGlaucus:
Still have some questions:

1. does XXX_zh means traditional chinese while zh_CN means simplified chinese? 

2. in the guide, "thrift" means the Apache Thrift or just means save?

3. also in the guide, what does url hit means? 

4. I added Service Name and Span Name labels in index, because in the guide, it referred to the Service Name and Span Name which are not pointed out in the index.

5. There are still many text that needs i18n support, such as error info, annotation query, some dialog titles... but some of them are pure text in html without a tag(which means you cannot add data-i18n attr), waiting for offical support.


codefromthecrypt:
> does XXX_zh means traditional chinese while zh_CN means simplified chinese?

dunno
>
> in the guide, "thrift" means the Apache Thrift or just means save?

english "Span names are generally thrift method names"
should be
"Span names are generally RPC method names"
thrift is too specific

> also in the guide, what does url hit means?
ex http.url=http://myservice/reset

http://myservice/reset would be the url hit (by a web browser or client)

> I added Service Name and Span Name labels in index, because in the guide, it referred to the Service Name and Span Name which are not pointed out in the index.
k

> There are still many text that needs i18n support, such as error info, annotation query, some dialog titles... but some of them are pure text in html without a tag(which means you cannot add data-i18n attr), waiting for offical support.
ok open to ideas if you have some. thanks for al the help!

shakuzen:
> does XXX_zh means traditional chinese while zh_CN means simplified chinese?

From [jquery-i18n-properties](https://github.com/jquery-i18n-properties/jquery-i18n-properties/):
> It loads and parses resource bundles (.properties) based on provided language and country codes (ISO-639 and ISO-3166) or language reported by browser.

If I understand correctly, the locale is supposed to be language code _ country code. However, for simplified vs traditional Chinese I found the following answer on SO: https://stackoverflow.com/a/4894634/3754710
Since the language code is the same for both, it is a bit more complicated than most other languages. I'm not sure what the convention is for the language code `zh` is by itself without country code.

MrGlaucus:
As my experience, singly "zh" always means Simplified Chinese, this happens when you just want to support Chinese only. if you want to support Simplified and Traditional Chinese both, we usually use "zh_CN" for Simplified Chinese and "zh_TW" for Traditional Chinese.(Of course, if you want to be more specific, there are zh_HK for Hong-Kong and zh_SG for Singapore and so on, but "zh_CN" and "zh_TW" are already enough.)

Issue #1827, opened 2017-11-29
Support amqps


codefromthecrypt:
Currently, our RabbitMQ collector (and sender) doesn't configure or test for amqps. We should fix that as this impacts the ability to use in environments such as cloud foundry

Thx for figuring this out @reshmik

related: spring-projects/spring-boot#6401

cc @shakuzen @marcingrzejszczak

shakuzen:
After significant time spent to get SSL configured on my local RabbitMQ instance, I have looked into this. The rabbit hole (pun intended) goes just a bit deeper than I originally expected.

## Status quo

### Current support
Our current support assumes (or has the constraint) that what is passed to `RABBIT_ADDRESSES` is a comma-separated list of RabbitMQ `Address`es (host:port pairs). The intention here was to easily support clustered RabbitMQ servers without a TCP load balancer. However, this does not allow specifying scheme, which is a problem here when the scheme is `amqps` instead of `amqp`. In retrospect, the constraint on `RABBIT_ADDRESSES` may have not been the best choice.

### Cloud providers
Depending on how the connection properties are available from cloud providers, the viable solutions may change. It's a shame I didn't investigate this more before.
@reshmik  @marcingrzejszczak
Can you confirm what is available to you from the RabbitMQ cloud providers you are using? Is it just one URI string like the following (with one host only, I assume)?

```
ampqs://user:pass@host[:port]/virtualHost
```

Or are the individual parts (username/password/host etc.) also provided as separate properties to some degree?

I'll try to do my due diligence on this tomorrow so we can best support what most Zipkin users may end up using. I'm out of time to look this up today, though.

---

Without a single property provided by the RabbitMQ Java client that can parse the [RabbitMQ URI spec](https://www.rabbitmq.com/uri-spec.html) **for multiple hosts**, it's not easy to support both of these options. To be fair, the URI spec doesn't define multiple hosts as valid.

## Quickest solution?

The flag for using TLS can be set separately, `ConnectionFactory#useSslProtocol`, if we want to expose that as a new property.
However, note that this spits out the following error because peer validation is not being used without configuring the trust manager (see [Key Managers, Trust Managers and Stores](https://www.rabbitmq.com/ssl.html#trust-levels)). Though no one in the linked Spring Boot issue seemed to be asking for this part to be configurable.
```
[...]

PR #1837, opened 2017-12-06
Option to use SSL with RabbitMQ

shakuzen:
Allows users to set a flag to use SSL when connecting to the RabbitMQ server via AMQPS.

See #1827

### DIFF
--- a/zipkin-collector/rabbitmq/README.md
+++ b/zipkin-collector/rabbitmq/README.md
@@ -12,11 +12,12 @@ Property | Environment Variable | Description
 --- | --- | ---
 `zipkin.collector.rabbitmq.addresses` | `RABBIT_ADDRESSES` | Comma-separated list of RabbitMQ addresses, ex. `localhost:5672,localhost:5673`
 `zipkin.collector.rabbitmq.concurrency` | `RABBIT_CONCURRENCY` | Number of concurrent consumers. Defaults to `1`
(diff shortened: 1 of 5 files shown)


shakuzen:
We still need to setup an integration test with amqps; I will try to add this later. For now, I've manually tested this with my local RabbitMQ server configured with AMQPS.

zipkin-collector/rabbitmq/README.md
```diff
@@ -17,6 +17,7 @@ Property | Environment Variable | Description
 `zipkin.collector.rabbitmq.queue` | `RABBIT_QUEUE` | Queue from which to collect span messages. Defaults to `zipkin`
 `zipkin.collector.rabbitmq.username` | `RABBIT_USER` | Username to use when connecting to RabbitMQ. Defaults to `guest`
 `zipkin.collector.rabbitmq.virtualHost` | `RABBIT_VIRTUAL_HOST` | RabbitMQ virtual host to use. Defaults to `/`
+`zipkin.collector.rabbitmq.useSsl` | `RABBIT_USE_SSL` | Set to `true` to use SSL when connecting to RabbitMQ
```
codefromthecrypt: nit when documenting and testing properties, we use lower-hyphen (although seems not the case in this README)

Ex from zipkin-server-shared.yml
```
      # Using ssl for connection, rely on Keystore
      use-ssl: ${CASSANDRA_USE_SSL:false}
```

shakuzen: Fixed

PR #1847, opened 2017-12-16
zipkin-ui: fix jumpiness on "more info" button

igorwwwwwwwwwwwwwwwwwwww:
The "more info" button on the span detail view is all jumpy. Upon clicking the button, the appearing table gets squished together.

This happens because the `display` CSS property gets set to `block`. Where in this case it should really be `table`.

We can avoid the issue by wrapping the table in an extra div, as the div does not mind the `display: block`.

The clearfix ensures the table is positioned below the "more info" button, fixing some more jumpiness during the sliding animation.

screen shot 2017-12-16 at 22 05 05


igorwwwwwwwwwwwwwwwwwwww:
Animation before:

jumpy

Animation after:

non-jumpy

codefromthecrypt:
another awesome

PR #1848, opened 2017-12-16
zipkin-ui: include minimal header on page load

igorwwwwwwwwwwwwwwwwwwww:
One of the reasons why the zipkin-ui _feels_ slow is because when you click a link, the page is first rendered as blank, and then re-rendered with the page content.

This can be improved significantly by rendering the header as part of the original `index.html` page load. This is much less visually disruptive, and makes everything feel a lot snappier.

I copied the navbar code from `layout.mustache`.

### DIFF
--- a/zipkin-ui/index.ejs
+++ b/zipkin-ui/index.ejs
@@ -13,5 +13,17 @@
     <title><%= htmlWebpackPlugin.options.title %></title>
   </head>
   <body>
+
+    <div class='navbar navbar-inverse' role='navigation'>
+      <div class='container'>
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
Maybe add xml comment above saying what is happening? Also can you screen
shot before (assuming slow enough that you can!)

igorwwwwwwwwwwwwwwwwwwww:
Page load before:

header_before

Page load after (also includes #1846):

header_after

codefromthecrypt:
lovely

PR #1850, opened 2017-12-16
zipkin-ui: re-design trace search form

igorwwwwwwwwwwwwwwwwwwww:
it's not perfect yet, but imo more organised than before.

screen shot 2017-12-17 at 00 03 24


codefromthecrypt:
Agreed! Thanks.

Lets wait for a couple more to give feedback before merge.

llinder (approved):

hexchain:
Does it feel better if the "Service name" and "Span name" line has the same width as the "Annotation Query" line?

jcchavezs:
:+1:

ImFlog:
Good initiative ! Some remarks :
- The "JSON" button is out of the box.
- I agree with @hexchain about the width
- I kinda liked the fact that the "Showing x of x ..." was under the search part, just above the results.

My 2 cents

eirslett:
Could use CSS flexbox instead of inline-block maybe... since all browsers support it now. The situation was different when the UI was first implemented by Twitter many years ago...

gianarb:
I like it @igorwwwwwwwwwwwwwwwwwwww ! Thanks

codefromthecrypt:
I guess this is awaiting small feedback fixies

FYI: I know folks have been used to me merging and releasing things, which is awesome because I have time to do it, and I prefer volunteers having time for code, docs and helping others.
Heads up: I will be out from 24 Dec back 3 Jan, so someone else will need to merge things in in that span (hehe) or wait till I'm back.

cc @openzipkin/core

igorwwwwwwwwwwwwwwwwwwww:
i've pushed some changes that I believe address all the feedback given so far. one additional change is that "showing n of k" and the "json" badge are only shown when the form was submitted.

before submit:

screen shot 2018-01-02 at 20 44 52

after submit:

screen shot 2018-01-02 at 20 45 11

codefromthecrypt:
thanks for all the help on this. We can take further changes as different PRs

PR #1851, opened 2017-12-17
Fix SelectFromSpan with 128bit TraceIDs and strict-trace-id=false

hexchain:
In `span` table, Trace IDs are stored in two fields: `trace_id` and `trace_id_high`, both of which are 16 characters long.

cc @adriancole


codefromthecrypt:
trying to make a test break around this https://github.com/openzipkin/zipkin/pull/1852

codefromthecrypt:
OK rebased over the tests that underpin this. Thanks again!

hexchain:
Guess you can just squash my prior commit into your second one, to make Git history a bit more cleaner, since that block of code changed a lot.

codefromthecrypt:
ok fixed mysql from under this change, so hopefully will build green

PR #1867, opened 2018-01-04
removes listing dupe tags from UI (only if both key & value are the same

basvanbeek:
In shared spans mode it is easy to wind up with duplicate tags (binary annotations). Some instrumentation has tried to avoid this collision by either removing this information from their client side or server side instrumentation at the expense of losing this information if one of the sides is not instrumented. 

By deduping tags in the UI view when both key and value are exactly the same we'll have all information available if one side is not instrumented while only showing dupe tags if client side and server side have different values (which can be valuable information to have and avoids us making an estimated decision on which tags are more important from either client or server side).

### DIFF
--- a/zipkin-ui/js/component_ui/spanPanel.js
+++ b/zipkin-ui/js/component_ui/spanPanel.js
@@ -19,11 +19,11 @@ function escapeHtml(string) {
 
 export function isDupeBinaryAnnotation(tagMap, anno) {
   if (!tagMap[anno.key]) {
-    tagMap[anno.key] = anno.value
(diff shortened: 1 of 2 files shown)


basvanbeek:
see also issue #989

codefromthecrypt:
looks like the right fix. thanks tons!

might be a mild pain to extract the code such that it can be unit tested..
but mind timeboxing an attempt?

basvanbeek:
sure. I'll also fix linter warnings/errors

codefromthecrypt (approved):

PR #1871, opened 2018-01-06
zipkin-ui: make dropdowns resize-friendly

igorwwwwwwwwwwwwwwwwwwww:
the jquery chosen plugin that is used for pretty dropdowns assigns the container element it creates a fixed width. this width is not updated on window resize, which can create strange effects.

this patch overrides the width style attribute of that element to be 100%, which means it will take up the full space of its container.

before (when resizing):

screen shot 2018-01-06 at 20 05 05

after:

screen shot 2018-01-06 at 20 30 00


codefromthecrypt:
thanks again!

PR #1872, opened 2018-01-06
zipkin-ui: relative interval-based lookback

igorwwwwwwwwwwwwwwwwwwww:
the default start-ts and end-ts elements in the form are imo a little difficult to use. commonly i am usually interested in some sort of relative interval, or lookback.

this patch replaces the start-ts and end-ts fields with a pre-configured dropdown of "1 hour", "3 hours", "6 hours", "12 hours", "1 day", "2 days", and "7 days".

screen shot 2018-01-06 at 21 46 48
screen shot 2018-01-06 at 21 47 11

there is also a "custom" option which reveals the start-ts and end-ts fields.

screen shot 2018-01-06 at 21 47 00

a nice side-effect from this is that refreshing the page will now show the latest traces, which can be nice when you are trying to reproduce an issue and "waiting" for the trace to show up.

### DIFF
--- a/zipkin-ui/js/component_data/default.js
+++ b/zipkin-ui/js/component_data/default.js
@@ -7,11 +7,18 @@ import {traceSummary, traceSummariesToMustache} from '../component_ui/traceSumma
 export function convertToApiQuery(windowLocationSearch) {
   const query = queryString.parse(windowLocationSearch);
   // zipkin's api looks back from endTs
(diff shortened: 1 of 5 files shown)


codefromthecrypt:
I like it. @openzipkin/core does anyone not want this?

codefromthecrypt (approved):

codefromthecrypt:
thanks a lot (Again!)

Issue #1869, opened 2018-01-06
100% sampling, aka firehose mode (mostly storage affecting)


codefromthecrypt:
to be fleshed out: this would be a mode that has more impact on tracers and transports than the server. That said, a 100% mode likely means less or no indexing on storage side, with a short (like 1hr) retention. There could be UI support at least fetch by ID which could query both the normal sampled storage and a firehose one, or it could be considered different apps. These details will be added and refreshed into a revised description with a plan.

codefromthecrypt:
here's a start https://github.com/openzipkin/zipkin/pull/1876 cc @drolando

codefromthecrypt:
Current status:

As of latest version, `SEARCH_ENABLED=false` works for elasticsearch, cassandra3 and in-memory storage, disabling all indexing. A future change #1893 should hide the Find UI as it no longer would work when search is disabled.

PR #1876, opened 2018-01-09
Adds ability to disable indexing for v2 storage implementations

codefromthecrypt:
This starts towards "firehose mode" by making it possible to disable
indexing, notably on cassandra where SASI amplifies writes into the
cluster.

When enabled, only retrieval by ID is allowed. That said, someone can
feel free to run the dependency linker if they like.. If 100% sampling
is on, the dependency links will relate to actual service calls made.

See #1869


michaelsembwever (approved):

michaelsembwever:
I notice there's no check that the SASIs have _actually_ been dropped.
With `indexing=false` do you still want the writes to be getting amplified (which will happen unless the SASI are dropped via `cql`)?

codefromthecrypt:
@michaelsembwever yeah incomplete on disabling things for CQL. I'm not sure we will check at runtime, just install without SASI on a fresh spot. We could check though

codefromthecrypt:
renamed to "search-enabled" as seems to imply more what's being disabled. feedback welcome

codefromthecrypt:
@igorwwwwwwwwwwwwwwwwwwww if /config.json has searchEnabled=false, you think you could hide the find screen?

igorwwwwwwwwwwwwwwwwwwww:
@adriancole that would be hiding the annotations query input, right? should be quite straight forward to change in zipkin-ui.

drolando:
@adriancole This is awesome! It'd have probably taken me a week to find all the places I needed to change :)

codefromthecrypt:
added a commit to not install the indexes unless search is enabled. Since a lot of prod sites will have their own copies of the schema files, they'll have to do the same on their own I guess. That or we further templatize the schema files with things like replication factor and ttl cc @openzipkin/cassandra

drolando:
I don't think you can add the replication factor in your schema since you've no idea about the datacenters names.

codefromthecrypt:
@drolando thx for the hint.. will not be ambitious then

codefromthecrypt:
added elasticsearch implementation. there are a few areas we need some unit tests, and generally some docs. We also need to have the UI conditional on the searchEnabled property

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/CassandraSpanStore.java
```diff
@@ -85,7 +99,8 @@
     final int traceIndexFetchSize = request.limit() * indexFetchMultiplier;
     List<Call<Map<String, Long>>> callsToIntersect = new ArrayList<>();
 
-    List<String> annotationKeys = CassandraUtil.annotationKeys(request);
```
codefromthecrypt: self-note: scrub this as the call is obviated higher up

zipkin-server/README.md
```diff
@@ -102,6 +102,7 @@ Attribute | Property | Description
 --- | --- | ---
 environment | zipkin.ui.environment | The value here becomes a label in the top-right corner. Not required.
 defaultLookback | zipkin.ui.default-lookback | Default duration in millis to look back when finding traces. Affects the "Start time" element in the UI. Defaults to 3600000 (1 hour in millis).
+searchEnabled | zipkin.ui.search-enabled | If the Find Traces screen is enabled. Defaults to true.
```
drolando, line 105: we also need to say something about why you'd ever want to do this

codefromthecrypt: PTAL cc also @bplotnick

codefromthecrypt:
current status: revised server README and rebased

still a few areas we need some unit tests, and generally some docs. We also need to have the UI conditional on the searchEnabled property

codefromthecrypt:
@igorwwwwwwwwwwwwwwwwwwww @Logic-32 can I ask one of you to figure out how to hide the "find a trace" and also maybe change the default content to empty or otherwise when the config.json has searchEnabled=false? I could possibly figure this out, but it will likely take orders of magnitude more time. I'll offer a trade to implement a favorite issue, like https://github.com/openzipkin/brave/issues/564 for example :D

codefromthecrypt:
pretend you don't know I'll implement the above issue anyway

drolando:
@adriancole do you want to update the UI as part of this CR? It feels long enough already :)

codefromthecrypt:
>
> @adriancole <https://github.com/adriancole> do you want to update the UI
> as part of this CR? It feels long enough already :)
>
well I like to keep master shippable. So, this question means.. is hiding
the UI an optional part of this change? Since you are the customer of this
change, if you say so, works for me!

codefromthecrypt:
So, plan is to "soft launch" this so that @drolando and others can try it and give feedback eventhough the UI won't be disabled. I'll do that as a patch release. We'll make a more broadcast announcement once feedback comes in (and any changes along with it), and ideally UI support is also in.

drolando:
> well I like to keep master shippable

Yeah, I agree with having a working master. But I usually tend to do multiple smaller PRs that change it gradually to not have a too long testing and reviewing cycle (and huge diffs). :)

You're the main maintainer though, so I'm fine with whatever you prefer.

codefromthecrypt:
ok tests are backfilled. I only need to write relevant README sections in cassandra and elasticsearch

codefromthecrypt:
ok all done with tests and README will merge on green

codefromthecrypt:
opened a follow-up issue for the UI change https://github.com/openzipkin/zipkin/issues/1893

PR #1878, opened 2018-01-10
Gives advice on how to handle cassandra's lack of duration support

### LINKED ISSUE #1877: Hide duration textbox when duration queries are not supported
What kind of issue is this?

 - [ ] Question. This issue tracker is not the place for questions. If you want to ask how to do
       something, or to understand why something isn't working the way you expect it to, use Gitter
       or Stack Overflow. https://gitter.im/openzipkin/zipkin https://stackoverflow.com/questions/tagged/zipkin

 - [ ] Bug report. If you’ve found a bug, spend the time to write a failing test. Bugs with tests
       get fixed and stay fixed. If you have a solution in mind, skip raising an issue and open a
       pull request instead.

 - [x] Feature Request. First, look at existing issues to see if the feature has been requested
       before. If you don't find anything, tell us what problem you’re trying to solve. Often a
       solution already exists! Don’t send pull requests to implement new features without first
       getting our support. Sometimes we leave features out on purpose to keep the project small.

-----

Would it be difficult to hide the "Duration (μs) >=" label and related text box when you cannot filter by duration (i.e. when using the old cassandra schema)?
Right now it's pretty confusing having that box there and then getting a "java.lang.IllegalArgumentException: getTraces with duration is unsupported" error if you write anything in it.

codefromthecrypt:
Fixes #1877

### DIFF
--- a/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanStore.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanStore.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2017 The OpenZipkin Authors
+ * Copyright 2015-2018 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
tx @drolando for the idea

drolando (approved):

codefromthecrypt:
PS the original design for cassandra duration query proved unsupportable, which was why it was pulled. It was redone in cassandra3.. one of the goals was to fix that

Issue #1460, opened 2016-12-28
zipkin ui very slow when has a lot of span


dragontree101:
we are trace is very big, and have a lot of spans, 
so zipkin's ui very slow when i choose span name or into some big trace.

i use zipkin 1.16 and use es store?  

does have any way to speed ui's performance?

codefromthecrypt:
>
> we are trace is very big, and have a lot of spans,
>
define big? spans are typically no more than 1KiB in size (often far less
than that, like hundreds of bytes)

> so zipkin's ui very slow when i choose span name or into some big trace.
>
can you show a profile from Chrome debug tools?

> i use zipkin 1.16 and use es store?
>
ok

> does have any way to speed ui's performance?
>
probably, but we'd need to try and understand what we are optimizing for.

dragontree101:
sorry i am not describe clearly
one of a trace is
![image](https://cloud.githubusercontent.com/assets/1763935/21535618/90af6a92-cdb2-11e6-83ca-e766e5cbfb53.png)

and chrome debug, i found get data from zipkin-server is fast
![image](https://cloud.githubusercontent.com/assets/1763935/21535711/c2a44d96-cdb3-11e6-8a7a-5181859c43b2.png)

and from chrome timeline i found scripting is slow, a lot of time use to execute this script
![image](https://cloud.githubusercontent.com/assets/1763935/21535760/9bb3cb70-cdb4-11e6-8f67-c2c3d6a8261a.png)

i guess exec jquery js is slow, i don't know this could be optimize?

codefromthecrypt:
yeah I think this could be optimized. @cburroughs are you interested in cracking knuckles on some performance work?

dragontree101:
i deploy by myself, not use docker, 
do you mean update zipkin to 1.18 or only update zipkin-ui from http://search.maven.org/#artifactdetails%7Cio.zipkin.ui%7Czipkin-ui%7C0.0.1%7Cjar ?

adriancole:
@dragontree101 yeah you could temporarily change the zipkin-ui jar

dragontree101:
how to change zipkin-ui?  

i change 
```
<dependency>
      <groupId>${project.groupId}</groupId>
      <artifactId>zipkin-autoconfigure-ui</artifactId>
      <optional>true</optional>
    </dependency>
```
to
```
<dependency>
    <groupId>io.zipkin.ui</groupId>
    <artifactId>zipkin-ui</artifactId>
[...]

PR #1884, opened 2018-01-15
Adding the ability to import traces from JSON into zipkin-ui

Logic-32:
I got tired of having no way to view previous traces in Zipkin.  Since it provides a way to download the trace JSON I figured there should be a way to add it back.  This adds a new link with the ability to select a JSON file for upload (as opposed to copy/paste due to the potential size of traces, ref: https://github.com/openzipkin/zipkin/issues/1460) and view it just like any other trace.

Relates to: #1747.

### DIFF
--- a/zipkin-ui/js/component_ui/uploadTrace.js
+++ b/zipkin-ui/js/component_ui/uploadTrace.js
@@ -1,0 +1,62 @@
+import {component} from 'flightjs';
+import FullPageSpinnerUI from '../component_ui/fullPageSpinner';
+import traceToMustache from '../../js/component_ui/traceToMustache';
+import {SPAN_V1} from '../spanConverter';
+
+function ensureV1(trace) {
+  if (trace == null || trace.length === 0
+          || (trace[0].localEndpoint === undefined && trace[0].remoteEndpoint === undefined)) {
+    return trace;
+  }
+
+  const newTrace = [];
+  for (let i = 0; i < trace.length; i++) {
+    newTrace.push(SPAN_V1.convert(trace[i]));
+  }
+
+  return newTrace;
+}
+
+export default component(function uploadTrace() {
+  this.doUpload = function() {
+    const files = this.node.files;
+    if (files.length === 0) {
+      return;
+    }
+
+    const reader = new FileReader();
+    reader.onload = evt => {
+      let model;
+      try {
+        let trace = JSON.parse(evt.target.result);
+        trace = ensureV1(trace);
+
+        const modelview = traceToMustache(trace);
+        model = {modelview, trace};
+      } catch (e) {
+        this.trigger('uiServerError',
+              {desc: 'Cannot parse file', message: e});
+        throw e;
+      } finally {
+        this.trigger(document, 'uiHideFullPageSpinner');
+      }
+
+      this.trigger(document, 'traceViewerPageModelView', model);
+    };
+
+    reader.onerror = evt => {
+      this.trigger(document, 'uiHideFullPageSpinner');
+      this.trigger('uiServerError',
+            {desc: 'Cannot load file', message: `${evt.target.error.name}`});
+    };
+
+    this.trigger(document, 'uiShowFullPageSpinner');
+    setTimeout(() => reader.readAsText(files[0]), 0);
+  };
+
+  this.after('initialize', function() {
+    this.on('change', this.doUpload);
+    FullPageSpinnerUI.teardownAll();
+    FullPageSpinnerUI.attachTo('#fullPageSpinner');
+  });
+});
--- a/zipkin-ui/js/component_ui/uploadTrace.js
+++ b/zipkin-ui/js/component_ui/uploadTrace.js
@@ -3,6 +3,16 @@ import FullPageSpinnerUI from '../component_ui/fullPageSpinner';
 import traceToMustache from '../../js/component_ui/traceToMustache';
 import {SPAN_V1} from '../spanConverter';
 
+function rootToFrontComparator(span1/* , span2*/) {
+  return span1.parentId === undefined ? -1 : 0;
+}
+
+function sort(trace) {
+  if (trace != null) {
+    trace.sort(rootToFrontComparator);
+  }
+}
+
 function ensureV1(trace) {
   if (trace == null || trace.length === 0
           || (trace[0].localEndpoint === undefined && trace[0].remoteEndpoint === undefined)) {
@@ -30,6 +40,7 @@ export default component(function uploadTrace() {
       try {
         let trace = JSON.parse(evt.target.result);
         trace = ensureV1(trace);
+        sort(trace);
 
         const modelview = traceToMustache(trace);
         model = {modelview, trace};
--- a/zipkin-ui/js/main.js
+++ b/zipkin-ui/js/main.js
@@ -8,6 +8,7 @@ import {compose, registry, advice, debug} from 'flightjs';
 import crossroads from 'crossroads';
 import initializeDefault from './page/default';
 import initializeTrace from './page/trace';
+import initializeTraceViewer from './page/traceViewer';
 import initializeDependency from './page/dependency';
 import CommonUI from './page/common';
 import loadConfig from './config';
@@ -21,6 +22,7 @@ loadConfig().then(config => {
 
   crossroads.addRoute(contextRoot, () => initializeDefault(config));
   crossroads.addRoute(`${contextRoot}traces/{id}`, traceId => initializeTrace(traceId, config));
+  crossroads.addRoute(`${contextRoot}traceViewer`, () => initializeTraceViewer(config));
   crossroads.addRoute(`${contextRoot}dependency`, () => initializeDependency(config));
   crossroads.parse(window.location.pathname);
 }, e => {
--- a/zipkin-ui/static/nav.properties
+++ b/zipkin-ui/static/nav.properties
@@ -2,3 +2,4 @@ nav.inves = Investigate system behavior
 nav.find = Find a trace
 nav.dep = Dependencies
 nav.search = Go to trace
+nav.viewSaved = View Saved Trace
--- a/zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiAutoConfiguration.java
+++ b/zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiAutoConfiguration.java
@@ -129,7 +129,7 @@ public class ZipkinUiAutoConfiguration extends WebMvcConfigurerAdapter {
   // If the path is a a file w/an extension, treat normally.
   // Otherwise instead of returning 404, forward to the index.
   // See https://github.com/twitter/finatra/blob/458c6b639c3afb4e29873d123125eeeb2b02e2cd/http/src/main/scala/com/twitter/finatra/http/response/ResponseBuilder.scala#L321
-  @RequestMapping(value = {"/zipkin/", "/zipkin/traces/{id}", "/zipkin/dependency"}, method = GET)
+  @RequestMapping(value = {"/zipkin/", "/zipkin/traces/{id}", "/zipkin/dependency", "/zipkin/traceViewer"}, method = GET)
   public ModelAndView forwardUiEndpoints() {
     return new ModelAndView("forward:/zipkin/index.html");
   }
--- a/zipkin-ui/js/page/traceViewer.js
+++ b/zipkin-ui/js/page/traceViewer.js
@@ -1,0 +1,75 @@
+import {component} from 'flightjs';
+import $ from 'jquery';
+import FilterAllServicesUI from '../component_ui/filterAllServices';
+import JsonPanelUI from '../component_ui/jsonPanel';
+import ServiceFilterSearchUI from '../component_ui/serviceFilterSearch';
+import SpanPanelUI from '../component_ui/spanPanel';
+import TraceUI from '../component_ui/trace';
+import FilterLabelUI from '../component_ui/filterLabel';
+import ZoomOut from '../component_ui/zoomOutSpans';
+import UploadTraceUI from '../component_ui/uploadTrace';
+import {traceViewerTemplate} from '../templates';
+import {contextRoot} from '../publicPath';
+
+const TraceViewerPageComponent = component(function TraceViewerPage() {
+  this.render = function(model) {
+    try {
+      this.$node.html(traceViewerTemplate({
+        contextRoot,
+        ...model
+      }));
+    } catch (e) {
+      this.trigger('uiServerError',
+                {desc: 'Failed to render template', message: e.message});
+    }
+
+    UploadTraceUI.attachTo('#traceFile');
+  };
+
+  this.attach = function() {
+    FilterAllServicesUI.attachTo('#filterAllServices', {
+      totalServices: $('.trace-details.services span').length
+    });
+    JsonPanelUI.attachTo('#jsonPanel');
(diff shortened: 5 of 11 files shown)


codefromthecrypt:
interesting! for a impl summary, is the trace in browser local storage or is it sortof a one-shot deal? Can you attach screen shots? Is there anything sharable between traceViewer.mustache and anything else?

Logic-32:
1. It's a one shot deal.  You refresh the page and you have to browse for the file again.
2. Attaching.
3. traceViewer.mustache _is_ trace.mustache with just a couple of small changes.  I did a horrible job committing it (it should've been a copy).  If you have a good reference for how I can share more between the two, accounting for the small changes, I'd be happy to refactor a bit.

traceimport-browse
traceimport-view

Edit: updated images to reflect new link name.

codefromthecrypt:
kindof chummy but I almost like "paste a trace" better. Import might
confuse people, or at least it did me.

Logic-32:
You da boss!  Say the words and I will make the change :)

codefromthecrypt:
some dangling concerns are that the UI doesn't natively speak v2 yet, so however worded, this implies the data is in v1 format.

The v2 format is in zipkin-js.. I wonder if there's a way we can leverage that to allow pasting of v2. If we have this feature, it should work with new data.

PS still need others feedback cc @openzipkin/core

Logic-32:
So, I just realized I messed up the commit a bit (new to and hating git, sorry).  So bear with me for a bit while I try to correct things.

Edit: Corrected but in 2 commits this time :\

Also, I do agree this should work with v2 data.  But at the moment I'm not sure where zipkin-js is or how to update it to work with v2 data.  I'd be happy to help with what I can given the right direction but for now I'll wait for the feedback from others.

zipkin-ui/js/component_ui/uploadTrace.js
```diff
@@ -0,0 +1,44 @@
...
+    const files = this.node.files;
+    if (files.length === 0) {
+      return;
+    }
+
+    const reader = new FileReader();
+    reader.onload = evt => {
+      let model;
+      try {
+        const trace = JSON.parse(evt.target.result);
```
codefromthecrypt: I think here, we should assume zipkin v2 format unless we know heuristically it is not. Worst case we can copy/paste a modified version of https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/src/jsonEncoder.js#L25 to map v2 json to v1 format.  Ex here are the tests for that: https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/test/jsonEncoder.JSON_V1.test.js

Reason is that it is only time resource which is why the UI isn't internally rewritten to v2 yet. I don't want to encourage people to keep using v1. The best heuristic to detect if incoming json is v2 or not is presence of the "localEndpoint" field, which is only absent on instrumentation error.

Logic-32: Are you suggesting that we iterate through the spans in `trace`, see if they have a 'localEndpoint', and if they do, then call `jsonEncoder.encodeV1()` (effectively) on each span so that the UI can display both v1 and v2 traces?

I'm midly afraid that v2 traces with 10k spans will suffer a bit from that.  But that concern aside, is there an easy way to extract jsonEncoder without having to copy/paste it into zipkin-ui?  Of course, I'm talking about extracting/sharing the conversion more so than the call to `JSON.stringify()` at the end.

> I think here, we should assume zipkin v2 format unless we know heuristically it is not.

Given that the UI will only show you traces in v1 format I think it is safer to assume v1 until zipkin-ui is updated to support v2.  But that doesn't mean we can't still support v2 pending the above discussion.

codefromthecrypt:
Haha if you have a trace with 9k spans you will have other problems.

In any case better to profile the concern and check what the impact is.

At some point the native format will be v2 which means this conversion will
not happen so it is temporary.

eirslett:
Looping through 10k elements in JS is still quicker than touching the DOM, I don't think it will be a problem. (But don't take my word for it, might be a good idea to profile it.) I would be concerned if it took more than 30ms.

codefromthecrypt:
Could copy/paste with a comment or add a dependency on zipkin-js and use
the functions directly. Usually we do rule of three on extracting code in
upstream projects (ex hopefully arent many use cases for converting to v1
so ideally we dont end up with 3 requests!) . Maybe see how it goes first
then we can handle that as it arises (easier to see impact this way?

Appreciate your effort and patience!

On 20 Jan 2018 7:43 am, "Adrian Cole" <adrian.f.cole@gmail.com> wrote:

Haha if you have a trace with 9k spans you will have other problems.

In any case better to profile the concern and check what the impact is.

At some point the native format will be v2 which means this conversion will
not happen so it is temporary.

Logic-32:
> Haha if you have a trace with 9k spans you will have other problems.

Lol.  Perhaps.  But it is an unfortunate fact of life in the system I work on.  It's not really all that common but I deal with the worst-of-the-worst so I have the unfortunate pleasure of seeing it daily :'(

> Looping through 10k elements in JS is still quicker than touching the DOM, I don't think it will be a problem. (But don't take my word for it, might be a good idea to profile it.) I would be concerned if it took more than 30ms.

Touche!  My head must be in the clouds right now :)

> Could copy/paste with a comment or add a dependency on zipkin-js and use
the functions directly.

Will start with copy/paste and update the pull request when I can.  No promise on ETA given work priorities and such :\

Logic-32:
Back to working on this.

> Is there anything sharable between traceViewer.mustache and anything else?

Just found out that with partials I can.
Ref:
- https://stackoverflow.com/questions/24267810/hogan-js-partials-in-pre-compiled-template?rq=1
- https://www.npmjs.com/package/mustache-loader <-- I have to add the 'noShortcut' option to do so.  Which isn't a hard refactor to do but before I go through that effort I wanted to run it by you to see how important it was before I did?

Also, I renamed "Import A Trace" to "View Saved Trace".  I added some missing i18n properties as well.  No translations, just the properties and default values.  Hopefully that doesn't mess with the foreign languages too much :\

Reference pictures in prior post updated to reflect these changes.

(commit to follow...)

zipkin-ui/static/nav.properties
```diff
@@ -1,4 +1,5 @@
 nav.inves = Investigate system behavior
 nav.find = Find a trace
 nav.dep = Dependencies
-nav.search = Go to trace
\ No newline at end of file
+nav.search = Go to trace
+nav.viewSaved = View Saved Trace
```
codefromthecrypt, line 5: @MrGlaucus @gianarb @drolando can you contribute chinese and italian text for this?

drolando:
Isn't the JSON returned by the UI valid V2 format? Couldn't you simply POST it back to /api/v2/spans and use the UI as usual?

codefromthecrypt:
@drolando right now the UI doesn't mount any write endpoints, and search etc is limited by original timestamp. to change to mount write endpoints might interfere with some folks assumptions about the security of their UI (right now, if you cut at /zipkin/* there's no way to poison or otherwise trace data). There is a different change that wasn't universally popular about writing to a separate storage service #1747.

On the technical bits, the v2 conversion etc code will end up replacing the v1 code in the project, so some of this is temporary technical debt.

hope this background is useful

Logic-32:
@drolando, in addition to what Adrian said, where I work we have a habit of attaching trace JSON to bug cases as output via the JSON button in the UI.  Without this feature, or some other tool to import the JSON, the trace JSON is pretty useless.  We have no desire to actually re-persist the JSON since our ES cluster is already pretty full of Zipkin data.  So this makes for a nice on-the-fly viewing of data with no storage overhead.

@adriancole, any updates on feedback from the core team?

codefromthecrypt:
PS I am also in favor of this feature as I routinely need to check people's
traces and right now I have to post in order to do so. For support, this is
a nice thing, as not always is the problem I'm looking at storage specific
(usually it is instrumentation specific)

zipkin-ui/js/main.js
```diff
@@ -21,6 +22,7 @@ loadConfig().then(config => {
 
   crossroads.addRoute(contextRoot, () => initializeDefault(config));
   crossroads.addRoute(`${contextRoot}traces/{id}`, traceId => initializeTrace(traceId, config));
+  crossroads.addRoute('zipkin/traceViewer', () => initializeTraceViewer(config));
```
codefromthecrypt: shouldn't this be using the contextRoot variable? (if not, those proxy-mounting zipkin will be broken I think.. look at the README for more)

codefromthecrypt:
@Logic-32 can you look at the test dir and see if you can make a similar test for anything that's new here (possibly for the conversion part, copy some test code from zipkin-js). This is a lot of new code and having at least some coverage will help if there's a bug we need to sort out later

codefromthecrypt:
PS the UI still says "view a saved trace" which is confusing as data in zipkin is saved.

I wonder if the upload functionality should be a widget next to "go to trace" Ex instead of going to a trace by ID, we are going to an offline trace (by choosing the file). Regardless, I feel a bit off about the json viewer being mixed in with server side stuff like dependencies, as it is offline..

One way is to change "go to trace" to "enter trace ID" and put another widget directly under that says "choose trace json"

I tried the functionality and notice I can't collapse any of the spans (even clicking expand-all collapse-all). Might want to check that.

zipkin-ui/js/main.js
Logic-32: Woops!  My bad.  Will fix.

Logic-32:
> the UI still says "view a saved trace" which is confusing as data in zipkin is saved.

That's because I changed it from "Import a Trace" ;)

> I wonder if the upload functionality should be a widget next to "go to trace"

Maybe?  It could certainly clear up the verbiage issue we're having.  I attached a rough mockup below of what it would look like (in IE).  Having it below the "go to trace" box isn't really practical as I don't think Bootstrap navs support multi-row very well (at least, I've never seen one).  Either way, I can't say it looks very clean.  It's not immediately obvious which textbox should be used for what purpose without some descriptive text around the file input.  Additionally, there are some workflow complications with doing it that way.  After selecting the file we'd have to change the URL to remove any reference to current state.  Otherwise somebody may try to copy/paste it and expect the same results.  Which won't happen.  We can use JS to fix the URL but we have to be careful with history management so the back/forward buttons work as expected.

So the short version is that I believe a dedicated page for viewing the traces is the simplest way to do it.  But I'm not sure how to disambiguate the online from offline stuff like you're concerned about.  We can consider a multi-level nav but that means extra clicks to get to a page.  With as simple as the UI is at the moment I don't think that's appropriate either.

mock-ie

> I tried the functionality and notice I can't collapse any of the spans (even clicking expand-all collapse-all). Might want to check that.

Works on my box in Zipkin 2.4.6 with this change applied.  Tried IE, FF, and Chrome.  Behaves identically to normal/stored traces.  Did those work for you?  Which version/branch were you using?

p.s. Will try to find a test to add when I can.

Logic-32:
Just got tests added for the span conversion stuff.  Unfortunately there is a somewhat higher level bug that may make them moot.

Currently, the /zipkin/v1/trace/{traceIdHex} has an optional query parameter to dump raw traces.  From what I can tell, getting that dump avoids some timestamp adjustments and sorting that a non-raw dump would do.  Unfortunately, it seems like /zipkin/v2/trace/{traceIdHex} defaults to the same /v1 raw behavior.  Now, it's unfortunate because the UI doesn't like it when the initial/root span isn't the first in the list of spans.  It actually fails quite miserably when that's the case.  Additionally, the lack of timestamp adjustments can cause spans to render in odd places on the timeline.  Is this something that would hold up this feature?  If so, I have no idea how I'd handle the situation at the moment.

For reference:
- We just deployed Zipkin 2.4.2 to our production servers so I was using it as my test case
- I dumped a normal v1 trace, a raw one, and a v2 trace for comparison
- The normal v1 trace rendered properly, neither of the latter did
- The raw v1 trace and v2 trace rendered identically after moving the root span up to be the first span

p.s.  How does "View Local Trace" sound?

codefromthecrypt:
Thanks for the update and all the hard work @Logic-32.

yes, we decided that in v2 it would be best to move the adjustment logic to javascript in the UI (which makes a fake root span etc), just this hasn't been done. I forgot about this.

The key part is this logic (around sorting the trace by the root span) https://github.com/openzipkin/zipkin/blob/master/zipkin/src/main/java/zipkin/internal/ApplyTimestampAndDuration.java

There is also clock skew logic, which would might be something we can punt slightly.
https://github.com/openzipkin/zipkin/blob/master/zipkin/src/main/java/zipkin/internal/CorrectForClockSkew.java

Logic-32:
Happy to help @adriancole :)  However, I'm not sure I'm up for porting those changes over.  I know Java/JS well enough but it's a lot of testing/etc... and some other priorities have come up for me.  Is it a requirement to have those working before the change can be merged?  If so, what other options do we have?

At a minimum, I could probably implement a sort routine to at least make sure the root span is first in the list.  That's about all I'd have time for.  And if I did that, would you want me to do it for raw v1 traces as well?

codefromthecrypt:
@Logic-32 thanks for being transparent about what you can budget attention for. It has been a great help so far. I think that the sort-based thing should happen after any potential conversion to v1 and before the existing logic is invoked. What I mean is that I'd solve order for v1 raw (which should transitively solve for v2). Limit to sorting and then we'll open an issue to progress further. Sound good?

Logic-32:
Sounds good to me!  Will update when I can.

zipkin-ui/static/nav.properties
MrGlaucus: I will work on the translation as the merge is done.

codefromthecrypt:
another release will be on the way soon. any chance this might be ready?

Logic-32:
Was able to look at sorting a little bit last week.  Going to try and polish/commit today :)

Edit: done.

codefromthecrypt:
testing now

codefromthecrypt:
I will test this with normal span. then merge and add code for messaging span. Good stuff!

codefromthecrypt:
there's some interesting offset rendering I'm looking into..

codefromthecrypt:
for example, the UI is misaligned unless you stretch the screen carefully

screen shot 2018-03-15 at 12 13 36 pm

screen shot 2018-03-15 at 12 13 45 pm

codefromthecrypt:
screen shot 2018-03-15 at 4 47 00 pm

Looks like the left positioning is present on trace viewer but not the normal trace screen

codefromthecrypt:
getting somewhere now..

codefromthecrypt:
ok it is because when we switch to v1 format we aren't merging by ID like the v1 api does.

screen shot 2018-03-15 at 5 08 06 pm

codefromthecrypt:
ok since this is on your master branch, I'm not going to push a commit to it. I'll make a commit to fix the merging thing and squash yours into it on the way in. will finish up tonight. Thanks for the help!

Logic-32:
To make sure I understand correctly:
1. You're trace is in v2 format and there is an issue converting to v1.
2. You're going to essentially merge this request while simultaneously fixing the bug?
3. Sorry for this being on my master branch.  Still new to git and didn't think about that.  I'm honestly going to delete the repo once it gets merged though so feel free to do whatever :)
    - Deleting the repo so I can start fresh/branch properly if I make more changes.
4.  You're welcome, and thank you!

codefromthecrypt:
Thanks @Logic-32 I squashed yours and added 355a96f2cd6b09e788cc0ecab044cd89291b6fa7 for the merge thing (as without it the UI was weird). will make follow-up now for messaging spans

Logic-32:
Awesome!  Thank you for fixing up that merging issue for me :)

PR #1890, opened 2018-01-19
Optimizes reference count of Span and squashes an errorprone warning

codefromthecrypt:
Based on https://www.youtube.com/watch?time_continue=2118&v=YLTD1GBEfwo
we were a bit noisy for no good reason. Thanks @raphw for noticing

This also squashes an error-prone warning, which while wasn't a bug
wasn't worth explaining in a suppression why it wasn't.

### DIFF
--- a/zipkin2/src/main/java/zipkin2/Span.java
+++ b/zipkin2/src/main/java/zipkin2/Span.java
@@ -148,9 +148,19 @@ public final class Span implements Serializable { // for Spark and Flink jobs
    * <p>Note: timestamps at or before epoch (0L == 1970) are invalid
    *
    * @see #duration()
+   * @see #timestampAsLong()
    */
   @Nullable public Long timestamp() {
     return timestamp > 0 ? timestamp : null;
+  }
+
+  /**
+   * Like {@link #timestamp()} except returns a primitive where zero implies absent.
+   *
+   * <p>Using this method will avoid allocation, so is encouraged when copying data.
+   */
+  public long timestampAsLong() {
+    return timestamp;
   }
 
   /**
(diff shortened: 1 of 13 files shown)


zipkin2/src/main/java/zipkin2/Span.java
```diff
@@ -243,26 +254,25 @@ public Builder toBuilder() {
     String traceId, parentId, id;
     Kind kind;
     String name;
-    Long timestamp, duration;
+    long timestamp, duration;
     Endpoint localEndpoint, remoteEndpoint;
     ArrayList<Annotation> annotations;
     TreeMap<String, String> tags;
-    Boolean debug, shared;
+    int flags = 0;
```
codefromthecrypt: ps I recall a conversation with @swankjesse that in java an int is better than a byte even if you only need a few flags. That the usage is easier and that it behaves like an int anyway in memory, iirc. correct me if wrong

codefromthecrypt:
Here are the benchmark results

Before
```
SpanBenchmarks.buildClientOnlySpan2        thrpt   15   4.222 ± 0.128  ops/us
SpanBenchmarks.buildClientOnlySpan2_clear  thrpt   15   4.303 ± 0.101  ops/us
SpanBenchmarks.buildClientOnlySpan2_clone  thrpt   15   5.754 ± 0.184  ops/us
```

After
```
SpanBenchmarks.buildClientOnlySpan2        thrpt   15   4.288 ± 0.152  ops/us
SpanBenchmarks.buildClientOnlySpan2_clear  thrpt   15   4.517 ± 0.167  ops/us
SpanBenchmarks.buildClientOnlySpan2_clone  thrpt   15   6.144 ± 0.074  ops/us
```

codefromthecrypt:
added this to follow-up in brave https://github.com/openzipkin/brave/pull/589

PR #1892, opened 2018-01-23
add trace_id_high for joining with zipkin_annotations

cgfork:
In Mysql 5.7 Reference Manual ([8.3.1 How Mysql Uses Indexes](https://dev.mysql.com/doc/refman/5.7/en/mysql-indexes.html)):

> If the table has a multiple-column index, any leftmost prefix of the index can be used by the optimizer to look up rows. For example, if you have a three-column index on (col1, col2, col3), you have indexed search capabilities on (col1), (col1, col2), and (col1, col2, col3).

In mysql.sql of zipkin:

> ALTER TABLE zipkin_spans ADD INDEX(`trace_id_high`, `trace_id`, `id`) COMMENT 'for joining with zipkin_annotations';

So, I think it should add trace_id_high condition for joining with zipkin_annotations.


codefromthecrypt:
Heh wow.. did you get a clash with mysql? Wonder if your trace id generator
is working ok as usually clashing on 128bit would require more load than
our schema would seem possible to support querying of

codefromthecrypt:
ahh I understand. So right now, our schema is biased towards 128-bit joins...

Is this the join you are looking to optimize? If so, wouldn't the index change need to occur with a query that uses it? Or is it fine to just do it like this?
```
      return ZIPKIN_SPANS.TRACE_ID_HIGH.eq(annotationTable.TRACE_ID_HIGH)
          .and(ZIPKIN_SPANS.TRACE_ID.eq(annotationTable.TRACE_ID))
          .and(ZIPKIN_SPANS.ID.eq(annotationTable.SPAN_ID));
```

zipkin-storage/mysql/src/main/java/zipkin/storage/mysql/MySQLSpanStore.java
```diff
@@ -271,7 +271,8 @@ private Endpoint endpoint(Record a) {
           .selectDistinct(ZIPKIN_SPANS.NAME)
           .from(ZIPKIN_SPANS)
           .join(ZIPKIN_ANNOTATIONS)
-          .on(ZIPKIN_SPANS.TRACE_ID.eq(ZIPKIN_ANNOTATIONS.TRACE_ID))
+          .on(ZIPKIN_SPANS.TRACE_ID_HIGH.eq(ZIPKIN_ANNOTATIONS.TRACE_ID_HIGH))
```
codefromthecrypt, line 274: guess I'm wondering if this needs to be refactored to use Schema.joinCondition() or similar?

cgfork: I think the index is not available without trace_id_high when seach or join because the trace_id_high is the first key of the INDEX(`trace_id_high`, `trace_id`, `id`)  and INDEX(`trace_id_high`, `trace_id`). So the trace_id_high added can optimize the join or the query. I think it is necessary to refactor the Schema.joinCondition(). Or change the index?

codefromthecrypt:
the index right now is not parameterized. This means it is setup for when
the trace_id_high field is present. The joinCondition primarily serves old
schemas who haven't upgraded to include the more recent trace_id_high
field. So, in this case we don't need to worry about the index part I think.

codefromthecrypt:
this should do what you want https://github.com/openzipkin/zipkin/pull/1909

cgfork:
I think this pr cloud be closed. can see #1909

PR #1894, opened 2018-01-23
Adds how to see traces to our quick start

codefromthecrypt:
A few people have asked what to do next, so I think we can improve. Last to ask was @abitrolly

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,9 @@ You can also start Zipkin via Docker.
 docker run -d -p 9411:9411 openzipkin/zipkin
 ```
 
-Once you've started, browse to http://your_host:9411 to find traces!
+Once the server is running, you can view traces with the Zipkin UI at `http://your_host:9411/zipkin/`.
+
+If your applications aren't sending traces, yet, configure them with [Zipkin instrumentation](https://zipkin.io/pages/existing_instrumentations.html) or try one of our [examples](https://github.com/openzipkin?utf8=%E2%9C%93&q=example).
 
 Check out the [`zipkin-server`](/zipkin-server) documentation for configuration details, or [`docker-zipkin`](https://github.com/openzipkin/docker-zipkin) for how to use docker-compose.
 


abitrolly:
I was following https://zipkin.io/pages/quickstart.html - does it sync?

Also `docker run -d -p 127.0.0.1:9411:9411 openzipkin/zipkin` is more safe, and http://127.0.0.1:9411 is more clickable.

codefromthecrypt:
ps no it doesn't sync. we have to copy/paste it. docker doesn't always listen on localhost, for example, if docker machine it might be 192.168.99.100

codefromthecrypt:
@abitrolly if docker is running native, wouldn't `docker run -d -p 127.0.0.1:9411:9411 openzipkin/zipkin` fail because it is already localhost?

codefromthecrypt:
@abesto @jcchavezs can either of you weigh-in on the docker advice? using docker-machine the instructions don't seem to work as just results in socket disconnect

codefromthecrypt:
merging to eliminate having to re-answer this again. any work on the links and/or most reliable docker syntax can happen later. ex in my machine (docker-machine) the port forwarding resulted in connection reset

abitrolly:
> @abitrolly if docker is running native, wouldn't `docker run -d -p 127.0.0.1:9411:9411 openzipkin/zipkin` fail because it is already localhost?

I don't see why it should fail. It will fail only if something is already listen at `127.0.0.1:9411`, not because somebody is using `127.0.0.1` with different port.

abesto:
Feels like I'm missing some context here, but would love to help out. What are you trying to accomplish? Something like this? “Start up the latest Zipkin release with Docker and send in some traces so that I can click around"? And if so, what's the problem, where's the connection reset / disconnect? Between the trace generator and Zipkin, or the browser and Zipkin?

codefromthecrypt:
The salient part here is the request to replace your_host with 127.0.0.1
and how to ensure such a suggestion won't result in more questions than it
clarifies.

For example, in the docker project we use the docker ip command vs assuming
we can map port 9411.

abesto:
Ah, I see. In that case this is what'd make sense for me: use  `127.0.0.1`, and add a note below it about `docker-machine ip`. Or we could use a `curl|sh` script that takes care of discovering `docker-machine` and figuring out the actual clickable link, though personally I like how magic-free the quick-start instructions are right now.

abitrolly:
I don't anything about `docker-machine`, so feel free to ignore the request. I just thought that `127.0.0.1` is more secure.

PR #1900, opened 2018-01-30
Fix the docs for ZIPKIN_UI_LOGS_URL for Kibana version 4, 5, and 6

xeraa:
This might need another iteration, but let's start the discussion to get the docs fixed:

IMO this has been working in Kibana 3, but has been broken with the full rewrite in version 4 (no major changes since then) though I have only tested it with Kibana 6.1. Unfortunately this got a bit more complicated after the rewrite.

### DIFF
--- a/zipkin-ui/README.md
+++ b/zipkin-ui/README.md
@@ -82,9 +82,11 @@ If you use zipkin, it's very likely that you have distributed logging sytem also
 
 One convenient way to switch from a trace to its logs is to get a button which directly links a trace to its logs.
 
-This feature can be activated by setting the property zipkin.ui.logs-url or its corresponding environment variable:
+This feature can be activated by setting the property `zipkin.ui.logs-url` or its corresponding environment variable:
 
-`ZIPKIN_UI_LOGS_URL=http://kibana.company.com/query={traceId}`
+* Kibana 3: `ZIPKIN_UI_LOGS_URL=http://kibana.company.com/query={traceId}`
+* Kibana 4 or newer: `ZIPKIN_UI_LOGS_URL=http://kibana.company.com/app/kibana#/discover?_a=(index:'filebeat-*',query:(language:lucene,query:'{traceId}'))`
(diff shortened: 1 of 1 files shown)


zipkin-ui/README.md
```diff
@@ -82,9 +82,11 @@ If you use zipkin, it's very likely that you have distributed logging sytem also
 
 One convenient way to switch from a trace to its logs is to get a button which directly links a trace to its logs.
 
-This feature can be activated by setting the property zipkin.ui.logs-url or its corresponding environment variable:
+This feature can be activated by setting the property `zipkin.ui.logs-url` or its corresponding environment variable:
 
-`ZIPKIN_UI_LOGS_URL=http://kibana.company.com/query={traceId}`
+* Kibana 3: `ZIPKIN_UI_LOGS_URL=http://kibana.company.com/query={traceId}`
+* Kibana 4 or newer: `ZIPKIN_UI_LOGS_URL=http://kibana.company.com/app/kibana#/discover?_a=(index:'filebeat-*',query:(language:lucene,query:'{traceId}'))`
+  This assumes that the log data is stored under the `filebeat-*` index pattern (replace this with the correct pattern if needed) and it uses the default timerange of 15 minutes (change it to 24 hours by adding `&_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-24h,mode:quick,to:now))` to the URL for example).
```
shakuzen, line 89: This sentence applies to the Kibana 4+ query parameters (and not the Kibana 3 URL). Perhaps this could be made more clear.

xeraa: The visualization actually looks like this:
zipkin 2018-01-30 05-38-29

But if you think this could be confusing with Kibana 3 I'll try to find another way to visualize it.

shakuzen: Sorry, I should have checked how it rendered. This looks good. Thank you!

shakuzen (approved):

codefromthecrypt:
Thanks very much. we had a question about this just a week ago

xeraa:
Yeah, they asked me that question on Gitter directly as well and that's how I started investigating. And I really like the feature :)

PR #1902, opened 2018-02-01
Keep Cassandra storage working even if the annotations query is dropped

michaelsembwever:
It should be possible to drop just the CONTAINS SASI and have Zipkin continue to start up.

(Trying to search by annotation will still ofcourse throw an exception, here a NPE)

### DIFF
--- a/zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/CassandraSpanStore.java
+++ b/zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/CassandraSpanStore.java
@@ -15,15 +15,17 @@ package zipkin2.storage.cassandra;
 
 import com.datastax.driver.core.KeyspaceMetadata;
 import com.datastax.driver.core.Session;
+import com.datastax.driver.core.exceptions.DriverException;
 import com.datastax.driver.core.utils.UUIDs;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


eirslett:
Maybe we can ship the NPE as a Server 500 error with a substatus that indicates it's due to SASI being dropped? That could eventually be used in the UI to show a sane error message.

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/CassandraSpanStore.java
```diff
@@ -73,6 +76,15 @@
     }
   }
 
+  private static SelectTraceIdsFromSpan.Factory initialiseSelectTraceIdsFromSpan(Session session) {
```
codefromthecrypt: maybe add a comment here with the same rationale as the PR desc (which this allows dropping the SASI fields)

michaelsembwever:
> Maybe we can ship the NPE as a Server 500 error with a substatus that indicates it's due to SASI being dropped? That could eventually be used in the UI to show a sane error message.

Good idea. A IllegalStateException with an appropriate message is now thrown.

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/CassandraSpanStore.java
michaelsembwever: done.

codefromthecrypt:
thx!

PR #1903, opened 2018-02-01
Index page now shows message if searching is disabled. 

### LINKED ISSUE #1893: Hide the "Find a Trace" screen when searchEnabled=false
In support of https://github.com/openzipkin/zipkin/pull/1876, we should hide the "find a trace" and also maybe change the default content to empty or otherwise when the config.json has searchEnabled=false

Logic-32:
Fixes #1893.

### DIFF
--- a/zipkin-ui/static/traces.properties
+++ b/zipkin-ui/static/traces.properties
@@ -6,6 +6,7 @@ traces.duration = Duration (μs) >=
 traces.limit = Limit
 traces.findBtn = Find Traces
 traces.sort = Sort
+traces.searchDisabled = Searching has been disabled via the searchEnabled property.
 service-percentage-desc = Service Percentage: Longest First
 service-percentage-asc = Service Percentage: Shortest First
 duration-desc = Longest First
--- a/zipkin-ui/js/page/default.js
+++ b/zipkin-ui/js/page/default.js
@@ -16,6 +16,7 @@ import TracesUI from '../component_ui/traces';
 import TimeStampUI from '../component_ui/timeStamp';
 import BackToTop from '../component_ui/backToTop';
 import {defaultTemplate} from '../templates';
+import {searchDisabled} from '../templates';
 import {contextRoot} from '../publicPath';
 import {i18nInit} from '../component_ui/i18n';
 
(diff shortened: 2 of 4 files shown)


codefromthecrypt:
sweet!

Logic-32:
:)

zipkin-ui/static/traces.properties
```diff
@@ -6,6 +6,7 @@ traces.duration = Duration (μs) >=
 traces.limit = Limit
 traces.findBtn = Find Traces
 traces.sort = Sort
+traces.searchDisabled = Searching has been disabled via the searchEnabled property.
```
codefromthecrypt, line 9: @MrGlaucus @gianarb @drolando can you contribute chinese and italian text for this?

codefromthecrypt:
I will merge this and if someone complains re-nag about i18n

drolando:
Where is that message supposed to appear? I'm running 2.5.2 with SEARCH_ENABLED=false but I don't see any difference from the normal UI

Logic-32:
I'm not sure where all @adriancole  merged that change.  According to the tags on the merge it should be in 2.5.2 but 2.5.3 seems the most likely to have it.

If there, it would disable the "Find a Trace" page and show a big red box instead of the search controls.

codefromthecrypt:
from latest it actually appears the property is set but not having the
desired effect...


ZIPKIN_UI_SEARCH_ENABLED=false java -jar zipkin.jar

curl -s localhost:9411/configprops|jq
'."zipkin.ui-zipkin.autoconfigure.ui.ZipkinUiProperties"'

*{*

  *"prefix"**: *"zipkin.ui"*,*

  *"properties"**: {*

    *"defaultLookback"**: *3600000*,*

    *"environment"**: *""*,*

    *"instrumented"**: *".*"*,*

    *"dependency"**: {*

      *"lowErrorRate"**: *0.5*,*

      *"highErrorRate"**: *0.75

*    },*

    *"basepath"**: *"/zipkin"*,*

    *"queryLimit"**: *10*,*

    *"logsUrl"**: null,*

    *"searchEnabled"**: *false

*  }*

*}*

Logic-32:
This is kinda funny...
```
// config.js
export default function loadConfig() {
  return $.ajax('config.json', {
    type: 'GET',
    dataType: 'json'
  }).then(data => function config(key) {
    return data[key] || defaults[key];
  });
```

Since we set `searchEnabled` to false, it defaults to the default.  Because that's how OR statements work :\

PR #1915, opened 2018-02-10
validate hasNext() on itrerator

igor-suhorukov:
By contract, any implementation of the java.util.Iterator.next() method should throw a NoSuchElementException exception when the iteration has no more elements.


codefromthecrypt:
fair enough. thanks!

PR #1916, opened 2018-02-12
zipkin-ui: fixed slow meaningless request in home page

### LINKED ISSUE #1914: zipkin-ui: slow meaningless request in home page
Recently we upgraded to zipkin 2.4.5 and found that there is a request(`/api/v1/traces`, without any params) will be sent when opening the home page.

This request is very slow (about 10s), causing the page stuck for a while, and the front end will not render the result to the page, so it is meaningless.

Before #1789, the request will be sent only when containing the `serviceName` param. But in #1872, which removed all constraints, an request would be sent even there is no params at all, and it will be very slow because there is no any filters in elasticsearch.

The request can still get result, but in `page/default.js`, because `serviceName` does not exist, `queryWasPerformed` will be `false`, then traces will not be rendered to the page.

I want bring the logic(before #1789) back (revert part of #1872, re-add the `serviceName` constraints), but in another way. Code below:

```
export default component(function DefaultData() {
  this.after('initialize', function() {
    const query = queryString.parse(window.location.search);
    if (query.serviceName) {    // <- serviceName in window.location.search
      const apiQuery = convertToApiQuery(query);    // <- then convert to api query
      const apiURL = `api/v1/traces?${queryString.stringify(apiQuery)}`;
      $.ajax(apiURL, {
        type: 'GET',
        dataType: 'json'
      }).done(traces => {
        const modelview = {
          traces: traceSummariesToMustache(query.serviceName, traces.map(traceSummary)),
          apiURL,
          rawResponse: traces
        };
        this.trigger('defaultPageModelView', modelview);
      }).fail(e => {
        this.trigger('defaultPageModelView', {traces: [],
                                              queryError: errToStr(e)});
      });
    } else {
      this.trigger('defaultPageModelView', {traces: []});
    }
  });
});
```

It would be nice if you agree with that, and I could submit a Pull Request to fix this issue.

ScienJus:
Fixed #1914

### DIFF
--- a/zipkin-ui/js/component_data/default.js
+++ b/zipkin-ui/js/component_data/default.js
@@ -31,14 +31,19 @@ export function convertToApiQuery(windowLocationSearch) {
 
 export default component(function DefaultData() {
   this.after('initialize', function() {
-    const query = convertToApiQuery(window.location.search);
-    const apiURL = `api/v1/traces?${queryString.stringify(query)}`;
+    const query = queryString.parse(window.location.search);
+    if (!query.serviceName) {
+      this.trigger('defaultPageModelView', {traces: []});
+      return;
+    }
+    const apiQuery = convertToApiQuery(window.location.search);
+    const apiURL = `api/v1/traces?${queryString.stringify(apiQuery)}`;
     $.ajax(apiURL, {
       type: 'GET',
       dataType: 'json'
     }).done(traces => {
       const modelview = {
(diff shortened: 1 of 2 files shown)


ScienJus:
@adriancole @igorwwwwwwwwwwwwwwwwwwww 

please review this.

codefromthecrypt:
thanks for the help!

PR #1917, opened 2018-02-13
Polish: lambdas should be replaced with method references

igor-suhorukov:


codefromthecrypt:
thanks for the polish.. shines better now

PR #1891, opened 2018-01-19
Removes UDT workaround using new Datastax driver

codefromthecrypt:
The ideal is to get rid of the dependency on the zipkin_udts keyspace and prevent below

```
Caused by: java.lang.IllegalArgumentException: Keyspace zipkin2_udts does not exist
    at com.datastax.driver.mapping.AnnotationParser.parseUDT(AnnotationParser.java:150) ~[cassandra-driver-mapping-3.4.0.jar!/:na]
    at com.datastax.driver.mapping.MappingManager.getUDTCodec(MappingManager.java:320) ~[cassandra-driver-mapping-3.4.0.jar!/:na]
    at com.datastax.driver.mapping.MappingManager.udtCodec(MappingManager.java:270) ~[cassandra-driver-mapping-3.4.0.jar!/:na]
    at com.datastax.driver.mapping.MappingManager.udtCodec(MappingManager.java:282) ~[cassandra-driver-mapping-3.4.0.jar!/:na]
    at zipkin2.storage.cassandra.DefaultSessionFactory.initializeUDTs(DefaultSessionFactory.java:89) ~[io.zipkin.zipkin2-zipkin-storage-cassandra-2.4.7.jar!/:na]
    at zipkin2.storage.cassandra.DefaultSessionFactory.create(DefaultSessionFactory.java:72) ~[io.zipkin.zipkin2-zipkin-storage-cassandra-2.4.7.jar!/:na]
    at zipkin2.storage.cassandra.CassandraStorage.session(CassandraStorage.java:161) ~[io.zipkin.zipkin2-zipkin-storage-cassandra-2.4.7.jar!/:na]
    at zipkin2.storage.cassandra.AutoValue_CassandraStorage.session(AutoValue_CassandraStorage.java:32) ~[io.zipkin.zipkin2-zipkin-storage-cassandra-2.4.7.jar
```

PR #1918, opened 2018-02-14
Installs Cassandra UDT keyspace only when allowed

codefromthecrypt:
Tested this locally to make sure we're all good. also re-pushed the docker test image to make sure it always installs the UDT keyspace

 cc @mvallebr @llinder @michaelsembwever

### DIFF
--- a/zipkin-server/src/main/java/zipkin/server/brave/BraveConfiguration.java
+++ b/zipkin-server/src/main/java/zipkin/server/brave/BraveConfiguration.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2017 The OpenZipkin Authors
+ * Copyright 2015-2018 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
--- a/zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/DefaultSessionFactory.java
+++ b/zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/DefaultSessionFactory.java
@@ -58,6 +58,7 @@ final class DefaultSessionFactory implements CassandraStorage.SessionFactory {
       if (cassandra.ensureSchema()) {
         session = closer.register(cluster.connect());
         Schema.ensureExists(cassandra.keyspace(), cassandra.searchEnabled(), session);
+        Schema.ensureExists(DEFAULT_KEYSPACE + "_udts", false, session);
         session.execute("USE " + cassandra.keyspace());
       } else {
         session = cluster.connect(cassandra.keyspace());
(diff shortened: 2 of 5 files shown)


zipkin-server/src/main/java/zipkin/server/brave/BraveConfiguration.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2017 The OpenZipkin Authors
+ * Copyright 2015-2018 The OpenZipkin Authors
```
codefromthecrypt, line 2: ignore this.. just didn't want to litter with a commit update to satisfy license plugin

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/DefaultSessionFactory.java
```diff
@@ -58,6 +58,7 @@
       if (cassandra.ensureSchema()) {
         session = closer.register(cluster.connect());
         Schema.ensureExists(cassandra.keyspace(), cassandra.searchEnabled(), session);
+        Schema.ensureExists(DEFAULT_KEYSPACE + "_udts", false, session);
```
mvallebr: Can I suggest you to add more debug logs on key places? It was really hard for me to find cassandra.ensureSchema was really set to false, I thought it could be an error to set the variable... I think it's worth logging it as well, as debug

Not on this PR, but I think it would be great as well to have a command line option on zipkin server to dump all spring config. It's not something hard to do, but extremely helpful for the user.

mvallebr:
This LGTM, from user's point of view. Please let me know as soon as you release a version with this, I am looking forward to it

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/DefaultSessionFactory.java
codefromthecrypt: Spring endpoints already exist for these things.. noted here https://github.com/openzipkin/zipkin/blob/master/zipkin-server/README.md#endpoints

for example, if you hit /configprops you'd see if ensureSchema was set to false or not. If I weren't distracted while you were troubleshooting, I would probably have mentioned to use that. there's also /autoconfig, too

https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html

I use both of these a lot.. while we actively decided to not replicate spring docs, it might make sense to note both of these in the server README.

It is true that there aren't many debug statements in the storage code. Will add a couple.

codefromthecrypt:
added more logging details in the READMEs

zipkin-storage/zipkin2_cassandra/src/main/java/zipkin2/storage/cassandra/DefaultSessionFactory.java
```diff
@@ -55,12 +58,15 @@
       Cluster cluster = closer.register(buildCluster(cassandra));
       cluster.register(new QueryLogger.Builder().build());
       Session session;
+      String keyspace = cassandra.keyspace();
       if (cassandra.ensureSchema()) {
         session = closer.register(cluster.connect());
-        Schema.ensureExists(cassandra.keyspace(), cassandra.searchEnabled(), session);
-        session.execute("USE " + cassandra.keyspace());
+        Schema.ensureExists(keyspace, cassandra.searchEnabled(), session);
+        Schema.ensureExists(DEFAULT_KEYSPACE + "_udts", false, session);
```
mvallebr, line 65: Just out of curiosity, when ensureSchema is enabled, aren't you creating 2 keyspaces here?

codefromthecrypt:
>
> Just out of curiosity, when ensureSchema is enabled, aren't you creating 2
> keyspaces here?
>
yes. one is a template for the UDTs (due to them being defined on
annotations I think) I tried to remove it by
https://github.com/openzipkin/zipkin/pull/1891, but seems like a different
topic.

Maybe @michaelsembwever @llinder can help us write a better comment or note
in the readme. Even if replying here, I don't mind copy/pasting

PR #1922, opened 2018-02-18
Allow the RabbitMQ collector to be configured via URI

### LINKED ISSUE #1901: RabbitMQ collector configuration should support a URI
Some cloud service offerings for RabbitMQ may only expose a connection URI. There is no easy way to use this with Zipkin Server now to connect to the RabbitMQ service. The application would have to parse the provided URI into its parts and set those.

The complication here is that multiple addresses cannot be used in the RabbitMQ [URI spec](https://www.rabbitmq.com/uri-spec.html). Originally one of the goals of the exposed configuration was to support RabbitMQ clusters that do not have a TCP load balancer or VIP in front of their nodes.

Some of this is explained in #1827.

Without changes to the RabbitMQ URI spec or Java client, I think our best option is to expose a `RABBIT_URI` configuration property. Behavior of the other configuration properties when this is set would have to be defined (i.e., when URI is set, ignore individual properties for any parts of the URI).

/cc @marcingrzejszczak

shakuzen:
In addition to the existing methods for configuration, this adds the ability to configure the RabbitMQ server to connect to via a RabbitMQ URI spec-compliant URI. When the URI is used, the other connection properties will be ignored, since they should be set via the URI.

Resolves #1901

### DIFF
--- a/zipkin-autoconfigure/collector-rabbitmq/src/main/java/zipkin/autoconfigure/collector/rabbitmq/ZipkinRabbitMQCollectorProperties.java
+++ b/zipkin-autoconfigure/collector-rabbitmq/src/main/java/zipkin/autoconfigure/collector/rabbitmq/ZipkinRabbitMQCollectorProperties.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2017 The OpenZipkin Authors
+ * Copyright 2015-2018 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -14,6 +14,8 @@
 package zipkin.autoconfigure.collector.rabbitmq;
 
 import com.rabbitmq.client.ConnectionFactory;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.util.List;
@@ -42,6 +44,9 @@ public class ZipkinRabbitMQCollectorProperties {
   private String virtualHost;
   /** Flag to use SSL */
   private Boolean useSsl;
+  /** RabbitMQ URI spec-compliant URI to connect to the RabbitMQ server.
+   * When used, other connection properties will be ignored. */
+  private URI uri;
 
   public List<String> getAddresses() {
     return addresses;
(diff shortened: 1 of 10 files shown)


zipkin-autoconfigure/collector-rabbitmq/src/main/java/zipkin/autoconfigure/collector/rabbitmq/ZipkinRabbitMQCollectorProperties.java
```diff
@@ -107,18 +112,31 @@ public void setUseSsl(Boolean useSsl) {
...
+    return uri;
+  }
+
+  public void setUri(URI uri) {
+    this.uri = uri;
+  }
+
   public RabbitMQCollector.Builder toBuilder()
-    throws KeyManagementException, NoSuchAlgorithmException {
+    throws KeyManagementException, NoSuchAlgorithmException, URISyntaxException {
```
codefromthecrypt, line 124: interesting that it throws so many things. It doesn't really matter though as we don't have api compatibility warranty on autoconfigure or zipkin-server types (though we should mention in a readme this hidden trivia)

zipkin-autoconfigure/collector-rabbitmq/src/test/java/zipkin/autoconfigure/collector/rabbitmq/ZipkinRabbitMQCollectorAutoConfigurationTest.java
```diff
@@ -58,9 +58,9 @@ public void doesNotProvideCollectorComponent_whenAddressNotSet() {
   }
```
codefromthecrypt, line 59: maybe just for good measure, add a test that when you set URI it is readable and hosts is ignored?

zipkin-collector/rabbitmq/README.md
```diff
@@ -10,11 +10,17 @@ The following configuration can be set for the RabbitMQ Collector.
...
 Property | Environment Variable | Description
 --- | --- | ---
-`zipkin.collector.rabbitmq.addresses` | `RABBIT_ADDRESSES` | Comma-separated list of RabbitMQ addresses, ex. `localhost:5672,localhost:5673`
 `zipkin.collector.rabbitmq.concurrency` | `RABBIT_CONCURRENCY` | Number of concurrent consumers. Defaults to `1`
 `zipkin.collector.rabbitmq.connection-timeout` | `RABBIT_CONNECTION_TIMEOUT` | Milliseconds to wait establishing a connection. Defaults to `60000` (1 minute)
-`zipkin.collector.rabbitmq.password` | `RABBIT_PASSWORD`| Password to use when connecting to RabbitMQ. Defaults to `guest`
 `zipkin.collector.rabbitmq.queue` | `RABBIT_QUEUE` | Queue from which to collect span messages. Defaults to `zipkin`
+`zipkin.collector.rabbitmq.uri` | `RABBIT_URI` | [RabbitMQ URI spec](https://www.rabbitmq.com/uri-spec.html)-compliant URI, ex. `amqp://user:pass@host:10000/vhost`
+
+If the URI is set, the following properties will be ignored.
```
codefromthecrypt, line 18: makes sense

codefromthecrypt (approved):
Looks good. one minor thing which is I think we should have a test that addresses are not set when uri is (eventhough it is easy to see).

zipkin-autoconfigure/collector-rabbitmq/src/test/java/zipkin/autoconfigure/collector/rabbitmq/ZipkinRabbitMQCollectorAutoConfigurationTest.java
shakuzen: Added https://github.com/openzipkin/zipkin/pull/1922/files#diff-e3e00988ebb1a8c1c756f3a67a6d533f

codefromthecrypt:
very nice!

codefromthecrypt:
haha nice branch name!

Issue #1874, opened 2018-01-07
Investigate an http.route tag


codefromthecrypt:
I'm going to recommend we make a standard tag of `http.route`, a way to help people make good low-cardinality names or tags from http requests. Main idea is that instead of parsing `http.url` which can be trouble, allow frameworks to provide `http.route` which might not always be the same format, but would always be variable and parameter free. It also has a higher likelihood of matching 255 char indexing constraints, which are common on storage backends.

Here are some examples of http route values for the same endpoint:

* vert.x web route "/items/:itemId"
* Play "/items/$id<[0-9]+>"
* Spring WebMVC request mapping "/items/{itemId}"

Until we have something like this, it is almost nonsense to answer some folks who want to push spans into metrics (see #1869). For example, in prometheus, counters etc will reset on each small deviation of tags, a problem far worse than ours wrt cardinality.

Example framework relationships:

* Jersey - response has the template
* ratpack - [response has the template](https://github.com/hyleung/ratpack-zipkin/blob/072e4cc3b95559a0e6a764b418db58dc52457d95/src/main/java/ratpack/zipkin/ServerResponse.java)
* spring - attribute includes the template
* sparkjava - [not yet supported](https://github.com/perwendel/spark/issues/959)
* play - [via tags](https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/core/routing/GeneratedRouter.scala#L234-L238) [kamon example](https://github.com/kamon-io/kamon-play/blob/master/kamon-play-2.6.x/src/main/scala/kamon/play/Play.scala)

In interop tests, we can't necessarily test for equality of this value as it will be framework specific. However, we can test for cardinality! Ex we make an endpoint with a path variable and execute it against two values of that. Then, we test `http.path` tag is different while `http.template` is the same. Ideally the http template is already "washed", but those derived from paths might need normalization. For example, you may need to ensure there's only one leading slash.

Besides a lookup tag, `http.template` can be used to create a better name for the span. For example, if a template is available, it could be a better choice than the http method. Implementation-wise, this implies a span rename feature, as templates are not always known before-the-fact. That said a primary use case is navigating to other systems like metrics, whether that's directly via the http.template tag or something derived from it.

[...]

PR #1924, opened 2018-02-22
Defines http.route key and ensures server span name is preferred

codefromthecrypt:
Especially when using http route based span names, we have to be careful
to not pick a bad name. In shared spans, for example, the UI prefers
server-side's name of the service. However, we have no name priority
logic, so it is arbitrary. This ensures the server's span name wins in a
race.

See https://github.com/openzipkin/zipkin/blob/9bf49aa7f77ea3eea894626439a38465786e9d90/zipkin-ui/js/component_ui/traceSummary.js#L64
See https://github.com/openzipkin/zipkin/issues/1874


codefromthecrypt:
Verified this fixes name priority using brave-webmvc-example (with route-based span names)

screen shot 2018-02-22 at 7 06 34 pm

PR #1934, opened 2018-03-01
Added ability to run zipkin-server as an executable jar, verified in …

zeagord:
…local and PCF

### DIFF
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -291,6 +291,7 @@
         <artifactId>spring-boot-maven-plugin</artifactId>
         <configuration>
           <classifier>exec</classifier>
+          <executable>true</executable>
         </configuration>
       </plugin>
       <!-- disable retrolambda since this module uses java 8 types -->


codefromthecrypt:
should we change the README, too? or leave that out cc @abesto 

Ex the quick start "could" strip the jar name and let people just ./zipkin

https://github.com/openzipkin/zipkin#quick-start

codefromthecrypt:
if we did something like this, we would have to think also about the extension approach (ex adding SQS via a module). in these scenarios, java -jar is still the right choice

codefromthecrypt:
for the readme we could add text below like:

If running as a service you can also leave out the "java -jar " part. Ex.
```bash
$ ./zipkin.jar
```

zeagord:
I think we should change the read me too. It will be easier for readers.

shakuzen:
Mentioning the option in the readme sounds reasonable, but it isn't usable for all Windows users for instance. It also could just generally be unfamiliar for Java developers used to working with normal JARs. I think the default advice remaining using `java -jar` is better. But mentioning that it can be registered as a service and linking to the corresponding Spring Boot documentation seems like a good idea to me.

codefromthecrypt:
ok maybe add a section to the zipkin-server/README and leave the root one
alone then? ex a section on running as a service

abesto:
My two cents: I like how explicit `java -jar` is. If the target audience is both Java experts and folks who want Zipkin to "just work", then I'd mention this as an alternative for people who know what this change does (personally, I have no idea :D )

eirslett:
Why would anyone want this? To be able to type 10 characters less, `java -jar `?

zeagord:
Running as a system service gives better control in production environments. If it is an executable jar or service it is straight forward to create to symlink and manage it. Secondly our Zipkin server is a springboot app  and springboot supports executable jar.

zeagord:
https://docs.spring.io/spring-boot/docs/current/reference/html/deployment-install.html

codefromthecrypt:
OK I think it is best to not document this :) Not documenting it won't confuse people who already work with this as a jar, yet still allows folks to do "pretend I'm a binary" mode if they want.

We can always add docs later if we get questions about it

PR #1945, opened 2018-03-11
Accomodates server spans who weren't propagated their parent

codefromthecrypt:
See #1745


sokac:
Thanks for a quick fix. lgtm

codefromthecrypt:
going out as 2.5.2

PR #1946, opened 2018-03-11
Injects configurable <base> tag to support reverse proxies

### LINKED ISSUE #1930: UI: Inject <base> into HTML configured via ENV var
As discussed in https://github.com/openzipkin/zipkin/issues/1731.

In situations where we run behind a reverse proxy with limited configurability (like the `nginx` ingress of Kubernetes), we can't rely on the reverse proxy injecting the required `<base>` tag to make the UI work on whatever URL path prefix it's assigned. The cleanest solution would be adding support to the Spring application serving the UI for adding the tag based on the presence and value of an ENV var.

I propose we call that ENV var `ZIPKIN_UI_BASEPATH`. Via @wdittmer, other applications all have custom names for it, there doesn't appear to be a standard:

> For example, for Prometheus we need to configure:
> - '--web.external-url=http://management.test.cbo:13370/prometheus'
> - '--web.route-prefix=/'

> For Grafana:
> - name: GF_SERVER_ROOT_URL
>   value: /grafana

> For Kibana:
> - name: SERVER_BASEPATH
>    value: "/kibana"

abesto:
Previously: #1930 

Unlike agreed there, this PR does _not_ change the ENV var (`proxy`) used in development mode, because based on the documentation, that seems to be a different animal: documentation for `proxy` says that `http(s)` and the port are required, while here we're preferably using relative paths. I'll be happy if someone who understands exactly what that ENV var does says "nope, that doc string is misleading, they do the same thing".

Full disclosure: I got `org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.7.0:testCompile (default-testCompile) on project zipkin: Compilation failure` when trying to compile the whole project for an end-to-end test, so I didn't do that. This is most likely a problem in my env, but I'm hoping I can get away with not tracking it down (at least not today). The newly introduced unit tests _do_ pass.

Fixes #1930

### DIFF
--- a/zipkin-autoconfigure/ui/pom.xml
+++ b/zipkin-autoconfigure/ui/pom.xml
@@ -52,5 +52,11 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>zipkin-ui</artifactId>
     </dependency>
+    <dependency>
+      <!-- HTML library for injecting configurable <base> tag -->
+      <groupId>org.jsoup</groupId>
+      <artifactId>jsoup</artifactId>
+      <version>1.11.2</version>
+    </dependency>
   </dependencies>
 </project>
--- a/zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiAutoConfiguration.java
+++ b/zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiAutoConfiguration.java
@@ -27,6 +27,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Lazy;
 import org.springframework.core.Ordered;
 import org.springframework.core.annotation.Order;
 import org.springframework.core.io.Resource;
@@ -80,17 +81,25 @@ public class ZipkinUiAutoConfiguration extends WebMvcConfigurerAdapter {
   Resource indexHtml;
 
   @Bean
+  @Lazy
   String processedIndexHtml() throws IOException {
-    InputStream is = indexHtml.getInputStream();
-    Document soup = Jsoup.parse(is, null, ui.getBasePath());
-    is.close();
-    if (soup.head().getElementsByTag("base").isEmpty()) {
-      soup.head().appendChild(
-        soup.createElement("base")
-      );
+    InputStream is = null;
+    try {
+      is = indexHtml.getInputStream();
+      Document soup = Jsoup.parse(is, null, ui.getBasePath());
+      is.close();
+      if (soup.head().getElementsByTag("base").isEmpty()) {
+        soup.head().appendChild(
+          soup.createElement("base")
+        );
+      }
+      soup.head().getElementsByTag("base").html(ui.getBasePath());
+      return soup.html();
+    } finally {
+      if (is != null) {
+        is.close();
+      }
     }
-    soup.head().getElementsByTag("base").html(ui.getBasePath());
-    return soup.html();
   }
 
   @Override
--- a/zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiAutoConfiguration.java
+++ b/zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiAutoConfiguration.java
@@ -93,7 +93,7 @@ public class ZipkinUiAutoConfiguration extends WebMvcConfigurerAdapter {
         soup.createElement("base")
       );
     }
-    soup.head().getElementsByTag("base").html(baseTagValue);
+    soup.head().getElementsByTag("base").attr("href", baseTagValue);
     return soup.html();
   }
 
--- a/zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiAutoConfiguration.java
+++ b/zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiAutoConfiguration.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2017 The OpenZipkin Authors
+ * Copyright 2015-2018 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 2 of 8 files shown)


zipkin-autoconfigure/ui/pom.xml
```diff
@@ -52,5 +52,11 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>zipkin-ui</artifactId>
     </dependency>
+    <dependency>
+      <!-- HTML library for injecting configurable <base> tag -->
+      <groupId>org.jsoup</groupId>
+      <artifactId>jsoup</artifactId>
+      <version>1.11.2</version>
```
codefromthecrypt, line 59: license is MIT (ok) and afaict no deps (good!)

zipkin-autoconfigure/ui/src/test/resources/zipkin-ui/index.html
```diff
@@ -0,0 +1,4 @@
+<html>
```
codefromthecrypt, line 1: clever

codefromthecrypt:
@wdittmer are you in a position to try this without building it first, or should we merge so you can test a snapshot? or do you need a release to try it?

codefromthecrypt:
cc @openzipkin/devops-tooling pretty tighter change than I was expecting to resolve the base path mapping concern!

zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiAutoConfiguration.java
```diff
@@ -75,6 +79,20 @@
   @Value("classpath:zipkin-ui/index.html")
   Resource indexHtml;
 
+  @Bean
+  String processedIndexHtml() throws IOException {
+    InputStream is = indexHtml.getInputStream();
```
jplock: Should this be in a try block to ensure the stream is always closed even if there’s a parsing exception?

wdittmer:
I can only use official containers at the moment (e.g. openzipkin/zipkin:2.4.2) to test in the testing/staging environments. I could try locally with the nginx test project as I described in the other ticket, but have to postpone to after our internal release date (~next week).
Thanks for picking it up so quick though, much appreciated!

codefromthecrypt:
We will get this merged and cut this week if it helps your release. If
somehow it doesnt work correctly we can revise

wdittmer:
If there is a new container image I can pull it in the testing environment easily and we know if it works quite quickly and provide feedback. Don't worry about the release, we can roll-out a patch release with the new Zipkin afterwards. For me it is more important that Zipkin doesn't break for others with this change :)

codefromthecrypt:
I've a slight revision to this locally which only uses the new logic if the property is set. testing and will p-p-push it

zipkin-autoconfigure/ui/src/main/java/zipkin/autoconfigure/ui/ZipkinUiAutoConfiguration.java
abesto: In terms of Java best practices, yes, for sure. In terms of failing gracefully, I'm not sure what the expected behavior is for Spring applications: if we fail to parse the HTML, that most likely means the HTML is invalid. Should we then:

 * Blow up, fail early and fail hard?
 * Serve the HTML anyway, without modification (which we _know_ will fail behind a reverse proxy with the `base-path` config set?)
 * Serve the HTML anyway, but only if `base-path` is not set. If it's set, blow up?
 * Serve a custom error page that describes what happened, maybe even includes the exception?

codefromthecrypt:
my take is to blow up and not do anything differently than spring boot
would. Mainly using the try with resources thing removes the distraction of
a dangling resource.. so worth doing.

codefromthecrypt:
BASEPATH will cause less config glitches I think, vs BASE_PATH, so taking a stand to match on that. Reason is that words between the dots when mapping env variables can be a bit crap (and since BASEPATH is used in Kibana... hey their fault)

abesto:
Sounds reasonable. I don't have any counterarguments.

codefromthecrypt:
ok checking local thx!

codefromthecrypt:
hmm props look good:

```bash
$ curl -s localhost:9411/configprops|jq '."zipkin.ui-zipkin.autoconfigure.ui.ZipkinUiProperties"'
{
  "prefix": "zipkin.ui",
  "properties": {
    "defaultLookback": 3600000,
    "environment": "",
    "instrumented": ".*",
    "dependency": {
      "lowErrorRate": 0.5,
      "highErrorRate": 0.75
    },
    "basepath": "/",
    "queryLimit": 10,
    "logsUrl": null,
    "searchEnabled": true
  }
}
```

codefromthecrypt:
somehow I was expecting to see "/" in the base tag..
```bash
$ curl -s localhost:9411/zipkin/
<!doctype html>
<html>
 <head> 
  <!--
      add 'base' tag to work around the fact that 'html-webpack-plugin' does not work
      with '__webpack_public_path__' being set as reported at https://github.com/jantimon/html-webpack-plugin/issues/119
    --> 
  <base href="/zipkin/">/</base> 
--snip--
```

abesto:
Oh fuck. Because I'm a stupid h00man, and put the link not in the `href` attribute, but in the html content of the tag.

abesto:
And even wrote nice unit tests that verify this totally borken behavior.

codefromthecrypt:
LOL do you want to sort it or should I? my IDE is open so no trouble. your call

abesto:
Please go ahead and take care of it then. I've since switched context to work stuff.

codefromthecrypt:
roger wilco, Roger Wilco

codefromthecrypt:
ok it works.. but I should try it through a proxy somehow

wdittmer:
If you have zipkin running locally you could try:
Dockerfile:

>FROM nginx:latest
>
>COPY nginx.conf /etc/nginx/nginx.conf


nginx.conf:
------------
worker_processes 1;

events { worker_connections 1024; }


http {

    log_format compression '$remote_addr - $remote_user [$time_local] '
        '"$request" $status $upstream_addr '
        '"$http_referer" "$http_user_agent" "$gzip_ratio"';

    server {
        listen 8080;
        server_name reverse-proxy;
        access_log /var/log/nginx/access.log compression;

        location /zipkin {
            proxy_pass          http://docker.for.mac.localhost:9411;
            proxy_set_header    Host             $host;
            proxy_set_header    X-Real-IP        $remote_addr;
            proxy_set_header    X-Forwarded-For  $proxy_add_x_forwarded_for;
            proxy_read_timeout 1800;
            proxy_connect_timeout 1800;
            rewrite /zipkin/(.*) /$1 break;
            rewrite /zipkin / break;
        }
    }
}
---------------
And in /etc/hosts:

------------
127.0.0.1 reverse-proxy
127.0.0.1 zipkin-server
-------------

codefromthecrypt:
will do a "dark release" of this as a patch as I'm confident in how it is coded it won't break anyone. If it works, the next minor we'll announce how do  ZIPKIN_UI_BASEPATH=/  or ZIPKIN_UI_BASEPATH=/foo/zipkin

so in other words 2.5.3 after builds finish etc
@wdittmer

wdittmer:
(man I suck at getting it to format straight...)
I will await 2.5.3 :)

PR #1949, opened 2018-03-16
Prefers empty array to undefined in UI span conversion; backfills ipv6

codefromthecrypt:
When writing json to the wire, we guard on creating empty arrays for
annotations and binaryAnnotations. However, in process, this just makes
life more difficult ex `(annotations || []).map...` This simplifies
things, cleaning up some code. It also backfills the missing ipv6.


codefromthecrypt:
cc @Logic-32 @zeagord

PR #1961, opened 2018-03-20
Makes it possible to implement a protobuf3 span encoder

### LINKED ISSUE #1942: Add proto3 encoding
We should eventually map out a proto3 encoding even if this is just a direct translation from the json. We don't yet have an issue for this, but it is in response to #1644 and its parent.

In the mean time, we should at least make it possible in the java library to do this. Minimally, this requires adding zipkin2.codec.Encoding.PROTO3 to permit this format when using the async reporter.

codefromthecrypt:
While we haven't yet defined a proto3 format, this allows one to be
defined. This also allows us to refine how stackdriver (or any future
translated format) works in zipkin reporter.

Implementation is relatively simple: this assumes a buffer of spans is
analogous to a proto3 repeated field, and that the field name itself is
ordinal <=127.

Fixes #1942

### DIFF
--- a/zipkin2/src/main/java/zipkin2/codec/Encoding.java
+++ b/zipkin2/src/main/java/zipkin2/codec/Encoding.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2017 The OpenZipkin Authors
+ * Copyright 2015-2018 The OpenZipkin Authors
  *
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
verified through https://github.com/openzipkin/zipkin-gcp/pull/62

releasing as a patch as it isn't really a real feature, yet

Issue #1962, opened 2018-03-21
upgrade to Spring Boot 2.0 NoClassDefFoundError UndertowEmbeddedServletContainerFactory


JrCx7scC:
UndertowEmbeddedServletContainerFactory has been renamed in Spring Boot 2.0. It's now UndertowServletWebServerFactory.

codefromthecrypt:
When we update to spring boot 2, it will be the only version we support. So
basically we just change everything to work with it.

Probably will do it soon. cc @shakuzen

codefromthecrypt:
@llinder @devinsba do you have any concerns with the next zipkin minor
moving to spring boot 2? (implies this also happens to zipkin-gcp -aws etc)

I know we are missing the layout factory change..

rahul-pharande:
@adriancole is there tentative date when we have spring boot 2.0 support available here.

codefromthecrypt:
only recently we are getting interest. there is a patch so I am personally
more comfortable. I would like an answer from the others before proceeding.


We still need https://github.com/openzipkin/zipkin-layout-factory

I'm curious.. since zipkin-server is packaged as an all-in-one exec jar,
and we don't support custom servers. Why do you care what version of spring
boot it uses underneath?

Is there some feature you are looking for?

JrCx7scC:
Because we recently used SpringCloud 2.0.0 and found this problem, there is a problem with the startup dependencies by Spring, or can there be other solutions?

JrCx7scC:
Thanks for your reply, let's try it without Spring Boot

bodiam:
Hi, just wanted to let you know I'm running into the same problem. All our applications are running on Spring Boot 2.0, which is why are running into this. I'll downgrade to 1.x for now.

JrCx7scC:
There are more bugs, waiting for the official version to be used again @bodiam

shakuzen:
> All our applications are running on Spring Boot 2.0, which is why are running into this. I'll downgrade to 1.x for now.

Everyone please note that your applications can run on Spring Boot 2 and still interact with the current version of the Zipkin server just fine.
We don't support custom builds of the Zipkin server, which is what I assume is being done here. It would be easiest and more supportable to use the Quick Start instructions for running the Zipkin Server: https://github.com/openzipkin/zipkin#quick-start
If there are reasons you need a custom build, please share them with us on Gitter https://gitter.im/openzipkin/zipkin.

bodiam:
[...]

PR #1968, opened 2018-03-26
Deprecates EnableZipkinServer to explain custom servers are unsupported

codefromthecrypt:
Especially lately, we have had a large number of people having problems
with unnecessarily custom servers. Some are due to not knowing Sleuth's
stream server is obviated by our rabbit support. Some are due to blogs
which unfortunately recommend starting Zipkin in the IDE via a custom
server.

Through discussion, we decided the easiest way to let users know custom
servers are unsupported is by deprecation. Deprecation shows up in the
IDE and will alert those doing blogs or otherwise that they are
suggesting discouraged practice. It also sends a clear signal to those
who need to make custom servers that while doing so is possible, it is
something the customizer needs to accept support responsibility of.

See https://github.com/openzipkin/zipkin/issues/1962#issuecomment-375872410


zipkin-server/src/main/java/zipkin/server/EnableZipkinServer.java
```diff
@@ -19,18 +19,22 @@
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import org.springframework.context.annotation.Import;
-import zipkin.server.brave.TracingConfiguration;
 
+/**
+ * @deprecated Custom servers are possible, but not supported by the community. Please use our
```
codefromthecrypt, line 24: here's the deprecation notice

shakuzen (approved):
:shipit: LGTM

codefromthecrypt:
thx for having a look, tommy

PR #1969, opened 2018-03-27
Corrects clock-skew in one-way RPC spans

codefromthecrypt:
In one-way RPC, we now correct obvious skew: when the server receives a
message before it was sent.


wdittmer (commented):
Only see a typo in the comments: "// the only way there is skey...."

codefromthecrypt:
@wdittmer thx good catch

PR #1971, opened 2018-03-27
Layout changes for the collector and storage using zipkin layout factory

zeagord:
Changed the the kafka10 collector build layout based on Zipkin-Layout-Factory

### DIFF
--- a/zipkin-autoconfigure/collector-kafka10/pom.xml
+++ b/zipkin-autoconfigure/collector-kafka10/pom.xml
@@ -43,22 +43,23 @@
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-maven-plugin</artifactId>
         <version>${spring-boot.version}</version>
-        <executions>
-          <execution>
-            <goals>
-              <goal>repackage</goal>
-            </goals>
-          </execution>
-        </executions>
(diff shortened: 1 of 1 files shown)


zipkin-autoconfigure/collector-kafka10/pom.xml
```diff
@@ -43,22 +43,23 @@
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-maven-plugin</artifactId>
         <version>${spring-boot.version}</version>
-        <executions>
-          <execution>
-            <goals>
-              <goal>repackage</goal>
```
codefromthecrypt, line 49: is repackage redundant? or is there another reason why this isn't needed anymore?

zeagord: Yeah it is redundant and IMHO repackage is not needed. If we need to we have to provide the launcher in the layout factory as optional. Is there a situation where we need to run this as a standalone jar?

codefromthecrypt: nope.. just asking!

codefromthecrypt:
thanks tons

Issue #1973, opened 2018-03-28
Skip filtering traces from result when STRICT_TRACE_ID is set to true


connectwithnara:
https://github.com/openzipkin/zipkin/blob/3f299cda195af7b1685a11e262dcd6a554dec2a1/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchSpanStore.java#L116

The check to remove traces from the result can be skipped if STRICT_TRACE_ID is enabled.

codefromthecrypt:
closed via #1974

codefromthecrypt:
We reverted this as we didn't have good test coverage. Basically the strict trace ID setting still includes the ability to read data prior to when the switch happened. That is `getTrace_retrievesBy128BitTraceId_afterSwitch` test.

This makes the issue more difficult because on one hand, if data is uniform, we should totally never even try to do mixed stuff. OTOH, we do want people to use strict ID, so we have to facilitate migrations.

I think we can handle this a little differently by looking at the length of the trace IDs inside the search results.

One approach is: look to see if any trace IDs clash on lower 64bits. If there is no clash, don't filter.

PR #1974, opened 2018-03-28
Skip filtering traces from result when STRICT_TRACE_ID is set to true…

connectwithnara:
… (#1973)


codefromthecrypt:
nice catch

PR #1976, opened 2018-03-28
Avoids more types absent in boot 2

codefromthecrypt:


zipkin-autoconfigure/ui/src/test/java/zipkin/autoconfigure/ui/ITZipkinUiAutoConfiguration.java
```diff
@@ -32,7 +33,7 @@
 )
 public class ITZipkinUiAutoConfiguration {
 
-  @LocalServerPort int zipkinPort;
+  @Autowired @Value("${local.server.port}") int zipkinPort;
```
shakuzen, line 36: I think it should just be `@Value("${local.server.port}")` without the `@Autowired`

PR #1978, opened 2018-03-28
Removes scribe from the default jar, as an optional module

codefromthecrypt:
This removes scribe and the 2 year abandoned library it uses from the
default server jar. A follow-up change will integrate it conditionally
in the docker image similar to how we integrate kafka10.

See #1192
See https://github.com/openzipkin/zipkin-reporter-java/issues/95


shakuzen (approved):

codefromthecrypt:
https://github.com/openzipkin/docker-zipkin/pull/173 adjusts the docker image

Issue #1979, opened 2018-03-30
Index chars limit used in ES SpanConsumer is inconsistent in docs and code


connectwithnara:
The limit checked is 255 in the [code](https://github.com/openzipkin/zipkin/blob/fdd5c4b6344012554756daafd3841c896d1a1ec7/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchSpanConsumer.java#L122) and 256 in the [docs](https://github.com/openzipkin/zipkin/tree/master/zipkin-storage/elasticsearch).

codefromthecrypt:
closed by #1980

PR #1980, opened 2018-03-30
Index chars limit used in ES SpanConsumer is inconsistent in docs and…

connectwithnara:
… code (#1979)

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchSpanConsumer.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchSpanConsumer.java
@@ -34,6 +34,8 @@ import zipkin2.storage.SpanConsumer;
 class ElasticsearchSpanConsumer implements SpanConsumer { // not final for testing
   static final Logger LOG = Logger.getLogger(ElasticsearchSpanConsumer.class.getName());
 
+  private static final int INDEX_CHARS_LIMIT = 256;
+
   final ElasticsearchStorage es;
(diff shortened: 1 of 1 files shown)


zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchSpanConsumer.java
```diff
@@ -119,13 +121,22 @@ void add(long indexTimestamp, Span span, long timestampMillis) {
...
         for (Map.Entry<String, String> tag : span.tags().entrySet()) {
-          if (tag.getKey().length() + tag.getValue().length() + 1 > 255) continue;
+          int length = tag.getKey().length() + tag.getValue().length() + 1;
+          if (length > INDEX_CHARS_LIMIT) {
+            continue;
+          }
           writer.value(tag.getKey()); // search is possible by key alone
-          writer.value(tag.getKey() + "=" + tag.getValue());
+          writer.value(
+            new StringBuilder(length)
```
codefromthecrypt, line 136: I'm pretty sure javac does this for us :P

codefromthecrypt: oic yeah suppose it wouldn't pre-define the length of the builder, though the length here is only correct if ascii (utf 8 will overrun this)

codefromthecrypt: though maybe ES length is chars anyway? regardless, that's a different issue.. I'm fine with this.

PR #1982, opened 2018-04-03
add note on security since it has come up several times on gitter

jorgheymans:
brief note basically saying that zipkin-ui does not handle security itself.

### DIFF
--- a/zipkin-ui/README.md
+++ b/zipkin-ui/README.md
@@ -155,3 +155,7 @@ And then it's observable in the UI:
 ```bash
 $ open http://localhost/proxy/foo/bar/zipkin/?serviceName=zipkin-server&startTs=1378193040000&endTs=1505463856013
 ```
+### How do I configure security (authentication, authorization)? 
+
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
sounds good, but should this note be in the server readme, here or both?

jorgheymans:
Good point, maybe we should also point out then that there are different
entry points to consider (browser URL vs rest api). Most people don't
distinguish between them initially. But layer they might want to have
different security rules for these. I'll revise and update. Maybe it will
be more clear where to put it as well then.

Op wo 4 apr. 2018 01:56 schreef Adrian Cole <notifications@github.com>:

jorgheymans:
added note on the endpoints.

as to how to secure data ingestion via the api that should be put in zipkin-server and somewhere in brave because it affects them both, should be done in a different PR.

codefromthecrypt:
Might be useful to link to our httpd doc somehow as proxying the UI is a
bit tricky
https://github.com/openzipkin/zipkin/tree/master/zipkin-ui#apache-http-as-a-zipkin-reverse-proxy

jorgheymans:
done.

codefromthecrypt:
thanks so much!

PR #1989, opened 2018-04-10
[WIP] Metrics test

zeagord:


zeagord:
Migrated to springboot2 and verified health endpoints. 
TODO: Implement the collector metrics via Micrometer

zeagord:
Added the Collector Counter using Micrometer and delegated the **HealthEnpoint** to the root  path`/health`

zeagord:
Polished the metrics collection and health enpoint. 
Sample output from `/metrics`
```
{
  "counter.zipkin_collector.messages.http": 1.0,
  "counter.zipkin_collector.spans": 0.0,
  "counter.zipkin_collector.spans_dropped": 0.0,
  "counter.zipkin_collector.spans.http": 0.0,
  "counter.zipkin_collector.spans_dropped.http": 0.0,
  "counter.zipkin_collector.bytes": 0.0,
  "gauge.zipkin_collector.message_bytes": 0.0,
  "gauge.zipkin_collector.message_spans.http": 0.0,
  "counter.zipkin_collector.messages_dropped": 0.0,
  "counter.zipkin_collector.messages": 0.0,
  "gauge.zipkin_collector.message_bytes.http": 0.0,
  "counter.zipkin_collector.bytes.http": 0.0,
  "gauge.zipkin_collector.message_spans": 0.0,
  "counter.zipkin_collector.messages_dropped.http": 0.0
}
```
And from the `/health`
```
{
  "status": "UP",
  "details": {
    "zipkin": {
      "status": "UP",
      "details": {
        "InMemoryStorage": {
          "status": "UP"
        }
      }
    },
    "diskSpace": {
      "status": "UP",
      "details": {
        "total": 121123069952,
        "free": 10457477120,
        "threshold": 10485760
      }
    }
  }
}
```

zeagord:
Implemented all the meters, polished tests.

zipkin-server/src/main/java/zipkin/server/internal/MetricsHealthController.java
```diff
@@ -0,0 +1,63 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.MeterRegistry;
+import java.util.ArrayList;
+import java.util.List;
+import org.omg.CORBA.PRIVATE_MEMBER;
```
shakuzen, line 21: OMG CORBA 😱

zipkin-server/src/main/java/zipkin/server/internal/MetricsHealthController.java
```diff
@@ -0,0 +1,63 @@
...
+      }
+      if (meter.getId().getName().contains("zipkin") && meter.getId().getName().contains("gauge")){
+        node.put(meter.getId().getName(), meterRegistry.get(meter.getId().getName()).gauge().value());
+      }
+    }
+    return node;
+  }
+
+  // Delegates the health endpoint from the Actuator to the root context path
+  @ReadOperation
```
shakuzen, line 58: `@ReadOperation` is for an Actuator endpoint, but this class is just a regular `RestController`. I think this annotation isn't having any effect.

zipkin-server/src/main/java/zipkin/server/internal/MetricsHealthController.java
```diff
@@ -0,0 +1,63 @@
...
+  @GetMapping("/health")
+  public ResponseEntity<Health> getHealth(){
+    return ResponseEntity.ok(healthEndpointDelegate.health());
```
shakuzen, line 61: In Spring Boot 2, the format of the `/health` endpoint changed slightly. That is, a `$.details` element was added, under which all of the health indicators' details are given. Before Spring Boot 2, each detail was at the root level.
We'll need to account for this difference if we want to remain backwards compatible with all parts of the `/health` response body.

Sample Spring Boot 1 `GET /health`
```json
{
    "diskSpace": {
        "free": 804290560,
        "status": "UP",
        "threshold": 10485760,
        "total": 1056858112
    },
    "status": "UP",
    "zipkin": {
        "ElasticsearchHttpStorage": {
            "status": "UP"
        },
        "status": "UP"
    }
}
```

zipkin-server/src/main/java/zipkin/server/internal/ActuateCollectorMetrics.java
```diff
@@ -11,17 +13,11 @@
...
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Date;
-import java.util.List;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.LongAdder;
-import java.util.function.Supplier;
-import org.springframework.boot.actuate.endpoint.PublicMetrics;
-import org.springframework.boot.actuate.metrics.Metric;
+import com.google.common.util.concurrent.AtomicDouble;
```
shakuzen, line 17: Do we need to use `AtomicDouble`? Avoiding Guava would seem preferable.
Do we need strict atomicity for these gauges? Maybe a `Double` is sufficient.

shakuzen (commented):
Thank you for working on this. It's looking pretty good so far. I left a few comments.

zipkin-server/src/main/java/zipkin/server/internal/ActuateCollectorMetrics.java
zeagord: Thanks @shakuzen for the review. I think Gauge can observe values that are settable so I was using AtomicDouble. 
http://micrometer.io/docs/concepts#_gauges
Please correct If I am wrong.

zeagord:
Closing this PR to breakdown the tests and the Micrometer/Boot2 upgrade in to separate ones.

zipkin-server/src/main/java/zipkin/server/internal/ActuateCollectorMetrics.java
shakuzen: You are correct. We need a strong reference to some object that can produce the double value when the registry is queried. It doesn't need to be an `AtomicDouble`, though, I believe. Zipkin Server doesn't have a direct dependency on Guava and there are version considerations due to transitive dependencies on Guava (which are hopefully going away in the future). My comment was mostly to point out I think we would like to avoid a direct dependency on Guava.

You are correct it needs to be settable per the warning in that section of the documentation, so my suggestion of `Double` (which is immutable) is indeed no good.

PR #1994, opened 2018-04-16
Adds direct transcoding of UTF-8 strings via code most similar to Netty

codefromthecrypt:
Before, we kicked out to `String.getBytes(UTF_8)` when encoding found
characters outside the normal ASCII range. This redundantly allocates
arrays and is generally less efficient than directly encoding.

This directly encodes UTF-8 using code most similar to Netty

https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java#L518

Thanks @wu-sheng for the sample data used in benchmarks

```
Benchmark                                      Mode  Cnt  Score   Error  Units
BEFORE:
CodecBenchmarks.writeChineseSpan_json_zipkin2  avgt   15  1.103 ± 0.006  us/op
AFTER:
CodecBenchmarks.writeChineseSpan_json_zipkin2  avgt   15  0.953 ± 0.008  us/op

Benchmark                                 Mode  Cnt  Score    Error  Units
BEFORE:
BufferBenchmarks.writeUtf8_chinese        avgt   15  0.098 ±  0.005  us/op
AFTER:
BufferBenchmarks.writeUtf8_chinese        avgt   15  0.061 ±  0.001  us/op
```

### DIFF
--- a/benchmarks/src/main/java/zipkin/benchmarks/CodecBenchmarks.java
+++ b/benchmarks/src/main/java/zipkin/benchmarks/CodecBenchmarks.java
@@ -218,6 +218,19 @@ public class CodecBenchmarks {
     return SpanBytesEncoder.JSON_V1.encodeList(tenSpan2s);
   }
 
+  static final byte[] zipkin2JsonChinese = read("/zipkin2-chinese.json");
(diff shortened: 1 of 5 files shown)


wu-sheng:
Glad to help. :)

PR #1997, opened 2018-04-16
Adds Endpoint IP address bytes accessors

codefromthecrypt:
This adds `Endpoint.ipv4Bytes()` and `Endpoint.ipv6Bytes()` needed to
prevent excess overhead marshaling into byte arrays. It is efficient to
do this once, as endpoint IPs are most often shared (local endpoint) for
all spans. Even in the case of remote endpoint, these are often parsed
from `Inet6Address` objects who already have a byte array reference.

This will be used in proto3 and possibly in a future retrofitted v1
thrift encoder.

See #1998

### DIFF
--- a/zipkin2/src/main/java/zipkin2/Endpoint.java
+++ b/zipkin2/src/main/java/zipkin2/Endpoint.java
@@ -48,13 +48,34 @@ public final class Endpoint implements Serializable { // for Spark and Flink job
   }
 
   /**
+   * IPv4 endpoint address packed into 4 bytes or null if unknown.
+   *
+   * @see #ipv6()
+   * @see java.net.Inet4Address#getAddress()
+   */
+  @Nullable public byte[] ipv4Bytes() {
+    return ipv4Bytes;
+  }
+
+  /**
    * The text representation of the primary IPv6 address associated with this a connection. Ex.
    * 2001:db8::c001 Absent if unknown.
    *
-   * <p>Prefer using the {@link #ipv4()} field for mapped addresses.
+   * @see #ipv4() for mapped addresses
+   * @see #ipv6Bytes()
    */
   @Nullable public String ipv6() {
     return ipv6;
+  }
+
+  /**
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
Here are the benchmarks. This time will happen once for all endpoints built from the same IP.

Ex. `endpoint.toBuilder().serviceName("foo").build()` does not re-parse

```
Benchmark                             Mode  Cnt  Score   Error  Units
EndpointBenchmarks.parseIpv4_addr     avgt   15  0.055 ± 0.003  us/op
EndpointBenchmarks.parseIpv4_literal  avgt   15  0.039 ± 0.001  us/op
EndpointBenchmarks.parseIpv6_addr     avgt   15  0.047 ± 0.002  us/op
EndpointBenchmarks.parseIpv6_literal  avgt   15  0.264 ± 0.003  us/op
```

zipkin2/src/main/java/zipkin2/Endpoint.java
```diff
@@ -515,4 +552,30 @@ Object readResolve() throws ObjectStreamException {
       }
     }
   }
+
+  static byte[] getIpv4Bytes(String ipv4) {
```
codefromthecrypt, line 556: the added tests 100% cover this new code

codefromthecrypt:
In normal proto3 generated code, this allows conversion like so:

```java
Endpoint toProto3(zipkin2.Endpoint input) {
  Endpoint.Builder result = Endpoint.newBuilder();
  if (input.ipv6Bytes() != null) {
    result.setIpv6(ByteString.copyFrom(input.ipv6Bytes()));
  }
  // continue..
  return result;
}
```

of course our encoder will be a lot more efficient than this.

codefromthecrypt:
might expose this functionality differently.. will let it sit until the rest of proto3 is done

codefromthecrypt:
this api proved to be the most convenient

PR #1999, opened 2018-04-17
Implements proto3 encoding

codefromthecrypt:
This is the first part of proto3 implementation, raised separately to
keep things readable.

See #1998

Fun fact. Our `TestObjects.CLIENT_SPAN`
 * in json v1 is 798 bytes
 * in json v2 is 438 bytes
 * in proto3 is 163 bytes

### DIFF
--- a/zipkin2/src/main/java/zipkin2/codec/Encoding.java
+++ b/zipkin2/src/main/java/zipkin2/codec/Encoding.java
@@ -14,7 +14,6 @@
 package zipkin2.codec;
 
 import java.util.List;
-import zipkin2.internal.Buffer;
 
 // ZIPKIN3 make this not an enum as it prevents non-standard encoding, for example reporting to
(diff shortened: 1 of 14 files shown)


codefromthecrypt:
interop tests for this are in #1998 I will backfill more unit tests tomorrow

codefromthecrypt:
going to revise the encoding approach as it isn't as quick as it could be due to multiple length calculations implicit in the format

codefromthecrypt:
actually moving speed even faster isn't urgent. will just complete the encoder

codefromthecrypt:
all encoding features are now in, but there is some tests left to port from the other pull request which is compat testing against protoc.

codefromthecrypt:
ok test coverage is now ... covered

codefromthecrypt:
tomorrow I'll add decoding as a separate PR

PR #2010, opened 2018-04-20
Upgrades to Springboot2 and Micrometer

zeagord:
... And keeps metrics, health endpoints backward compatible


zipkin-server/src/main/java/zipkin/server/internal/ZipkinQueryApiV1.java
```diff
@@ -13,6 +13,7 @@
  */
 package zipkin.server.internal;
 
+import io.micrometer.core.annotation.Timed;
```
codefromthecrypt: heh :P

zipkin-server/src/main/java/zipkin/server/internal/MetricsHealthController.java
```diff
@@ -43,4 +59,31 @@ public ObjectNode fetchMetricsFromMicrometer(){
...
+    Map health = new HashMap();
+    health.put("status", healthEndpointDelegate.health().getStatus().getCode());
+    health.put("zipkin", healthEndpointDelegate.health().getDetails().get("zipkin"));
+    return health;
+  }
+
+  // Prometheus scrape code ported from springboot actuator and can be deprecated later
+  // through root context path for backward compatibility, Mimics the same code from
+  // org.springframework.boot.actuate.metrics.export.prometheus.PrometheusScrapeEndpoint
+  @GetMapping("/prometheus")
```
shakuzen: If this endpoint does the same thing as the Actuator version, we could just redirect or forward that route. This is what the Spring Initializr did to maintain backwards compatibility with its `/info` endpoint: https://github.com/spring-io/initializr/blob/540918567b60da18341883750a5a327f4456fd84/initializr-web/src/main/java/io/spring/initializr/web/autoconfigure/InitializrWebConfig.java#L47

zeagord: Makes sense. I will change it. But i makes the tests fail :) `writeSpans_updatesPrometheusMetrics`

codefromthecrypt:
Thanks for the hard work and review folks. Feel free to make any later polishing post merge

codefromthecrypt:
ps added this to sort the test that failed 9d1c94b9ded7a316140d694866eda25ec98bee97

PR #2013, opened 2018-04-22
Clarifies Span text for the purpose of synchronizing code with api

codefromthecrypt:
The zipkin-api project has no text to define what a span is. This
clarifies text so we can use it consistently.

See https://github.com/openzipkin/zipkin-api/pull/47


codefromthecrypt:
cc @openzipkin/core

PR #2021, opened 2018-04-26
 Implements Proto3 decoding and collector detection

codefromthecrypt:
This allows collectors to detect if proto3 is in use. When using http,
the media type "application/x-protobuf" can also be used.

See #1998


codefromthecrypt:
I am working through changes for the collector now, to detect proto3 so things like kafka etc will work

PR #2024, opened 2018-04-27
Fixes naming convention and removes empty prometheus metrics and fixes skips

codefromthecrypt:
When we switched to micrometer's prometheus in 2.7.3, we accidentally
broke the naming convention used before.

Notably, in prometheus.yml, we had this:

```yaml
      - source_labels: [__name__]
        regex: '(?:gauge|counter)_zipkin_collector_(.*)_([^_]*)'
        replacement: '${2}'
        target_label: transport

```

As the default naming convention adds "_total", this causes dashboards
to end up with a transport literally named "_total".

This also removes empty metrics eagerly initialized and patterns that should have been skipped.

### DIFF
--- a/zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin/autoconfigure/prometheus/ZipkinPrometheusMetricsAutoConfiguration.java
+++ b/zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin/autoconfigure/prometheus/ZipkinPrometheusMetricsAutoConfiguration.java
@@ -18,6 +18,7 @@ import io.micrometer.core.instrument.Clock;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.Timer;
+import io.micrometer.prometheus.PrometheusMeterRegistry;
 import io.undertow.server.HandlerWrapper;
 import io.undertow.server.HttpHandler;
 import io.undertow.server.HttpServerExchange;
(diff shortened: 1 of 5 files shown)


zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin/autoconfigure/prometheus/ZipkinPrometheusMetricsAutoConfiguration.java
```diff
@@ -18,6 +18,7 @@
 import io.micrometer.core.instrument.MeterRegistry;
```
zeagord, line 18: We may not need this since we use the `PrometheusMeterRegistry`

zeagord (approved):

zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin/autoconfigure/prometheus/ZipkinPrometheusMetricsAutoConfiguration.java
codefromthecrypt: HttpRequestDurationHandler is still referencing it.. maybe later, we will move that one to the zipkin-server, but I failed to get it working :P

codefromthecrypt:
reverted the naming convention as too many other things broke. going with the document approach:

For your prometheus, moving forward make sure your config has a _total on the end for the counters.
```diff
       # Received message count
       - source_labels: [__name__]
-        regex: '(?:gauge|counter)_zipkin_collector_(.*)_([^_]*)'
+        regex: '(?:gauge|counter)_zipkin_collector_(.*)_([^_]*)_total'
         replacement: '${2}'
         target_label: transport
       - source_labels: [__name__]
-        regex: '(?:gauge|counter)_zipkin_collector_(.*)_([^_]*)'
+        regex: '(?:gauge|counter)_zipkin_collector_(.*)_([^_]*)_total'
         replacement: 'zipkin_collector_${1}'
         target_label: __name__
```

Also, replace `http_requests_total` with `http_request_duration_seconds_count`

PR #2025, opened 2018-04-27
Reverts old naming convention as so many metrics changed

codefromthecrypt:
This reverts part of b5402bf481866463fead10764b84e323176cd6bb

Basically almost all metrics changed in the last version. We'll announce
and take a one-time hit to ensure the metrics used are standard moving
forward.


codefromthecrypt:
this is the ultimate fix https://github.com/openzipkin/zipkin/pull/2026

PR #2026, opened 2018-04-28
Changes prometheus metrics to use tags instead of name encoding

codefromthecrypt:
Before, prometheus metrics needed parsing to determine the transport.
This leaves "/metrics" alone, but maps the collector metrics to look
normal, using tags instead.


zeagord (commented):
Looks good

PR #2030, opened 2018-04-28
Support for setting  the json field names (filebeat).

WiFeng:
When the application architecture as follow:
`application -> filebeat -> kafka -> zipkin`
The kafka10 collector would print error  , because of it can't parse the packaged json string. 

The origin kafka topic data :
```
{"@timestamp":"2019-04-11T03:28:20.985Z","beat":{"hostname":"vm-10-112-33-39","name":"vm-10-112-33-39","version":"5.1.2"},"fields":{"topic":"lefan_tracing_log"},"input_type":"log","message":"[{\"traceId\":\"7e93c1680995fcdc\",\"name\":\"GET\",\"version\":\"php-4\",\"id\":\"79fcb72dfaf3a691\",\"timestamp\":1554953300670084,\"duration\":10517,\"annotations\":[{\"value\":\"sr\",\"timestamp\":1554953300670084,\"endpoint\":{\"serviceName\":\"odp\",\"ipv4\":\"10.112.33.39\",\"port\":80}},{\"value\":\"ss\",\"timestamp\":1554953300680601,\"endpoint\":{\"serviceName\":\"odp\",\"ipv4\":\"10.112.33.39\",\"port\":80}}],\"binaryAnnotations\":[{\"key\":\"http.url\",\"value\":\"http:\\/\\/odp.my.le.com\\/tv\\/growth?uid=238737344\u0026times=1554953300\u0026channel=0\u0026saltkey=7f76101bb7e0027bad315ee0b7f103c9\",\"endpoint\":{\"serviceName\":\"odp\",\"ipv4\":\"10.112.33.39\",\"port\":80}},{\"key\":\"path\",\"value\":\"\\/letv\\/www\\/odp.my.le.com\\/index.php\",\"endpoint\":{\"serviceName\":\"odp\",\"ipv4\":\"10.112.33.39\",\"port\":80}}]}]","offset":18446626,"source":"/letv/logs/phpzipkin/tracing-20190411.log","tags":["service-lefan","web-tier"],"type":"log"}
```
But zipkin-collector only need the `message` data，and can't handle other format.
```
[{"traceId":"b018df0c8d465265","name":"GET","version":"php-4","id":"8abd9e83ed3f9af5","timestamp":1524900779537352,"duration":12561,"annotations":[{"value":"sr","timestamp":1524900779537352,"endpoint":{"serviceName":"growth","ipv4":"10.112.156.88","port":80}},{"value":"ss","timestamp":1524900779549913,"endpoint":{"serviceName":"growth","ipv4":"10.112.156.88","port":80}}],"binaryAnnotations":[{"key":"http.url","value":"http:\/\/le.com\/growth\/v2\/values\/get?app=1&uid=125614239","endpoint":{"serviceName":"growth","ipv4":"10.112.156.88","port":80}},{"key":"path","value":"\/letv\/www\/usergrowth-lwf\/public\/index.php","endpoint":{"serviceName":"growth","ipv4":"10.112.156.88","port":80}}]}]
```
 So it does't work.

This version supports for setting the json filed name. Default value is `message`, it's for filebeat friendly. 
```
    java \
    -Dloader.path='kafka10.jar,kafka10.jar!/lib' \
    -Dspring.profiles.active=kafka \
    -Dzipkin.collector.kafka.bootstrap-servers=10.183.1.1:9092,10.183.1.2:9092,10.183.1.3:9092 \
    -Dzipkin.collector.kafka.topic=tracing_log \
    -Dzipkin.collector.kafka.field=message \
    -cp zipkin.jar \
    org.springframework.boot.loader.PropertiesLauncher \
    --logging.level.zipkin=DEBUG 2>&1 >>/tmp/zipkin2.log &
```

### DIFF
--- a/zipkin-autoconfigure/collector-kafka10/src/main/java/zipkin/autoconfigure/collector/kafka10/ZipkinKafkaCollectorProperties.java
+++ b/zipkin-autoconfigure/collector-kafka10/src/main/java/zipkin/autoconfigure/collector/kafka10/ZipkinKafkaCollectorProperties.java
@@ -26,6 +26,8 @@ class ZipkinKafkaCollectorProperties {
   private String groupId;
   /** Kafka topic span data will be retrieved from. */
   private String topic;
+  /** Kafka topic span data field (if it's json). */
+  private String field;
   /** Number of Kafka consumer threads to run. */
   private Integer streams;
   /** Additional Kafka consumer configuration. */
@@ -55,6 +57,14 @@ class ZipkinKafkaCollectorProperties {
     this.topic = emptyToNull(topic);
   }
 
+  public String getField() {
+    return field;
+  }
+
+  public void setField(String field) {
+    this.field = emptyToNull(field);
+  }
+
   public Integer getStreams() {
     return streams;
   }
@@ -76,6 +86,7 @@ class ZipkinKafkaCollectorProperties {
     if (bootstrapServers != null) result.bootstrapServers(bootstrapServers);
     if (groupId != null) result.groupId(groupId);
     if (topic != null) result.topic(topic);
+    if (field != null) result.field(field);
     if (streams != null) result.streams(streams);
     if (overrides != null) result.overrides(overrides);
     return result;
--- a/zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaCollector.java
+++ b/zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaCollector.java
@@ -58,6 +58,7 @@ public final class KafkaCollector implements CollectorComponent {
     Collector.Builder delegate = Collector.builder(KafkaCollector.class);
     CollectorMetrics metrics = CollectorMetrics.NOOP_METRICS;
     String topic = "zipkin";
+    String field = "message";
     int streams = 1;
 
     @Override public Builder storage(StorageComponent storage) {
@@ -82,6 +83,14 @@ public final class KafkaCollector implements CollectorComponent {
      */
     public Builder topic(String topic) {
       this.topic = checkNotNull(topic, "topic");
+      return this;
+    }
+
+    /**
+     * The field of the span json data will be parsed. Defaults to "message"
+     */
+    public Builder field(String field) {
+      this.field = checkNotNull(field, "message");
       return this;
     }
 
--- a/zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaCollectorWorker.java
+++ b/zipkin-collector/kafka10/src/main/java/zipkin/collector/kafka10/KafkaCollectorWorker.java
@@ -13,6 +13,8 @@
  */
 package zipkin.collector.kafka10;
 
+import com.google.gson.Gson;
+import java.util.HashMap;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -86,6 +88,20 @@ final class KafkaCollectorWorker implements Runnable {
               } catch (RuntimeException e) {
                 metrics.incrementMessagesDropped();
               }
+            } else if (bytes[0] == 123) {
+              // If the string start with "{", we parse the Json Object, then get specified field
+              Gson gson;
+              String str, newStr;
+              byte[] newBytes;
+              HashMap<String, String> map;
+
+              gson = new Gson();
+              str = new String(bytes);
+              map = gson.fromJson(str, map.getClass());
+              newStr = map.get("message");
+              newBytes = newStr.getBytes();
+
+              collector.acceptSpans(newBytes, DETECTING_DECODER, NOOP);
             } else {
               collector.acceptSpans(bytes, DETECTING_DECODER, NOOP);
             }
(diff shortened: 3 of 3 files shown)


codefromthecrypt:
hmm this would affect things besides Kafka, right? or is Kafka the only thing filebeat uses?

Can you post the code to your filebeat based zipkin reporter? Generally speaking we only accept our format..  I wonder if there's a way to have filebeat not envelope message data.

cc @openzipkin/elasticsearch 

feel free to chat here https://gitter.im/openzipkin/zipkin

WiFeng:
@adriancole  It wouldn't affect others things. I've also thought about other options, such as modifying the filebeat configuration to achieve the format zipkin requires, but it won't work.

If someone use filebeat, perhaps it's useful.

codefromthecrypt:
have you asked someone at elasticseatch to make it possible to send raw
data? or otherwise unwrap via config?

finally there is no zipkin filebeat setup I am aware of. can you please
describe how this is ending up in filebeats?

WiFeng:
@adriancole I haven't asked anyone at elasticsearch. I hope they can help me that make it possible to send raw data.

I can't understand your last words. Can you explain again ?

codefromthecrypt:
what I mean to ask is how filebeats is used anyway. for example no library
I know of uses filebeats.

so I am asking how filebeats is involved in zipkin data.

WiFeng:
Now the filebeat supports these output methods ([Configure the output](https://www.elastic.co/guide/en/beats/filebeat/current/configuring-output.html)):
```
Elasticsearch
Logstash
Kafka
Redis
File
Console
```
These outputs are all common components, and AFAIK the filebeat only can config one output method. In production environment a various of the log file should be collected, so couldn't specify the output is the zipkin api.

@tsg  @ruflin

codefromthecrypt:
a change to our main server should be due to a supported transport.

as far as I can guess, you have a custom zipkin sender or maybe even custom
tracer which puts json into a log.

instead of having us guess can you mention exactly what is generating this
trace data and why filebeats is your only choice?

we have other types of issues and custom setups especially unknown ones
should not affect our build. we have a "rule of 3" policy which means at
least a few different sites need a common concern before we add a feature
to the server everyone uses.

first step is showing on github exactly what you are doing.

xeraa:
> These outputs are all common components, and AFAIK the filebeat only can config one output method. In production environment a various of the log file should be collected, so couldn't specify the output is the zipkin api.

I'm not sure I fully understand that sentence. Though I get the general appeal of just sending out information once and then using it multiple times — for example for logs and traces.

Since you already seem to have Filebeat -> Kafka our general approach would then be Logstash to get the data to Elasticsearch for logging and you might be able to forward the data to Zipkin for tracing as well. There is a community plugin for a [Logstash to Zipkin output](https://github.com/whitemerry/logstash-output-zipkin); while it seems to be unfinished that would be a good approach in general. We see it pretty much as a feature that the Beats only have a single output (at the moment) and for any further distribution and splitting you'd use Logstash — optionally with a queue in between.

WiFeng:
@adriancole thanks for your replies. The PR may be more appropriate as my custom zipkin server, instead of merging to master branch. 

@xeraa thanks for your suggestion. We used the ELK tools to collect the server others logs, it's so pretty. But I also think the traces shoudn't forward to logstash, because of the traces process seem  to become  more complex.

If someone has the same technology stack, maybe it's useful. thanks~~

codefromthecrypt:
had a shower moment..

maybe if you implement unwrapping as a Kafka bytes deserializer.. just
unwrapping part.. we could make this work with no changes except putting
your type in the classpath and setting a property.

Kafka deserializer is a property iirc and so this could allow you to use
standard server.

WiFeng:
@adriancole It's very good. Can you give a example or add that into the documents? 

thank you!

ghost:
:)

shakuzen:
@WiFeng could you share the reasoning why you are outputting trace data to a log file rather than using a Zipkin reporter directly to Kafka?

WiFeng:
@shakuzen Thanks for your reply! It's for performance and reliability.

codefromthecrypt:
you'll need to get this into your classpath and set a property to use it via the docs in zipkin-server README. I'm closing this issue as I don't expect us to generalize this at this point.

```java
import com.fasterxml.jackson.core.TreeNode;
import com.fasterxml.jackson.databind.ObjectMapper;
import java.io.IOException;
import java.io.UncheckedIOException;
import org.apache.kafka.common.serialization.Deserializer;
import org.apache.kafka.common.utils.Bytes;

public final class UnwrappingJsonDeserializer implements Deserializer<Bytes> {
  static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();

  public static void main(String[] args) {
    String span = "{\n"
      + "  \"traceId\": \"6b221d5bc9e6496c\",\n"
      + "  \"id\": \"6b221d5bc9e6496c\"\n"
      + "}";
    String json = "{\"message\": " + span + "}";

    Bytes decoded = new UnwrappingJsonDeserializer().deserialize(null, json.getBytes());
    System.err.println(new String(decoded.get()));
  }

  @Override public Bytes deserialize(String topic, byte[] data) {
    if (data == null || data.length == 0) return null;
    try {
      TreeNode tree = OBJECT_MAPPER.readTree(data).at("/message");
      return Bytes.wrap(tree.toString().getBytes());
    } catch (IOException e) {
      throw new UncheckedIOException(e); // unexpected with a byte array
    }
  }
}
```

PR #2031, opened 2018-04-28
Fixes metrics response code and handles uris same as Micrometer WebMVC

zeagord:
This is to ensure dimensions don't explode on variables

### DIFF
--- a/zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin/autoconfigure/prometheus/ZipkinPrometheusMetricsAutoConfiguration.java
+++ b/zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin/autoconfigure/prometheus/ZipkinPrometheusMetricsAutoConfiguration.java
@@ -28,8 +28,17 @@ import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.web.embedded.undertow.UndertowDeploymentInfoCustomizer;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.util.StringUtils;
 
 @Configuration class ZipkinPrometheusMetricsAutoConfiguration {
+  // from io.micrometer.spring.web.servlet.WebMvcTags
+  private static final Tag URI_NOT_FOUND = Tag.of("uri", "NOT_FOUND");
+  private static final Tag URI_REDIRECTION = Tag.of("uri", "REDIRECTION");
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
It isn't obvious to me why we would intentionally not track UI latency... Mind elaborating a bit more? I might just not understand.

zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin/autoconfigure/prometheus/ZipkinPrometheusMetricsAutoConfiguration.java
```diff
@@ -75,11 +74,22 @@ UndertowDeploymentInfoCustomizer httpRequestDurationCustomizer() {
...
+        , Tag.of("path", cleanPathTemplate(exchange.getRelativePath()))
         , Tag.of("status", Integer.toString(exchange.getStatusCode()))
       );
     }
 
+    private String cleanPathTemplate(String path){
+      path = path.replaceAll("//", "/");
+      if (path.contains("/api/v1/trace"))
+        return "/api/v1/trace/{traceId}";
+      if (path.contains("/zipkin/traces/"))
```
codefromthecrypt: ahh this must be for crossroads

codefromthecrypt:
thanks.. I'll pull locally also split the logging thing from the other thing

codefromthecrypt:
we can't remove the log4j thing to zipkin-autoconfigure as it breaks kafka when run as an exec jar..

```
[INFO] 2018-04-29 10:43:46.809  WARN [/] 50622 --- [           main] ConfigServletWebServerApplicationContext : Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'kafka' defined in class path resource [zipkin/autoconfigure/collector/kafka/ZipkinKafkaCollectorAutoConfiguration.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [zipkin.collector.kafka.KafkaCollector]: Factory method 'kafka' threw exception; nested exception is java.lang.NoClassDefFoundError: org/apache/log4j/Logger
```

codefromthecrypt:
moved log4j stuff to a separate PR https://github.com/openzipkin/zipkin/pull/2032

shakuzen (commented):
After the fact, but LGTM

codefromthecrypt:
reviews before during and after the fact greatly appreciated, especially by
users like yourself!

PR #2036, opened 2018-04-30
Redirect /info endpoint for Boot 2 Actuator

shakuzen:
Similar to the other endpoints we mention in the Server README, the /info endpoint should be redirected since it was moved along with other Actuator endpoints during the Spring Boot 2 upgrade. Adds a test to ensure this is checked going forward.

I didn't put this config with the redirect for the prometheus endpoint since the info endpoint isn't exactly related to metrics or health. Actually, I was tempted to move the prometheus redirect configuration to where I made this to keep the redirect config in one place, but for now I submit these changes for review.


zeagord (approved):

codefromthecrypt:
just pushed a deflake commit and picked your commit and pused as https://travis-ci.org/openzipkin/zipkin/builds/373098445 if passes will rebase and merge yours in

codefromthecrypt:
closed via 452692479b8424b01f69cfe5b77704612475bfeb there was a slight test glitch

codefromthecrypt:
thanks for helping tommy!

Issue #2037, opened 2018-04-30
Upgrading to Springboot 2 adds double slash in the url


zeagord:
After upgrading to Springboot 2 the UI URLs after `/zipkin//` for the requests.

shakuzen:
Note that the URLs still work this way, but it is distracting and a difference we should probably fix if easy to do.

codefromthecrypt:
agreed should sort it. we should make the existing properties work like
they did before as they are added for things like bastion hosts. in other
words if feasible change via code not property value

PR #2038, opened 2018-04-30
[WIP] Fix for Log4J and trailing double slash in the URL of UI

zeagord:
This should fix the logging issue and the tests did not fail in Kafka collector. And has the fix for #2037 as well.

### DIFF
--- a/zipkin-server/src/main/resources/zipkin-server-shared.yml
+++ b/zipkin-server/src/main/resources/zipkin-server-shared.yml
@@ -140,7 +140,7 @@ zipkin:
     # Default is "match all websites"
     instrumented: .*
     # URL placed into the <base> tag in the HTML
-    base-path: /zipkin/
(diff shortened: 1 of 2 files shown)


zeagord:
Did not work. :/

zeagord:
Closing this as this is not a valid fix.

zipkin-server/src/main/resources/zipkin-server-shared.yml
```diff
@@ -140,7 +140,7 @@ zipkin:
     # Default is "match all websites"
     instrumented: .*
     # URL placed into the <base> tag in the HTML
-    base-path: /zipkin/
+    base-path: /zipkin
```
codefromthecrypt, line 143: ps people will set this with properties.. if there's a behavior change we might want to attack that in the UiConfiguration class

PR #2039, opened 2018-05-01
Fixes false positive where health status returned ok when not

codefromthecrypt:
Kafka 0.8 is busted since 2.0 but we didn't know as our health endpoint was also busted!

### DIFF
--- a/zipkin-server/src/main/java/zipkin/server/internal/MetricsHealthController.java
+++ b/zipkin-server/src/main/java/zipkin/server/internal/MetricsHealthController.java
@@ -20,7 +20,10 @@ import io.micrometer.core.instrument.MeterRegistry;
 import io.prometheus.client.CollectorRegistry;
 import java.util.HashMap;
 import java.util.Map;
+import org.springframework.boot.actuate.health.Health;
 import org.springframework.boot.actuate.health.HealthEndpoint;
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
verified manually. when #2040 is in, this would break the build

```bash
curl -s localhost:9411/health -v
*   Trying ::1...
* TCP_NODELAY set
> GET /health HTTP/1.1
> Host: localhost:9411
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 503 Service Unavailable
< Connection: keep-alive
< vary: origin
< Transfer-Encoding: chunked
< Content-Type: application/json;charset=UTF-8
< Date: Tue, 01 May 2018 06:46:54 GMT
< 
{"zipkin":{"status":"DOWN","details":{"InMemoryStorage":{"status":"UP"},"KafkaCollector":{"status":"DOWN","details":{"error":"org.I0Itec.zkclient.exception.ZkException: Unable to connect to foo"}}}},"status":"DOWN"}
```

shakuzen (approved):
We should probably add a test for this also, but that can be done afterwards.

codefromthecrypt:
we actually already have a test for it :) in the integration tests we make
sure health is down when kafka is.

problem was that kafka was always down!

PR #2041, opened 2018-05-01
Fixes UI basepath default

### LINKED ISSUE #2037: Upgrading to Springboot 2 adds double slash in the url

codefromthecrypt:
Fixes #2037


codefromthecrypt:
I looked at the code.. the actual code is fine, it believes what's here is actually the default!

PR #2045, opened 2018-05-09
To distinguish from the zipkin module.

tbwork:
This will cause a error in Eclipse when importing the project due to that there are two same artifacts in the maven project.

### DIFF
--- a/zipkin2/pom.xml
+++ b/zipkin2/pom.xml
@@ -24,7 +24,7 @@
   </parent>
 
   <groupId>io.zipkin.zipkin2</groupId>
-  <artifactId>zipkin</artifactId>
+  <artifactId>zipkin2</artifactId>
   <name>Zipkin v2</name>
 
   <properties>


zipkin2/pom.xml
```diff
@@ -24,7 +24,7 @@
   </parent>
 
   <groupId>io.zipkin.zipkin2</groupId>
-  <artifactId>zipkin</artifactId>
+  <artifactId>zipkin2</artifactId>
```
codefromthecrypt, line 27: Appreciate the thought, but this change would have broken the entire ecosystem if published to maven central. Basically, people working now would not be able to simply update their versions.
io.zipkin.zipkin2:zipkin is a completely different artifact if changed to io.zipkin.zipkin2:zipkin2

codefromthecrypt:
As mentioned on #2046, the only reason you are experiencing a glitch is likely eclipse doesn't know what to name this on override. For example, Intellij names it "zipkin (2)" or something. There's likely a way to do this neater in eclipse, but also mentioned in that issue that this concern will go away when the old library (io.zipkin.java) is removed.

In the future, I wouldn't recommend us dual-publishing libraries for more than a release cycle. Rather, we'd keep a long running `io.zipkin.zipkin3` branch until there is no dependencies on `io.zipkin.zipkin2`. However, we are were we are: there's work to do namely around the collector, UI and also mysql which implies we have at least a person-month of effort to get rid of dependencies on the v1 code.. maybe more than that.

Personally, I will be working on this effort when I return.

codefromthecrypt:
see https://github.com/openzipkin/zipkin/issues/2047

Issue #1232, opened 2016-08-10
PostgreSQL support


codefromthecrypt:
_From @adriancole on March 11, 2016 1:19_

@Koizumi85 asked on gitter about PostgreSQL support. spanstore-jdbc takes advantage of some SQL syntax which might not be supported in PostgreSQL. Before we start down this path, let's figure out some basics. For example, which version of PostgreSQL is needed, so that we can test the configuration desired.

_Copied from original issue: openzipkin/zipkin-java#93_

codefromthecrypt:
@Koizumi85 nag..

codefromthecrypt:
_From @eirslett on March 21, 2016 10:21_

Could it be supported as an external/optional module? I'm not sure it's a good idea to reuse the JDBC code, even if that's tempting.

codefromthecrypt:
_From @Koizumi85 on March 21, 2016 10:28_

Actually i think with postgresql 9.5 the actual code could possibly work if there were a possibility to switch the database type in Jooq..
Background: psql9.5 supports "upsert" semantics which are heavily used here.
But postgresql 9.5 is fairly new and is not hosted in the repositories of Ubuntu LTS and other linux distributions and there are a lot of sysadmins who don't want to take the postgresql repositories for LTS reasons.

codefromthecrypt:
_From @eirslett on March 21, 2016 12:11_

9.1 or 9.3 are the safest choices, I think.

codefromthecrypt:
_From @Koizumi85 on March 21, 2016 13:10_

9.1 and 9.3 should be almost identically unless the json column type of 9.3. but i don't think we would need this here, or?
So I think a 9.1 compatibility would be nice.

codefromthecrypt:
_From @eirslett on March 21, 2016 13:17_

Sounds good to me. The JSON query capabilites would be nice to have (for things like annotations), but it wasn't added in MySQL until 5.7.8, I think many linux distros are at MySQL 5.5 or so. And the existing SQL span store doesn't use JSON.

dwelch2344:
Any word on this? We dev against PG and would love to not have another database dep just for zipkin (especially on dev machines)

codefromthecrypt:
your are lucky number 3 person to be interested in this! https://blog.codinghorror.com/rule-of-three/

if we started this, would you help maintain and/or test it? Which version of postgresql can you use?

dwelch2344:
[...]

PR #2052, opened 2018-05-18
add support postgresql as storage backend

tian-junwei:
add support postgresql as storage backend


codefromthecrypt:
Thanks for the help.

This looks like a straight port of the mysql one, right?

We do have demand, but if we added it, we should use the v2 codebase (ex
you can look at the cassandra or elasticsearch ones for an example). Also
we'd want to consider the schema, if it is best or not.

codefromthecrypt:
if doing postgresql, let's make the schema a bit quicker. Everything is v2 type now, so it should be possible!
 #1232

Issue #2047, opened 2018-05-09
Get rid of the v1 codebase


codefromthecrypt:
We don't have a lot of hands to do "huge maintenance work". There was a lot of demand for a v2 format, so we released a v2 core library as soon as we could (september 2017).

With this in place, we were able to decouple libraries from the v1 codebase. This included brave and zipkin-reporter, which no longer have mandatory dependencies on it. However, there's still a large amount of work to do. This issue tracks that work:

- [x] Port mysql (as-is) to the zipkin2 codebase, or redo it
- [x] Port "cassandra" (as-is) to the zipkin2 codebase, or drop it
- [x] Remove zipkin v1 dep from zipkin-junit and repackage under io.zipkin.zipkin2 namespace
- [x] Make a new collector, which accepts even the old v1 thrift: make the server depend on it.
- [x] redo all the storage tests to port to v2 codebase
- [x] implement the zipkin UI in v2 javascript (implies clock skew and merge logic)

Making this issue was Influenced, but not caused by a strange recent flurry of issues #2044 #2045 #2046 It has always been implicitly on the backlog

codefromthecrypt:
cc @openzipkin/core @shakuzen @marcingrzejszczak @devinsba I will start on this when I return from holiday 21 May

codefromthecrypt:
here's a corresponding issue in brave https://github.com/openzipkin/brave/issues/699

tbwork:
@adriancole Enjoy your vacation~

codefromthecrypt:
@tbwork thanks tons! enjoying the last of it now :)

PS thought while away:

io.zipkin.zipkin2:zipkin-collector could be a new java 8+ only jar. The only consumer of this code is the server, which is java 8 anyway. This will make things easier as formerly there was tension about the codebase which is only used by the server/collector.

Note: A future version of this collector could do reactive streams (1.0 or via java 9) or other fanciness, but I don't think we should do that now. There is so much work just to migrate off the v1 code we should make sure this completes before working on candy.

codefromthecrypt:
https://github.com/openzipkin/zipkin/pull/2066 is the conversion library.. need to vet this through storage and the dependency linker code

tbwork:
Rogar that!

codefromthecrypt:
storage are converted. next step is to make the server not require zipkin v1 storage component, which simplifies autoconfig. This also implies redoing autoconfig to not export it either.

codefromthecrypt:
[...]

PR #2060, opened 2018-05-22
Kafka storage backend

afalko:
This is one of the requirements for the architecture described here: https://docs.google.com/document/d/11_MB69A6d_9_N5ClxVcUhKJRMvVI18voV5cGv1Ji5oc/edit?usp=sharing

This PR is experimental and works our canary environment. I'm sending it out before it is fully baked so to get early feedback. 

TODO list before we can merge:
- [ ] Unit tests
- [ ] Integration tests
- [ ] Friendlier error handling


codefromthecrypt:
nice start and thanks for looking at this. cc @ImFlog @bplotnick @drolando

One thing I'm wondering is if we can somehow reuse the code from the async reporter which can bundle messages at an appropriate size for the topic. It looked like your design doc abandoned that idea... maybe because it was awkward..  At any rate it would be nice to find a way to get the sizing code into the right place. Maybe we can address this concern in a replacement collector module perhaps.. https://github.com/openzipkin/zipkin/issues/2047#issuecomment-390358196

afalko:
@adriancole Yeh, my initial PoC of this used the KafkaSender async reporter, but then I had no way to pull in the server-components from zipkin-server. I built my own undertow server that would go out of date with zipkin-server improvements. 

In theory, I could include the async reporter kafka module into the storage module here; but that seems like it would be a strange dependency chain. Happy to reconsider that; I might need to copy some of the callback stuff to handle errors. 

Although the current code sends a kafka message per span, it'll carry us through for a while in terms of scale. With 3x acks=all kafka configuration, we can push 25k 1k messages/sec, which is 2.2 billion spans per day. If you don't replicate on kafka and set acks=0, then you can do 100k+ messages/sec.

codefromthecrypt:
Thanks for the notes, Andrey. One little thing is that we don't have
bundling at the storage side at the moment... what this means is the other
side would make a new storage request for each span. Maybe in some
environments that's ok, notably cassandra would be OK. However, bundling
could help those who can take advantage of batching commands.. What are you
using for ultimate storage?

afalko:
Ahh, I didn't think about what happens on the other end. Do you want me to do the bundling work as part of this PR or a followup PR?

We are running ES right now, but it is on my todo list to research whether ES or Cassandra scales better. Has anyone in the community done an evaluation between the two already?

codefromthecrypt:
It is probably better to have a bundling option, and I do wonder if we
shouldn't revisit this as a sender or at least a message queue based
sender, so that we don't end up with a complete copy/paste just for
rabbitMQ instead. cc @shakuzen

afalko:
Thinking about it more, bundling seems be to be better done on the storage layer-side; e.g. why should the sender know the optimal bundle size for a storage backend they might not even know about. 

Building off your idea for an MQ based sender; what if Zipkin were to be made to run storage-less and instead an admin could configure MQ ingest instead? The current kafka code probably wouldn't change too much; but get shifted around to be behind a more suitable MQ interface. Batching config could be added later, so there is a batch for MQ, then a separate batch for storage.  

Once I submit the filter interface PR w/ quota filters; putting that stuff in the senders will be dangerous because there is no guarantee that service owners will depend on the correct sender lib and configure it correctly.

codefromthecrypt:
>
> Thinking about it more, bundling seems be to be better done on the storage
> layer-side; e.g. why should the sender know the optimal bundle size for a
> storage backend they might not even know about.
>
this is more a point of sending 1 span at a time which is known to not be
efficient, vs believing we know the best bunding size. For example, the
storage layer can rebundle smaller, but only if the input was bundled in
the first place. We don't want to introduce state between different
collector messages (Which is how we'd do bundling in the case where people
only send 1 span)..

> Building off your idea for an MQ based sender; what if Zipkin were to be
> made to run storage-less and instead an admin could configure MQ ingest
> instead? The current kafka code probably wouldn't change too much; but get
> shifted around to be behind a more suitable MQ interface. Batching config
> could be added later, so there is a batch for MQ, then a separate batch for
> storage.
>
I think this is more the proxy use case in general. For example, we've long
had requests for http proxying (which isn't quite as safe as MQ based as
there's no buffer), but could help simplify a backend as that backend only
needs to talk zipkin2. The key here though is to not eliminate batching
which is very much a norm in MQ, without forcing a change to other code
which doesn't have batching by default. The reporter's async reporter does
have working bundling and it is all about the transport, less about
storage. Using that, or an approach derived from that has the least risk.


> Once I submit the filter interface PR w/ quota filters; putting that stuff
> in the senders will be dangerous because there is no guarantee that service
> owners will depend on the correct sender lib and configure it correctly.
>
I'm not sure about the danger part, because we often limit things here
anyway. For example, I don't expect us to make a configuration that is
completely arbitrary. For example, I meant literally rabbit as that's the
other supported MQ. Basically consider how this impacts design. If someone
does something custom using the same code, and misconfigures it.. it is on
them. We are rather strict with what's supported and not, so yeah I would
defer this part (the policy from the mechanism)

On the quota filters and this change in general, I'd caution that depending
on how many features and how the code looks in general it might not be
merged as is. So please do proceed, but keep in mind it might be moved
around, possibly to another repo like we do with zipkin-gcp depending on
how things go. It is really important to know the general features that
will be used in filtering to see the least code needed to accomplish it,
and for example, some of it might move to collector v2.

In other words, please do put in code that does "exactly" what you need as
well, including filtering examples, so that when we think together about
how to enable this, we don't break it in the process.

Cheers and thanks tons!

afalko:
After working on https://github.com/openzipkin/zipkin/pull/2072 and talking about it internally. We think it would be easiest to pull our Kafka storage out to a separate repo that we would be happy to maintain and make changes (in needed) to make storage backends more pluggable. Thoughts on this @adriancole?

rmichela:
I like the idea of being able to load custom storage backends from the classpath as plugins.

codefromthecrypt:
I think a separate repo is a great place to experiment and solidify
something. When you create it, link back to here the initial import.

zipkin-autoconfigure/storage-kafka/src/main/java/zipkin/autoconfigure/storage/kafka/ZipkinKafkaStorageAutoConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+@Configuration
+@EnableConfigurationProperties(ZipkinKafkaStorageProperties.class)
+@ConditionalOnProperty(name = "zipkin.storage.type", havingValue = "kafka")
+@ConditionalOnMissingBean(StorageComponent.class)
+class ZipkinKafkaStorageAutoConfiguration {
+
+  @Bean
+  @ConditionalOnMissingBean
+  V2StorageComponent storage(ZipkinKafkaStorageProperties properties) {
```
codefromthecrypt, line 38: fyi as of latest, this dance is no longer needed as we don't need to support v1 storage anymore from a bean POV

afalko:
Good news @adriancole I was able to get this working by pulling the kafka impl out of code base: https://github.com/afalko/zipkin-storage-kafka . The only thing I haven't tested out yet is whether you can plop config values into `zipkin.yml`. That is low priority for me at the moment. 

Here is how I launch the new storage module:
```
STORAGE_TYPE=kafka java -Dzipkin.storage.kafka.bootstrap-servers=localhost:9092 -Dloader.path=../zipkin-storage-kafka/zipkin-autoconfigure-storage-kafka/target/zipkin-autoconfigure-storage-kafka-2.9.3.jar,../zipkin-storage-kafka/zipkin-storage-kafka/target/zipkin-storage-kafka-2.9.3.jar -cp zipkin-server/target/zipkin-server-*exec.jar org.springframework.boot.loader.PropertiesLauncher
```

cc: @rmichela

codefromthecrypt:
whoot! PS I was going to mention a trick for experimenting with the
adjuster stuff.

you can auto-configure a wrapper like we do for tracing. in your storage
wrapper, you wrap the span consumer.. inject and test your filter/adjusters
this way (as a pre-step to storage which is correct anyway). Might help
decoupling


@Configuration
static class AdjustingStorageComponentEnhancer implements BeanPostProcessor {

  @Autowired(required = false)
  YourStuff yourStuff;

  @Override
  public Object postProcessBeforeInitialization(Object bean, String beanName) {
    return bean;
  }

  @Override
  public Object postProcessAfterInitialization(Object bean, String beanName) {
    if (yourStuff == null) return bean;
    if (bean instanceof StorageComponent) {
      return new AdjustingStorageComponent(yourStuff, (StorageComponent) bean);
    }
    return bean;
  }
}

codefromthecrypt:
@jeqo I think we can close this in favor of your experimental repo.. is that the case?

jeqo:
@adriancole that would be awesome! didn't know this effort before, but it would be great if @afalko @rmichela could check if https://github.com/jeqo/zipkin-storage-kafka covers their use-cases before.

afalko:
@jeqo Thanks for doing that work; very cool you got writes working with Streams. I'm no longer on the company and team maintaining this branch/abovementioned write-only Kafka solution. @rmichela will be in a better position to make the call on how to proceed.

rmichela:
I’ve been on paternity leave for the past few months and am really out of the loop.

codefromthecrypt:
@rmichela proceed with parenting :P

I'm going to close this as much has changed since.
* For those looking for an http/kafka forwarder, there's https://github.com/HotelsDotCom/pitchfork by @worldtiki and folks
* For those looking for near-realtime processing of spans (ex dependency graph linking) there's  https://github.com/jeqo/zipkin-storage-kafka

I'm glad we had this to start the fire, but I don't think we need to re-own this PR. Thanks @afalko for being the early bird!

PR #2064, opened 2018-05-29
Filters

afalko:
@rmichela

This didn't work, so I'm going to pump something out by doing reflection myself rather than trying to get Spring to do it. 

Here is the example filter: https://github.com/afalko/ExampleSpanFilter

### DIFF
--- a/zipkin-collector/kafka/src/main/java/zipkin/collector/kafka/KafkaCollector.java
+++ b/zipkin-collector/kafka/src/main/java/zipkin/collector/kafka/KafkaCollector.java
@@ -15,6 +15,7 @@ package zipkin.collector.kafka;
 
 import java.io.IOException;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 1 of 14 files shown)


codefromthecrypt:
Hi there. don't forget to link issues when making pull requests ad folks might not know the overall plan.

as mentioned this might end up in the v2 package but regardless thanks for the spike

afalko:
Please accept my appologies @adriancole. I meant to make this a PR to my own fork before blasting it out to the community. I'll close it and create a proper one when I'm ready.

codefromthecrypt:
no worries.. glad to see wheels turning

codefromthecrypt:
ps this might give a general structure for add-ons https://github.com/openzipkin-attic/zipkin-sparkstreaming-example I'd expect we'd want something like this, like you can just add to the classpath and go

afalko:
Thanks for the pointer! @rmichela was able to get the add-classpath-and-go part working in the sample plugin: https://github.com/afalko/ExampleSpanFilter/pull/1

It works e2e, so stay tuned :)

PR #2066, opened 2018-05-30
Adds V1 conversion library

codefromthecrypt:
This allows us to port old storage to zipkin2 packaged code while still
being able to read v1 structs.

Note: this intentionally ignores unsupported Binary Annotation types
such as I64. In doing so, the code is less difficult to maintain.

See #2047


codefromthecrypt:
due to the likelihood of classpath mismatch and the relatively small size of this library, I will fold it into the main zipkin jar. This also simplifies format detection, having all formats in one place.

codefromthecrypt:
ok rewrote under the zipkin2 library and backported tests. Will do more integration experiments tomorrow.

PS the jar is still <200KiB, so I don't feel so bad :P

```
$ du -k  /Users/acole/oss/zipkin/zipkin2/target/zipkin-2.8.5-SNAPSHOT.jar
196	/Users/acole/oss/zipkin/zipkin2/target/zipkin-2.8.5-SNAPSHOT.jar
```

zipkin2/src/main/java/zipkin2/SpanBytesDecoderDetector.java
```diff
@@ -0,0 +1,110 @@
...
+//
+// When byte(0) == '[' (91), assume it is a list of json-encoded spans
+//
+// When byte(0) == 10, assume it is a proto3-encoded span or trace ID field
+//
+// When byte(0) <= 16, assume it is a TBinaryProtocol-encoded thrift
+// .. When serializing a Span (Struct), the first byte will be the type of a field
+// .. When serializing a List[ThriftSpan], the first byte is the member type, TType.STRUCT(12)
+// .. As ThriftSpan has no STRUCT fields: so, if the first byte is TType.STRUCT(12), it is a list.
+public final class SpanBytesDecoderDetector {
```
codefromthecrypt, line 38: cc @drolando this will be the replacement for that internal class you were looking at earlier

cc also @wu-sheng

zipkin2/src/main/java/zipkin2/internal/ThriftEndpointCodec.java
```diff
@@ -0,0 +1,95 @@
...
+  static Endpoint read(ByteBuffer bytes) {
+    Endpoint.Builder result = Endpoint.newBuilder();
+
+    while (true) {
+      ThriftField thriftField = ThriftField.read(bytes);
+      if (thriftField.type == TYPE_STOP) break;
+
+      if (thriftField.isEqualTo(IPV4)) {
+        int ipv4 = bytes.getInt();
+        result.parseIp(
```
adriancole: TODO: make parsing from int to avoid alloc

zipkin2/src/main/java/zipkin2/internal/V1JsonSpanReader.java
```diff
@@ -0,0 +1,187 @@
...
+    if (stringValue != null) {
+      return V1BinaryAnnotation.createString(key, stringValue, endpoint);
+    }
+    if (booleanValue != null && booleanValue && endpoint != null) {
+      return V1BinaryAnnotation.createAddress(key, endpoint);
+    }
+    return null; // toss unsupported data
+  }
+
+  static final JsonReaderAdapter<Endpoint> ENDPOINT_READER =
```
codefromthecrypt: TODO: delete and reuse same code from v2 as it is the same

zipkin2/src/main/java/zipkin2/SpanBytesDecoderDetector.java
wu-sheng: Sure. Travel around in the whole weekend. :)

PR #2071, opened 2018-05-31
Switches java license format to code comment style

codefromthecrypt:
Otherwise latest code formatter adds javadoc tags to the license!


pom.xml
```diff
@@ -723,6 +723,10 @@
         <configuration>
           <!-- session.executionRootDirectory resolves properly even with nested modules -->
           <header>${main.basedir}/src/etc/header.txt</header>
+          <mapping>
```
codefromthecrypt, line 726: here is the actual change

PR #2072, opened 2018-06-01
Support for plugin span filters

afalko:
This is one of the requirements for the architecture described here: https://docs.google.com/document/d/11_MB69A6d_9_N5ClxVcUhKJRMvVI18voV5cGv1Ji5oc/edit?usp=sharing

In addition to SpanFilters we'd like to introduce HttpFilters so that we can respond to things found in HttpHeaders. That would be a follow up PR. 

This is an enrichment filter we plan to deploy to production as our first filter: https://github.com/afalko/zipkin-generic-enrichment-filter . I'm happy to move this filter to openzipkin project if requested. 

Additional filters that we plan to implement:

> ServiceTagMappingFilter
	This filter would take in a static configuration where services claim tag and annotation identifiers. If another service comes in and tries to use a reserved tag, it will get a 409 (conflict) error from our server. An alternative scheme can be developed whereby annotations and tags are namespaced, the drawback of that would be an increase in storage requirements.
	Another way to implement this filter is to remove conflicting tags, but not drop the message. In that case, the regular 202 http message is sent back to the caller. 

> GDPRFilters
	These filters depend on the company and the nature of the data processed by the apps of that company.  There might be cases where the implementer might want to enrich spans or replace the data within the spans. These would likely never have any use if they are open sourced. They are not in the minimum viable change path.

An additional example filter that does blocking can be found here: https://github.com/afalko/ExampleSpanFilter

- [] Utests for failure cases

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,27 @@ $ ./mvnw -DskipTests --also-make -pl zipkin-server clean install
 $ java -jar ./zipkin-server/target/zipkin-server-*exec.jar
 ```
 
+## Zipkin Filters
+
+Zipkin supports a plugin model for filters. You can take filters off the shelf or create your own. Right now you can 
+filter incoming Spans. This can be useful if you want to block certain Spans from being stored for capacity 
+ reasons, edit contents of Spans for prevent PII data from getting stored, or enrich Spans with additional information.     
+ Please follow the examples mentioned into the links below to find out how to create filters. 
+
+### Zipkin Span Filters
+
+Here is an example enrichment filter that allows you to inject a tag to your spans from environment passed to Zipkin:
+[zipkin-generic-enrichment-filter](https://github.com/afalko/zipkin-generic-enrichment-filter) 
+
+### Configuring Filters
+
+You are responsible for placing any filters onto Spring's loader.path and using `PropertiesLancher` to run Zipkin:
+```
+java -Dloader.path=/path/to/my/filters/ExampleSpanFilter-1.0.0.jar \
+  -cp zipkin-server-*exec.jar \
+  org.springframework.boot.loader.PropertiesLauncher
+```
+
 ## Artifacts
 ### Library Releases
 Releases are uploaded to [Bintray](https://bintray.com/openzipkin/maven/zipkin).
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -286,6 +286,8 @@
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-maven-plugin</artifactId>
         <configuration>
+          <!-- This forces Spring Boot to use the PropertiesLauncher for startup, enabling loader.path -->
+          <layout>ZIP</layout>
           <classifier>exec</classifier>
           <executable>true</executable>
           <excludeArtifactIds>logback-classic</excludeArtifactIds>
--- a/zipkin-server/src/main/java/zipkin/server/internal/ZipkinHttpCollector.java
+++ b/zipkin-server/src/main/java/zipkin/server/internal/ZipkinHttpCollector.java
@@ -21,6 +21,9 @@ import io.undertow.util.HttpString;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
 import java.util.zip.GZIPInputStream;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
@@ -32,7 +35,10 @@ import zipkin2.codec.SpanBytesDecoder;
 import zipkin2.collector.Collector;
 import zipkin2.collector.CollectorMetrics;
 import zipkin2.collector.CollectorSampler;
+import zipkin2.collector.filter.SpanFilter;
 import zipkin2.storage.StorageComponent;
+
+import javax.xml.ws.http.HTTPException;
 
 /** Implements the POST /api/v1/spans and /api/v2/spans endpoints used by instrumentation. */
 @Configuration
@@ -51,13 +57,15 @@ class ZipkinHttpCollector implements HttpHandler, HandlerWrapper {
 
   @Autowired
   ZipkinHttpCollector(
-      StorageComponent storage, CollectorSampler sampler, CollectorMetrics metrics) {
+      StorageComponent storage, CollectorSampler sampler, CollectorMetrics metrics,
+      Optional<List<SpanFilter>> filters) {
     this.metrics = metrics.forTransport("http");
     this.collector =
         Collector.newBuilder(getClass())
             .storage(storage)
             .sampler(sampler)
             .metrics(this.metrics)
+            .filters(filters.orElse(Collections.emptyList()))
             .build();
     this.JSON_V2 = new HttpCollector(SpanBytesDecoder.JSON_V2);
     this.PROTO3 = new HttpCollector(SpanBytesDecoder.PROTO3);
@@ -152,7 +160,13 @@ class ZipkinHttpCollector implements HttpHandler, HandlerWrapper {
 
   static void error(HttpServerExchange exchange, Throwable e) {
     String message = e.getMessage();
-    int code = message == null || message.startsWith("Cannot store") ? 500 : 400;
+    int code;
+    if (e instanceof HTTPException) {
+      HTTPException httpException = (HTTPException)e;
+      code = httpException.getStatusCode();
+    } else {
+      code = message == null || message.startsWith("Cannot store") ? 500 : 400;
+    }
     if (message == null) message = e.getClass().getSimpleName();
     exchange.setStatusCode(code).getResponseSender().send(message);
   }
--- a/zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
+++ b/zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
@@ -22,6 +22,7 @@ import zipkin2.Callback;
 import zipkin2.Span;
 import zipkin2.SpanBytesDecoderDetector;
 import zipkin2.codec.BytesDecoder;
+import zipkin2.collector.filter.SpanFilter;
 import zipkin2.storage.StorageComponent;
 
 import static java.lang.String.format;
@@ -48,6 +49,7 @@ public class Collector { // not final for mock
     StorageComponent storage = null;
     CollectorSampler sampler = null;
     CollectorMetrics metrics = null;
+    List<SpanFilter> filters = null;
 
     Builder(Logger logger) {
       this.logger = logger;
@@ -74,15 +76,23 @@ public class Collector { // not final for mock
       return this;
     }
 
+    /** @see {@link CollectorComponent.Builder#filters(List<SpanFilter)} */
+    public Builder filters(List<SpanFilter> filters) {
+      if (filters == null) throw new NullPointerException("filters == null");
+      this.filters = filters;
+      return this;    }
+
     public Collector build() {
       return new Collector(this);
     }
+
   }
 
   final Logger logger;
   final CollectorMetrics metrics;
   final CollectorSampler sampler;
   final StorageComponent storage;
+  final List<SpanFilter> filters;
 
   Collector(Builder builder) {
     if (builder.logger == null) throw new NullPointerException("logger == null");
@@ -91,6 +101,7 @@ public class Collector { // not final for mock
     if (builder.storage == null) throw new NullPointerException("storage == null");
     this.storage = builder.storage;
     this.sampler = builder.sampler == null ? CollectorSampler.ALWAYS_SAMPLE : builder.sampler;
+    this.filters = builder.filters;
   }
 
   public void accept(List<Span> spans, Callback<Void> callback) {
@@ -133,6 +144,7 @@ public class Collector { // not final for mock
     List<Span> spans;
     try {
       spans = decodeList(decoder, serializedSpans);
+      spans = filterSpans(spans, callback);
     } catch (RuntimeException e) {
       callback.onError(errorReading(e));
       return;
@@ -144,6 +156,22 @@ public class Collector { // not final for mock
     List<Span> out = new ArrayList<>();
     if (!decoder.decodeList(serialized, out)) return Collections.emptyList();
     return out;
+  }
+
+  /**
+   * Take the list of spans and pump them through pre-configured filters
+   *
+   * @param spans
+   */
+  List<Span> filterSpans(List<Span> spans, Callback<Void> callback) {
+    List<Span> processed = spans;
+    if (filters == null) {
+      return spans;
+    }
+    for (SpanFilter filter : filters) {
+      processed = filter.process(processed, metrics, callback);
+    }
+    return processed;
   }
 
   void record(List<Span> sampled, Callback<Void> callback) {
--- a/zipkin-collector/core/src/main/java/zipkin2/collector/CollectorComponent.java
+++ b/zipkin-collector/core/src/main/java/zipkin2/collector/CollectorComponent.java
@@ -15,6 +15,7 @@ package zipkin2.collector;
 
 import java.util.List;
 import zipkin2.Component;
+import zipkin2.collector.filter.SpanFilter;
 import zipkin2.storage.SpanConsumer;
 import zipkin2.storage.StorageComponent;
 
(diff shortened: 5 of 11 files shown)


zipkin-server/pom.xml
```diff
@@ -283,6 +283,8 @@
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-maven-plugin</artifactId>
         <configuration>
+          <!-- This forces Spring Boot to use the PropertiesLauncher for startup, enabling loader.path -->
```
codefromthecrypt, line 289: any impact to this? especially for existing modules? cc @zeagord cc @shakuzen

zipkin-server/src/main/java/zipkin/server/internal/ZipkinHttpCollector.java
```diff
@@ -56,10 +61,10 @@
   private HttpHandler next;
 
   @Autowired ZipkinHttpCollector(StorageComponent storage, CollectorSampler sampler,
-    CollectorMetrics metrics) {
+    CollectorMetrics metrics, Optional<List<SpanFilter>> filters) {
```
codefromthecrypt: we don't have to inject optional right? can't we just get an empty list?

zipkin/src/main/java/zipkin/collector/Collector.java
```diff
@@ -74,23 +78,34 @@ public Builder sampler(CollectorSampler sampler) {
       return this;
     }
 
+    /** @see {@link CollectorComponent.Builder#filters(List<SpanFilter>)} */
+    public Builder filters(List<SpanFilter> filters) {
+      logger.info("Total filters loaded: " + filters.size());
```
codefromthecrypt: remove this.. as we don't log builder invocations. probably can accomplish the same at higher abstraction ex /autoconfig endpoint

zipkin/src/main/java/zipkin/filter/FilterActivatedException.java
```diff
@@ -0,0 +1,14 @@
+package zipkin.filter;
+
+public class FilterActivatedException extends Throwable {
+  private final int httpCode;
+
+  public FilterActivatedException(String message, int httpCode) {
+    super(message);
+    this.httpCode = httpCode;
+  }
+
```
codefromthecrypt: this seems leaky abstraction as it isnt necessarily http but has http code. also extending throwable is a bit weird. maybe use a built-in exception or document how this is used as a signal. we don't currently define custom exceptions anywhere else in the codebase

zipkin/src/main/java/zipkin/filter/SpanFilter.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.util.List;
+
+public interface SpanFilter {
+  /**
+   * Process filters given a set of v1 spans. The callback should return a FilterActivatedException if filter
+   * implementor wants to produce custom return codes back to the user.
+   * @param spans
+   * @param callback
+   */
```
codefromthecrypt: hmm are you hinting there is a partial success here? ex you can return without exception?

zipkin/src/main/java/zipkin/internal/Collector.java
```diff
@@ -55,13 +63,35 @@ protected void acceptSpans(byte[] serializedSpans, D decoder, Callback<Void> cal
...
 
+  /**
+   * Take the list of spans and pump them through pre-configured filters
+   *
+   * @param spans
+   */
+  List<S> filterSpans(List<S> spans, Callback<Void> callback) {
+    // These should never be processed in parallel.
+    // If you need parallel processing put that into your specific filter implementation.
+    boolean isV1Span = spanType.getType().equals(Span.class);
```
codefromthecrypt: we won't be filtering on two types. you can assume later it will be only zipkin2.span

zipkin2/pom.xml
```diff
@@ -60,6 +60,12 @@
       <version>2.8.2</version>
     </dependency>
```
codefromthecrypt: heh no way jose ;) gave can't be a core dep as it conflicts

README.md
```diff
@@ -107,6 +107,30 @@ $ ./mvnw -DskipTests --also-make -pl zipkin-server clean install
...
+Zipkin supports a plugin model for filters. You can take filters off the shelf or create your own. There are two classes
+of filters: a) Span Filters - where you want to operate on individual spans. b) Http Filters - where you want to operate
+on headers. If you want to create your own, please follow the examples mentioned into the links below.
+
+### Zipkin Span Filters
+
+Here is an example enrichment filter that allows you to inject a tag to your spans from environment passed to Zipkin:
+[zipkin-generic-enrichment-filter](https://github.com/afalko/zipkin-generic-enrichment-filter)
+
+### Zipkin Http Filters
```
codefromthecrypt: please remove forward refs to http filters which I am not sure what they will be. makes the change easier to pin down.

codefromthecrypt (commented):
thanks for spiking this. I think filter might not be the best word as it seems this is also mutating data.. maybe put in the PR the use cases this is for?

zipkin-server/pom.xml
afalko: @rmichela helped me with all of the Spring stuff, so he might be able to go into more detail on this.

zipkin-server/src/main/java/zipkin/server/internal/ZipkinHttpCollector.java
afalko: I had this with `@Nullable` before, but I kind of like `Optional` from a readability standpoint. Happy to change it back. @rmichela, is it better to use `Optional` or `@Nullable`?

zipkin/src/main/java/zipkin/filter/FilterActivatedException.java
afalko: I'll look around from suitable built-ins. I'll add better documentation too. The reason I extended `Throwable` was because the error call takes a throwable: https://github.com/openzipkin/zipkin/pull/2072/files#diff-1b30356bd2b01dbbfd9152c17a1ac4baR148

In theory, I could copy what you did in that same function: `message.startsWith("Cannot store")`, but that isn't strongly typed and requires an implementor to have to be careful about their error strings. I wanted to avoid that drawback and make it easier on filter implementors. Happy to send another PR to clean replace `message.startsWith`.

zipkin-server/src/main/java/zipkin/server/internal/ZipkinHttpCollector.java
rmichela: The `Optional<T>` is required so Spring knows the constructor injected `filters` parameter is not required. Without it, the bean graph will fail to load at startup when no `SpanFilter` plugin implementations are registered with Spring.

https://stackoverflow.com/a/42135423

zipkin/src/main/java/zipkin/filter/SpanFilter.java
afalko: If the implementor so chooses they could consider ingesting all spans, but throwing an error. I think that would be a rare case, but I can't think of reasons to get in someone's way of doing that; nor can I think of a strong reason to go out of our way to support that very well. Do you want me to have the call of the interface stop if exception is added to callback?

zipkin/src/main/java/zipkin/internal/Collector.java
afalko: Ahh, that's good news; I won't need the guava token type once that is that case. In terms of "later", do you want me to make this only supported for v2 as part of this PR?

zipkin2/pom.xml
afalko: I only need it for `TokenType` to determine v1 vs. v2. If zipkin2.Span is that only thing that is supported in the next week or two, then I can remove this. Otherwise, I need suggestions :).

afalko:
Most of our use cases are outlined in the doc I linked, but I'll paste them in the PR description shortly. Other names I can think of are `SpanProcessors`, `SpanManipulators`, `SpanManglers`. Open to other suggestions or votes; naming is hard :)

codefromthecrypt:
ya only handle v2 as it makes it simpler.

codefromthecrypt:
ah ok. i think sometimes we are using a defaulted optional autowired field
for cases like this (I think we do in elasticsearch storage for long to
interceptors) but yeah what you are doing sounds fine.

codefromthecrypt:
personally I like the idea of the filters being pure functions so that we
don't have to setup a separate sort of mechanism for exceptions there. if
there are "on error resume next" type concerns, a decorating filter might
be able to silence those things. test cases for what we are expecting here
can help drive design.

this made me also think that this should probably accept collector metrics
so that it can increase dropped spans if it drops some.

codefromthecrypt:
ok this can be redone over the new zipkin-collector library, which is an abstract class so no api break :P

I'd recommend just using the Adjuster type as we did some work on this for the sparkstreaming api.

https://github.com/openzipkin-attic/zipkin-sparkstreaming/blob/master/sparkstreaming/src/main/java/zipkin/sparkstreaming/Adjuster.java

The only difference is that it is unlikely the spans will be in the same trace..

Adjustment should be a pure function, meaning it shouldn't throw exceptions. I think we should start with this assumption and move forward from there.. basically in hindsight I think we can get most mileage by taking some of the work @naoman helped with and make it into the core collector library along with your input too.. two for one that way :P

zipkin-server/src/main/java/zipkin/server/internal/ZipkinHttpCollector.java
```diff
@@ -140,7 +147,13 @@
 
   static void error(HttpServerExchange exchange, Throwable e) {
     String message = e.getMessage();
-    int code = message == null || message.startsWith("Cannot store") ? 500 : 400;
+    int code;
+    if (e instanceof FilterActivatedException) {
+      FilterActivatedException filterException = (FilterActivatedException)e;
+      code = filterException.getHttpCode();
```
afalko: @adriancole I tried to find a built-in exception class that carries something like http code, but unfortunately there isn't one that I can bring in without big dependencies. 

These two are the most commonly used: 
1) http://hc.apache.org/httpclient-3.x/apidocs/org/apache/commons/httpclient/HttpException.html
2) https://docs.oracle.com/javase/7/docs/api/javax/xml/ws/http/HTTPException.html

I still think the most reasonable option is to have the custom exception. Without this, a filter implementor cannot adjust the http code emitted if there is an error, which would make for a bad user experience.

afalko: I take part of this back. The latter exception will suffice for the error code and is bolted in java. The only drawback is that it doesn't provide a way to set a custom message.

afalko:
@adriancole I took a closer look at adjuster and I think is too specific for what we are trying to do. I can imagine a filter implementation having that adjuster class internal to it and it would be invoked within the `process` implementation. I'm meeting with @naoman next week to find out more about his work and if my soon-to-be-revised PR will work for his use-cases.

codefromthecrypt:
thanks.

main thing to keep in mind is coupling. there is no reason to assume
collector has a relationship with http. so try to not taint the collector
layer with assumptions about http.

If you want to do something for user experience, try the metrics adapter or
any other integrated test (spring boot test) so we can tell which code is
doing what and change that as need be.

Keep in mind that most will never see the http message bubbled up in an
async reporter. User experience is mostly a dev issue when developing
customizations.


For these scenarios logging or Exception metrics callbacks (normal ones)
can help. You can look at the existing unit tests for the collector where
logging is verified. We did a lot on collector logging and the fact that
there is a new place in lifecycle, adjustment, minimally hints at logging
needed here. For example after decoding and before storage, we can have a
debug message about any adjustment error.

Meanwhile adjustment errors are bugs. Please keep this in mind that
adjusters should absolutely never throw. In other words please dont try to
use exception as flow control logic.

If all things above said, the error message in http needs to be customised,
I would suggest a hook for this, specific to the http collector, probably a
callback. However we need to ensure this happens synchronously as otherwise
you cant report an error anyway. This might be a challenge as I don't
recall where in the lifecycle of http collector we release a 202. If we
want a nice message back to what is the requestor, usually an async lib,
this needs to be kept in mind.


all above basically hints lets keep abstractions clean, not use error as
flow control, which means exceptions in adjusters are bugs, and finally
make the least api for collectors possible.

codefromthecrypt:
from a metrics mapping pov, recall we have spans dropped metrics. if
adjuster blew up, or it decided to drop, this should be incremented. in a
way this impacts the api. for example if the adjuster is a function of one
span (which might be appropriate considering no trace id grouping is
likely), returning null should increment span drop as should any unexpected
error. it should be logged to fine as well.

this isn't to say we cant add more hooks but bear in mind monitoring is
very simple now.. it is messages and spans accepted or dropped. we dont
have any splitting hairs about why from metrics pov.

now consider a site wants to be very specific. they could make a micrometer
decorator which adds metrics on unexpected exception classified by type
etc. for example finagle does similar by showing a hierarchy of exception
types. In other words, you could iron clad beta adjusters by decorating the
function with one that adds new metrics you can look at.

hope these hints help and sorry timing didnt work when I was in town!

afalko:
I've rebased on latest master and removed the custom exception. I also pass `CollectorMetrics` through into the `SpanFilter` interface. I've added a utest that drops spans too.

benjefferies:
This looks like it could be a solution to our current kubernetes problem. We have many environments tracing and all of the traces are stored in one backend. We want to add meta tags such as the environment for all tracing.

It would be nice if https://github.com/afalko/zipkin-generic-enrichment-filter/blob/master/src/main/java/net/afalko/zipkin/filter/GenericEnrichmentFilter.java was bundled as a part of zipkin. It seems to work in the spirit of 12FA.

Any reasons this hasn't progressed?

codefromthecrypt:
hi there are a couple reasons this hasn't progressed. firstly the initial PR has been obviated by a Kafka alternative at Salesforce I think.

the other thing is that there is a glitch not everyone is aware of. at the collector layer we aren't guaranteed to see the whole trace. this means only stateless ops limited to scrubbing or adding tags based only on the current span is possible. to consider the whole trace we need a buffer like Kafka or a tiered storage like voltdb.

if your requirements are limited to things like data filtering or addition (and not pruning layers of the trace or other trace scoped things) indeed this would sort it (as would an elasticsearch pipeline stage)

can you verify that all you need to do are stateless ops that apply to all spans?

benjefferies:
Thanks for getting back to me @adriancole.

I don't think we would require the whole trace as we can search the trace via a tag in any span. So in theory this should be good enough.

We're using stackdriver as a backend so can't leverage any ES pipeline stages.

To give some context we want to add an environment tag for context. This will allow us to know which environment the trace has come from i.e. prod, pre-prod, integration

codefromthecrypt:
@benjefferies I'm inclined to pick this back up with a zipkin2.Span variant of https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/handler/FinishedSpanHandler.java#L30

Naming is a funny one, but I like CollectedSpanHandler at the moment. I would add a lot of javadoc mentioning limitations I said including that a "collected span" isn't necessarily a finished one (partial data can be here, such as a span holding nothing but a single tag). Later, you could imagine a CollectedTraceHandler. regardless, these types would be in the zipkin-collector jar.

If you are willing to champion this by testing things I'm willing to pick this up in a different PR, citing this one and the other prior art. I agree we can push this through.. we just need a user who this is valid for to keep it real.

benjefferies:
@adriancole Sounds good. We can test these changes for sure as it's going to be something we need and feels like the most elegant way of implementing it from our perspective.

afalko:
@benjefferies @adriancole this what we would using at Salesforce when I was there 7 months ago: https://github.com/afalko/zipkin-storage-kafka and https://github.com/afalko/zipkin-storage-filters...it is essentially a set of jars that you add to your Zipkin deploy to intercept spans and do what you will with them. You can change anything about the Spans except http-level things like headers. For those, we had a proxy sitting in front of Zipkin. 

+1 to the FinishedSpanHandler idea. That would simplified things for us.

zipkin-server/src/main/java/zipkin/server/internal/ZipkinHttpCollector.java
shakuzen: The StackOverflow answer is for injecting a single bean. Spring is aware of collection types (after all, it is injecting the list of `SpanFilter` beans, not a single bean of type `List<SpanFilter>`) and will inject an empty collection if no beans of the collection type exist, as far as I can tell.

codefromthecrypt:
will pick this up in https://github.com/openzipkin/zipkin/issues/2802

PR #2075, opened 2018-06-04
Migrates legacy cassandra code to v2 apis

codefromthecrypt:
This is a fairly mammoth change needed to ensure we can stop publishing the v1 libraries and complete the server migration to v2 libraries.

What this does is rebase the cassandra v1 (thrift based) schema over the api style and dependencies of the cassandra v3 library.

This preserves all functionality it had before, including request deduplication. It is just using different base libraries to accomplish the same.

FYI, I don't expect folks to review this, but feel free if you like!

cc @openzipkin/cassandra


zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/v1/CassandraStorage.java
```diff
@@ -119,18 +123,15 @@ public Builder maxConnections(int maxConnections) {
     }
 
     /**
-     * Ensures that schema exists, if enabled tries to execute script io.zipkin:zipkin-cassandra-core/cassandra-schema-cql3.txt.
-     * Defaults to true.
+     * Ensures that schema exists, if enabled tries to execute script
+     * io.zipkin:zipkin-cassandra-core/cassandra-schema-cql3.txt. Defaults to true.
```
codefromthecrypt, line 127: wow this javadoc is a bit out of date :P

PR #2076, opened 2018-06-06
Migrates legacy mysql code to v2 apis

codefromthecrypt:
This is a fairly mammoth change needed to ensure we can stop publishing the v1 libraries and complete the server migration to v2 libraries.

What this does is rebase the mysql v1 (thrift based) schema over the `zipkin2` package, via converters.

See #2047


codefromthecrypt:
Note: this requires a little more work, notably to use the provided executorService

PR #2089, opened 2018-06-11
Attempts to compile with JDK 10

codefromthecrypt:
Note: there are some warnings like this using JDK 10.0.1 on osx

```
[WARNING] /Users/acole/oss/zipkin/zipkin-autoconfigure/collector-scribe/src/main/java/zipkin/autoconfigure/collector/scribe/ZipkinScribeCollectorAutoConfiguration.java:[-1,-1] /modules/java.base/java/io/InputStream.class: major version 54 is newer than 53, the highest major version supported by this compiler.
  It is recommended that the compiler be upgraded.

```

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -15,8 +15,8 @@
 # NOTE: Travis tests Elasticsearch and MySQL using Docker, and Kafka manually. We don't redundantly test them here
 machine:
   environment:
-    # Use java 9 here and 1.8 in travis (so that we test both)
-    JAVA_HOME: /usr/lib/jvm/zulu-9-amd64
+    # Use java 10 here and 1.8 in travis (so that we test both)
+    JAVA_HOME: /usr/lib/jvm/zulu-10-amd64
     # Quiet Maven invoker logs (Downloading... when running zipkin-server/src/it)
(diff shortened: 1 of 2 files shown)


shakuzen (approved):

codefromthecrypt:
There's more work to complete this as mockito isn't happy

```
org.mockito.exceptions.base.MockitoException: 

Mockito cannot mock this class: interface zipkin.storage.StorageAdapters$SpanConsumer.

Mockito can only mock non-private & non-final classes.
If you're not sure why you're getting this error, please report to the mailing list.


Java               : 10
JVM vendor name    : Azul Systems, Inc.
JVM vendor version : 10.0.1+9
JVM name           : OpenJDK 64-Bit Server VM
JVM version        : 10.0.1+9
JVM info           : mixed mode
OS name            : Linux
OS version         : 3.13.0-149-generic


Underlying exception : java.lang.UnsupportedOperationException: Cannot define class using reflection
Caused by: java.lang.UnsupportedOperationException: Cannot define class using reflection
Caused by: java.lang.IllegalArgumentException: Unknown Java version: 10
```

codefromthecrypt:
obviated as we now compile with JDK 11

PR #2101, opened 2018-06-22
Fixes accidental ID zero when parsing empty strings

codefromthecrypt:
Glitches noticed when updating Wingtips, which has tests for these.


codefromthecrypt:
before a traceid of "" would end up coercing to a string of zeros, which is invalid

PR #2104, opened 2018-06-25
Completes javascript port of v2 -> v1 span conversion

codefromthecrypt:
There are a number of edge cases that our Java code handles which the
Javascript does not. This will complete the port as a part of migrating
the UI to use v2 endpoints.

See #2047


codefromthecrypt:
I've actually ported all the code, but there are a bunch of tests I need to port, yet

codefromthecrypt:
done

Issue #1744, opened 2017-09-22
UI should skip self-referencing trace


codefromthecrypt:
the following trace will break the UI (white screen)

```
Uncaught RangeError: Maximum call stack size exceeded
    at u (traceToMustache.js:34)
    at traceToMustache.js:37
    at Array.map (<anonymous>)
    at u (traceToMustache.js:37)
    at traceToMustache.js:37
    at Array.map (<anonymous>)
    at u (traceToMustache.js:37)
    at traceToMustache.js:37
    at Array.map (<anonymous>)
    at u (traceToMustache.js:37)
```

```
curl -s localhost:9411/api/v2/spans -H 'Content-Type: application/json' -d '[
  {
    "traceId": "ade3bd64985e5693",
    "parentId": "ade3bd64985e5693",
    "id": "ade3bd64985e5693",
    "name": "offer",
    "timestamp": 1506066778476000,
    "duration": 5,
    "localEndpoint": {
      "serviceName": "offerservice",
      "ipv4": "10.139.144.225",
      "port": 8080
    },
    "annotations": [
      {
        "timestamp": 1506066778476000,
        "value": "kafka produce"
      }
    ],
    "tags": {
      "lc": "unknown"
    }
  },
  {
    "traceId": "ade3bd64985e5693",
    "parentId": "ade3bd64985e5693",
    "id": "ade3bd64985e5693",
    "name": "offer",
    "timestamp": 1506066778568000,
    "duration": 78,
    "localEndpoint": {
      "serviceName": "deliverystatus",
      "ipv4": "10.138.144.243",
      "port": 8080
    },
    "annotations": [
      {
        "timestamp": 1506066778568000,
        "value": "kafka consume"
      }
    ],
    "tags": {
      "lc": "unknown"
    }
  },
  {
    "traceId": "ade3bd64985e5693",
    "parentId": "ade3bd64985e5693",
    "id": "ade3bd64985e5693",
    "name": "offer",
    "timestamp": 1506066778624000,
    "duration": 76,
    "localEndpoint": {
      "serviceName": "offerhistory",
      "ipv4": "10.139.144.194",
      "port": 8080
    },
    "annotations": [
      {
        "timestamp": 1506066778624000,
        "value": "kafka consume"
      }
    ],
    "tags": {
      "lc": "unknown"
    }
  },
  {
    "traceId": "ade3bd64985e5693",
    "parentId": "ade3bd64985e5693",
    "id": "ade3bd64985e5693",
[...]

PR #2109, opened 2018-06-28
Deletes self-referencing spans in conversion javascript

codefromthecrypt:
See #1744


codefromthecrypt:
with this change, the javascript doesn't crash anymore.. even if the actual data from #1744 is quite wrong due to 3 spans having constant ids

screen shot 2018-06-28 at 4 50 26 pm

PR #2112, opened 2018-06-29
Ports Node (used by ClockSkew) to javascript

codefromthecrypt:
See #2047


llinder (approved):
Not an area of the code I have looked at much.  That said the implementation looks clean and understandable.

PR #2115, opened 2018-07-02
Initial pass for adding multiple trace lookup

llinder:


llinder:
Moving this to a separate branch https://github.com/openzipkin/zipkin/tree/multifetch

Issue #1206, opened 2016-07-29
Search by dependency link


codefromthecrypt:
One thing @yurishkuro mentioned in his presentation was that we don't support search by multiple services out of the box. This is interesting, because we do actually support search without a service.

The use-case is moving from a link point of view (ex an edge on a dependency tree) to a trace query.

We can investigate how search by service works on a per-datastore basis, as well the UI affects of such a thing.

calm2016:
In microservice applications, searching among all services is really useful and required. Please consider to at least allow "all" keywords to indicate searching in all services.
Btw, it would be great if simple regular expression is supported, such as xx*\* in the annotation/key-value search function. Thanks

codefromthecrypt:
currently the api supports leaving out the serviceName parameter.
http://zipkin.io/zipkin-api/#/paths/%252Ftraces/get/parameters/serviceName

regex isn't likely to work across all storage options server-side

codefromthecrypt:
all service query is out in 2.3

yurishkuro:
@adriancole does this work in either of the Cassandra implementations? Looks like it might in v2 because of `zipkin2.span (annotation_query)` and `zipkin2.trace_by_service_span (duration)` SASI indices, but v1 uses service name as a PK field iirc.

codefromthecrypt:
>
> @adriancole <https://github.com/adriancole> does this work in either of
> the Cassandra implementations? Looks like it might in v2 because of zipkin2.span
> (annotation_query) and zipkin2.trace_by_service_span (duration) SASI
> indices, but v1 uses service name as a PK field iirc.
>
It works (everything works) in "cassandra3". In "cassandra" it is expensive
as it does a fan-out across services.

yurishkuro:
ah, fan-out, nice trick. Not an option for us, unfortunately (3k services).

thanks.

codefromthecrypt:
> ah, fan-out, nice trick. Not an option for us, unfortunately (3k services).

I'll raise a PR to minimize the impact of this for those using
cassandra and not yet upgraded.. If someone is just clicking search, a
cheaper way is to scan the traces table until you've collected enough
trace ids.

codefromthecrypt:
actually, we can't do this in the legacy impl because we don't have
any secondary index on the traces table (because we need to support
2.2+ which has no SASI)

Those who use cassandra and want more efficient all-services query
will need to upgrade I think... cc @openzipkin/cassandra in case I'm
mistaken

[...]

PR #2116, opened 2018-07-02
Adds trace IDs to dependency links

codefromthecrypt:
TODO: make this conditional, as existing dependency linker jobs won't
need IDs and keeping them might make spark jobs move more data than
they currently need.

See #1206


zeagord (approved):

codefromthecrypt:
ps not merging intentionally until we get some experience with the
experiment. reason is that this is api breaking

codefromthecrypt:
for those wondering.. pausing while we investigate what else is out there (ex haystack has some aggregation tools as well.. not sure if there is any overlap until we ask)

codefromthecrypt:
I think we can pick this back up in javascript after #2731

codefromthecrypt:
moved most code out in this https://github.com/openzipkin/zipkin/pull/2803

codefromthecrypt:
moved to branch attic/links-with-traceIds

PR #2123, opened 2018-07-06
Sorts api results in the UI

codefromthecrypt:
reported by @drolando


codefromthecrypt:
FYI we have zero tests in this area of the codebase. I will sort (pun intended) that later. For now, tested manually.

PR #2127, opened 2018-07-10
Ports dependency linking tests to v2 model and corrects MySQL impl

codefromthecrypt:
While porting dependency linking tests, noticed a small bug and
corrected the MySQL impl.

See #2047

### DIFF
--- a/zipkin2/src/test/java/zipkin2/storage/ITDependencies.java
+++ b/zipkin2/src/test/java/zipkin2/storage/ITDependencies.java
@@ -1,0 +1,701 @@
+/*
+ * Copyright 2015-2018 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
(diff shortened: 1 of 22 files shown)


zipkin2/src/test/java/zipkin2/storage/ITDependencies.java
```diff
@@ -0,0 +1,701 @@
...
+
+/**
+ * Base test for {@link SpanStore} implementations that support dependency aggregation. Subtypes
+ * should create a connection to a real backend, even if that backend is in-process.
+ *
+ * <p>This is a replacement for {@code zipkin.storage.DependenciesTest}. There is some redundancy
+ * as {@link DependencyLinkerTest} also defines many of these tests. The redundancy helps ensure
+ * integrated storage doesn't fail due to mismapping of data, for example.
+ */
+public abstract class ITDependencies {
```
codefromthecrypt, line 55: This is the ported test suite

PR #2128, opened 2018-07-11
Adds UI for Links with trace ids

zeagord:
Adds the UI, service endpoint and the integration

TODO: UI integration tests which will done when we actually implement this feature.

### DIFF
--- a/zipkin2/src/main/java/zipkin2/storage/QueryRequest.java
+++ b/zipkin2/src/main/java/zipkin2/storage/QueryRequest.java
@@ -98,6 +98,22 @@ public final class QueryRequest {
     return limit;
   }
 
+  /** Parent service name to fetch for the link dependency */
+  @Nullable
+  public String parentServiceName(){
+    return parentServiceName;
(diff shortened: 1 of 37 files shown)


drolando:
Is there a related issue I can read to understand what you're trying to do?

zeagord:
This is related to #1206

codefromthecrypt:
wanna pop in some screen shots?

zeagord:
Here's the screen capture of the feature.
qch7pkts9s

zipkin2/src/main/java/zipkin2/storage/QueryRequest.java
```diff
@@ -98,6 +98,22 @@ public int limit() {
     return limit;
   }
 
+  /** Parent service name to fetch for the link dependency */
+  @Nullable
+  public String parentServiceName(){
```
llinder, line 103: From what I recall we decided to use a separate query request for this called `DependencyQueryRequest`.

PR #2131, opened 2018-07-12
Removes v1 codebase

### LINKED ISSUE #2047: Get rid of the v1 codebase

codefromthecrypt:
Fixes #2047


zipkin-server/pom.xml
```diff
@@ -18,11 +18,14 @@
...
 
   <parent>
-    <groupId>io.zipkin.java</groupId>
+    <groupId>io.zipkin.zipkin2</groupId>
     <artifactId>parent</artifactId>
     <version>2.10.2-SNAPSHOT</version>
   </parent>
 
+  <!-- Group ID left alone due to folks downloading it with curl from maven central.
+       TODO: setup a maven redirect -->
```
devinsba: An issue should probably be created for this before merge

zipkin-collector/rabbitmq/src/test/java/zipkin2/collector/rabbitmq/ITRabbitMQCollector.java
```diff
@@ -117,12 +117,12 @@ public void skipsMalformedData() throws Exception {
 
   /** Guards against errors that leak from storage, such as InvalidQueryException */
   @Test
-  public void skipsOnSpanConsumerException() throws Exception {
+  public void skipsOnSpanConsumerException() {
     // TODO: reimplement
   }
 
   @Test
-  public void messagesDistributedAcrossMultipleThreadsSuccessfully() throws Exception {
```
devinsba, line 125: What changed to cause this?

devinsba (approved):

zipkin-collector/rabbitmq/src/test/java/zipkin2/collector/rabbitmq/ITRabbitMQCollector.java
codefromthecrypt: sorry this was intellij cleanup

zipkin-server/pom.xml
codefromthecrypt: https://github.com/openzipkin/zipkin/issues/2132 thanks for the reminder!

PR #2133, opened 2018-07-12
Adds some configuration to reduce work at server startup

codefromthecrypt:
More interesting things can be found here: https://github.com/dsyer/spring-boot-allocations

Thx @dsyer

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -247,6 +247,10 @@
         <version>${spring-boot.version}</version>
         <exclusions>
           <exclusion>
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
In docker, best time on normal startup before this change:
Started ZipkinServer in 5.258 seconds (JVM running for 6.475)

In docker, best time on normal startup after this change:
Started ZipkinServer in 4.995 seconds (JVM running for 6.16)

PR #2142, opened 2018-07-19
Fixes UI display of json annotation values caused by opentracing log messages

anandshah05:
UI Panel goes out of screen for json values.

Same issue in log messages (normal annotation) as given in this issue https://github.com/openzipkin/zipkin/issues/1113

Request you to merge to the master branch

### DIFF
--- a/zipkin-ui/js/component_ui/spanPanel.js
+++ b/zipkin-ui/js/component_ui/spanPanel.js
@@ -44,10 +44,12 @@ export function maybeMarkTransientError(row, anno) {
 export function formatAnnotationValue(value) {
   const type = $.type(value);
   if (type === 'object' || type === 'array' || value == null) {
-    return escapeHtml(JSON.stringify(value));
-  } else {
-    return escapeHtml(value.toString()); // prevents false from coercing to empty!
+    return `<pre><code>${escapeHtml(JSON.stringify(value, null, 2))}</code></pre>`;
   }
+  const result = value.toString();
+  // Preformat if the text includes newlines
+  return result.indexOf('\n') === -1 ? escapeHtml(result)
+    : `<pre><code>${escapeHtml(result)}</code></pre>`;
 }
 
 // Binary annotations are tags, and sometimes the values are large, for example
--- a/zipkin-ui/js/component_ui/spanPanel.js
+++ b/zipkin-ui/js/component_ui/spanPanel.js
@@ -38,18 +38,15 @@ export function maybeMarkTransientError(row, anno) {
 // end up becoming javascript objects later. For this reason, we have to guard
 // and stringify as necessary.
 
-// annotations are named events which shouldn't hold json. If someone passed
-// json, format as a single line. That way the rows corresponding to timestamps
-// aren't disrupted.
+// annotations are named events which can hold json. If someone passed
+// json, wrap it with <pre><code> to display it properly on UI.
+// That way the rows corresponding to timestamps aren't disrupted.
 export function formatAnnotationValue(value) {
   const type = $.type(value);
   if (type === 'object' || type === 'array' || value == null) {
     return `<pre><code>${escapeHtml(JSON.stringify(value, null, 2))}</code></pre>`;
   }
-  const result = value.toString();
-  // Preformat if the text includes newlines
-  return result.indexOf('\n') === -1 ? escapeHtml(result)
-    : `<pre><code>${escapeHtml(result)}</code></pre>`;
+  return escapeHtml(value.toString());
 }
 
 // Binary annotations are tags, and sometimes the values are large, for example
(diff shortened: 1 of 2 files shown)


zipkin-ui/test/component_ui/spanPanel.test.js
```diff
@@ -59,18 +59,18 @@ describe('formatAnnotationValue', () => {
...
   it('should format array as one-line json', () => {
     formatAnnotationValue([{foo: 'bar'}, {baz: 'qux'}]).should.equal(
-      '[{&quot;foo&quot;:&quot;bar&quot;},{&quot;baz&quot;:&quot;qux&quot;}]'
+      '<pre><code>[{&quot;foo&quot;:&quot;bar&quot;},{&quot;baz&quot;:&quot;qux&quot;}]</code></pre>'
     );
   });
 
   it('should format null as json', () => {
-    formatAnnotationValue(null).should.equal('null');
+    formatAnnotationValue(null).should.equal('<pre><code>null</code></pre>');
```
codefromthecrypt, line 74: is this required? I mean to preformat and code the word null.. seems like an accidental change.

zipkin-ui/js/component_ui/spanPanel.js
```diff
@@ -44,10 +44,12 @@ export function maybeMarkTransientError(row, anno) {
...
   const type = $.type(value);
   if (type === 'object' || type === 'array' || value == null) {
-    return escapeHtml(JSON.stringify(value));
-  } else {
-    return escapeHtml(value.toString()); // prevents false from coercing to empty!
+    return `<pre><code>${escapeHtml(JSON.stringify(value, null, 2))}</code></pre>`;
   }
+  const result = value.toString();
+  // Preformat if the text includes newlines
+  return result.indexOf('\n') === -1 ? escapeHtml(result)
```
codefromthecrypt: unit test both with and without newline to ensure you've done what you want (and it stays like that). Or, remove the special casing of newlines, as annotations should certainly not have newlines

codefromthecrypt:
regardless of whether this is merged, please let us know what instrumentation caused this as annotations are not intended to be used as logs or include newlines. When we add workarounds, it is a reasonable request to ask why, so in case whatever we are working around is fixed, the workaround can be considered for removal

zipkin-ui/test/component_ui/spanPanel.test.js
anandshah05: wrapped null with <pre><code> to fix unit tests

zipkin-ui/js/component_ui/spanPanel.js
anandshah05: I actually copied from the formatBinaryAnnotationValue func, which works fine

anandshah05:
I understand, annotation is meant for one line only, but appending full request/response in log helps in tracing the message. Changes are minor and has no impact on any other functionality.

This is to fix the UI Display when there is json message in log, attaching UI screenshots

before fix

before-fix

after fix

after-fix

zipkin-ui/js/component_ui/spanPanel.js
anandshah05: I agree with your point, let me delete these unnecessary code

anandshah05:
Please verify latest changes and merge

codefromthecrypt:
full request and response is better for a tag than annotation. can you try that instead?

each strange thing we support makes changing UI later more difficult. I will close this until there are more people asking to do what you are.

codefromthecrypt:
ps even if we have requests to make zipkin a logging system, due to opentracing for example, we cannot afford that responsibility and it is an unfair burden. usually logging ends up with new problems in search, bulk and indexing.

you can see issues in jaeger for example. eventhough they initially encouraged tracing as logging and have a lot of staff, they have many issues including search indexing and volume and are even looking at rearchitecting to collect logs from elsewhere. 
https://github.com/jaegertracing/jaeger/search?p=2&q=logging&type=Issues

many libraries have the ability to correlate with trace and span IDs and you can configure the UI to use a search expression link to get to logs. zipkin basically is used to integrate with logging and metrics not be a one stop shop.

Ex here are the docs on integration which can be improved as well.

https://github.com/openzipkin/zipkin/blob/master/zipkin-ui/README.md#how-do-i-find-logs-associated-with-a-particular-trace

anandshah05:
Thanks Adrian for the explaination

anandshah05:
I agree with your point of zipkin is not a logging system. The fix which in this PR is just a minor UI display fix, where dialog box goes out of screen. It has no functional impact. Look at the UI screenshot attached above.

I think we should add this fix to mainline, request you to re-think about it.

codefromthecrypt:
You don't have a history of helping in the UI, so it isn't yet a good time
for you to assess the maintenance impact of this. That said, I will
reconsider once others ask for the same thing.

Meanwhile, if you must do what you are doing, tag instead of annotation.

anandshah05:
Since we use tag for searching, we might be creating index on tag at backend database. So using tag for a json message will have performance impact and is not helpful

codefromthecrypt:
>
> Since we use tag for searching, we might be creating index on tag at
> backend database. So using tag for a json message will have performance
> impact and is not helpful
>
if you are using zipkin storage I think most if not all limit to first 255
characters. Depending on storage impl there is also some questions about
which tags to index. Nevertheless tags are what are used for key/value and
where the value is potentially escaped json

PR #2148, opened 2018-07-24
remove es aggregation,improve es search performance

zwukong:
i have two account,sorry for the bother,please @ JarVzhao ~

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchSpanStore.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchSpanStore.java
@@ -88,7 +88,6 @@ final class ElasticsearchSpanStore implements SpanStore {
 
     SearchRequest esRequest =
         SearchRequest.create(indices).filters(filters);
-
     Call<List<Span>> spanResult= search.newCall(esRequest, BodyConverters.SPANS);
     //convert to List<List<Span>>,each list has only one span
     Call<List<List<Span>>> result =spanResult.map(spanList->{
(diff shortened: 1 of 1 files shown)


zwukong:
i have tested codes locally, if you have any questions,please @ JarVzhao ~

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchSpanStore.java
```diff
@@ -63,6 +60,8 @@
     long beginMillis = Math.max(endMillis - request.lookback(), EARLIEST_MS);
 
     SearchRequest.Filters filters = new SearchRequest.Filters();
+    //only search for SERVER span,which is simpler and faster than aggregation
+    filters.addTerm("kind", "SERVER");
```
codefromthecrypt, line 64: we cannot assume this.. there are spans that are messaging in nature (PRODUCER/CONSUMER). also, RPC spans themselves are optional..

codefromthecrypt (commented):
Thanks for the spike.. I'm quite surprised this passed for you... maybe you don't have docker installed?

Tests like https://github.com/openzipkin/zipkin/blob/master/zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ITElasticsearchStorageV6.java will need to pass. If you can't run docker, you it should look for localhost elasticsearch...

codefromthecrypt:
right, so I'm sorry the tests will "ignore" if elasticsearch isn't running and docker is not installed. You can see the failure here https://travis-ci.org/openzipkin/zipkin/builds/407517059

First thing is to run the tests then you will know what works or not with this approach. Thanks again for giving a try.

zwukong:
sorry,i forgot to run the unit tests. i use a local es,spans are from sleuth, the ui query is ok.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchSpanStore.java
zwukong: i see,in our company,we use different queries ,for example in messaging circumstance,the kind may be messaging.UI splits them,not combine them all together

zwukong: aggregation is very slow in es especially in large indices

codefromthecrypt:
are you planning to continue this?

codefromthecrypt:
closing until effort resumes

PR #2150, opened 2018-07-26
Makes java.util.logging route to log4j2 by default

### LINKED ISSUE #2149: Make debug logging work with zipkin-server
I may be wrong, but there may be a packaging/config problem preventing JUL logs (like from the collector) from showing up even when DEBUG is available. This could possibly be a regression if verified.

Ex. the following bad request should show up in console output when the server has `--logging.level.zipkin2=DEBUG`

```
$ curl -X POST localhost:9411/api/v2/spans -H'Content-Type: 
application/json' -d '[ "mary", "had", "a", "little", "lamb"]'
Cannot decode spans
```

seems https://docs.spring.io/spring-boot/docs/current/reference/html/howto-logging.html#howto-configure-log4j-for-logging says to use the log manager, but I don't think that works by default using java -jar

```bash
$ java -Djava.util.logging.manager=org.apache.logging.log4j.jul.LogManager -jar zipkin.jar --logging.level.zipkin2=TRACE
Could not load Logmanager "org.apache.logging.log4j.jul.LogManager"
java.lang.ClassNotFoundException: org.apache.logging.log4j.jul.LogManager
```

codefromthecrypt:
This fixes the inconvenience of having to specify it in various CLIs.

Fixes #2149

### DIFF
--- a/zipkin-server/src/main/java/zipkin/server/ZipkinServer.java
+++ b/zipkin-server/src/main/java/zipkin/server/ZipkinServer.java
@@ -21,6 +21,11 @@ import zipkin2.server.internal.RegisterZipkinHealthIndicators;
 @SpringBootApplication
 @EnableZipkinServer
 public class ZipkinServer {
+  static {
+    // Make sure java.util.logging goes to log4j2
+    // https://docs.spring.io/spring-boot/docs/current/reference/html/howto-logging.html#howto-configure-log4j-for-logging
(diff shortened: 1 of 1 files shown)


zeagord (approved):

codefromthecrypt:
why.. travis.. why are you broke!

shakuzen (approved):

PR #2156, opened 2018-08-03
Updates javascript dependencies in efforts to kill security warning

codefromthecrypt:
node-growl is flagged, and indirectly required by some test infra.

To squash that implied a rediculous amount of dependency updates :P


codefromthecrypt:
double-checked minification. looks a bit better than before due to things inherited by moving up a bit:

```
1784	./zipkin-ui/target/zipkin-ui-2.10.5-SNAPSHOT.jar
2112	/Users/acole/.m2/repository/io/zipkin/java/zipkin-ui/2.10.4-SNAPSHOT/zipkin-ui-2.10.4-SNAPSHOT.jar
```

PR #2157, opened 2018-08-03
Bump Spring-Boot and Micrometer

oscerd:
I noticed that we would like to be in sync with SB stuff, Jooq would need to stay at 3.10.5 instead of 3.11.2 then


codefromthecrypt:
cool you are fast to notice!

codefromthecrypt:
Thanks andrea!

oscerd:
My pleasure, just minor things :-)

PR #2158, opened 2018-08-04
Accept a Cassandra running 3.11.3 as well as later

nollbit:
The current code in `Schema.java` seems to only accept servers running a later version than C* 3.11.3 (i.e. 3.11.4). Since 3.11.3 is recently released, I'm assuming that the code is meant to accept that version and thus this modification is needed.

### DIFF
--- a/zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/Schema.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/Schema.java
@@ -101,7 +101,7 @@ final class Schema {
   static KeyspaceMetadata ensureExists(String keyspace, boolean searchEnabled, Session session) {
     session.getCluster().getMetadata().getAllHosts().forEach((host) -> {
       Preconditions.checkState(
-              0 < VersionNumber.parse("3.11.3").compareTo(host.getCassandraVersion()),
+              0 <= VersionNumber.parse("3.11.3").compareTo(host.getCassandraVersion()),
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
absolutely sensible. thanks!

codefromthecrypt:
PS I will backfill a unit test for this type pre-release, but not now as I am readying for airport :P

codefromthecrypt:
I think a typo in integration tests existed, which means maybe this actually is a bug that prevents 3.11.3 from working indeed. Is that the case? If so, we should patch immediately

https://github.com/openzipkin/zipkin/pull/2159

zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/Schema.java
```diff
@@ -101,7 +101,7 @@ static KeyspaceMetadata getKeyspaceMetadata(Session session) {
   static KeyspaceMetadata ensureExists(String keyspace, boolean searchEnabled, Session session) {
     session.getCluster().getMetadata().getAllHosts().forEach((host) -> {
       Preconditions.checkState(
-              0 < VersionNumber.parse("3.11.3").compareTo(host.getCassandraVersion()),
+              0 <= VersionNumber.parse("3.11.3").compareTo(host.getCassandraVersion()),
```
codefromthecrypt, line 104: not terribly intuitive, but this should have been >=

codefromthecrypt:
ok please update to 2.11.1 Thanks again!

PR #2160, opened 2018-08-05
Fixes cassandra version check and backfills tests

codefromthecrypt:
See #2158

### DIFF
--- a/zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/Schema.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/Schema.java
@@ -24,7 +24,6 @@ import com.datastax.driver.core.TypeCodec;
 import com.datastax.driver.core.VersionNumber;
 import com.datastax.driver.core.exceptions.InvalidTypeException;
 import com.datastax.driver.mapping.annotations.UDT;
-import com.google.common.base.Preconditions;
 import com.google.common.io.CharStreams;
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
cc @nollbit @drolando @michaelsembwever 

bug could happen to anyone.. hard to look at comparisons like this. Anyway tests backfilled now and 2.11.1 should be cut immediately cc @openzipkin/cassandra

zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/Schema.java
```diff
@@ -84,27 +85,30 @@ static Metadata readMetadata(Session session) {
...
     }
     return keyspaceMetadata;
   }
 
+  static KeyspaceMetadata getKeyspaceMetadata(Session session, String keyspace) {
+    Cluster cluster = session.getCluster();
+    com.datastax.driver.core.Metadata metadata = cluster.getMetadata();
+    for (Host host : metadata.getAllHosts()) {
+      checkState(
+        0 >= VersionNumber.parse("3.11.3").compareTo(host.getCassandraVersion()),
```
codefromthecrypt, line 104: the last patch worked, but wouldn't allow newer versions :P

codefromthecrypt:
FYI credit for this goes to @drolando who somehow was able to think through the unit tests and notice the problem :P Now, we have unit tests so less thinking needed later

PR #2162, opened 2018-08-06
Makes build work with JDK 10

codefromthecrypt:


codefromthecrypt:
will look at the failure later

PR #2166, opened 2018-08-10
Block on http calls to elastic search

malonso1976:


zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
```diff
@@ -147,7 +156,11 @@ public void acceptSpans(
   }
 
   void record(List<Span> sampled, Callback<Void> callback) {
-    storage.spanConsumer().accept(sampled).enqueue(callback);
+    if(blockOnStorage) {
```
codefromthecrypt: I would prefer the logic be handled here without changing the Call api. What if you just did spanConsumer.execute() instead in a try/finally where after or on exception you call the callback

codefromthecrypt (commented):
Thanks for starting this.. made a code note for now. once this is closer I'll sweep for where docs should be updated.

zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
malonso1976: Ok, I understand. But will not be able to lock on the HttpCall semaphore, because its is not available  for this use.

Some other contention mechanism must be develop here when storage begins to drop spans

codefromthecrypt: what if we move the semaphore here?

codefromthecrypt: actually we don't need the semaphore right? because in the case of kafka and rabbit (the only valid pull consumers) concurrency is controlled at that abstraction, and blocking will limit concurrency to stream count..

I suppose you are asking how to expose how to block *more* ex on a certain exception?

malonso1976: Hi Adrian,
  when using zipking with a Kafka collector and heavy load of spans, we tried with several storage alternatives. Mysql does not scale as well as others so it was a quick discard. Cassandra could not handle so many Future calls and began throwing driver errors, no matter how we tuned it (it was zipkin 2.7.1). Elastic did the job, but began receiving exceptions like this:

```
10:13:25.604 [pool-1-thread-1] DEBUG z.collector.kafka10.KafkaCollector - Cannot store spans [1420afe62427f4ce559a0e797bb1b777/13468e1e3b2d9e26] due to IllegalStateException(over capacity)
java.lang.IllegalStateException: over capacity
	at zipkin2.elasticsearch.internal.client.HttpCall.enqueue(HttpCall.java:87)
	at zipkin.internal.V2Collector.record(V2Collector.java:57)
	at zipkin.internal.Collector.accept(Collector.java:79)
	at zipkin.collector.Collector.accept(Collector.java:145)
	at zipkin.collector.kafka10.KafkaCollectorWorker.run(KafkaCollectorWorker.java:84)
	at zipkin.collector.kafka10.KafkaCollector$LazyKafkaWorkers$1.run(KafkaCollector.java:196)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```
The max-connections parameter determined the semaphore count:

https://github.com/openzipkin/zipkin/blob/master/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java#L77

This count must be shared between query calls and collector insertion calls. In order to guarantee that insertion calls wait to be executed to avoid dropping spans, I suggested to change the semaphore.tryAcquire to semaphore.acquire, and that fixed the case. 

Our problem now is that the HttpCall is shared for the mixed use of query api operations and collector api insertions. There is no chance for reservation of the same amount of HttpCalls as Kafka worker threads: KAFKA_STREAMS parameter.

Alternatives:
- Split pool of HttpCalls used for collector processing from those used for query prrocessing
- Implement a retry behaviour when java.lang.IllegalStateException or RuntimeException is caught 

```
void record(List<Span> sampled, Callback<Void> callback) {
    if(blockOnStorage) {
      try {
        storage.spanConsumer().accept(sampled).execute();
        callback.onSuccess(null);
      } catch (IOException | RuntimeException | Error e) {
        // TODO: retry on error
        callback.onError(e);
      }
    } else {
      storage.spanConsumer().accept(sampled).enqueue(callback);
    }
  }
```

 What do you think? Retry behaviour requires additional parameters such as maximum number of retries and wait interval between retries...

codefromthecrypt:
I think most who are in a high load scenario dont try to run only one
instance of zipkin. if you split your collector deployment from your
ui/read, the problem becomes simpler doesn't it?

malonso1976:
Yes, we are running 2 instances of zipkin. One for the querying api/ui and another one for collector processing. 
Do you still find it sensible to work on this pull request?

malonso1976:
One more comment: Kafka collector can use as many thread workers (stream parameter) as topic's partitions, RabbitMQ can use an unbounded number of workers. In order to scale up Kafka there is a limitation regarding the number of consumers.

codefromthecrypt:
The thing I'm trying to focus here on is solving at the right layer of abstraction. Usually folks never touch this code and don't want it to become complex.  I do understand there is some finesses here.

## why isolate read traffic? it makes concurrency easier and without multiple settings
So, my question about the query vs collector was this comment:

https://github.com/openzipkin/zipkin/pull/2166#discussion_r209559144
> Our problem now is that the HttpCall is shared for the mixed use of query api operations and collector api insertions. There is no chance for reservation of the same amount of HttpCalls as Kafka worker threads: KAFKA_STREAMS parameter.

So, what I was getting at was that if there was a semaphore at the collector level, or some bounding by stream count, or a combination.. if only collector traffic went to that instance (via QUERY_ENABLED=false), wouldn't we be able to simplify because we don't need to concern with read traffic to the HttpCall api?

## Should we continue this PR?
You asked about continuing this PR. As mentioned many times, I will not merge tight coupling between http call and the collector tier. Please stop proposing this. If that's your goal, maybe you should put this change down and wait for someone else to pick it up as reviewing is taking longer than actually doing the change. There are plenty of options and tight coupling between storage impl and collector impl is the least favorable for maintenance.

## Let's review prior art: SQS
For example, in SQS exceptions except IllegalArgumentException are implicitly retried (due to lack of a follow-up delete message) https://github.com/openzipkin/zipkin-aws/blob/master/collector-sqs/src/main/java/zipkin2/collector/sqs/SQSSpanProcessor.java#L130 They also aren't affecting storage layer, rather assuming IllegalArgumentException is not retryable. Of course, this retry policy might not be awesome, but it is easier to reason about and a function to select exceptions would be straight forward to discuss (especially if separate). SQS also has a backoff mechanism which is simple in nature. What I like very much is that SQS did not ask to change the storage layer or borrow a storage specific semaphore to do its backoff etc. That made the SQS code less scary to me: that the retry semantics were isolated.

## Let's review options!
This like concurrency gates have options, and zipkin-server is made with spring. You can use AOP and all sorts of things to weave in advanced things as a means to experiment. Let's not forget this as it is better than committing experiments or code that only one site needs. For example, I don't think that most would find it optimal to spend a lot of time on a site who wants to handle a lot of load with only one collector node and also limited storage capacity. Focusing on how that site can use other tools can also be possible.. ex answering the spring question of how to weave in other libraries like https://github.com/Netflix/concurrency-limits

You can also make a custom collector! For example, if you are convinced of one approach, you can package your own collector the same way that kafka08 does (module). It is actually a good idea to try and test things, and if you need back/forth style advice on this, gitter is best to do that: https://gitter.im/openzipkin/zipkin

## OK back to work.. what are we trying to do?

The most discussed part of this change is that a polling collector is in a loop, and there's no value to pulling off spans you can't store! block on storage will definitely slow the amount of requests going to the backend with minimal impact when polling.  For example, the polling loop will then pause until it knows there is an error or not. 

Our goal was to make some cheap change, like cheap to make cheap to review, isolated.

Here are some thoughts, but overall use tests to drive your ideas. without tests it is really hard to reason with the full set of options especially when one of the reviewers is asleep!

### option 1:  polling collector ~= limiting the concurrent with synchronous api
For example, in elasticsearch this isn't waiting until indexing is done iirc, this is only waiting to know if it was accepted. IOTW, we are blocking until we know there is an error then proceeding. Blocking on storage can be implemented with semaphores and things (possible change 2), or it can be simply using the synchronous api. IMHO, the latter is easier to reason with especially in a loop. @llinder @devinsba or others have feedback on this.

### option 2: polling collector ~= limiting the concurrent requests with semaphores
if you do block on storage (change 1), yes there is a limit to how many concurrent requests can be in-flight as no more than the concurrency limit for queue processors. In Kafka, yes it would be stream count, in Rabbit the channel count. If the limit should be higher than queue count, this means you can't use the synchronous api (execute), rather a collector tier semaphore. If proposed, I don't personally even want to review this without tests.

### separate concern: on exception re-process spans
In Kafka, we could already implement a means to resubmit work similar to SQS. We can do this with or without a collector-tier change. For example, and especially if in a blocking loop, we can hold a reference to the initial spans and retry them on a condition, whether or not that condition is pluggable. If we did this, ideally there's a way to stop trying (avoiding poison messages from stopping everything). This thing should be heavily tested.

### separate concern: on exception sleep
Collector or storage tier can sleep as opposed to immediately redoing the same command. SQS has a naive but probably fine approach for at least one site. If there needs to be more finesse on how to decide which way to sleep, that could be done in a relatively decoupled way, as simply as an okhttp elasticsearch interceptor for example! It could also be possible to make a generic exception handler, but I personally think it is a bit early for this. At any rate this is a separate change!

## Wrapping up
Please choose an option which makes sense to you and decide which you'd like to do or not. Please consider simplest first and do this in separate steps to conserve time for others who also need review or coding help.

malonso1976:
Hi Adrian,
   I am really interested in solving this issue and contributing back to zipkin. I really appreciate the time you are investing in reviewing my pull request and I thank you for that.
   Having said that, I want to be sure that the solution proposed is acceptable for you, so that the pull-request succeeds. All we are doing together is finding a suitable solution that meets your acceptance criteria.

   Alternatives:

- Semaphore at collector level: this would mimic the behaviour defined by elastic HttpCall at the collector level. It is easy to implement, requires and additional configuration parameter to setup the semaphore count. This solution works is like an intermediate queue/buffer between collector and storage to overcome capacity problems on the storage side, limiting maximum storage running span insertions.
- Retry policy at collector level: we need to define under which exceptions retry shall be performed. On elastic IllegalStateException("over capacity"). On cassandra it might be a DriverException subclass (BusyPoolException). This solution is similar to SQS: max number of retries and an adaptative wait time, upper bounded by a max wait time. Optionally a concurrency limit algorithm can be used to control congestion.

   I'd rather go for Semaphore at collector level, but if I choose a solution on my own without agreeing it with you, makes this a "trial-error" processs none of us would like to enter into.

   What do you think?

Regards,

codefromthecrypt:
Thanks for the loopback.

Personally, I link to https://github.com/Netflix/concurrency-limits as
once we get into semaphores and such, the code becomes higher
maintenance. I'd rather shade a dependency on that vs start building
the code (incidentally concurrency-limits core has no dep except
slf4j)

One approach we can take external to this PR which could help in
guiding this is to use RejectedExecutionException when there's a
retryable exception. This is the same runtime exception that
concurrency-limits uses.
https://github.com/Netflix/concurrency-limits/blob/744f8ae10523b0963b5d9453ce5d0d5c90d32139/concurrency-limits-core/src/main/java/com/netflix/concurrency/limits/executors/BlockingAdaptiveExecutor.java#L16

We can add javadoc to the SpanConsumer saying that implementations
should callback with RejectedExecutionException when they know an
external limit such as a BusyPoolException has happened. This should
break nothing, for example it would not break the code in SQS.
However, it could prepare us in a way where we can automatically
address resource constraints. It is also nicely decoupled IMHO. what
do you think?

If you like it, maybe go ahead and make elasticsearch and cassandra
use RejectedExecutionException (even if by wrapping) as a separate
change?

malonso1976:
Hi Adrian,
   I have commited a pull request with semaphore implementation.
   
   I will review your proposal and start a new branch to integrate netflix's concurrency-limits, it looks good.

Regards

codefromthecrypt:
looking forward to it!

zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
Logic-32: @malonso1976, @adriancole, two things here:
1. The code here mimics the code in #2169, meaning that if both get merged a conflict will have to be resolved.  The obvious resolution being to make multiple if-conditions.  I would recommend that, instead of having any if-statements, we make use of some composition and wrap the `spanConsumer()` in some abstractions that do the appropriate things based on the settings.  
2. I think what I've been describing in #2166 can actually help resolve the current complication of what to do with the Semaphore while also giving Kafka a hook to indicate it needs to slow down.  Basically, the idea would be to us an ExecutorService with a custom RejectedExecutionHandler that would call any specific method to say slow down/retry.  I'm not familiar with Kafka so I can't say what that'd look like.  But it's an idea.
    - Further, it looks like the concurrency limiter in #2169 already has an ExecutorService internally.  If that were made a little more accessible to external configuration then we could directly share it for the purposes of pushing back on Kafka and making use of Netflix's limiter
    - It would also give me a place to put a non-zero length queue for my issue (spike load in the HTTP Collector) but could probably default to 0 for most other purposes.

I may be a bit off-base with this comment.  Especially since I'm not entirely sure if there is intent to merge both this and the Netflix limiter eventually or just one of them.  I figured I'd at least point out the conflict at the very least and hope the ExecutorService could help out.

@malonso1976, if you don't mind, I could checkout (or fork) this branch and make the change so you could see what it would look like.  As I have it implemented (locally) it already separates bulk indexing requests from GET requests so there are no load conflicts there.  Which you mentioned as a problem and I've seen myself as well.

malonso1976:
Hi,
  I submitted 2 pr. The first one  #2166
<https://github.com/openzipkin/zipkin/pull/2166> is our current fix in our
production environment, but it does not protect http api span feed.
  The second one #2169 <https://github.com/openzipkin/zipkin/pull/2169> is
not tested properly and surely will need some fix. Let me review it next
week and we can talk about what needs to be adjusted and explain the ideas
behind the pr  and also regarding Kafka message consumption which I am sure
is not well done.

  Regarding the fork question, feel free to fork any branch and I will have
a look at your contribution

Kind regards,

Mariano

codefromthecrypt:
#2502 will address this use case, please give feedback and/or try there. thank you for this initial work!

PR #2168, opened 2018-08-15
Check-health for Kafka Collector

### LINKED ISSUE #2098: Make kafka health check meaningful
Right now, the kafka health check returns ok eventhough it is not. Let's improve that.

jeqo:
Use Admin Client API to validate that communication with Kafka brokers is working correctly.

Attempt to fix #2098

### DIFF
--- a/zipkin-collector/kafka/src/main/java/zipkin2/collector/kafka/KafkaCollector.java
+++ b/zipkin-collector/kafka/src/main/java/zipkin2/collector/kafka/KafkaCollector.java
@@ -154,7 +154,7 @@ public final class KafkaCollector extends CollectorComponent {
 
   final LazyKafkaWorkers kafkaWorkers;
   final Properties properties;
-  AdminClient adminClient;
+  volatile AdminClient adminClient;
 
   KafkaCollector(Builder builder) {
     kafkaWorkers = new LazyKafkaWorkers(builder);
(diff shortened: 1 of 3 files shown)


zipkin-collector/kafka/src/main/java/zipkin2/collector/kafka/KafkaCollector.java
```diff
@@ -148,9 +153,12 @@ public KafkaCollector build() {
   }
 
   final LazyKafkaWorkers kafkaWorkers;
+  private final Properties properties;
+  private AdminClient adminClient;
```
zeagord: I would change this to package specific access to reduce the size of the bytecode.

zeagord (approved):

zipkin-collector/kafka/src/main/java/zipkin2/collector/kafka/KafkaCollector.java
jeqo: @zeagord didn't know that, thanks! :) Fixing it.

zipkin-collector/kafka/src/main/java/zipkin2/collector/kafka/KafkaCollector.java
```diff
@@ -164,15 +172,19 @@ public CheckResult check() {
     try {
       CheckResult failure = kafkaWorkers.failure.get(); // check the kafka workers didn't quit
       if (failure != null) return failure;
+      if (adminClient == null) adminClient = AdminClient.create(properties);
+      KafkaFuture<String> maybeClusterId = adminClient.describeCluster().clusterId();
+      maybeClusterId.get(1, TimeUnit.SECONDS);
       return CheckResult.OK;
-    } catch (RuntimeException e) {
+    } catch (RuntimeException | TimeoutException | ExecutionException | InterruptedException e) {
```
codefromthecrypt: probably easier to just catch exception

zipkin-collector/kafka/src/main/java/zipkin2/collector/kafka/KafkaCollector.java
```diff
@@ -164,15 +172,19 @@ public CheckResult check() {
     try {
       CheckResult failure = kafkaWorkers.failure.get(); // check the kafka workers didn't quit
       if (failure != null) return failure;
+      if (adminClient == null) adminClient = AdminClient.create(properties);
```
codefromthecrypt: if the goal is to initialize (then later close) the admin client once, I think you will want to double check initialize with volatile ref like some other code in this repo

reason is close isnt guaranteed to be same thread as check

codefromthecrypt (commented):
small note otherwise lgtm thanks!

zipkin-collector/kafka/src/main/java/zipkin2/collector/kafka/KafkaCollector.java
jeqo: Agree. I'll fix it.

zipkin-collector/kafka/src/main/java/zipkin2/collector/kafka/KafkaCollector.java
jeqo: Thanks for the advise, I'll check it.

codefromthecrypt:
lgtm!

codefromthecrypt:
can you do the same in zipkin-reporter?

PR #2169, opened 2018-08-18
Concurrency limiter from Netflix to avoid throttling storage

malonso1976:


codefromthecrypt (commented):
thanks for giving this a go.

first question: have you tried this successfully?

second comment: I think the real constraint is storage so this should be a property at global level not nested for each collector. otherwise your constraints won't apply if both http and Kafka spans are accepted.


we can work through other design things after consideration of above 2 things.

malonso1976:
Hi Adrian,
   I just made a initial unit test (removed from code now) to see if ConcurrencyLimiter worked as expected. I want to test it in one of our environments under load to see if it works for us.
   For your second comment, I did not take into consideration http collector... my fault. In our setup we do not received spans directly through http endpoint.

   Do you have any design concern how shall concurrency limiter applied on http collector? The http collector is a special case, because http client knows synchronously whether collection succeeded or not, and current ConcurrentLimiter works as a pool of threads, processing requests asynchronously.

malonso1976:
Hi again,
  just tried to move up the configuration and shared the concurrency limiter with the http collector. Please see if it is closer to what you suggested and I will try to test it in a real environment as soon as possible.

Regards,

codefromthecrypt:
yeap on this being where I was hoping it would be (will apply to all collectors, not just http but good start)

once you have your results in, we can chat further. thanks so far.

shakuzen:
@malonso1976 were you able to try this out in a real environment?

malonso1976:
Not really, but I am on it.

I stopped working on it because I had a family emergency that took me away
from work a couple of months.

I am back and want to test it soon.

shakuzen:
@malonso1976 No problem at all. Let us know when you have a chance to test it out.

pom.xml
```diff
@@ -64,6 +64,7 @@
     <!-- Be careful to set this as a provided dep, so that it doesn't interfere with dependencies
          from other projects. For example, cassandra and spring boot set guava versions -->
     <guava.version>19.0</guava.version>
+    <netflix.concurrency.limits.version>0.0.49</netflix.concurrency.limits.version>
```
Logic-32, line 67: @adriancole, concurrency-limits v0.2.0 is the latest tag but uses Java 8 features (such as Optional).  Do you know of any current plans to get io.zipkin.zipkin2:zipkin off of 1.6 and up to at least 1.8?  I saw #777 but it doesn't appear to have actually upped the `main.signature.artifact`.

It also looks like [retrolambda does not support recompiling dependencies](https://github.com/evant/gradle-retrolambda/issues/80) or I'd explore that option further.

Logic-32: Never mind, made a new artifact outside of zipkin-core so it's not subject to this limitation.  Affords me some other luxuries around not-shading dependencies as well :)

codefromthecrypt:
zipkin-collector is java 8 for this reason. we can shade some smaller deps
if needed

codefromthecrypt:
#2502 will address this use case. Thanks

codefromthecrypt:
https://github.com/apache/incubator-zipkin/pull/2502 now includes test instructions. please give a try!

codefromthecrypt:
note: concurrency-limits-core has been abandoned by Netflix by over a year at this point. We may have to revisit this

codefromthecrypt:
I opened this noticing no release since last july even if there were a couple more months of changes, nothing was released https://github.com/Netflix/concurrency-limits/issues/157

anuraaga:
I've heard the word resilience4j lately - never used it but just a reference

Logic-32:
Reslience4j and Failsafe are interesting options but, when evaluating them, make sure they solve the problem this is looking to address.

Namely, storage gets backed up so we need to slow down how much we're writing to it but do so in a way that doesn't cause the server to exhaust memory.  In this particular case, a limited-size queue is used to buffer some data while we have a small bottleneck we're working with.  Some simple retries with an appropriate backoff policy could address the same basic problem of slow storage but may be harder to tame with respect to pending actions.  Meaning, I can't say how hard it is to prevent an "infinite queue" with the retry handlers.

FWIW, the ExecutorService that backs this implementation is pretty universal and supported by Java.  The only part that is not is the one that adjusts how many threads are allowed to be active in it.  To a certain extent, that logic could be directly incorporated into Zipkin instead of pulled in via Netflix.

anuraaga:
@Logic-32 Thanks for summarizing the problem, it'll help if we look into any migration! I was indeed incorrectly thinking the `RateLimiter` in resilience4j would be relevant to us, but based on your explanation, my reading is that their bulkhead may provide what we need.

https://github.com/resilience4j/resilience4j#632-threadpoolbulkhead

PR #2182, opened 2018-08-30
make zipkin2.Span.tag field mutable while it's initialized as empty

anton-kolhun:
I faced the problem for the following scenario:

Given Zipkin2.Span.Builder with 'tags' = null
Then Span(builder) gets initialized with immutable 'tags' field:
`tags =  Collections.emptyMap()`
Then while reporting the span( `zipkin2.reporter.Reporter.report(..))`)
 and modifying it over SpanAdjuster hook:
`span -> span.putTag(..)`
The UnsupportedOperationException occurs

This change implies initializing zipkin2.Span object with a mutable 'tags' field.

### DIFF
--- a/zipkin/src/main/java/zipkin2/Span.java
+++ b/zipkin/src/main/java/zipkin2/Span.java
@@ -637,7 +637,7 @@ public final class Span implements Serializable { // for Spark and Flink jobs
   final long timestamp, duration; // zero means null, saving 2 object references
   final Endpoint localEndpoint, remoteEndpoint;
   final List<Annotation> annotations;
-  final Map<String, String> tags;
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
sorry this type is intentionally immutable. Span.Builder is mutable (you can call span.toBuilder() to get one). https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/internal/recorder/MutableSpan.java is an example of a more efficient mutable impl which will be public api soon.

codefromthecrypt:
feel free to talk more about this here https://gitter.im/openzipkin/zipkin

anton-kolhun:
the point is clear now. thanks for explaining

codefromthecrypt:
no problem!

Issue #1906, opened 2018-02-05
Overlap of texts in trace GUI


zsojma:
Hello, we just started using zipkin together with linkerd. Unfortunatelly, because of long names of linkerd services, the trace GUI is not very clear... Here is a screen:

![image](https://user-images.githubusercontent.com/28751135/35827845-cf4d01c6-0abd-11e8-96c9-6507ef84399f.png)

Please, is there a possibility / some setting / configuration how to improve it so texts do not overlap each other? Or do you plan to improve the appereance?

Thank you

codefromthecrypt:
Yeah these names are unfortunate. It would be better to use a shorter name
and add a tag with the huge ones.

There is no plan per se as this is all volunteer. This will sit until
someone who knows how to do something better comes up with an idea.

zeagord:
Clipping the text by setting the text-overflow in the css may help and the full name can be displayed on mouseover . I can try do that If it is an agreeable behaviour.

codefromthecrypt:
@zeagord I don't know any better option for this situation. I'd go with the service name clipping until there's another way cc @klingerf

zeagord:
This is fixed by #2184

shakuzen:
Thanks as always, @zeagord!

PR #2184, opened 2018-09-07
Fixes long service names

zeagord:
This will fix long service names in the spans #1906 . 
Provide the value of the service name in a visually pleasing tooltip

screen shot 2561-09-07 at 14 11 58


codefromthecrypt:
@klingerf I think this should help with linkerd names

Thanks so much Raja!

codefromthecrypt (approved):

PR #2191, opened 2018-09-17
Add the ability to specify the complete JDBC url

### LINKED ISSUE #2190: [Feature request] More flexible MySQL datasource configuration
Currently the MySQL data storage is configured roughly here: https://github.com/openzipkin/zipkin/blob/d47e80433cdf97a78e92712f313ae607e6f61dec/zipkin-autoconfigure/storage-mysql/src/main/java/zipkin2/autoconfigure/storage/mysql/ZipkinMySQLStorageProperties.java#L89-L102

This doesn't give a whole lot of flexibility in the datasource setup. In my case I have a Galera cluster for which I can specify a JDBC url like this: `jdbc:mariadb:failover://host1:3306,host2:3306,host3:3306/database?prop1=true` This allows for failover in case one of the nodes is down.

Even if you use 'plain' MySQL there is still a host of failover configuration options: https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-config-failover.html.

Can we perhaps add an optional property to `ZipkinMySQLStorageProperties` to specify the JDBC url directly and if it doesn't exist use the current implementation as a fallback?

Given a property `jdbcUrl`, kind of like this?

```java
    StringBuilder url = new StringBuilder();
    if (getJdbcUrl() != null) {
      url.append(getJdbcUrl());
    } else {
      url.append("jdbc:mysql://");
      url.append(getHost()).append(":").append(getPort());
      url.append("/").append(getDb());
      url.append("?autoReconnect=true");
      url.append("&useSSL=").append(isUseSsl());
      url.append("&useUnicode=yes&characterEncoding=UTF-8");
    }
```

aukevanleeuwen:
I know this storage component is deprecated, but I'd still like to be able to specify it while we are implementing another storage backend. 

Closes #2190

### DIFF
--- a/zipkin-autoconfigure/storage-mysql/src/main/java/zipkin2/autoconfigure/storage/mysql/ZipkinMySQLStorageProperties.java
+++ b/zipkin-autoconfigure/storage-mysql/src/main/java/zipkin2/autoconfigure/storage/mysql/ZipkinMySQLStorageProperties.java
@@ -17,11 +17,13 @@ import com.zaxxer.hikari.HikariDataSource;
 import java.io.Serializable;
 import javax.sql.DataSource;
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.util.StringUtils;
 
 @ConfigurationProperties("zipkin.storage.mysql")
 class ZipkinMySQLStorageProperties implements Serializable { // for Spark jobs
   private static final long serialVersionUID = 0L;
 
+  private String jdbcUrl;
   private String host = "localhost";
   private int port = 3306;
   private String username;
@@ -29,6 +31,14 @@ class ZipkinMySQLStorageProperties implements Serializable { // for Spark jobs
   private String db = "zipkin";
   private int maxActive = 10;
   private boolean useSsl;
+
+  public String getJdbcUrl() {
+    return jdbcUrl;
+  }
+
+  public void setJdbcUrl(String jdbcUrl) {
+    this.jdbcUrl = jdbcUrl;
+  }
 
   public String getHost() {
     return host;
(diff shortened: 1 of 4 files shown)


zipkin-autoconfigure/storage-mysql/src/main/java/zipkin2/autoconfigure/storage/mysql/ZipkinMySQLStorageProperties.java
```diff
@@ -87,18 +97,27 @@ public void setUseSsl(boolean useSsl) {
...
     result.setMaximumPoolSize(getMaxActive());
     result.setUsername(getUsername());
     result.setPassword(getPassword());
     return result;
   }
+
+  private String determineJdbcUrl() {
+    StringBuilder url = new StringBuilder();
+    if (StringUtils.hasText(getJdbcUrl())) {
+      url.append(getJdbcUrl());
```
codefromthecrypt: nit: there's no value to using stringbuilder here. just exit early then retain existing logic if not present

codefromthecrypt (approved):
This sounds reasonable to me.. at first I thought MYSQL_URL would be better but I think yours is better actually.

zipkin-autoconfigure/storage-mysql/src/main/java/zipkin2/autoconfigure/storage/mysql/ZipkinMySQLStorageProperties.java
aukevanleeuwen: 👍 done

aukevanleeuwen:
Not sure what's up with the checks, should I `amend` my commit and try a couple of times? Doesn't seem related to my changes?

codefromthecrypt (approved):

codefromthecrypt:
thanks!

PR #2209, opened 2018-10-15
Manually sets UI depedendencies to avoid cryptiles 3.x

codefromthecrypt:
We had a vulnerability, albeit only dev deps, associated with cryptiles
3.x. This hard-codes versions of libraries that depended on it in favor
of 4.x.


codefromthecrypt:
manually verified that stylesheets etc look fine

PR #2212, opened 2018-10-15
Migrates to headless UI tests to work around broken CircleCI build

### LINKED ISSUE #2211: Move UI tests off PhantomJS to something else
Our UI tests in CircleCI (not travis) fail on PhantomJS

```
Exit code: 1 
Output:
[INFO] 
[INFO] 15 10 2018 02:19:29.067:ERROR [launcher]: Cannot start PhantomJS
[INFO] 	Auto configuration failed
[INFO] 140336306921088:error:25066067:DSO support routines:DLFCN_LOAD:could not load the shared library:dso_dlfcn.c:185:filename(libssl_conf.so): libssl_conf.so: cannot open shared object file: No such file or directory
```

Seems PhantomJS is dead anyway...

https://www.puzzle.ch/de/blog/articles/2018/02/12/phantomjs-is-dead-long-live-headless-browsers

Headless chrome seems possible.. not sure which circleci images would be implied.

Notes: Sheng mentioned that circleci is not supported for apache anyway, so maybe we can avoid this by also stopping our dependency on circleci cc @openzipkin/devops-tooling

codefromthecrypt:
This moves to headless chrome. While this depends on Chrome being
installed, it avoids problems with the abandoned PhantomJS library.

Also, we do need CircleCI to pass else we don't know if things only
tested there work (such as JDK 11).

Fixes #2211


zipkin-ui/test/spanConverter.test.js
```diff
@@ -1050,18 +1050,18 @@ describe('SPAN v1 merge by ID', () => {
     expect(spans).to.deep.equal([
       {
         traceId: '00000000000000022222222222222222',
-        id: '000000000000000a',
```
codefromthecrypt, line 1053: our ordering code orders root spans first

shakuzen (approved):

PR #2223, opened 2018-10-29
Fix Issue Templates

beckje01:
Remove markdown from about sections.


codefromthecrypt:
thanks for the fast turnaround!

Issue #2217, opened 2018-10-23
Refactor UI to internally use the v2 model


codefromthecrypt:
Right now, we pull data in v2 format #1802, but we still operate on it in v1.

GuangmingLuo:
Hi, 
I recently find that I can't search trace by id in this UI [Go to trace] when I use 128 bit traceId for both version 2.10.4 and 2.11, even though it does exist in the DB. 
Does this relate to this issue, or maybe raise a new issue ?
BR

codefromthecrypt:
I will try start this by changing the UI logic step-by-step to eventually use v2 data:

1. break shared v1 spans into single host spans, but still v1 format
Ex. {"cs" "sr" "sa" "ca" "ss" "cr"} -> {"cs" "cr", "sa"},  {"sr" "ss", "ca"}
1. add span.kind, remoteEndpoint annotation to the UI's model temporarily and use it
 {"cs" "cr", "sa"},  {"sr" "ss", "ca"} -> {kind = CLIENT, "cs" "cr", remoteEndpoint},  {kind = SERVER, "sr" "ss", remoteEndpoint}
1. add localEndpoint annotation to the UI's model temporarily and use it
{kind = CLIENT, "cs" "cr", remoteEndpoint},  {kind = SERVER, "sr" "ss", remoteEndpoint} -> {kind = CLIENT, "cs" "cr", localEndpoint, remoteEndpoint},  {kind = SERVER, "sr" "ss",  localEndpoint, remoteEndpoint}
1. complete switch to v2 model
{kind = CLIENT, "cs" "cr", localEndpoint, remoteEndpoint},  {kind = SERVER, "sr" "ss",  localEndpoint, remoteEndpoint} -> {kind = CLIENT, timestamp, duration, localEndpoint, remoteEndpoint},  {kind = SERVER, timestamp, duration,  localEndpoint, remoteEndpoint, shared=true}

codefromthecrypt:
one interesting design luck is that in the current UI we are a little inefficient. We make one traces api query for the index, then redundantly make a trace api query to show the trace view. A better UI will reuse the same data. However when we are porting logic, this is a bit lucky. Since each screen uses a different api, I can break one screen and not break the other one. For example, I can port the search screen to use v2 without touching the trace view screen.

codefromthecrypt:
Another thing to consider. We should have stub/mock data that includes some interesting traces:
* trace with clock skew
* regular RPC trace
* messaging trace

[...]

PR #2228, opened 2018-10-31
Simplifies summary logic by not doing skew correction

codefromthecrypt:
The trace summary screen and individual trace screens are served by
different data. At the moment, offsets are not considered when computing
the trace duration broken down by service. Maybe it should be, but
removing clock skew makes migration to v2 simpler and we can then
revisit this logic.

See #2217

### DIFF
--- a/zipkin-ui/js/component_data/default.js
+++ b/zipkin-ui/js/component_data/default.js
@@ -4,7 +4,6 @@ import $ from 'jquery';
 import queryString from 'query-string';
 import {traceSummary, traceSummariesToMustache} from '../component_ui/traceSummary';
 import {SPAN_V1} from '../spanConverter';
-import {correctForClockSkew} from '../skew';
 
 export function convertToApiQuery(source) {
   const query = Object.assign({}, source);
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
To test this, I used "skew.json" before and after the change and noticed the service percentage was exactly the same. This is how I know the percentage function doesn't consider offset even without looking at the code (I also looked at the code though!)

screen shot 2018-10-31 at 11 26 44 am

zeagord (approved):

tacigar:
I think that is good!
But I have a question.

I think that `adjustTimestamps` in `skew.js` may change the timestamps of the spans, and `traceDuration` in `traceSummary.js` will be used for duration calculation.
And I think that in many case `traceDuration` function returns `span[0].timestamp + span[0].duration - span[0].timestamp` (== `span[0].duration`).
So in many case, I think the process of clock-skew is unnecessary.
Is my opinion correct?
And is this even in all cases?

Sorry for newbie question 😞

codefromthecrypt:
This may indeed be a problem. I will look into it now as we don't have data yet to break this.

adriancole:
What we did is add an async span that runs past the critical path. 

When we don't do skew correction, the result is this:
screen shot 2018-10-31 at 5 39 54 pm


When we do skew correction, the result is this:
screen shot 2018-10-31 at 5 40 14 pm

While we don't know which result is correct, we should keep the historical behavior until we know what the correct result should be. So, this reverts the change.

PR #2229, opened 2018-10-31
Reverts skew "uncorrection" in index as it affects service percentages

codefromthecrypt:
In #2228 my assumption was incorrect as @tacigar highlighted. This makes
a test to ensure the logic stays the same for now.


codefromthecrypt:
What we did is add an async span that runs past the critical path. 

When we don't do skew correction, the result is this:
screen shot 2018-10-31 at 5 39 54 pm


When we do skew correction, the result is this:
screen shot 2018-10-31 at 5 40 14 pm

While we don't know which result is correct, we should keep the historical behavior until we know what the correct result should be. So, this reverts the change.

PR #2231, opened 2018-11-01
Upgrades to Springboot 2.1 and Zipkin Layout Factory v0.0.3

zeagord:


zipkin-server/src/main/resources/zipkin-server-shared.yml
```diff
@@ -160,6 +160,9 @@ spring:
     exclude:
       # otherwise we might initialize even when not needed (ex when storage type is cassandra)
       - org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration
+  # Added this to override the default bean definition as the CharacterEncodingFilter bean already
+  # available from HttpEncodingAutoConfiguration
+  main.allow-bean-definition-overriding: true
```
codefromthecrypt: good comment

codefromthecrypt:
travis is angry https://travis-ci.org/openzipkin/zipkin/builds/449208466?utm_source=github_status&utm_medium=notification

zeagord:
Removed `CharacterEncodingFilter` as it is no longer required since Springboot doesn't enforce the response encoding
Removed `WebMvcConfigurerAdapter` as it is deprecated and we can `WebMvcConfigurer` interface to add the resource handlers

codefromthecrypt:
lemme know if you think this is ready to go

shakuzen:
I suppose we could merge, but we wouldn't be able to release with this as currently logging is broken. As mentioned in chat, we think this is the same issue as https://github.com/spring-projects/spring-boot/issues/15074.

codefromthecrypt:
I would suggest against merge as skippable master is important. there might
be change here that works also with boot v2. we can pull that out and merge.

shakuzen (approved):
I think we're good to merge as long as logging still works with Java 8/9/10. I realized after my previous comment that the status quo should be maintained with these changes. Logging won't work with Java 11, but it didn't work with Java 11 before these changes either. So we should still be able to make a release, we just shouldn't claim full runtime compatibility with Java 11 until the logging issue is resolved (most likely by upgrading to a patched Spring Boot version).

codefromthecrypt:
Oh I see. Thanks for the clarification, I agree then.. Click merge!

PR #2232, opened 2018-11-01
Brings back relative width in trace query screen

codefromthecrypt:
This fixes a bug in the trace query screen and reorganizes code to more
easily tell which screens need which data from the trace summary.

Before:
screen shot 2018-11-01 at 4 56 19 pm

After:
screen shot 2018-11-01 at 4 55 35 pm

### DIFF
--- a/zipkin-ui/js/component_ui/traceSummary.js
+++ b/zipkin-ui/js/component_ui/traceSummary.js
@@ -108,11 +108,19 @@ export function getServiceName(span) {
   return allServiceNames.length === 0 ? null : allServiceNames[0];
 }
 
-function getSpanTimestamps(spans) {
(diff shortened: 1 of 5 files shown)


zipkin-ui/js/component_ui/traceSummary.js
```diff
@@ -227,40 +231,32 @@ export function traceSummariesToMustache(serviceName = null, traceSummaries, utc
     return [];
   } else {
     const traceSummariesCleaned = removeEmptyFromArray(traceSummaries);
-    const maxDuration = Math.max(...traceSummariesCleaned.map((s) => s.duration)) / 1000;
```
codefromthecrypt, line 230: this was the bug. all the data are in microseconds, yet somehow this line thought it should be millis :P

PR #2233, opened 2018-11-02
Refactors entrypoints into traceSummary and backfills UI tests

codefromthecrypt:
This refactors some of the call sites into traceSummary to reduce
redundant work. To help expose what UI templates use, this adds some
tests that push an api template up to the point of invoking mustache.

This serves two purposes: sets up trip wires to ensure we don't break
something when migrating to v2. This also will help us identify dead
data or needlessly different named data.

### DIFF
--- a/zipkin-ui/js/component_data/default.js
+++ b/zipkin-ui/js/component_data/default.js
@@ -4,7 +4,6 @@ import $ from 'jquery';
 import queryString from 'query-string';
 import {traceSummary, traceSummariesToMustache} from '../component_ui/traceSummary';
 import {SPAN_V1} from '../spanConverter';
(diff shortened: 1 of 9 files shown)


codefromthecrypt:
pushing this up here @tacigar and @zeagord. Main thing I'm trying to get to is an understanding of what data is being used by what and why. I'm not there yet, but it looks like there will be some opportunities for simplification.

codefromthecrypt:
small bug here.. have a fix which I will add tomorrow

codefromthecrypt:
Fixed and verified percentages show exact same after refactoring:
screen shot 2018-11-03 at 8 01 28 am

zeagord (approved):

Issue #998, opened 2016-02-24
The total time of each span seems incorrect


lycplus:
In the summary page (home page)，zipkin will show a summary time of each span below each trace. We have many redis spans each costs 300us+, but in the summary page, it shows `redis * 232 5ms`,`5ms/232=21.5us`, so, I think, may be the summary data is incorrect.

shakuzen:
As far as I can tell from spot checking with the latest version of Zipkin, it appears the time shown on the traces view for each service name is the **longest** span duration for that service name, which perhaps isn't apparent.

codefromthecrypt:
@liyichao check this out.. it is confirmed (even if not necessarily the right thing to do) #2234

lycplus:
Then I misunderstand the meaning, which is not that straightforward. Close the issue if you think it is not a problem.

codefromthecrypt:
there was a bug actually.. https://github.com/openzipkin/zipkin/pull/2235

codefromthecrypt:
#2488 will add some clarifications also

PR #2234, opened 2018-11-03
Changes index page to order services descending by max span duration

codefromthecrypt:
I figured out what "max" was in the index page. It corresponds to the
longest duration span in the trace, as attributed to a specific service.

This orders the search results descending by the most likely culprit,
using relevant time units to prevent minutes from showing as huge counts
of milliseconds, or short traces showing up as zero.

Before:
screen shot 2018-11-03 at 3 18 38 pm

After: (notice the slowest service is the first one listed)
screen shot 2018-11-03 at 3 18 30 pm

### DIFF
--- a/zipkin-ui/test/component_data/default.test.js
+++ b/zipkin-ui/test/component_data/default.test.js
@@ -13,10 +13,10 @@ describe('convertSuccessResponse', () => {
       duration: 168.731,
       durationStr: '168.731ms',
       width: 100,
-      totalSpans: 2,
-      serviceDurations: [
-        {name: 'backend', count: 1, max: 111},
-        {name: 'frontend', count: 2, max: 168}
+      spanCount: 2,
+      serviceSummaries: [
+        {serviceName: 'frontend', spanCount: 2, maxSpanDurationStr: '168.731ms'},
+        {serviceName: 'backend', spanCount: 1, maxSpanDurationStr: '111.121ms'}
       ],
       infoClass: ''
     };
@@ -38,10 +38,10 @@ describe('convertSuccessResponse', () => {
       duration: 168.731,
       durationStr: '168.731ms',
       width: 100,
-      totalSpans: 2,
-      serviceDurations: [
-        {name: 'backend', count: 1, max: 111},
-        {name: 'frontend', count: 2, max: 168}
+      spanCount: 2,
+      serviceSummaries: [
+        {serviceName: 'frontend', spanCount: 2, maxSpanDurationStr: '168.731ms'},
+        {serviceName: 'backend', spanCount: 1, maxSpanDurationStr: '111.121ms'}
       ],
       infoClass: '',
       servicePercentage: 65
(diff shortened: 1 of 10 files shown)


codefromthecrypt:
cc @tacigar

codefromthecrypt:
ps I asked @narayaruna offline and basically a "don't care" response as this isn't a very important change. I agree that it isn't important for analysis.. the main goal of this specific change is understanding what the code does so that we can complete v2 refactoring (#2217) which will let us do a UI replacement easier as discussed at LINE https://cwiki.apache.org/confluence/display/ZIPKIN/2018-10-29+Zipkin+UI+at+LINE+Tokyo

so yeah, this is in the weeds, likely not much notable help to end users, but it is a part of becoming better in a different sense (the UI revamp by @tacigar)

/me ends blathering

zipkin-ui/test/component_data/default.test.js
```diff
@@ -61,9 +61,9 @@ describe('convertSuccessResponse', () => {
       duration: 0.017,
       durationStr: '17μ',
       width: 100,
-      totalSpans: 1,
-      serviceDurations: [
-        {name: 'backend', count: 1, max: 0} // TODO: figure out what max means
```
codefromthecrypt, line 66: most important change was removing this TODO :P

zipkin-ui/test/component_data/trace.test.js
```diff
@@ -212,10 +212,10 @@ describe('convertSuccessResponse', () => {
       duration: '168.731ms',
       services: 2,
       depth: 2,
-      totalSpans: 2,
-      serviceCounts: [
-        {name: 'backend', count: 1, max: 111},
-        {name: 'frontend', count: 2, max: 168}
+      spanCount: 2,
+      serviceNameAndSpanCounts: [
+        {serviceName: 'backend', spanCount: 1},
```
codefromthecrypt, line 217: second most important change (to me) was removing dead data as it helps us understand what is needed by which screen

zipkin-ui/test/component_data/trace.test.js
```diff
@@ -262,10 +262,10 @@ describe('convertSuccessResponse', () => {
 
     const timeMarkers = [
       {index: 0, time: ''},
-      {index: 1, time: '3.4000000000000004μ'},
-      {index: 2, time: '6.800000000000001μ'},
-      {index: 3, time: '10.2μ'},
-      {index: 4, time: '13.600000000000001μ'},
```
codefromthecrypt, line 268: 3rd change is the actual UI nicies, like this horrible rendering of fractions of a us :D

shakuzen:
What that duration even meant was confusing to users and a ticket was opened long ago: https://github.com/openzipkin/zipkin/issues/998. I had never really paid much attention to it before, but I agree it isn't clear what it means. This change seems to be an improvement for people who understand what it means in the first place.

codefromthecrypt:
merged for reason tommy mentioned. we can kill the confusion beast with each change like this.

PR #2238, opened 2018-11-06
Works around i18n bug for chinese users by making zh_CN default for zh

### LINKED ISSUE #2175: i18n failed
when i visit index html, error occur below.
it might be because `data-i18n="traces.lookback">` (templates/index.mustache)have not been defined in i18n files.

```
jquery.i18n.properties.js:271 Uncaught TypeError: Cannot read property 'length' of undefined
    at Object.$.i18n.prop (jquery.i18n.properties.js:271)
    at HTMLSpanElement.<anonymous> (i18n.js:22)
    at Function.each (jquery.js:354)
    at De.fn.init.each (jquery.js:189)
    at Object.callback (i18n.js:18)
    at callbackIfComplete (jquery.i18n.properties.js:306)
    at loadAndParseFiles (jquery.i18n.properties.js:322)
    at jquery.i18n.properties.js:319
    at Object.success (jquery.i18n.properties.js:349)
    at l (jquery.js:3268)
```

codefromthecrypt:
Eventually, we may need to change the javascript library used as it is
broken.

Fixes #2175


codefromthecrypt:
@MrGlaucus @gsta-cloudins @gzchenyong @ChinaTelecomDragon @sheng-wu @v35715009 @JoveYu @GuangmingLuo @jaydensun can one of you confirm this is ok by running a build off this branch? If not, could you confirm if we built a snapshot?

codefromthecrypt:
while waiting on feedback here, we can also consider an option which is to replace this library we use https://github.com/openzipkin/zipkin/issues/2239

codefromthecrypt:
I heard from @v35715009 that this doesn't help. We probably need to change the library

PR #2241, opened 2018-11-06
Adds messaging and database traces from Jorge 

zeagord:


codefromthecrypt:
only suggestion is to name the files also based on site. this helps folks know they are real, maybe answer questions later or link to site docs.

zeagord:
Sure. Let me do that.

codefromthecrypt:
he opted for no attribution, which is totally ok

codefromthecrypt:
oops the first is in v1 format :) we should wash this through the server and re-add it.

(ex POST through the v1 endpoint, then GET via the v2 one)

PR #2245, opened 2018-11-09
Migrates index screen to v2 format

codefromthecrypt:
This allows the index screen to work in v2 format by re-implementing clock skew and merging logic against v2 structs.

### DIFF
--- a/zipkin-ui/js/component_ui/traceToMustache.js
+++ b/zipkin-ui/js/component_ui/traceToMustache.js
@@ -2,9 +2,9 @@ import _ from 'lodash';
 import {
   traceSummary,
   getServiceNameAndSpanCounts,
-  getServiceNames,
   mkDurationStr
 } from './traceSummary';
+import {SPAN_V1} from '../spanConverter';
 import {Constants, ConstantNames} from './traceConstants';
 
 export function getRootSpans(spans) {
@@ -14,6 +14,20 @@ export function getRootSpans(spans) {
 
 function compareSpan(s1, s2) {
   return (s1.timestamp || 0) - (s2.timestamp || 0);
+}
+
+function endpointsForSpan(span) {
+  return _.union(
+    (span.annotations || []).map(a => a.endpoint),
+    (span.binaryAnnotations || []).map(a => a.endpoint)
+  ).filter(h => h != null);
+}
+
+function getServiceNames(span) {
+  return _(endpointsForSpan(span))
+      .map((ep) => ep.serviceName)
+      .filter((name) => name != null && name !== '')
+      .uniq().value();
 }
 
 function childrenToList(entry) {
@@ -154,8 +168,9 @@ export function getServiceName(span) { // export for testing
   return allServiceNames.length === 0 ? null : allServiceNames[0];
 }
 
-export default function traceToMustache(trace, logsUrl = undefined) {
-  const t = traceSummary(trace);
+export default function traceToMustache(v2Trace, logsUrl = undefined) {
+  const trace = SPAN_V1.convertTrace(v2Trace);
+  const t = traceSummary(v2Trace);
   const traceId = t.traceId;
   const duration = t.duration;
   const serviceNameAndSpanCounts = getServiceNameAndSpanCounts(t.groupedTimestamps);
--- a/zipkin-ui/js/skew.js
+++ b/zipkin-ui/js/skew.js
@@ -1,15 +1,16 @@
+import {mergeV2ById} from './spanCleaner';
+
 /*
- * Convenience type representing a tree. This is here because multiple facets in zipkin require
- * traversing the trace tree. For example, looking at network boundaries to correct clock skew, or
- * counting requests imply visiting the tree.
+ * Convenience type representing a trace tree. Multiple Zipkin features require a trace tree. For
+ * example, looking at network boundaries to correct clock skew and aggregating requests paths imply
+ * visiting the tree.
  */
-// originally zipkin2.internal.Node.java
-class Node {
-  constructor(value) {
+// originally zipkin2.internal.SpanNode.java
+class SpanNode {
+  constructor(span) {
     this._parent = undefined; // no default
-    this._value = value; // undefined is possible when this is a synthetic root node
+    this._span = span; // undefined is possible when this is a synthetic root node
     this._children = [];
-    this._missingRootDummyNode = false;
   }
 
   // Returns the parent, or undefined if root.
@@ -17,9 +18,13 @@ class Node {
     return this._parent;
   }
 
-  // Returns the value, or undefined if a synthetic root node
-  get value() {
-    return this._value;
+  _setParent(newParent) {
+    this._parent = newParent;
+  }
+
+  // Returns the span, or undefined if a synthetic root node
+  get span() {
+    return this._span;
   }
 
   // Returns the children of this node
@@ -28,22 +33,20 @@ class Node {
   }
 
   // Mutable as some transformations, such as clock skew, adjust the current node in the tree.
-  setValue(newValue) {
-    if (!newValue) throw new Error('newValue was undefined');
-    this._value = newValue;
+  setSpan(span) {
+    if (!span) throw new Error('span was undefined');
+    this._span = span;
   }
 
-  _setParent(newParent) {
-    this._parent = newParent;
-  }
-
+  // Adds the child IFF it isn't already a child.
   addChild(child) {
+    if (!child) throw new Error('child was undefined');
     if (child === this) throw new Error(`circular dependency on ${this.toString()}`);
     child._setParent(this);
     this._children.push(child);
   }
 
-  // Returns an array of values resulting from a breadth-first traversal at this node
+  // Returns an array of spans resulting from a breadth-first traversal at this node
   traverse() {
     const result = [];
     const queue = [this];
@@ -51,8 +54,8 @@ class Node {
     while (queue.length > 0) {
       const current = queue.shift();
 
-      // when there's a synthetic root span, the value could be undefined
-      if (current.value) result.push(current.value);
+      // when there's a synthetic root span, the span could be undefined
+      if (current.span) result.push(current.span);
 
       const children = current.children;
       for (let i = 0; i < children.length; i++) {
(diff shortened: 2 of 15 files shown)


codefromthecrypt:
ported more logic to javascript after testing in java. need some test backporting.

the hard part next will be to rewrite the clock skew to be single side. this code was deleted from java after porting to javascript. I might temporarily put it back until I figure out the logic as I am very slow with javascript due to nuance of the language and not as familiar with how to debug similar to java breakpoints.

zipkin/src/test/java/zipkin2/internal/NodeTest.java
```diff
@@ -88,8 +89,80 @@ public void addChild_selfNotAllowed() {
...
+    Node<Span> current = root;
+    for (int i = 1, length = trace.size() - 1; i < length; i++) {
+      current = current.children.get(0);
+      assertThat(current.value).isEqualTo(trace.get(i));
+      assertThat(current.children).extracting(Node::value)
+        .containsExactly(trace.get(i + 1));
+    }
+  }
+
+  @Test public void constructsTraceTree_qualifiesChildrenOfDuplicateServerSpans() {
```
codefromthecrypt: this test now works.. although a lot of others fail now.. whoot!

codefromthecrypt:
this work can be a bit overwhelming so I am going to first port the v2 merging to java (ironically tested in javascript first!), and in a separate PR. Then, I'll change the tree building stuff (Node.java) to work off an assumption that spans with the same (id, kind) are merged. This might be somewhat tricky as late completed server spans might not report kind (but do report localEndpoint!).

Anyway that strangles the nastiest stuff into one place. After that, it makes the rest of the work easier.

drolando:
Does this move back the client/server span merging to Java? So the UI expects to receive pre-merged spans?

codefromthecrypt:
java is easier for me to figure out problems in java before porting to
javascript. but also dependency linking is still java now also.


it might be that some pipelines will want code like this anyway.. it would
help them too once done.

codefromthecrypt:
@drolando sorry to answer you directly. don't worry the UI will have all logic it needs. There's no difference in the query response from the server

codefromthecrypt:
ok rebased over a bunch of internal work. The current code is not correct in any way, but I think I can make it correct now :) IOTW don't pay attention to current code on this branch.. it will be redone with the new Trace.merge stuff

zipkin-ui/js/skew.js
```diff
@@ -68,6 +68,11 @@ class Node {
   }
 }
 
+// In javascript, dict keys can't be objects
+function keyString(id, shared) {
```
codefromthecrypt: notable (id, shared) is not enough. This has to be kindof different. will think on a plane trip tonight about it

codefromthecrypt:
especially now with special casing due to shared IDs, it is getting strange to keep this generic. we only use it for spans. I'm finding myself having to explain awkardly and it is probably more time to make it concrete. I think I originally did this as we had 2 versions of Span (v1.Span v2.Span)

Later I will change the "Node" to SpanNode or TraceNode as it makes documenting a lot easier and less confusing to read or write

zipkin/src/main/java/zipkin2/internal/SpanNode.java
```diff
@@ -0,0 +1,313 @@
...
+    Map<Key, Key> keyToParent = new LinkedHashMap<>();
+
+    /**
+     * Builds a trace tree by merging and processing the input or returns an empty tree.
+     *
+     * <p>While the input can be incomplete or redundant, they must all be a part of the same trace
+     * (e.g. all share the same {@link Span#traceId()}).
+     */
+    public SpanNode build(List<Span> spans) {
+      if (spans.isEmpty()) return new SpanNode(null);
```
codefromthecrypt: self-note make this illegal

zipkin/src/main/java/zipkin2/internal/SpanNode.java
```diff
@@ -0,0 +1,313 @@
...
+      } else if (span.parentId() != null) {
+        // We are not a root span, and not a shared server span. Proceed in most specific to least.
+
+        // We could be the child of a shared server span (ex a local (intermediate) span on the same
+        // endpoint). This is the most specific case, so we try this first.
+        parentKey = new Key(span.parentId(), true, endpoint);
+        if (keyToParent.containsKey(parentKey)) {
+          keyToParent.put(noEndpointKey, parentKey);
+        } else {
+          // Next, prefer the same endpoint in case data was incorrectly send w/o a shared endpoint
```
codefromthecrypt: self-note "send w/o a shared endpoint" rewrite

codefromthecrypt:
going to restore https://github.com/openzipkin/zipkin/blob/610b8bf796c5f91d2b73d8914a48308b2a3786a4/zipkin/src/main/java/zipkin/internal/CorrectForClockSkew.java and https://github.com/openzipkin/zipkin/blob/610b8bf796c5f91d2b73d8914a48308b2a3786a4/zipkin/src/test/java/zipkin/internal/CorrectForClockSkewTest.java to rework the old clock skew correcter to v2 types before porting back to javascript (and deleting the temporary java code version)

codefromthecrypt:
finally clock skew is now v2 in java. last step is porting to javascript

codefromthecrypt:
ok! I think we are ready for a look

zipkin-ui/js/component_ui/traceToMustache.js
```diff
@@ -154,8 +168,9 @@ export function getServiceName(span) { // export for testing
   return allServiceNames.length === 0 ? null : allServiceNames[0];
 }
 
-export default function traceToMustache(trace, logsUrl = undefined) {
-  const t = traceSummary(trace);
+export default function traceToMustache(v2Trace, logsUrl = undefined) {
```
codefromthecrypt, line 171: this page is still v1 format. that will change in a later pull request

zipkin-ui/js/skew.js
```diff
@@ -63,105 +66,170 @@ class Node {
...
 }
 
-/*
- * Some operations do not require the entire span object. This creates a tree given (parent id,
- * id) pairs.
- */
-class TreeBuilder {
+// In javascript, dict keys can't be objects
+function keyString(id, shared = false, endpoint) {
+  const sharedV = shared === true;
```
tacigar: nit: How about `!!shared`?

zipkin-ui/js/skew.js
```diff
@@ -63,105 +66,170 @@ class Node {
...
+   * RPC. If two different servers respond to the same client, the only way for us to tell which
+   * is which is by endpoint. Our goal is to retain full paths across multiple endpoints. Even
+   * though instrumentation should be configured in such a way that a client never sends the same
+   * span ID to multiple servers, it can happen. Accordingly, we index defensively including any
+   * endpoint data that might be available.
+   */
+  _index(span) {
+    let idKey;
+    let parentKey;
+    const shared = span.shared === true; // guards against undefined
```
tacigar: nit: dito

zipkin-ui/js/skew.js
```diff
@@ -63,105 +66,170 @@ class Node {
...
+
+    if (shared) {
+      // we need to classify a shared span by its endpoint in case multiple servers respond to the
+      // same ID sent by the client.
+      idKey = keyString(span.id, true, span.localEndpoint);
```
tacigar, line 110: I think `span.localEndpoint` is a little large for using as key string...
I think that it is better to use a hash value if it is possible... 🤔

zipkin-ui/js/skew.js
```diff
@@ -63,105 +66,170 @@ class Node {
...
+   * For example, if this is a server span, it would be a different node, and a child of its client
+   * even if they share the same span ID.
+   *
+   * Processing is defensive of typical problems in span reporting, such as depth-first. For
+   * example, depth-first reporting implies you can see spans missing their parent. Hence, the
+   * result of processing all spans can be a virtual root node.
+   */
+  _process(span) {
+    const endpoint = span.localEndpoint;
+    const key = keyString(span.id, span.shared, span.localEndpoint);
```
tacigar, line 132: dito

zipkin-ui/js/skew.js
codefromthecrypt: great

zipkin-ui/js/skew.js
codefromthecrypt: maybe.. doesn't seem obvious for a hashCode replacement. do you know one or should I manually write one like we have in java?

zipkin-ui/js/skew.js
```diff
@@ -63,105 +66,170 @@ class Node {
...
+    const noEndpointKey = endpoint ? keyString(span.id, span.shared) : key;
+
+    let parent;
+    if (span.shared === true) {
```
tacigar: nit: How about `if (span.shared) { ... }` ?

zipkin-ui/js/skew.js
```diff
@@ -63,105 +66,170 @@ class Node {
...
     // other words, assume that the first root is the "real" root.
-    if (!parentId && !this._rootNode) {
-      this._rootNode = node;
-      this._rootId = id;
-    } else if (!parentId && this._rootId === id) {
-      this._rootNode.setValue(this._mergeFunction(this._rootNode.value, node.value));
+    if (!parent && !this._rootSpan) {
+      this._rootSpan = node;
+      delete this._spanToParent[noEndpointKey];
+    } else if (span.shared === true) {
```
tacigar: nit: dito

codefromthecrypt:
I think we can't always do span.shared because difference between undefined
and false.

funny but I think you can see some tests fail travis not circleci I suppose
due to different browser.

anyway I prefer simplest so if idea that keeps tests passing good by me!

zipkin-ui/js/skew.js
```diff
@@ -385,54 +350,42 @@ function adjust(node, skewFromParent) {
...
-  } else if (dataError) {
-    if (debug) {
-      console.log(`skipping clock skew adjustment due to data errors: traceId=${traceId}`);
-    }
+  }
+
+  const childrenOfRoot = trace.children;
+  for (let i = 0; i < childrenOfRoot.length; i++) {
+    const next = childrenOfRoot[i].span;
+    if (next.parentId || !!next.shared) continue;
```
tacigar: ~nit: How about `if (next.parentId || next.shared) ` ?~

zipkin-ui/js/spanCleaner.js
```diff
@@ -123,7 +123,7 @@ function compareShared(left, right) {
   if (left === right) {
     return 0;
   } else {
-    return left ? 1 : -1;
+    return left === true ? 1 : -1;
```
tacigar: nit: I think `left ? 1 : -1` is OK.

And sorry, this is not comments for this line, but I noticed wrong codes in L111 & L112, so I add comments.
I think those code should be the following.

```
L111:  if (typeof(left) === 'undefined') return -1;
L112:  if (typeof(right) === 'undefined') return 1;
```

zipkin-ui/js/skew.js
```diff
@@ -192,189 +259,87 @@ function ipsMatch(a, b) {
...
-    const b = binaryAnnotations[i];
-    if (!b.endpoint) continue;
-    if (b.key === 'lc' && ipsMatch(skew.endpoint, b.endpoint)) {
-      const result = Object.assign({}, span);
-      result.timestamp = spanTimestamp - skew.skew;
-      return result;
-    }
-  }
-  return span;
+  return result;
```
tacigar, line 272: Great!
V2 is very simple! 👍

tacigar:
> I think we can't always do span.shared because difference between undefined and false.

Oh, I see...
I think you are completely correct.
Sorry, actually I think that `shared === true` is also valid now.

tacigar:
hmmmm...
Sorry, I thought again, but I couldn't see problems caused by the difference between `undefined` and `false` in this case...
What do you mean? 🤔 

But I think that `shared === true` is OK too 👍

codefromthecrypt:
easy enough to try and simplify. test will say so!

tacigar:
I think tests fail after https://github.com/openzipkin/zipkin/pull/2245/commits/519255f339fcc269a0b6d9938de3a82645ebc1dd.

adriancole:
only in travis but yes. will try to figure out. the code before this was
wrong as dropped some spans

On 23 Nov 2018 12:00, "tacigar" <notifications@github.com> wrote:

I think tests fail after 519255f
<https://github.com/openzipkin/zipkin/commit/519255f339fcc269a0b6d9938de3a82645ebc1dd>
.

—
You are receiving this because you authored the thread.
<https://github.com/openzipkin/zipkin/pull/2245#issuecomment-441156763>,
or mute
the thread

codefromthecrypt:
totally right on span.shared @tacigar

zipkin-ui/js/spanCleaner.js
codefromthecrypt: thanks

tacigar:
Why does test fail... 🤔

codefromthecrypt:
heh travis version of chrome still hates me

tacigar:
😢

codefromthecrypt:
it is an ordering problem I cannot reproduce on my laptop... it is related to the shared flag

```
[INFO] HeadlessChrome 0.0.0 (Linux 0.0.0) mergeV2ById should order spans by shared, timestamp then name FAILED
[INFO] 	AssertionError: expected [ '0000000000000004-true-1',
[INFO] 	  '0000000000000004-false-2',
[INFO] 	  '0000000000000003-false-2',
[INFO] 	  '0000000000000002-false-3' ] to deeply equal [ '0000000000000004-false-2',
[INFO] 	  '0000000000000004-true-1',
[INFO] 	  '0000000000000003-false-2',
[INFO] 	  '0000000000000002-false-3' ]
```

tacigar:
Yeah, I think so.
But I cannot reproduce on my PC, too...

codefromthecrypt:
used FireFox locally to figure out the problem (used browser  `Firefox` with `karma-firefox-launcher` in `zipkin-ui/karma.conf.js`)

codefromthecrypt:
rebased etc and now 🙏

codefromthecrypt:
Thank you so much @tacigar for the help with this!

tacigar:
LGTM!
Sorry for late.

PR #2246, opened 2018-11-14
Adds internal type to merge spans in a trace together

codefromthecrypt:
This merges spans that have the same identity together for the purpose
of display or aggregation. This works around instrumentation that
accidentally or purposefully send spans in pieces.

This work was described in issue #2245. Basically, we have multiple
tools that need a stable algo for merging spans. This approach is a
row-processing one. There is a little redundancy as we have to scan for
corrections before performing them.


zeagord (approved):
LGTM

PR #2248, opened 2018-11-15
Fixes collapse problem in UI

### LINKED ISSUE #1774: Make collapse button work by default
Currently, the UI defaults to expand all, which implicitly selects all services like so..

screen shot 2017-11-01 at 4 01 14 pm

When this is in place, you cannot collapse a the calls under root (clicking on the + just blinks red in the top-left corner).

Conversely, if you click Collapse All, then expand root, you can now collapse the child

screen shot 2017-11-01 at 4 02 47 pm

If you look closely, the difference is that in the Collapse All, both services are not highlighted. Looking at the trace.js, hide is muted when the service is in the filtered list

```javascript
function hideSpan(span) {
  if (span.inFilters > 0 || span.openChildren > 0 || span.openParents > 0) return;
  span.hide();
}
``` 

Probably, the default should be similar to clicking Collapse All, click + as intuitively, someone would like to be able to hide parts of the tree. This is especially important in large traces. Thanks to @sigerber for reporting this!

codefromthecrypt:
tested with all the zipkin-ui/testdata

Fixes #1774

### DIFF
--- a/zipkin-ui/js/component_ui/trace.js
+++ b/zipkin-ui/js/component_ui/trace.js
@@ -35,7 +35,7 @@ export function showSpans(spans, parents, children, selectedSpans) {
 }
 
 function hideSpan(span) {
-  if (span.inFilters > 0 || span.openChildren > 0 || span.openParents > 0) return;
+  if (span.openChildren > 0 || span.openParents > 0) return;
   span.hide();
 }
 


drolando:
There's still a bug. The button now collapses all its immediate children that are "leaves", but won't hide children that are "roots" of a subtree.

drolando:
You can reproduce that with the `messaging-kafka.json` example trace. If you click on `servicea on-message` (the 2nd span) it doesn't collapse the entire subtree, it only hides a few spans

codefromthecrypt:
thanks for checking!

codefromthecrypt:
considering the UI is to be rewritten in less than a month, not sure how much effort will try on this one

PR #2250, opened 2018-11-15
Adds the gitter link in the question sumary

jcchavezs:
This PR provides the link to gitter and stack overflow before the human clicks on `Get started`.

Ping @beckje01 @adriancole

PR #2254, opened 2018-11-15
Revert "Adds the gitter link in the question sumary"

jcchavezs:
Reverts openzipkin/zipkin#2250

Ping @beckje01 @adriancole


drolando (approved):

drolando:
What about just adding the url for gitter? Would that fit?

codefromthecrypt:
added a commit for plain links

drolando:
That string is still more than 200 chars...

```python
>>> len("This issue tracker is not the place for questions. If you want to ask how to do something, or to understand why something isn't working the way you expect it to, please use https://gitter.im/openzipkin/zipkin or https://stackoverflow.com/questions/tagged/zipkin.")
262
```

codefromthecrypt:
maybe we just give gitter link then..

codefromthecrypt:
thanks for the watchdog service @beckje01 @drolando!

beckje01:
Looks good now

Issue #2208, opened 2018-10-10
UI enhancement: allow different units for duration input


shakuzen:
Currently, the duration input field on the UI forces users to use microseconds as the unit. Microseconds is not a familiar time unit for most people, I assume. We even convert trace durations displayed in trace search results into more familiar units when the duration is long enough (ms or s). I think we can provide a better experience by allowing users to input different units.

Some ideas:
1) dropdown selection of units next to the text field (could still default to microseconds if we want to be least jarring from the status quo)
2) allow specifying of units in the text field while defaulting to microseconds if no units are provided. ex. `1s` for 1 second, `50ms` for 50 milliseconds.

I assume we'll want to restrict the changes to the UI and not touch the actual API, so the UI would need to convert whatever is input to microseconds. If we did want to consider changing the API (in a backwards compatible way, I think), we could look at [Spring's support for `Duration` conversion](https://jira.spring.io/browse/SPR-13721) as long as we can default the units to microseconds.

/cc @zeagord

tacigar:
I prefer 2nd idea! :)

shakuzen:
Resolved by #2260

PR #2260, opened 2018-11-17
Allow different units for duration input

tacigar:
I tried to implement 2nd idea of #2208 .

μs
---
e2fd44d074cb28b784c46bc6fcab9982

ms
---
c2ed6da055f891b75dfd760297cbb1fb

s
---
c85c551de9b1969da64cf54ef14e051f


codefromthecrypt:
cool! looks good


nit maybe the text above no longer needs to say us? maybe it does..

tacigar:
Yes.
I considered microseconds as the default unit and treat durations as microsecond if we do not specify a unit.
But currently microsecond units can not be explicitly specified.
I will change this...

tacigar:
Support μs units.

cb87ac65f3c3d9f1203cc4f974fbe3b2

drolando:
Lgtm!

Can you also add a tip in the textbox showing the possible options?

codefromthecrypt:
screen shot 2018-11-19 at 10 14 18 am
@drolando how about this? there's not much room..

codefromthecrypt:
just added a commit

tacigar:
Great!
Thank you~ 👍

codefromthecrypt:
@drolando if you want change after no problem. hopefully we don't have to describe traces that took minutes or hours too much :D

zipkin-ui/js/component_data/default.js
```diff
@@ -6,8 +6,31 @@ import {traceSummary, traceSummariesToMustache} from '../component_ui/traceSumma
 import {mergeV2ById} from '../spanCleaner';
 import {SPAN_V1} from '../spanConverter';
 
+export function convertDurationToMicrosecond(duration) {
+  const match = duration.match(/^(\d+)(μs|ms|s)$/i);
+  if (match) {
+    const unit = match[2];
+    switch (unit) {
+      case 'μs':
```
shakuzen, line 14: Not that I imagine I'll be using microsecond units, as not having to use that was the point of the change, but I think many people don't know how to easily type `μ` on their keyboard. What if we also accept `us` for microseconds?

codefromthecrypt: good point.. I think lack of suffix should be μs. Is it not already the case? good suggestion on us as well.

tacigar: yes, lack of suffix will be μs now.
but I think `us` is also good.
I think we should support both of them.
I'll make another PR.

PR #2263, opened 2018-11-20
Conditionally filters when strictTraceId == true

### LINKED ISSUE #1973: Skip filtering traces from result when STRICT_TRACE_ID is set to true

codefromthecrypt:
In order to read data written before switching from strictTraceId == false, we
filter client side even when strictTraceId == true. This broke some behavior
Netflix relied on that allowed multi-tag based query on "fit" experiments.

There are a few ways to address this, including adding another flag to say if
you have any old data or not. In this case, you'd never want to filter client-
side. OTOH introducing a new flag is burdensome from a documentation POV.

This addresses the concern in a different way. The only reason we would have to
filter client side, is if there is a clash on the lower 64-bits of the trace ID.
Knowing this, we can achieve the same with no configuration just by looking for
a clash and doing no filtering if there is no clash.

Fixes #1973

### DIFF
--- a/zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/SelectFromTraces.java
+++ b/zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/SelectFromTraces.java
@@ -28,6 +28,7 @@ import java.util.function.BiConsumer;
 import java.util.function.Supplier;
 import zipkin2.Call;
 import zipkin2.Span;
+import zipkin2.internal.FilterTraces;
 import zipkin2.internal.HexCodec;
 import zipkin2.internal.Nullable;
 import zipkin2.internal.V1ThriftSpanReader;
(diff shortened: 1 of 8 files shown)


codefromthecrypt:
hmm cassandra docker test might be legit broke :P will look more carefully

codefromthecrypt:
[ERROR] getTraces_minDuration(zipkin2.storage.cassandra.ITCassandraStorage$ITSpanStore)  Time elapsed: 1.751 s  <<< FAILURE!

fails locally as well...

codefromthecrypt:
I have to redo how this works because cassandra impl always needs to filter due to how trace IDs are indexed

codefromthecrypt:
I was wrong.. this has nothing to do with cassandra. I had a bug in my impl

codefromthecrypt:
ergh.. there might be a behaviour needed still..

codefromthecrypt:
gonna wait until travis passes or fails.. it is too time consuming to keep running the cassandra tests and I have other things to do today

PR #2264, opened 2018-11-20
Fixes Expand All and Collapse All

zeagord:
cc @tacigar


codefromthecrypt:
formatting nit on Expand All

codefromthecrypt:
looks good.. if you have time for before after otherwise I dont mind
checking my installing local

@tacigar curious what you use to make animated gif for before and after..

zeagord:
**Before**
screen shot 2561-11-20 at 18 00 34

**After**
screen shot 2561-11-20 at 18 00 22

zipkin-ui/templates/trace.mustache
```diff
@@ -23,8 +23,11 @@
 
   <form class='form-row mt-3' role='form'>
     <div class='form-group col-md-3' id='filterAllServices'>
-      <button type='button' class='btn btn-primary active' value='uiExpandAllSpans' data-i18n="trace.expand">Expand All</button>
-      <button type='button' class='btn btn-primary' value='uiCollapseAllSpans' data-i18n="trace.collapse">Collapse All</button>
+      <button id= 'expandAll' type='button' class='btn btn-primary' value='uiExpandAllSpans'
+      data-i18n="trace
```
codefromthecrypt: formatting.. not sure if it will hurt or not, but guessing the newline might

codefromthecrypt (approved):
LG one nit

tacigar:
This PR seems a solution for #1774.
And I think this PR make `Collapse All` as default, right?

zeagord:
Yes, this makes "collapse all" by default

codefromthecrypt:
there is always debate about whether collapse all, expand all etc, is good or not

the original logic was to only expand spans matching the service from the previous screen. However, we have to consider now we can select all services in the query. Also, one can navigate to a trace directly. It could make sense to expand the service if one was selected in the prior screen, still, though.

codefromthecrypt:
PS regardless of last commentary, expand and collapse being buggy is far more important than the behaviour :) I think people will like a not buggy screen more than they care about default

tacigar:
> Yes, this makes "collapse all" by default

OK.

>there is always debate about whether collapse all, expand all etc, is good or not

I see...

I agree that `Collapse All` with a less than 4 nested traces go well.
But I tried `Collapse All` with a deep nested (4 nested) trace, it did not go well... :sob: 
(I could not hide the root's child span)
ezgif-4-d174882667f2
So I guess essential problems haven't been solved yet... :cry:

tacigar:
I think there are some problems in `component_ui/trace.js`
For example, in https://github.com/openzipkin/zipkin/blob/510434c88ecd1750d5a36164d46c06e3e2958b07/zipkin-ui/js/component_ui/trace.js#L56-L59 I think the code should be like the following.

```
    $.each(children[$selected.id], (j, cId) => {
      family.add(cId);
      $selected.openChildren -= 1;
      spans[cId].openParents -= 1;
      spans[cId].inFilters -= 1;
    });
```
I think the essential problems are in these processes.

drolando:
> Yes, this makes "collapse all" by default

Can you make this configurable? It shouldn't be too hard.

I'm pretty sure that defaulting to everything collapsed I'd get a ton of questions from developers asking why there's only one span and if zipkin is broken. Given that the UI has defaulted to everything expanded for a long time (and that most people don't even know those 2 button exists) changing it to collapsed would be an unexpected change.

drolando:
> the original logic was to only expand spans matching the service from the previous screen.

I'm -1 on this. That a button called "expand ALL" only expands some subtrees is very unintuitive. If people want this I'd add a third button "expand service" or something like that.

---------------------

@zeagord thanks for trying to fix this!

codefromthecrypt:
@drolando  on this:

> the original logic was to only expand spans matching the service from the previous screen.
>
> I'm -1 on this. That a button called "expand ALL" only expands some subtrees is very >unintuitive. If people want this I'd add a third button "expand service" or something like that.

that wasn't the suggestion at all. What I meant was when navigating to the screen, and service "edge" was what was queried for, the edge services were expanded IIRC. I agree it would be senseless to have an explicit press of a button called "expand ALL" result in not expanding all. I was referring to the act of navigation.

codefromthecrypt:
> I'm pretty sure that defaulting to everything collapsed I'd get a ton of questions from developers asking why there's only one span and if zipkin is broken. Given that the UI has defaulted to everything expanded for a long time (and that most people don't even know those 2 button exists) changing it to collapsed would be an unexpected change.

configuration is a burden, it needs to be carried over to the new UI etc. I don't think many UIs, if any, default to collapse all. We just default to what we had originally (expand service highlighted or all if no service), or default to expand all.

This was intending to fix a bug, not change the feature in other words.

codefromthecrypt:
Thanks for taking a stab @zeagord all the feedback folks! Great to hear people championing for good :)

zeagord:
Still filters need to fixed and more tests. 

expandall

zeagord:
Removed filter service by label and dropdown

codefromthecrypt:
so clarifying default is expand all looks good.

there is one dangling format concern I mentioned which makes me nervous
about the i18n plugin. might want to look at that.

codefromthecrypt:
I'll test locally and fix the lint in a bit

codefromthecrypt:
ok last thing this needs to to make expand all back to default. then good to go!

zeagord:
Sure. I will do it and fix the i18n along with it.

zeagord:
Added font-awesome icon for expand 
screen shot 2561-11-23 at 11 12 02

codefromthecrypt:
fails like this:
```
[INFO] 	TypeError: $selected.$expander.html is not a function
[INFO] 	    at Object.<anonymous> (webpack:///js/component_ui/trace.js:15:24 <- test/component_ui/trace.test.js:49605:25)
[INFO] 	    at Function.each (webpack:///node_modules/jquery/dist/jquery.js:360 <- test/component_ui/trace.test.js:483:19)
[INFO] 	    at showSpans (webpack:///js/component_ui/trace.js:11:4 <- test/component_ui/trace.test.js:49601:20)
[INFO] 	    at Context.<anonymous> (webpack:///test/component_ui/trace.test.js:181:14 <- test/component_ui/trace.test.js:49495:26)
```

codefromthecrypt:
thanks so much!

tacigar:
Cool!

PR #2267, opened 2018-11-22
Moves UI tests to v2 data, fixes a couple glitches

codefromthecrypt:
This fixes a glitch that causes NaN to show in the trace summary. It
also adds some TODOs which will be fixed when the new code merges that
uses v2 native data.

Before (netflix trace):

screen shot 2018-11-22 at 3 11 00 pm

After
screen shot 2018-11-22 at 3 10 55 pm

### DIFF
--- a/zipkin-ui/js/component_ui/traceSummary.js
+++ b/zipkin-ui/js/component_ui/traceSummary.js
@@ -42,72 +42,6 @@ export function getServiceNames(span) {
       .uniq().value();
 }
 
-function findServiceNameForBinaryAnnotation(span, key) {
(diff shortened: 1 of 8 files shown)


zipkin-ui/js/component_ui/traceSummary.js
```diff
@@ -159,15 +88,13 @@ export function traceSummary(trace = []) {
   const timestamp = trace[0].timestamp;
   const duration = traceDuration(trace);
   const groupedTimestamps = getGroupedTimestamps(trace);
-  const endpoints = _(trace).flatMap(endpointsForSpan).uniqWith(endpointEquals).value();
```
codefromthecrypt, line 162: nothing ever read this!

PR #2272, opened 2018-11-24
Corrects counts in trace detail screen

codefromthecrypt:
While I don't currently agree with the counts being helpful. This fixes
the counts such that they add up. Before, they were count of unique span
ID, not actual spans. This is confusing..

More work to follow, but this is enough to review on its own.


zipkin-ui/js/component_ui/traceToMustache.js
```diff
@@ -168,9 +168,19 @@ export function getServiceName(span) { // export for testing
   return allServiceNames.length === 0 ? null : allServiceNames[0];
 }
 
-export default function traceToMustache(v2Trace, logsUrl = undefined) {
+export default function traceToMustache(tree, logsUrl = undefined) {
+  // TODO: this is the max depth of spans, but it probably isn't all that useful vs endpoint depth
```
codefromthecrypt, line 172: cc @ui interesting nuance here.. remember zipkin was originally written to only have remote spans. local (intermediate) spans were added later. This accidentally degraded the usefulness of what originally was equivelant to remote span depth to .. well a number :) cc also @drolando as I don't remember if you are on ui list or not.

codefromthecrypt: cc also @vsen as you may have same concern in haystack. basically I'm not sure the value of tree depth when the tree is any kind of span. Seems more important to have depth of remote links.

zipkin-ui/js/component_ui/traceToMustache.js
```diff
@@ -181,8 +191,6 @@ export default function traceToMustache(v2Trace, logsUrl = undefined) {
   const traceTimestamp = trace[0].timestamp || 0;
   const spanDepths = toSpanDepths(trace);
```
codefromthecrypt, line 192: fyi the logic below I will rewrite in a different PR after I figure out what it is doing.

zipkin-ui/js/spanNode.js
```diff
@@ -0,0 +1,272 @@
...
+      const children = current.children;
+      for (let i = 0; i < children.length; i++) {
+        queue.push(children[i]);
+      }
+    }
+    return result;
+  }
+
+  // Returns the depth of the tree resulting from a depth-first traversal at this node
+  maxDepth() {
```
codefromthecrypt, line 69: this is new.. the rest of the file is extracted from another with no change

zipkin-ui/test/component_data/trace.test.js
```diff
@@ -453,8 +453,8 @@ describe('convertSuccessResponse', () => {
       traceId: '1e223ff1f80f1c69',
       duration: '99.411ms',
       services: 2,
-      depth: 3,
-      spanCount: 3,
+      depth: 4,
```
codefromthecrypt, line 456: look below! see how we used to have serviceNameAndSpanCounts summing to a different count than spanCount!

zipkin-ui/test/spanNode.test.js
```diff
@@ -0,0 +1,173 @@
...
+   *
+   *          a
+   *        / | \
+   *       b  c  d
+   *      /|\
+   *     e f g
+   *          \
+   *           h
+   */
+  it('should report correct max depth', () => {
```
codefromthecrypt, line 77: these two tests are new. the rest of the file is extracted

PR #2273, opened 2018-11-24
Fix back-to-top button

tacigar:
before
current

after
fixed

### DIFF
--- a/zipkin-ui/js/component_ui/traces.js
+++ b/zipkin-ui/js/component_ui/traces.js
@@ -17,7 +17,6 @@ export default component(function traces() {
     if (this.sortFunctions.hasOwnProperty(data.order)) {
       this.$traces.sort(this.sortFunctions[data.order]);
       this.$node.html(this.$traces);
-      this.triggerUpdateTraces();
     }
   };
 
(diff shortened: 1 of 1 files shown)


zipkin-ui/js/component_ui/traces.js
```diff
@@ -29,8 +28,6 @@ export default component(function traces() {
       this.timestamp = parseInt($this.data('timestamp'), 10);
       this.percentage = parseInt($this.data('servicePercentage'), 10);
     });
-    this.on(document, 'uiAddServiceNameFilter', this.addFilter);
```
tacigar, line 32: `addFilter` and `removeFilter` are removed when filter features are removed.
So this process does not finish successfully.
The process of this line seems to be called in https://github.com/openzipkin/zipkin/blob/2e32676489acd3c807cfcc57a6e0e456a00fd607/zipkin-ui/js/page/default.js#L89 , so later processes of this line will not be executed.
So https://github.com/openzipkin/zipkin/blob/2e32676489acd3c807cfcc57a6e0e456a00fd607/zipkin-ui/js/page/default.js#L92 is not called, and then https://github.com/openzipkin/zipkin/blob/2e32676489acd3c807cfcc57a6e0e456a00fd607/zipkin-ui/js/component_ui/backToTop.js#L11 is not called, too.

As a result, this callback https://github.com/openzipkin/zipkin/blob/2e32676489acd3c807cfcc57a6e0e456a00fd607/zipkin-ui/js/component_ui/backToTop.js#L13 is not set... 😢 

This is the cause of this bug.

zipkin-ui/js/component_ui/traces.js
```diff
@@ -17,7 +17,6 @@ export default component(function traces() {
     if (this.sortFunctions.hasOwnProperty(data.order)) {
       this.$traces.sort(this.sortFunctions[data.order]);
       this.$node.html(this.$traces);
-      this.triggerUpdateTraces();
```
tacigar, line 20: There is not this method now

codefromthecrypt:
you are a detective!

codefromthecrypt:
thank you again for the close eyes and able hands @tacigar!

Issue #2274, opened 2018-11-26
Miscategorization of service labels


codefromthecrypt:
The current code in the UI prefers the remote service name to the local service name. This is incorrect as the local service name should always be preferred. For example, a client can mislabel the remote side.

That said, there are cases where we want to read the remote service names. For example, root spans who were originated by uninstrumented clients (such as kinesis), or leaf client spans which are calling uninstrumented services (such as s3 or another SaaS). Dependency linker does this, for example. It will read the remote service names for the root and leaves.

One way we can address this is to prefer the remote name only in these use cases. Another way is to display both in these use cases. In the current UI, I think we can only choose to prefer one. In the next UI (project lens), it may be possible to present both. I think haystack-ui presents both for all cases.

codefromthecrypt:
cc @openzipkin/ui

codefromthecrypt:
relates to #1794 

so I think what we can do is float the topic of prefer local vs remote until the rest of v2 is done, preferring local for now. 

I think best case is that the mustache template has a remoteServiceName iff it is a root or leaf
then the UI logic can do whatever it wants. This sort of decoration will be easier when v2 is done.

drolando:
Can you elaborate a bit more on what you mean by "remote service name" and "local service name"?

There are 3 different service names (maybe 4) in a RPC call:
1) CLIENT local service name
2) CLIENT remote service name
3) SERVER local service name

Which one would you show in the UI? I think the one that makes the most sense is 3, using 2 in case there's no server span. Which I believe it's the current UI behavior.

codefromthecrypt:
3 is best. it reflects the same logic as dependency graph. we still have an
issue to help identify remote uninstrumented callers or receivers but I
think we should do that without removing the ability to label the server
correctly.

codefromthecrypt:
fyi @fyi-coursera opened #2345 just now choosing to display only the remote service name when a client-only trace (which would be very weird in practice unless a bug).

I think we had discussions in the past about potentially showing two labels similar to I think how haystack works if I recall. I'm not sure if we want to special-case single-span traces in any way. 

However, I do think that the bottom of the tree (leaves which are clients) should have handling.

codefromthecrypt:
copying discussion from  #2345 as it is very hard to follow decisions and rationale when discussion occurs with partial context in multiple issues.

from @fyi-coursera
> This PR makes the UI use the client remote service name instead of the client local service name if no server local service name exists.

from @drolando 
[...]

PR #2276, opened 2018-11-26
Converts v1 binary annotation conversion logic to mustache format

codefromthecrypt:
Currently, v1 conversion is only a temporary stage on the way to
mustache. This starts the "v2 native" part of the trace detail screen
by skipping right into mustache. Later, we will rename Binary Annotation
to Tag. Another later change is to do the same for annotation, then
refactor out remaining indirection.

See #2274 as this also simplifies service name selection logic,
preferring the authoritative (local) one.

### DIFF
--- a/zipkin-ui/js/component_ui/trace.js
+++ b/zipkin-ui/js/component_ui/trace.js
@@ -91,6 +91,7 @@ export function initSpans($node) {
   const spans = {};
   const children = {};
   const parents = {};
+  // this includes both local and remote service names
   const spansByService = {};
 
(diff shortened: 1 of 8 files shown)


zeagord (commented):
Looks good to me. Tested with the uninstrumented spans like kafka and db traces and the changes prefers the local service name over remote service name.

codefromthecrypt:
#2274 has some follow-up about the service label ranking, which I think will be far easier once all code is in v2, particularly in a tree format

PR #2277, opened 2018-11-26
Removes a UI transformation step processing span timeline

codefromthecrypt:
This converts zipkin v2 timestamp, duration and annotations to mustache
annotations without an intermediate step. This has some interesting
advantages:

* More clear which events are derived
* Less reliance on keywords like "sr", in favor of XXX Start, XXX Finish
* A future change can clarify the mustache event list as a timeline
   (instead of annotaions which is imprecise)

Before:
screen shot 2018-11-26 at 7 03 01 pm

After:
screen shot 2018-11-26 at 7 03 06 pm


codefromthecrypt:
I'll raise alert to @openzipkin/core and @openzipkin/ui as this change decides to be explicit about lifecycle vs annotation. Some may prefer we continue to map the "sr" etc style names from v1, but anyway please consider and raise your POV if you have any strong feelings.

codefromthecrypt:
Here's a great reason why start/finish is better.. you might recall we had to do some hacks to backport messaging to v1 format. We no longer have to play this game cc @ImFlog @jeqo @jonathan-lo

Before:
screen shot 2018-11-27 at 7 52 03 am

Now: we don't need to hijack wire send, which is very helpful as the instrumentation might have added it! cc also @anuraaga who is recently using this.
After:
screen shot 2018-11-27 at 7 51 58 am

codefromthecrypt:
For the sake of getting work complete before we need it next week, I'll merge this. If after the fact, folks want me to revert the names I'll be ok doing that as it will be somewhat mechanical.

PR #2279, opened 2018-11-27
Makes UI merge logic lenient to missing shared flag on server side

codefromthecrypt:
Last PR, I cheated by backfilling missing shared flag on netflix data.
This, removes it as now the logic can work around missing shared flag
under the most common circumstances.

cc @narayaruna

### DIFF
--- a/zipkin-ui/js/spanCleaner.js
+++ b/zipkin-ui/js/spanCleaner.js
@@ -101,6 +101,10 @@ export function compare(a, b) {
   return (a > b) - (a < b);
 }
 
+function isUndefined(ref) {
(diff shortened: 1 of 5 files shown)


codefromthecrypt:
tested by viewing the netflix trace with the missing shared flag and it visualizes properly

codefromthecrypt:
CI glitch is likely due to difference between travis and circleci chrome. Will use FireFox locally to investigate

codefromthecrypt:
using FireFox locally to figure out the problem (browser Firefox with karma-firefox-launcher in zipkin-ui/karma.conf.js)

PR #2280, opened 2018-11-27
Deduplicates annotation processing logic

codefromthecrypt:


zipkin-ui/test/spanConverter.test.js
```diff
@@ -1286,3 +1286,56 @@ describe('SPAN v1 merge by ID', () => {
     ]);
   });
 });
+
+describe('formatEndpoint', () => {
```
codefromthecrypt, line 1290: no change in tests... just moved from the other file

PR #2282, opened 2018-11-27
Fix bugs in UI skew logic relating to headless traces

codefromthecrypt:


zipkin-ui/test/component_ui/traceToMustache.test.js
```diff
@@ -7,30 +7,46 @@ import {httpTrace} from './traceTestHelpers';
...
+    const headless = new SpanNode(); // headless as there's no root span
+
+    // make a copy of the cleaned http trace as adding a child is a mutation
+    treeCorrectedForClockSkew(httpTrace).children.forEach(child => headless.addChild(child));
+
+    const {traceId, duration, services, depth, spanCount, serviceNameAndSpanCounts} =
+      traceToMustache(headless);
+
+    traceId.should.equal('bb1f0e21882325b8');
+    duration.should.equal('111.121ms'); // TODO: is this right?
```
codefromthecrypt: I need to check this out.. I suspect it will not look right at all :P

codefromthecrypt: it is fine

PR #2285, opened 2018-11-27
Adds UI test data for SmartThings OAuth Authorization flow.

llinder:


codefromthecrypt:
I see nothing! ps when you submit can you change to v2 format? easy way is to post local to v1 endpoint then GET on the v2

llinder:
Ha guess it would help if I saved the file before making the pull request!

codefromthecrypt:
thanks tons lance!

codefromthecrypt:
context was @kellabyte was asking about a sample. Here's the chat description from gitter:

> To describe it in a few sentences… we add trace propegation across web redirects to tie several systems together, especially when they involve round trips to other web sites we don’t own. Further we add user IDs to these traces to investigate specific customer issues or complaints. Lastly we also embed the trace context in things like oauth authorization codes so we can tie a full user actions together even when part of that action is performed on backend systems we don’t own.

PR #2294, opened 2018-11-28
Tightens UI logic so we don't have to consider empty span nodes

codefromthecrypt:


codefromthecrypt:
saving power by merging now verifying the zipkin-ui part passed :P

PR #2298, opened 2018-11-30
Hunts down all the things that didn't actually rely on empty variables

codefromthecrypt:
It was really difficult to tell what data we need to convert when we are
producing a lot of data explicitly marked null or empty.


zipkin-ui/js/component_ui/trace.js
```diff
@@ -179,6 +178,11 @@ export default component(function trace() {
     }
   };
 
+  this.originalDuration = function() {
+    const markerText = $('#timeLabel-backup .time-marker-5').text();
+    return markerText ? parseFloat(markerText) : 0;
```
codefromthecrypt, line 183: before we were getting NaN

PR #2301, opened 2018-11-30
Uses SpanNode to generate the trace tree for the detail screen

codefromthecrypt:
There's still more work to do, but this removes some redundant tree
walking code.

### DIFF
--- a/zipkin-ui/js/component_ui/traceToMustache.js
+++ b/zipkin-ui/js/component_ui/traceToMustache.js
@@ -1,75 +1,5 @@
-import _ from 'lodash';
 import {getTraceDuration, mkDurationStr} from './traceSummary';
 import {SPAN_V1} from '../spanConverter';
-
-export function getRootSpans(spans) {
-  const ids = spans.map((s) => s.id);
-  return spans.filter((s) => ids.indexOf(s.parentId) === -1);
-}
-
-function compareSpan(s1, s2) {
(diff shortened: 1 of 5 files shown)


zipkin-ui/js/spanNode.js
```diff
@@ -65,40 +65,6 @@ class SpanNode {
     return result;
   }
 
-  // Returns the depth of the tree resulting from a depth-first traversal at this node
-  maxDepth() {
```
codefromthecrypt, line 69: @tacigar this might be useful for the new UI. the current one is still one row per ID not per span. So can't use this logic really..

zipkin-ui/js/component_ui/traceToMustache.js
```diff
@@ -103,96 +33,116 @@ function incrementServiceNameCount(span, results) {
...
+    let current = queue.shift();
+
+    let span = SPAN_V1.convert(current.span);
+
+    // This is more than a normal tree traversal, as we are merging any server spans that share the
+    // same ID. When that's the case, we pull up any of their children as if they are our own.
+    const childIds = [];
+    current.children.forEach(child => {
+      if (child.span.id === span.id) {
+        span = SPAN_V1.merge(span, SPAN_V1.convert(child.span));
```
codefromthecrypt, line 79: ps in a later change we can even pull up servers to fix the single host span issue

PR #2302, opened 2018-11-30
WIP: merge single-host spans in the UI

### LINKED ISSUE #1799: zipkin-ui does not support non-shared spans
Feature Request

With the advent of the X-Ray compatible modes for zipkin/brave comes the fact that X-Ray spans (segments) are not shared by client and server so the UI ends up losing the hierarchy information. It still orders correctly because of timestamps. I would be great to support this model of spans in the native zipkin UI

codefromthecrypt:
Finally we can fix #1799 not that the UI code uses a tree internally

Consider you are using single-host spans (ex disabling join), with this change, the UI can merge them together as if they were in the same span. The new UI (Zipkin Lens) will likely handle this differently.. this is just a stop-gap.

Note: the UI currently re-parents the children of the server to the client span. I hypothesise most won't notice (as they don't l look at parentId in the UI only span Id). This is the cheapest way to support the feature, but we can do more if important. Do mention if it is.

cc @openzipkin/ui @openzipkin/core 

Before:
screen shot 2018-12-01 at 9 20 01 am

screen shot 2018-12-01 at 9 26 21 am
screen shot 2018-12-01 at 9 26 15 am

After:
screen shot 2018-12-01 at 9 19 41 am
screen shot 2018-12-01 at 9 25 49 am

### DIFF
--- a/zipkin-ui/js/component_ui/traceToMustache.js
+++ b/zipkin-ui/js/component_ui/traceToMustache.js
@@ -76,8 +76,22 @@ export function traceToMustache(root, logsUrl) {
     const childIds = [];
     current.children.forEach(child => {
       if (child.span.id === span.id) {
+        // merge any spans with the same ID as the current UI has a single ID per row
         span = SPAN_V1.merge(span, SPAN_V1.convert(child.span));
         child.children.forEach(grandChild => {
+          queue.push(grandChild);
+          childIds.push(grandChild.span.id);
+        });
+      } else if (current.span.kind === 'CLIENT' && child.span.kind === 'SERVER') {
+        // merge any child that's a server, and reparent its kids accordingly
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
will fix add tests etc after a few votes come in

bogdandrutu:
@adriancole why not modify the client span-id to be server span-id?

codefromthecrypt:
this is only internal to UI because javascript needs string key. the ids
here are used for collapse feature and we need a primary ID string per row.
could make that a concatenation as it wont matter.

@drolando asked offline to have the "show ids" pane show both client and
server ID. that is fine and will do.

codefromthecrypt:
@tacigar probably better to do this in lens and not in the old UI. sound good?

PR #2304, opened 2018-12-02
Completes refactor of UI to v2 native

### LINKED ISSUE #2217: Refactor UI to internally use the v2 model

codefromthecrypt:
This is the last step where the UI no longer uses any v1 types.

What it does now is a dual-pass where at first v2 spans are collected
into events such that spans that share an ID can be in the same row.

After that, it decorates them with positional details, such as where
to put annotations or how wide the span line should be.

fixes #2217


codefromthecrypt:
@openzipkin/ui so happy this is done.. so much work. I tested by viewing a netflix trace before and after as that's our most complex one.

PR #2306, opened 2018-12-02
SmartThings Traces Test Data

beckje01:
Added another special case trace.
Added description file for explinations of the traces.


zipkin-ui/testdata/testdata-explination.md
```diff
@@ -0,0 +1,15 @@
+smartthings-oauth-authoization.json
```
codefromthecrypt: nit typo in the file enclosing this explination/explanation

codefromthecrypt (commented):
thanks and good idea. It might be fine to just append to the README instead though? could be easier for discovery..wdyt?

beckje01:
I'll redo this into the readme I was thinking it should go there but wasn't sure.

codefromthecrypt:
thanks much!

codefromthecrypt:
Thank you SOOO much it is great to have a big trace like this. We've been waiting a while to have data like this!

once "lens" is out, we can try to add screenshots of trace and/or dependency graph to the README cc @openzipkin/ui

Issue #2236, opened 2018-11-06
UI should support generic and site-specific tags


codefromthecrypt:
**Feature:**
Right now, the "annotation query" section is flexible by allowing people to enter free-form text. However, this is not ideal for some sites. It would be good to allow some tag names to be pre-populated with value auto-completion. Notably, things with low cardinality like the standard "http.route" or a site-specific "phase".

**Rational**
Sometimes people make mistakes in the UI search by using incorrect key/value encoding. Also, folks like LINE and also Expedia already know there is a need for pre-defined site specific tags. This was discussed at length at the LINE workshop, who made a new UI that supports this.

https://cwiki.apache.org/confluence/display/ZIPKIN/2018-10-29+Zipkin+UI+at+LINE+Tokyo

This was also previously discussed at Netflix who have some site-specific tags that help correlate with Spinnaker deployments: https://cwiki.apache.org/confluence/display/ZIPKIN/2018-07-18+Aggregation+and+Analysis+at+Netflix

**Example Scenario**
For example, LINE has a phase tag, which is similar to an environment. This allows the user to easily reduce results without thinking about search syntax or making mistakes.

**Prior Art**
 * https://github.com/ExpediaDotCom/haystack-ui

tacigar:
I think the first step is creating a new API for fetching annotations like the following.

* **endpoint**: `/api/v2/annotations`
* **response format**: 
```
{
  "serviceName1": { "annotation1", "annotation2", "key1=value1", ... },
  "serviceName2": ...,
  ...
}
```
Of course, I think that it is good to fetch at the timing when the service name is changed, like span names.
However, since the default value of service name is `all` and I think that annotation query suggestion should be valid in this case, so I think that it is necessary to fetch annotations of all service names after all.
So I think fetching all annotations of all services at first is better.
After that, with the change of the service name as a trigger, we can filter these data at local.

tacigar:
The second step is about UI, but I do not know the component library to do such functions well... :cry:

codefromthecrypt:
> The second step is about UI, but since I do not know the library to
> express such functions well... 😢

you seem to express them in code well enough ;)

codefromthecrypt:
as annotations are rarely used (the term annotation in annotationQuery is historical) For the purpose of dropdown I would suggest

/tags the response could indeed be broken down by service, but in this case the value is more simpler as a map per service name

ex

```
GET /api/v2/tags
[...]

Issue #2219, opened 2018-10-25
Determine how to address the next Elasticsearch index pattern break


codefromthecrypt:
Many of you will recall that we had to redo everything last change to the index constraints: [dropping of multi-type](https://github.com/openzipkin/zipkin/issues/1676).

One thing we decided to do to accommodate this was to use a delimited as clearly we need to have separate indexing and retention for spans vs dependency links.

Unfortunately, we seem bit again.. Looks like Elasticsearch will also drop support for the colon pattern soon.

```
[2018-10-24T11:32:00,033][WARN ][o.e.d.c.m.MetaDataCreateIndexService] index or alias name [zipkin:span-2018-10-24] containing ':' is deprecated and will not be supported in Elasticsearch 7.0+
```

First, we should strongly push back on Elasticsearch as this causes another big round of work for us amplified by the ecosystem. It isn't quite cool to have to keep hitting mines like this. The knock on effects include double-reads, data migrations, all sorts of things, it isn't fun especially for a volunteer group.

Anyway, assuming pushback fails we need to address this thing:
* figure out which character is ok now.
* make conditional code (again) to address this not only here, but also in the dependency linker which is completely different codebase
* communicate to custom ingesters about this and why and what to do about it.

xeraa:
Background: https://github.com/elastic/elasticsearch/issues/23892

codefromthecrypt:
@xeraa yeah so I think the extent of this, tell me if I'm wrong, is that all people's stuff, their cleanup jobs, kibana searches etc will now have to do split mode right? How are you communicating to customers the work they have to do? I would like to be able to link to that as this is likely a big chore not just on our codebase.

codefromthecrypt:
I made a note requesting the same to the upstream issue

xeraa:
I might be reading https://github.com/elastic/elasticsearch/pull/26247/files wrong, but this should only apply to the creation of new indices in 7.0+. So you can keep using the old pattern for reads / searches, but it will indeed need a migration path for new indices and searching the combination of old and new indices.

We might not be aware how common colons are in index names, so this is definitely good feedback. With the collision for cross cluster search I'd assume that we want to make this switch, but maybe we can ease the migration path.

Hackish workaround: Configurable separator and existing indices get an alias to conform to the new pattern (though that was a 30s thought process at 3am, so this might totally not work for various reasons).

codefromthecrypt:
[...]

PR #2309, opened 2018-12-03
Adds Cassandra support for Autocomplete tags

zeagord:


zipkin-storage/cassandra/src/main/resources/zipkin2-schema-indexes.cql
```diff
@@ -47,3 +47,17 @@ CREATE CUSTOM INDEX IF NOT EXISTS ON zipkin2.span (l_service) USING 'org.apache.
 
 CREATE CUSTOM INDEX IF NOT EXISTS ON zipkin2.trace_by_service_span (duration) USING 'org.apache.cassandra.index.sasi.SASIIndex'
    WITH OPTIONS = {'mode': 'PREFIX'};
+
+CREATE TABLE IF NOT EXISTS zipkin2.tags (
```
codefromthecrypt: I presume this is exact same as service-span

zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/SelectTagKeys.java
```diff
@@ -0,0 +1,82 @@
...
+  @Override
+  protected ResultSetFuture newFuture() {
+    return factory.session.executeAsync(factory.preparedStatement.bind());
+  }
+
+  @Override public Call<ResultSet> clone() {
+    return new SelectTagKeys(factory);
+  }
+
+  static class AccumulateTagsAllResults extends AccumulateAllResults<List<String>> {
```
codefromthecrypt: possible refactor possibility, as I recognise this class :P

zipkin/src/main/java/zipkin2/storage/TagStore.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public interface TagStore {
+
+  /**
+   * Retrieves the list of tag keys
+   */
+  Call<List<String>> getTagKeys();
+
+  /**
+   * Retrieves the list of tag keys
```
codefromthecrypt:
```suggestion
   * Retrieves the list of tag values
```

codefromthecrypt (commented):
good start... next in-memory I think as then we can add some tests

zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/SelectTagKeys.java
```diff
@@ -0,0 +1,82 @@
...
+  static class Factory {
+    final Session session;
+    final PreparedStatement preparedStatement;
+    final AccumulateTagsAllResults accumulateTagsIntoSet =
+      new AccumulateTagsAllResults();
+
+    Factory(Session session) {
+      this.session = session;
+      this.preparedStatement =
+        session.prepare(QueryBuilder.select("key").distinct().from(TABLE_TAGS));
```
michaelsembwever: I don't know what the cardinality of tags will be? But this won't scale.

But, for example, if there's 1 million tags in a Zipkin storage this full-table scan is going to be painful (if not timeout).

jcchavezs:
Right. I would do a progressive query (by pagination but only simple one,
no ordering and deifnirively not with `offset` but with `after`.

Den tir. 4. des. 2018, 09:41 skrev mck <notifications@github.com:

zipkin-storage/cassandra/src/main/resources/zipkin2-schema-indexes.cql
zeagord: Yep. Same as service-span

zeagord:
I have added the in-memory api. I will revisit the cassandra tomorrow and address the above concerns.

zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/SelectTagKeys.java
jcchavezs: Right. I would do a progressive query (by pagination but only simple one, no ordering and deifnirively not with `offset` but with `after`.

codefromthecrypt:
made a comment about fixed cardinality.. we definitely need to document this as it is indeed inappropriate for unbounded. https://github.com/openzipkin/zipkin/issues/2236#issuecomment-444302547

One thing @zeagord and I discussed is initially inheriting the config for the other names (service/span). This is for simplicity. In the future we could add a timestamp/lookback parameter to only fetch the values for a range. However, same problem would apply to service/span so thinking of that later

zeagord:
Need to add tests and revisit the Elastic search design.

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java
```diff
@@ -130,6 +131,19 @@ public String getTrace(@PathVariable String traceIdHex, WebRequest request) thro
     return new String(SpanBytesEncoder.JSON_V2.encodeList(trace), UTF_8);
   }
 
+  @GetMapping(
+    value = "/tagKeys",
+    produces = APPLICATION_JSON_VALUE)
+  public List getTagKeys() throws IOException {
```
tacigar: I think the type of return value should be `List<String>` or `ResponseEntity<List<String>>`.

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java
```diff
@@ -130,6 +131,19 @@ public String getTrace(@PathVariable String traceIdHex, WebRequest request) thro
...
+    return storage.tagStore().getTagKeys().execute();
+  }
```
tacigar: I think should insert empty line.

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java
```diff
@@ -130,6 +131,19 @@ public String getTrace(@PathVariable String traceIdHex, WebRequest request) thro
...
+  @GetMapping(
+    value = "/tagValues",
+    produces = APPLICATION_JSON_VALUE)
+  public List getTagValues(@RequestParam String key) throws IOException {
```
tacigar: dito

tacigar:
In elasticsearch, how about using `_q` instead of changing mapping?

codefromthecrypt:
> reuse _q field in elasticsearch


we can think about it but the performance might be bad.

for example, getting the key names would require an expression I am not
sure how to express unless we hard code the possible key names.

if we hard code the possible key names yes it could work, but we have to
check the performance of scanning all span documents to get the values

codefromthecrypt:
good news is we can try it. actually I think we need to hard code key names
anyway especially in Cassandra.

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java
codefromthecrypt: agree

codefromthecrypt:
so on cassandra (and elasticsearch) we'll need to ensure the "deduper" is in use to avoid thrashing writes.

In both cases, it might be helpful to reverse-engineer the service-span mapping to re-use the same table. ex PRIMARY KEY ((type, key), value) This could make data management in general easier long term.

Since there is a time bomb on elasticsearch #2219, we might want to solve that first before merging this (or at least before cutting a release with it).

Meanwhile, we can allow UI testing to work with static managed list of tags. (ex there may be only several values associated with phase, for example.. so one way is to allow the UI to configure predefined where it is small)

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java
zeagord: Fixed

zeagord:
> is this tagkey because "key" is reserved? I suspect if not easier to just do key/value

Bodyconverters for ES looks for "key" from the result of aggregations which clashes with the key in the tag {k,v}. It could be the name of the aggregation. I will change the name of the aggs alone and see if it works.

zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/SelectTagKeys.java
drolando: If we want to have a whitelist of tags to index and not just index all tags you might get away without doing this at all. And just return the list of whitelisted tags.

zipkin-storage/cassandra-v1/src/main/resources/cassandra-schema-cql3-upgrade-1.txt
```diff
@@ -28,3 +28,7 @@ ALTER TABLE zipkin.service_names
 ALTER TABLE zipkin.traces
     WITH compaction = {'class': 'org.apache.cassandra.db.compaction.DateTieredCompactionStrategy', 'max_window_size_seconds': '86400'}
     AND default_time_to_live =  604800;
+
+ALTER TABLE zipkin.tags
```
drolando: I don't think this is needed since this is a brand new table that already has a default ttl, no?

michaelsembwever:
> In both cases, it might be helpful to reverse-engineer the service-span mapping to re-use the same table. ex PRIMARY KEY ((type, key), value) This could make data management in general easier long term.

if service-span and tags are collapsed to one table it does make full table scans (eg `QueryBuilder.select("key").distinct().from(TABLE__)`) more painful…
i suspect keeping two separate tables is in fact wiser, thanks to @drolando for raising this.

zipkin-storage/cassandra-v1/src/main/resources/cassandra-schema-cql3-upgrade-1.txt
zeagord: I wasn't aware of the reason why this one exists. I will remove it.

zeagord: Remvoed

zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/SelectTagKeys.java
zeagord: Added the whitelist implementation. Let me know your thoughts.

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CassandraStorage.java
```diff
@@ -86,6 +91,11 @@ public Builder searchEnabled(boolean searchEnabled) {
       return this;
     }
 
+    @Override public Builder tag(List<String> keys) {
+      this.tag = keys;
+      return this;
```
codefromthecrypt, line 97: check null

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CassandraSpanConsumer.java
```diff
@@ -80,7 +84,9 @@
         if (span.name() == null) continue;
         insertSpanNames.add(insertSpanName.newInput(serviceName, span.name()));
       }
-
+      if (null != v2.tags() && !v2.tags().isEmpty()) {
```
codefromthecrypt: tags wont be null. I would also expect this to filter key names that actually should be indexed

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CassandraTagStore.java
```diff
@@ -0,0 +1,38 @@
...
+public class CassandraTagStore implements TagStore {
+  private final Session session;
+  private final Call<List<String>> tagKeys;
+  private final SelectTagValues.Factory tagValues;
+  CassandraTagStore(CassandraStorage storage){
+    this.session = storage.session.get();
+    tagKeys = new SelectTagKeys.Factory(session).create();
+    tagValues = new SelectTagValues.Factory(session);
+  }
+  @Override public Call<List<String>> getTagKeys() {
```
codefromthecrypt: no need if keys are immutable (collections.unmodifiable or util.sortedlist)

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CassandraTagStore.java
```diff
@@ -0,0 +1,38 @@
...
+    return tagKeys.clone();
+  }
+
+  @Override public Call<List<String>> getTagKeyValues(String key) {
```
codefromthecrypt: since api is named TagStore redundant to say TagXxx in methods

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CassandraTagStore.java
```diff
@@ -0,0 +1,38 @@
...
+    if (key.isEmpty()) return Call.emptyList();
```
codefromthecrypt: throw exception on null or empty string instead

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/SelectTagKeys.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.function.BiConsumer;
+import java.util.function.Supplier;
+import zipkin2.Call;
+import zipkin2.storage.cassandra.internal.call.AccumulateAllResults;
+import zipkin2.storage.cassandra.internal.call.ResultSetFutureCall;
+
+import static zipkin2.storage.cassandra.v1.Tables.TABLE_TAGS;
+
+public class SelectTagKeys extends ResultSetFutureCall {
+
```
codefromthecrypt: just return static list from config? less code

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/InsertTags.java
```diff
@@ -0,0 +1,148 @@
...
+import java.io.IOException;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.TimeUnit;
+import zipkin2.Call;
+import zipkin2.Callback;
+import zipkin2.storage.cassandra.internal.call.ResultSetFutureCall;
+
+import static zipkin2.storage.cassandra.v1.Tables.TABLE_TAGS;
+
+public class InsertTags extends ResultSetFutureCall {
```
codefromthecrypt: unpublic

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/Tables.java
```diff
@@ -81,5 +82,11 @@
    */
   static final String ANNOTATIONS_INDEX = "annotations_index";
 
+  /**
+   * This index supports get tag keys {@link TagStore#getTagKeys()} and then query for the values
+   * for that key {@link TagStore#getTagKeyValues(String key)}
+   */
+  static final String TABLE_TAGS = "tags";
```
codefromthecrypt: be more specific in names. tags is too generic maybe siteTags

zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/CassandraSpanConsumer.java
```diff
@@ -18,8 +18,10 @@
 import com.datastax.driver.core.utils.UUIDs;
 import com.google.common.base.Preconditions;
 import java.util.ArrayList;
+import java.util.HashMap;
```
codefromthecrypt: always linkedhashmap to prevent test flakes on order

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/InsertTags.java
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  @Override
+  protected void doCancel() {
+    factory.cache.remove(input, InsertTags.this); // invalidate
+    super.doCancel();
+  }
+
+  @Override public Call<ResultSet> clone() {
+    return null;
```
codefromthecrypt: ooops

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/SelectTagValues.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.function.BiConsumer;
+import java.util.function.Supplier;
+import zipkin2.Call;
+import zipkin2.storage.cassandra.internal.call.AccumulateAllResults;
+import zipkin2.storage.cassandra.internal.call.ResultSetFutureCall;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static zipkin2.storage.cassandra.v1.Tables.TABLE_TAGS;
+
+public class SelectTagValues extends ResultSetFutureCall {
```
codefromthecrypt: unpublic

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CassandraTagStore.java
```diff
@@ -0,0 +1,38 @@
...
+ * the License.
+ */
+package zipkin2.storage.cassandra.v1;
+
+import com.datastax.driver.core.Session;
+import java.util.List;
+import zipkin2.Call;
+import zipkin2.storage.TagStore;
+
+public class CassandraTagStore implements TagStore {
```
codefromthecrypt: unpublic

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CassandraTagStore.java
codefromthecrypt: Make this derived from a constant list from config I think.  maybe abstract base class instead of interface

zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/CassandraSpanConsumer.java
```diff
@@ -35,27 +37,33 @@
   private static final long WRITTEN_NAMES_TTL =
       Long.getLong("zipkin2.storage.cassandra.internal.writtenNamesTtl", 60 * 60 * 1000);
 
+  private static final long WRITTEN_TAGS_TTL =
```
codefromthecrypt: fine to reuse other constant I think

codefromthecrypt (commented):
made most of a pass!

zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/CassandraSpanConsumer.java
zeagord: Unused import

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/SelectTagKeys.java
zeagord: I am returning the static list in the TagStore here https://github.com/zeagord/zipkin/blob/b64e7950527d83f49598247ccf0f2385220cff4b/zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/CassandraTagStore.java#L32

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchTagStore.java
```diff
@@ -0,0 +1,77 @@
...
+
+  ElasticsearchTagStore(ElasticsearchStorage es) {
+    this.search = new SearchCallFactory(es.http());
+    this.indexNameFormatter = es.indexNameFormatter();
+    this.searchEnabled = es.searchEnabled();
+    this.namesLookback = es.namesLookback();
+    this.tag = es.tag();
+  }
+
+  @Override public Call<List<String>> getKeys() {
```
codefromthecrypt: if you make an abstract base type for the TagStore, it will be less tempting to make a specific api for this. Then, your base type has a builder which assigns the list tagKeys and then returns that (wrapped in a call to fit the api, but used directly elsewhere.. for example, if asking for values of a tag not in the list, you just return empty). That would be my advice.. note: the builder could be StorageComponent.Builder as it would be unexpected for TagStore to be built independently

codefromthecrypt:
this chops off the basic functionality and will allow the  UI work to start immediately when merged. I can help rework the other commits similarly https://github.com/openzipkin/zipkin/pull/2332

codefromthecrypt:
other storage impls pulled out in #2333 and #2334

codefromthecrypt:
I think this is nearly ready. we need to test the auto-upgrade logic and also update the README files to talk about how autocomplete works

michaelsembwever:
Code changes LGTM.

One comment though: most of the diff is whitespace/code-style changes. For the reviewer there's a huge waste of time reading diffs that have nothing to do with the actual PR. It would be great if those changes where separated out to a separate follow-up commit (still within the PR). That way by reviewing just the first commit of the PR a lot of time could be saved.

codefromthecrypt:
@michaelsembwever sorry about the formatting thing. I was in a rush to get something stable before I turned off internet for the vacation, but that amplified efforts to others.. not sure the better call but I apologize nevertheless. thanks for reviewing despite this.

codefromthecrypt:
FYI travis is failing still on the same tests as last push this needs to be looked into prior to merge

PR #2310, opened 2018-12-04
Reverts accidental inclusion of annotation timestamps in trace plotting

codefromthecrypt:
annotations are relative to span, not a trace. This was a bug. sorry @drolando, as this affected you.


drolando:
Can you add a similar annotation to one of the test traces? That way we might notice of we break this again in the future.

Or add a unit test for this with a comment

codefromthecrypt:
>
> Can you add a similar annotation to one of the test traces? That way we
> might notice of we break this again in the future.
>
> Or add a unit test for this with a comment
>

read my mind.. I am putting together some high-level test for this now

>

PR #2316, opened 2018-12-07
Move zipkin-lens over from https://github.com/line/zipkin-lens

huydx:
# Current Status
We have some steps left before we can merge into master. However, you can test now by checking out the `zipkin-lens` branch

## Server packaged with Lens
The lens UI is bundled into the default zipkin server. If you want to try, you can [build locally](https://github.com/openzipkin/zipkin/blob/master/zipkin-server/README.md#building-locally), then pass the argument `--zipkin.ui.source-root=classpath:zipkin-lens` to enable lens.

Ex.
```bash
$ ./mvnw -Dlicense.skip=true -DskipTests --also-make -pl zipkin-server clean install
$ java -jar ./zipkin-server/target/zipkin-server-*exec.jar --zipkin.ui.source-root=classpath:zipkin-lens
```

# TODO
- [x] Remove code of conduct, contributing details as they are inherited from the main project
- [x] Clarify if we can replace the license headers with the normal ones (as they will later be rewritten with apache ones anyway)
- [x] Integration with spring boot
- [x] Add CI integration for zipkin-lens


codefromthecrypt (approved):

codefromthecrypt:
all good. will merge on green which means people can try snapshot build

PR #2318, opened 2018-12-07
Documents Zipkin Lens development URL.

llinder:
- for backwards compatability, Zipkin Lens router prefix is /zipkin.  This can be confusing during development mode so adding some docs for clarity.


codefromthecrypt:
cc @ScienJus

PR #2322, opened 2018-12-12
Collections.sort replace with more object oriented .sort() approach

igor-suhorukov:


codefromthecrypt:
sorry this is a java 6+ library so we cant use list.sort in the main source
tree.

nice thought though!

igor-suhorukov:
@adriancole thank you for feedback

PR #2324, opened 2018-12-12
Replace indexed loop with for-each java5 language construction

igor-suhorukov:


codefromthecrypt:
Not sure you were thinking about this.. we don't document our rationale for all things, but this will actually cause our core library to be less performant. This library is used in the main path of applications, so overhead in terms of cpu and allocation is important. For this reason, we prefer to use non-allocating form of iteration across a list. List.get(index) for does not require allocation of an iterator object.

In other words: for (int...) vs forEach, the forEach part implicitly new's up an Iterator object even if you don't see it happening in the code.

While not all the code you affected here has a benchmark on it that includes data iterated, running one would show up. Let me know if you'd like more info!

Issue #2323, opened 2018-12-12
Make search component for lens


tacigar:
**Feature:**
Create a more convenient search component for lens.

**Rational**
Make it easier for users to search for traces.

**Example Scenario**
When users search for traces

**Prior Art**
 * https://github.com/ExpediaDotCom/haystack-ui

tacigar:
Done by #2325

PR #2325, opened 2018-12-13
Make search component for lens

tacigar:
related #2323


tacigar:
Currently...
GIF1: 
global-search

GIF2:
global-search-lookback

codefromthecrypt:
If I weren't already married, I would marry your search component

tacigar:
@zeagord Thank you for telling me the typo!

drolando:
@tacigar Can you make it so that when you click on the `+1` the scroll down that lets you select the key appears automatically?

Nice work anyway!

tacigar:
@drolando 
Thank you for feedback!
You mean such behavior, right?
default-open

codefromthecrypt:
I think he meant to only allow this feature if you clicked star on zipkin :D

drolando:
@tacigar Yep!

tacigar:
> I think he meant to only allow this feature if you clicked star on zipkin :D

That's a good idea 😆

tacigar:
> @tacigar Yep!

OK! Thank you!

tacigar:
Currently, the search condition is owned by the state of the GlobalSearch component.
However, I think that search conditions should be in Redux state for making this search component global.
I'll solve this problem when I start the next PR.

codefromthecrypt:
I like this, but I am tempted to type. a big whitespace makes me think this is a search form. If I click into this space, I accidentally create a new search control. it is easy to accidentally make many, and backspace/delete doesn't get rid of them. This might frustrate people who like to use keyboard.

maybe easy to make a lot of search by accident

codefromthecrypt:
also the result conditions might be better to stay on the right.. for example, max results. not sure if others will feel this way, but it seems odd to mix data conditions with filtering ones

codefromthecrypt:
one solution could be to not have the background white! white looks like a text box, so makes me want to type in it

codefromthecrypt:
I really like how much space we have for results, now. It is my favorite part

tacigar:
> also the result conditions might be better to stay on the right.. for example, max results. not sure if others will feel this way, but it seems odd to mix data conditions with filtering ones

OK!
I'll put `limit` and `lookback` condition on the right.

> one solution could be to not have the background white! white looks like a text box, so makes me want to type in it

I see.
I'll rethink the design a bit more.

tacigar:
Thank you for your feedback :+1:

tacigar:
OK, I will do the following for improving design and UX.

- [ ] Change color (as @adriancole says, current color is a bit confusing)
- [ ] When user selects search condition, open option values
- [ ] When user input delete key, delete focused condition component

tacigar:
OK. 
I think that UX related to operation has improved considerably.

tacigar:
I start to improve design.

codefromthecrypt:
hehe "Improve lookback design because I don't like it."

tacigar:
OK. I have improved design a bit.
I start to fix NaN bugs occured in `<input type="number" />`

codefromthecrypt:
noticed this.. is there a missing file? or is this just me?

```
[INFO] /Users/acole/oss/zipkin/zipkin-lens/src/components/Browser/Traces/index.test.js
[INFO]   5:28  error  Unable to resolve path to module './InitialMessage'  import/no-unresolved
[INFO] 
[INFO] ✖ 1 problem (1 error, 0 warnings)
[INFO] 
[ERROR] (node:32732) [ESLINT_LEGACY_OBJECT_REST_SPREAD] DeprecationWarning: The 'parserOptions.ecmaFeatures.experimentalObjectRestSpread' option is deprecated. Use 'parserOptions.ecmaVersion' instead. (found in "node_modules/eslint-config-airbnb-base/index.js")
```

tacigar:
Oops!

tacigar:
before:



after:

drolando:
@tacigar annotationQuery seems broken to me. I never get back any result whatever query I input there

tacigar:
Oh...
In my environment, it works... 🤔 
Could you tell me the API path being called after pressing the search button (like `http://localhost:9000/zipkin/api/v2/traces?serviceName=serviceA&annotationQuery=error....` )

drolando:
@tacigar it works if you only filter on the annotation name but what I want to do is to filter on something like `region=uswest1-prod`. I think that `=` doesn't get urlencoded properly

tacigar:
Oops! That's true!
I'll fix it.
Thank you!

tacigar:
I'll move the search condition state to redux from react component.

tacigar:
About last one.
See: https://github.com/facebook/react/issues/4494

tacigar:
zipkin-browser

tacigar:
At first, I did not intend to change the design of existing components much in this PR.
But I realized that it is essential to consider other components in building GlobalSearch components.
So I decided to add the minimal changes to Browser page & DetailedTrace page components to consider the design of the GlobalSearch component.
But I didn't add any changes to Dependencies page & Timeline components, because these components need a lot of changes in future(e.g Replacing vizceral / Using SVG for timeline, etc...)

zipkin-browser
detailed-trace-summary

tacigar:
@adriancole 
PTAL 🙇

basvanbeek:
@tacigar I'd personally would drop the lighter (white) background all together. It clutters the display unnecessarily. 
By having the + and search icon clearly visible and having a contrasting color for the "Please select the criteria..." the intent of the top bar is already apparent.

tacigar:
@basvanbeek 
Thank you for your feedback!
I tried to change the design a little.
Which do you think is better?

## Current design (white background)

current-1
current-2

## Dark

dark-1
dark-2

## More dark

I personally like this.

more-dark-1
more-dark-2

## No border

Maybe this is not good, because not having a contrasting color for the "Please select the criteria...".

no-border-1
no-border-2

cc @adriancole

codefromthecrypt:
I like no border best

codefromthecrypt:
@tacigar can you rebase and force push the conflicts away :)

codefromthecrypt:
actually no problem I will help

codefromthecrypt:
I think I like this right now. One thing maybe not user friendly is the name "annotationQuery" and similarly the raw labels like "serviceName". I think eventually these should be satisified by i18n

ex "annotationQuery" in english label could be "tag or annotation query"

Other thing is now it seems we have to click the double-down arrow to expand the trace inline. This might be ok as it hints where to go to go out or not. However, I found myself double-clicking the bar eventhough the jump down icon is intuitive..
screen shot 2018-12-31 at 1 38 40 pm

codefromthecrypt:
IMHO minimum change probably is getting rid of the light grey framing of the condition buttons until we figure out something better.

basvanbeek:
I agree with @adriancole in liking the no border option best.

tacigar:
@adriancole @basvanbeek 
Thank you!
I'll do so.

tacigar:
Finally I'll add tests for GlobalSearch

tacigar:
Umm...
I'll revert the design.
The reasons are below.

* If there are no borders, it is ugly when the search condition becomes multiple lines.
* I want to add a button for deleting all search conditions. If there are no borders, there is no place to put it. (If I put it by force, it may be bad)
* if there are no borders, limit condition and lookback condition appear strangely floating on the gray background.

basvanbeek:
Still have an issue with the background cluttering the design. How about instead of a solid background color you do a thin border to box the tags. This also allows the delete icon to be grouped in.

tacigar:
I see...
What do you think about this design?
new

or

new2

basvanbeek:
I'd remove the shading on the border for a cleaner more modern look. You can then choose to stick with the previously shown darker background or the lighter one as shown here. The border box color should not have too much contrast so depending on background you'd offset the color just a bit so it's visible but not distracting.

tacigar:
@basvanbeek 
How about this?
Should I remove shadow of search condition badges, too?

new3

tacigar:
Current design

light-buttons

Thank you @basvanbeek @adriancole for a lot of valuable feedback!!

codefromthecrypt:
I'm happy with this. I think there will be some further changes in style maybe about the colors used in top bar, but meanwhile we have a lot of code held hostage here. I think it is progress and we can let more feedback occur separately. That said up to @basvanbeek if he's good merge!

basvanbeek:
I will test locally and if I see no blockers I will merge in a few hours when back in my office.

basvanbeek:
@tacigar 
How much work is it to complete the color scheme change so detail panels are not in the old dark theme? Also I notice a huge amount of alignment issues.
Personally I'd get rid of all shadow box effects, they make the design cluttered and look more outdated.
Also some inconsistency issues with fonts. Most are Roboto but for instance the service pills are in a serif font.
UX speaking I'm confused about the + and - icons for tree collapse/expand. I expect clicking + to expand and - to collapse. Currently it is the other way around.
I've attached a screenshot tackling some of the issues I saw.
schermafbeelding 2019-01-10 om 13 27 42

codefromthecrypt:
the thing that is most visually noticeable to me about @basvanbeek suggestions (not to deflate the others) is removal of the shadow boxes. I have to say I really do prefer them removed clicking back and forth looking at both ways.

codefromthecrypt:
hah ps the expand collapse being reversed... I never noticed, but now I do :D

tacigar:
@basvanbeek 
Thank you for your suggestions!!
Looking at the screenshots, I think your design is better than the current design completely.
However, the active highlight of the sidebar may be better on the right side...

drolando:
@tacigar when clicking on a trace I get a blank screen and I see this error in the console:

```
Warning: Failed prop type: The prop `skewCorrectedTrace` is marked as required in `TraceSummary`, but its value is `undefined`.
    in TraceSummary (created by Browser)
    in Browser (created by Connect(Browser))
    in Connect(Browser) (created by Route)
    in Route (created by App)
    in div (created by Layout)
    in div (created by Layout)
    in Layout (created by Route)
    in Route (created by withRouter(Layout))
    in withRouter(Layout) (created by App)
    in Router (created by BrowserRouter)
    in BrowserRouter (created by App)
    in Provider (created by App)
    in App checkPropTypes.js:19
TypeError: root is undefined[Learn More]
bundle.js line 12003 > eval:298:7
The above error occurred in the <TraceSummary> component:
    in TraceSummary (created by Browser)
    in div (created by Browser)
    in div (created by Browser)
    in div (created by Browser)
    in Browser (created by Connect(Browser))
    in Connect(Browser) (created by Route)
    in Route (created by App)
    in div (created by Layout)
    in div (created by Layout)
    in Layout (created by Route)
    in Route (created by withRouter(Layout))
    in withRouter(Layout) (created by App)
    in Router (created by BrowserRouter)
    in BrowserRouter (created by App)
    in Provider (created by App)
    in App

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://fb.me/react-error-boundaries to learn more about error boundaries.
```

If I open the trace in the trace view page it works fine though

drolando:
There's also some weird issue when I refresh the search page (at least on firefox). The dropdown for all selected filters appear open and even clicking around I'm not able to close the one for serviceName. https://www.dropbox.com/s/rx5b8w9wwwn06jq/Screenshot%202019-01-10%2019.03.30.png

Also, there's a missing space (or 2) between the service name and the "x 10"

tacigar:
@drolando 
Thank you for your feedback!!
I'll fix them 🙇

tacigar:
Fixed

zipkin-lens/src/components/GlobalSearch/ConditionLimit.js
```diff
@@ -0,0 +1,84 @@
...
+      <div className="condition-limit">
+        {
+          isFocused
+            ? (
+              <input
+                ref={(ref) => { this.inputRef = ref; }}
+                type="number"
+                value={limit}
+                onChange={(event) => {
+                  let newLimit = parseInt(event.target.value, 10);
```
zeagord: Can you move this outside the component. Just for better readability.

zipkin-lens/webpack.dev.config.js
```diff
@@ -35,6 +35,20 @@ module.exports = {
...
         loader: 'url-loader',
       },
+      {
+        test: /\.svg$/,
+        use: [
+          {
+            loader: 'babel-loader',
+          },
+          {
+            loader: 'react-svg-loader',
```
zeagord, line 45: Did you happen to try the `loader: 'svg-inline-loader'` from webpack itself?

zipkin-lens/src/components/GlobalSearch/ConditionLimit.js
tacigar: Thank you for your comment!

codefromthecrypt:
discussed here we think it is better to just merge this.

PR #2320, opened 2018-12-08
 Update a preprocessing logic to be applied to traces

tacigar:
Currently zipkin lens reuses the zipkin ui's preprocessing logic to be applied to traces.
This PR updates the preprocessing logic of zipkin lens with the latest zipkin ui's logic.
However, some changes have been made according to Linter's warning, and some processing has also been changed.

PR #2327, opened 2018-12-15
Fix bugs that lens shows a blank page after something is clicked

### LINKED ISSUE #2326: Lens Blank Page after Plus Button
## Describe the Bug
When you click on the plus button when the span is expanded it goes to a black screen.

## Steps to Reproduce

1. Expand a span with details that has children
1. Click the plus button on the left while the span details are showing

## JS Stack
```
app-01a4167….min.js:41 TypeError: Cannot read property 'length' of undefined
    at app-01a4167….min.js:25
    at Array.map (<anonymous>)
    at t.value (app-01a4167….min.js:25)
    at Ta (app-01a4167….min.js:41)
    at La (app-01a4167….min.js:41)
    at fo (app-01a4167….min.js:41)
    at ho (app-01a4167….min.js:41)
    at Xo (app-01a4167….min.js:41)
    at Wo (app-01a4167….min.js:41)
    at Vo (app-01a4167….min.js:41)
app-01a4167….min.js:25 Uncaught TypeError: Cannot read property 'length' of undefined
    at app-01a4167….min.js:25
    at Array.map (<anonymous>)
    at t.value (app-01a4167….min.js:25)
    at Ta (app-01a4167….min.js:41)
    at La (app-01a4167….min.js:41)
    at fo (app-01a4167….min.js:41)
    at ho (app-01a4167….min.js:41)
    at Xo (app-01a4167….min.js:41)
    at Wo (app-01a4167….min.js:41)
    at Vo (app-01a4167….min.js:41)
```

tacigar:
Fix #2326 
Thank you @beckje01 !

In #2320 , some fields of objects related to trace information have been renamed.
These bugs occurred because I forgot to use this new field name in some components.

### DIFF
--- a/zipkin-lens/src/components/Browser/Traces/index.js
+++ b/zipkin-lens/src/components/Browser/Traces/index.js
@@ -48,8 +48,8 @@ class Traces extends React.Component {
     return traceSummaries
       .filter((summary) => {
         for (let i = 0; i < filters.length; i += 1) {
-          if (!summary.serviceDurations.find(
-            service => service.name === filters[i],
+          if (!summary.serviceSummaries.find(
+            serviceSummary => serviceSummary.serviceName === filters[i],
           )) {
             return false;
           }
--- a/zipkin-lens/src/components/Timeline/index.js
+++ b/zipkin-lens/src/components/Timeline/index.js
@@ -123,8 +123,8 @@ class Timeline extends React.Component {
               /* Skip closed spans */
               if (closed[span.spanId]) {
                 if (hasChildren) {
-                  for (let i = 0; i < span.children.length; i += 1) {
-                    closed[span.children[i]] = true;
+                  for (let i = 0; i < span.childIds.length; i += 1) {
+                    closed[span.childIds[i]] = true;
                   }
                 }
                 return null;


shakuzen:
Do we have tests for Lens that run as part of the build and would catch this kind of thing in the future?

tacigar:
No... not yet.
There are some tests, but these are not enough.
I have to add tests.

But I think what I have to do at first is refactoring codes before adding tests, because I believe that these bugs will never occur again by refactoring and appropriate use prop types.
The refactoring I believe is necessary is below.

1. **Split props properly**

Currently traces related props are large objects such as `traceSummary` like the following.
https://github.com/openzipkin/zipkin/blob/9b91acbaa3b90ad1c9e4ab62489db05dfd9edf74/zipkin-lens/src/components/Browser/Traces/Trace.js#L15
This is too bad... :sob: 
props should be primitive as much as possible.
It helps us find such bugs early.

2. **Define prop types strictly**

Currently the definitions of the prop types are too rough like the following.
https://github.com/openzipkin/zipkin/blob/9b91acbaa3b90ad1c9e4ab62489db05dfd9edf74/zipkin-lens/src/components/Browser/Traces/Trace.js#L15
This is a bad way as prop types tell us a type bug.
So I should define prop types strictly like the following

```
traceSummary: PropTypes.shape({
  serviceSummaries: PropTypes.arrayOf({
    serviceName: PropTypes.string,
    ...
  }),
  ...
}).isRequired,
```

---

But these changes are quite large.
Should I do it now or later ... :thinking: ?

shakuzen:
Thanks for the info, @tacigar! I think those can be done as a separate pull request. Could you make an issue to track it just so we don't forget? I don't want to block resolving this pull request that fixes an issue that a user is facing trying out Lens.

beckje01:
We have lens running in parallel to the original up in our dev to find issues. So anything we submit isn't urgent to us just trying to get early testing in.

llinder (approved):

tacigar:
@shakuzen 
OK! I'll make issues

@beckje01 
I see! Thank you very much 👍

PR #2328, opened 2018-12-16
Adds gRPC endpoint /zipkin.proto3.SpanService/Report

ewhauser:
This adds an opt-in gRPC endpoint for reporting spans. A request to the 
gRPC `/zipkin.proto3.SpanService/Report` service is the same proto used
with our `POST /api/v2/spans`, `Content-Type: application/x-protobuf` 
endpoint: `zipkin.proto3.ListOfSpans`.

This is currently disabled by default, as the feature is experimental.
Enable it like so:
```bash
$ COLLECTOR_GRPC_ENABLED=true java -jar zipkin.jar
```

Under the scenes the runtime is the same as the POST endpoint (Armeria), 
and uses the same port (9411).

---
former description:

This PR is an offshoot of a discussion in openzipkin/zipkin-api#57 . There are a number of issues that need to be addressed before consideration of merge:

- [x] Finalize the API and merge it into `zipkin-api`
- [ ] Find a zero-duplication and zero-copy approach for the deserialization of spans on the server side
- [x] Add additional configuration options for configuring the server (i.e. SSL)
- [ ] Integration test with GCP
- [ ] Performance testing

Also, there is still some discussion about whether the project as a whole wants to support a gRPC endpoint in `zipkin-server` due to potential bloat. See the discussion in zipkin-api#57 for background.

### DIFF
--- a/zipkin-server/README.md
+++ b/zipkin-server/README.md
@@ -370,6 +370,19 @@ Example usage:
 $ RABBIT_ADDRESSES=localhost java -jar zipkin.jar
 ```
 
+### gRPC Collector (Experimental)
+You can enable a gRPC span collector endpoint by setting `COLLECTOR_GRPC_ENABLED=true`. The
+`zipkin.proto3.SpanService/Report` endpoint will run on the same port as normal http (9411).
+
+
+Example usage:
+
+```bash
+COLLECTOR_GRPC_ENABLED=true java -jar zipkin.jar
+```
+
+As this service is experimental, it is not recommended to run this in production environments.
+
 ### 128-bit trace IDs
 
 Zipkin supports 64 and 128-bit trace identifiers, typically serialized
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinGrpcCollector.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinGrpcCollector.java
@@ -1,0 +1,78 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package zipkin2.server.internal;
+
+import com.linecorp.armeria.common.grpc.protocol.AbstractUnaryGrpcService;
+import com.linecorp.armeria.spring.ArmeriaServerConfigurator;
+import java.util.concurrent.CompletableFuture;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
+import zipkin2.Callback;
+import zipkin2.codec.SpanBytesDecoder;
+import zipkin2.collector.Collector;
+import zipkin2.collector.CollectorMetrics;
+import zipkin2.collector.CollectorSampler;
+import zipkin2.storage.StorageComponent;
+
+/** Collector for receiving spans on a gRPC endpoint. */
+@ConditionalOnProperty(name = "zipkin.collector.grpc.enabled") // disabled by default
+final class ZipkinGrpcCollector {
+  static final byte[] EMPTY = new byte[0];
+
+  @Bean ArmeriaServerConfigurator grpcCollectorConfigurator(StorageComponent storage,
+    CollectorSampler sampler, CollectorMetrics metrics) {
+    CollectorMetrics grpcMetrics = metrics.forTransport("grpc");
+    Collector collector = Collector.newBuilder(getClass())
+      .storage(storage)
+      .sampler(sampler)
+      .metrics(grpcMetrics)
+      .build();
+
+    return sb ->
+      sb.service("/zipkin.proto3.SpanService/Report", new SpanService(collector, grpcMetrics));
+  }
+
+  static final class SpanService extends AbstractUnaryGrpcService {
+
+    final Collector collector;
+    final CollectorMetrics metrics;
+
+    SpanService(Collector collector, CollectorMetrics metrics) {
+      this.collector = collector;
+      this.metrics = metrics;
+    }
+
+    @Override protected CompletableFuture<byte[]> handleMessage(byte[] bytes) {
+      metrics.incrementMessages();
+      CompletableFutureCallback result = new CompletableFutureCallback();
+      collector.acceptSpans(bytes, SpanBytesDecoder.PROTO3, result);
+      return result;
+    }
+  }
+
+  static final class CompletableFutureCallback extends CompletableFuture<byte[]>
+    implements Callback<Void> {
+
+    @Override public void onSuccess(Void value) {
+      complete(EMPTY);
+    }
+
+    @Override public void onError(Throwable t) {
+      completeExceptionally(t);
+    }
+  }
+}
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
@@ -98,7 +98,10 @@ public class ZipkinServerConfiguration implements WebMvcConfigurer {
       // could split the collector's CORS policy into a different property, still allowing POST
       // with content-type by default.
       .allowRequestMethods(HttpMethod.GET, HttpMethod.POST)
-      .allowRequestHeaders(HttpHeaderNames.CONTENT_TYPE);
+      .allowRequestHeaders(HttpHeaderNames.CONTENT_TYPE,
+        // Use literals to avoid a runtime dependency on armeria-grpc types
+        HttpHeaderNames.of("X-GRPC-WEB"))
+      .exposeHeaders("grpc-status", "grpc-message", "armeria.grpc.ThrowableProto-bin");
     return builder -> builder.decorator(corsBuilder::build);
   }
 
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -80,8 +80,21 @@
       <version>${wire.version}</version>
     </dependency>
     <dependency>
-      <groupId>org.apache.zipkin.proto3</groupId>
+      <groupId>io.zipkin.proto3</groupId>
       <artifactId>zipkin-proto3</artifactId>
+    </dependency>
+    <!-- TODO: This only resolves google empty. Remove this once using org.apache.zipkin.proto3 -->
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+      <version>3.7.1</version>
+      <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>*</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
   </dependencies>
 
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -80,7 +80,7 @@
       <version>${wire.version}</version>
     </dependency>
     <dependency>
-      <groupId>io.zipkin.proto3</groupId>
+      <groupId>org.apache.zipkin.proto3</groupId>
       <artifactId>zipkin-proto3</artifactId>
     </dependency>
   </dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -154,6 +154,26 @@
     <url>https://github.com/openzipkin/zipkin/issues</url>
   </issueManagement>
 
+  <repositories>
+    <repository>
+      <id>apache.snapshots.https</id>
+      <url>https://repository.apache.org/content/repositories/snapshots/</url>
+      <releases>
+        <enabled>false</enabled>
+      </releases>
+    </repository>
+  </repositories>
+
+  <pluginRepositories>
+    <pluginRepository>
+      <id>apache.snapshots.https</id>
+      <url>https://repository.apache.org/content/repositories/snapshots/</url>
+      <releases>
+        <enabled>false</enabled>
+      </releases>
+    </pluginRepository>
+  </pluginRepositories>
+
   <dependencyManagement>
     <dependencies>
       <dependency>
@@ -362,9 +382,9 @@
       </dependency>
 
       <dependency>
-        <groupId>io.zipkin.proto3</groupId>
+        <groupId>org.apache.zipkin.proto3</groupId>
         <artifactId>zipkin-proto3</artifactId>
-        <version>0.1.0</version>
+        <version>0.2.1-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>com.squareup.wire</groupId>
@@ -627,6 +647,7 @@
           <mapping>
             <!-- Don't use javadoc style as this makes code formatters break it by adding tags! -->
             <java>SLASHSTAR_STYLE</java>
+            <kt>SLASHSTAR_STYLE</kt>
           </mapping>
           <excludes>
             <exclude>.travis.yml</exclude>
--- a/pom.xml
+++ b/pom.xml
@@ -154,26 +154,6 @@
     <url>https://github.com/openzipkin/zipkin/issues</url>
   </issueManagement>
 
-  <repositories>
-    <repository>
-      <id>apache.snapshots.https</id>
-      <url>https://repository.apache.org/content/repositories/snapshots/</url>
-      <releases>
-        <enabled>false</enabled>
-      </releases>
-    </repository>
-  </repositories>
-
-  <pluginRepositories>
-    <pluginRepository>
-      <id>apache.snapshots.https</id>
-      <url>https://repository.apache.org/content/repositories/snapshots/</url>
-      <releases>
-        <enabled>false</enabled>
-      </releases>
-    </pluginRepository>
-  </pluginRepositories>
-
   <dependencyManagement>
     <dependencies>
       <dependency>
@@ -382,9 +362,9 @@
       </dependency>
 
       <dependency>
-        <groupId>org.apache.zipkin.proto3</groupId>
+        <groupId>io.zipkin.proto3</groupId>
         <artifactId>zipkin-proto3</artifactId>
-        <version>0.2.1-SNAPSHOT</version>
+        <version>0.2.0</version>
       </dependency>
       <dependency>
         <groupId>com.squareup.wire</groupId>
@@ -546,7 +526,8 @@
                 <goal>unpack-dependencies</goal>
               </goals>
               <configuration>
-                <includeArtifactIds>zipkin-proto3</includeArtifactIds>
+                <!-- TODO: Remove protobuf-java once using org.apache.zipkin.proto3 -->
+                <includeArtifactIds>protobuf-java,zipkin-proto3</includeArtifactIds>
                 <includes>**/*.proto</includes>
                 <outputDirectory>${unpack-proto.directory}</outputDirectory>
               </configuration>
@@ -566,6 +547,8 @@
               <configuration>
                 <protoSourceDirectory>${unpack-proto.directory}</protoSourceDirectory>
                 <includes>
+                  <!-- TODO: Remove google.protobuf.Empty once using org.apache.zipkin.proto3 -->
+                  <include>google.protobuf.Empty</include>
                   <include>zipkin.proto3.*</include>
                 </includes>
               </configuration>
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -34,7 +34,9 @@
     <main.java.version>1.8</main.java.version>
     <main.signature.artifact>java18</main.signature.artifact>
     <start-class>zipkin.server.ZipkinServer</start-class>
+    <kotlin.version>1.3.30</kotlin.version>
     <maven-invoker-plugin.version>3.2.0</maven-invoker-plugin.version>
+    <proto.generatedSourceDirectory>${project.build.directory}/generated-test-sources/wire</proto.generatedSourceDirectory>
   </properties>
 
   <dependencyManagement>
@@ -82,6 +84,10 @@
       <groupId>com.linecorp.armeria</groupId>
       <artifactId>armeria-zipkin</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.linecorp.armeria</groupId>
+      <artifactId>armeria-grpc-protocol</artifactId>
+    </dependency>
 
     <!-- zipkin requires exporting /health endpoint -->
     <dependency>
@@ -123,6 +129,25 @@
       <groupId>org.jsoup</groupId>
       <artifactId>jsoup</artifactId>
       <version>1.11.3</version>
+    </dependency>
+
+    <!-- to test the experimental grpc endpoint with the square/wire library -->
+    <dependency>
+      <groupId>org.apache.zipkin.proto3</groupId>
+      <artifactId>zipkin-proto3</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.squareup.wire</groupId>
+      <artifactId>wire-grpc-client</artifactId>
+      <version>${wire.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jetbrains.kotlin</groupId>
+      <artifactId>kotlin-stdlib</artifactId>
+      <version>${kotlin.version}</version>
+      <scope>test</scope>
     </dependency>
 
     <!-- Cassandra and Cassandra 3 backends -->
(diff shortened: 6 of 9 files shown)


zipkin-server-grpc/src/main/java/zipkin2/server/grpc/ZipkinGrpcServer.java
```diff
@@ -0,0 +1,163 @@
...
+import zipkin2.proto3.PublishSpansResponse;
+import zipkin2.proto3.Span;
+import zipkin2.proto3.SpanServiceGrpc;
+import zipkin2.storage.StorageComponent;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ZipkinGrpcServer extends SpanServiceGrpc.SpanServiceImplBase {
```
codefromthecrypt: I would make this `GrpcCollector extends CollectorComponent`, then make a static inner class and/or package private that extends `SpanServiceGrpc.SpanServiceImplBase`. Then the start on collector would call this, and close, could call stop. you can look at the kafka one for an example. Notably, this allows you to add a health check so that the normal health status can return (though in this case it probably isn't of much consequence unless it fails to start for some reason).

codefromthecrypt:
curios... maybe @trustin or @anuraaga would know.. is there a way to run both normal http and grpc's http on the same port? Currently, we use undertow underneath, but we could consider changing that.

anuraaga:
If you use armeria :) Upstream gRPC doesn't support it yet I think, or even if it did it would only be HTTP/2, not HTTP/1.

codefromthecrypt:
>
> If you use armeria :) Upstream gRPC doesn't support it yet I think, or
> even if it did it would only be HTTP/2, not HTTP/1.
>

frankly I think we should switch to armeria then. we have several
contributors who are also contributors to armeria. zipkin always tries to
be simplest possible setup and adding ports betrays this. if we are talking
about grpc being shaded anyway we can just make stackdriver use the shaded
one (or potentially convert its storage module also to armeria)

>

anuraaga:
I guess the talking points would be that armeria still isn't GA and has more dependencies. But agree that more ports is annoying and a big reason I love armeria :)

codefromthecrypt:
we dont promise any api compat for our server, so not worries. if we did
armeria we would just add grpc endpoint directly to the zipkin-server imho
and not mess with extra modules.

codefromthecrypt:
also good point on the deps of armeria. similar audit would apply to ensure
things still work ;)

FWIW current Cassandra driver supports a range of guava starting at 16
iirc. later it doesnt depend on guava. almost everything else has no
conflicts due to okhttp directly. we would need to watch that zipkin-aws
and zipkin-gcp extensions dont conflict.

ewhauser:
I can look into armeria as an alternative.

The biggest trick here is going to be in the serializer though. We have to use Zipkin’s Proto3Codec instead of relying on the server implementation to deserialize the span payload. I’m not sue if any of the server implementations support the concept of lazy deserialization. If they don’t, then I’ll have to plug into the serialization pipeline.

anuraaga:
For the zero copy aspect, it might be simplest to define the client API in zipkin-api as is but use an internal proto here that replaces Span with bytes. Embedded messages and strings are encoded exactly the same in proto.

https://developers.google.com/protocol-buffers/docs/encoding

anuraaga:
Also I guess you would want to use the new support for plugging in as a spring webflux server - I don't know what that means since I haven't used spring recently but probably relevant.

https://github.com/line/armeria/commit/70785f8438e25385c5a30947ffc26a1ad3ab148a

And if you use this unsafe option you'd be able to have real zero-copy with no heap arrays involved until the zipkin2.Span

https://github.com/line/armeria/blob/master/grpc/src/main/java/com/linecorp/armeria/server/grpc/GrpcServiceBuilder.java#L181

ewhauser:
There are a few more dependencies with `armeria`:

```console
[INFO] +- com.linecorp.armeria:armeria-grpc-shaded:jar:0.75.0:compile
[INFO] |  +- com.linecorp.armeria:armeria-shaded:jar:0.75.0:compile
[INFO] |  |  +- com.fasterxml.jackson.core:jackson-annotations:jar:2.9.7:compile
[INFO] |  |  +- com.fasterxml.jackson.core:jackson-core:jar:2.9.7:compile
[INFO] |  |  +- com.fasterxml.jackson.core:jackson-databind:jar:2.9.7:compile
[INFO] |  |  +- io.micrometer:micrometer-core:jar:1.1.0:compile
[INFO] |  |  |  +- org.hdrhistogram:HdrHistogram:jar:2.1.9:compile
[INFO] |  |  |  \- org.latencyutils:LatencyUtils:jar:2.0.3:compile
[INFO] |  |  +- io.netty:netty-codec-haproxy:jar:4.1.31.Final:compile
[INFO] |  |  |  \- io.netty:netty-codec:jar:4.1.31.Final:compile
[INFO] |  |  +- io.netty:netty-codec-http2:jar:4.1.31.Final:compile
[INFO] |  |  |  +- io.netty:netty-codec-http:jar:4.1.31.Final:compile
[INFO] |  |  |  \- io.netty:netty-handler:jar:4.1.31.Final:compile
[INFO] |  |  +- io.netty:netty-resolver-dns:jar:4.1.31.Final:compile
[INFO] |  |  |  +- io.netty:netty-resolver:jar:4.1.31.Final:compile
[INFO] |  |  |  \- io.netty:netty-codec-dns:jar:4.1.31.Final:compile
[INFO] |  |  +- io.netty:netty-tcnative-boringssl-static:jar:2.0.19.Final:compile
[INFO] |  |  +- io.netty:netty-transport:jar:4.1.31.Final:compile
[INFO] |  |  |  \- io.netty:netty-buffer:jar:4.1.31.Final:compile
[INFO] |  |  +- io.netty:netty-transport-native-epoll:jar:linux-x86_64:4.1.31.Final:compile
[INFO] |  |  |  +- io.netty:netty-common:jar:4.1.31.Final:compile
[INFO] |  |  |  \- io.netty:netty-transport-native-unix-common:jar:4.1.31.Final:compile
[INFO] |  |  +- io.netty:netty-transport-native-unix-common:jar:linux-x86_64:4.1.31.Final:compile
[INFO] |  |  \- org.reactivestreams:reactive-streams:jar:1.0.2:compile
[INFO] |  +- com.google.code.findbugs:jsr305:jar:3.0.2:compile
[INFO] |  +- org.curioswitch.curiostack:protobuf-jackson:jar:0.3.0:compile
[INFO] |  |  +- javax.annotation:javax.annotation-api:jar:1.3.1:runtime
[INFO] |  |  +- com.google.protobuf:protobuf-java-util:jar:3.6.1:runtime
```

Neither `netty` or `jackson` being shaded is less than ideal. Appears that `armeria` is choosing not to shade `netty` in the future - https://github.com/line/armeria/issues/705.

On the plus side, I believe you can run HTTP/1, gRPC, and gRPC-Web all from the same server instance. GRPC-Web only supports unary RPC at present, so would need to provide a unary method for sending spans as well (or alternatively drop streaming support).

ewhauser:
@anuraaga - For implementing the `Span` payload directly as a `ByteString`, I was looking into implementing `io.grpc.BindableService` directly and replacing the input parameter. I haven't tried that yet though so if you have a pointer or example then I'd be happy to take it.

codefromthecrypt:
to see differences more directly probably add it as a dep on zipkin-server
then build the exec jar. do a jar -tvf|sort before and after and diff
those. I suspect Jackson is already in the main exec jar even if we don't
use it (I think zipkin-aws does)

anuraaga:
I was thinking this would be simplest by having

Public API proto, used by clients

```
message WriteSpansRequest {
  repeated zipkin.proto3.Span span = 1;
}

service ZipkinService {
  rpc WriteSpans (stream WriteSpansRequest) returns (whatever);
}
```

while the server actually implements a different, non-exposed proto 

```
message WriteSpansRequest {
  repeated bytes span = 1;
}

service ZipkinService {
  rpc WriteSpans (stream WriteSpansRequest) returns (whatever);
}
```

The protos are compatible so the clients get the advantage of the generated code and easy-to-populate structs while the server would only see the `ByteString` and decode straight into a Java span.

It has the downside of duplicate protos but it's probably a lot less complex than a custom `BindableService`

ewhauser:
Got it. Yes, that is a definitely easier and a nice trick. Will give it a shot. Thanks!

ewhauser:
So, @anuraaga's trick doesn't work - at least with gRPC Java. The issue is that the marshaller will have already copied the input stream into a `ByteString` prior to getting access to it.

So I took the approach of implementing a custom `BindableService`. This has the added benefit of removing all dependencies on Protocol Buffers entirely. Here are the dependencies introduced now:

```console
[INFO] +- io.grpc:grpc-netty-shaded:jar:1.17.1:runtime
[INFO] |  \- io.grpc:grpc-core:jar:1.17.1:compile (version selected from constraint [1.17.1,1.17.1])
[INFO] |     +- io.grpc:grpc-context:jar:1.17.1:compile
[INFO] |     +- com.google.code.gson:gson:jar:2.7:compile
[INFO] |     +- com.google.errorprone:error_prone_annotations:jar:2.2.0:compile
[INFO] |     +- com.google.code.findbugs:jsr305:jar:3.0.2:compile
[INFO] |     +- org.codehaus.mojo:animal-sniffer-annotations:jar:1.17:compile
[INFO] |     +- com.google.guava:guava:jar:26.0-android:compile
[INFO] |     |  +- org.checkerframework:checker-compat-qual:jar:2.5.2:compile
[INFO] |     |  \- com.google.j2objc:j2objc-annotations:jar:1.1:compile
[INFO] |     +- io.opencensus:opencensus-api:jar:0.17.0:compile
[INFO] |     \- io.opencensus:opencensus-contrib-grpc-metrics:jar:0.17.0:compile
[INFO] +- io.grpc:grpc-stub:jar:1.17.1:compile
```

So this ends up being much less of a dependency burden than `armeria` - no netty, protobufs - with the downside of having the collector run on a different port. We can keep this version of gRPC in sync with the version that is being used by `zipkin-gcp` to avoid any issues.

However, `armeria` does have the advantage of everything being on the same port and it would support gRPC-Web out of the box without a proxy. Would switching to `armeria` mean that it would eliminate `undertow` and just use it for everything? If so, we shoud probably find a path to finalizing the gRPC API in this PR and moving forward with this experimental version on a different port. A separate issue could be opened for moving to `armeria` - given the size of that change, I would want to make sure there was plenty of support for it.

codefromthecrypt:
yes armeria would imply geting rid of undertow and its deps. undertow was
only added for better perf.

plan sounds fine, but I'm really not keen merging another port to expose.
meaning if helpful for development can separate, but please don't release
us needing to support a second http port perpetually unless the change to
undo that immediately follows.

codefromthecrypt:
PS it isn't fair to put the burden of switching all to armeria on Eric :P I
will spike a PR so in case this happens while I'm away (I'm out tomorrow)
the grunt work is ready.

ewhauser:
I started this whole mess so I'm fine rolling up my sleeves on the switch. Just want to make sure there is support behind it.

codefromthecrypt:
> I started this whole mess so I'm fine rolling up my sleeves on the switch. Just want to make sure there is support behind it.

I think there's some sort of "jury's still out" in general.

The trick is that adding an optional module will likely become
effectively required as there is so much marketing buzz around gRPC.
So, we have to be quite careful how to introduce this. scribe being on
a different port caused a bunch of pain back in the day with people
invariably choosing the wrong port, having to go through work to
figure it out (matters not if docs say use the right port), then going
onto support and moving the concern there, or triple posted issues
etc.

Basically, if we're going to do this, in the case of gRPC, I think we
ought to consider this becoming effectively required, and how to stop
the most type of problems we can't control. We do have a lot of
control over the server and extensions of it as we explicitly give
zero warranty on custom builds. We have little control over config
mistakes even with docs.

codefromthecrypt:
ps I'll graciously accept your offer to do the armeria spike if comes
to that. I'll raise an issue on it.

codefromthecrypt:
here's the issue for armeria https://github.com/openzipkin/zipkin/issues/2331

trustin:
@ewhauser wrote:
> I would want to make sure there was plenty of support for it.

The Armeria project maintainers including LINE employees like me (@hyangtack and @minwoox as well) would be happy to support Zipkin community for adoption of Armeria and its sustainable development. I believe you don't need to worry too much about the future of Armeria given that it's being used in various critical parts of LINE services.

anuraaga:
For the copying issue, just to clarify, using the proto approach makes

`ByteBuf` -> `zipkin.proto3.Span` -> `byte[]` -> `zipkin2.Span` become

`ByteBuf` -> `byte[]` -> `zipkin2.Span`

and then using Armeria's unsafe option can make it become

`ByteBuf` -> `zipkin2.Span` 

`Proto3Codec` needs to be updated to accept `ByteBuffer`, which is pretty simple change I think. On the flip side, I think it's not practical for it to parse the `InputStream` directly since doing that efficiently would probably require using the `KnownLength` class inside of `Proto3Codec` which shouldn't have gRPC dependencies.

Of course it depends on the decision of moving to armeria, but if we do, I'd still recommend using the proto trick + unsafe option.

ewhauser:
Agreed. If armeria is the way forward, then this is the better option.

ewhauser:
Trustin - armeria seems like a great project. More so wanted to make sure
there was support for using it within zipkin-server. There are a lot of
consumers that I don’t want to break; thank you for the offer to help
support with the adoption!

codefromthecrypt:
I was just looking at the proto. It seems probably unary ListOfSpans style request would make more sense as the initial impl as all infrastructure we have is based on ListOfSpans (Collector, Storage etc). It is nice that all of our impls accept literally the same messages. For example, if our initial method was putSpans(ListOfSpans) then the exact codec we use for normal proto over http/2 could be used to process that message, right?

Starting with just a unary likely also means the largest amount of client support in general. Does that sound right?

ewhauser:
Yes. I’m was planning on swapping out in the next rev as streaming is not
yet supported by GRPC-Web.

Re: putSpans and ListOfSpans - I think that’s the right approach. armeria
says it supports gRPC over HTTP/1 so I need to explore how that works. If
it’s similar to GRPC Gateway where you use annotations to define the HTTP/1
mappings, then you could potentially share the code for both
implementations.

ewhauser:
For inspiration, I took a look at the [API](https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/agent/trace/v1/trace_service.proto#L45) provided by OpenCensus trace collector.

They actually use bidirectional streaming in their [implementation](https://github.com/census-instrumentation/opencensus-service/blob/f35ec164877d1f26b695d6dcece9130849dd8633/receiver/opencensus/octrace/opencensus.go#L70) although there are some nuances in their protocol which do not exist for Zipkin.

I still think implementing unary makes sense since gRPC-Web does not support streaming, but I'll probably perf test both solutions before finishing.

anuraaga:
I think streaming is interesting if the client doesn't do any batching so if you try comparing the perf, I'd recommend doing it without batching in the streaming version at least as an option

codefromthecrypt:
@anuraaga I don't expect many clients to not do bundling at all as that implies a different internal reporting arch only used for grpc, wouldn't it?

For example, even kafka is a bundle-at-a-time transport. This was done for reasons including that some spans are so small that sending in a single message is wasteful. Also, often people mistakenly assume that sending one span onto a transport is protecting the app from overhead when usually even kafka clients can (and have) crashed an app enqueueing.

I wonder if we can decouple this completely as having two supported ways to do grpc seems tech debt creating not just here but the inevitable "it is streaming on server, so why not also in X (where X is every client library)". I really hesitate strongly about adding more complexity especially in the initial impl as there's already quite a lot of work to do just for unary.

Assuming unary is in, it would seem possible for someone who felt strongly about streaming to add a module to their setup that is a streaming alternate as well.. meaning we can completely defer this whole thing as well the impact discussions which now only stall the work entirely. What I mean is bundling the option for streaming means supporting it and has almost the same impact as supporting it enabled by default. I don't quite feel we should add that burden today considering we still have a lot of other things to do to get this in.

anuraaga:
Yeah just going unary seems fine. I think streaming without batching is interesting since HTTP/2 flow control is basically a type of batching - having two layers of batching feels like it should be unnecessary, so it would be interesting to see if it makes things simpler to just use the flow control and stream a Span at a time.

codefromthecrypt:
@anuraaga PS I like the idea of re-using http/2 flow-control (and also ping for that matter) in theory, and will also help with whatever infra is needed when we experiment with streaming.

codefromthecrypt:
fyi I didn't knock out all the kinks here but here's a work in progress for direct armeria integration https://github.com/openzipkin/zipkin/pull/2348

codefromthecrypt:
this can be now be rebased in master, which is armeria, avoiding a second http engine

codefromthecrypt:
@ewhauser you mind if I mutate your branch?

ewhauser:
Of course not. Little backed up at the moment; but feel free to ping me if
I can help

codefromthecrypt:
ok I rebased this over master and did some refactoring in the last commit. 

cc @openzipkin/armeria 

A cheap test to show that the service mounted
```bash
$ curl -X POST localhost:9411/zipkin.proto3.SpanService/Report -H'Content-Type: application/grpc' -v
*   Trying ::1...
* TCP_NODELAY set
> POST /zipkin.proto3.SpanService/Report HTTP/1.1
> Host: localhost:9411
> User-Agent: curl/7.54.0
> Accept: */*
> Content-Type: application/grpc
> 
< HTTP/1.1 200 OK
< content-type: application/grpc+proto
< grpc-status: 13
< grpc-message: Half-closed without a request
< content-length: 0
< 
```

zipkin-server/src/test/proto/zipkin-service.proto
```diff
@@ -0,0 +1,39 @@
...
+
+// In Java, the closest model type to this proto is in the "zipkin2" package
+option java_package = "zipkin2.proto3";
+option java_multiple_files = true;
+
+// TODO Once this API is finalized it should be moved to zipkin-api's zipkin.proto prior to
+// this PR being merged into zipkin. We don't actually use the generated stubs anywhere in the main project -
+// they are only being used for integration tests.
+
+// Use this to report spans onto the gRPC transport. Implementations are asynchronous and may drop
```
codefromthecrypt: @openzipkin/core here's my take on the grpc endpoint definition. you can ignore over this line because it would weave into the normal proto.

codefromthecrypt:
here's the change to the proto https://github.com/openzipkin/zipkin-api/pull/59

codefromthecrypt:
the jar grew a lot again. will track things I can't control here in https://github.com/line/armeria/issues/1703

cc @trustin @anuraaga 

```bash
$ du -k zipkin-nogrpc.jar
47396	zipkin-nogrpc.jar
$ du -k  zipkin-grpc.jar
55840	zipkin-grpc.jar
```

diff zipkin-nogrpc.jars zipkin-grpc.jars
```diff
1a2
> 3277212 BOOT-INF/lib/byte-buddy-1.9.12.jar
9a11
> 1424558 BOOT-INF/lib/protobuf-java-3.7.0.jar
13a16
> 1200130 BOOT-INF/lib/proto-google-common-protos-1.12.0.jar
15a19
> 825158 BOOT-INF/lib/grpc-services-1.19.0.jar
16a21
> 766263 BOOT-INF/lib/grpc-core-1.19.0.jar
35a41
> 327600 BOOT-INF/lib/opencensus-api-0.19.2.jar
41a48
> 241622 BOOT-INF/lib/gson-2.8.5.jar
43c50
< 226228 BOOT-INF/lib/zipkin-2.12.9-SNAPSHOT.jar
---
> 226229 BOOT-INF/lib/zipkin-2.12.9-SNAPSHOT.jar
46a54
> 142031 BOOT-INF/lib/armeria-grpc-0.83.0.jar
51a60
> 126747 BOOT-INF/lib/re2j-1.2.jar
53a63
> 113966 BOOT-INF/lib/protobuf-jackson-0.3.1.jar
61a72
> 68840 BOOT-INF/lib/protobuf-java-util-3.6.1.jar
70a82
> 39549 BOOT-INF/lib/grpc-stub-1.19.0.jar
77a90
> 29016 BOOT-INF/lib/grpc-context-1.19.0.jar
89a103
> 14961 BOOT-INF/lib/opencensus-contrib-grpc-metrics-0.19.2.jar
91a106
> 13694 BOOT-INF/lib/error_prone_annotations-2.2.0.jar
98a114
> 7614 BOOT-INF/lib/grpc-protobuf-lite-1.19.0.jar
99a116,117
> 6834 BOOT-INF/lib/zipkin-proto3-0.2.0.jar
> 6684 BOOT-INF/lib/zipkin-collector-grpc-2.12.9-SNAPSHOT.jar
105a124
> 4994 BOOT-INF/lib/grpc-protobuf-1.19.0.jar
108a128
> 3448 BOOT-INF/lib/animal-sniffer-annotations-1.17.jar
109a130
> 2363 BOOT-INF/lib/zipkin-autoconfigure-collector-grpc-2.12.9-SNAPSHOT.jar
```

codefromthecrypt:
note that almost all of the big jars should be unnecessary as our service endpoint is well defined and can parse its own message without the help of google's proto libraries

ewhauser:
The way I was trying to do this previously was make the gRPC endpoint optional - similar to how the storage integrations work. Obviously, a tradeoff there.

codefromthecrypt:
@ewhauser ack we could do that. I should have mentioned the goal I'm trying for here is to not pollute our maven namespace at all, and embed the grpc listener in the server impl. That way, layered integrations such as zipkin-aws and zipkin-gcp magically can use it without module composition (which would likely lead to more dep clashes)

codefromthecrypt:
basically I think almost all of the extra deps are unnecessary which could mean <1MiB overhead in terms of jar size for a pretty sweet feature even if we start by turning it off by default.

We have had some grumbles about how our recent jar hike (prior to this) is slowing people with wan connections, so hoping we can just obviate that.

codefromthecrypt:
rebasing this now. then will try with https://github.com/line/armeria/pull/1727

codefromthecrypt:
FYI: latest changes in master before this have shrunk the zipkin jar a little bit.

```
$ du -k zipkin.jar 
47104	zipkin.jar
```

codefromthecrypt:
so before the grpc'less grpc here's the before and after

before: 47104k
after:    56384k

codefromthecrypt:
@anuraaga while the last commit doesn't quite work, yet. the jar is 3MiB down so far..

before: 47104k
before grpc-less: 56384k
now: 53312k

codefromthecrypt:
While there is a little work to do, especially around avoiding tripping up classloader based on docservice, once this works, the overhead to add grpc is roughly 1MiB and no more guava fight (at the moment anyway)

cc @trustin @anuraaga 

```
before: 47104k
before grpc-less: 56384k
now: 48192k (back to below 50Mib!)
```

let's pick this back up when https://github.com/line/armeria/pull/1727 changes again

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinGrpcCollector.java
```diff
@@ -0,0 +1,73 @@
...
+import zipkin2.Callback;
+import zipkin2.codec.SpanBytesDecoder;
+import zipkin2.collector.Collector;
+import zipkin2.collector.CollectorMetrics;
+import zipkin2.collector.CollectorSampler;
+import zipkin2.storage.StorageComponent;
+
+/** Collector for receiving spans on a gRPC endpoint. */
+@ConditionalOnProperty(name = "zipkin.collector.grpc.enabled") // disabled by default
+final class ZipkinGrpcCollector {
```
codefromthecrypt, line 33: @ewhauser can't get much smaller and still be java, right?

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinGrpcCollector.java
```diff
@@ -0,0 +1,73 @@
...
+      collector.acceptSpans(bytes, SpanBytesDecoder.PROTO3, result);
+      return result;
+    }
+  }
+
+  static final class CompletableFutureCallback extends CompletableFuture<byte[]>
+    implements Callback<Void> {
+
+    @Override public void onSuccess(Void value) {
+      complete(new byte[0] /* empty result */);
```
anuraaga: May as well use `static final byte[] EMPTY = new byte[0]`

codefromthecrypt: good idea

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinGrpcCollector.java
ewhauser: 👍

codefromthecrypt:
@ewhauser I think this is ready to merge. We aren't releasing right away as the project is in the middle of an ASF move. However, the remaining task of perf testing will be easier once merged as people can build off master. The feature is currently disabled under a flag.

I don't expect perf testing to be surprising as this is not much different than vanilla armeria http, but yeah once we have a grpc sender we can hook it up and compare.

codefromthecrypt:
The zero copy thing is in a different change, btw, it isn't grpc specific #2435

shakuzen:
> However, the remaining task of perf testing will be easier once merged as people can build off master.

As long as this PR is up-to-date with master, maybe this is a good opportunity to try out Jitpack for testing pre-merge?

codefromthecrypt:
@shakuzen sure.. anyone want to translate this into eric's branch name? "https://jitpack.io/com/github/apache/incubator-zipkin/zipkin-server/2.12.9/zipkin-server-2.12.9-exec.jar"

codefromthecrypt:
ps this is up to date with master which means the maven package is org.apache.zipkin

(with much rejoicin)

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
```diff
@@ -98,7 +98,10 @@ public void addViewControllers(ViewControllerRegistry registry) {
       // could split the collector's CORS policy into a different property, still allowing POST
       // with content-type by default.
       .allowRequestMethods(HttpMethod.GET, HttpMethod.POST)
-      .allowRequestHeaders(HttpHeaderNames.CONTENT_TYPE);
+      .allowRequestHeaders(HttpHeaderNames.CONTENT_TYPE,
```
codefromthecrypt, line 101: @trustin fyi in armeria repo I mentioned something about different modules needing to collaborate on CORS config. since this is now baked into zipkin server directly now, there is minimal impact to us.

codefromthecrypt:
oops on rebase lost update to zipkin proto 0.2.0

codefromthecrypt:
@shakuzen I just now realize how extremely complex our test setup is vs just using wire (which doesn't require a shell out to protoc). I raised a small revision to our proto as I think now's the time https://github.com/apache/incubator-zipkin-api/pull/64

codefromthecrypt:
will rebase once the following are in:

* reduce dependencies in our service definition https://github.com/apache/incubator-zipkin-api/pull/64
* switch to wire https://github.com/apache/incubator-zipkin/pull/2532

codefromthecrypt:
ok now using a JVM impl square/wire to compile the protos needed for the integration tests. maybe jitpack works now

codefromthecrypt:
jitpack build off last commit is here https://jitpack.io/com/github/ewhauser/zipkin/zipkin-server/experimental-grpc-1.5.0-g94ac6cf-1164/zipkin-server-experimental-grpc-1.5.0-g94ac6cf-1164-exec.jar

codefromthecrypt:
I broke the build settings for integration test (cli only, as intellij is fine).. will fix. here's the command to get to the same fail as travis

```
$ ./mvnw clean verify -pl zipkin-server -o -Dinvoker.skip=true
```

zipkin-server/README.md
```diff
@@ -370,6 +370,20 @@ Example usage:
...
+### gRPC Collector (Experimental)
+
+You can enable a gRPC span collector endpoint by setting `GRPC_SERVER_ENABLED=true`. The
+`zipkin.proto3.SpanService/Report` endpoint will run on the same port as normal http (9411).
+
+
+If you need to deploy a collector-only server, you can disable other http endpoints like so:
+
+```bash
+GRPC_COLLECTOR_ENABLED=true HTTP_COLLECTOR_ENABLED=false QUERY_ENABLED=false java -jar zipkin.jar
```
codefromthecrypt: ps I'm just now noticing our env convention is backwards.

it might be ok to keep it, to be in parity with http. or we can add COLLECTOR_HTTP_ENABLED as a fallback instead of propagating the backwards convention to grpc

codefromthecrypt:
those who want to test other language reporters against this before release can do the following:

```bash
$ curl -sSL https://jitpack.io/com/github/ewhauser/zipkin/zipkin-server/experimental-grpc-1.5.0-g94ac6cf-1164/zipkin-server-experimental-grpc-1.5.0-g94ac6cf-1164-exec.jar > zipkin.jar
$ GRPC_COLLECTOR_ENABLED=true java -jar zipkin.jar 
```

codefromthecrypt:
added a commit to unhook this from snapshots. will merge on green

anuraaga (approved):

codefromthecrypt:
updated description

PR #2332, opened 2018-12-17
Adds AutocompleteTags and corresponding REST api

codefromthecrypt:
This adds an optional storage interface `AutocompleteTags` to the
`StorageComponent` and corresponding endpoints to support the UI.

`/api/v2/autocompleteKeys` and `/api/v2/autocompleteValues?key=http.host`

Cache control is always set on the keys, and it is set conditionally on values when there are more than 3. The "more than 3" allows for a freshly booting system to not confuse users (we do the same thing on `/api/v2/spans?serviceName=foo` endpoint).

The server accepts an environment variable `AUTOCOMPLETE_KEYS`
which acts as a whitelist as storing values for every tag would be
expensive.


codefromthecrypt:
@tacigar @zeagord @shakuzen PTAL I suspect we can merge this on green

zeagord:
Sure. I think so too.

shakuzen (approved):

PR #2334, opened 2018-12-17
Adds MySQL support for Autocomplete tags

codefromthecrypt:


codefromthecrypt:
tested with

```bash
$ STORAGE_TYPE=mysql MYSQL_USER=root AUTOCOMPLETE_KEYS=environment  java -jar ./zipkin-server/target/zipkin-server-*exec.jar
```

then, used curl
```bash
$ curl -s localhost:9411/api/v2/autocompleteKeys
["environment"]
```

zeagord (approved):
Looks good to me. I missed the proper conversion with jooq. Good catch.

shakuzen (approved):

PR #2335, opened 2018-12-17
Adds UI for site tags

zeagord:
screen shot 2561-12-17 at 17 48 51


codefromthecrypt:
personally I would not put arbitrary heading like Tag Key. Instead make a drop-down per key, or auto-complete in the annotation query section (with tabs such as haystack do). We aren't expecting more than several site specific tags.

codefromthecrypt:
also we should mention that redesign in the zipkin-ui code is optional. we'd expect to land lens in a month.. given that, a conservative approach like auto-complete inside the annotationQuery section can make sense (if not hard to impl)

codefromthecrypt:
closing this as we should let this happen in lens

PR #2341, opened 2019-01-03
Adds delay limiter for autocomplete tags

codefromthecrypt:
This reduces index load by obviating calls to store the same autocomplete key/value pair for an hour. An hour allows a storage change to occur and autocomplete tags to re-populate without restarting the server.


codefromthecrypt:
@anuraaga @trustin @huydx this code looks up your alley if you are interested :P

codefromthecrypt:
cc also @llinder @michaelsembwever as I plan to use this to replace the existing stuff in cassandra from muting redundant requests

zeagord (approved):

codefromthecrypt:
pending integration test of elasticsearch limiter. later we should choose a config property for this stuff.

Ex.
```
     * @param suppressionTtl milliseconds to obviate suppress calls to write the same service name
     * @param suppressionMaxSize maximum service names to suppress at a time
```

ideally we have the same property for all autocomplete names, regardless of whether they are service, span or tag autocompletion. I think this would be ideal as it would be less config to present to people.

zipkin/src/main/java/zipkin2/internal/DelayLimiter.java
```diff
@@ -0,0 +1,142 @@
...
+
+  public static <C> DelayLimiter<C> create() {
+    return new Builder().build();
+  }
+
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+
+  public static final class Builder {
```
anuraaga, line 33: Can we add a Ticker here and inject it so no need for the powermocks?

zipkin/src/main/java/zipkin2/internal/DelayLimiter.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  public void clear() {
+    cache.clear();
+    suppressions.clear();
+  }
+
+  /** This attempts to remove the oldest entry to free up one slot */
+  void tryClearOneSuppression() {
+    // This double-checks that another thread didn't remove the same context we peeked
```
anuraaga: Would put this comment on the if statement instead of this, was confused how a var declaration can have that much power

zipkin/src/main/java/zipkin2/internal/DelayLimiter.java
```diff
@@ -0,0 +1,142 @@
...
+
+    Suppression<C> suppression = new Suppression<>(context, System.nanoTime() + ttlNanos);
+
+    if (cache.putIfAbsent(context, suppression) != null) return false; // lost race
+
+    suppressions.offer(suppression);
+
+    // If we added an entry, it could make us go over the max size. This loops until we are no more
+    // than max size.
+    while (suppressions.size() > maximumSize) {
```
anuraaga: This is a bit scary, it seems like in a bad case, which maximumSize wants to protect against, threads may just be spinning on this.

zipkin/src/main/java/zipkin2/internal/DelayLimiter.java
```diff
@@ -0,0 +1,142 @@
...
+    cache.clear();
+    suppressions.clear();
+  }
+
+  /** This attempts to remove the oldest entry to free up one slot */
+  void tryClearOneSuppression() {
+    // This double-checks that another thread didn't remove the same context we peeked
+    Suppression<C> eldest;
+    if ((eldest = suppressions.peek()) != null && suppressions.remove(eldest)) {
+      // This double-checks that another thread didn't resolve the space problem
```
anuraaga: This seems like overkill if the impact is an extra lost suppression or so. It seems like it's safe to remove the while loop above if we loosen the preciseness here which might be safer.

zipkin/src/main/java/zipkin2/internal/DelayLimiter.java
codefromthecrypt: ticker is a guava type.. if it is really important to not use powermock we can..
* make DelayLimiter not final and expose a nanoTime method overridden in tests
* add a custom Ticker type only used in tests

In either case it would only be used in tests, as this is an internal type anyway. Probably the least messy and class bloat way would be to make an overridable method. otherwise we can leave it.

zipkin/src/main/java/zipkin2/internal/DelayLimiter.java
codefromthecrypt: I'm not sure I follow how threads could spin here. we guard max size can't be zero. how can we turn this concern into a test?

zipkin/src/main/java/zipkin2/internal/DelayLimiter.java
codefromthecrypt: I thought so too at first, but then I put in a parallel test and we leaked 12%. The problem this is trying to solve is to make the data structure bounded, which maybe isn't completely clear in the docs..

zipkin/src/main/java/zipkin2/internal/DelayLimiter.java
codefromthecrypt: going with overridable method for now as I also don't like the powermock cruft

codefromthecrypt: ergh it bloats other things to do this as each suppression would also need a ref to the ticker. I think the cure is worse than the disease on this one. We use powermock elsewhere and the same pattern in RateLimiter in brave too.

codefromthecrypt: I'll take it back.. there's only likely 5000 of these suppression references per instance, and max 4 instances (in the cassandra impl: ES has only 1), so who cares :P

codefromthecrypt: now the tests run much much faster :)

zipkin/src/main/java/zipkin2/internal/DelayLimiter.java
codefromthecrypt: yeah we can't change this to an if statement as it leaks bad. For example, in 10k invocations under 2 contended threads it leaks between 100 and 400 with bound set to 1000.

zipkin/src/main/java/zipkin2/internal/DelayLimiter.java
codefromthecrypt: will try a different way before giving up :P

codefromthecrypt: yeah I think you are right. the main goal is not to go over size, vs worrying about going under. I will remove the check here and just let it go to 999 in a race condition which is better than the other way!

codefromthecrypt: I think I got what you mean now. check me!

zipkin/src/main/java/zipkin2/internal/DelayLimiter.java
codefromthecrypt: I changed to not loop on max size and instead just loop on suppressions.peek

zipkin/src/main/java/zipkin2/internal/DelayLimiter.java
codefromthecrypt: rewrote so maybe it is ok now check me!

codefromthecrypt:
great feedback @anuraaga I like the code better now and also the parallel tests run much much faster when there's no powermock instrumentation

michaelsembwever (approved):
lgtm

codefromthecrypt:
@zeagord @huydx ok I think this is ready to test. Only thing I have left is to write the README info on elasticsearch

codefromthecrypt:
ps I'll let the following idea sink in for a day or two. We need a property for this although only cassandra and elasticsearch will currently use it (mysql doesn't)

zipkin.storage.autocomplete-ttl or STORAGE_AUTOCOMPLETE_TTL
* How long in milliseconds to suppress calls to write the same autocomplete key/value pair. Default 3600000 (1 day)

zipkin.storage.autocomplete-max or STORAGE_AUTOCOMPLETE_MAX
* How many autocomplete key/value pairs to suppress at a time. Default 5000

codefromthecrypt:
note STORAGE_AUTOCOMPLETE_MAX (not thrilled with the name) is more about not crashing the server. we can have a number much higher than 5000. This will be the total cardinality of autocomplete tag values, bounded only to stop things from crashing if someone goofed cardinality.

codefromthecrypt:
I'm inclined to rename zipkin.storage.autocomplete-max to zipkin.storage.autocomplete-cardinality set it to a higher number like 20k and not expose an env variable for it.

we can add internal notes but I wouldn't expect users to modify this. we then draw attention to TTL which is the more important config.

wdyt?

zipkin/src/test/java/zipkin2/internal/DelayLimiterTest.java
```diff
@@ -0,0 +1,151 @@
...
+    LongStream.range(0, count).forEach(i -> exec.execute(() -> {
+      if (delayLimiter.shouldInvoke(i)) trueCount.incrementAndGet();
+      latch.countDown();
+    }));
+
+    assertThat(latch.await(1, TimeUnit.SECONDS)).isTrue();
+    exec.shutdown();
+    assertThat(exec.awaitTermination(1, TimeUnit.SECONDS)).isTrue();
+
+    assertThat(trueCount.get()).isEqualTo(count);
```
anuraaga: Dunno if it's kosher for this codebase but awaitility makes these sort of tests simpler without having to worry about latches

anuraaga: Think assertj has an override for atomicinteger, something like `assertThat(trueCount).hasValue(Foo)`

anuraaga (approved):
Thanks for the cleanups, this looks like what I was imagining!

Not exposing the cardinality SGTM.

llinder (approved):

zipkin/src/test/java/zipkin2/internal/DelayLimiterTest.java
codefromthecrypt: thanks for the tip on atomicinteger. on awaitility I'm deferring this by just removing the redundant latch thing as awaitTermination does the same thing anyway :P

codefromthecrypt:
will merge on green. thanks all!

PR #2345, opened 2019-01-09
use the client remote service name instead of the client local service name

fyi-coursera:
If I have a span like below with no "SERVER" span, the UI will display the service name as `sample` but this makes no sense as it is a call into `mysql` from `sample`:

```
{
  ...
  "kind": "CLIENT",
  "name": "query",
  "timestamp": 1547069325788000,
  "duration": 10000,
  "localEndpoint": {
    "serviceName": "sample",
    "ipv4": "127.0.0.1",
    "port": 12345
  },
  "remoteEndpoint": {
    "serviceName": "mysql",
    "ipv4": "10.0.0.0",
    "port": 12345
  },
  "annotations": [
    {
      "timestamp": 1547069325788000,
      "value": "cs"
    },
    {
      "timestamp": 1547069325798000,
      "value": "cr"
    }
  ]
}
```

This PR makes the UI use the client remote service name instead of the client local service name if no server local service name exists. I believe this fixes the leaf span issue mentioned in #2274

### DIFF
--- a/zipkin-ui/js/component_ui/spanRow.js
+++ b/zipkin-ui/js/component_ui/spanRow.js
@@ -289,13 +289,18 @@ export function newSpanRow(spansToMerge) {
       if (!res.duration && next.duration) res.duration = next.duration;
     }
 
-    const nextServiceName = getServiceName(next.localEndpoint);
-    if (nextServiceName && (!res.serviceName || next.kind === 'SERVER')) {
-      res.serviceName = nextServiceName; // prefer the server's service name
+    const nextLocalServiceName = getServiceName(next.localEndpoint);
+    const nextRemoteServiceName = getServiceName(next.remoteEndpoint);
+    if (nextLocalServiceName && next.kind === 'SERVER') {
+      res.serviceName = nextLocalServiceName; // prefer the server's service name
+    } else if (nextRemoteServiceName && next.kind === 'CLIENT' && !res.serviceName) {
+      res.serviceName = nextRemoteServiceName;
+    } else if (nextLocalServiceName && !res.serviceName) {
+      res.serviceName = nextLocalServiceName;
     }
 
-    maybePushServiceName(res.serviceNames, nextServiceName);
-    maybePushServiceName(res.serviceNames, getServiceName(next.remoteEndpoint));
+    maybePushServiceName(res.serviceNames, nextLocalServiceName);
+    maybePushServiceName(res.serviceNames, nextRemoteServiceName);
 
     parseAnnotationRows(next).forEach((a) => maybePushAnnotation(res.annotations, a));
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
hi there. FYI your span data is incorrect as it redundantly has "cr" "cs" and also span.kind. what is generating this?

will consider the change in separate comments. I think we have some issue about remote name policy

codefromthecrypt:
I don't agree that it is nonsense to display the name of the service, btw. that's the primary thing used for everything including query and aggregation.

I think it is safe to say that your concern is that you don't see mysql correct? for example, if you saw both names in some visualization you would be fine correct?

drolando:
I'm +1 on showing the remoteEndpoint name if the server span is missing. For exactly the same reason @fyi-coursera mentioned.

My mysql spans now show up as "serviceX" rather than "mysql" which is very annoying. Same for calls to outside services i.e. AWS or google.

I remember we discussed this a month or so ago and I thought this was already what we agreed upon. And was about to open an issue as well.

codefromthecrypt:
I think this should only apply to leaf spans. otherwise it will definitely mess up association with the dependency graph. please keep rationale discussion on issue #2274 as it is split enough times already

fyi-coursera:
> hi there. FYI your span data is incorrect as it redundantly has "cr" "cs" and also span.kind. what is generating this?
> 
> will consider the change in separate comments. I think we have some issue about remote name policy

sorry, it's generated by some WIP code in the middle of moving from the v1 model to the v2 one. Can I just drop `cs` and `cr` from that span data and be ok?

Will move the rest of the discussion to #2274

codefromthecrypt:
sorry, it's generated by some WIP code in the middle of moving from the v1
> model to the v2 one. Can I just drop cs and cr from that span data and be
> ok?
>
> Will move the rest of the discussion to #2274
> <https://github.com/openzipkin/zipkin/issues/2274>
>

thanks! incidentally, we are closer than might seem. basically the code
must use SpanNode check for children in order to know if it is leaf or not.
if there are no children it is a leaf. In this case, it can prioritize the
client name.. of course the lens UI could show both names

fyi-coursera:
updated to only use client remote service name for leaf spans. Also added a test

codefromthecrypt (approved):
good job.

can you visually verify? Also, @tacigar I think we would need this on lens unless logic is different there.

drolando:
It took me a while to run it, but it works fine for me.

drolando:
yes lens has the same problem

fyi-coursera:
screen shot 2019-01-11 at 10 42 30 am

looks good

codefromthecrypt:
thanks we can sort lens independently

fyi-coursera:
thanks for the help and guidance! really appreciate it

PR #2347, opened 2019-01-11
Fix URL in README

rilleralle:
Existing URL returns a 404.
Use URL https://zipkin.io/pages/tracers_instrumentation


shakuzen (approved):
Thank you for catching and fixing this!

PR #1806, opened 2017-11-20
Switches embedded server from Tomcat to Undertow

codefromthecrypt:
This switches from Tomcat to Undertow, then MVC to raw Undertow handler to process http messages (for apps sending data directly to zipkin).

The test uses a real backend even if local elasticsearch. The test results in a lot more data than my local ES can swallow, so it is also a test of how fast we can drop data, and how well our rate-limiting works.

So, after this change, the same zipkin server against local elasticsearch can swallow 3x more traffic than before.

The below results were a best of 5 for before and after.. tests were interleaved to ensure elasticsearch wasn't skewing.

```
# before
Running 30s test @ http://localhost:9411/api/v2/spans
  4 threads and 128 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    14.50ms   37.19ms 527.63ms   93.43%
    Req/Sec     2.02k     0.94k    6.66k    71.75%
  Latency Distribution
     50%    4.20ms
     75%    8.61ms
     90%   24.53ms
     99%  198.98ms
  212147 requests in 30.05s, 24.92MB read
  Socket errors: connect 0, read 1, write 0, timeout 19
Requests/sec:   7058.96
Transfer/sec:    849.19KB

# after
Running 30s test @ http://localhost:9411/api/v2/spans
  4 threads and 128 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    16.28ms   33.11ms 388.07ms   90.37%
    Req/Sec     6.43k     2.86k   14.55k    64.65%
  Latency Distribution
     50%    3.03ms
     75%   14.59ms
     90%   47.73ms
     99%  155.01ms
  761153 requests in 30.05s, 76.22MB read
Requests/sec:  25332.61
Transfer/sec:      2.54MB
```

Here's the command which starts a server and uses elasticsearch local storage:
```bash
$ STORAGE_TYPE=elasticsearch java -jar ./zipkin-server/target/zipkin-server-*-exec.jar
$ wrk -t4 -c128 -d30s http://localhost:9411/api/v2/spans -s zipkin-json.lua --latency
```

Here's the script:
```lua
wrk.method = "POST"
wrk.headers["Content-Type"] = "application/json"
[...]

PR #2348, opened 2019-01-12
Convert to armeria for http engine

codefromthecrypt:
This is mostly a shotgun attempt. The integration tests are likely
failing as the spring boot runner likely wants to open a tomcat port
which armeria would conflict with. Not sure about it actually. Anyway,
this is a rough start.

cc @trustin @anuraaga @huydx


codefromthecrypt:
fyi if someone is up to help advise or otherwise add a commit here it will progress quicker. due to travel etc I will have limited focus for a while

pom.xml
```diff
@@ -249,6 +250,18 @@
         <version>${jooq.version}</version>
       </dependency>
 
+      <dependency>
+        <groupId>com.linecorp.armeria</groupId>
+        <artifactId>armeria-spring-boot-autoconfigure</artifactId>
+        <version>${armeria.version}</version>
+        <exclusions>
+          <exclusion>
+            <groupId>com.linecorp.armeria</groupId>
```
codefromthecrypt, line 259: so we use log4j2 and this dep makes boot up angry

zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
```diff
@@ -103,6 +103,10 @@ HttpTracing httpTracing(Tracing tracing) {
         .build();
   }
 
+  @Bean ArmeriaServerConfigurator tracingConfigurator(Tracing tracing) {
```
codefromthecrypt, line 108: also I haven't used the special currenttracecontext armeria provider here yet

zipkin-server/src/test/java/zipkin2/server/internal/ITZipkinMetricsHealth.java
```diff
@@ -45,7 +45,10 @@
 @SpringBootTest(
   classes = ZipkinServer.class,
   webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
-  properties = "spring.config.name=zipkin-server"
+  properties = {
+    "spring.config.name=zipkin-server",
+    "spring.main.web-application-type=none"
```
codefromthecrypt: this might be redundant but this test currently dies starting. making that not the case would be progress!

anuraaga: Just a quick note, did you try setting webEnvironment above to `NONE`?

codefromthecrypt:
nope didn't try that

codefromthecrypt:
fyi battery holding I plan to put a couple more hours into this today while
on a flight

codefromthecrypt:
I made significant progress. will update later today.

codefromthecrypt:
current status:

a lot works except for some reason..
* POST endpoints that have content-type headers aren't getting invoked
* actuator endpoints don't work

see also: probably we want to write the "zipkin-ui" forwarding rules into some equivalent in armeria though I don't know what that is (possibly just normal java method dispatch). Also zipkin-ui loads assets from a jar file, so there's probably some armeria equivalent of that (right now MVC stuff doesn't work at all)

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
```diff
@@ -99,19 +89,19 @@ CollectorMetrics metrics(MeterRegistry registry) {
   @Bean
   public MeterRegistryCustomizer meterRegistryCustomizer() {
     return registry ->
-        registry
```
codefromthecrypt, line 106: indent changes aren't intentional.. just I've not cleaned them out

zipkin-server/src/main/resources/zipkin-server-shared.yml
```diff
@@ -143,14 +143,24 @@ zipkin:
...
     mime-types: application/json,application/javascript,text/css,image/svg
 
+# Disable the Web Server.
+# See https://docs.spring.io/spring-boot/docs/current/reference/html/howto-embedded-web-servers.html.
+spring.main.web-application-type: none
+
+# Additional server ports specified with Armeria configuration.
+armeria:
+  ports:
+    - port: ${QUERY_PORT:${server.port:9411}}
```
codefromthecrypt: oooo nesting...

trustin:
Will take a look soon.

codefromthecrypt:
ok I have a commit not yet pushed from flight. you can delay until then. on
flight got far except getting actuator and webmvc working

codefromthecrypt:
@trustin pushed.. so I think if someone can get the zipkin-autoconfigure/ui module working I can probably get the rest to fall in line. I think it is only the spring related things not working at this point (actuator, mvc forwarding rules for the UI which imho can be redone in pure armeria if helps just I don't know how to get the resource tricks spring does to work in armeria)

anuraaga:
I want to mess with this branch. Is the best way to create a PR to the `armeria-server` branch from a fork? Just wondering what's easiest.

Also concept-wise, armeria can serve servlet-api using Tomcat as a translation layer - not a separate port / tcp handler, just uses the business logic. What do you think of serving Zipkin API using armeria directly for optimal performance while also letting low QPS endpoints like actuator to be served using the translation layer? It adds a dependency on Tomcat, but has the upside of guaranteeing full spring compatibility on endpoints we delegate to that layer.

codefromthecrypt:
@anuraaga a pull against this branch works for me. I didn't realize you can use tomcat as a translation layer. Your rationale makes sense.

codefromthecrypt:
actually I just gave you karma to push directly. that is mutually easier as
then you can get it to run travis etc.

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
```diff
@@ -13,168 +13,155 @@
...
-      int len;
-      while ((len = in.read(buf)) > 0) {
-        outputStream.write(buf, 0, len);
-      }
-      return outputStream.toByteArray();
+  @Override
+  public HttpResponse handleException(RequestContext ctx, HttpRequest req, Throwable cause) {
+    ZipkinHttpCollector.metrics.incrementMessagesDropped();
+    if (cause instanceof IllegalArgumentException) {
+      return HttpResponse.of(BAD_REQUEST, MediaType.ANY_TEXT_TYPE, cause.getMessage());
```
hyangtack, line 181: `MediaType.ANY_TEXT_TYPE` is for `text/*`, so I think you need to use some other media type such as `MediaType.PLAIN_TEXT_UTF_8`.

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java
```diff
@@ -71,116 +64,109 @@
...
-      @Nullable @RequestParam(value = "annotationQuery", required = false) String annotationQuery,
-      @Nullable @RequestParam(value = "minDuration", required = false) Long minDuration,
-      @Nullable @RequestParam(value = "maxDuration", required = false) Long maxDuration,
-      @Nullable @RequestParam(value = "endTs", required = false) Long endTs,
-      @Nullable @RequestParam(value = "lookback", required = false) Long lookback,
-      @RequestParam(value = "limit", defaultValue = "10") int limit)
-      throws IOException {
+  @Get("/api/v2/traces")
+  public HttpResponse getTraces(
+    @Param("serviceName") Optional<String> serviceName,
```
hyangtack, line 91: You may know but you can omit the value of `@Param` annotation if you specify `-parameters` javac option.
https://line.github.io/armeria/server-annotated-service.html#parameter-injection

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
```diff
@@ -65,23 +73,21 @@ public void addViewControllers(ViewControllerRegistry registry) {
...
-    return factory;
+  @Bean ArmeriaServerConfigurator corsConfigurator(
+    @Value("${zipkin.query.allowed-origins:*}") String allowedOrigins) {
+    CorsServiceBuilder corsBuilder = allowedOrigins.equals("*") ? CorsServiceBuilder.forAnyOrigin()
+      : CorsServiceBuilder.forOrigins(allowedOrigins.split(","));
+
+    Function<Service<HttpRequest, HttpResponse>, CorsService>
+      corsDecorator = corsBuilder.allowRequestMethods(HttpMethod.GET).newDecorator();
+
+    return server -> server
```
hyangtack, line 98: Actually it's enough that you just specify `return sb -> sb.decorator(corsDecorator);`. But I guess we have a problem CORS with annotated services. Let me check it, then let you know.

zipkin-server/src/main/resources/zipkin-server-shared.yml
hyangtack: I'm not sure but isn't it enough to specify `${QUERY_PORT:9411}`?

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
hyangtack: Filed an issue: https://github.com/line/armeria/issues/1531

anuraaga:
Added a commit that forwards unhandled requests through the tomcat layer. Interestingly, we're back to the standard armeria-spring example instead of the minimal one. It seems to make sense for zipkin-server to include a servlet context for things like the UI unless we instead ported the latter off of webmvc.

codefromthecrypt:
thanks for the unblocking. there is definitely progress here. the jar is much larger now, which is a bit of a problem as it is a lot larger. A good bit of it is tomcat, though some of the space is due to things like boringssl which is possible for us to punt or just accept...

codefromthecrypt:
```
# this version
$ du -k  ./zipkin-server/target/zipkin-server-*exec.jar
51572	./zipkin-server/target/zipkin-server-2.11.13-SNAPSHOT-exec.jar

# master before this change
$ du -k zipkin.jar
38760	zipkin.jar
```

codefromthecrypt:
I'll try to figure out which jars are responsible.. a few mins..

codefromthecrypt:
here are the differences... besides the huge tomcat dep validation is also suspect. I'm not sure we'd strictly want this as we have so few things to validate...

left is armeria right is before
```diff
1,3d0
< 5817106 Mon Dec 24 15:44:12 IST 2018 BOOT-INF/lib/armeria-0.78.1.jar
< 3235448 Fri Nov 02 14:29:18 IST 2018 BOOT-INF/lib/tomcat-embed-core-9.0.13.jar
< 3016315 Thu Nov 15 13:36:22 IST 2018 BOOT-INF/lib/netty-tcnative-boringssl-static-2.0.20.Final.jar
6a4
> 2256120 Tue Nov 27 11:39:30 IST 2018 BOOT-INF/lib/undertow-core-2.0.16.Final.jar
14d11
< 1155681 Wed Aug 22 22:16:56 IST 2018 BOOT-INF/lib/hibernate-validator-6.0.13.Final.jar
22d18
< 586456 Mon Oct 29 15:49:58 IST 2018 BOOT-INF/lib/netty-common-4.1.31.Final.jar
24c20,21
< 563216 Mon Oct 29 16:03:30 IST 2018 BOOT-INF/lib/netty-codec-http-4.1.31.Final.jar
---
> 582362 Mon Aug 13 11:09:14 IST 2018 BOOT-INF/lib/xnio-api-3.6.5.Final.jar
> 531113 Tue Nov 27 11:42:22 IST 2018 BOOT-INF/lib/undertow-servlet-2.0.16.Final.jar
26d22
< 463017 Mon Oct 29 15:54:22 IST 2018 BOOT-INF/lib/netty-transport-4.1.31.Final.jar
30,31d25
< 419511 Mon Oct 29 16:00:18 IST 2018 BOOT-INF/lib/netty-handler-4.1.31.Final.jar
< 404008 Mon Oct 29 16:05:56 IST 2018 BOOT-INF/lib/netty-codec-http2-4.1.31.Final.jar
37d30
< 316560 Mon Oct 29 15:55:48 IST 2018 BOOT-INF/lib/netty-codec-4.1.31.Final.jar
40d32
< 273373 Mon Oct 29 15:51:52 IST 2018 BOOT-INF/lib/netty-buffer-4.1.31.Final.jar
42d33
< 249951 Fri Nov 02 14:29:20 IST 2018 BOOT-INF/lib/tomcat-embed-el-9.0.13.jar
44a36,37
> 224014 Thu Apr 25 16:03:30 IST 2013 BOOT-INF/lib/javax.el-3.0.0.jar
> 221719 Thu Jan 11 10:35:18 IST 2018 BOOT-INF/lib/wildfly-common-1.3.0.Final.jar
45a39
> 166279 Fri Mar 16 08:35:42 IST 2018 BOOT-INF/lib/jboss-threads-2.3.2.Final.jar
48d41
< 141019 Mon Oct 29 16:42:10 IST 2018 BOOT-INF/lib/netty-transport-native-epoll-4.1.31.Final-linux-x86_64.jar
51d43
< 128535 Mon Oct 29 16:17:42 IST 2018 BOOT-INF/lib/netty-resolver-dns-4.1.31.Final.jar
52a45
> 121323 Mon Aug 13 11:10:02 IST 2018 BOOT-INF/lib/xnio-nio-3.6.5.Final.jar
58d50
<  93107 Tue Dec 19 16:23:28 IST 2017 BOOT-INF/lib/validation-api-2.0.1.Final.jar
61d52
<  66540 Tue Mar 27 18:35:34 IST 2018 BOOT-INF/lib/classmate-1.4.0.jar
66,67d56
<  55421 Mon Oct 29 16:30:18 IST 2018 BOOT-INF/lib/netty-transport-native-unix-common-4.1.31.Final-linux-x86_64.jar
<  54859 Mon Oct 29 15:57:40 IST 2018 BOOT-INF/lib/netty-codec-dns-4.1.31.Final.jar
68a58
>  47189 Thu Aug 10 11:34:20 IST 2017 BOOT-INF/lib/wildfly-client-config-1.0.0.Final.jar
70d59
<  42659 Mon Dec 24 15:43:38 IST 2018 BOOT-INF/lib/armeria-tomcat-0.78.1.jar
73d61
<  33471 Mon Oct 29 18:10:02 IST 2018 BOOT-INF/lib/netty-transport-native-unix-common-4.1.31.Final.jar
75d62
<  32800 Mon Oct 29 15:52:28 IST 2018 BOOT-INF/lib/netty-resolver-4.1.31.Final.jar
80,81d66
<  25499 Mon Dec 24 15:43:42 IST 2018 BOOT-INF/lib/armeria-spring-boot-autoconfigure-0.78.1.jar
<  25305 Mon Oct 29 15:58:58 IST 2018 BOOT-INF/lib/netty-codec-haproxy-4.1.31.Final.jar
87,89c72
<  19936 Fri Mar 31 10:55:30 IST 2017 BOOT-INF/lib/jsr305-3.0.2.jar
<  16515 Thu Mar 16 17:37:30 IST 2017 BOOT-INF/lib/jcl-over-slf4j-1.7.25.jar
<  16465 Wed Jul 18 22:35:12 IST 2018 BOOT-INF/lib/metrics-json-4.0.3.jar
---
>  19834 Wed Jan 17 13:26:46 IST 2018 BOOT-INF/lib/jboss-annotations-api_1.2_spec-1.0.2.Final.jar
91d73
<  12966 Fri Nov 02 14:28:20 IST 2018 BOOT-INF/lib/tomcat-annotations-api-9.0.13.jar
99,100c81
<   2497 Tue Oct 13 16:07:04 IST 2009 BOOT-INF/lib/javax.inject-1.jar
<   2097 Mon Dec 18 21:06:08 IST 2017 BOOT-INF/lib/reactive-streams-1.0.2.jar
---
>    425 Fri Nov 30 08:48:34 IST 2018 BOOT-INF/lib/spring-boot-starter-undertow-2.1.1.RELEASE.jar
```

anuraaga:
Thank for listing up the diffs. I remembered that we should be able to use jetty similar to tomcat here, as a servlet-api translator. It might be smaller. Though actually I didn't look too much into the webmvc code yet should we just port it to annotated services?

Since the default config for zipkin uses http instead of https, excluding boringssl could make sense. JDK9+ users wouldn't really need it anyways. Another option is to include the linux-only jar, armeria includes the uber jar for ease of getting started on various OS.

hibernate-validator I had always assumed is a requirement of spring boot - if not I suspect there's sometime that could change on armeria side to make it optional.

And finally a large part of the armeria jar is the doc service webapp. I had originally thought of making it a separate jar in case users want to exclude it but @trustin suggested keeping it in core for simplicity. @trustin any thoughts? Though the doc service's swagger support might be useful in zipkin server anyways.

anuraaga:
Realized ui is just a simple controller mostly returning static files so it was easy to port to armeria. Actually @trustin just added some much easier APIs to add cache headers + redirect to index.html so we should wait until next armeria release before finishing those up. But managed to remove tomcat and validator.

codefromthecrypt:
thanks @anuraaga! I will try to help with polishing on this

codefromthecrypt:
fyi I am stepping off this branch again as I have to pack and move tomorrow. I didn't look into deeply but there is a test failing in zipkin-ui autoconfig

anuraaga:
Yeah sorry should have mentioned there's some known failing tests. After the next armeria release, which I think is soon, it'll be much easier to fix them :)

trustin:
Armeria 0.79.0 has been released. :-)

trustin:
Didn't think much about making DocService as an optional dependency but we may have to eventually once we add more bells and whistles to it, like metric dashboard and on demand profiler.

codefromthecrypt:
fyi I will work on this next 3hrs offline

zipkin-autoconfigure/ui/src/main/java/zipkin2/autoconfigure/ui/ZipkinUiAutoConfiguration.java
```diff
@@ -94,66 +103,35 @@ String processedIndexHtml() throws IOException {
...
-    response.setStatus(HttpStatus.FOUND.value());
+    return sb -> {
+      sb
+        .service("/zipkin/config.json",
+          (((ctx, req) -> HttpResponse.of(CONFIG_HEADERS, HttpData.of(config)))))
+        .service("/zipkin/index.html",
+          ((ctx, req) -> HttpResponse.of(INDEX_HEADERS, HttpData.of(index))))
+        .serviceUnder("/zipkin/", uiFileService)
+        .service("/favicon.ico", new RedirectService(HttpStatus.FOUND, "/zipkin/favicon.ico"))
+        .service("/", new RedirectService(HttpStatus.FOUND, "/zipkin/"));
```
huydx: you may want to pass meter registry to armeria too (`sb.meterRegistry(registry)`)

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java
```diff
@@ -71,116 +64,109 @@
...
+      .contentType(MediaType.JSON)
+      .setInt(HttpHeaderNames.CONTENT_LENGTH, body.length);
+    if (shouldCacheControl) {
+      headers = headers.add(
+        HttpHeaderNames.CACHE_CONTROL,
+        CacheControl.maxAge(namesMaxAge, TimeUnit.SECONDS).mustRevalidate().getHeaderValue()
+      );
     }
-    return response.body(names);
+    return AggregatedHttpMessage.of(headers, HttpData.of(body));
```
huydx, line 169: not sure it must be `AggregatedHttpMessage` or we could just use HttpResponse.of() here ? cc @hyangtack

trustin: Yep. A user can use `HttpResponse.of(...)`. `AggregatedHttpMessage` will be translated into `HttpResponse` anyway.

trustin:
See https://line.github.io/armeria/server-http-file.html for `HttpFile` API.

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java
anuraaga: I think HttpReaponse will be on the event loop, not blocking executor. There's some blocking here (though easily made async which we should at some point)

codefromthecrypt:
if someone's up to progress this farther today, I think I have unblocked the config-related problems.

quick glance at test failures suggest actuator isn't working, then there is some slight mismatch in cors and redirect tests

codefromthecrypt:
PS I have another flight or two today so will stab at this a bit more

anuraaga:
Realized that even though we ported UI, actuator will still use servlet-api and would need it included. Could either manually implement actuator endpoints as armeria services or try servlet-api again, maybe this time with jetty to see if it's smaller.

codefromthecrypt:
makes sense. yeah it is possible jetty could be smaller.. good thinking.

I wouldnt guess it viable to try to implement actuator manually as there
are a bunch of things in there.. maybe we can raise an issue so spring boot
3 admin endpoints are not strictly requiring servlet.. possibly that or
looking at what an armeria servlet context would look like.

codefromthecrypt:
@anuraaga I realized on a flight that you probably meant implementing support for generic actuator endpoints. I spiked this on the flight and seems to begin progress though the last commit is just proof-concept in nature. It doesn't appear actuator has a hard-dependency on servlet context iotw. So, we need to think about whether to continue this work here or in armeria.

trustin:
> we need to think about whether to continue this work here or in armeria.

Sounds like fun. Would be a nice addition to our Spring integration.

zipkin-server/src/main/java/zipkin2/server/internal/actuate/TolerantPathMapping.java
```diff
@@ -0,0 +1,149 @@
...
+import com.linecorp.armeria.server.PathMapping;
+import com.linecorp.armeria.server.PathMappingContext;
+import com.linecorp.armeria.server.PathMappingResult;
+import com.linecorp.armeria.server.VirtualHost;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import javax.annotation.Nullable;
+
+final class TolerantPathMapping implements PathMapping {
```
anuraaga: I'm looking at making an armeria module for actuator, mostly by just following this code. Just wondering, why did you need this class? I'm not really following what it does, but suspect a change in core could be made to avoid it.

codefromthecrypt: took me a few minutes to remember!

yeah so we want to handle when no content type header is sent at all.

anuraaga: Thanks, will check it. Wouldn't have figured that out since from what I can tell, `/loggers` is the only default endpoint that sets `consumes`!

codefromthecrypt:
yeah this is for the zipkin http collector endpoints POST /api/v2/spans for
example

anuraaga:
Ah - in that case I suspect it's not needed. That API is using annotated service, not actuator. From what I can tell from these docs, actuator requires content type header so the default path mapping probably works

https://docs.spring.io/spring-boot/docs/current/actuator-api/html/#loggers-single

codefromthecrypt:
dunno probably something must have failed else I wouldnt have written this.
the version of me that knew was force pushed

codefromthecrypt:
@anuraaga I played with building from your branch of armeria, but I get a mysterious NPE in the exploratory commit

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-invoker-plugin:3.2.0:install (integration-test) on project zipkin-server: Failed to install project dependencies: MavenProject: io.zipkin.java:zipkin-server:2.12.2-SNAPSHOT @ /Users/acole/oss/zipkin/zipkin-server/pom.xml: Failed to install project artifacts: null: Failed to install POM: null: NullPointerException -> [Help 1]
```

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
anuraaga: Hi @hyangtack - as that issue has been fixed, I tried deleting the below preflight hack but it still didn't work. Is there something else that needs to be done too?

anuraaga:
Updated this branch and CircleCI passes now. I made one behavior change in that when collector is disabled, it returns 404 instead of 405 - the latter seemed complicated and also surprised me, if the API isn't on shouldn't it be 404? Anyways, hopefully it's a small enough change.

I think there are some more code cleanups possible but at least things are merged and probably working - Travis seems to get stuck on not being able to connect to Cassandra but I'm hoping that's not related to armeria.

codefromthecrypt:
will take a look. thanks!

codefromthecrypt:
quick look noticed we drifted you. if you dont have time to adjust for the
new dual mounted ui stuff, I can take a look later maybe Tuesday.

have a presentation to write then a family visit meanwhile


thanks for the continued swing on this. I also want this merged asap

anuraaga:
While going through my github.com/pulls, I found this blast from the past!

https://github.com/anuraaga/zipkin-java/pull/2

@trustin remember those old APIs? :D

anuraaga:
@adriancole I noticed the drift, I think I merged it ok and the integration tests run (admittedly only running the tests, not the UI itself) but did you notice something that doesn't look right related to the dual mounted? Anyways, in no rush so could wait until Tuesday or whenever

trustin:
> @trustin remember those old APIs? :D

Ohh, that's a very long time ago. :rofl: Never knew you tried to replace Spring Web at that time! I hope we are getting better and better in terms of API usability and user happiness. :crossed_fingers:

codefromthecrypt:
I'll try to figure out what this is tonight.

```
[ERROR]   ITZipkinMetricsHealth.writeSpans_updatesPrometheusMetrics:165 NullPointer
```

codefromthecrypt:
actually ITZipkinMetricsHealth is an unrelated flake. there's no reason it should pass or not pass based on this code change.

codefromthecrypt:
using in mem and the script here one request bombed like this. not critical but after merge possibly interesting to folks maybe @trustin 

https://github.com/openzipkin/zipkin/pull/1806#issue-153649052

```
2019-03-06 17:15:58.938  WARN 69546 --- [worker-nio-2-15] c.l.a.s.HttpServerHandler                : [id: 0xba9745a4, L:0.0.0.0/0.0.0.0:9411 ! R:/0:0:0:0:0:0:0:1:62432][h1c] Unexpected exception:

java.io.IOException: Protocol wrong type for socket
	at sun.nio.ch.FileDispatcherImpl.write0(Native Method) ~[?:?]
	at sun.nio.ch.SocketDispatcher.write(SocketDispatcher.java:47) ~[?:?]
	at sun.nio.ch.IOUtil.writeFromNativeBuffer(IOUtil.java:113) ~[?:?]
	at sun.nio.ch.IOUtil.write(IOUtil.java:58) ~[?:?]
	at sun.nio.ch.IOUtil.write(IOUtil.java:50) ~[?:?]
	at sun.nio.ch.SocketChannelImpl.write(SocketChannelImpl.java:466) ~[?:?]
	at io.netty.channel.socket.nio.NioSocketChannel.doWrite(NioSocketChannel.java:405) ~[netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.AbstractChannel$AbstractUnsafe.flush0(AbstractChannel.java:938) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.flush0(AbstractNioChannel.java:360) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.AbstractChannel$AbstractUnsafe.flush(AbstractChannel.java:905) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.DefaultChannelPipeline$HeadContext.flush(DefaultChannelPipeline.java:1370) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:776) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush(AbstractChannelHandlerContext.java:768) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.AbstractChannelHandlerContext.flush(AbstractChannelHandlerContext.java:749) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.ChannelDuplexHandler.flush(ChannelDuplexHandler.java:117) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:776) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush(AbstractChannelHandlerContext.java:768) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.AbstractChannelHandlerContext.flush(AbstractChannelHandlerContext.java:749) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.handler.flush.FlushConsolidationHandler.flushNow(FlushConsolidationHandler.java:206) [netty-handler-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.handler.flush.FlushConsolidationHandler.flushIfNeeded(FlushConsolidationHandler.java:199) [netty-handler-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.handler.flush.FlushConsolidationHandler.resetReadAndFlushIfNeeded(FlushConsolidationHandler.java:194) [netty-handler-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.handler.flush.FlushConsolidationHandler.channelReadComplete(FlushConsolidationHandler.java:147) [netty-handler-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelReadComplete(AbstractChannelHandlerContext.java:398) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelReadComplete(AbstractChannelHandlerContext.java:380) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelReadComplete(AbstractChannelHandlerContext.java:373) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.handler.timeout.IdleStateHandler.channelReadComplete(IdleStateHandler.java:295) [netty-handler-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelReadComplete(AbstractChannelHandlerContext.java:398) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelReadComplete(AbstractChannelHandlerContext.java:380) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelReadComplete(AbstractChannelHandlerContext.java:373) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelReadComplete(DefaultChannelPipeline.java:1413) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelReadComplete(AbstractChannelHandlerContext.java:398) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelReadComplete(AbstractChannelHandlerContext.java:380) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.DefaultChannelPipeline.fireChannelReadComplete(DefaultChannelPipeline.java:936) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:168) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:677) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:612) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:529) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:491) [netty-transport-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:905) [netty-common-4.1.33.Final.jar!/:4.1.33.Final]
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) [netty-common-4.1.33.Final.jar!/:4.1.33.Final]
	at java.lang.Thread.run(Thread.java:834) [?:?]
```

codefromthecrypt:
ps above is not a problem as it was one of hundreds of thousands of requests

codefromthecrypt:
I'm very happy with this change. I ran wrk and found that this results in far more consistent latency, which is great for apps

BEFORE
```
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     9.03ms   27.60ms 275.16ms   92.96%
    Req/Sec    16.09k     4.18k   23.37k    82.48%
  Latency Distribution
     50%    1.78ms
     75%    2.01ms
     90%    3.92ms
     99%  148.75ms
  1874690 requests in 30.03s, 212.75MB read
Requests/sec:  62421.95
Transfer/sec:      7.08MB
```

AFTER
```
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     4.92ms    5.08ms  63.50ms   95.37%
    Req/Sec     7.62k     1.29k   10.95k    70.99%
  Latency Distribution
     50%    3.46ms
     75%    4.07ms
     90%    6.93ms
     99%   32.97ms
  909271 requests in 30.00s, 84.98MB read
Requests/sec:  30304.72
Transfer/sec:      2.83MB
```

codefromthecrypt:
@anuraaga you are the hero of this change. thank you so much

trustin:
I'm curious why we have half the original throughput. Is this expected? I'd like to do some profiling later. Would you mind letting me know how to reproduce the benchmark, e.g. how to set up and what wrk command to run.

codefromthecrypt:
@trustin 

so here's the thing.. build the and start the server per the bottom of zipkin-server/README.md

Make zipkin-json.lua 
```lua
wrk.method = "POST"
wrk.headers["Content-Type"] = "application/json"
wrk.body   = '[{"traceId":"4d1e00c0db9010db86154a4ba6e91385","parentId":"86154a4ba6e91385","id":"4d1e00c0db9010db","kind":"CLIENT","name":"get","timestamp":1472470996199000,"duration":207000,"localEndpoint":{"serviceName":"frontend","ipv6":"7::0.128.128.127"},"remoteEndpoint":{"serviceName":"backend","ipv4":"192.168.99.101","port":9000},"annotations":[{"timestamp":1472470996238000,"value":"foo"},{"timestamp":1472470996403000,"value":"bar"}],"tags":{"http.path":"/api","clnt/finagle.version":"6.45.0"}}]'
```

run it 5 times picking best

```bash
wrk -t4 -c128 -d30s http://localhost:9411/api/v2/spans -s zipkin-json.lua --latency
```

you can compare this vs the previous commit

codefromthecrypt:
@trustin ps I think the previous had better throughput as the P<=90 was faster

jcarres-mdsol:
Just curious, in your performance numbers above there are less than half the req/sec in the after compared with the before

codefromthecrypt:
@jcarres-mdsol yes though the >90% latency is much better which is indicative of what zipkin would be under load. Also, this is a dummy request.. I doubt most real servers were servicing 60k requests/second in real life with the resources of my laptop.

codefromthecrypt:
reason I'm less concerned about the raw throughput difference iotw, is that usually storage falls over faster than this.

codefromthecrypt:
for those watching from home, a more realistic test would be to load up a megabyte of bundled span data per POST message as opposed to a single tiny span. Then see about 240MB/s (peak throughput site like netflix)/ however many servers that site uses. Also consider non-peak. Also consider that most high throughput sites don't even use http (for example most use kafka). That's how I end up not terribly concerned about the amount of requests per second even if this "hello span" might seem damning if looking mostly at the raw throughput.

That plus there's likely some optimizations to follow. make sense?

codefromthecrypt:
here's another datapoint.. my laptop before this change with docker and mysql attached as opposed to in-memory
```
    Latency    27.29ms   17.77ms 164.65ms   80.97%
    Req/Sec     1.24k   430.38     2.42k    75.96%
  Latency Distribution
     50%   23.40ms
     75%   33.27ms
     90%   47.81ms
     99%   94.82ms
  12846 requests in 2.60s, 1.46MB read
  Socket errors: connect 0, read 75, write 0, timeout 0
Requests/sec:   4931.43
Transfer/sec:    573.09KB
```

basically we can isolate some things about http, but all-in this "wrk" test is not accurate for measuring integrated actual usage unless you are hitting your own install.

codefromthecrypt:
a cooler test is also to hit one of the -example apps, like https://github.com/openzipkin/zipkin-ruby-example. This will make more realistic span data although definitely unrealistic to most sites as only 2 services.

ex `wrk -t4 -c128 -d30s http://localhost:8081  --latency`

then look at prometheus stats on the zipkin server instead of wrk stats (which will only tell you about the example app). anyway this is fun.

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
hyangtack: Really? I will take a look again!

jcarres-mdsol:
Ruby will definitely be the bottleneck.
Thanks for the details!

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
codefromthecrypt: follow-up PRs welcome. I mainly merged this to get @anuraaga out of the work pit of rebasing massive shifts. looks good enough to move forward with, but yeah any follow-ups very welcome!

trustin:
@adriancole Thanks. Will take a look to find out what we can do to make it perform better. 
@anuraaga Great job!

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
anuraaga: @hyangtack Thanks! By the way I should have mentioned that by "didn't work", I meant the integration test didn't succeed. I didn't dig at all so there is also the chance the test itself needs an update and not Armeria (I don't know cors very well and it's not something I want to add to my life ;) )

codefromthecrypt:
Now that we support http/2, I was able to make a few runs to show some differences, using the same test json, against the same server. While this is laptop, it is an interesting thing we can now use.

http 1 (text framing) best run in 5, because toes..
```
$ h2load -n100000 -c100 -m10 http://localhost:9411/api/v2/spans -d fooooo.json --h1
starting benchmark...
spawning thread #0: 100 total client(s). 100000 total requests
Application protocol: http/1.1
progress: 10% done
progress: 20% done
progress: 30% done
progress: 40% done
progress: 50% done
progress: 60% done
progress: 70% done
progress: 80% done
progress: 90% done
progress: 100% done

finished in 3.07s, 32558.26 req/s, 3.04MB/s
requests: 100000 total, 100000 started, 100000 done, 100000 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 100000 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 9.35MB (9800000) total, 5.05MB (5300000) headers (space savings 0.00%), 1.14MB (1200000) data
                     min         max         mean         sd        +/- sd
time for request:      107us    341.96ms      2.96ms     12.15ms    99.32%
time for connect:      215us      6.37ms      2.58ms      1.61ms    64.00%
time to 1st byte:     7.42ms     17.79ms     12.71ms      2.53ms    63.00%
req/s           :     325.86      363.59      337.85        8.32    70.00%
```

http 2 (binary framing) best run in 5, because toes..
```
$ h2load -n100000 -c100 -m10 http://localhost:9411/api/v2/spans -d fooooo.json 
starting benchmark...
spawning thread #0: 100 total client(s). 100000 total requests
Application protocol: h2c
progress: 10% done
progress: 20% done
progress: 30% done
progress: 40% done
progress: 50% done
progress: 60% done
progress: 70% done
progress: 80% done
progress: 90% done
progress: 100% done

finished in 2.01s, 49772.07 req/s, 2.24MB/s
requests: 100000 total, 100000 started, 100000 done, 100000 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 100000 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 4.50MB (4713900) total, 295.41KB (302500) headers (space savings 95.20%), 1.14MB (1200000) data
                     min         max         mean         sd        +/- sd
time for request:      550us    231.94ms     19.42ms     26.36ms    95.22%
time for connect:      457us      7.64ms      4.09ms      2.11ms    58.00%
time to 1st byte:    16.73ms    105.63ms     31.54ms     14.78ms    89.00%
req/s           :     498.31      537.56      511.34        9.70    70.00%
```

For those using okhttp sender, you can use http/2 like below, but make sure you are using the latest zipkin server as prior to that, it didn't accept http/2 requests!

```java
  @Bean Sender sender() {
    OkHttpSender.Builder senderBuilder = OkHttpSender.newBuilder();
    senderBuilder.clientBuilder().protocols(Collections.singletonList(Protocol.H2_PRIOR_KNOWLEDGE));
    return senderBuilder.endpoint("http://127.0.0.1:9411/api/v2/spans").build();
  }
```

Issue #2321, opened 2018-12-10
Make zipkin-lens use the same or similar color scheme to the previous UI


codefromthecrypt:
We've had a number of informal feedback that people like the old color scheme. Let's switch zipkin-lens to something that is less shocking?

tacigar:
OK!

tacigar:
Done.
Can I close this issue?

codefromthecrypt:
I will close in a sec.. just waiting for latest docker build and will
verify manually :D

codefromthecrypt:
oops I think I released docker too fast :) few more minutes

codefromthecrypt:
<img width="971" alt="screen shot 2019-01-24 at 5 06 12 pm" src="https://user-images.githubusercontent.com/64215/51667364-e40f3e00-1ffa-11e9-9114-fd05022c69f3.png">

PR #2354, opened 2019-01-17
Change theme of Dependencies page

tacigar:
Related #2321


tacigar:
dependencies

zeagord (approved):
LGTM. Font is little lighter. But it is ok, we can change that later.

tacigar:
Thank you!!!

Issue #2351, opened 2019-01-16
zipkin lens UI bugs / improvements


drolando:
These issues / improvements are not worth of separate tickets so I'm filing them together

- [x] search page result rows and traces are mismatched: https://www.dropbox.com/s/60s8vmz4ogmbt7g/Screenshot%202019-01-16%2014.13.56.png As you can see in the screenshot the result row says there are a lot of services and spans, but the trace viewer shows very few. I think the result lines and the trace data are simply mismatched.
- [x] opinionated change: it'd be nice if the search button was on the right or at least it was it bigger to make it obvious you need to click that
- [x] there's no favicon
- [ ] it's be nice if the search dropdowns had tooltips on what should go in them, mostly the annotation query. Like some examples of what people can do with it
- [x] I'd also remove the default "error" annotation query since we don't default any other field
- [x] remove the `-` from the custom time picker since it makes the boxes unaligned: https://www.dropbox.com/s/tzu82pyczl5lzya/Screenshot%202019-01-16%2016.25.24.png
- [x] the `+` and `-` symbols from collapsing / expanding subtrees are inverted. I'd expect pressing `+` to expand, not to collapse the tree
- [x] leaf spans should use remoteEndpoint service name rather than the localEndpoint. If the SERVER span exists, the server's localEndpoint should be used (same as now).
- [x] it'd be nice if pressing enter executed the search
- [x] the services list is not alphabetically ordered (span names are sorted instead)
- [x] in the search results there should be a space between the number and "spans"
- [x] since we have microsecond precision we should show also the microseconds when clicking on a span
- [ ] the "serviceName" column is larger than it needs to (larger than the longer service name), the spanName column instead is super narrow
- [ ] when I try adding a tags filter the ui gets very slow. The characters that I type appear very slowly, as if the JS is trying to do something with each of them. Querying the autocomplete api maybe? But I never enabled it.

tacigar:
@drolando Thank you!!!

tacigar:
* the services list is not alphabetically ordered (span names are sorted instead)
*  in the search results there should be a space between the number and "spans"

resolved  by #2434

codefromthecrypt:
@drolando I missed this list. thank you so much!

jorgheymans:
- reinstate the 'x minutes ago' feature when listing traces.

PR #2362, opened 2019-01-24
Change timeline theme of zipkin-lens

tacigar:
Related #2321 

**Previous**:

previous

**Current**:

current

---

In the future I am planning to use SVG for Timeline.
But it will take some time.
So I changed only the color of the current Timeline first.


tacigar:
And I fixed 2 problems of #2351 in this PR.

* search page result rows and traces are mismatched: https://www.dropbox.com/s/60s8vmz4ogmbt7g/Screenshot%202019-01-16%2014.13.56.png As you can see in the screenshot the result row says there are a lot of services and spans, but the trace viewer shows very few. I think the result lines and the trace data are simply mismatched.
* the + and - symbols from collapsing / expanding subtrees are inverted. I'd expect pressing + to expand, not to collapse the tree

Thank you @drolando !!

Issue #2352, opened 2019-01-16
Support search/open a trace by id in lens


drolando:
In the old UI there's a textbox where you can enter the `traceId` and open directly that trace. We should add it also to zipkin-lens

PR #2368, opened 2019-01-29
Support search/open a trace by traceId in lens

tacigar:
Related: #2352

### DIFF
--- a/zipkin-lens/src/components/GlobalSearch/ConditionAnnotationQuery.js
+++ b/zipkin-lens/src/components/GlobalSearch/ConditionAnnotationQuery.js
@@ -19,8 +19,8 @@ class ConditionAnnotationQuery extends React.Component {
     this.handleKeyDown = this.handleKeyDown.bind(this);
   }
 
-  handleKeyDown(e) {
-    if (e.keyCode === 13) {
+  handleKeyDown(event) {
+    if (event.key === 'Enter') {
       this.inputRef.blur();
     }
   }
--- a/zipkin-lens/src/components/GlobalSearch/ConditionTraceId.js
+++ b/zipkin-lens/src/components/GlobalSearch/ConditionTraceId.js
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 
-const maxCharacters = 20;
+const maxCharacters = 16;
 
 const propTypes = {
   value: PropTypes.string.isRequired,
@@ -12,7 +12,7 @@ const propTypes = {
   isFocused: PropTypes.bool.isRequired,
 };
 
-class ConditionAnnotationQuery extends React.Component {
+class ConditionTraceId extends React.Component {
   constructor(props) {
     super(props);
     this.inputRef = undefined;
@@ -36,7 +36,7 @@ class ConditionAnnotationQuery extends React.Component {
     } = this.props;
 
     return (
-      <div className="condition-annotation-query">
+      <div className="condition-trace-id">
         <input
           ref={(ref) => {
             setNextFocusRef(ref);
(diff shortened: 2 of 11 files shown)


tacigar:
I used GlobalSearch when implementing this feature.
The user can specify traceId in the same way as other search conditions.
If traceId is specified when the search button is pressed, jump to tracePage without searching even if other search conditions are specified.

zipkin-lens/src/components/GlobalSearch/ConditionAnnotationQuery.js
```diff
@@ -12,8 +12,6 @@ const propTypes = {
   isFocused: PropTypes.bool.isRequired,
 };
 
-// TODO: After specific-tags feature is completed in server-side,
```
codefromthecrypt, line 15: :D

tacigar: :)

tacigar:
traceid

codefromthecrypt:
haven't installed locally yet. is there a way to have this work with hyperlink if not already? I mean a link with a trace ID embedded shortcut to the trace view screen?

codefromthecrypt:
so navigating from the search bar is a go!

http://localhost:9411/zipkin/trace/5c50102aa5ef50d0ba078c79dcd66bef doesn't work because we haven't re-routed "zipkin/trace/{traceid}" back to the index page I think.. looking into it

codefromthecrypt:
looks like the old route was like this: /zipkin/traces/{id} not /zipkin/trace/{id}

I think we should follow this for now because it makes lens drop-in replacement of the old one. Later we can consider what @llinder proposed earlier which was hash routing

codefromthecrypt:
fyi https://github.com/openzipkin/zipkin/blob/master/zipkin-autoconfigure/ui/src/main/java/zipkin2/autoconfigure/ui/ZipkinUiAutoConfiguration.java#L134

tacigar:
@adriancole 
OK, thank you!!
I'll change them in this PR now.

tacigar:
Fixed

tacigar:
> haven't installed locally yet. is there a way to have this work with hyperlink if not already? I mean a link with a trace ID embedded shortcut to the trace view screen?

Yes, there is.

You mean this link, right?
link

codefromthecrypt:
this question was about people who use bookmarks. you already addressed
this with the /zipkin/traces change.

basically we want existing bookmarks to work at least for a while I think.

codefromthecrypt (approved):
thanks it works

zipkin-lens/src/components/GlobalSearch/ConditionTraceId.js
```diff
@@ -0,0 +1,65 @@
...
+
+class ConditionTraceId extends React.Component {
+  constructor(props) {
+    super(props);
+    this.inputRef = undefined;
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+  }
+
+  handleKeyDown(e) {
+    if (e.keyCode === 13) {
```
zeagord: KeyCode is deprecated. We must use the `e.key` instead. More better to follow the instruction here everywhere we handle keyboard events. 
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode#Example

zeagord (changes requested):

zipkin-lens/src/components/GlobalSearch/ConditionTraceId.js
zeagord: LGTM other than that. Cheers.

tacigar: Thank you for information 🙇 
I only use `event.key` for now.

tacigar:
> this question was about people who use bookmarks. you already addressed
this with the /zipkin/traces change. basically we want existing bookmarks to work at least for a while I think.

Ah, I got it.
Thank you~

codefromthecrypt:
good work folks

PR #2372, opened 2019-01-31
Fix autocompleteTags bugs for Elasticsearch

tacigar:
If the type of tagValue is `text`, it will be splitted with some characters (hyphen, space, etc...) when aggregating.
So I use `keyword` type for tagValue.

---

## Example

If I post the following trace data,

```
[{
  ...
  "tags":{"instanceId":"hello-zipkin-world"}
  ...
}],
```

* Previous response for `api/v2/autocompleteValues?key=instanceId`

```
["hello","world","zipkin"]
```

* Current(This PR) response for `api/v2/autocompleteValues?key=instanceId`

```
["hello-zipkin-world"]
```

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchAutocompleteTags.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchAutocompleteTags.java
@@ -57,7 +57,7 @@ final class ElasticsearchAutocompleteTags implements AutocompleteTags {
     if (indices.isEmpty()) return Call.emptyList();
 
     SearchRequest.Filters filters =
-      new SearchRequest.Filters().addTerm("tagKey", key.toLowerCase(Locale.ROOT));
+      new SearchRequest.Filters().addTerm("tagKey", key);
 
     SearchRequest request = SearchRequest.create(indices)
       .filters(filters)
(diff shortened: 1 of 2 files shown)


zeagord (approved):

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchAutocompleteTags.java
```diff
@@ -57,7 +57,7 @@
     if (indices.isEmpty()) return Call.emptyList();
 
     SearchRequest.Filters filters =
-      new SearchRequest.Filters().addTerm("tagKey", key.toLowerCase(Locale.ROOT));
```
shakuzen, line 60: Why do we no longer want to normalize (lowercase) the input? Does the change to `keyword` make it a case insensitive search? Or if we want to be case sensitive, we should probably have a test for that.

shakuzen (commented):
Just have a question about what seems to be a change unrelated to the mentioned bug. Also, for anyone else's reference, here is the documentation on the involved things:
* [text datatype](https://www.elastic.co/guide/en/elasticsearch/reference/current/text.html)
* [keyword datatype](https://www.elastic.co/guide/en/elasticsearch/reference/current/text.html)

codefromthecrypt:
I suspect we dont have tests for this concern. I know in service names we
downcase values to prevent accidental conflicts and knowing not all indexes
are case insensitive. that said we do nothing for tags to downcase them. in
such case it is possibly asymmetric to try and downcase here knowing the
input is not specially handled.

iotw I think leaving case alone is the right choice for autocomplete tags
(to ensure it matches data which is not specially handled). we should have
a test to prove it ex that shows case is sensitive in both keys and values.

if we do desire case insensitive auto complete tags likely we would have to
change span ingest to downcase the raw data for reasons like keyword
search.. that would be a bigger change I think.

great point.

huydx:
IMHO I want to preserve the same mapping with span. From user's point of view, not normalize things may lead to less surprise.

codefromthecrypt:
I'll add case sensitivity tests in a separate PR

PR #2375, opened 2019-02-01
Add links of GitHub/Twitter/Gitter to sidebar

tacigar:
This is an experimental PR.
If you don't like this, please close this PR... 🙇 

**Before**:

before

**After**

after

### DIFF
--- a/zipkin-lens/scss/components/_sidebar.scss
+++ b/zipkin-lens/scss/components/_sidebar.scss
@@ -12,8 +12,7 @@
 
 .sidebar__menu {
   & a {
-    color: #dedede;
-  color: $gray-11;
+    color: $gray-11;
     text-decoration: none;
   }
 }
@@ -56,4 +55,23 @@
   & * {
     fill: $gray-11;
   }
+}
+
+.sidebar__other-links {
+  position: absolute;
+  bottom: 24px;
+  width: 100%;
+  font-size: 24px;
+
+  & a {
+    color: $gray-11;
+    text-decoration: none;
+  }
+}
+
+.sidebar__other-link {
+  padding: 10px 0;
+  width: 100%;
+  display: flex;
+  justify-content: center;
 }
--- a/zipkin-lens/src/components/App/Sidebar.js
+++ b/zipkin-lens/src/components/App/Sidebar.js
@@ -58,6 +58,17 @@ class Sidebar extends React.Component {
           {this.renderPageOption('browser')}
           {this.renderPageOption('dependencies')}
         </div>
+        <div className="sidebar__other-links">
+          <a href="https://github.com/openzipkin/zipkin" target="_blank" rel="noopener noreferrer">
+            <div className="sidebar__other-link fab fa-github" />
+          </a>
+          <a href="https://twitter.com/zipkinproject" target="_blank" rel="noopener noreferrer">
+            <div className="sidebar__other-link fab fa-twitter" />
+          </a>
+          <a href="https://gitter.im/openzipkin/zipkin/" target="_blank" rel="noopener noreferrer">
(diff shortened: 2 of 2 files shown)


shakuzen (approved):
Personally, I like it. Zipkin users don't always know it is an open source project with a community, so it is probably nice to lead them those things. What about a link to our site (https://zipkin.io or soon to just be https://zipkin.apache.org)?

tacigar:
@shakuzen 
Thank you for your valuable opinion and suggestion!!
I added a link to `https://zipkin.apache.org/`.

after2

codefromthecrypt:
good idea

jcarres-mdsol:
I'm ambivalent. 
Companies (as ourselves) will deploy this in their internal networks as part of the monitoring infra. 
In our particular case we do have in some homegrown software links to slack rooms and I'm concerned people will confuse those.  If you start seeing people in gitter asking why the User service is slow, you know why 💃

codefromthecrypt:
oh interesting point. I suspect we can revert this if it would cause confusion.

shakuzen:
I'm not sure this change should be any more confusing than Grafana or Prometheus having links to their stuff in their UIs (they do). Arguably this is a little more noticeable than theirs, but not much if you ask me.

codefromthecrypt:
noticed some feedback about the amount of links being perhaps excessive https://twitter.com/Monkeyanator/status/1236069513701085184

If we consider design impact of removing them.. if it weren't for the links, we could have the entire bottom of the layout dedicated to presenting actual stuff..

codefromthecrypt:
also in hind sight I'm not sure how many have come to gitter due to the links in lens.. we have no analytics to suggest if they are helping or not.

jcchavezs:
I would just keep the gitter one with a icon of help or just don't add any
icon at all (just keep the language one).

shakuzen:
I can understand feedback about the number of links or the current design of displaying them. I still would argue at least a link back to the open source project/community should be somewhere, as it is for other open source monitoring tools. Otherwise users within a company may think Zipkin is a company tool (especially if integrated into an internal tool) or the only route to feedback/contribution is the operator within their company, which they may never know who that is. Giving a path directly to us encourages engagement and proper attribution/recognition of the project. I'm no designer but would be happy to review alternate designs.

anuraaga:
My personal opinion is a single link to the home page. Gives a pointer to the open source without being too direct to chat. I think it would be great to have an awesome conversation start because of the gitter link in the UI but I there's at least as good a chance that all parties involved would prefer going through the zipkin operator who will be able to answer e.g., "Why don't traces show up" by first checking if their storage is down.

Of course we could just wait and see if this is actually a problem or not.

anuraaga:
Another thought is keep the help icon defaulted to gitter but add a knob to configure it so users can link to internal jira or something. This would just be a few lines of code but seems pretty useful for operators.

shakuzen:
I think a single link to our website is sufficient as it has links from there to GitHub and Gitter and Twitter as well as an explanation of what Zipkin is - which might be what users are looking for initially. Having an optional additional link that users can configure could be useful, but seems like a separate decision. I don't know of any other monitoring tool that offers such a feature and if they did, I'm sure they wouldn't let it replace links to their own website, and I don't think we should either, again for purposes of attribution/recognition of the open source project. If people really want to replace something like that, they can fork the project like they would have to for other open source monitoring tools. Just my feelings on it.

anuraaga:
> I'm sure they wouldn't let it replace links to their own website, and I don't think we should either

Right sorry for confusion meant the configurable help link could be in addition to the website link which would be static.

For reference, Grafana allows adding arbitrary links in various places including top level, and I've used it to tie together internal tools a little. We shouldn't go that far though :)

PR #2377, opened 2019-02-01
Fixes regression in cassandra autocomplete tags

codefromthecrypt:
We didn't defensively check if autocomplete tags were in use. This means
people needed to upgrade their schema just to continue working. This
puts defense around autocomplete tags and avoids using the tables unless
autocomplete is actually enabled.


codefromthecrypt:
before this, if `ENSURE_SCHEMA=false`, the using span consumer would crash like this:

```
com.datastax.driver.core.exceptions.InvalidQueryException: unconfigured table autocomplete_tags

	at com.datastax.driver.core.exceptions.InvalidQueryException.copy(InvalidQueryException.java:49)
	at com.datastax.driver.core.DriverThrowables.propagateCause(DriverThrowables.java:35)
	at com.datastax.driver.core.AbstractSession.prepare(AbstractSession.java:97)
	at zipkin2.storage.cassandra.v1.InsertAutocompleteValue$Factory.<init>(InsertAutocompleteValue.java:42)
	at zipkin2.storage.cassandra.v1.CassandraSpanConsumer.<init>(CassandraSpanConsumer.java:52)
	at zipkin2.storage.cassandra.v1.CassandraStorage.spanConsumer(CassandraStorage.java:358)
```

PR #2378, opened 2019-02-01
WIP address people sending lookback zero

### LINKED ISSUE #2356: Use es to store data, hit  /dependencies  api directly and set lookback to 0, then it returns data.
## while i use /dependencies api directly and set lookback to zero, i hope it return empty list. but it still returns data.  Using es to store data.
image

```
IndexNameFormatter.java
public List<String> formatTypeAndRange(@Nullable String type, long beginMillis, long endMillis) {
    GregorianCalendar current = midnightUTC(beginMillis);
    GregorianCalendar end = midnightUTC(endMillis);
    if (current.equals(end)) {
      return Collections.singletonList(formatTypeAndTimestamp(type, current.getTimeInMillis()));
    }
……
```

codefromthecrypt:
Lookback zero is invalid and should return empty.

Fixes #2356

### DIFF
--- a/zipkin/src/test/java/zipkin2/storage/ITDependencies.java
+++ b/zipkin/src/test/java/zipkin2/storage/ITDependencies.java
@@ -125,6 +125,14 @@ public abstract class ITDependencies {
     );
   }
 
+  @Test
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
obviated by #2399

PR #2381, opened 2019-02-04
Rename annotationQuery to tags

tacigar:
`annotationQuery` is difficult to understand...

@adriancole @llinder


codefromthecrypt (approved):

drolando:
I think this is pretty confusing right now. Maybe if you have autocomplete tags enable it looks better but without them it just seems broken.

When I select tags it just gives me an empty textbox with no instruction on what to do. At least in the old ui we have the example query.

codefromthecrypt:
any suggestions? maybe use OSX preview to annotate a screenshot with
something more intuitive?

PR #2383, opened 2019-02-07
Use the remote service name instead of the local service name in Lens

tacigar:
Synchronized zipkin-lens with zipkin-ui.

Related: #2345


tacigar:
Related #2351 

> leaf spans should use remoteEndpoint service name rather than the localEndpoint. If the SERVER span exists, the server's localEndpoint should be used (same as now).

codefromthecrypt:
consider the merge an approval :)

thanks!

Issue #2365, opened 2019-01-24
[zipkin-lens] Show both client and server timings in rpc spans


drolando:
This is something we talked about a while ago, but decided that it was too hard to do until we moved to SVG rendering. Filing here for reference so we don't forget.

Right now the blue bar for RPC spans shows client timings. It starts at the client ts and ends after the client duration. To check how much the network delay was and how long the server actually took you have to click on it and manually check the timestamps.

It'd be nice if that bar was actually composed of 2 different bars: a darker blue one that still represents client timings like now and an overlapping lighter bar that shows server timings. That way you can see the network delay just by looking at the trace view.

I don't know how hard it is to implement in CSS though... @tacigar @zeagord

codefromthecrypt:
https://github.com/openzipkin/zipkin/pull/2386 fixes this.. note this was one popular part of the last UI attempt https://github.com/openzipkin-attic/zipkin-ui

PR #2386, opened 2019-02-12
Show both client and server timings

tacigar:
Related #2365 

### How to implement

If the span's annotation contains `Client Start` and `Client Finish` and `Server Start` and `Server Finish`, use them for rendering the span bar instead of `span.left` and `span.width`. 

---

### Example1

**Before**
before1
**After**
after1

### Example2

**Before**
before2
**After**
after2

cc @drolando

### DIFF
--- a/zipkin-lens/src/components/Timeline/index.js
+++ b/zipkin-lens/src/components/Timeline/index.js
@@ -135,6 +135,7 @@ class Timeline extends React.Component {
                   startTs={startTs}
                   endTs={endTs}
                   traceDuration={traceSummary.duration}
+                  traceTimestamp={traceSummary.spans[0].timestamp}
                   numTimeMarkers={DEFAULT_NUM_TICKS}
                   serviceNameColumnWidth={serviceNameColumnWidth}
                   spanNameColumnWidth={spanNameColumnWidth}
--- a/zipkin-lens/scss/components/_timeline.scss
+++ b/zipkin-lens/scss/components/_timeline.scss
@@ -149,24 +149,36 @@
   position: relative;
 }
 
+.timeline__double-span-bar {
+  width: 100%;
+  height: 100%;
+  position: absolute;
+  display: flex;
+  align-items: center;
+}
+
 .timeline__span-bar {
   position: absolute;
   border-radius: 4px;
-  opacity: 0.8;
-  height: 80%;
+  height: 70%;
+
+  &--client {
+    background-color: $gray-9;
+  }
+
+  &--server {}
 }
 
 .timeline__span-duration {
   color: $dark-4;
   font-size: 12px;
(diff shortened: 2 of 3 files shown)


codefromthecrypt:
looks good. maybe in a test show with good and bad data? ex missing client receive or server send (one-way)

drolando:
This looks amazing! Thanks!

tacigar:
These are [netflix.json](https://github.com/openzipkin/zipkin/blob/master/zipkin-ui/testdata/netflix.json) and [ascend.json](https://github.com/openzipkin/zipkin/blob/master/zipkin-ui/testdata/ascend.json).

Maybe ascend.json is a bit strange.

It has following span data.

```
    ...,
    {
        ...,
        "id": "52b1ab4956917c39",
        "kind": "CLIENT",
        "timestamp": 1531282088533834,
        "duration": 14077,
        ...,
    },
    {
        "id": "52b1ab4956917c39",
        "kind": "SERVER",
        "name": "http:/parent/auth-service/v1/authentications/",
        "timestamp": 1531282088553000,
        "duration": 4157,
        "localEndpoint": {
            "serviceName": "auth-service",
            ...,
        },
        ...,
        "shared": true
    },
    ...
```

`Client` → `...33834` ~ `...47911`
`Server` → `...53000` ~ `...57157`

zipkin-lens/src/components/Timeline/index.js
```diff
@@ -135,6 +135,7 @@ class Timeline extends React.Component {
                   startTs={startTs}
                   endTs={endTs}
                   traceDuration={traceSummary.duration}
+                  traceTimestamp={traceSummary.spans[0].timestamp}
```
codefromthecrypt, line 138: is this different than traceSummary.timestamp?

codefromthecrypt:
one other thing to test.. is to make a really really short client/server span.. like one that completes in less than a millisecond, whereas other spans have a different scale like more than a millisecond. This might show some rendering issues

zipkin-lens/src/components/Timeline/index.js
tacigar: `traceSummary` does not have `timestamp` field now...

codefromthecrypt: heh ok!

codefromthecrypt:
lovely

tacigar:
> one other thing to test.. is to make a really really short client/server span.. like one that completes in less than a millisecond, whereas other spans have a different scale like more than a millisecond. This might show some rendering issues

I tried.
That's the third span.



**Zoom up**



@adriancole 
Sorry for late 🙇

codefromthecrypt:
looks good. not late :) I figured there would be no problem

PR #2387, opened 2019-02-12
Fix a bug of custom lookback in Safari

tacigar:
I tried using lens in safari.
Then I found an UI bug.

**Before**

before-custom

**After**

after-custom


codefromthecrypt:
thanks again!

PR #2385, opened 2019-02-11
Sorts children by timestamp order in the UI

codefromthecrypt:
This prevents rows from looking out-of-place.

Fixes #2370

Using this json:
```json
[
  {
    "traceId": "a52abd6393af6ae8",
    "parentId": "a52abd6393af6ae8",
    "id": "82dd139ac538faeb",
    "kind": "CLIENT",
    "name": "iuserservice/getcurrentloginuser",
    "timestamp": 1548829911741874,
    "duration": 3123,
    "localEndpoint": {
      "serviceName": "hapi-demo",
      "ipv4": "192.168.3.116"
    },
    "remoteEndpoint": {
      "ipv4": "192.168.3.36",
      "port": 9046
    },
    "tags": {
      "__appCode": "100",
      "__appId": "0010100001",
      "__appName": "hapi-demo",
      "__clientId": "2da2420c-c836-41e8-b455-4b8358fc87b6",
      "__orgName": "ARC-Framework",
      "__ownerName": "liangdongcheng",
      "__requestId": "22838be6-1959-4614-b25d-04f9ddd1a6f7",
      "__serviceId": "0010100001",
      "__serviceName": "hapi-demo"
    }
  },
  {
    "traceId": "a52abd6393af6ae8",
    "parentId": "a52abd6393af6ae8",
    "id": "7dd1a7019c810b64",
    "kind": "CLIENT",
    "name": "idemoservice/sayhello",
    "timestamp": 1548829911747695,
    "duration": 6586,
    "localEndpoint": {
      "serviceName": "hapi-demo",
      "ipv4": "192.168.3.116"
    },
    "remoteEndpoint": {
      "ipv4": "192.168.3.147",
      "port": 9999
    },
    "tags": {
      "__appCode": "100",
      "__appId": "0010100001",
      "__appName": "hapi-demo",
      "__clientId": "2da2420c-c836-41e8-b455-4b8358fc87b6",
      "__orgName": "ARC-Framework",
      "__ownerName": "liangdongcheng",
      "__requestId": "22838be6-1959-4614-b25d-04f9ddd1a6f7",
      "__serviceId": "0010100001",
      "__serviceName": "hapi-demo"
    }
  },
  {
    "traceId": "a52abd6393af6ae8",
    "parentId": "a52abd6393af6ae8",
    "id": "7dd1a7019c810b64",
    "kind": "SERVER",
    "name": "idemoservice/sayhello",
    "timestamp": 1548829658845021,
    "duration": 3711,
[...]

PR #2391, opened 2019-02-13
Fixes sort order in java impl of SpanNode

codefromthecrypt:
Follow-up to #2385 cc @drolando

### DIFF
--- a/zipkin-ui/test/component_data/spanNode.test.js
+++ b/zipkin-ui/test/component_data/spanNode.test.js
@@ -200,4 +200,27 @@ describe('SpanNodeBuilder', () => {
     expect(root.children.map(n => n.span))
       .to.eql([trace[3], trace[2], trace[1]]); // null first
   });
+
+  it('should order children by timestamp when IPs change ', () => {
+    const trace = [
+      {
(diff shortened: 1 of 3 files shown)


zipkin-ui/test/component_data/spanNode.test.js
```diff
@@ -200,4 +200,27 @@ describe('SpanNodeBuilder', () => {
     expect(root.children.map(n => n.span))
       .to.eql([trace[3], trace[2], trace[1]]); // null first
   });
+
+  it('should order children by timestamp when IPs change ', () => {
```
codefromthecrypt, line 204: interestingly this passes which means I need to keep looking at that test trace! Maybe I missed the real culprit

drolando:
Yeah this commit still doesn't fix my traces.

On the other hand it the issue is simply that the ips don't match I can probably fix that, but it'd still be nice if the code didn't rely on ip values.

Why does it btw? You already have the parentId and spanId, why do you care about the ip?

codefromthecrypt:
This is a work in progress to investigate what is actually interfering with the sort order. I didn't actually change the javascript code for this test to pass, which means I got something wrong.

PR #2395, opened 2019-02-16
Fixes UI ordering of subtrees

codefromthecrypt:
Before, order was inconsistent for subtrees with regards to timestamps.
Particularly, when a shared span was present, and endpoint details were
mixed between local spans, order could end up backwards. This fixes it.

Below is a real trace from Yelp, zoomed out to the point where you can
see the problem, but not the actual details.

Before
screen shot 2019-02-16 at 6 21 09 pm
screen shot 2019-02-16 at 6 20 56 pm

After
screen shot 2019-02-16 at 6 20 11 pm
screen shot 2019-02-16 at 6 20 38 pm

See #2385

### DIFF
--- a/zipkin-lens/src/zipkin/span-node.test.js
+++ b/zipkin-lens/src/zipkin/span-node.test.js
@@ -213,4 +213,39 @@ describe('SpanNodeBuilder', () => {
     expect(root.children.map(n => n.span))
       .toEqual([trace[3], trace[2], trace[1]]); // null first
   });
+
+  it('should order children by timestamp when IPs change ', () => {
+    const trace = [
+      {
+        traceId: '1',
+        parentId: 'a',
+        id: 'c',
+        kind: 'SERVER',
+        shared: true,
+        timestamp: 1,
+        localEndpoint: { serviceName: 'my-service', ipv4: '10.2.3.4' },
+      },
+      {
+        traceId: '1',
+        parentId: 'c',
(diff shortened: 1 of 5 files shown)


zipkin-lens/src/zipkin/span-node.test.js
```diff
@@ -213,4 +213,39 @@ describe('SpanNodeBuilder', () => {
     expect(root.children.map(n => n.span))
       .toEqual([trace[3], trace[2], trace[1]]); // null first
   });
+
+  it('should order children by timestamp when IPs change ', () => {
```
codefromthecrypt, line 217: nevermind this.. it is just a backport of another test. Eventhough there is some labor, I wanted to ensure tests are the same across zipkin-ui and zipkin-lens until we complete

zipkin-lens/src/zipkin/trace.js
```diff
@@ -266,22 +271,21 @@ export function detailedTraceSummary(root, logsUrl) {
...
-        toPrefix.push(child);
-        childIds.push(child.span.id);
+        children.push(child);
       }
     });
-    queue = toPrefix.concat(queue);
+
+    // Pulling up children may affect our sort order. We re-sort to ensure rows are added in
+    // timestamp order.
+    children.sort(nodeByTimestamp);
```
codefromthecrypt, line 286: this is the part that fixes it

PR #2397, opened 2019-02-17
Shows different tag values when client and server differ

### LINKED ISSUE #2390: If client and server side has a same tag name, should it be display server side tag value ?
Client side (hapi-demo) and server side (soa-demo-server) has same tag names '__serviceId & __serviceName', but tag values is not same. tag value is '0010100001/hapi-demo' and '0010100003/soa-demo-service', and now display client tag values, shoulde it display server side tag value ?

Trace JSON: 

[trace.zip](https://github.com/openzipkin/zipkin/files/2858841/trace.zip)

image

codefromthecrypt:
This does the minimal needed to show that there are different values for
the same tag, between a client and a server. Future work could be
considered to answer who sent what value. In the mean time, users can
inspect the JSON button when this occurs.

Fixes #2390

## Before: different values were swallowed

screen shot 2019-02-17 at 11 31 02 am
screen shot 2019-02-17 at 11 31 15 am

## After: different values are displayed

screen shot 2019-02-17 at 11 27 56 am
screen shot 2019-02-17 at 11 27 42 am

### DIFF
--- a/zipkin-lens/src/zipkin/span-row.js
+++ b/zipkin-lens/src/zipkin/span-row.js
@@ -196,7 +196,7 @@ function parseTagRows(span) {
         key: ConstantNames[key] || key,
         value: span.tags[key],
       };
-      if (localFormatted) tagRow.endpoint = localFormatted;
+      if (localFormatted) tagRow.endpoints = [localFormatted];
       tagRows.push(tagRow);
     });
   }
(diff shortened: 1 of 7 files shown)


tacigar:
LGTM

tacigar (approved):

codefromthecrypt:
thanks for taking a look!

ldcsaa:
If show like these would be better ?

http.path (C [or Client])    /content/persional
http.path (S [ or Server])    /content/persionals

codefromthecrypt:
@ldcsaa how to design UX for this is a separate issue as there are many ways to do that. want to open one? It will take some time to get feedback on it, which is why I solved the underlying problem seperately

ldcsaa:
OK~ 
I create a new issue [#2403](https://github.com/openzipkin/zipkin/issues/2403)

PR #2404, opened 2019-02-19
Suggests use of Zipkin Lens when running a standard server

### LINKED ISSUE #2402: Allow switching between Zipkin Lens and prior UI during transition phase
**Feature:**
Allow Zipkin end-users to switch between Zipkin Lens and the historic UI.

**Rational**
We are transitioning between the two UIs and it would be a nicer user experience if individual users can switch between the two. Currently, only operators can decide which UI is used for all users of that deployment at startup. If users can switch, this gives them the choice of which UI to use when, and notably allows early adopters/testers of Zipkin Lens, while providing a fallback between when any issue is found and when a release is made with it fixed in Zipkin Lens.

**Example Scenario**
Power users want to try the latest and greatest UI for Zipkin without forcing all users to try it at once or all the time.

**Prior Art**
@adriancole may have found a way to hack this together ⛏️

codefromthecrypt:
This adds a "Try Lens UI" button to the classic UI when running in the
standard build (which can server both asset trees at once). This is
controlled by `SUGGEST_LENS` env variable set to true by default. Those
simply hosting zipkin-ui assets will not be impacted.

This works via a normal browser cookie. If someone wants to stop using
lens, then close their browser and reload the page.

switcheroo

This button can be disabled by setting `ZIPKIN_UI_SUGGEST_LENS=false`

Thanks for tips by @kaiyzen @bsideup. While the approach here is dirty,
one nice thing is that it leaves no permanent trace in published assets
etc. All the controls can be removed with no change to users in the next
month or two when transition completes.

cc @openzipkin/ui 

Fixes #2402


shakuzen:
Awesome work, Adrian! I think this will really help the transition to Lens.
One piece of feedback is that it doesn't seem obvious to a user how to go back to the classic UI right now. What do you think about also adding a button to Lens that's the complement of the current button? "Go back to classic UI" or something.

codefromthecrypt:
> One piece of feedback is that it doesn't seem obvious to a user how to go
> back to the classic UI right now. What do you think about also adding a
> button to Lens that's the complement of the current button? "Go back to
> classic UI" or something.
>
I think lens would be cluttered as the sidebar has already a lot of things
in it. I would prefer to just let someone close their browser as this is
really a short term thing..

drolando:
Some people don't really want to close their browsers to not lose all their tabs.

Also this is not intuitive at all and we'd get a bunch of questions on how to go back and forth

codefromthecrypt:
to set expectations.. lens has been in progress for months. the only reason
we are doing this at all is to remove FUD about pulling the trigger and
actually releasing lens as default.

Meanwhile, people are still looking at zipkin as if lens doesnt exist.

The primary motive of this wasnt to create a flip flop for an extended
beta, but rather the minimum infra to reduce the burden of maintaining 2
uis simultaneously.. and dispelling myth that lens doesnt exist.

if we add an icon on lens to go back, people may have impression that we
will be keeping the old ui which we certainly wont. it will be deleted
ideally within a month

make sense?

codefromthecrypt:
ok so let me ask another way then.

of people offering for us to do more work.. who is offering to do it? and
revert it back later?

drolando:
I'll look into it tomorrow. Isn't it enough to set the cookie to false and refresh the page?

codefromthecrypt:
@drolando yes, but you will still need a control for that somewhere. I didn't see an obvious place that wouldn't clutter the new UI which without this revision request would be lacking said clutter.

drolando:
I'm fine with merging as is and add the other button in another pr

codefromthecrypt:
@drolando thanks for stepping up. I'll merge this, then as that will let folks building from master or using snapshots give it a try.

PR #2407, opened 2019-02-21
Fix css bugs of limit condition and lookback condition components in zipkin lens

tacigar:
I found some css bugs in limit condition and lookback condition components.

**BEFORE**

before-1

before-2

**AFTER**

after-1

after-2


shakuzen (approved):

codefromthecrypt:
thanks again

PR #2409, opened 2019-02-22
Refactor prop-types

tacigar:
* Collect the definitions of proptypes in various files into one file and use them.
* Fix type bug of limit condition component.

### DIFF
--- a/zipkin-lens/src/prop-types/index.js
+++ b/zipkin-lens/src/prop-types/index.js
@@ -1,0 +1,77 @@
+import PropTypes from 'prop-types';
+
+export const spanTagPropTypes = PropTypes.shape({
+  key: PropTypes.string.isRequired,
+  value: PropTypes.string.isRequired,
+});
+
+export const spanTagsPropTypes = PropTypes.arrayOf(
+  spanTagPropTypes,
+);
+
+export const spanAnnotationPropTypes = PropTypes.shape({
+  value: PropTypes.string.isRequired,
+  timestamp: PropTypes.number.isRequired,
+  endpoint: PropTypes.string.isRequired,
+  relativeTime: PropTypes.string.isRequired,
+});
+
+export const spanAnnotationsPropTypes = PropTypes.arrayOf(
+  spanAnnotationPropTypes,
+);
+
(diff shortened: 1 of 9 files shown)


zipkin-lens/src/prop-types/index.js
```diff
@@ -0,0 +1,77 @@
+import PropTypes from 'prop-types';
```
tacigar, line 1: This file has all prop-type definitions of spans and traces.

zipkin-lens/src/prop-types/index.js
```diff
@@ -0,0 +1,77 @@
...
+  timestamp: PropTypes.number.isRequired,
+  endpoint: PropTypes.string.isRequired,
+  relativeTime: PropTypes.string.isRequired,
+});
+
+export const spanAnnotationsPropTypes = PropTypes.arrayOf(
+  spanAnnotationPropTypes,
+);
+
+export const detailedSpanPropTypes = PropTypes.shape({
```
codefromthecrypt, line 25: maybe add a TODO as while I think your assessment is correct based on span row conversion, the underlying data only requires trace and span ID. We need to revisit things like this once UX is working as desired https://github.com/openzipkin/zipkin-api/blob/master/zipkin2-api.yaml

codefromthecrypt (approved):
I like the centralization

zeagord (approved):

zipkin-lens/src/prop-types/index.js
tacigar: @adriancole 
Sorry, I don't understand what you mean 😭 
What should I do?

codefromthecrypt: Add above this line

```
// TODO: Verify which fields we should enforce here, as some are optional per https://github.com/openzipkin/zipkin-api/blob/master/zipkin2-api.yaml
```

tacigar: OK
Thank you!

tacigar:
PTAL 🙇

codefromthecrypt (approved):

codefromthecrypt:
just awaiting travis to finish

PR #2413, opened 2019-02-25
Refactor Timeline component

tacigar:
The codes of Timeline component are dirty.
So I refactored them.

* Use BEM notation for CSS
* Rename some variables


codefromthecrypt (approved):

codefromthecrypt:
thanks for the refactoring!

PR #2415, opened 2019-02-26
Works around encoding bug in zipkin-ruby by leniently parsing

### LINKED ISSUE #2414: possible bug in zipkin-ruby or in the visualization
After updating to 2.12.2 we have seen that many of our traces get smashed to the left of the graph
screen shot 2019-02-25 at 11 41 28 am

When clicking on one of those I get:

screen shot 2019-02-25 at 11 42 36 am

Only SA information.
Looking at the json, these could be the relevant sections:
```
  {
    "traceId": "7309633b16097a5c",
    "parentId": "7309633b16097a5c",
    "id": "4f829b6d0eb7bfb5",
    "name": "post",
    "localEndpoint": {
      "serviceName": "hermes-sandbox",
      "ipv4": "54.162.48.195",
      "port": 443
    },
    "tags": {
      "sa": "1"
    }
  },
```

```
  {
    "traceId": "7309633b16097a5c",
    "parentId": "7309633b16097a5c",
    "id": "4f829b6d0eb7bfb5",
    "kind": "SERVER",
    "name": "post /path_info_in_here_edited_out",
    "timestamp": 1551123015539229,
    "duration": 999430,
    "localEndpoint": {
      "serviceName": "hermes-sandbox",
      "ipv4": "172.17.0.20",
      "port": 3000
    },
    "tags": {
      "http.path": "/apps/exd/api/mauth_proxy"
    }
  },
```

Note how the SA notation is alone in a different section of the JSON even when they have the same ID.

I assume the UI is rendering any one of the other depending on the orther it finds this information. Thus it gets an annotation with no further info and gets rendered this way.

These traces are being sent by the zipkin-ruby tracer. I suspect a bug there,  but please verify.
Zipkin-ruby does *not* have support for v2 traces yet which can be the root of this.

If you determine this is a bug in the current UI, I do not care really as long as Lens do not have the same bug as we will migrate as soon as it is available in the stable image.

codefromthecrypt:
In zipkin-ruby (zipkin-tracer up to 0.32.3), the "sa" annotation was
incorrectly encoded as "1" not `true`. This likely was due to the fact
that thrift encodes boolean as 1. This caused the remote endpoint to be
mistaken for a local endpoint as the parser thought "sa" was a normal
string tag with the local endpoint attached. The impact was
mis-rendering and also service graph aggregation would likely be
incorrect in some cases as well.

Definitely, zipkin-ruby should update as even if we leniently parse,
clones will not likely be as tolerant and this could hurt things.
However, it is best to be lenient as it is cheap and low impact. There
is unlikely any valid tag named "sa" "ca" or "sa" used for a purpose
besides the remote endpoint.

Fixes #2414

### DIFF
--- a/zipkin-storage/mysql-v1/src/test/java/zipkin2/storage/mysql/v1/LazyMySQLStorage.java
+++ b/zipkin-storage/mysql-v1/src/test/java/zipkin2/storage/mysql/v1/LazyMySQLStorage.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2018 The OpenZipkin Authors
+ * Copyright 2015-2019 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -37,9 +37,9 @@ public class LazyMySQLStorage implements TestRule {
     if (storage != null) return storage;
 
     try {
-      container = new ZipkinMySQLContainer(version);
-      container.start();
-      System.out.println("Will use TestContainers MySQL instance");
+      //container = new ZipkinMySQLContainer(version);
+      //container.start();
+      //System.out.println("Will use TestContainers MySQL instance");
     } catch (Exception e) {
       // Ignored
     }
--- a/zipkin-storage/mysql-v1/src/test/java/zipkin2/storage/mysql/v1/LazyMySQLStorage.java
+++ b/zipkin-storage/mysql-v1/src/test/java/zipkin2/storage/mysql/v1/LazyMySQLStorage.java
@@ -37,9 +37,9 @@ public class LazyMySQLStorage implements TestRule {
     if (storage != null) return storage;
 
     try {
-      //container = new ZipkinMySQLContainer(version);
-      //container.start();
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
I tried this and it worked. Note that since conversion happens at ingress, it will only help with new spans.

Normal:
screen shot 2019-02-26 at 1 06 13 pm

Lens:
screen shot 2019-02-26 at 1 06 43 pm

zipkin-storage/mysql-v1/src/test/java/zipkin2/storage/mysql/v1/LazyMySQLStorage.java
```diff
@@ -37,9 +37,9 @@ MySQLStorage get() {
     if (storage != null) return storage;
 
     try {
-      container = new ZipkinMySQLContainer(version);
-      container.start();
-      System.out.println("Will use TestContainers MySQL instance");
+      //container = new ZipkinMySQLContainer(version);
```
codefromthecrypt: haha whoops

shakuzen (approved):
LGTM

jcarres-mdsol:
Fixed by https://github.com/openzipkin/zipkin-ruby/pull/139

PR #2419, opened 2019-02-27
Use HTTPS for downloading/uploading artifacts

shakuzen:
This will help protect Zipkin and its users by ensuring our artifacts and dependencies we use are authentic.


shakuzen:
I opted to do a find-and-replace on `http:` in the package-lock.json files rather than regenerate them because regenerating them changes a number of versions and I don't know if that is okay or not.

codefromthecrypt:
green works for me. Thanks, tommy

PR #2421, opened 2019-03-01
Fails on incorrect format sent to http endpoints

codefromthecrypt:
This gives a message instead of parsing the valid bits of json encoding. Thanks for the nudge by @drolando though I think we've been nudged before, too!

Ex.
```bash
$ curl -X POST -s localhost:9411/api/v1/spans -H'Content-Type: application/json' -d @netflix.json
Expected a JSON_V1 encoded list, but received: JSON_V2
```

```bash
$ curl -X POST -s localhost:9411/api/v1/spans -H'Content-Type: application/json' -d 'adssa'
Expected a JSON_V1 encoded list
```

```bash
$ curl -X POST -s localhost:9411/api/v2/spans -H'Content-Type: application/json' -d '[{"traceId":"49d20272a7a8f41d","id":"49d20272a7a8f41d","name":"post /cxf/catalog","timestamp":1536918355807036,"duration":10341,"annotations":[{"timestamp":1536918355807036,"value":"sr","endpoint":{"serviceName":"catalog-service","ipv4":"192.168.1.113"}},{"timestamp":1536918355817377,"value":"ss","endpoint":{"serviceName":"catalog-service","ipv4":"192.168.1.113"}}],"binaryAnnotations":[{"key":"http.path","value":"/cxf/catalog","endpoint":{"serviceName":"catalog-service","ipv4":"192.168.1.113"}}]}]'
Expected a JSON_V2 encoded list, but received: JSON_V1
```

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
@@ -135,9 +135,8 @@ class ZipkinHttpCollector implements HttpHandler, HandlerWrapper {
         }
       }
 
-      BytesDecoder<Span> detectedDecoder;
       try {
-        detectedDecoder = SpanBytesDecoderDetector.decoderForListMessage(body);
+        SpanBytesDecoderDetector.decoderForListMessage(body);
       } catch (IllegalArgumentException e) {
         metrics.incrementMessagesDropped();
         exchange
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
tests pending

codefromthecrypt:
ok tests are in

drolando (approved):

shakuzen (approved):

zeagord (approved):

codefromthecrypt:
this won't work as there is some overlap when we send data that can match either format.. will adjust

shakuzen (approved):

Issue #2350, opened 2019-01-16
Support upload trace in Zipkin Lens UI


zeagord:
In the legacy UI there is a feature to upload traces and visualise it. We should support the same in the Lens UI

tacigar:
To tell the truth, I am wondering whether this function is necessary...
At least I have never used this...
WDYT? @openzipkin/ui

shakuzen:
Here is the original pull request that added it: https://github.com/openzipkin/zipkin/pull/1884

I definitely used it at my previous job and encouraged others to use it. It is one way of getting around the storage period limitation. That is, for a particular interesting trace, you can save it to your local machine or attach it to an issue (e.g. JIRA) so that others can check that trace at any time easily, without having to worry about the original data expiring from the data store. It's a nicer experience than everyone wanting to do this learning how to POST the data to a locally running instance or something.

codefromthecrypt:
I concur with tommy's summary

tacigar:
@shakuzen @adriancole 
Thank you for your opinions!!!
OK, I see.
I'll implement this feature 🙇

shakuzen:
Resolved via #2447

PR #2424, opened 2019-03-04
Support upload trace in Zipkin Lens UI

tacigar:
#2350 

traceviewer-1
traceviewer-2


zipkin-lens/scss/components/_mini-timeline.scss
```diff
@@ -1,37 +1,37 @@
...
 
-.mini-trace-viewer__time-marker-label {
+.mini-timeline__time-marker-label {
   color: $dark-2;
   font-size: $font-size-xs;
   position: absolute;
   left: -20px;
 
-  &.first {
+  &--first {
```
tacigar, line 22: Use BEM notation

zipkin-lens/scss/components/_mini-timeline.scss
```diff
@@ -1,37 +1,37 @@
-.mini-trace-viewer {
+.mini-timeline {
```
tacigar, line 1: Rename `trace-viewer` -> `timeline`

zipkin-lens/scss/components/_trace-page.scss
```diff
@@ -0,0 +1,3 @@
+.trace-page {
```
tacigar, line 1: Sine there is a possibility of writing later, I made a file

zipkin-lens/src/components/App/index.js
```diff
@@ -4,8 +4,9 @@ import { BrowserRouter, Route } from 'react-router-dom';
 
 import Layout from './Layout';
 import BrowserContainer from '../../containers/Browser/BrowserContainer';
-import DetailedTraceSummaryContainer from '../../containers/DetailedTraceSummary/DetailedTraceSummaryContainer';
+import TracePageContainer from '../../containers/TracePage/TracePageContainer';
```
tacigar, line 7: I refactored to reuse `DetailedTraceSummary`.

* Remove a process of fetching trace data from `DetailedTraceSummary` component.
* Make a new component `TracePage`, and add that process and `DetailedTraceSummary` component to the new component.

zipkin-lens/src/components/DetailedTraceSummary/index.js
```diff
@@ -2,43 +2,24 @@ import PropTypes from 'prop-types';
...
 
 const propTypes = {
-  isLoading: PropTypes.bool.isRequired,
-  traceId: PropTypes.string.isRequired, /* From url parameter */
-  traceSummary: PropTypes.shape({}),
-  fetchTrace: PropTypes.func.isRequired,
+  startTs: PropTypes.number,
+  endTs: PropTypes.number,
+  onStartAndEndTsChange: PropTypes.func.isRequired,
+  traceSummary: detailedTraceSummaryPropTypes.isRequired,
```
tacigar: Strict prop types.

zipkin-lens/src/components/MiniTimeline/index.js
```diff
@@ -3,29 +3,30 @@ import React from 'react';
...
 
-const GRAPH_HEIGHT = 75;
-const NUM_TICKS = 5;
-const LEFT_MOUSE_BUTTON = 0;
+const graphHeight = 75;
+const numTimeMarkers = 5;
+const leftMouseButton = 0;
 
-const renderTickLines = () => {
+const renderTimeMarkers = () => {
```
tacigar, line 19: Rename ticks → timeMarker

zipkin-lens/src/components/Timeline/index.js
```diff
@@ -98,8 +98,8 @@ class Timeline extends React.Component {
           serviceNameColumnWidth={serviceNameColumnWidth}
           spanNameColumnWidth={spanNameColumnWidth}
           numTimeMarkers={defaultNumTimeMarkers}
-          onServiceNameColumnWidthChange={this.handleServiceNameColumnChange}
-          onSpanNameColumnWidthChange={this.handleSpanNameColumnChange}
+          onServiceNameColumnWidthChange={this.handleServiceNameColumnWidthChange}
```
tacigar, line 101: I mistook method name...
These are overlooked bugs...
Maybe current lens does not work correctly... 😭

zipkin-lens/src/components/TraceViewer/index.js
```diff
@@ -0,0 +1,163 @@
...
+          isInvalidFormat: true,
+        });
+      }
+      const traceSummary = detailedTraceSummary(
+        treeCorrectedForClockSkew(rawTrace),
+      );
+
+      this.setState({
+        startTs: 0,
+        endTs: traceSummary.duration,
```
tacigar, line 61: `startTs` and `endTs` must be set when `traceSummary` is changed.
If I don't do so, MiniTimeline's range will be displayed as the previous range...

zipkin-lens/src/components/TraceViewer/index.js
```diff
@@ -0,0 +1,163 @@
...
+          const styles = isDragActive
+            ? { ...dropzoneBaseStyle, dropzoneActiveStyle }
+            : { ...dropzoneBaseStyle };
+          return (
+            <div
+              {...getRootProps()}
+              style={styles}
+            >
+              <input {...getInputProps()} />
+              <p>Choose file</p>
```
shakuzen, line 107: I'm not sure the right verbosity here, but I think we should guide users more. The classic UI has a "Browse" button at least, which I think gives more an indication that clicking it will allow one to select a file from their computer to upload. An improvement on that might be to reword with something like "Upload a file by drag/drop or click to browse".
As an aside, we haven't gotten any reports of confusion, but maybe even more explanation somehow about what kind of file/contents are expected would be good. Anyways, that could be a separate discussion from this pull request.

shakuzen (approved):
I tried it out locally with some of our test data. I left one comment about UI text, but I'm happy to consider it post-merge as well.

jcarres-mdsol:
Does the trace lookup header have any purpose in this screen?

shakuzen:
> Does the trace lookup header have any purpose in this screen?

@jcarres-mdsol do you mean the search bar at the top where you can select criteria to search? If so, I agree it doesn't have a use on this page and could be confusing, but I think it is there for consistency as a global search context. I'm not a designer so not sure what's best to do about it (hide it for this view only? show it as disabled in this view?), but thank you for bringing it up. What do others think? Maybe something to spin off into a separate issue so we don't block adding this functionality that helps bring Lens up to feature parity with the classic UI.

jcarres-mdsol:
If it has no use, then should not be there. 
Another option would be to, have a "search traceid" component instead, which is also a functionality in the classic UI.

That would make sense, this would be the place where you look at one single trace, either by searching by its traceid in the storage or uploading it from outside storage

tacigar:
@jcarres-mdsol Thank you for your feedback.

> Another option would be to, have a "search traceid" component instead, which is also a functionality in the classic UI.

Lens already has that feature.

traceid

I tried implementing the feature to upload JSON in the same way as above.
What do you think about this?

fromjson

cc: @shakuzen

jcarres-mdsol:
It is an intriguing idea, to me it is less discoverable as UI (I did not even see the traceid search option there) but I can see the appeal of a single search bar, something like the browser.

codefromthecrypt:
what I like about this is we can also do a dependency graph of that
one trace selected...

codefromthecrypt:
ohh and the index graph and screen could work equally with multiple
files.. ex json + a trace ID, shows in the index screen, and
dependency graph

codefromthecrypt:
to be clear.. I like the universal search bar approach with a "traceJSON" button

shakuzen:
Personally, I worry that if we put a json option in the universal search dropdown, we're overloading the meaning of options there too much. Currently you select a property of a span (in some sense). Putting "fromJSON" or something similar there feels a bit confusing. If we want the JSON upload to be part of the universal search instead of a dedicated view/tab, maybe we should make a dedicated upload button in the universal search? Though that might give the impression that the uploaded JSON is saved.

codefromthecrypt:
I guess we need to think on this a bit more?

codefromthecrypt:
grab 15m here? https://cwiki.apache.org/confluence/display/ZIPKIN/2019-03-11+Tapping+into+Zipkin+data

codefromthecrypt:
@shakuzen I agree that there could be some confusion about the query aspect of this. Though it does seem to be in some means related to traceId field... that's how it made sense to me. Also, what we are optimizing for.

For example, by putting the concept of locally sourced (which is indeed a bit icky), into the background.. we benefit by retaining the universal search capability which is what I think we want in the foreground.

Maybe we can down-rank the "from json" criteria to the bottom, or possibly quasi hide it as most users won't be using this? For example, if you "expert mode" or click something, then the from json option appears, but remains in a place where it is universally considered.

wdyt?

codefromthecrypt:
added for 9am tokyo time monday as that's the only time reasonable and in the schedule. We can collect local feedback even if it is too early. https://cwiki.apache.org/confluence/display/ZIPKIN/2019-03-11+Tapping+into+Zipkin+data

tacigar:
Personally I agree with @shakuzen...

> Maybe we can down-rank the "from json" criteria to the bottom, or possibly quasi hide it as most users won't be using this? For example, if you "expert mode" or click something, then the from json option appears, but remains in a place where it is universally considered.

umm...
I feel a little complicated.

I think it would be better to add another dropdown menu for traceId and upload JSON like the following.

dropdown

codefromthecrypt:
sounds good to me. this would move trace id out if the universal search
then right? and the action of this form will be like old ui where it takes
you to expanded trace screen?

tacigar:
Yes

tacigar:
I'll close this pr
and make a new pr

PR #2431, opened 2019-03-07
Adds integration test for messaging linking and fixes MySQL impl

codefromthecrypt:
Other storage options were document oriented, so weren't subject to
field constraints. This fixes the fields cherry-picked from MySQL,
which are needed to link messaging spans.

This adds the missing integration test, which only ensures data is
there. The complete logic test is `DependencyLinkerTest`.

Thanks @mbrade for reporting


shakuzen (commented):
post merge, but for the record: looks good to me

PR #2432, opened 2019-03-07
Configure CorsService without catch-all path mapping

hyangtack:
Modifications:
- Specified Spring `@Order` annotation to the `corsConfigurator` in order that the bean is evaluated at the last.
  - Removed a dummy service of the catch-all path mapping with OPTIONS header.
- Added `access-control-request-method: GET` header when sending a preflight request in IT.

### DIFF
--- a/zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin2/autoconfigure/prometheus/ZipkinPrometheusMetricsAutoConfiguration.java
+++ b/zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin2/autoconfigure/prometheus/ZipkinPrometheusMetricsAutoConfiguration.java
@@ -71,7 +71,8 @@ import org.springframework.util.StringUtils;
   @Bean
   @Order(1)
   ArmeriaServerConfigurator notFoundMetricCollector() {
-    return sb -> sb.service(PathMapping.ofCatchAll(), (ctx, req) -> HttpResponse.of(HttpStatus.NOT_FOUND));
+    // Use glob instead of catch-all to avoid adding it to the trie router.
+    return sb -> sb.service(PathMapping.ofGlob("/**"), (ctx, req) -> HttpResponse.of(HttpStatus.NOT_FOUND));
   }
 
   static final class MetricCollectingService<I extends Request, O extends Response>
--- a/zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin2/autoconfigure/prometheus/ZipkinPrometheusMetricsAutoConfiguration.java
+++ b/zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin2/autoconfigure/prometheus/ZipkinPrometheusMetricsAutoConfiguration.java
@@ -20,6 +20,7 @@ import com.linecorp.armeria.common.RequestContext;
 import com.linecorp.armeria.common.Response;
 import com.linecorp.armeria.common.logging.RequestLog;
 import com.linecorp.armeria.common.logging.RequestLogAvailability;
+import com.linecorp.armeria.server.PathMapping;
 import com.linecorp.armeria.server.Service;
 import com.linecorp.armeria.server.ServiceRequestContext;
 import com.linecorp.armeria.server.SimpleDecoratingService;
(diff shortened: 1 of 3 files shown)


hyangtack:
@adriancole @anuraaga This is PR for https://github.com/openzipkin/zipkin/pull/2348#discussion_r261855064. I found that the preflight request didn't contain `access-control-request-method: GET` header in the IT. :-)

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
```diff
@@ -87,21 +83,12 @@ public void addViewControllers(ViewControllerRegistry registry) {
     registry.addRedirectViewController("/info", "/actuator/info");
   }
 
-  @Bean ArmeriaServerConfigurator corsConfigurator(
+  /** Configures the server at the last because of the specified {@link Order} annotation. */
+  @Order @Bean ArmeriaServerConfigurator corsConfigurator(
```
anuraaga, line 87: Does this need to be run after the other catch all bean? If the two have an ordering dependencies, we should specify values to `@Order` to guarantee the order I think.

hyangtack: Yes, you're right. But the orders for the other `ArmeriaServerConfigurator` beans are `0` which is a default when no `@Order` annotation is specified. And just specifying `@Order` makes its bean lowest precedence because the annotation's default value is `Integer.MAX_VALUE` (which means lowest precedence). 

So it's okay to specify `@Order` only from here. But if you want to make their order clear, it would also be okay to specify order on all `ArmeriaServerConfigurator` beans. :-)

See: 
- https://github.com/line/armeria/pull/1535
- https://line.github.io/armeria/apidocs/com/linecorp/armeria/spring/ArmeriaServerConfigurator.html#getOrder()

anuraaga: Ah no don't think we need to always specify it when there is no order dependency for the default value. But just wanted to confirm there is no ordering issue with the bean here which has the same order.

https://github.com/openzipkin/zipkin/blob/master/zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin2/autoconfigure/prometheus/ZipkinPrometheusMetricsAutoConfiguration.java

hyangtack: Ah, got it. I missed that. I guess there's no ordering issue between them, but I will change their order in order to make it clear. Thank you!

zipkin-autoconfigure/metrics-prometheus/src/main/java/zipkin2/autoconfigure/prometheus/ZipkinPrometheusMetricsAutoConfiguration.java
```diff
@@ -65,12 +66,13 @@
...
-  // adding metrics. We add a lowest-precedence path mapping so anything not mapped by another
+  // adding metrics. We add a lower precedence path mapping so anything not mapped by another
   // service is handled by this.
   @Bean
-  @Order
+  @Order(1)
   ArmeriaServerConfigurator notFoundMetricCollector() {
-    return sb -> sb.serviceUnder("/", (ctx, req) -> HttpResponse.of(HttpStatus.NOT_FOUND));
+    // Use glob instead of catch-all to avoid adding it to the trie router.
+    return sb -> sb.service(PathMapping.ofGlob("/**"), (ctx, req) -> HttpResponse.of(HttpStatus.NOT_FOUND));
```
hyangtack, line 75: @adriancole I fixed here to pass the test https://github.com/openzipkin/zipkin/blob/master/zipkin-server/src/test/java/zipkin2/server/internal/ITZipkinServerHttpCollectorDisabled.java#L49-L56. From the previous version, `405 method not allowed` returned because `/api/v1/spans` path exists but `POST` method is not allowed due to `zipkin.collector.http.enabled=false`. In this case `405 method not allowed` seems more reasonable, so would it better to fix the test code? Please let me know what you think! :-)

codefromthecrypt:
thanks so much for the help!

PR #2434, opened 2019-03-07
Fixes some lens UI bugs

drolando:
Service names are now alphabetically sorted. The API seems to return
them out of order, so let's sort them just to be sure.

Added missing space between span count and the "span" word.

Timestamps in the span details view now have microsecond precision.
moment.js doesn't support that, so I had to append the last 3 digits of
the timestamp to the human readable string.

CSS is weird. Apparently it's enough to set `white-space: normal` to
make long text wrap correctly in the span details view. I also tested
with a single very long word and if that doesn't fit in the screen it
gets split automatically.

### DIFF
--- a/zipkin-lens/src/actions/services-action.js
+++ b/zipkin-lens/src/actions/services-action.js
@@ -22,7 +22,9 @@ export const fetchServices = () => async (dispatch) => {
       throw Error(res.statusText);
     }
     const services = await res.json();
-    dispatch(fetchServicesSuccess(services));
+    // alphabetically sort service names since the api returns
+    // them out of order
+    dispatch(fetchServicesSuccess(services.sort()));
   } catch (err) {
     dispatch(fetchServicesFailure());
   }
--- a/zipkin-lens/src/actions/services-action.js
+++ b/zipkin-lens/src/actions/services-action.js
@@ -22,8 +22,8 @@ export const fetchServices = () => async (dispatch) => {
       throw Error(res.statusText);
     }
     const services = await res.json();
-    // alphabetically sort service names since the api returns
-    // them out of order
+    // alphabetically sort service names since the api might
+    // return them out of order
     dispatch(fetchServicesSuccess(services.sort()));
   } catch (err) {
     dispatch(fetchServicesFailure());
--- a/zipkin-lens/src/components/Timeline/TimelineSpanData.js
+++ b/zipkin-lens/src/components/Timeline/TimelineSpanData.js
@@ -41,7 +41,11 @@ const renderInfo = span => (
       data={
         span.annotations.map(a => (
           {
-            duration: moment(a.timestamp / 1000).format('MM/DD HH:mm:ss:SSS'),
+            // moment.js only supports millisecond precision, however our timestamps have microsecond
+            // precision. So we use moment.js to generate the human readable time with just milliseconds
+            // and then append the last 3 digits of the timestamp which are the microseconds.
+            // NOTE: a.timestamp % 1000 would save a string conversion but drops leading zeros.
+            duration: moment(a.timestamp / 1000).format('MM/DD HH:mm:ss.SSS') + a.timestamp.toString().slice(-3),
             relativeTime: a.relativeTime,
             annotation: a.value,
             address: a.endpoint,
@@ -64,7 +68,8 @@ const renderInfo = span => (
       columns={
         [
           { Header: 'Key', accessor: 'key' },
-          { Header: 'Value', accessor: 'value' },
+          // `white-space: normal` makes long text wrap correctly.
+          { Header: 'Value', accessor: 'value', style: {'white-space': 'normal'} },
         ]
       }
     />
--- a/zipkin-lens/src/components/Timeline/TimelineSpanData.js
+++ b/zipkin-lens/src/components/Timeline/TimelineSpanData.js
@@ -41,11 +41,13 @@ const renderInfo = span => (
       data={
         span.annotations.map(a => (
           {
-            // moment.js only supports millisecond precision, however our timestamps have microsecond
-            // precision. So we use moment.js to generate the human readable time with just milliseconds
-            // and then append the last 3 digits of the timestamp which are the microseconds.
+            // moment.js only supports millisecond precision, however our timestamps have
+            // microsecond precision. So we use moment.js to generate the human readable time
+            // with just milliseconds and then append the last 3 digits of the timestamp
+            // which are the microseconds.
             // NOTE: a.timestamp % 1000 would save a string conversion but drops leading zeros.
-            duration: moment(a.timestamp / 1000).format('MM/DD HH:mm:ss.SSS') + a.timestamp.toString().slice(-3),
+            duration:
+              moment(a.timestamp / 1000).format('MM/DD HH:mm:ss.SSS') + a.timestamp.toString().slice(-3),
             relativeTime: a.relativeTime,
             annotation: a.value,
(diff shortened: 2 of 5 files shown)


codefromthecrypt:
lookin to double your commits to this repo, huh? :D

seriously, thanks for the help, daniele!

zipkin-lens/src/components/Timeline/TimelineSpanData.js
```diff
@@ -41,7 +41,11 @@ const renderInfo = span => (
       data={
         span.annotations.map(a => (
           {
-            duration: moment(a.timestamp / 1000).format('MM/DD HH:mm:ss:SSS'),
+            // moment.js only supports millisecond precision, however our timestamps have microsecond
+            // precision. So we use moment.js to generate the human readable time with just milliseconds
+            // and then append the last 3 digits of the timestamp which are the microseconds.
+            // NOTE: a.timestamp % 1000 would save a string conversion but drops leading zeros.
+            duration: moment(a.timestamp / 1000).format('MM/DD HH:mm:ss.SSS') + a.timestamp.toString().slice(-3),
```
codefromthecrypt: an unrelated question is: why is a timestamp field named duration?

zipkin-lens/src/actions/services-action.js
```diff
@@ -22,7 +22,9 @@ export const fetchServices = () => async (dispatch) => {
       throw Error(res.statusText);
     }
     const services = await res.json();
-    dispatch(fetchServicesSuccess(services));
+    // alphabetically sort service names since the api returns
+    // them out of order
```
codefromthecrypt: being a geek, and ignore me. Technically it isn't required to return them out of order :) it is just not required to return them in order either. iotw order is undefined.

codefromthecrypt (approved):
the changes here make sense to me even if I have unrelated low priority questions.

codefromthecrypt:
there is a lint fail here:

```
[INFO] /home/travis/build/openzipkin/zipkin/zipkin-lens/src/components/Timeline/TimelineSpanData.js
[INFO]   44:1   error  Line 44 exceeds the maximum line length of 100  max-len
[INFO]   45:1   error  Line 45 exceeds the maximum line length of 100  max-len
[INFO]   72:56  error  A space is required after '{'                   object-curly-spacing
[INFO]   72:80  error  A space is required before '}'                  object-curly-spacing
[INFO] 
[INFO] ✖ 4 problems (4 errors, 0 warnings)
[INFO]   2 errors and 0 warnings potentially fixable with the `--fix` option.
```

npm run lint iirc will tell you about it locally

zipkin-lens/src/actions/services-action.js
drolando: Involuntarily you raise a good point: this is the wrong place to fix this. I should fix it in the java API so then it's fixed for everything.

codefromthecrypt: it won't be fixed for everything because we never defined the order in the api and we know there are clones etc. iotw, it would only be "fixed" for that version of the impl. I think since the api doesn't define order, what you have is correct.

codefromthecrypt: PS I do like the idea of us sorting even if it won't obviate doing it here. When you like I think all you'd need to do is a sort here https://github.com/openzipkin/zipkin/blob/master/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java#L159 and then add a test in ITZipkinServer.  That line should cover service, span names, and tag values in one go

drolando: I just tried that and it looks like sorting an array of strings in java is not as easy as I thought: https://circleci.com/gh/openzipkin/zipkin/3097?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
```java
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project zipkin-server: Compilation failure
[ERROR] /home/circleci/project/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java:[80,10] error: no suitable method found for sort(List<String>)
[ERROR]     method Arrays.sort(int[]) is not applicable
[ERROR]       (argument mismatch; List<String> cannot be converted to int[])
[ERROR]     method Arrays.sort(long[]) is not applicable
[ERROR]       (argument mismatch; List<String> cannot be converted to long[])
[ERROR]     method Arrays.sort(short[]) is not applicable
[ERROR]       (argument mismatch; List<String> cannot be converted to short[])
[ERROR]     method Arrays.sort(char[]) is not applicable
[ERROR]       (argument mismatch; List<String> cannot be converted to char[])
[ERROR]     method Arrays.sort(byte[]) is not applicable
[ERROR]       (argument mismatch; List<String> cannot be converted to byte[])
[ERROR]     method Arrays.sort(float[]) is not applicable
[ERROR]       (argument mismatch; List<String> cannot be converted to float[])
[ERROR]     method Arrays.sort(double[]) is not applicable
[ERROR]       (argument mismatch; List<String> cannot be converted to double[])
[ERROR]     method Arrays.sort(Object[]) is not applicable
[ERROR]       (argument mismatch; List<String> cannot be converted to Object[])
[ERROR]     method Arrays.<T#1>sort(T#1[],Comparator<? super T#1>) is not applicable
[ERROR]       (cannot infer type-variable(s) T#1
[ERROR]         (actual and formal argument lists differ in length))
[ERROR]     method Arrays.<T#2>sort(T#2[],int,int,Comparator<? super T#2>) is not applicable
[ERROR]       (cannot infer type-variable(s) T#2
[ERROR]         (actual and formal argument lists differ in length))
[ERROR]   where T#1,T#2 are type-variables:
[ERROR]     T#1 extends Object declared in method <T#1>sort(T#1[],Comparator<? super T#1>)
[ERROR]     T#2 extends Object declared in method <T#2>sort(T#2[],int,int,Comparator<? super T#2>)
```
so I dropped that.

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java
```diff
@@ -156,6 +157,7 @@ public AggregatedHttpMessage getAutocompleteValues(@Param("key") String key) thr
    * active use, and active use usually implies more than 3 services.
    */
   AggregatedHttpMessage maybeCacheNames(boolean shouldCacheControl, List<String> values) {
+    Collections.sort(values);
```
codefromthecrypt, line 160: did a quick check and I don't think any backends return immutable lists. so this should be fine

codefromthecrypt (approved):

tacigar (approved):
LGTM!!!

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java
shakuzen, line 160: Why sort them in the server and in the UI? I think just sorting in the UI is fine until we have a reason to sort them in the API response.

codefromthecrypt:
sorting in the api was mainly because daniele felt strongly about it. it
does add a slight risk that a downstream storage module uses immutable
collection and this would break them.

personally I dont care too much about it. there is a great deal of ui code
that adjusts span data so it seems odd to focus so much on the service list
part of the api knowing that.

codefromthecrypt:
I'm ok being on the hook if something breaks, which I don't think will.

Thanks @drolando really nice to fix some broken windows

PR #2437, opened 2019-03-07
Adds patch until quoting bug is fixed upstream

codefromthecrypt:
We need to cut this immediately as prometheus scraping is broke.

See https://github.com/line/armeria/issues/1637

Before the prom data was quoted by accident as it was accidentally set to json
```bash
$ curl -vs http://localhost:9411/actuator/prometheus|head -c 140
*   Trying ::1...
* TCP_NODELAY set
> GET /actuator/prometheus HTTP/1.1
> Host: localhost:9411
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< content-type: application/json; charset=utf-8
< content-length: 8833
< 
{ [8833 bytes data]
"# HELP armeria_server_router_virtualHostCache_evictions_total  \n# TYPE armeria_server_router_virtualHostCache_evictions_total counter\narm
```

Now, it is ok
```bash
$ curl -vs http://localhost:9411/actuator/prometheus|head -c 140
*   Trying ::1...
* TCP_NODELAY set
> GET /actuator/prometheus HTTP/1.1
> Host: localhost:9411
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< content-type: text/plain; version=0.0.4; charset=utf-8
< content-length: 8986
< 
{ [8986 bytes data]
# HELP zipkin_collector_message_bytes size of a message containing serialized spans
# TYPE zipkin_collector_message_bytes gauge
zipkin_colle
```

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
@@ -14,19 +14,29 @@
 package zipkin2.server.internal;
 
 import brave.Tracing;
+import com.google.gson.JsonParser;
+import com.linecorp.armeria.common.AggregatedHttpMessage;
 import com.linecorp.armeria.common.HttpMethod;
+import com.linecorp.armeria.common.HttpRequest;
+import com.linecorp.armeria.common.HttpResponse;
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
going out immediately as 2.12.5

Issue #2425, opened 2019-03-05
Lens UI no space for the used-by pie


jcarres-mdsol:
## Describe the Bug
Using the Lens UI, dependency view, when clicking on a service, I can see on the right I can see information about used-by and using services.

When I have a lot (40) services using one of my services, the tags take a lot of space and there is no space for the pie chart which gets crunched and almost invisible.

## Steps to Reproduce
Have a crazy architecture, lots of services connected to one particular service, go to dependencies, click on the service which is used by everyone.

## Expected Behaviour
I would expect the pie charts to have the same size and not depend on the size taken by the tags.

tacigar:
Thank you for your feedback!! I'll take a look.

PR #2442, opened 2019-03-15
Fix piechart no-space bug

tacigar:
#2425 

Delete legends that display service names because it puts pressure on pie chart.
These legends are not necessary because similar information is on the top.

**before**

**after**


codefromthecrypt (approved):
good idea

jcarres-mdsol:
Cool, thanks!

PR #2457, opened 2019-03-20
Use Elasticsearch REST Client instead of OKHTTP

gquintana:
This is an attempt to replace OK HTTP library with Elasticsearch Low level Client.

Pros:
* Supports multiple HTTPS endpoints
* Discovers dynamically new/removed nodes (called sniffing)

Cons:
* Based on Apache HTTP Client, different threading/queuing/IO model
* Mostly Elasticsearch specific HTTP Client

TODOs:
- [ ] Fix failing tests
- [ ] Fix todos and clean code
- [ ] Test Zipkin Server on real ES cluster
- [ ] Sniffing autoconfiguration
- [ ] ...
 
At this point, this a proof of concept. The question is: Is it worth continuing? Has this contribution any chance to get accepted (after fixing everything)?


codefromthecrypt:
My guess is this will introduce problems like the following if you are able to get the tests to pass (usually ES doesn't work with a wide version range, at least right now we are rev-locked using their library in zipkin-dependencies). We need to make sure the cure isn't worse than the disease.

Here are some things you'd hit if you continue:

This will break the auth integration in the zipkin-aws repository. For example, an interceptor doing AWS authentication exists there.

This will break the env variables used to debug connections with the logging interceptor.

This will break compile on the builder, which we try not to do, as eventhough we don't support custom servers the storage library might be used directly (even if probably few would).

Finally, this might have some library skew management besides elasticsearch themselves. Should look at the dependencies especially what their v7 library will have.

codefromthecrypt:
I do have hesitations about the elasticsearch (apache httpclient based) library, mentioned above. Full disclosure I personally prefer okhttp and enjoy coding with it... this module has been easy to support for years and often times authors of change switch direction. Wanting to maintain something is important.

Having something others are happy to maintain is an important part of the decision. For example, @anuraaga wrote the original one, but he uses stackdriver now. IIRC @jcarres-mdsol uses this in AWS, eventhough the first AWS user was @devinsba. People change roles and integrations like this fade out of focus. It may fade out of focus for you in a year who knows.. that's why we need buy in.

Let's say tests pass without dropping versions, aws still works etc.. To answer your question directly about mergability.  I'd like to see another maintainer who isn't currently required to use elasticsearch ok with helping maintain this. We should also revise our support policy in that case to follow elasticsearch's which will affect end sites. For example, if they haven't already dropped a version we support, you can guarantee they will. We need to be transparent with users that our support range will be in elasticsearch's hands if this is merged. If other committers are on-board even if you or I step away, I'm fine on it.

I would prefer that similar effort was made to address whatever the problem is in okhttp though as that would break less people. However, yeah I mean what I said, if other folks can get behind this, I'd also be ok.

gquintana:
I understand. I won't argue, thanks for this detailed answer.

PR #2460, opened 2019-03-21
Fixes dependency scope of vendored gson

codefromthecrypt:
this saves 240KiB for an unused dep


codefromthecrypt:
circleci does parallel tests I think it isn't getting the change to core propagated to collector. since ASF we can't use circleci anyway I'm merging and hoping for the best :P

PR #2462, opened 2019-03-21
Fix /metrics reporting wrong values

### LINKED ISSUE #2430: /metrics doesn't increment for kafka
possibly because of the two kafka configs, the kafka section of the /metrics endpoint doesn't increment while it does on /actuator/prometheus

shakuzen:
Multiple metrics can have the same name with different tag values, so looking up the metric by name was giving the wrong value sometimes before.

Fixes #2430


codefromthecrypt (approved):
hoping for the best!

Issue #1990, opened 2018-04-10
Zipkin support for ActiveMQ as a collector


Shatakshi26:
We are using ActiveMQ in our project and we don't want to go for any other queue, so it will be good if Zipkin provides support for ActiveMQ as a collector.

---
This issue is now closed. Until a release happens with this change, you can look at the PR description of #2639 for testing

Shatakshi26:
We are using ActiveMQ in our project and we don't want to go for any other queue, so it will be good if Zipkin provides support for ActiveMQ as a collector.

codefromthecrypt:
looks like there is an implementation started here https://github.com/openzipkin/zipkin/pull/2466

codefromthecrypt:
Let's assume we are talking about activemq (not artemis or apollo)

I like that activemq has very few dependencies, however they do add 2MiB weight to the server.
```
# I bundled the dependencies
$ jar -tvf activemq.jar
     0 Tue Mar 26 08:30:02 ICT 2019 META-INF/
   214 Tue Mar 26 08:30:02 ICT 2019 META-INF/MANIFEST.MF
     0 Tue Mar 26 08:30:02 ICT 2019 lib/
1431933 Fri Mar 15 07:58:56 ICT 2019 lib/activemq-client-5.15.9.jar
 41203 Thu Mar 16 17:36:32 ICT 2017 lib/slf4j-api-1.7.25.jar
 32359 Fri Feb 01 11:35:24 ICT 2008 lib/geronimo-jms_1.1_spec-1.1.1.jar
 50155 Fri Jun 20 11:32:38 ICT 2014 lib/hawtbuf-1.11.jar
 20220 Fri Feb 01 11:42:20 ICT 2008 lib/geronimo-j2ee-management_1.1_spec-1.0.1.jar
$ du -k activemq.jar
2112	/Users/acole/.m2/repository/io/zipkin/custom/security/1.0-SNAPSHOT/security-1.0-SNAPSHOT-module.jar
```

Moreover, I am concerned that one of its non-annotation dependencies has not been updated since 2014: hawtbuf

I'm not sure which liabilities exist here https://github.com/fusesource/hawtbuf but this library is used all over activemq. @gtully can you weigh in on this point?
```
[INFO] \- org.apache.activemq:activemq-client:jar:5.15.9:compile
[INFO]    +- org.slf4j:slf4j-api:jar:1.7.25:compile
[...]

PR #2466, opened 2019-03-25
add collector support activemq

tian-junwei:
like collector-rabbitmq the collector-activemq support collector consumes a ActiveMQ queue for messages that contain a list of spans


codefromthecrypt:
@IAMTJW I like your energy and eagerness to do work. We have a change policy though which suggests multiple people need to buy-in before we make a standard feature. For example, adding this here also has impact as people using tracers will also want it. there's a lot of work around it.

I see that there are a couple votes here which is good. https://github.com/openzipkin/zipkin/issues/1990

We do need buy-in from folks in @openzipkin/core before adding this. Let's try to see if anyone is interested. Whatever we add becomes years of support work, so it is only fair to make sure people are ok with that.

tian-junwei:
> @IAMTJW I like your energy and eagerness to do work. We have a change policy though which suggests multiple people need to buy-in before we make a standard feature. For example, adding this here also has impact as people using tracers will also want it. there's a lot of work around it.
> 
> I see that there are a couple votes here which is good. #1990
> 
> We do need buy-in from folks in @openzipkin/core before adding this. Let's try to see if anyone is interested. Whatever we add becomes years of support work, so it is only fair to make sure people are ok with that.

get it , our company use activemq as collector，use postgresql as storage and I want to make it available to others.

codefromthecrypt:
@IAMTJW would you be willing to contribute senders in zipkin-reporter-java and other languages for this?

zipkin-autoconfigure/collector-activemq/src/main/resources/zipkin-server-shared.yml
```diff
@@ -0,0 +1,221 @@
+zipkin:
```
codefromthecrypt: you don't need this file, I think you want to make activemq.yml right?

zipkin-collector/activemq/README.md
```diff
@@ -0,0 +1,56 @@
...
+Property | Environment Variable | Description
+--- | --- | ---
+`zipkin.collector.activemq.connection-timeout` | `ACTIVE_CONNECTION_TIMEOUT` | Milliseconds to wait establishing a connection. Defaults to `60000` (1 minute)
+`zipkin.collector.activemq.queue` | `ACTIVE_QUEUE` | Queue from which to collect span messages. Defaults to `zipkin`
+
+If the URI is set, the following properties will be ignored.
+
+Property | Environment Variable | Description
+--- | --- | ---
+`zipkin.collector.activemq.addresses` | `ACTIVE_ADDRESSES` | Comma-separated list of ActiveMQ addresses, ex. `tcp://localhost:61616,tcp://localhost2:61616`
```
codefromthecrypt: is the naming convention `ACTIVE` normal in activemq or is it `ACTIVEMQ` or different? can you give example of some commandline tool which uses this convention?

zipkin-collector/activemq/src/main/java/zipkin2/collector/activemq/ActiveMQCollector.java
```diff
@@ -0,0 +1,233 @@
...
+        builder.connectionFactory.setBrokerURL("failover:("+builder.addresses+")?initialReconnectDelay=100");
+        connection = builder.connectionFactory.createConnection();
+        Session session=connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
+        connection.start();
+        Destination destination = session.createQueue(builder.queue);
+        MessageConsumer messageConsumer = session.createConsumer(destination);
+        Collector collector = builder.delegate.build();
+        CollectorMetrics metrics = builder.metrics;
+        messageConsumer.setMessageListener(new ActiveMQSpanConsumerMessageListener(collector,metrics));
+      }catch (Exception e){
```
codefromthecrypt: use multi-catch for what is actually thrown, unless literally Exception is thrown

codefromthecrypt (commented):
this is good progress. please look at tests for rabbit or kafka to ensure yours are equally comprehensive.

ghost (approved):
Im ready

codefromthecrypt:
@IAMTJW @moekelley76 can you put your vote on the original issue with thumbsup? https://github.com/openzipkin/zipkin/issues/1990

tian-junwei:
> this is good progress. please look at tests for rabbit or kafka to ensure yours are equally comprehensive.

Is there  a activemq server to be used test ?

jbertram:
How important is performance here, and what's the performance like for this implementation? ActiveMQ 5.x isn't particularly high-performance at this point given its underlying architecture, and storing messages in a database further slows things down. I'm new to Zipkin so I'm curious. Thanks!

codefromthecrypt:
usually performance of the collector is secondary to applicability. for
example, the most applicable transport to a site is a differentiator as it
allows them to use it. a high performance but not applicable answer is just
code to maintain. that is why we do things like "rule of 3" and other
guards to ensure the transports are viable to the most folks.

jbertram:
I was asking about performance more generally and also as it applies to this particular implementation. Is the performance of a collector in general an important factor? Assuming this collector is "applicable," what kind of performance does it provide?

codefromthecrypt:
it is usually the case that storage is slower than the collection
transport. I would be surprised if any activemq option is too slow

codefromthecrypt:
for example some sites are only taking a few hundred messages per second at
100% sample rate.

jbertram:
Is a few hundred messages per second considered near the top-end of the potential volume?

codefromthecrypt:
@jbertram depending on bundling mechanics, higher throughput can be thousands or tens of thousands per second. Higher throughput sites often run multiple collectors though. I would say that most users will not be doing more than a thousand of messages/second.

In the future, this may change as more people explore the costs involved in unsampled (usually this means late sampled, or tiered) collection. For example, we have a https://github.com/adriancole/zipkin-voltdb experiment which can swallow a lot more data than most storage backends. If someone doing 1k messages per second at 1% decides to do 100%, that means multiplying the throughput. For example, some doing what we call "firehose mode" like yelp, they had to write custom collectors as they went over what their kafka setup could do. This is more a thought for you vs what's commonly the case now. I mean to hint there could be some sites doing 100k messages/second into a short term buffer, even if I have no idea when that would be practiced and what the sharding of that rate would typically look like.

Back to reality :)

You'll notice some mentioning the order of thousands of spans per second here. No one currently is using activemq, but that's what this issue is about :) spans should be bundled (ex many spans per message), so if you have say.. 8k spans/second, ideally this is bundled into <1k messages/second of size in orders of low kilobytes.

https://cwiki.apache.org/confluence/display/ZIPKIN/Sites

I'll ask some of the sites to provide more fine grained details on bundling mechanics. Thanks very much for the interest!

codefromthecrypt:
@IAMTJW are you using this in production now (from a fork?) if so, what's your ingest rate?

PS it may be useful to bring this work into a separate repo to evaluate. for example, I made a project like this to help people test/create mongo without affecting the core repository. I can make something like this for you if it would help. https://github.com/adriancole/zipkin-mongodb

tian-junwei:
> @IAMTJW are you using this in production now (from a fork?) if so, what's your ingest rate?
> 
> PS it may be useful to bring this work into a separate repo to evaluate. for example, I made a project like this to help people test/create mongo without affecting the core repository. I can make something like this for you if it would help. https://github.com/adriancole/zipkin-mongodb

We are using this in production(release 2.8.3), but we have only dozens of applications and have not much spans per second here

codefromthecrypt:
@moekelley76 by "I'm ready" does this mean you can try this code? what are you using now and how much throughput of messages do you expect?

codefromthecrypt:
@ImFlog PS we are still waiting for another person to mention they will test this.

As this repository will shortly move to the Apache Software Foundation (ASF) and this is non-trivial code, you should probably prepare by sending a contributor agreement to the ASF. If you have already done this in the past, no work to do. This is a once in life thing.

If you haven't already, easiest way is to download the ICLA template. After filling the [icla.pdf](https://www.apache.org/licenses/icla.pdf) with personal information correctly, print, sign, scan, and send it in mail as an attachment to the secretary (secretary@apache.org)

codefromthecrypt:
whoops I meant @IAMTJW not @ImFlog !

tian-junwei:
> whoops I meant @IAMTJW not @ImFlog !

I  have send mail to secretary@apache.org

codefromthecrypt:
PS this repo is no longer moving to ASF (none are)

thanhctES:
I am waiting for this. Please merge it.

codefromthecrypt:
superceded by https://github.com/openzipkin/zipkin/pull/2639

Issue #2467, opened 2019-03-25
Basic Auth setup for the new (armeria) http collector endpoint


gquintana:
**Feature:**

Add authentication and autorisation control on HTTP REST API and UIs. This is a continuation of #782, specific to adding basic auth example integration instructions.

**Rational**

* Traces may content sensitive information, not everyone should be able to see.
* Avoid inappropriate spans in storage: control who's allowed to push spans over HTTP API

I admit security is a complicated topic. The MVP for me would be file based basic auth.

**Example Scenario**

* Do canary testing, giving access to Zipkin to a subset of users
* Protect Zipkin server against attacks (or bad usages)

**Prior Art**

* Zipkin doesn't support HTTPS yet https://github.com/openzipkin/zipkin/issues/2449 which is a requirement
* Spring Security https://spring.io/guides/tutorials/spring-security-and-angular-js/

codefromthecrypt:
there is an existing issue about this. hard coding basic auth credentials is probably not something we would want to standardize on for obvious reasons. however, it should be possible.

maybe the below no longer works due to drift from the armeria update will have a look

https://github.com/openzipkin/zipkin/issues/782#issuecomment-433306013

gquintana:
> there is an existing issue about this. hard coding basic auth credentials is probably not something we would want to standardize on for obvious reasons. 

Reasons are not so obvious to me. Can you explain?

I tried the recipe described in the comment:
1. Add spring-security to class path 
```
$ jar -tvf zipkin-server-2.12.7-SNAPSHOT-exec.jar |grep security
   408 Fri Feb 15 10:42:02 CET 2019 BOOT-INF/lib/spring-boot-starter-security-2.1.3.RELEASE.jar
778579 Wed Feb 13 17:37:46 CET 2019 BOOT-INF/lib/spring-security-config-5.1.4.RELEASE.jar
433224 Wed Feb 13 17:37:14 CET 2019 BOOT-INF/lib/spring-security-core-5.1.4.RELEASE.jar
537635 Wed Feb 13 17:37:30 CET 2019 BOOT-INF/lib/spring-security-web-5.1.4.RELEASE.jar
```
2. Add properties to Spring boot config:
```
spring:
  security:
    user:
      name: zipkin
      password: zipkin
```

The security filter chain appears to be created:
```
[...]

PR #2479, opened 2019-03-31
Updates to latest versions, notably Armeria

codefromthecrypt:


codefromthecrypt:
See https://github.com/openzipkin/zipkin/issues/2467 as this version of armeria support TLS setup. Note: we need a actual test case to imply this works.

PR #2482, opened 2019-04-03
Make hover effects in lens better

tacigar:
Make hover effects in lens better for better UX.

**BEFORE**

before-hover

**AFTER**

after-hover


codefromthecrypt:
So as I understand it, this makes what you are hovering over darker, so it is harder to lose your cursor? Am I missing anything else?

tacigar:
That's all :)

codefromthecrypt (approved):
I agree this is helpful to see what you are doing

codefromthecrypt:
thank you so much!

PR #2483, opened 2019-04-03
Polishes MySQL schema and resurrects test initialization

codefromthecrypt:
I noticed the init scripts in our docker mysql tests were never invoked.

When fixing them, I noticed some cases where we were using more
complicated unique key + index statements when we had no nullable fields
declared. Particularly, this caused a syntax glitch in mariadb on the
dependencies table. Easiest way out was to switch these to primary keys.


zipkin-storage/mysql-v1/src/main/resources/mysql.sql
```diff
@@ -42,7 +41,6 @@ CREATE TABLE IF NOT EXISTS zipkin_dependencies (
   `parent` VARCHAR(255) NOT NULL,
   `child` VARCHAR(255) NOT NULL,
   `call_count` BIGINT,
-  `error_count` BIGINT
+  `error_count` BIGINT,
+  PRIMARY KEY (`day`, `parent`, `child`)
 ) ENGINE=InnoDB ROW_FORMAT=COMPRESSED CHARACTER SET=utf8 COLLATE utf8_general_ci;
-
-ALTER TABLE zipkin_dependencies ADD UNIQUE KEY(`day`, `parent`, `child`);
```
codefromthecrypt, line 48: this syntax was crashing initialization on the docker container.. not sure why, but anyway fixed now

zipkin-storage/mysql-v1/src/main/resources/mysql.sql
```diff
@@ -6,11 +6,10 @@ CREATE TABLE IF NOT EXISTS zipkin_spans (
   `parent_id` BIGINT,
   `debug` BIT(1),
   `start_ts` BIGINT COMMENT 'Span.timestamp(): epoch micros used for endTs query and to implement TTL',
-  `duration` BIGINT COMMENT 'Span.duration(): micros used for minDuration and maxDuration query'
+  `duration` BIGINT COMMENT 'Span.duration(): micros used for minDuration and maxDuration query',
+  PRIMARY KEY (`trace_id_high`, `trace_id`, `id`)
 ) ENGINE=InnoDB ROW_FORMAT=COMPRESSED CHARACTER SET=utf8 COLLATE utf8_general_ci;
 
-ALTER TABLE zipkin_spans ADD UNIQUE KEY(`trace_id_high`, `trace_id`, `id`) COMMENT 'ignore insert on duplicate';
```
codefromthecrypt, line 12: since no fields here are nullable in the schema, it is simpler to just use primary key

zipkin-storage/mysql-v1/src/test/java/zipkin2/storage/mysql/v1/ZipkinMySQLContainer.java
```diff
@@ -14,47 +14,24 @@
...
 
-  public ZipkinMySQLContainer(String version) {
+  ZipkinMySQLContainer(String version) {
     super("openzipkin/zipkin-mysql:" + version);
-
     withExposedPorts(3306);
-
-    setWaitStrategy(new AbstractWaitStrategy() {
-      @Override
-      protected void waitUntilReady() {
```
codefromthecrypt, line 43: this basically never succeeded, which delayed container startup ignoring the fact that the init scripts never ran

Issue #1794, opened 2017-11-14
Decide what to do about remote service name and search


codefromthecrypt:
Right now, we return all service names, local and remote in /api/v2/services (and v1 for that matter). This is used for the query search drop-down, and they also correspond to the dependency graph. 

We currently don't have any tests for being able to search by remote service name, so storage implementations handle this differently. I'm of the opinion we should either remove the remote service names from the list, or decide how to handle a query like this. For example, do we want to allow a query like "where I called service X" or do we want to leave like many are doing which is "either I am this service or this service was called"?

Until we do something, we should ack that we write extra data and I'm pretty sure more than one implementation only consider the local service name in queries!

cc @openzipkin/core

xqliang:
Some times i  need to known "whether calling X remote service is slowing".

codefromthecrypt:
right, so the span name is attached to a local service (the services query
mixes local and remote service when choosing a span name). I wonder if we
shouldn't have a separate drop-down for remoteService? Or maybe the mix is
ok.. what do you think?

xqliang:
mix is ok for me

goller:
I like the mix option because from a UI point-of-view one is looking for a specific service regardless if it is local or remote.  

I think it is better that there is a UI that allows one to "move up" towards parent spans or "down" to children.

codefromthecrypt:
>
> I think it is better that there is a UI that allows one to "move up"
> towards parent spans or "down" to children.
>
mind raising a separate issue? maybe with a screen shot with apple preview
annotation sketch (or your favorite technique) to describe this?

codefromthecrypt:
This issue came up again in #2337 I think that end users are paying a lot for this, and we should simply stop writing indexes for remote and local service names against the same query conditions.

One of the more damaging things conceptually is when one conflates local and remote service names, so what we do in the services index actually contributes to this.

If we do stop indexing the spans against a remote service name, we will lose the ability to search for remote services.. unless we introduce another query which is remoteService=X. I'm personally still better with this as we reduce expense and in some cases restore usability meanwhile.

codefromthecrypt:
re-pinging @openzipkin/core @openzipkin/ui @openzipkin/cassandra as IMHO we can't keep tripping over the same problem year after year.

[...]

PR #2484, opened 2019-04-03
De-conflate local from remote service names

codefromthecrypt:
Especially mentioned in #1794, local and remote service names are undeniably serving different use cases, yet currently conflated in the api.

/services -> only local serviceNames
/remoteServices?serviceName=X -> new: only remote serviceNames for auto-complete
/spans?remoteServiceName=X -> new: to restore functionality @llinder mentioned

cc @openzipkin/ui @bulicekj

### DIFF
--- a/zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/Schema.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/Schema.java
@@ -82,25 +82,25 @@ final class Schema {
         UPGRADE_1);
     }
 
-    boolean hasRemoteServiceByService = hasUpgrade2_remoteService(keyspaceMetadata);
-    if (!hasRemoteServiceByService) {
+    boolean hasRemoteService = hasUpgrade2_remoteService(keyspaceMetadata);
+    if (!hasRemoteService) {
       LOG.warn(
         "schema lacks remote service indexing: apply {}, or set CassandraStorage.ensureSchema=true",
         UPGRADE_2);
     }
 
-    return new Metadata(compactionClass, hasAutocompleteTags, hasRemoteServiceByService);
+    return new Metadata(compactionClass, hasAutocompleteTags, hasRemoteService);
   }
 
   static final class Metadata {
     final String compactionClass;
-    final boolean hasAutocompleteTags, hasRemoteServiceByService;
+    final boolean hasAutocompleteTags, hasRemoteService;
 
     Metadata(String compactionClass, boolean hasAutocompleteTags,
-      boolean hasRemoteServiceByService) {
+      boolean hasRemoteService) {
       this.compactionClass = compactionClass;
       this.hasAutocompleteTags = hasAutocompleteTags;
-      this.hasRemoteServiceByService = hasRemoteServiceByService;
+      this.hasRemoteService = hasRemoteService;
     }
   }
 
--- a/zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/Schema.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/Schema.java
@@ -75,11 +75,21 @@ final class Schema {
     String compactionClass =
       keyspaceMetadata.getTable("span").getOptions().getCompaction().get("class");
 
-    return new Metadata(
-      compactionClass,
-      hasUpgrade1_autocompleteTags(keyspaceMetadata),
-      hasUpgrade2_remoteService(keyspaceMetadata)
-    );
+    boolean hasAutocompleteTags = hasUpgrade1_autocompleteTags(keyspaceMetadata);
+    if (!hasAutocompleteTags) {
+      LOG.warn(
+        "schema lacks autocomplete indexing: apply {}, or set CassandraStorage.ensureSchema=true",
+        UPGRADE_1);
+    }
+
+    boolean hasRemoteServiceByService = hasUpgrade2_remoteService(keyspaceMetadata);
+    if (!hasRemoteServiceByService) {
+      LOG.warn(
+        "schema lacks remote service indexing: apply {}, or set CassandraStorage.ensureSchema=true",
+        UPGRADE_2);
+    }
+
+    return new Metadata(compactionClass, hasAutocompleteTags, hasRemoteServiceByService);
   }
 
   static final class Metadata {
--- a/zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/Schema.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/Schema.java
@@ -43,6 +43,7 @@ final class Schema {
   static final String TABLE_SPAN = "span";
   static final String TABLE_TRACE_BY_SERVICE_SPAN = "trace_by_service_span";
   static final String TABLE_SERVICE_SPANS = "span_by_service";
+  static final String TABLE_SERVICE_REMOTE_SERVICES = "remote_service_by_service";
   static final String TABLE_DEPENDENCY = "dependency";
   static final String TABLE_AUTOCOMPLETE_TAGS = "autocomplete_tags";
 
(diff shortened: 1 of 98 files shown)


codefromthecrypt:
oh wow.. cassandra tests pass now :) someone quick take a photo!

zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/Schema.java
```diff
@@ -73,14 +75,22 @@ static Metadata readMetadata(Session session) {
     String compactionClass =
       keyspaceMetadata.getTable("span").getOptions().getCompaction().get("class");
 
-    return new Metadata(compactionClass);
+    return new Metadata(
+      compactionClass,
+      hasUpgrade1_autocompleteTags(keyspaceMetadata),
```
codefromthecrypt: @zeagord you and I both forgot to check this!

zipkin-storage/cassandra/src/test/java/zipkin2/storage/cassandra/ITEnsureSchema.java
```diff
@@ -77,12 +90,19 @@
     try (CassandraStorage storage = CassandraStorage.newBuilder()
       .contactPoints(contactPoint.getHostString() + ":" + contactPoint.getPort())
       .ensureSchema(false)
+      .autocompleteKeys(asList("environment"))
```
codefromthecrypt, line 98: @zeagord here was the secret to causing us to figure out if we *really* work when schema isn't updated

codefromthecrypt:
I am nearly done the hard parts. just have to implement query support for cassandra and cassandra v1. after that, it is easy

codefromthecrypt:
@openzipkin/cassandra I'm a little stuck on the best way to model the remoteServiceName=X query.

I'm thinking about adding this for the remote service index. only other option I can think of is adding a SASI field on spans of r_service. thoughts?


```
CREATE TABLE IF NOT EXISTS zipkin2.trace_by_service_remote_service (
    service       text,             //-- service name, or blank to support queries without service name
    remote_service  text,             //-- remote service name
    bucket        int,              //-- time bucket, calculated as ts/interval (in microseconds), for some pre-configured interval like 1 day.
    ts            timeuuid,         //-- start timestamp of the span, truncated to millisecond precision
    trace_id      text,             //-- trace ID
    PRIMARY KEY ((service, remote_service, bucket), ts)
)
   WITH CLUSTERING ORDER BY (ts DESC)
    AND compaction = {'class': 'org.apache.cassandra.db.compaction.TimeWindowCompactionStrategy'}
    AND default_time_to_live =  259200
    AND gc_grace_seconds = 3600
    AND read_repair_chance = 0
    AND dclocal_read_repair_chance = 0
    AND speculative_retry = '95percentile'
    AND comment = 'Secondary table for looking up a trace by remote service name. service column may be blank (when only looking up by remote_service). bucket column adds time bucketing to the partition key, values are microseconds rounded to a pre-configured interval (typically one day). ts column is start timestamp of the span as time-uuid, truncated to millisecond precision.';
```
Note: this would be in addition to [trace_by_service_span](https://github.com/openzipkin/zipkin/blob/master/zipkin-storage/cassandra/src/main/resources/zipkin2-schema-upgrade-1.cql), and require a client-side join if other parameters are sent (similar to how we do with annotation query).

Another option is to extend trace_by_service_span multiplying rows there by remote service name. This would be simpler to query... Not sure what's better? throwing this out to get feedback.

zipkin-storage/cassandra/src/test/java/zipkin2/storage/cassandra/ITEnsureSchema.java
zeagord: :) I totally forgot about this.

zipkin-storage/cassandra/src/main/resources/zipkin2-schema-upgrade-2.cql
```diff
@@ -0,0 +1,30 @@
+CREATE TABLE IF NOT EXISTS zipkin2.remote_service_by_service (
```
codefromthecrypt, line 1: @michaelsembwever @llinder @drolando can one of you have a look at this file? I'm just sort of replacing inline with what we did for span name indexing, but someone who knows cassandra should weigh in.

zipkin-storage/cassandra/src/main/resources/zipkin2-schema-indexes.cql
```diff
@@ -33,20 +62,19 @@ CREATE TABLE IF NOT EXISTS zipkin2.span_by_service (
...
-    'analyzed': 'true',
-    'analyzer_class':'org.apache.cassandra.index.sasi.analyzer.DelimiterAnalyzer',
-    'delimiter': '░'};
-
-CREATE CUSTOM INDEX IF NOT EXISTS ON zipkin2.span (l_service) USING 'org.apache.cassandra.index.sasi.SASIIndex'
-   WITH OPTIONS = {'mode': 'PREFIX'};
-
-CREATE CUSTOM INDEX IF NOT EXISTS ON zipkin2.trace_by_service_span (duration) USING 'org.apache.cassandra.index.sasi.SASIIndex'
-   WITH OPTIONS = {'mode': 'PREFIX'};
+CREATE TABLE IF NOT EXISTS zipkin2.remote_service_by_service (
```
michaelsembwever, line 65: is adding the tables `trace_by_service_remote_service` and `remote_service_by_service` the only actual change in the file's diff?

zipkin-storage/cassandra/src/main/resources/zipkin2-schema-indexes.cql
```diff
@@ -33,20 +62,19 @@ CREATE TABLE IF NOT EXISTS zipkin2.span_by_service (
     AND speculative_retry = '95percentile'
     AND comment = 'Secondary table for looking up span names by a service name.';
 
-DROP INDEX IF EXISTS zipkin2.span_annotation_query_idx;
```
codefromthecrypt, line 36: this index didn't exist

zipkin-storage/cassandra/src/main/resources/zipkin2-schema-upgrade-2.cql
michaelsembwever: It's an anti-pattern, you want to be careful with this, as it involves hot partitions. (That's why LCS with a low threshold and row caching. Deduping the writes is also important.)
So long as you apply the caution we took with `span_by_service` it's cool.

btw there's the stress profile if you want to see how the performance/throughput numbers differ…

zipkin-storage/cassandra/src/main/resources/zipkin2-schema-indexes.cql
codefromthecrypt: yes, the other changes were moving the indexing near the tables they indexed as it was confusing me (at the cost of confusing you in this review.. sorry)

zipkin-storage/cassandra/src/main/resources/zipkin2-schema-upgrade-2.cql
codefromthecrypt: I will add comments to all the tables that have deduplication on them, as yeah in the schema now we make no such notes yet.

codefromthecrypt: added a commit revising comments, feel free to suggest revisions to those if needed. thanks!

codefromthecrypt: also backfilled missing stress tests

codefromthecrypt:
cassandra 3.11.3+ done (except some unit tests which all around need to be added but saving for later)
next up cassandra 2.2+ impl (cassandra-v1)

codefromthecrypt:
will merge this prior to the cassandra v1 impl as it was too hard to see what changed https://github.com/openzipkin/zipkin/pull/2492

codefromthecrypt:
ok cassandra v1 "done" (same quality caveat)

now I will do a quality sweep including unit tests and README updates. if the build is green, it is testable

codefromthecrypt:
only task left is unit test sweep.. doing that now.

PR #2489, opened 2019-04-05
Allows POST with content-type CORS config by default

### LINKED ISSUE #2486: /api/v2 fails to handle CORS
## Describe the Bug
When a reporter sends an OPTIONS request to /api/v2, the server gives an empty response and no tracing is done (no actual POST with data is done). See https://github.com/ewhauser/angular-tracing/issues/11

## Steps to Reproduce
In the above mentioned repo, run the demo with `yarn install` and `yarn quick`. When observing the developer tools of your browser, you should see OPTIONS-requests to localhost:9411, but no POST's with actual span data.

codefromthecrypt:
We missed a spot when porting our CORS config. We didn't realize that
before we switched to Armeria, that we allowed POST when we thought we
were controlling GET policy on our api endpoints.

This restores that access as we broke browser-based tracing.

Fixes #2486

Thanks @JonasVerhofste @trustin for the help!

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
@@ -92,7 +92,7 @@ public class ZipkinServerConfiguration implements WebMvcConfigurer {
       //
       // The reason is that our former CORS implementation accidentally allowed POST. People doing
       // browser-based tracing relied on this, so we can't remove it by default. In the future, we
-      // could split the collector's CORS policy into a different property, still allowing POST by
+      // could split the collector's CORS policy into a different property, still allowing POST
       // with content-type by default.
       .allowRequestMethods(HttpMethod.GET, HttpMethod.POST)
       .allowRequestHeaders(HttpHeaderNames.CONTENT_TYPE);
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
@@ -14,6 +14,7 @@
 package zipkin2.server.internal;
 
 import brave.Tracing;
+import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.common.HttpMethod;
 import com.linecorp.armeria.server.RedirectService;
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
I verified this by building off the branch locally then running the angular example on FF, chrome and safari. cc @ewhauser 

## Building locally

To build and run the server from the currently checked out source, enter the following.
```bash
$ ./mvnw -Dlicense.skip=true -DskipTests --also-make -pl zipkin-server clean install
# Run the server
$ java -jar ./zipkin-server/target/zipkin-server-*exec.jar
```

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
```diff
@@ -86,8 +87,15 @@ public void addViewControllers(ViewControllerRegistry registry) {
   /** Configures the server at the last because of the specified {@link Order} annotation. */
   @Order @Bean ArmeriaServerConfigurator corsConfigurator(
     @Value("${zipkin.query.allowed-origins:*}") String allowedOrigins) {
-    CorsServiceBuilder corsBuilder =
-      CorsServiceBuilder.forOrigins(allowedOrigins.split(",")).allowRequestMethods(HttpMethod.GET);
+    CorsServiceBuilder corsBuilder = CorsServiceBuilder.forOrigins(allowedOrigins.split(","))
+      // NOTE: The property says query, and the UI does not use POST, but we allow POST?
+      //
+      // The reason is that our former CORS implementation accidentally allowed POST. People doing
```
codefromthecrypt, line 93: @drolando this is one of the gotchas that probably works in your favor when thinking about having the UI post-back. I didn't realize we accidentally allowed POST anyway!

chlab:
@adriancole thanks for fixing this. I'm coming from https://github.com/ewhauser/angular-tracing/issues/11 and just ran into that same problem. How do I update my docker image to get this fix? Or will this be released at some later date?

codefromthecrypt:
@chlab this isn't released yet, but we will release something in the next few days.

codefromthecrypt:
@chlab an option for you now is to "java -jar" the -exec jar here. http://oss.jfrog.org/oss-snapshot-local/io/zipkin/java/zipkin-server/2.12.8-SNAPSHOT/

note.. don't try until +45m from now the change just merge hasn't published, yet

PR #2497, opened 2019-04-12
Adds workaround for gzip compression regression

### LINKED ISSUE #2495: Regression: No gzip for search results in Armeria server
After moving to Armeria gzip was disabled with no way of enabling it back

* Gzip compression in Armeria has to be enabled explicitly, looks like it's not enabled yet
* JSON is highly compressible, sending 1MB responses over the network is better than 10MB
* Documentation points to `zipkin.yml`, which has documented compression options that are ignored. They seemingly configure Spring server, which is not covered when Armeria is used

Proposed fix:
```java
@Configuration public class SpringConfiguration {
    @Bean @Order(0) ArmeriaServerConfigurator httpResponseGzipCompressionConfigurator() {
        return serverBuilder -> serverBuilder.decorator(
                s -> new HttpEncodingService(s,
                        mt -> mt.belongsTo(MediaType.ANY_APPLICATION_TYPE) || mt.belongsTo(MediaType.ANY_TEXT_TYPE),
                        1024));
    }

}
```
Note: This does not allow one to configure it, we probably want to reuse "compression" parameters in `Zipkin.yml`

_Reproducible on 12.5.8 version_

codefromthecrypt:
It is costly to ship uncompressed json to browsers. This puts in a
stop-gap.

Thanks @timothybasanov for reporting

FYI @anuraaga @hyangtack we will remove this if addressed upstream

Fixes #2495

### DIFF
--- a/zipkin-autoconfigure/ui/src/main/java/zipkin2/autoconfigure/ui/CompressionProperties.java
+++ b/zipkin-autoconfigure/ui/src/main/java/zipkin2/autoconfigure/ui/CompressionProperties.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2015-2019 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin2.autoconfigure.ui;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.web.server.Compression;
+
+@ConfigurationProperties("server")
+class CompressionProperties {
+  public Compression getCompression() {
+    return compression;
+  }
+
+  public void setCompression(Compression compression) {
+    this.compression = compression;
+  }
+
+  private Compression compression;
+}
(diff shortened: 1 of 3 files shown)


zipkin-autoconfigure/ui/src/main/java/zipkin2/autoconfigure/ui/ZipkinUiAutoConfiguration.java
```diff
@@ -133,36 +142,44 @@ String processedIndexHtml(Resource indexHtml) {
...
-      .service("/zipkin/index.html", indexSwitchingService)
+        .asService());
+
+      sb.serviceUnder("/zipkin/", uiFileService);
 
       // TODO This approach requires maintenance when new UI routes are added. Change to the following:
       // If the path is a a file w/an extension, treat normally.
       // Otherwise instead of returning 404, forward to the index.
       // See https://github.com/twitter/finatra/blob/458c6b639c3afb4e29873d123125eeeb2b02e2cd/http/src/main/scala/com/twitter/finatra/http/response/ResponseBuilder.scala#L321
-      .service("/zipkin/", indexSwitchingService)
```
codefromthecrypt, line 157: don't worry about the red.. I was just re-organizing as we no longer have to fluent everything

zipkin-autoconfigure/ui/src/main/java/zipkin2/autoconfigure/ui/CompressionProperties.java
```diff
@@ -0,0 +1,30 @@
...
+ * the License.
+ */
+package zipkin2.autoconfigure.ui;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.web.server.Compression;
+
+@ConfigurationProperties("server")
+class CompressionProperties {
+  public Compression getCompression() {
```
anuraaga, line 21: Not a big deal but I presume spring already exposes a bean with this

codefromthecrypt: surprisingly no.. I tried :)

codefromthecrypt:
verified the browser moves less bytes now..

Screen Shot 2019-04-12 at 5 34 33 PM

codefromthecrypt:
cutting this shortly as I need to release a guava fix anyway

PR #2499, opened 2019-04-14
Adds UI for remoteServiceName

### LINKED ISSUE #1794: Decide what to do about remote service name and search

codefromthecrypt:
This needs some help from a @openzipkin/ui person as the layout in classic isn't working. I do think we need classic to work as we have an undetermined amount of time until we turn it off.

This is using the two smartthings test traces, which both include the same (service, remote service) pair.

Screen Shot 2019-04-14 at 3 41 26 PM
Screen Shot 2019-04-14 at 3 42 59 PM

Fixes #1794


codefromthecrypt:
@zeagord @tacigar can one of you help me with the styling of old ui? feel free to add commit to this branch

zipkin-lens/src/util/global-search.js
```diff
@@ -12,6 +12,7 @@ export const lookbackDurations = {
 
 export const orderedConditionKeyList = autocompleteKeys => ([
   'serviceName',
+  'remoteServiceName',
```
tacigar: `orderedConditionKeyList` determines the order of condition badges.
In other words, this defines the priority of the condition display.
Do users frequently search traces using `remoteServiceName` as search criteria?
If not, I think this search condition should be placed a bit further down.

tacigar: Many users are used to `select serviceName → select spanName` flow.
So if the `remoteServiceName` is in the current position, they may be a little confused...

codefromthecrypt: I guess we don't know how frequently this will be used. I can move it down and people can ask for it to be moved up if they want :)

tacigar: In my opinion, to avoid confusion `remoteServiceName` should be after `spanName` or after `maxDuration` at first.

codefromthecrypt:
need some help on the layout in zipkin classic before this can be merged

codefromthecrypt:
thanks for the css fix @tacigar  all that's needed now is for me to kindly handle lack of the remote services endpoint in classic

zipkin-lens/src/util/global-search.js
codefromthecrypt: roger

codefromthecrypt: actually I put it last as not sure how often people will use this

codefromthecrypt:
lens gracefully handles absence of remoteServiceName (404). checking into zipkin-ui now

codefromthecrypt:
done!

codefromthecrypt:
thanks for the help @tacigar

PR #2505, opened 2019-04-18
Matches path pattern of classic UI

codefromthecrypt:
The current UI is anchored to trailing slash and if requests go to
zipkin without a trailing slash, there will be a redirect. This
redirect loses the query parameters of the search. Rather than write
a custom redirect handler that carries-over query parameters on
redirect, this matches the expected path and thus obviates a redirect
in the first place.


anuraaga:
By the way, I filed https://github.com/line/armeria/issues/1725 since I think it's good for the redirect to make it easy or default to preserving query

tacigar (approved):
LGTM!

PR #2518, opened 2019-04-19
Show IDs in lens

### LINKED ISSUE #2507: Support "Show IDs" in lens
Right now, classic ui has the ability to show the span IDs without looking at the json. Because this has been requested many times, we should support that use case here even if the UI elements are different.

Screen Shot 2019-04-18 at 4 12 45 PM

most recently discussed in #1947

tacigar:
showids

Closes #2507

### DIFF
--- a/zipkin-lens/scss/components/_timeline-span-data.scss
+++ b/zipkin-lens/scss/components/_timeline-span-data.scss
@@ -52,18 +52,19 @@
 
 .timeline-span-data__ids {
   display: flex;
-  justify-content: flex-end;
 }
 
 .timeline-span-data__id-data {
   display: flex;
-  margin: 5px 8px 0 8px;
+  margin: 5px 8px 0 3px;
+  font-size: 11px;
 }
 
 .timeline-span-data__id-data-key {
   color: $gray-5;
+  font-weight: lighter;
 }
 
 .timeline-span-data__id-data-value {
-  font-weight: bold;
+  color: $gray-4;
 }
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
can you update the image?

tacigar:
Sorry, I don't know what should I update 🤔

codefromthecrypt:
oh when asked we thought maybe best to always show ids in light font (not bold) instead of clicking button

jcchavezs:
Can I ask for the reason of displaying it inline instead of a list? Not a
UI expert but feels not natural to me (personal opinion).

fre. 19. apr. 2019, 10:37 skrev Adrian Cole <notifications@github.com>:

codefromthecrypt:
about same line is less space taken. in the new ui we can compare spans
with another but real estate is a problem. one id per line makes this worse
especially as almost no one will want the parent id

jcchavezs:
:+1: makes totally sense.

fre. 19. apr. 2019, 10:51 skrev Adrian Cole <notifications@github.com>:

tacigar:
Ah, I see.
Sorry, I thought that you said only we should use light font (Above image uses light font already).

As you said, id info are unnecessary for many users, so I think it is not bad to hide them by default.
What do you think?

codefromthecrypt:
I meant to say no boxing only light font. no outline just very light font so looks almost invisible and not distracting. personally I think the "show ids" button is distracting vs ids mentioned in a light font with no boxing or bolding... single same font for name of id and value of it

tacigar:
Ah, I got it.
I'll do so.

tacigar:
Update!

update

codefromthecrypt (approved):
love

shakuzen (approved):
The screenshot looks good to me. It takes up just as much space as having an expand button and less space than when that would be expanded.

PR #2521, opened 2019-04-20
Fixes edge case in dependency resolution when ./mvnw test is used

codefromthecrypt:
Detailed notes in the comments. Basically, there's some malpractice
that occurs when we use a test-jar instead of a main artifact named
`-tests`. This works around the dependency resolution bug until we
refactor that (likely before our first apache release).

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -90,6 +90,13 @@
           </execution>
         </executions>
       </plugin>
+      <!-- The benchmark test suite is not a valid dependency to others -->
+      <plugin>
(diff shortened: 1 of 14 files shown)


codefromthecrypt:
technically, only the collector modules would trip this as they use the codec. I plan to leave this uniform anyway to avoid land mines. by next week I will address the root problem and make zipkin-tests jar like we have in brave

codefromthecrypt:
https://github.com/apache/incubator-zipkin/pull/2517 now builds, so merging

PR #2526, opened 2019-04-22
Add unit tests for Browser component

tacigar:
**Before**

Before

**After**

After


zipkin-lens/src/components/Browser/BrowserHeader.test.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import BrowserHeader from './BrowserHeader';
+import { sortingMethods } from './sorting';
+
+describe('<BrowserHeader />', () => {
+  it('should render correct the number of results', () => {
```
codefromthecrypt: I am very happy you put in a "toehold test" like mentioned in @swankjesse's blog https://publicobject.com/2019/02/17/toehold-test/

this means later, we can test edge cases like zero results. Starting the first test class reduces burden of later change!

zipkin-lens/src/components/Browser/BrowserResults.test.js
```diff
@@ -0,0 +1,48 @@
...
+    width: 1,
+  };
+  const defaultProps = {
+    traceSummaries: [
+      { traceId: '1', duration: 5, ...traceSummaryCommonProps },
+      { traceId: '2', duration: 3, ...traceSummaryCommonProps },
+      { traceId: '3', duration: 7, ...traceSummaryCommonProps },
+    ],
+    sortingMethod: sortingMethods.LONGEST,
+    skewCorrectedTracesMap: {
```
codefromthecrypt: this highlights to me if "skewCorrectedTracesMap" is a great name or not. clock skew is only one type of adjustment/correction. It might be helped to call this plainly tracesMap and differentiate something uncorrected as "raw traces"

zipkin-lens/src/components/Browser/sorting.test.js
```diff
@@ -0,0 +1,61 @@
+import { sortingMethods, sortTraceSummaries } from './sorting';
+
+const numberArrayToTraceSummaries = (numArray, key) => numArray.map(
+  n => ({ [key]: n }),
+);
+
+describe('sortTraceSummaries', () => {
+  it('LONGEST', () => {
```
codefromthecrypt: nit: for a test like this, sometimes I like to use tuples. for example (timestamp, duration). it is easier to see when there's a discrepancy between order when both values are literals also.

ex.

input = [
 (2, 1),
 (3, 2),
 (1, 3)
]
expect(sortedByTimestamp(input)).toEqual(
 (1, 3),
 (2, 1),
 (3, 2)
)

codefromthecrypt (approved):
good progress. I made some style comments which may not be relevant

zipkin-lens/src/components/Browser/BrowserResults.test.js
tacigar: I see. I'll rename them 👍

zipkin-lens/src/components/Browser/sorting.test.js
tacigar: I see. Thank you for your advice.

tacigar:
PTAL 🙇

codefromthecrypt (approved):
great

PR #2527, opened 2019-04-22
Add unit tests for TraceSummary component

tacigar:
**Before**

Before

**After**

After


codefromthecrypt:
test fail is due to timezone. if you have a way to lock timezone for test, use GMT

tacigar:
I'll remove the test for now, because it is a bit hard for me to test formatted date...

shakuzen:
> I'll remove the test for now, because it is a bit hard for me to test formatted date...

Do we want to open an issue to track testing this or is it not something we should test?

codefromthecrypt:
even a TODO in the lines that would be covered by the test could help avoid
forgetting

tacigar:
Sorry I added tests.

codefromthecrypt:
thanks!

PR #2532, opened 2019-04-23
Replaces fragile proto test setup with Square Wire

### LINKED ISSUE #2530: Skip protoc execution when tests are skipped
We tried to use Jitpack to test a build of the Zipkin Server from a pull request (#2328), but the build failed with the following.

https://jitpack.io/com/github/ewhauser/zipkin/experimental-grpc-1.5.0-g7a0c3ce-1160/build.log

```
[INFO] --- protobuf-maven-plugin:0.6.1:test-compile (test-compile-proto) @ zipkin-server ---
[INFO] Compiling 1 proto file(s) to /home/jitpack/build/zipkin-server/target/generated-test-sources/protobuf/java
[ERROR] PROTOC FAILED: /home/jitpack/build/zipkin-server/target/protoc-plugins/protoc-3.7.0-linux-x86_64.exe: error while loading shared libraries: libatomic.so.1: cannot open shared object file: No such file or directory

[ERROR] /home/jitpack/build/zipkin-server/target/test/proto/zipkin.proto [0:0]: /home/jitpack/build/zipkin-server/target/protoc-plugins/protoc-3.7.0-linux-x86_64.exe: error while loading shared libraries: libatomic.so.1: cannot open shared object file: No such file or directory
```

Since this protoc execution is only for tests and the build is skipping tests, we should be able to disable it when tests are skipped so Jitpack succeeds. This can probably be done in a Maven profile.
Alternatively, we can customize the command Jitpack uses to build, but I do like it working out-of-the-box (as it did before the pull request added GRPC collector support) and I wonder if Zipkin developers wouldn't also run into this problem on their machines.
/cc @bsideup

codefromthecrypt:
Square Wire is a lot simpler setup, and it doesn't require installation
of binaries to compile proto files. This helps us and also anyone who
will be doing ASF source verification.

Fixes #2530

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -45,6 +45,11 @@
       <artifactId>guava</artifactId>
       <version>${guava.version}</version>
     </dependency>
+    <!-- for codec benchmarks -->
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+    </dependency>
 
     <dependency>
       <groupId>com.esotericsoftware.kryo</groupId>
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -69,19 +69,26 @@
       <artifactId>zipkin-server</artifactId>
       <version>${project.version}</version>
     </dependency>
-
+    <dependency>
+      <groupId>com.squareup.wire</groupId>
+      <artifactId>wire-runtime</artifactId>
+      <version>${wire.version}</version>
+    </dependency>
     <dependency>
       <groupId>io.zipkin.proto3</groupId>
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
thanks to @jessewilson who reminded me that square wire does proto3 and grpc services now.

Ex our grpc service could be called like this in kotlin when #2328 is merged

```kotlin
interface SpanService : Service {
  @WireRpc(
      path = "/zipkin.proto3.SpanService/Report",
      requestAdapter = "zipkin.proto3.ListOfSpans#ADAPTER",
      responseAdapter = "zipkin.proto3.ReportResponse#ADAPTER"
  )
suspend fun Report(request: ListOfSpans): ReportResponse
```

PR #2533, opened 2019-04-24
Moves to apache layout factory dist

codefromthecrypt:
pending official release see https://github.com/apache/incubator-zipkin-layout-factory/pull/9


codefromthecrypt:
will add a commit to use Apache snapshot repo. after that works I think @jeqo can start release

codefromthecrypt:
LOOKS GOOD

tested manually both collectors

codefromthecrypt:
now with formal release

Issue #2535, opened 2019-04-24
Allow filtering the displayed services in dependency graph


shakuzen:
Feedback from LINE's SRE team was that when you have many services, the service graph becomes practically unusable. The ability to filter the displayed services would go a long way in helping this. Users usually want to see the dependencies for their service in particular, or perhaps all subsequent dependencies from an edge service.

Jason from Haystack mentioned that the vizceral library has filtering capabilities built-in, so we could easily leverage that to display a filtered dependencies graph.

bulicekj:
Filtering without recreating the graph is built into Vizceral, and I believe Lens already has that (the drop down in the top right) but it only greys out nodes not connected with the highlighted one. What you are suggesting we do in Haystack, but unfortunately not via built-in a method. Given a serviceName, [we simply filter edge source or destination to include the serviceName](https://github.com/ExpediaDotCom/haystack-ui/blob/master/src/components/serviceGraph/serviceGraphContainer.jsx#L153) and return the new graph. It should still be a fairly easy addition - I would keep the dropdown you guys have and simply do this graph recreation instead of the current highlighting function that it does.

bulicekj:
maybe something like above ^

PR #2537, opened 2019-04-24
dependency graph filtering

bulicekj:
Changing the dropdown select on the dependency graph filtering to recreate the graph with one level up and down from the selected service (#2535)


codefromthecrypt:
thanks! gotta screen shot?

codefromthecrypt:
fyi this is likely delaying review due to "golden week" in japan

tacigar:
Sorry for delaying review 😭 
Thank you for your great work!

tacigar:
Linter has issued a warning.
Could you fix them?

tacigar:
There are some lint errors.
Would you fix them? 🙇

zipkin-lens/src/components/Dependencies/DependenciesGraph.js
```diff
@@ -71,8 +79,8 @@ class DependenciesGraph extends React.Component {
             name: 'dependencies-graph',
             updated: new Date().getTime(),
             maxVolume: maxVolume * 2000,
-            nodes: graph.allNodes(),
-            connections: graph.allEdges(),
+            nodes,
+            connections
           }}
           objectHighlighted={this.handleObjectHighlighted}
           match={filter}
```
tacigar: `match` prop is no longer needed.
Would you remove this line?

bulicekj:
Before: 
Screen Shot 2019-04-29 at 11 09 02 AM
After:
Screen Shot 2019-04-29 at 11 16 36 AM


FYI - there's an bug with Vizceral (the dependency graph framework) stacking nodes when you switch to a service that isn't present on the graph. For example, selecting `frontend` after previously selecting `frontendtwo` like the above screenshot will cause this to happen: 
Screen Shot 2019-04-29 at 11 15 32 AM

This should probably not be merged until we can figure out how to prevent nodes from stacking.

zipkin-lens/src/components/Dependencies/DependenciesGraph.js
bulicekj: Good call 👍

tacigar (approved):

tacigar:
Thank you so much!
LGTM 👍 
But we need to investigate the above issue a little more before merging

codefromthecrypt:
any news on the stacking front (or lack of stacking front)?

bulicekj:
I spent some time trying to change the way the graph is re-rendered upon the state change, but no success thus far. @tacigar any ideas?

tacigar:
Ummm...
I tried to re-render, but I could not do...
Maybe it is better to use **Vizceral** directly instead of **Vizceral-react**.
If do so, we can handle more.

codefromthecrypt:
naver pinpoint has an upstream and downstream depth control to their filter. for example, if you select a service you can show one link up and down by default, but you can change that to show 2 links down. I don't think their data model is precise enough to truly filter that the downstream were caused by the selected service though. I think it doesn't

zeagord (approved):

bulicekj:
@tacigar @adriancole I was playing around today and found a solution that fixes the above mentioned bug. If we completely un-mount and re-mount VizceralExt, it re-renders with the correct layout -> 
zipkin-vizceral-2 

It's not the most elegant solution, as it hacks thorough some of the react re-rendering magic, but fixes the "stacking" issue. Let me know what you think

bulicekj:
Also worth a note, I tried playing around with [vizceral's built in filtering](https://github.com/Netflix/Vizceral/wiki/Configuration#filters) and got it to work with one state change but couldn't get it to filter again after, even though [state should update](https://github.com/Netflix/vizceral-react/blob/b40765c9e22cba4b363e74ead011b578a8e65935/src/vizceral.jsx#L77) 🤔

tacigar:
Wow...wonderful hack...

tacigar (approved):

tacigar:
Thank you for your great work!!

codefromthecrypt:
@tacigar which merges first? this or the other?

tacigar:
Please merge this first!
I think @bulicekj 's great work should be on the commit history!

PR #2538, opened 2019-04-26
Tolerates empty POST requests with debug logging

codefromthecrypt:
This backfills tests to show that our normal codec library is ok with
empty data. What's different, is this doesn't fail on POST when there
are empty messages. Instead, if debug logging is enabled, we say which
IP and user agent posted it.

See https://github.com/apache/incubator-zipkin-reporter-java/issues/138

### DIFF
--- a/zipkin-server/src/test/kotlin/zipkin2/server/internal/ITZipkinHttpCollector.kt
+++ b/zipkin-server/src/test/kotlin/zipkin2/server/internal/ITZipkinHttpCollector.kt
@@ -1,0 +1,230 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package zipkin2.server.internal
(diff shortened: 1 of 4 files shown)


zipkin-server/src/test/kotlin/zipkin2/server/internal/ITZipkinHttpCollector.kt
```diff
@@ -0,0 +1,230 @@
...
+    val message = SpanBytesEncoder.JSON_V2.encodeList(asList(
+      Span.newBuilder().traceId("1").id("1").name("test").build()
+    ))
+
+    assertThat(post("/api/v1/spans", message).code()).isEqualTo(202)
+    assertThat(post("/api/v2/spans", message).code()).isEqualTo(202)
+  }
+
+  @Test @Throws(IOException::class)
+  fun emptyIsOk() {
```
codefromthecrypt, line 112: this is the important change here.

codefromthecrypt:
here's example log when debug is turned on
```
2019-04-26 12:34:48.782 DEBUG 33554 --- [-worker-nio-2-1] z.s.i.UnzippingBytesRequestConverter     : Empty POST body sent by user-agent->curl/7.54.0, x-forwarded-for->null
```

here's the request that produced an empty trace. You can see from the client POV it is ok
```bash
$ curl -s localhost:9411/api/v2/spans -X POST -v
*   Trying ::1...
* TCP_NODELAY set
> POST /api/v2/spans HTTP/1.1
> Host: localhost:9411
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 202 Accepted
< content-type: text/plain; charset=utf-8
< content-length: 12
< 
202 Accepted
```

codefromthecrypt:
revised the log output a bit

```
2019-04-26 12:45:30.890 DEBUG 36360 --- [-worker-nio-2-2] z.s.i.ZipkinHttpCollector                : Empty POST body sent by clientAddress->/0:0:0:0:0:0:0:1, userAgent->curl/7.54.0
2019-04-26 12:45:51.850 DEBUG 36360 --- [-worker-nio-2-3] z.s.i.ZipkinHttpCollector                : Malformed gzip body sent by clientAddress->/0:0:0:0:0:0:0:1, userAgent->curl/7.54.0
```

PR #2541, opened 2019-04-26
Adds license header to DDL

codefromthecrypt:


codefromthecrypt:
oops this was a bad idea :) we can't license header CQL files :P

codefromthecrypt:
actually looks like it was just bad naming convention.. let's see

codefromthecrypt:
yeah the license header thing isn't working even if files are valid. I will skip it for CQL

PR #2546, opened 2019-04-27
Extracts zipkin-tests jar before we publish to o.a.z.zipkin2 namespace

codefromthecrypt:
This fixes an anti-pattern before we commit to it in the new namespace.


codefromthecrypt:
the following release command failed before I made this change.. not sure if this is the only change left or not..

```bash
$ VERSION=2.13.0 NEXT_VERSION=2.13.1-SNAPSHOT GPG_TTY=$(tty) ./mvnw release:prepare -Papache-release  --batch-mode -nsu -DreleaseVersion=$VERSION -Dtag=v$VERSION -DdevelopmentVersion=$NEXT_VERSION -Darguments="-Dmaven.test.skip=true" -DpushChanges=false -DautoVersionSubmodules=true
```

codefromthecrypt:
yep the release command works after this.

PR #2552, opened 2019-05-01
Fixes logging and metrics for collectors

codefromthecrypt:
Before, we were inconsistent about incrementing metrics, and in worst
case, we double-incremented. This adds tests for the messaging based
collectors and backfills tests to make the base Collector class 100%
coverage.

Separately, we have recently changed our web endpoint to produce its
own error messages to the client. In other words, we no longer rely on
(often unnecessary) wrapping of exceptions to correct the message being
sent back. This changes the logic around message formatting so it is
only used when logging is enabled.

### DIFF
--- a/zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
+++ b/zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.function.Supplier;
 import java.util.logging.Logger;
 import zipkin2.Callback;
 import zipkin2.Span;
@@ -29,6 +30,7 @@ import zipkin2.storage.StorageComponent;
 
 import static java.lang.String.format;
 import static java.util.logging.Level.FINE;
+import static zipkin2.Call.propagateIfFatal;
 
 /**
  * This component takes action on spans received from a transport. This includes deserializing,
@@ -39,6 +41,13 @@ import static java.util.logging.Level.FINE;
  * threads.
  */
 public class Collector { // not final for mock
+  static final Callback<Void> NOOP_CALLBACK = new Callback<Void>() {
+    @Override public void onSuccess(Void value) {
+    }
+
+    @Override public void onError(Throwable t) {
+    }
+  };
 
   /** Needed to scope this to the correct logging category */
   public static Builder newBuilder(Class<?> loggingClass) {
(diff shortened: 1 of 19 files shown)


codefromthecrypt:
fyi @Logic-32 this will make your pull request drift. I am ok helping rebase if needed.

zipkin-collector/scribe/src/main/java/zipkin2/collector/scribe/ScribeSpanConsumer.java
```diff
@@ -42,22 +43,23 @@
...
+    List<Span> spans = new ArrayList<>();
+    int byteCount = 0;
     try {
-      spans =
-          messages
-              .stream()
-              .filter(m -> m.category.equals(category))
-              .map(m -> m.message.getBytes(StandardCharsets.ISO_8859_1))
-              .map(b -> Base64.getMimeDecoder().decode(b)) // finagle-zipkin uses mime encoding
-              .peek(b -> metrics.incrementBytes(b.length))
```
codefromthecrypt, line 55: this logic was plain wrong

zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
```diff
@@ -110,34 +119,46 @@ public void accept(List<Span> spans, Callback<Void> callback) {
...
 
     try {
+      // adding a separate callback intentionally decouples collection from storage
       record(sampled, acceptSpansCallback(sampled));
       callback.onSuccess(null);
-    } catch (RuntimeException e) {
-      callback.onError(errorStoringSpans(sampled, e));
-      return;
+    } catch (RuntimeException | Error e) {
+      handleStorageError(spans, e, callback);
```
Logic-32, line 133: Technically speaking, isn't this still a "Collector error" at this point?  Since `record()` is async, the "storage error" doesn't happen until later.  Arguably, we may want to handle them the same way but I wanted to bring up the potential issue just in case.

codefromthecrypt: thanks for asking. so the most likely cause at this point is in the "record" method, which I suppose we should rename to store. There's always a chance that there's an error in the storage implementation prior to returning a call. For example, in Elasticsearch a failure to install the index template would propagate here.

```
  void record(List<Span> sampled, Callback<Void> callback) {
    storage.spanConsumer().accept(sampled).enqueue(callback);
  }
```

Logic-32: Interesting point!  Would it be worthwhile to separate the `...accept()` call from `enqueue()` for the sake of error messaging?  I'm not exactly sold on the difference being worthwhile at the moment.

codefromthecrypt: added a commit on this topic. good question

codefromthecrypt: sorry wires crossed. So, In any case I think we'll still need to invoke either callback.onSuccess or onError.

So I guess you are asking if the code added to separate out which of the failures into separate log prefixes would be worth it..

ex instead of just "storage error" a separate if statement and log message for `storage.spanConsumer()` vs `spanConsumer.accept()` vs `call.enqueue()`.

I can understand that in the logs you might not have the stack trace (though the caller may), so this could be a way to know the difference, if the message wasn't already clear what happened..

codefromthecrypt: sorry it would be like instead of

"Cannot store spans ... due to Error installing ES template" (or whatever the cause would be)
"Cannot get instance of span consumer ...
or
"Cannot create storage call to accept spans ...
or
"Cannot enqueue spans for storage...

I think classifying like this would add a dozen lines of main and 3 or 4 more test cases, but thinking right now, I'm not sure it would be terribly more helpful.

zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
```diff
@@ -178,60 +198,56 @@ void warn(String message, Throwable e) {
...
-        && (error.startsWith("Malformed") || error.startsWith("Truncated"))) {
-      if (shouldWarn()) warn(error, e);
-      return (RuntimeException) e;
-    } else {
-      if (shouldWarn()) {
-        message = format("%s due to %s(%s)", message, e.getClass().getSimpleName(), error);
-        warn(message, e);
-      }
-      return new RuntimeException(message, e);
+    // We have specific code that customizes log messages. Use this when the case.
```
codefromthecrypt, line 238: PS I really don't think this special casing is helpful anymore

codefromthecrypt:
no rush on this PR except that it will stumble @Logic-32's one. the release cut is happening (a good thing)

Logic-32:
I'd actually like to see this in before #2502 if possible.  You're welcome to rebase it in if you'd like.  If not, I'll take care of it when I can.

PR #2558, opened 2019-05-06
Removes support for Elasticsearch 2.x

codefromthecrypt:
As discussed, this removes support for Elasticsearch 2.x

See https://lists.apache.org/thread.html/ac01ebd12b6ecaac6496e146570652a3e804d45d9b9d2100001f5c75@%3Cdev.zipkin.apache.org%3E


codefromthecrypt:
cc @xeraa @jcarres-mdsol

codefromthecrypt:
as a side effect, this could make travis work again. When we merged ES 7, we broke travis' 50m hard timeout

xeraa:
I‘m afraid this is the sanest choice and it accurately describes our support policy (we‘ll probably provide some more guidance around that soon).

codefromthecrypt:
travis may still be unhappy after this change (regarding the 50m limit). Any further efforts will be in a different PR.

PR #2563, opened 2019-05-07
Add a netty scribe server.

### LINKED ISSUE #2461: re-implement scribe over armeria
considering armeria was originally a thrift implementation.. I'm wondering if we can't replace our old swift variant of scribe with armeria?

https://github.com/openzipkin/zipkin/tree/master/zipkin-collector/scribe

cc @openzipkin/armeria

anuraaga:
This ports the scribe functionality to use the armeria library underneath. This reduces the size dependency liability as the former implementation was based on an abandoned library (facebook swift).

Here's how to get a build off this branch from jitpack
```bash
$ curl -sSL https://jitpack.io/com/github/anuraaga/zipkin/zipkin-server/netty-scribe-SNAPSHOT/zipkin-server-netty-scribe-SNAPSHOT-exec.jar > zipkin.jar
$ curl -sSL https://jitpack.io/com/github/anuraaga/zipkin/zipkin-autoconfigure-collector-scribe/netty-scribe-SNAPSHOT/zipkin-autoconfigure-collector-scribe-netty-scribe-SNAPSHOT-module.jar > scribe.jar
$ SCRIBE_ENABLED=true \
    java \
    -Dloader.path='scribe.jar,scribe.jar!/lib' \
    -Dspring.profiles.active=scribe \
    -cp zipkin.jar \
    org.springframework.boot.loader.PropertiesLauncher
```

Fixes #2461

This adds a TCP server which reads thrift framed transport (https://github.com/facebookarchive/scribe/wiki/Logging-Messages indicates scribe uses thrift framed transport) and creates synthetic HTTP requests with the frames and serves them with armeria's thrift service.

/cc @trustin for any thoughts on this approach


codefromthecrypt:
I am excited

codefromthecrypt:
taking a look now thanks!

anuraaga:
For reference, the IDL is at https://github.com/facebookarchive/scribe/blob/master/if/scribe.thrift

It'd be a pain to have to wire up the thrift compiler to the build :(

codefromthecrypt:
ack. the types are very simple though.. it is possible that we temporarily use thrift compiler, then inline it.

codefromthecrypt:
ps here is the old thrift def https://github.com/openzipkin-attic/zipkin-csharp/blob/master/tools/thrift/scribe.thrift

codefromthecrypt:
ps almost integrated. will push a commit

codefromthecrypt:
ok I ported the impl but not sure what I'm doing :) the ScribeTest fails because the state checks used aren't valid anymore or maybe the server isn't started yet

anuraaga:
Sweet thanks! I should be able to finish it up now.

Since there is an `isRunning`, I wasn't sure if `start` is supposed to block or not, it's easy to make it do so with `syncUninterruptibly()`. Will play with it and the test.

codefromthecrypt:
yeah it doesn't have to block, but we should have a way in the test to know it has started. I think there are no other listeners so this point is not particularly common. The old one blocked I think..

anuraaga:
Ok think it works.

@trustin if possible, would be great for you to review since you're much better at Netty than me :)

zipkin-collector/scribe/src/main/java/zipkin2/collector/scribe/NettyScribeServer.java
```diff
@@ -0,0 +1,76 @@
...
+    ServerBootstrap b = new ServerBootstrap();
+    channel = b.group(bossGroup, workerGroup)
+      .channel(EventLoopGroups.serverChannelType(bossGroup))
+      .childHandler(new ChannelInitializer<SocketChannel>() {
+        @Override
+        protected void initChannel(SocketChannel ch) {
+          ch.pipeline().addLast(new ScribeInboundHandler(scribe));
+        }
+      })
+      .childOption(ChannelOption.AUTO_READ, true)
```
trustin: AUTO_READ is on by default, i.e. no need to set.

zipkin-collector/scribe/src/main/java/zipkin2/collector/scribe/NettyScribeServer.java
```diff
@@ -0,0 +1,76 @@
...
+      .childOption(ChannelOption.SO_KEEPALIVE, true)
```
trustin: As we chat at Slack, this is practically meaningless.

zipkin-collector/scribe/src/main/java/zipkin2/collector/scribe/NettyScribeServer.java
```diff
@@ -0,0 +1,76 @@
...
+      .bind(port)
+      .syncUninterruptibly()
```
trustin: You might want to catch an exception and handle/rethrow it. One surprising behavior of `sync()` and `syncUninterruptibly()` is that it can throw a checked exception even though it doesn't have the `throws Exception` clause.

zipkin-collector/scribe/src/main/java/zipkin2/collector/scribe/ScribeInboundHandler.java
```diff
@@ -0,0 +1,187 @@
...
+      logger.warn("Unexpected exception servicing thrift request. "
+        + "This usually indicates a bug in armeria's thrift implementation.", e);
+      exceptionCaught(ctx, e);
+      return;
+    }
+
+    response.aggregateWithPooledObjects(ctx.executor(), ctx.alloc()).handle((msg, t) -> {
+      if (t != null) {
+        logger.warn("Unexpected exception reading thrift response. "
+          + "This usually indicates a bug in armeria's thrift implementation.", t);
```
trustin: Hmm, really? I think aggregation may fail on other cases such as OOME.

zipkin-collector/scribe/src/main/java/zipkin2/collector/scribe/ScribeInboundHandler.java
```diff
@@ -0,0 +1,187 @@
...
+    maybeReadPayload(ctx);
+  }
+
+  void maybeReadPayload(ChannelHandlerContext ctx) {
+    if (pending.readableBytes() < nextFrameSize) {
+      return;
+    }
+
+    ByteBuf payload = ctx.alloc().buffer(nextFrameSize);
+    pending.readBytes(payload, nextFrameSize);
```
trustin, line 131: You need to call `pending.discardSomeReadBytes()` so that the `reader/writerIndex` of `pending` does not increase infinitely.

zipkin-collector/scribe/src/main/java/zipkin2/collector/scribe/NettyScribeServer.java
anuraaga: Thanks - by the way, any thoughts on removing it from this tutorial?

https://netty.io/wiki/user-guide-for-4.x.html#writing-a-discard-server

zipkin-collector/scribe/src/main/java/zipkin2/collector/scribe/ScribeInboundHandler.java
anuraaga: Ah that's true - guess the exception will probably be self-explanatory enough, removed the armeria blaming ;)

zipkin-collector/scribe/src/main/java/zipkin2/collector/scribe/NettyScribeServer.java
anuraaga: Based on the behavior of this test, which checks for the actual exception raised, I'll go ahead and leave as is

https://github.com/apache/incubator-zipkin/pull/2563/files#diff-00a19d279ad37c9a9847b8ba5680bfaa

@adriancole let me know if you think it's better to wrap the exception.

anuraaga (commented):
Thanks!

codefromthecrypt:
@anuraaga I prefer not wrapping exceptions if that's the question :)

codefromthecrypt:
updated description with jitpack instructions. testing with finagle now

codefromthecrypt:
ok I ran the https://github.com/openzipkin/zipkin-finagle-example/compare/raw-finagle with 100s of load on laptop (`wrk -t4 -c128 -d100s  http://localhost:8081 --latency`).

I received some messages like below. However, I've not checked the base case.. will post back after doing that.

on zipkin:
```
2019-05-08 16:50:29.204  WARN 68201 --- [-worker-nio-2-6] z.c.s.ScribeInboundHandler               : Unexpected error handling connection.

java.io.IOException: Connection reset by peer
	at sun.nio.ch.FileDispatcherImpl.read0(Native Method) ~[?:?]
	at sun.nio.ch.SocketDispatcher.read(SocketDispatcher.java:39) ~[?:?]
	at sun.nio.ch.IOUtil.readIntoNativeBuffer(IOUtil.java:223) ~[?:?]
	at sun.nio.ch.IOUtil.read(IOUtil.java:192) ~[?:?]
	at sun.nio.ch.SocketChannelImpl.read(SocketChannelImpl.java:382) ~[?:?]
```

on the traced application:
```
May 08, 2019 4:50:28 PM com.twitter.finagle.liveness.FailureAccrualFactory$$anon$4$$anon$5 didMarkDead
INFO: FailureAccrualFactory marking connection to "zipkin-tracer" as dead. Remote Address: Inet(localhost/127.0.0.1:9410,Map())
```

anuraaga:
Thanks - also added some performance optimizations and more consistency with the swift server.

zipkin-collector/scribe/src/main/java/zipkin2/collector/scribe/ScribeInboundHandler.java
```diff
@@ -0,0 +1,184 @@
...
+    pending.discardSomeReadBytes();
+
+    state = ReadState.HEADER;
+
+    HttpRequest request = HttpRequest.of(
+      THRIFT_HEADERS.toMutable(),
+      new ByteBufHttpData(payload, true));
+    ServiceRequestContext requestContext = ServiceRequestContextBuilder.of(request)
```
anuraaga: @trustin Can you confirm this code looks like what you'd expect for invoking a service manually?

codefromthecrypt:
FWIW the previous code somehow doesn't work as all spans are dropped sending to http somehow..

anuraaga:
What do you mean by `dropped sending to http` exactly?

codefromthecrypt:
will try this codebase again all the same :)

anuraaga:
By the way, while I would expect this server to support 128 connections no problem, I guess finagle limits to 5 if looking at performance of a single client

https://github.com/twitter/finagle/blob/develop/finagle-zipkin-scribe/src/main/scala/com/twitter/finagle/zipkin/thrift/ScribeRawZipkinTracer.scala#L59

codefromthecrypt:
@anuraaga meaning when I switched the server to the version before this.. and ran the same wrk script.. I got the following. It is possible I am somehow fat fingering, but it is odd..

```bash
$ curl -s localhost:9411/health
{"status":"UP","zipkin":{"status":"UP","details":{"InMemoryStorage":{"status":"UP"},"ScribeCollector":{"status":"UP"}}}}
{
  "counter.zipkin_collector.messages.http": 40,
  "counter.zipkin_collector.spans_dropped.http": 0,
  "gauge.zipkin_collector.message_bytes.scribe": 0,
  "counter.zipkin_collector.bytes.scribe": 0,
  "gauge.zipkin_collector.message_spans.scribe": 0,
  "counter.zipkin_collector.spans.scribe": 0,
  "counter.zipkin_collector.messages_dropped.scribe": 0,
  "gauge.zipkin_collector.message_bytes.http": 1601652,
  "counter.zipkin_collector.messages.scribe": 0,
  "counter.zipkin_collector.spans_dropped.scribe": 0,
  "counter.zipkin_collector.bytes.http": 89401054,
  "gauge.zipkin_collector.message_spans.http": 3916,
  "counter.zipkin_collector.spans.http": 257426,
}
```

anuraaga:
Ah I understand the test now - you're `wrk`ing to a server that then uses a finagle client, then it means it's using the defaults :)

zipkin-collector/scribe/src/main/java/zipkin2/collector/scribe/NettyScribeServer.java
anuraaga: @adriancole I went ahead and wrapped the exception since the exception type is different depending on the platform (e.g., Windows vs Travis) so I guess it's good to add some sanity there.

codefromthecrypt:
ignore above as I found the fat fingering. I will run against base case now.

FWIW latest commit still gets these (when using defaults in finagle on laptop)

anuraaga:
Updated to not log client connection close, same as normal armeria does. The interesting comparison will be whether base case and new case `didMarkDead` at a similar rate. If not, then I might be missing something needed to make sure finagle knows the connection's still alive.

codefromthecrypt:
in a "fair fight" I got one connection fail client side on the old vs new (several)

Is it possible that the scribe handler is happening on the I/O thread by chance?

anuraaga:
Yeah it's on the I/O thread - but the collector is asynchronous so it should be ok right?

codefromthecrypt:
ps to make this more fair can you go rebase/force push on latest master? I don't expect it to change things, but trying for apples/apples

codefromthecrypt:
also this time when I compare I will compare request count vs trace count to see if the errors are not important

anuraaga:
Merged master. I'll also try to run the wrk after getting the binaries tomorrow.

codefromthecrypt:
getting log messages like this:

```
2019-05-08 18:25:48.325  WARN 84511 --- [entloop-nio-4-1] i.n.b.ServerBootstrap                    : Unknown channel option 'SO_BACKLOG' for channel '[id: 0x975d50ac, L:/127.0.0.1:9410 - R:/127.0.0.1:63794]'
```

anuraaga:
Yeah I'll remove that option I think, it only works on Linux and probably doesn't have much effect.

Current hypothesis is I need to ensure responses go in order, which the other servers seem to implement. Going to try to repro first here with the binaries though

codefromthecrypt:
> Yeah it's on the I/O thread - but the collector is asynchronous so it
should be ok right?
the in-memory storage is not offloaded to a different thread. It is
possible that the swift listener is offloading onto another thread always..

codefromthecrypt:
I'm gonna backfill a test on master and rebase your branch over unless you scream :P

codefromthecrypt:
@anuraaga in case interested.. yeah the sequence id dance happens if that's what you are referring to. In the sender we only have one thread that does anything so we use a simple counter https://github.com/apache/incubator-zipkin-reporter-java/blob/master/libthrift/src/main/java/zipkin2/reporter/libthrift/ScribeClient.java#L49

codefromthecrypt:
PS because my test itself had a bug this PR is no longer at risk of me rebasing :P

codefromthecrypt:
ps whenever this is ready for next look, I will test again using finagle, except using our libthrift impl [1] instead of finagle's native one. This will reduce the amount of variables around socket resets.

[1] ours, but with this socket reset fix https://github.com/apache/incubator-zipkin-reporter-java/pull/143

anuraaga:
Haven't gotten to try to repro locally yet but went ahead and added the code to guarantee response order.

codefromthecrypt:
re-checking

codefromthecrypt:
over 1M messages with current code. no disconnects, which makes it better at this than the swift codebase

codefromthecrypt (approved):
fantastic

codefromthecrypt:
fuzz PR https://github.com/apache/incubator-zipkin/pull/2575

PR #2564, opened 2019-05-07
Removes Kafka 0.8 support (KAFKA_ZOOKEEPER)

codefromthecrypt:
This removes support for Kafka 0.8 (last updated almost 4 years ago).
Notably, this means those using `KAFKA_ZOOKEEPER` to configure their
broker need to switch to `KAFKA_BOOTSTRAP_SERVERS` instead.

Note: Kafka 0.8 was not packaged into zipkin-server, it was an optional
add-on. However, our docker image was created in such a way that it felt
like it was available by default.

See https://lists.apache.org/thread.html/432df5a806ee27dd959ded5ebf5e7cc6bd4370f6b1b1daf7bf594e80@%3Cdev.zipkin.apache.org%3E


codefromthecrypt:
note: this would occur in zipkin 2.14 (the next pending release is 2.13)

codefromthecrypt:
updated release notes to warn folks about this https://github.com/apache/incubator-zipkin/releases/tag/v2.13.0

PR #2570, opened 2019-05-08
Ports all zipkin-server tests to Kotlin

codefromthecrypt:
This reduces the amount of code to maintain and helps us learn Kotlin.


zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
```diff
@@ -185,6 +185,10 @@ static void maybeLog(String prefix, ServiceRequestContext ctx, AggregatedHttpMes
     }
 
     if (content.isEmpty()) ZipkinHttpCollector.maybeLog("Empty POST body", ctx, request);
+    if (content.length() == 2 && "[]".equals(content.toStringAscii())) {
```
codefromthecrypt, line 188: found a bug while doing this!!

codefromthecrypt:
PS I don't expect any reasonable human being to review 2.5K lines. I will merge this on green and take beatings as necessary.

anuraaga:
Was thinking a bit about this - I know it's after the fact, and it's not really a big deal, but what do you think about the effect on random contributions from this change? I guess people may be less motivated to contribute changes to a Java project if they have to understand a different programming language to write tests.

codefromthecrypt:
On Fri, May 10, 2019 at 1:17 PM Anuraag Agrawal <notifications@github.com>
wrote:

> Was thinking a bit about this - I know it's after the fact, and it's not
> really a big deal, but what do you think about the effect on random
> contributions from this change? I guess people may be less motivated to
> contribute changes to a Java project if they have to understand a different
> programming language to write tests.
>

Good question.. as this is the flip side of the coin. Some are interested
in learning more Kotlin (several on gitter were interested in this). We
also need more experience in Kotlin as it comes up in Brave frequently. The
only thing I can say about this is that few non-committers contribute
directly to the server anyway, and we aren't planning on changing the core
library to kotlin. Folks wanting to write tests in Java have plenty of
places to do that.

anuraaga:
Cool makes sense

PR #2571, opened 2019-05-08
Backfills functional test of Scribe

codefromthecrypt:
cc @anuraaga I think this will make it easier to tell what's going on without spinning up a server. PS you can also look at the reporter as it makes scribe messages without the thrift compiler :P


codefromthecrypt:
this test isn't good enough. our sender code had a problem in it (not sure if finagle's has a similar problem or not) https://github.com/apache/incubator-zipkin-reporter-java/pull/143

codefromthecrypt:
obviated

PR #2572, opened 2019-05-08
annotation query parsing improvements

jorgheymans:
- whitespace is trimmed from annotation query parameters
- when both an annotation and tag are specified with the same key, (ie `foo and foo=bar`) the tag now wins because it's considered to be more specific. Previously last one in the query would win.

### DIFF
--- a/zipkin/src/main/java/zipkin2/storage/QueryRequest.java
+++ b/zipkin/src/main/java/zipkin2/storage/QueryRequest.java
@@ -186,7 +186,8 @@ public final class QueryRequest {
     }
 
     /**
-     * Corresponds to query parameter "annotationQuery". Ex. "http.method=GET and error"
+     * Corresponds to query parameter "annotationQuery". Ex. "http.method=GET and error". Parameter keys and values are
+     * trimmed.
      *
      * @see QueryRequest#annotationQueryString()
      */
@@ -196,9 +197,12 @@ public final class QueryRequest {
       for (String ann : annotationQuery.split(" and ", 100)) {
         int idx = ann.indexOf('=');
         if (idx == -1) {
-          map.put(ann.trim(), "");
+          // put the annotation only if there is no key present already, prevents overriding more specific tags
+          map.putIfAbsent(ann.trim(), "");
         } else {
+          // tag
           String[] keyValue = ann.split("=", 2);
+          // tags are put regardless, i.e. last tag wins
           map.put(ann.substring(0, idx).trim(), keyValue.length < 2 ? "" : ann.substring(idx + 1).trim());
         }
       }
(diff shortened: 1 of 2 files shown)


zipkin/src/test/java/zipkin2/storage/QueryRequestTest.java
```diff
@@ -75,6 +76,26 @@
...
+    // @adriancole said this would be hard to parse, annotation containing spaces
+    assertThat(queryBuilder.parseAnnotationQuery("L O L").annotationQuery)
+      .containsOnly(entry("L O L", ""));
+    // annotation with spaces combined with tag
+    assertThat(queryBuilder.parseAnnotationQuery("L O L and http.path = /a").annotationQuery)
+      .containsOnly(entry("L O L", ""), entry("http.path", "/a"));
+    assertThat(queryBuilder.parseAnnotationQuery("bar =123 and L O L and http.path = /a and A B C").annotationQuery)
+      .containsOnly(entry("L O L", ""), entry("http.path", "/a"), entry("bar", "123"), entry("A B C", ""));
+    // border case, L O L as a tag and annotation, seems that the annotation takes precedence
+    assertThat(queryBuilder.parseAnnotationQuery("L O L=123 and L O L and http.path = /a").annotationQuery)
```
jorgheymans: what is the expected behavior when a query contains a tag with the same key as an annotation in said query ?

codefromthecrypt: the backend will do a redundant query. Technically, the most specific would win based on the query rules.

codefromthecrypt (approved):
Thanks for the help!

the fact than someone can enter a redundant query is unrelated to this change, though interesting.

jorgheymans:
The query string is parsed into a map keyed with the annotation keys. So in
fact the last one specified in the query wins, even though it can be the
least specific one. An easy fix would be to only overwrite keys in the map
in case of a tag parameter and drop annotation parameters if a tag value is
present already. WDYT?

Op do 9 mei 2019 02:20 schreef Adrian Cole <notifications@github.com>:

codefromthecrypt:
sounds good

codefromthecrypt:
will merge when you say you're done :) thanks!

jorgheymans:
good to go now, feel free to merge :)

codefromthecrypt:
travis is upset ... lemme look

zipkin/src/main/java/zipkin2/storage/QueryRequest.java
```diff
@@ -196,10 +197,13 @@ public Builder parseAnnotationQuery(@Nullable String annotationQuery) {
       for (String ann : annotationQuery.split(" and ", 100)) {
         int idx = ann.indexOf('=');
         if (idx == -1) {
-          map.put(ann, "");
+          // put the annotation only if there is no key present already, prevents overriding more specific tags
+          map.putIfAbsent(ann.trim(), "");
```
codefromthecrypt: this is a JDK 1.8 function so failing because this type is in the core library.
```suggestion
          ann = ann.trim();
          if (!map.containsKey(ann)) map.put(ann, "");
```

PR #2573, opened 2019-05-09
Implements deduplication in Elasticsearch

codefromthecrypt:
## Overview

This implements document deduplication based on trace ID and MD5 of what we generate. This will help partial failures of bulk requests, transparent retries, etc from using extra resources at rest or when analysis jobs like dependency aggregation occur.

## Why set document ID to `${traceID}-${MD5(json)}`
There can be multiple updates per trace ID and span ID, for example, client and server, or a late tag. For this reason, we need some sort of content hash. MD5 is a cheap hash, but has a mild collision risk. Adding the trace ID removes that risk and also helps to classify documents for purposes of knowing how often traces include a collision.

## Implementation Details

Before, we used a combination of our buffer and Okio's Buffer, and a
combination of our json encoder and Moshi's. This changes all the code
in Elasticsearch to use Okio buffers which reduces the amount of array
allocations and allows us to do cheap content hashes .

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -70,8 +70,13 @@
     </dependency>
 
     <dependency>
-      <groupId>org.apache.zipkin</groupId>
+      <groupId>${project.groupId}</groupId>
       <artifactId>zipkin-server</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}.zipkin2</groupId>
+      <artifactId>zipkin-storage-elasticsearch</artifactId>
       <version>${project.version}</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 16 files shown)


codefromthecrypt:
I ran a 5m load test against before and after. As it is, this doesn't make a big difference, likely because by default under load senders will produce 3MiB or so json per message. This will still result in big arrays. OTOH, we now have a chance to optimize.

the left hand side is before and the right hand side is with this change. The test is brave-webmvc-example against zipkin+elasticsearch 6.7 for 5 minutes

Screen Shot 2019-05-09 at 12 08 03 PM

codefromthecrypt:
the above load test result was before MD5 was added.. I'll run it again

codefromthecrypt:
with hashing the load is up a bit from the base case (based on my not entirely scientific scenario). I will try to buy back some performance from elsewhere to pay for the hashing.

Screen Shot 2019-05-09 at 2 02 56 PM

codefromthecrypt:
ok fixed some memory overhead

codefromthecrypt:
OK I'm personally happy with this. the line corresponding to GC time (100ms) and heap used (<100MiB) is stable throughout the load (8.5Kspans/second indexing into ES in my laptop)

Screen Shot 2019-05-10 at 12 16 44 PM

PR #2576, opened 2019-05-10
Simplifies build by bundling Scribe (still disabled by default)

codefromthecrypt:
We used to bundle scribe, but due to library conflicts, moved it
optional. As Armeria natively supports thrift (incidentally armeria
means sea thrift), we no longer have this problem.

Bundling removes one of the most complex things from our project for
sites that still use Scribe (most folks who use Finagle will).

This adds 1MiB to the server jar, which could be reduced further if
needed by:
* inlining scribe types instead of using the large generated files
* somehow removing the libthrift dep

I don't think this is worthwhile yet.


shakuzen (approved):
LGTM with a small whisper

codefromthecrypt:
heh cool :)

Issue #2435, opened 2019-03-07
Add ByteBuffer versions to BytesDecoder


anuraaga:
**Feature:**
Description of the feature

Currently, `BytesDecoder` only supports operating on `byte[]`. It could define similar methods that work on `ByteBuffer`.

**Rational**
Why this feature should be in zipkin

As zipkin is migrating to using Netty via Armeria and looking into adding gRPC ingestion, optimizations become possible that use Netty's `ByteBuf`, which is a manually allocated buffer. Using `ByteBuf` more would reduce garbage and probably increase throughput of zipkin-server, and this can be done by making methods that accept `ByteBuffer`.

**Example Scenario**
When will this be used by operators?

zipkin-server would use it when ingesting spans.

Implementation would look like

- JSON - currently uses `ByteArrayInputStream`. While there is no `ByteBufferInputStream` in the JDK for some reason, it would be trivial to add one like this and the rest of the business logic would be unchanged

  https://stackoverflow.com/questions/4332264/wrapping-a-bytebuffer-with-an-inputstream/6603018#6603018

  Though it would also be worth considering moving from gson to Jackson - in my experience I've gotten better performance from the latter, and it would natively support working on `ByteBuffer`. Just for reference, here is some code that uses Jackson's incremental parser https://github.com/curioswitch/curiostack/blob/master/common/grpc/protobuf-jackson/src/main/java/org/curioswitch/common/protobuf/json/ParseSupport.java#L65

- Thrift - currently uses `ByteBuffer.wrap`. As it already operates on `ByteBuffer`, there is basically no work involved

- Proto - Uses lots of manual operations on `byte[]`. Would need the most logic rewriting. My recommendation would be going ahead and converting to `ByteBuffer` and using `ByteBuffer.wrap` for the `byte[]` path

@adriancole does this seem like a reasonable idea?

codefromthecrypt:
I would be ok with this except please don't redo internals without benchmarking. It is my experience that not only are most payload types byte arrays (ex kafka) but also buffers are a lot slower in benchmarks.

In other words I don't think this is important enough to hurt perf... a lot of energy was spent on getting perf to the place it is.. if you can manage to keep or beat those numbers with internal rewriting, though, knock yourself out :)

last thing is to watch for api level.. zipkin core is java 6.

anuraaga:
[...]

PR #2579, opened 2019-05-10
Benchmark bytes / bytebuffer, protobuf vs zipkin vs wire.

anuraaga:
For #2435 

This is just a hacky set of code to compare performance of decoding `ByteBuffer` and `byte[]`. I used the protobuf library mainly since it has native decoders for each type so is easiest to compare. Naturally zipkin's `Buffer` could be updated to act on `ByteBuffer` directly to avoid this, though I'm not sure it'd be faster since protobuf uses `Unsafe` when it can. wire does especially poorly with `ByteBuffer` it seems, since it just copies it into a `byte[]` but doesn't have netty's unsafe optimizations like `ByteBufUtil.getBytes` would have.

As expected, wire is especially fast in time here, probably because it doesn't do hex decoding - it's nowhere near an apple comparison and mainly for reference. Otherwise, `byte[]` seems to be a bit faster, which isn't unexpected given the JVM is optimized for it, but the difference is mainly in the noise throughout. This could possibly be considered enough reason to switch to `ByteBuffer` though as given similar parsing time, `ByteBuffer` version is a bit better since it generates no garbage and could potentially reduce GC.

```
Benchmark                                        (num)  Mode  Cnt    Score    Error  Units
ProtoCodecBenchmarks.bytebuffer_protobufDecoder      1  avgt   15      1.267 ±   0.014  us/op
ProtoCodecBenchmarks.bytebuffer_protobufDecoder     10  avgt   15     12.244 ±   0.165  us/op
ProtoCodecBenchmarks.bytebuffer_protobufDecoder    100  avgt   15    121.105 ±   1.388  us/op
ProtoCodecBenchmarks.bytebuffer_protobufDecoder   1000  avgt   15   1209.171 ±  18.580  us/op
ProtoCodecBenchmarks.bytebuffer_protobufDecoder  10000  avgt   15  12839.910 ± 771.469  us/op
ProtoCodecBenchmarks.bytebuffer_wireDecoder          1  avgt   15      0.214 ±   0.018  us/op
ProtoCodecBenchmarks.bytebuffer_wireDecoder         10  avgt   15      0.969 ±   0.064  us/op
ProtoCodecBenchmarks.bytebuffer_wireDecoder        100  avgt   15      8.317 ±   0.600  us/op
ProtoCodecBenchmarks.bytebuffer_wireDecoder       1000  avgt   15     90.706 ±   2.031  us/op
ProtoCodecBenchmarks.bytebuffer_wireDecoder      10000  avgt   15   1039.518 ±  65.438  us/op
ProtoCodecBenchmarks.bytebuffer_zipkinDecoder        1  avgt   15      1.389 ±   0.049  us/op
ProtoCodecBenchmarks.bytebuffer_zipkinDecoder       10  avgt   15     13.393 ±   0.802  us/op
ProtoCodecBenchmarks.bytebuffer_zipkinDecoder      100  avgt   15    111.140 ±   2.660  us/op
ProtoCodecBenchmarks.bytebuffer_zipkinDecoder     1000  avgt   15   1234.412 ± 106.794  us/op
ProtoCodecBenchmarks.bytebuffer_zipkinDecoder    10000  avgt   15  11658.781 ± 694.221  us/op
ProtoCodecBenchmarks.bytes_protobufDecoder           1  avgt   15      1.143 ±   0.031  us/op
ProtoCodecBenchmarks.bytes_protobufDecoder          10  avgt   15     11.733 ±   0.545  us/op
ProtoCodecBenchmarks.bytes_protobufDecoder         100  avgt   15    122.660 ±   7.020  us/op
ProtoCodecBenchmarks.bytes_protobufDecoder        1000  avgt   15   1287.893 ±  41.410  us/op
ProtoCodecBenchmarks.bytes_protobufDecoder       10000  avgt   15  11838.844 ± 184.314  us/op
ProtoCodecBenchmarks.bytes_wireDecoder               1  avgt   15      0.136 ±   0.005  us/op
ProtoCodecBenchmarks.bytes_wireDecoder              10  avgt   15      0.598 ±   0.024  us/op
ProtoCodecBenchmarks.bytes_wireDecoder             100  avgt   15      5.603 ±   0.113  us/op
ProtoCodecBenchmarks.bytes_wireDecoder            1000  avgt   15     67.543 ±   1.314  us/op
ProtoCodecBenchmarks.bytes_wireDecoder           10000  avgt   15    775.430 ±  72.922  us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder             1  avgt   15      1.132 ±   0.061  us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder            10  avgt   15     10.942 ±   0.225  us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder           100  avgt   15    117.941 ±   8.454  us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder          1000  avgt   15   1279.525 ±  73.367  us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder         10000  avgt   15  12400.006 ± 469.528  us/op
```

### DIFF
--- a/benchmarks/src/main/java/zipkin2/codec/ProtoCodecBenchmarks.java
+++ b/benchmarks/src/main/java/zipkin2/codec/ProtoCodecBenchmarks.java
@@ -54,17 +54,14 @@ public class ProtoCodecBenchmarks {
   static final byte[] clientSpanJsonV2 = read("/zipkin2-client.json");
   static final Span clientSpan = SpanBytesDecoder.JSON_V2.decodeOne(clientSpanJsonV2);
 
-  @Param({"1", "10", "100", "1000", "10000"})
-  public int num;
+  // Assume a message is 1000 spans (which is a high number for as this is per-node-second)
+  static final List<Span> spans = Collections.nCopies(1000, clientSpan);
+  static final byte[] encodedBytes = SpanBytesEncoder.PROTO3.encodeList(spans);
 
-  private byte[] encodedBytes;
   private ByteBuf encodedBuf;
 
   @Setup
   public void setup() {
-    List<Span> spans = Collections.nCopies(num, clientSpan);
-    encodedBytes = SpanBytesEncoder.PROTO3.encodeList(spans);
-
     encodedBuf = PooledByteBufAllocator.DEFAULT.buffer(encodedBytes.length);
     encodedBuf.writeBytes(encodedBytes);
   }
@@ -109,7 +106,6 @@ public class ProtoCodecBenchmarks {
     Options opt = new OptionsBuilder()
       .include(".*" + ProtoCodecBenchmarks.class.getSimpleName() + ".*bytebuffer_")
       .addProfiler("gc")
-      .param("num", "10000")
       .build();
 
     new Runner(opt).run();
--- a/benchmarks/src/main/java/zipkin2/codec/ProtoCodecBenchmarks.java
+++ b/benchmarks/src/main/java/zipkin2/codec/ProtoCodecBenchmarks.java
@@ -8,7 +8,6 @@ import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
-import okio.ByteString;
 import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.BenchmarkMode;
 import org.openjdk.jmh.annotations.Fork;
@@ -23,7 +22,6 @@ import org.openjdk.jmh.annotations.TearDown;
 import org.openjdk.jmh.annotations.Threads;
 import org.openjdk.jmh.annotations.Warmup;
 import org.openjdk.jmh.runner.Runner;
-import org.openjdk.jmh.runner.RunnerException;
 import org.openjdk.jmh.runner.options.Options;
 import org.openjdk.jmh.runner.options.OptionsBuilder;
 import zipkin2.Span;
@@ -31,7 +29,7 @@ import zipkin2.Span;
 @Measurement(iterations = 5, time = 1)
 @Warmup(iterations = 10, time = 1)
 @Fork(3)
-@BenchmarkMode(Mode.AverageTime)
+@BenchmarkMode(Mode.SampleTime)
 @OutputTimeUnit(TimeUnit.MICROSECONDS)
 @State(Scope.Thread)
 @Threads(1)
@@ -71,8 +69,18 @@ public class ProtoCodecBenchmarks {
   }
 
   @Benchmark
-  public List<zipkin2.proto3.Span> bytes_wireDecoder() throws IOException {
-    return zipkin2.proto3.Span.ADAPTER.asRepeated().decode(encodedBytes);
+  public List<Span> bytes_copy_protobufDecoder() {
+    return ProtobufSpanDecoder.decodeList(ByteBufUtil.getBytes(encodedBuf));
+  }
+
+  @Benchmark
+  public List<Span> bytes_wireDecoder() throws IOException {
+    return WireSpanDecoder.decodeList(encodedBytes);
+  }
+
+  @Benchmark
+  public List<Span> bytes_copy_wireDecoder() throws IOException {
+    return WireSpanDecoder.decodeList(ByteBufUtil.getBytes(encodedBuf));
   }
 
 
@@ -87,14 +95,16 @@ public class ProtoCodecBenchmarks {
   }
 
   @Benchmark
-  public List<zipkin2.proto3.Span> bytebuffer_wireDecoder() throws IOException {
-    return zipkin2.proto3.Span.ADAPTER.asRepeated().decode(ByteString.of(encodedBuf.nioBuffer()));
+  public List<Span> bytebuffer_wireDecoder() {
+    return WireSpanDecoder.decodeList(encodedBuf.nioBuffer());
   }
 
   // Convenience main entry-point
-  public static void main(String[] args) throws RunnerException {
+  public static void main(String[] args) throws Exception {
     Options opt = new OptionsBuilder()
-      .include(".*" + ProtoCodecBenchmarks.class.getSimpleName())
+      .include(".*" + ProtoCodecBenchmarks.class.getSimpleName() + ".*bytes_")
+      .addProfiler("gc")
+      .param("num", "10000")
       .build();
 
     new Runner(opt).run();
--- a/benchmarks/src/main/java/zipkin2/codec/ProtoCodecBenchmarks.java
+++ b/benchmarks/src/main/java/zipkin2/codec/ProtoCodecBenchmarks.java
@@ -1,3 +1,19 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package zipkin2.codec;
 
 import com.google.common.io.Resources;
@@ -69,20 +85,9 @@ public class ProtoCodecBenchmarks {
   }
 
   @Benchmark
-  public List<Span> bytes_copy_protobufDecoder() {
-    return ProtobufSpanDecoder.decodeList(ByteBufUtil.getBytes(encodedBuf));
-  }
-
-  @Benchmark
-  public List<Span> bytes_wireDecoder() throws IOException {
+  public List<Span> bytes_wireDecoder() {
     return WireSpanDecoder.decodeList(encodedBytes);
   }
-
-  @Benchmark
-  public List<Span> bytes_copy_wireDecoder() throws IOException {
-    return WireSpanDecoder.decodeList(ByteBufUtil.getBytes(encodedBuf));
-  }
-
 
   @Benchmark
   public List<Span> bytebuffer_zipkinDecoder() {
@@ -102,7 +107,7 @@ public class ProtoCodecBenchmarks {
   // Convenience main entry-point
   public static void main(String[] args) throws Exception {
     Options opt = new OptionsBuilder()
-      .include(".*" + ProtoCodecBenchmarks.class.getSimpleName() + ".*bytes_")
+      .include(".*" + ProtoCodecBenchmarks.class.getSimpleName() + ".*bytebuffer_")
       .addProfiler("gc")
       .param("num", "10000")
       .build();
--- a/benchmarks/src/main/java/zipkin2/codec/ProtoCodecBenchmarks.java
+++ b/benchmarks/src/main/java/zipkin2/codec/ProtoCodecBenchmarks.java
@@ -1,0 +1,110 @@
+package zipkin2.codec;
+
+import com.google.common.io.Resources;
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufUtil;
+import io.netty.buffer.PooledByteBufAllocator;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import okio.ByteString;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import zipkin2.Span;
+
+@Measurement(iterations = 5, time = 1)
+@Warmup(iterations = 10, time = 1)
+@Fork(3)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@State(Scope.Thread)
+@Threads(1)
+public class ProtoCodecBenchmarks {
+
+  static final byte[] clientSpanJsonV2 = read("/zipkin2-client.json");
+  static final Span clientSpan = SpanBytesDecoder.JSON_V2.decodeOne(clientSpanJsonV2);
+
+  @Param({"1", "10", "100", "1000", "10000"})
+  public int num;
+
+  private byte[] encodedBytes;
+  private ByteBuf encodedBuf;
+
+  @Setup
+  public void setup() {
+    List<Span> spans = Collections.nCopies(num, clientSpan);
+    encodedBytes = SpanBytesEncoder.PROTO3.encodeList(spans);
+
+    encodedBuf = PooledByteBufAllocator.DEFAULT.buffer(encodedBytes.length);
+    encodedBuf.writeBytes(encodedBytes);
+  }
+
+  @TearDown
+  public void tearDown() {
+    encodedBuf.release();
+  }
+
+  @Benchmark
+  public List<Span> bytes_zipkinDecoder() {
+    return SpanBytesDecoder.PROTO3.decodeList(encodedBytes);
+  }
+
+  @Benchmark
+  public List<Span> bytes_protobufDecoder() {
+    return ProtobufSpanDecoder.decodeList(encodedBytes);
+  }
+
+  @Benchmark
+  public List<zipkin2.proto3.Span> bytes_wireDecoder() throws IOException {
+    return zipkin2.proto3.Span.ADAPTER.asRepeated().decode(encodedBytes);
+  }
+
+
+  @Benchmark
+  public List<Span> bytebuffer_zipkinDecoder() {
+    return SpanBytesDecoder.PROTO3.decodeList(ByteBufUtil.getBytes(encodedBuf));
+  }
+
+  @Benchmark
+  public List<Span> bytebuffer_protobufDecoder() {
+    return ProtobufSpanDecoder.decodeList(encodedBuf.nioBuffer());
+  }
+
+  @Benchmark
+  public List<zipkin2.proto3.Span> bytebuffer_wireDecoder() throws IOException {
+    return zipkin2.proto3.Span.ADAPTER.asRepeated().decode(ByteString.of(encodedBuf.nioBuffer()));
+  }
+
+  // Convenience main entry-point
+  public static void main(String[] args) throws RunnerException {
+    Options opt = new OptionsBuilder()
+      .include(".*" + ProtoCodecBenchmarks.class.getSimpleName())
+      .build();
+
+    new Runner(opt).run();
+  }
+
+  static byte[] read(String resource) {
+    try {
+      return Resources.toByteArray(Resources.getResource(CodecBenchmarks.class, resource));
+    } catch (IOException e) {
+      throw new IllegalStateException(e);
+    }
+  }
+}
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -70,6 +70,12 @@
     </dependency>
 
     <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+      <version>3.7.1</version>
+    </dependency>
+
+    <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>zipkin-server</artifactId>
       <version>${project.version}</version>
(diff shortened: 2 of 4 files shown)


benchmarks/src/main/java/zipkin2/codec/ProtoCodecBenchmarks.java
```diff
@@ -0,0 +1,110 @@
...
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import zipkin2.Span;
+
+@Measurement(iterations = 5, time = 1)
+@Warmup(iterations = 10, time = 1)
+@Fork(3)
+@BenchmarkMode(Mode.AverageTime)
```
codefromthecrypt:
```
@BenchmarkMode(Mode.SampleTime)
```

then you can also look at the following, which I use grep for

where /tmp/foo is output from this
```bash
$ egrep '(p0.999 |gc.alloc.rate.norm)' /tmp/foo
```

note that sometimes things that make sense also don't because things aren't in isolation. For example, you could imagine the combinatory effect of shared buffers parsing and encoding to be of interest.

codefromthecrypt:
PS some background always helps...

our core library includes a dependency free codec. I think this is still important eventhough the parsing side is less so (it does help in wonky huge dep clash environments like spark)

in our server, we have some flexibility, and to that end I was thinking about having the ability to override the codecs towards your initial issue #2435. When I started optimizing the Elasticsearch side, I noticed that our v2 format is so easy now, even converting from a codec-specific type like proto.span *is still cheaper* than us doing it directly for some reasons you mentioned.  While using google's proto library would re-introduce their version thrash I do like doing comparisons. I've thought most about wire as that is so incredibly more efficient than ours, and it has no notable deps we don't already depend on.

Thanks for doing this comparison!

anuraaga:
Yeah it'd definitely only be for the server. And while `protobuf-java-util` has some dependencies, just `protobuf-java` is dependency free and I believe has never broken backwards compatibility so I wouldn't write it off so fast ;)

I added a wire-based decoder too, with the same code flow as the protobuf-based one, just uses wire for the decoding. Now that it is doing hex decoding, it's a proper comparison. Wire is quite a bit slower, probably since it has to do an array copy before converting to hex since it doesn't expose the ability to read single bytes, and maybe because protobuf uses `Unsafe` though I don't know if that really makes a big impact on modern JVMs (it would have a larger impact with java 8 than java 11 I guess).

Protobuf vs zipkin seems to show protobuf a few percent faster. 

Garbage-wise, as expected the lack of an initial array copy does remove a whole buffer's worth of allocation, but due to the lots of allocations to create hex and other strings, it's a very small percentage overall, ~3%, so just a little improvement. This is after applying a change to reuse `char[]`, which had a similar improvement in the base case too.

There do seem to be some differences shown here, though they're not so huge as to make it clear whether any approach is the best.

```
Benchmark                                                                           (num)    Mode   Cnt         Score         Error   Units
ProtoCodecBenchmarks.bytes_copy_protobufDecoder                                     10000  sample  1309     11510.855 ±     138.341   us/op
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:bytes_copy_protobufDecoder·p0.00    10000  sample           10010.624                 us/op
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:bytes_copy_protobufDecoder·p0.50    10000  sample           10977.280                 us/op
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:bytes_copy_protobufDecoder·p0.90    10000  sample           13156.352                 us/op
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:bytes_copy_protobufDecoder·p0.95    10000  sample           15319.040                 us/op
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:bytes_copy_protobufDecoder·p0.99    10000  sample           16967.270                 us/op
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:bytes_copy_protobufDecoder·p0.999   10000  sample           24244.060                 us/op
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:bytes_copy_protobufDecoder·p0.9999  10000  sample           25493.504                 us/op
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:bytes_copy_protobufDecoder·p1.00    10000  sample           25493.504                 us/op
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:·gc.alloc.rate                      10000  sample    15      1212.778 ±      22.951  MB/sec
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:·gc.alloc.rate.norm                 10000  sample    15  21949689.300 ±     213.505    B/op
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:·gc.churn.G1_Eden_Space             10000  sample    15      1110.207 ±     113.046  MB/sec
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:·gc.churn.G1_Eden_Space.norm        10000  sample    15  20094766.898 ± 2030230.821    B/op
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:·gc.churn.G1_Old_Gen                10000  sample    15       112.514 ±      11.349  MB/sec
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:·gc.churn.G1_Old_Gen.norm           10000  sample    15   2036516.677 ±  203864.604    B/op
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:·gc.churn.G1_Survivor_Space         10000  sample    15         0.443 ±       0.438  MB/sec
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:·gc.churn.G1_Survivor_Space.norm    10000  sample    15      8009.395 ±    7904.907    B/op
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:·gc.count                           10000  sample    15        82.000                counts
ProtoCodecBenchmarks.bytes_copy_protobufDecoder:·gc.time                            10000  sample    15       346.000                    ms
ProtoCodecBenchmarks.bytes_copy_wireDecoder                                         10000  sample  1068     14170.764 ±     135.522   us/op
ProtoCodecBenchmarks.bytes_copy_wireDecoder:bytes_copy_wireDecoder·p0.00            10000  sample           12648.448                 us/op
ProtoCodecBenchmarks.bytes_copy_wireDecoder:bytes_copy_wireDecoder·p0.50            10000  sample           13713.408                 us/op
ProtoCodecBenchmarks.bytes_copy_wireDecoder:bytes_copy_wireDecoder·p0.90            10000  sample           15810.560                 us/op
ProtoCodecBenchmarks.bytes_copy_wireDecoder:bytes_copy_wireDecoder·p0.95            10000  sample           16846.029                 us/op
ProtoCodecBenchmarks.bytes_copy_wireDecoder:bytes_copy_wireDecoder·p0.99            10000  sample           19930.808                 us/op
ProtoCodecBenchmarks.bytes_copy_wireDecoder:bytes_copy_wireDecoder·p0.999           10000  sample           24186.126                 us/op
ProtoCodecBenchmarks.bytes_copy_wireDecoder:bytes_copy_wireDecoder·p0.9999          10000  sample           24281.088                 us/op
ProtoCodecBenchmarks.bytes_copy_wireDecoder:bytes_copy_wireDecoder·p1.00            10000  sample           24281.088                 us/op
ProtoCodecBenchmarks.bytes_copy_wireDecoder:·gc.alloc.rate                          10000  sample    15      1129.202 ±      28.012  MB/sec
ProtoCodecBenchmarks.bytes_copy_wireDecoder:·gc.alloc.rate.norm                     10000  sample    15  25120909.386 ±     202.273    B/op
ProtoCodecBenchmarks.bytes_copy_wireDecoder:·gc.churn.G1_Eden_Space                 10000  sample    15      1040.511 ±      54.956  MB/sec
ProtoCodecBenchmarks.bytes_copy_wireDecoder:·gc.churn.G1_Eden_Space.norm            10000  sample    15  23153464.194 ± 1194946.337    B/op
ProtoCodecBenchmarks.bytes_copy_wireDecoder:·gc.churn.G1_Old_Gen                    10000  sample    15        90.519 ±       4.650  MB/sec
ProtoCodecBenchmarks.bytes_copy_wireDecoder:·gc.churn.G1_Old_Gen.norm               10000  sample    15   2014291.467 ±  102517.266    B/op
ProtoCodecBenchmarks.bytes_copy_wireDecoder:·gc.churn.G1_Survivor_Space             10000  sample    15         0.838 ±       0.323  MB/sec
ProtoCodecBenchmarks.bytes_copy_wireDecoder:·gc.churn.G1_Survivor_Space.norm        10000  sample    15     18578.527 ±    6846.810    B/op
ProtoCodecBenchmarks.bytes_copy_wireDecoder:·gc.count                               10000  sample    15        76.000                counts
ProtoCodecBenchmarks.bytes_copy_wireDecoder:·gc.time                                10000  sample    15       173.000                    ms
ProtoCodecBenchmarks.bytes_protobufDecoder                                          10000  sample  1340     11252.910 ±     126.595   us/op
ProtoCodecBenchmarks.bytes_protobufDecoder:bytes_protobufDecoder·p0.00              10000  sample            9912.320                 us/op
ProtoCodecBenchmarks.bytes_protobufDecoder:bytes_protobufDecoder·p0.50              10000  sample           10731.520                 us/op
ProtoCodecBenchmarks.bytes_protobufDecoder:bytes_protobufDecoder·p0.90              10000  sample           12546.867                 us/op
ProtoCodecBenchmarks.bytes_protobufDecoder:bytes_protobufDecoder·p0.95              10000  sample           14974.976                 us/op
ProtoCodecBenchmarks.bytes_protobufDecoder:bytes_protobufDecoder·p0.99              10000  sample           16912.548                 us/op
ProtoCodecBenchmarks.bytes_protobufDecoder:bytes_protobufDecoder·p0.999             10000  sample           19200.541                 us/op
ProtoCodecBenchmarks.bytes_protobufDecoder:bytes_protobufDecoder·p0.9999            10000  sample           19267.584                 us/op
ProtoCodecBenchmarks.bytes_protobufDecoder:bytes_protobufDecoder·p1.00              10000  sample           19267.584                 us/op
ProtoCodecBenchmarks.bytes_protobufDecoder:·gc.alloc.rate                           10000  sample    15      1130.991 ±      25.879  MB/sec
ProtoCodecBenchmarks.bytes_protobufDecoder:·gc.alloc.rate.norm                      10000  sample    15  20009660.550 ±     259.757    B/op
ProtoCodecBenchmarks.bytes_protobufDecoder:·gc.churn.G1_Eden_Space                  10000  sample    15      1136.615 ±     111.000  MB/sec
ProtoCodecBenchmarks.bytes_protobufDecoder:·gc.churn.G1_Eden_Space.norm             10000  sample    15  20105313.287 ± 1867925.592    B/op
ProtoCodecBenchmarks.bytes_protobufDecoder:·gc.churn.G1_Old_Gen                     10000  sample    15         0.319 ±       0.197  MB/sec
ProtoCodecBenchmarks.bytes_protobufDecoder:·gc.churn.G1_Old_Gen.norm                10000  sample    15      5659.194 ±    3507.786    B/op
ProtoCodecBenchmarks.bytes_protobufDecoder:·gc.churn.G1_Survivor_Space              10000  sample    15         0.354 ±       0.454  MB/sec
ProtoCodecBenchmarks.bytes_protobufDecoder:·gc.churn.G1_Survivor_Space.norm         10000  sample    15      6261.311 ±    8110.558    B/op
ProtoCodecBenchmarks.bytes_protobufDecoder:·gc.count                                10000  sample    15        84.000                counts
ProtoCodecBenchmarks.bytes_protobufDecoder:·gc.time                                 10000  sample    15       352.000                    ms
ProtoCodecBenchmarks.bytes_wireDecoder                                              10000  sample  1027     14696.033 ±     274.890   us/op
ProtoCodecBenchmarks.bytes_wireDecoder:bytes_wireDecoder·p0.00                      10000  sample           12697.600                 us/op
ProtoCodecBenchmarks.bytes_wireDecoder:bytes_wireDecoder·p0.50                      10000  sample           13991.936                 us/op
ProtoCodecBenchmarks.bytes_wireDecoder:bytes_wireDecoder·p0.90                      10000  sample           16420.045                 us/op
ProtoCodecBenchmarks.bytes_wireDecoder:bytes_wireDecoder·p0.95                      10000  sample           17812.685                 us/op
ProtoCodecBenchmarks.bytes_wireDecoder:bytes_wireDecoder·p0.99                      10000  sample           26746.552                 us/op
ProtoCodecBenchmarks.bytes_wireDecoder:bytes_wireDecoder·p0.999                     10000  sample           63427.052                 us/op
ProtoCodecBenchmarks.bytes_wireDecoder:bytes_wireDecoder·p0.9999                    10000  sample           64225.280                 us/op
ProtoCodecBenchmarks.bytes_wireDecoder:bytes_wireDecoder·p1.00                      10000  sample           64225.280                 us/op
ProtoCodecBenchmarks.bytes_wireDecoder:·gc.alloc.rate                               10000  sample    15      1003.950 ±      33.760  MB/sec
ProtoCodecBenchmarks.bytes_wireDecoder:·gc.alloc.rate.norm                          10000  sample    15  23180993.238 ±     327.812    B/op
ProtoCodecBenchmarks.bytes_wireDecoder:·gc.churn.G1_Eden_Space                      10000  sample    15      1001.456 ±     113.305  MB/sec
ProtoCodecBenchmarks.bytes_wireDecoder:·gc.churn.G1_Eden_Space.norm                 10000  sample    15  23105838.211 ± 2265066.780    B/op
ProtoCodecBenchmarks.bytes_wireDecoder:·gc.churn.G1_Old_Gen                         10000  sample    15         0.585 ±       0.150  MB/sec
ProtoCodecBenchmarks.bytes_wireDecoder:·gc.churn.G1_Old_Gen.norm                    10000  sample    15     13499.422 ±    3403.060    B/op
ProtoCodecBenchmarks.bytes_wireDecoder:·gc.churn.G1_Survivor_Space                  10000  sample    15         0.752 ±       0.249  MB/sec
ProtoCodecBenchmarks.bytes_wireDecoder:·gc.churn.G1_Survivor_Space.norm             10000  sample    15     17272.859 ±    5177.973    B/op
ProtoCodecBenchmarks.bytes_wireDecoder:·gc.count                                    10000  sample    15        73.000                counts
ProtoCodecBenchmarks.bytes_wireDecoder:·gc.time                                     10000  sample    15       158.000                    ms
ProtoCodecBenchmarks.bytes_zipkinDecoder                                            10000  sample  1220     12361.137 ±     243.588   us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:bytes_zipkinDecoder·p0.00                  10000  sample            9863.168                 us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:bytes_zipkinDecoder·p0.50                  10000  sample           11452.416                 us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:bytes_zipkinDecoder·p0.90                  10000  sample           15382.938                 us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:bytes_zipkinDecoder·p0.95                  10000  sample           18246.861                 us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:bytes_zipkinDecoder·p0.99                  10000  sample           22308.127                 us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:bytes_zipkinDecoder·p0.999                 10000  sample           26914.292                 us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:bytes_zipkinDecoder·p0.9999                10000  sample           27066.368                 us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:bytes_zipkinDecoder·p1.00                  10000  sample           27066.368                 us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:·gc.alloc.rate                             10000  sample    15       984.706 ±      72.183  MB/sec
ProtoCodecBenchmarks.bytes_zipkinDecoder:·gc.alloc.rate.norm                        10000  sample    15  19129801.132 ±     226.638    B/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:·gc.churn.G1_Eden_Space                    10000  sample    15       993.104 ±     126.048  MB/sec
ProtoCodecBenchmarks.bytes_zipkinDecoder:·gc.churn.G1_Eden_Space.norm               10000  sample    15  19270497.313 ± 1683336.845    B/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:·gc.churn.G1_Old_Gen                       10000  sample    15         0.434 ±       0.268  MB/sec
ProtoCodecBenchmarks.bytes_zipkinDecoder:·gc.churn.G1_Old_Gen.norm                  10000  sample    15      8434.596 ±    5290.023    B/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:·gc.churn.G1_Survivor_Space                10000  sample    15         0.753 ±       0.651  MB/sec
ProtoCodecBenchmarks.bytes_zipkinDecoder:·gc.churn.G1_Survivor_Space.norm           10000  sample    15     14673.383 ±   12535.509    B/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:·gc.count                                  10000  sample    15        83.000                counts
ProtoCodecBenchmarks.bytes_zipkinDecoder:·gc.time                                   10000  sample    15       335.000                    ms

**Comment by codefromthecrypt** [comment]
rebased your branch as there were some merge conflicts due to drifts

**Comment by codefromthecrypt** [comment]
I'm playing around a bit about the hex it is yeah double array allocation indeed

**Comment by codefromthecrypt** [comment]
one thing I was thinking about in this change is that when you are ultimately backed by an array, google sneakily looks to see if that's the case and special cases accordingly.

When we have an array is unlikely something we'd optimize for in the grpc listener. I'm wondering if we can get a more realistic source like what armeria would be using and apply that to the bytebuffer (or otherwise) case...

PS I also looked at wire and started to play with inlining a hex utility.

**Comment by codefromthecrypt** [comment]
cc also @swankjesse as this sort of thing is fun..

I think special peeking aside, it might boil down to if we can manage to read proto bytes as hex. if we can definitely wire should be better than it is now.. 

I added this to wire's ProtoReader though haven't tested it quite yet
```kotlin
  // stolen from okio
  private val HEX_DIGITS =
      charArrayOf('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f')

  @Suppress("NOTHING_TO_INLINE")
  internal inline fun hex(data: BufferedSource, byteCount: Int): String {
    val result = CharArray(byteCount * 2)
    var i = 0
    while (i < result.size) {
      val b = data.readByte().toInt();
      result[i++] = HEX_DIGITS[b shr 4 and 0xf]
      result[i++] = HEX_DIGITS[b       and 0xf] // ktlint-disable no-multi-spaces
    }
    return String(result)
  }

  /**
   * Reads a `bytes` field value from the stream as a lower-hex string. The length is read from the
   * stream prior to the actual data.
   */
  @Throws(IOException::class)
  fun readBytesAsHex(): String {
    val byteCount = beforeLengthDelimitedScalar()
    source.require(byteCount) // Throws EOFException if insufficient bytes are available.
    return hex(source, byteCount.toInt())
  }
```

codefromthecrypt:
I'll revise what I said above @anuraaga has in the bench setup is fairly realistic as the bytebuffer isn't a cover for a byte array.

However, like Rag said, google's library is peeking and using Unsafe under the covers. I'm not sure Okio does that (or assumes it could).

Specifically, google's returns [CodedInputStream.UnsafeDirectNioDecoder](https://github.com/protocolbuffers/protobuf/blob/5b232b8ecbce13286be09e703997e887ae0d464d/java/core/src/main/java/com/google/protobuf/CodedInputStream.java#L1278)

```java
  static CodedInputStream newInstance(ByteBuffer buf, boolean bufferIsImmutable) {
    if (buf.hasArray()) {
      return newInstance(buf.array(), buf.arrayOffset() + buf.position(), buf.remaining(), bufferIsImmutable);
    } else if (buf.isDirect() && CodedInputStream.UnsafeDirectNioDecoder.isSupported()) {
      return new CodedInputStream.UnsafeDirectNioDecoder(buf, bufferIsImmutable);
    } else {
```

anuraaga:
Yup using Unsafe is probably the biggest difference and in the Armeria case it would end up in the Unsafe path if we so choose. From what I understand, wire / okio is mostly for Android, so it wouldn't make sense for them to implement support for Unsafe which doesn't work there. Protobuf had to support calling Java from C++... but interestingly the optimizations that went into doing that do apply to Netty-based servers too like here since they use direct memory.

I'm personally leaning towards not changing anything at this point though since for single digit gains, having two parser implementations seems like it's not worth it. What do you think?

codefromthecrypt:
yep.. if you revise from 10k spans per message down to 1k spans (which is a fairly high number anyway considering this is per traced host/second), there is not much difference between google and wire anyway. However, it does help memory wise to have wire decode bytes into hex strings.

with wire making hex manually
```
ProtoCodecBenchmarks.bytebuffer_protobufDecoder:bytebuffer_protobufDecoder·p0.999   sample            2716.557                us/op
ProtoCodecBenchmarks.bytebuffer_protobufDecoder:·gc.alloc.rate.norm                 sample     15  2271332.980 ±     49.158    B/op
ProtoCodecBenchmarks.bytebuffer_wireDecoder:bytebuffer_wireDecoder·p0.999           sample            3082.207                us/op
ProtoCodecBenchmarks.bytebuffer_wireDecoder:·gc.alloc.rate.norm                     sample     15  2707555.118 ±     43.790    B/op
```

with wire native hex stringing
```
ProtoCodecBenchmarks.bytebuffer_protobufDecoder:bytebuffer_protobufDecoder·p0.999   sample            2808.340                us/op
ProtoCodecBenchmarks.bytebuffer_protobufDecoder:·gc.alloc.rate.norm                 sample     15  2271332.557 ±     37.638    B/op
ProtoCodecBenchmarks.bytebuffer_wireDecoder:bytebuffer_wireDecoder·p0.999           sample            3045.130                us/op
ProtoCodecBenchmarks.bytebuffer_wireDecoder:·gc.alloc.rate.norm                     sample     15  2579549.017 ±     59.592    B/op
```

codefromthecrypt:
@anuraaga leaning towards not changing anything I guess you mean changing the parsing approach? If so, I would say even if we wanted to, it can wait.

we are allocating more per request in parsing than necessary, but it is hard to tell the impact without looking at the total stuff allocated along the way. Regardless, I think we should merge your benchmarks so it can be picked up later.

codefromthecrypt:
PS @anuraaga I think the context you mention which is there are optimizations on the server that would be different than android-focused libraries is certainly spot-on. We shouldn't prevent ourselves from taking advantage of these optimizations considering this is server code.

I do remember some discussions last time about the guava sensitivity and heft of google libraries (ex. protobuf-java is 2MiB). I'm not super keen to re-enter that at the moment, but there could be some sense in using an unsafe mechanism such as they are.

At any rate, I learned thought this work that there's actually a measurable amount of gc time to recover. Eventhough I mentioned a little less priority, that's my personal 2p, if you are interested in taking the "attack the overhead" issue forward, please continue.

anuraaga:
So in the interest of having these benchmarks and comparison points, made this PR non-draft, we may as well get it in and play with it some more.

anuraaga:
Merging since this only touches anything in the not-so critical `benchmarks` module.

PR #2580, opened 2019-05-10
Makes it an error to store during assembly of a call

codefromthecrypt:
Before this, there was some extra code in the throttle package handling
a bug in our in memory storage. This fixes that and removes the extra
code.

See #2502

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
@@ -24,10 +24,8 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
-import java.util.function.Supplier;
 import zipkin2.Call;
 import zipkin2.Callback;
-import zipkin2.storage.InMemoryStorage;
 
 /**
  * {@link Call} implementation that is backed by an {@link ExecutorService}. The ExecutorService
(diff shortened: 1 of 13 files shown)


zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
```diff
@@ -41,39 +39,21 @@
...
-   * supplier call needs to be supplied later to avoid having it take action when it is created
-   * (like {@link InMemoryStorage} and thus avoid being throttled.
-   */
-  final Supplier<? extends Call<V>> supplier;
-  volatile Call<V> delegate;
-  volatile boolean canceled;
-
-  public ThrottledCall(ExecutorService executor, Limiter<Void> limiter,
-    Supplier<? extends Call<V>> supplier) {
+  final Call<V> delegate;
```
codefromthecrypt, line 45: look ma. no volatiles :)

codefromthecrypt:
finally Jenkins is green

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
```diff
@@ -153,17 +126,17 @@ static String setCurrentThreadName(String name) {
 
   final class QueuedCall implements Runnable {
```
Logic-32: Can this be `static` now?

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
Logic-32: :D

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
codefromthecrypt: yep fixing in the pending change thx

PR #2586, opened 2019-05-11
Folds Buffer tree into UnsafeBuffer to improve performance

codefromthecrypt:
Recent optimizations required peeking at the underlying byte array. This
change formalizes the buffer as one that is specialized for being array
backed (and writes are unsafe eg not bounds checked).

The results are now almost the same as google's proto library, despite
this library being smaller and including proto, json and thrift codecs
and a functional Zipkin storage engine:

Comparison: decoding 1000 proto3 encoded spans

```
ProtoCodecBenchmarks.bytes_protobufDecoder:bytes_protobufDecoder·p0.99       sample            1953.710                us/op
ProtoCodecBenchmarks.bytes_protobufDecoder:·gc.alloc.rate.norm               sample     15  2023206.265 ±     32.203    B/op
ProtoCodecBenchmarks.bytes_wireDecoder:bytes_wireDecoder·p0.99               sample            2351.104                us/op
ProtoCodecBenchmarks.bytes_wireDecoder:·gc.alloc.rate.norm                   sample     15  2435408.568 ±     53.612    B/op
```

Before
```
ProtoCodecBenchmarks.bytes_zipkinDecoder:bytes_zipkinDecoder·p0.99           sample            2162.688                us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:·gc.alloc.rate.norm                 sample     15  2095172.026 ±     45.379    B/op
```

After
```
ProtoCodecBenchmarks.bytes_zipkinDecoder:bytes_zipkinDecoder·p0.99           sample            1996.800                us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:·gc.alloc.rate.norm                 sample     15  2095168.617 ±     36.985    B/op
```

FYI: While this is array backed, it doesn't imply allocating a new array
per span. Those looking to read large amounts of data can use this by
slicing a buffer and filling for each length-prefixed byte string.


codefromthecrypt:
merging as interested in doing a perf test prior to release and there are multiple PRs involved

anuraaga (commented):
LGTM

PR #2589, opened 2019-05-12
Makes SpanBytesDecoder work on ByteBuffer

### LINKED ISSUE #2435: Add ByteBuffer versions to BytesDecoder

codefromthecrypt:
Internally, this also covers InputStream.

This was easier (though not easy) due to splitting how we handle
encoding from decoding.

Incidentally, the jar size of zipkin is smaller still vs the last
release.

Fixes #2435


zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/SelectFromTraces.java
```diff
@@ -163,7 +164,7 @@ public String toString() {
       return (row, result) -> {
         V1ThriftSpanReader reader = V1ThriftSpanReader.create();
         V1SpanConverter converter = V1SpanConverter.create();
-        V1Span read = reader.read(row.getBytes("span"));
+        V1Span read = reader.read(ReadBuffer.wrapUnsafe(row.getBytes("span")));
```
codefromthecrypt, line 167: here cassandra is using the bytebuffer variant directly

benchmarks/src/main/java/zipkin2/codec/ProtoCodecBenchmarks.java
```diff
@@ -87,7 +87,7 @@ public void tearDown() {
 
   @Benchmark
   public List<Span> bytebuffer_zipkinDecoder() {
-    return SpanBytesDecoder.PROTO3.decodeList(ByteBufUtil.getBytes(encodedBuf));
+    return SpanBytesDecoder.PROTO3.decodeList(encodedBuf.nioBuffer());
```
codefromthecrypt, line 90: here is us decoding directly the bytebuffer in benchmarks

zipkin/src/main/java/zipkin2/internal/JsonCodec.java
```diff
@@ -44,14 +44,14 @@
...
   // Hides gson types for internal use in other submodules
   public static final class JsonReader {
     final com.google.gson.stream.JsonReader delegate;
 
-    JsonReader(byte[] bytes) {
-      delegate =
-        new com.google.gson.stream.JsonReader(
-          new InputStreamReader(new ByteArrayInputStream(bytes), UTF_8));
+    JsonReader(ReadBuffer buffer) {
+      delegate = new com.google.gson.stream.JsonReader(new InputStreamReader(buffer, UTF_8));
```
codefromthecrypt, line 54: our buffer type now backports inputstream so it works directly with gson

zipkin/src/main/java/zipkin2/codec/SpanBytesDecoder.java
```diff
@@ -35,123 +37,180 @@
...
       return decodeList(this, spans);
     }
   };
 
+  /**
+   * ByteBuffer implementation of {@link #decodeList(byte[])}.
+   *
+   * <p>Note: only use this when it is ok to modify the underlying {@link ByteBuffer#array()}.
+   */
+  public abstract boolean decodeList(ByteBuffer spans, Collection<Span> out);
```
codefromthecrypt, line 189: adding the methods here allows us to not break the BytesDecoder interface, yet still allow us the overloads for our builtin types

codefromthecrypt:
With this change, our decoder is significantly faster and GC efficient also with bytebuffer.

```
ProtoCodecBenchmarks.bytebuffer_protobufDecoder:bytebuffer_protobufDecoder·p0.99    sample            2232.320                us/op
ProtoCodecBenchmarks.bytebuffer_protobufDecoder:·gc.alloc.rate.norm                 sample     15  2023331.587 ±     39.087    B/op
ProtoCodecBenchmarks.bytebuffer_wireDecoder:bytebuffer_wireDecoder·p0.99            sample            2330.624                us/op
ProtoCodecBenchmarks.bytebuffer_wireDecoder:·gc.alloc.rate.norm                     sample     15  2435472.372 ±     54.488    B/op
ProtoCodecBenchmarks.bytebuffer_zipkinDecoder:bytebuffer_zipkinDecoder·p0.99        sample            1964.032                us/op
ProtoCodecBenchmarks.bytebuffer_zipkinDecoder:·gc.alloc.rate.norm                   sample     15  1735287.795 ±     34.431    B/op
ProtoCodecBenchmarks.bytes_protobufDecoder:bytes_protobufDecoder·p0.99              sample            1949.696                us/op
ProtoCodecBenchmarks.bytes_protobufDecoder:·gc.alloc.rate.norm                      sample     15  2023208.424 ±     43.028    B/op
ProtoCodecBenchmarks.bytes_wireDecoder:bytes_wireDecoder·p0.99                      sample            2360.566                us/op
ProtoCodecBenchmarks.bytes_wireDecoder:·gc.alloc.rate.norm                          sample     15  2435409.404 ±     48.839    B/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:bytes_zipkinDecoder·p0.99                  sample            1810.596                us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:·gc.alloc.rate.norm                        sample     15  1735156.400 ±     30.538    B/op
```

codefromthecrypt:
results on JRE 11 are even better

```
ProtoCodecBenchmarks.bytebuffer_protobufDecoder:bytebuffer_protobufDecoder·p0.99    sample            1776.026                us/op
ProtoCodecBenchmarks.bytebuffer_protobufDecoder:·gc.alloc.rate.norm                 sample     15  2007317.528 ±     17.585    B/op
ProtoCodecBenchmarks.bytebuffer_wireDecoder:bytebuffer_wireDecoder·p0.99            sample            2129.469                us/op
ProtoCodecBenchmarks.bytebuffer_wireDecoder:·gc.alloc.rate.norm                     sample     15  1971521.040 ±     20.325    B/op
ProtoCodecBenchmarks.bytebuffer_zipkinDecoder:bytebuffer_zipkinDecoder·p0.99        sample            1586.258                us/op
ProtoCodecBenchmarks.bytebuffer_zipkinDecoder:·gc.alloc.rate.norm                   sample     15  1591251.722 ±     17.422    B/op
ProtoCodecBenchmarks.bytes_protobufDecoder:bytes_protobufDecoder·p0.99              sample            1520.026                us/op
ProtoCodecBenchmarks.bytes_protobufDecoder:·gc.alloc.rate.norm                      sample     15  1919189.226 ±     23.622    B/op
ProtoCodecBenchmarks.bytes_wireDecoder:bytes_wireDecoder·p0.99                      sample            1952.563                us/op
ProtoCodecBenchmarks.bytes_wireDecoder:·gc.alloc.rate.norm                          sample     15  1971395.380 ±     20.724    B/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:bytes_zipkinDecoder·p0.99                  sample            1371.259                us/op
ProtoCodecBenchmarks.bytes_zipkinDecoder:·gc.alloc.rate.norm                        sample     15  1591151.095 ±     14.826    B/op
```

codefromthecrypt:
here are relative read comparisons. clearly if someone wants most efficient client and server side, they should switch to proto3. I'm not sure the end to end on http+proto3 vs grpc, but I suspect either will be less overhead.

cc @dio @apache/zipkin-committers 

```
CodecBenchmarks.readChineseSpan_JSON_V1:readChineseSpan_JSON_V1·p0.99     sample             19.232             us/op
CodecBenchmarks.readChineseSpan_JSON_V1:·gc.alloc.rate.norm               sample      15  18879.162 ±  77.971    B/op
CodecBenchmarks.readChineseSpan_JSON_V2:readChineseSpan_JSON_V2·p0.99     sample             13.488             us/op
CodecBenchmarks.readChineseSpan_JSON_V2:·gc.alloc.rate.norm               sample      15  13976.995 ±   0.076    B/op
CodecBenchmarks.readChineseSpan_PROTO3:readChineseSpan_PROTO3·p0.99       sample              2.892             us/op
CodecBenchmarks.readChineseSpan_PROTO3:·gc.alloc.rate.norm                sample      15   2154.954 ±  20.872    B/op
CodecBenchmarks.readChineseSpan_THRIFT:readChineseSpan_THRIFT·p0.99       sample             11.423             us/op
CodecBenchmarks.readChineseSpan_THRIFT:·gc.alloc.rate.norm                sample      15   5805.938 ±  41.742    B/op
CodecBenchmarks.readClientSpan_JSON_V1:readClientSpan_JSON_V1·p0.99       sample             18.720             us/op
CodecBenchmarks.readClientSpan_JSON_V1:·gc.alloc.rate.norm                sample      15  19887.021 ±  33.413    B/op
CodecBenchmarks.readClientSpan_JSON_V2:readClientSpan_JSON_V2·p0.99       sample             13.680             us/op
CodecBenchmarks.readClientSpan_JSON_V2:·gc.alloc.rate.norm                sample      15  14345.030 ±   0.083    B/op
CodecBenchmarks.readClientSpan_PROTO3:readClientSpan_PROTO3·p0.99         sample              2.046             us/op
CodecBenchmarks.readClientSpan_PROTO3:·gc.alloc.rate.norm                 sample      15   1744.276 ±   0.020    B/op
CodecBenchmarks.readClientSpan_THRIFT:readClientSpan_THRIFT·p0.99         sample              8.608             us/op
CodecBenchmarks.readClientSpan_THRIFT:·gc.alloc.rate.norm                 sample      15   5065.300 ± 188.452    B/op
```

codefromthecrypt:
I double-checked the GC relating to span decoding and it is still good compared to moshi (what we use to read nested elasticsearch responses)

```
JsonCodecBenchmarks.bytebuffer_moshiDecoder:bytebuffer_moshiDecoder·p0.99      sample           6416.712                us/op
JsonCodecBenchmarks.bytebuffer_moshiDecoder:·gc.alloc.rate.norm                sample    15  4793152.590 ±    165.315    B/op
JsonCodecBenchmarks.bytebuffer_zipkinDecoder:bytebuffer_zipkinDecoder·p0.99    sample           4600.463                us/op
JsonCodecBenchmarks.bytebuffer_zipkinDecoder:·gc.alloc.rate.norm               sample    15  3234258.812 ±    103.979    B/op
JsonCodecBenchmarks.bytes_moshiDecoder:bytes_moshiDecoder·p0.99                sample           5892.997                us/op
JsonCodecBenchmarks.bytes_moshiDecoder:·gc.alloc.rate.norm                     sample    15  5240585.937 ±  16702.252    B/op
JsonCodecBenchmarks.bytes_zipkinDecoder:bytes_zipkinDecoder·p0.99              sample           5038.080                us/op
JsonCodecBenchmarks.bytes_zipkinDecoder:·gc.alloc.rate.norm                    sample    15  3234750.088 ±   2432.093    B/op
```

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
```diff
@@ -0,0 +1,379 @@
...
+import static zipkin2.internal.HexCodec.HEX_DIGITS;
+import static zipkin2.internal.JsonCodec.UTF_8;
+
+/** Read operations do bounds checks, as typically more errors occur reading than writing. */
+public abstract class ReadBuffer extends InputStream {
+  /** Do not use the buffer passed here after, as it may be manipulated directly. */
+  public static ReadBuffer wrapUnsafe(ByteBuffer buffer) {
+    if (buffer.hasArray()) {
+      int pos = buffer.position();
+      return wrap(buffer.array(), buffer.arrayOffset() + pos);
```
anuraaga: Think we should also look at `remaining()` - it can be efficient to use `slice`, in which case the buffer doesn't go to the end of the array.

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
```diff
@@ -0,0 +1,379 @@
...
+    }
+    return new ReadBuffer.Buff(buffer);
+  }
+
+  public static ReadBuffer wrap(byte[] bytes, int pos) {
```
anuraaga: The above will also require adding `length` here - `offset` and `length` are pretty idiomatic for these sort of buffers.

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    @Override final byte[] readBytes(int length) {
+      require(length);
+      byte[] copy = new byte[length];
+      buf.get(copy);
+      return copy;
+    }
+
+    @Override boolean tryReadAscii(char[] destination, int length) {
```
anuraaga, line 117: Since these are internal, I think we can follow up later but just a thought - I still want to do a deeper investigation on whether we should be using `char[]` for ASCII buffers. On JVM9+, I think `byte[]` should be faster since they internally store as `byte[]`.

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
```diff
@@ -0,0 +1,379 @@
...
+      buf.mark();
```
anuraaga, line 118: Think you want to either `require` or `return false` when there aren't enough bytes.

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
```diff
@@ -0,0 +1,379 @@
...
+          buf.reset();
+          return false;  // Not 7-bit ASCII character
+        }
+        destination[i] = (char) b;
+      }
+      return true;
+    }
+
+    @Override final String doReadUtf8(int length) {
+      return new String(readBytes(length), UTF_8);
```
anuraaga, line 131: `require(length)`

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    @Override int readInt() {
+      require(4);
+      return buf.getInt();
+    }
+
+    @Override long readLong() {
+      require(8);
+      return buf.getLong();
```
anuraaga: I think you need to check `buf.order()` for these two methods. Alternatively, if we're ok with having this buffer restricted to big endian (the default Java endianness), then it'd be good to throw `IllegalArgumentException` in the constructor when it doesn't match expectations

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
```diff
@@ -0,0 +1,379 @@
...
+    @Override public long skip(long maxCount) {
+      return skip(buf, maxCount);
+    }
+
+    @Override public int available() {
+      return buf.remaining();
+    }
+
+    static long skip(ByteBuffer buf, long maxCount) {
+      // avoid java.lang.NoSuchMethodError: java.nio.ByteBuffer.position(I)Ljava/nio/ByteBuffer;
```
anuraaga: Any more details on this? All the buffer methods seem to be available since 1.6

https://docs.oracle.com/javase/7/docs/api/java/nio/Buffer.html#position()

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
```diff
@@ -0,0 +1,379 @@
...
+        byte b = buf[pos + i];
+        if ((b & 0x80) != 0) return false;  // Not 7-bit ASCII character
+        destination[i] = (char) b;
+      }
+      pos += length;
+      return true;
+    }
+
+    @Override final String doReadUtf8(int length) {
+      String result = new String(buf, pos, length, UTF_8);
```
anuraaga: `require(length)`

zipkin/src/main/java/zipkin2/internal/WriteBuffer.java
```diff
@@ -369,86 +190,102 @@ void writeLongLe(long v) {
...
-
-    if ((b = readByte()) >= 0) {
-      return result | b << 21;
-    }
-    result |= (b & 0x7f) << 21;
-
-    b = readByte();
-    if ((b & 0xf0) != 0) {
-      throw new IllegalArgumentException("Greater than 32-bit varint at position " + (pos - 1));
+  public static int utf8SizeInBytes(CharSequence string) {
```
anuraaga, line 207: Might want to look at this one. It has a warning about being optimized. Main difference I see is for strings without surrogates, it has slightly less branches.

https://github.com/protocolbuffers/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/Utf8.java#L240

zipkin/src/main/java/zipkin2/internal/WriteBuffer.java
```diff
@@ -369,86 +190,102 @@ void writeLongLe(long v) {
...
+    return 5;
   }
 
-  void require(int byteCount) {
-    if (pos + byteCount > buf.length) {
-      throw new IllegalArgumentException(
-        "Truncated: length " + byteCount + " > bytes remaining " + remaining());
-    }
+  /** Like {@link #varintSizeInBytes(int)}, except for uint64. */
+  public static int varintSizeInBytes(long v) {
```
anuraaga, line 287: Can consider this one with less if's for 64-bit (the 32-bit version is the same as protobuf). Since trace IDs are random, I guess they cover the entire spectrum of varint sizes and probably would especially benefit from this.

https://github.com/protocolbuffers/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/CodedOutputStream.java#L770

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
codefromthecrypt: good idea

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
codefromthecrypt: renamed

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
codefromthecrypt: added a bit higher. thanks

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
codefromthecrypt: added higher (this is an internal method)

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
codefromthecrypt: nope you are right

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
codefromthecrypt: was an old JRE. I was thinking actually about not worrying about it. old code doesn't need to use this as it is read-side (write side is trickier)

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
codefromthecrypt: ditto

zipkin/src/main/java/zipkin2/internal/WriteBuffer.java
codefromthecrypt: added TODO

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
codefromthecrypt: dunno kindof doubt it because there is decoding going on in the CTOR as opposed to the char array one which does a plain copy. feel free to test it though

Ex from JDK 11
```
    public String(byte bytes[], int offset, int length) {
        checkBoundsOffCount(offset, length, bytes.length);
        StringCoding.Result ret = StringCoding.decode(bytes, offset, length);
        this.value = ret.value;
        this.coder = ret.coder;
    }
```
```

File: zipkin/src/main/java/zipkin2/internal/WriteBuffer.java (hunk 13 above)
Comment (codefromthecrypt) [reply zipkin/src/main/java/zipkin2/internal/WriteBuffer.java:207]: added TODO

File: zipkin/src/main/java/zipkin2/internal/WriteBuffer.java (hunk 14 above)
Comment (codefromthecrypt) [reply zipkin/src/main/java/zipkin2/internal/WriteBuffer.java:287]: another TODO!

**Comment by codefromthecrypt** [comment]
I will revert the utf decoder thing.. that can wait

File: zipkin/src/main/java/zipkin2/internal/ReadBuffer.java [hunk 15]
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    @Override long readLongLe() {
+      require(8);
+      return getLong(ByteOrder.LITTLE_ENDIAN);
+    }
+
+    long getLong(ByteOrder wanted) {
+      if (buf.order() == wanted) return buf.getLong();
+      return (readByteUnsafe() & 0xffL)
```
anuraaga: I think this always returns little endian, even if someone called readLong on a little endian buffer. Though maybe the TODO already expects the test to fail? ;)

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
anuraaga: Might be good to rename these two methods to `...Unsafe` to prevent wondering about it when reading this later.

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
```diff
@@ -0,0 +1,384 @@
...
+        if ((b & 0x80) != 0) {
+          buf.reset();
+          return false;  // Not 7-bit ASCII character
+        }
+        destination[i] = (char) b;
+      }
+      return true;
+    }
+
+    @Override final String doReadUtf8(int length) {
```
anuraaga, line 130: For reference, when working on https://github.com/protocolbuffers/protobuf/pull/3824 I couldn't get a `CharsetDecoder` to be faster than copying into an array like here. I can't remember if I tried using a threadlocal though to reuse the decoder.

Anyways, at some point we could think about whether it makes sense to copy over some Unsafe code to read the bytes directly.

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
```diff
@@ -0,0 +1,384 @@
...
+        | (buf[pos + 5] & 0xffL) << 16
+        | (buf[pos + 6] & 0xffL) << 8
+        | (buf[pos + 7] & 0xffL);
+    }
+
+    @Override long readLongLe() {
+      require(8);
+      int pos = this.offset;
+      this.offset = pos + 8;
+      return (buf[pos] & 0xffL)
```
anuraaga, line 226: It's hardly better, but could create a method that takes 8 `byte` parameters to share the logic between buf / array

anuraaga (approved):

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
codefromthecrypt: hah sorry this was sloppy of me. will fix.. I still owe some unit tests on this one

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
codefromthecrypt: didn't think about that either.. glad you are on the case!

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
codefromthecrypt: tests done

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
codefromthecrypt: I think I saw in gson code that using a local ref to an array is faster than field. I'll bench first about using that approach.

codefromthecrypt: I was wrong about local array at least in JDK8. If any of the other approaches are more performant probably need to use a different JDK to notice. closing this topic for now.. thanks!

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
codefromthecrypt: makes sense

codefromthecrypt:
gotta run. squash/merge if you're happy on green. thanks for all the help!

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
anuraaga: I think the local variable helps with a volatile field, but probably nothing much for non-volatile. Anyways, `reverseBytes` is cool, now we don't have the duplication so this looks good.

PR #2595, opened 2019-05-15
Decode ByteBuffer directly where possible.

anuraaga:
I recalled something I noticed a while ago but forgot to address - the armeria gzip decoder currently doesn't return pooled objects even if it's input is pooled. So we'll still have the array conversion for gzipped requests, but at least non-gzipped get to avoid it.

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
@@ -119,17 +119,17 @@ public class ZipkinHttpCollector {
       }
 
       try {
-        // logging already handled upstream in UnzippingBytesRequestConverter where request context exists
-        if (msg.content().isEmpty()) {
-          result.onSuccess(null);
-          return null;
-        }
-
         final HttpData content;
         try {
           content = UnzippingBytesRequestConverter.convertRequest(ctx, msg);
         } catch (IllegalArgumentException e) {
           result.onError(e);
+          return null;
+        }
+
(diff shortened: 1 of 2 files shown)


codefromthecrypt (approved):
sensible

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
```diff
@@ -139,7 +139,7 @@ HttpResponse validateAndStoreSpans(SpanBytesDecoder decoder, ServiceRequestConte
         } else {
           // Currently this will happen for gzip spans. Need to fix armeria's gzip decoder to allow
           // returning pooled buffers on request.
-          nioBuffer = ByteBuffer.wrap(content.array());
+          nioBuffer = ByteBuffer.wrap(content.array(), content.offset(), content.length());
```
codefromthecrypt, line 142: good eye :)

codefromthecrypt:
I compared this against the last release (2.13) running the same load test:11K span/second in 2 proto messages/second decode only (aka COLLECTOR_SAMPLE_RATE=0)

Before:
Screen Shot 2019-05-15 at 2 39 16 PM

Now:
Screen Shot 2019-05-15 at 2 30 16 PM

while this is not the only change involved, overall there's a heck of a lot less GC activity. bravo!

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
anuraaga: Filed https://github.com/line/armeria/issues/1771 too

Issue #2602, opened 2019-05-17
Correct license annotations for files that will be deleted soon


codefromthecrypt:
We eagerly want to delete zipkin-ui, as it has problems such as vendored files. There's an existing issue on this (#2569). Meanwhile, we should correct our license annotation of it.  There's another file (AbstractUnsafeUnaryGrpcService), but it will be deleted in a week or so.

> To ensure you are abiding by 3rd party license terms the full license text generally needs be included (e.g. MIT, BSD). Usually this is done by pointing a pointer to a file in the distribution in the LICENSE file (not a URL). Some projects put all of these licenses into a license directory. You might want to consider doing that.
> 
> Thanks,
Justin
> 
> 1. ./zipkin-2.13.0/zipkin-ui/libs/dagre-d3/js/dagre-d3.js
> 2. ./zipkin-2.14.0/zipkin-server/src/main/java/zipkin2/server/internal> /AbstractUnsafeUnaryGrpcService.java
> 3. ./zipkin-ui/libs/chosen/chosen.css
> 4. https://github.com/line/armeria/blob/master/NOTICE.txt

codefromthecrypt:
we solved this by re-prioritizing deletion of the files

PR #2604, opened 2019-05-18
Moves to Armeria 0.86

codefromthecrypt:
implicitly addresses one of the license nits in #2602


zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
```diff
@@ -213,7 +214,7 @@ static HttpData convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage
     String encoding = request.headers().get(HttpHeaderNames.CONTENT_ENCODING);
     HttpData content = request.content();
     if (!content.isEmpty() && encoding != null && encoding.contains("gzip")) {
-      content = GZIP_DECODER_FACTORY.newDecoder().decode(content);
+      content = GZIP_DECODER_FACTORY.newDecoder(ByteBufAllocator.DEFAULT).decode(content);
```
codefromthecrypt: @anuraaga @trustin this was a guess. Is it correct to do this or is there a request-specific means to choose the most relevant byte buff allocator?

trustin: You can use ctx.alloc() 😄

codefromthecrypt: nice figured it was something like that!

anuraaga (approved):

Issue #2609, opened 2019-05-23
zipkin-server cannot work in IDEA , but when i use " java -jar zipkin.jar " in windows cmd , it works.


ruansheng8:
![image](https://user-images.githubusercontent.com/24709538/58224174-2fd6b780-7d4f-11e9-8311-fe173e682366.png)

### **Start log in IDEA :**

```
:: Powered by Spring Boot ::         (v2.1.5.RELEASE)

2019-05-23 11:29:51.915  INFO 12864 --- [           main] z.s.ZipkinServer                         : Starting ZipkinServer on YY-shengR with PID 12864 (D:\IdeaProjects\incubator-zipkin\zipkin-server\target\classes started by ruansheng in D:\IdeaProjects\incubator-zipkin)
2019-05-23 11:29:51.918  INFO 12864 --- [           main] z.s.ZipkinServer                         : The following profiles are active: shared
2019-05-23 11:29:53.394  INFO 12864 --- [           main] c.l.a.c.Flags                            : com.linecorp.armeria.verboseExceptions: false (default)
2019-05-23 11:29:53.395  INFO 12864 --- [           main] c.l.a.c.Flags                            : com.linecorp.armeria.verboseSocketExceptions: false (default)
2019-05-23 11:29:53.395  INFO 12864 --- [           main] c.l.a.c.Flags                            : com.linecorp.armeria.verboseResponses: false (default)
2019-05-23 11:29:53.398  INFO 12864 --- [           main] c.l.a.c.Flags                            : com.linecorp.armeria.useEpoll: false (default)
2019-05-23 11:29:53.497  INFO 12864 --- [           main] c.l.a.c.Flags                            : com.linecorp.armeria.useOpenSsl: true (default)
2019-05-23 11:29:53.497  INFO 12864 --- [           main] c.l.a.c.Flags                            : com.linecorp.armeria.dumpOpenSslInfo: false (default)
2019-05-23 11:29:53.497  INFO 12864 --- [           main] c.l.a.c.Flags                            : com.linecorp.armeria.maxNumConnections: 2147483647 (default)
2019-05-23 11:29:53.497  INFO 12864 --- [           main] c.l.a.c.Flags                            : com.linecorp.armeria.numCommonWorkers: 24 (default)
2019-05-23 11:29:53.497  INFO 12864 --- [           main] c.l.a.c.Flags                            : com.linecorp.armeria.numCommonBlockingTaskThreads: 200 (default)
[...]

PR #2611, opened 2019-05-23
Handle exceptions without message.

anuraaga:
While this doesn't address the root issue in #2609 it is needed to prevent a proper exception being turned into a NPE.

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
@@ -241,12 +241,12 @@ final class BodyIsExceptionMessage implements ExceptionHandlerFunction {
   public HttpResponse handleException(RequestContext ctx, HttpRequest req, Throwable cause) {
     ZipkinHttpCollector.metrics.incrementMessagesDropped();
 
-    LOGGER.warn("Unexpected error handling request.", cause);
-
     String message = cause.getMessage() != null ? cause.getMessage() : "";
     if (cause instanceof IllegalArgumentException) {
       return HttpResponse.of(BAD_REQUEST, MediaType.ANY_TEXT_TYPE, message);
     } else {
+      LOGGER.warn("Unexpected error handling request.", cause);
+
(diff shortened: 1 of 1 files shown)


zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
```diff
@@ -235,13 +235,19 @@ static HttpData convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage
 
 final class BodyIsExceptionMessage implements ExceptionHandlerFunction {
 
+  static final Logger LOGGER = LogManager.getLogger();
+
   @Override
   public HttpResponse handleException(RequestContext ctx, HttpRequest req, Throwable cause) {
     ZipkinHttpCollector.metrics.incrementMessagesDropped();
+
+    LOGGER.warn("Unexpected error handling request.", cause);
```
shakuzen: I'm not sure we want to add this logging in all cases. It seems too verbose to me for bad requests.

shakuzen (commented):
We definitely need the change to prevent a NPE; I'm not so sure about the additional logging

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
anuraaga: Good point, moved the logging so it's for cases that aren't bad requests, which we probably do need to see for debugging issues.

shakuzen (approved):

codefromthecrypt:
tribal knowledge deserving to be in the readme.

yep we tend to use debug or finer for anything that can end up on request
scope. this limits info/warn to things lifecycle in nature. metrics should
indicate when to turn up logging.

currently our exception metrics to degree exist aren't likely fancy. I
remember finagle having the last 3 causes or something in the metric name
which is handy for scala where there is a different type for everything
maybe less so java excepting that it can identify NPE per second heh.

anyway sorry about the curt message earlier.. blipped online and will soon
be stably so. thanks for the help!

Issue #2610, opened 2019-05-23
Refactor GlobalSearch component


tacigar:
Currently, GlobalSearch component is implemented too complicated.
I think before #2475 we need to refactor whole of this component.

Issue #2485, opened 2019-04-03
Clearly separate autocomplete tags in lens dropdown


drolando:
It's confusing why some tags would show up in the drop down and some wouldn't since right now everything's just mixed.

It'd be nice if we could split the list in 2 sections and move the autocomplete tags to the 2nd. With ideally a title that explains that only some whitelisted tags are there and they can use the generic "tags" search for everything else.

drolando:
This is @tacigar's  comment on what'd be needed:

> I use react-select(https://github.com/jedwatson/react-select) to implement global-search component. (Please see https://github.com/openzipkin/zipkin/blob/7ac793e1261c4e03b1f1066cf9718c6c7c740d18/zipkin-lens/src/components/GlobalSearch/SearchCondition.js#L132-L169)
This component library is not very extensible.
So I think it is hard to add separator and descriptions to the dropdown as long as we use it.
It seems that we need to throw away react-select and create similar component ourself…
But it is a bit hard...

codefromthecrypt:
I agree that for existing sites, they may be confused about which properties of the span are by default searchable. If we want to optimize for this, we could have a separate element for tags including predefined keys or blank to enter your own. Remember also that predefined should one day include site-specific tags that we don't auto-complete values for! ex antennae-id or request id.

on hover it could say how to control this.

codefromthecrypt:
what if tags had a tag: or tag. prefix? not necessarily elegant but it
could reduce confusion perhaps. other data are top level

codefromthecrypt:
I mean ex duration is really span.duration and name is span.name tag is
span.tags.foo

possibly namespacing could clarify, skipping the span. part

codefromthecrypt:
I think haystack doesn't have this problem because you cannot search on and
hoc values iirc

Issue #2451, opened 2019-03-19
Use CSS framework in lens


tacigar:
Currently, most of the components are so special that CSS framework can not be used as it is. 
So I thought we did not need to use CSS framework.

However, it is expected that more general purpose components will added in the future.
So I'm going to begin to use CSS framework partially to develop such general components efficiently.

@zeagord Thank you for your great idea!!

zeagord:
I think it is nice to use the css frameworks such as semantic-ui or material-ui. It will quicken the development and visually pleasing as well. Let me know if you need any help from my end.

tacigar:
We use material-ui now.

shakuzen:
Just for tracking purposes, I believe it was #2633 that introduced Material UI

PR #2613, opened 2019-05-24
[WIP] Refactor global search component

tacigar:
Related: #2610 

**Work in progress**


tacigar:
We use **React-Select** in GlobalSearch component.

React-select is a nice component library.
But, I think it is unsuitable for usage like GlobalSearch in the following points.

* In GlobalSearch, react-select's style needs to be modified a lot for design. I think it is not a beautiful way.
* As mentioned in #2485 , we want to display the dropdown menu more rich, but it seems impossible by using react-select.
* I want to control **focus** completely for providing more intuitive operation, but react-select black-boxes focus a bit...

So, I want to create a new **Select** component.

tacigar:
I think now that we should use the desgin(CSS) Framework (related: #2451).
There are a lot of frameworks, but I want to use Material-UI (https://material-ui.com/) especially.
So I decided to introduce it within this GlobalSearch component at first in this PR.

codefromthecrypt:
@tacigar I am sorry that the switch back to openzipkin will make this conflict a lot. If you want me to fix your branch and rebase I can. otherwise you will need to do this prior to merge.

Note: for license update, run: `./mvnw com.mycila:license-maven-plugin:format`

tacigar:
OK!
Thank you for your information.
I'll update license myself 🙇

tacigar:
Sorry, I'll close this PR and make an other PR.

PR #2618, opened 2019-05-28
[zipin-lens] Fix #2615: Bundle Fontawesome into the app

### LINKED ISSUE #2615: Fontawesome is loaded from CDN, so Zipkin Lens doesn't work well offline
## Describe the Bug
index.html directly refers to fontawesome via CDN

## Expected Behaviour
Fontawesome should be an NPM dependency, and packed with WebPack. I would also recommend using Create React App, rather than doing all the build script stuff yourselves. It's much easier, IMO!

abesto:
On #2615 the idea of migrating to Create React App was discussed. That feels like a bigger undertaking, I took the more direct route to fixing this issue.

Note: https://github.com/webpack-contrib/extract-text-webpack-plugin says " Since webpack v4 the extract-text-webpack-plugin should not be used for css. Use mini-css-extract-plugin instead.", and we're on Webpack v4, so I did that.

#### Dev mode

```
cd zipkin-lens
rm -rf node_modules
npm install
npm install --only=dev
npm run start
```

http://localhost:8000 now looks like this:

2019-05-28-231159_478x580

#### Production mode

```
cd zipkin-lens
rm -rf node_modules
npm install
npm install --only=dev
rm -rf target
npm run build
cd target/classes
```

Notice that we now have a new directory with the fonts files (the `file-loader` specified for the fonts filed can be replaced with a `url-loader` to inline the fonts into CSS if preferred):

```
$ tree zipkin-lens
zipkin-lens/
├── app-80c52891f798f53ed6c5.min.js
├── favicon.ico
├── index.html
├── main.80c52891f798f53ed6c5.css
└── webfonts
    ├── fa-brands-400-921f1150167369cf4c400135a4905728.woff2
    ├── fa-brands-400-9d56249d09070f656a1c52e566448f0d.woff
    ├── fa-brands-400-ab673bac3a9e36cb1d4e54777c413fe3.ttf
    ├── fa-brands-400-ab6ad0116ae2dfcfb5f850c9d0fc363f.svg
    ├── fa-brands-400-f2594ef62455697f61dc99862c19afba.eot
    ├── fa-solid-900-1dc5b6dd4bf409a6f919be38603f76a0.woff2
    ├── fa-solid-900-20b351a6af2d523589fd193785e7d7f0.eot
    ├── fa-solid-900-63726a69fa60cb67459140ccaf679f96.ttf
    ├── fa-solid-900-997f6b1819184c54248de33c81b7bb0c.woff
    └── fa-solid-900-ab6ad0116ae2dfcfb5f850c9d0fc363f.svg

1 directory, 14 files
```

Now let's set up a simple webserver so we can look at the output:

```
mv zipkin-lens zipkin
python3 -m http.server
```

http://localhost:8000/zipkin now looks like this:

2019-05-28-231828_471x573


zipkin-lens/scss/main.scss
```diff
@@ -15,10 +15,15 @@
 // limitations under the License.
 //
 
-@import '../node_modules/rc-calendar/assets/index.css';
-@import '../node_modules/rc-slider/assets/index.css';
-@import '../node_modules/rc-time-picker/assets/index.css';
-@import '../node_modules/react-table/react-table.css';
+@import '~rc-calendar/assets/index.css';
```
abesto, line 18: `~` is Webpack notation for "load from `node_modules`, wherever it is"

zipkin-lens/webpack.prod.config.js
```diff
@@ -40,24 +46,12 @@ module.exports = {
...
-              options: {
-                sourceMap: true,
-              },
-            },
-          ],
-        }),
+        use: [
+          { loader: MiniCssExtractPlugin.loader },
+          { loader: 'css-loader', options: { sourceMap: true } },
+          { loader: 'resolve-url-loader' },
```
abesto, line 52: This is needed for some magical Webpack path wrangling that, I'll admit, I only mostly understand. Either way, it makes the site go. Something along the lines of: it replaces `~@fortawesome/...woff2` with wherever the files end up in the actual output maybe, but also Webpack doesn't find the source files without this I think. Didn't dig deeper; it works, I commits :)

zipkin-lens/webpack.prod.config.js
```diff
@@ -40,24 +46,12 @@ module.exports = {
       },
       {
         test: /\.scss$/,
-        use: ExtractTextPlugin.extract({
-          fallback: 'style-loader',
-          use: [
-            {
-              loader: 'css-loader',
-              options: {
-                url: false,
```
abesto, line 49: The `url` option of `css-loader` controls whether `url()` values in CSS are processed. Not sure why we had `url: false`, but we definitely need `url: true` to get the font paths into the Webpack pipeline (both for URL rewriting, and to actually output the font files into the target directory)

zipkin-lens/webpack.prod.config.js
```diff
@@ -66,6 +60,14 @@ module.exports = {
           minimize: true,
         },
       },
+      {
+        test: /webfonts\/.*\.(svg|woff|woff2|ttf|eot)$/,
+        loader: "file-loader",
```
abesto, line 65: This can be replaced with `url-loader` if `url:` base64 inlined fonts are preferred. I prefer my fonts in separate files, but up to you.

zipkin-lens/static/index.html
```diff
@@ -21,7 +21,6 @@
   <head>
     <base href="/" />
     <meta charset="utf-8">
-    <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.4.1/css/all.css" integrity="sha384-5sAR7xN1Nv6T6+dT2mhtzEpVJvfS3NScPQTrOxhwjIuvcA67KV2R5Jz6kr4abQsz" crossorigin="anonymous">
```
abesto, line 24: Bye-bye fonts loaded from the internets!

devinsba:
Just something to keep in mind, I believe we will need to include the FontAwesome license in our various files, disclaimer...

codefromthecrypt:
@devinsba as we aren't adding fontawesome to the source tree (vendoring like we did in zipkin-ui), and what we are verifying is a source distribution, not a binary one.. we don't need to add to the LICENSE file (FWIW I added notes about vendored files to DISCLAIMER once and was told to move them to LICENSE as DISCLAIMER is about us being not a real ASF project yet).

https://github.com/apache/incubator-zipkin/issues/2548 should be the one for fontawesome and other things that are dependencies in the lens build iiuc. IOTW, DEPENDENCIES is the drone you're looking for

codefromthecrypt:
Thanks tons for the effort here @abesto 

@pete-woods can you try a build off this branch. Ex

```bash
$ TAG=042c201fe01eb5d8bd88127b7730f5795743c655
$ curl -sSL https://jitpack.io/com/github/apache/incubator-zipkin/zipkin-server/${TAG}/zipkin-server-${TAG}-exec.jar > zipkin.jar
```

codefromthecrypt:
@tacigar PTAL

pete-woods:
From looking at the diff, the changes are exactly what I was looking for.

tacigar:
Sorry for too late review!!
I'll review this PR tomorrow morning, because I don't have a development environment now... :sob:

tacigar:
LGTM!!
Thank you 👍

codefromthecrypt (approved):
really happy about this, thanks Zoltán!

PR #2623, opened 2019-06-04
Clarifies storage versions supported vs tested

codefromthecrypt:


README.md
```diff
@@ -120,20 +126,29 @@ data layouts based on Zipkin's V1 Thrift model, as opposed to the
...
+features, but is tested against MariaDB 10.3.
+
+The schema was designed to be easy to understand and get started with;
+it was not designed for performance. Ex spans fields are columns, so
 you can perform ad-hoc queries using SQL. However, this component has
 [known performance issues](https://github.com/apache/incubator-zipkin/issues/1233): queries will eventually take seconds to return
 if you put a lot of data into it.
 
+This store does not require a [job to aggregate](https://github.com/apache/incubator-zipkin-dependencies) dependency links.
+However, running the job will improve performance.
```
shakuzen: It might be good to clarify that the performance improvement is only for retrieving dependency links - that is, it won't improve overall performance of read/write on span data.

codefromthecrypt: great point

shakuzen (approved):

PR #2632, opened 2019-06-14
Fix ThrottledStorageComponent not passing through health check

lambcode:
When using the ThrottledStorageComponent the health check (/actuator/health) always reports "UP" even when the backing storage component is in a bad state.

This change passes through the check from the delegate. I also formalized the name of components so that instead of reporting on the "ThrottledStorageComponent" in the health endpoint, you will get a more useful name such as "Throttled(ElasticsearchStorage)".

### DIFF
--- a/zipkin-server/src/test/kotlin/zipkin2/server/internal/throttle/ThrottledStorageComponentTest.kt
+++ b/zipkin-server/src/test/kotlin/zipkin2/server/internal/throttle/ThrottledStorageComponentTest.kt
@@ -16,7 +16,9 @@ package zipkin2.server.internal.throttle
 import com.linecorp.armeria.common.metric.NoopMeterRegistry
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Test
+import org.mockito.Mockito.*
 import zipkin2.storage.InMemoryStorage
+import zipkin2.storage.StorageComponent
 
 class ThrottledStorageComponentTest {
   val delegate = InMemoryStorage.newBuilder().build()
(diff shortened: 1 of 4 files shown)


zipkin/src/main/java/zipkin2/Component.java
```diff
@@ -38,6 +38,11 @@ public CheckResult check() {
     return CheckResult.OK;
   }
 
+  public String name() {
```
codefromthecrypt: please revert this method as it is adding a public method to a base type that existed for a long time, for a single call site. We take public methods on base types more seriously here vs other projects.

zipkin-server/src/test/kotlin/zipkin2/server/internal/throttle/ThrottledStorageComponentTest.kt
```diff
@@ -16,7 +16,9 @@ package zipkin2.server.internal.throttle
 import com.linecorp.armeria.common.metric.NoopMeterRegistry
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Test
+import org.mockito.Mockito.*
```
codefromthecrypt: please no wildcard imports

codefromthecrypt (commented):
I like the main idea here.

I think it is a better choice to use `toString()` vs exposing something similar. Let's do that instead of the `name()` refactoring?

zipkin/src/main/java/zipkin2/Component.java
lambcode: I'm okay with this change. Just want to note that the name reported by the health endpoint will change for some components. E.g "InMemoryStorage" will become "InMemoryStorage{traceCount=123}"

zipkin-server/src/test/kotlin/zipkin2/server/internal/throttle/ThrottledStorageComponentTest.kt
lambcode: Fixed in latest commit

codefromthecrypt:
Ack. thanks for mentioning but I think this is the better optimization

codefromthecrypt:
thanks very much!

Issue #2340, opened 2018-12-31
[storage-elasticsearch] why all the es ports must be the same?


maoling:
read the [doc ](https://github.com/openzipkin/zipkin/tree/master/zipkin-storage/elasticsearch) to see this:
> You can alternatively supply a list of http base urls. This list is used to recover from failures. Note that all ports must be the same, and the scheme must be http, not https.

that means: **http://elasticsearch-1:9200,http://elasticsearch-2:19200** isn't  supprted?
if need this to be improved,I can give a PR!

codefromthecrypt:
currently, we use okhttp DNS trick to round robin across the multiple
hosts. if you know an alternative way that works with mixed ports feel feel
free to propose!

codefromthecrypt:
I can take a second look to see if there's a way

maoling:
Let me take a look

codefromthecrypt:
https://github.com/square/okhttp/issues/4530 will solve this the right way. meanwhile I will take a quick look to see if any way now (hack)

codefromthecrypt:
we are currently using okhttp3.Dns which is InetAddress, not InetSocketAddress. Sorry we cannot support mixed ports until some other feature like loadbalancer happens or someone else is more clever.

cc @swankjesse

maoling:
@adriancole 
- yes,if `okhttp3.Dns` use `InetSocketAddress`,the problem will be solved.
- the problem seems to be not very urgent,if really need,we can implement a simple `round robin` using the `InetSocketAddress` inside（：D）.
- I don't know why choose` okhttp3` which's more suitable for mobile development?

codefromthecrypt:
- I don't know why choose okhttp3 which's more suitable for mobile
   development?

We chose okhttp because it is a clean, well designed library, that keeps
api portability in mind. It is not exclusively targeted to mobile dev even
if they care about android.

codefromthecrypt:
everything works now. we changed the implementation. please try 2.16.2 or later!

PR #2653, opened 2019-07-08
Migrate from okhttp to armeria

### LINKED ISSUE #2646: Consider what to do about OkHttp switching to Kotlin
Currently, our elasticsearch module and server tests use OkHttp.. most important to end users is the elasticsearch module. I switched the version and it broke api a little

```diff
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
@@ -132,7 +132,7 @@ public final class HttpCall<V> extends Call.Base<V> {
 
   public static <V> V parseResponse(Response response, BodyConverter<V> bodyConverter)
     throws IOException {
-    if (!HttpHeaders.hasBody(response)) {
+    if (!HttpHeaders.promisesBody(response)) {
       if (response.isSuccessful()) {
         return null;
       } else {
```

More importantly, this adds yet another 2MiB as that's the size of the kotlin standard library. Tests, we can use anything, but we should be considerate about the actual server code.

asking @openzipkin/core Java maintainers an opinion on this. I suspect it is either continue and not upgrade to OkHttp 4, upgrade to OkHttp 4 and accept the kotlin dep, or switch to Armeria and remember that we have some integrations downstream like zipkin-aws signatures which would need to be redone if we chose that path.

anuraaga:
Fixes #2646

I haven't gotten to skim through this myself yet, so there's probably silly cleanups needed still, but feel free to take an initial look too if you have time. I also haven't run the integration tests yet.

In a followup PR, I'll probably migrate from moshi to jackson since we already depend on the latter through armeria / spring, so we can remove moshi and okio for a few hundred K of library savings. The fact that we don't natively use okio buffers now makes using moshi more complicated, and probably slower.


codefromthecrypt:
this is the second time you've written elasticsearch support now. this time
you dont need it yourself! many thanks for the help

codefromthecrypt:
ps I plan to dig deeper into this over the weekend because trying to finish
javascript and it is taking all my willpower to resist other things ;)

anuraaga:
Updated `zipkin-server` too and this is getting better, but one WIP is setting connect timeout, which isn't too hard but will bring back the `close` method on the storage which I thought wasn't needed by using Armeria's default clientfactory.

Also for the record, want to say it was quite painful updating the `zipkin-server` tests due to Kotlin. Not only is it unfamiliar and a big context switch, but I find it has quite strange syntax and quirks when integrating with normal Java libraries - for the life of me I couldn't get `assertThat(Throwable).satisfies { checks }` to even compile. So another post-humous vote against the Kotlin tests ;)

codefromthecrypt:
appreciate the feedback rag. it isnt an irreversible thing (kotlin tests)
though I am not sure if intellij will help the other direction or not. in
hindsight I think more may have thought it was a nice idea in abstract than
were directly impacted in a good or bad way. you are the first to feel
impact and while it might be water under the bridge.. reversing is an
option.

anuraaga:
V6 integration tests are working fine now, for some reason the armeria client can't connect to V7. The build passing is deceptive, since the V7 tests got skipped due to the connection failure will look more.

anuraaga:
Ok V7 should be all good too

zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/ZipkinElasticsearchStorageAutoConfiguration.java
```diff
@@ -59,126 +46,79 @@
...
-      @Override protected Runnable wrap(Runnable task) {
-        return RequestContext.current().makeContextAware(traceCtx.wrap(task));
-      }
+  @Bean @Qualifier(QUALIFIER) @ConditionalOnSelfTracing Consumer<HttpClientBuilder>
+  elasticsearchTracing(Optional<Tracing> tracing) {
+    if (!tracing.isPresent()) {
+      return client -> {};
     }
-    return new TracingCurrentRequestContextExecutorService();
+    return client -> client.decorator(BraveClient.newDecorator(tracing.get(), "elasticsearch"));
```
codefromthecrypt, line 160: cool

zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/ZipkinElasticsearchStorageAutoConfiguration.java
```diff
@@ -59,126 +46,79 @@
...
-      .writeTimeout(timeout, TimeUnit.MILLISECONDS).build();
+    return factory -> factory.connectTimeoutMillis(timeout);
   }
 
+
   @Bean @Qualifier(QUALIFIER) @Conditional(HttpLoggingSet.class)
-  Interceptor zipkinElasticsearchHttpLoggingInterceptor(ZipkinElasticsearchStorageProperties es) {
-    Logger logger = Logger.getLogger(ElasticsearchStorage.class.getName());
-    return new HttpLoggingInterceptor(logger::info).setLevel(es.getHttpLogging());
+  Consumer<HttpClientBuilder> zipkinElasticsearchHttpLogging(
```
codefromthecrypt, line 79: cool. was curious how you'd do this

zipkin-server/src/test/kotlin/zipkin2/elasticsearch/ZipkinElasticsearchStorageAutoConfigurationTest.kt
```diff
@@ -178,10 +195,10 @@ class ZipkinElasticsearchStorageAutoConfigurationTest {
     Access.registerElasticsearchHttp(context)
     context.refresh()
 
-    val client = context.getBean(OkHttpClient::class.java)
-    assertThat(client.connectTimeoutMillis()).isEqualTo(timeout)
-    assertThat(client.readTimeoutMillis()).isEqualTo(timeout)
-    assertThat(client.writeTimeoutMillis()).isEqualTo(timeout)
+    val storage = context.getBean(ElasticsearchStorage::class.java)
+    // TODO(anuraaga): Verify connect timeout after https://github.com/line/armeria/issues/1890
```
codefromthecrypt, line 199: thx

zipkin-server/src/test/kotlin/zipkin2/server/internal/elasticsearch/BasicAuthInterceptorTest.kt
```diff
@@ -13,29 +13,53 @@
...
+import java.util.concurrent.CompletionException
 
 class BasicAuthInterceptorTest {
-  @Rule @JvmField val mockWebServer = MockWebServer()
-  @Rule @JvmField val thrown: ExpectedException = ExpectedException.none()
 
-  var client: OkHttpClient = OkHttpClient.Builder()
-    .addNetworkInterceptor(BasicAuthInterceptor(ZipkinElasticsearchStorageProperties(false, 0)))
-    .build()
+  companion object {
```
codefromthecrypt, line 36: sorry.. this looks like it probably wasn't fun to figure out

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -341,41 +357,67 @@ IndexTemplates ensureIndexTemplates() {
...
+      .decorator(HttpDecodingClient.newDecorator());
+
+    clientCustomizer().accept(client);
+
+    return client.build();
   }
 
-  @Override
-  public void close() {
-    if (!shutdownClientOnClose()) return;
```
codefromthecrypt, line 377: intentional to remove the shutdownClientOnClose?

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -341,41 +357,67 @@ IndexTemplates ensureIndexTemplates() {
...
+          }
+        })
+        .collect(Collectors.toList());
+      List<Endpoint> endpoints = urls.stream()
+        .map(url -> Endpoint.parse(url.getAuthority()))
+        .collect(Collectors.toList());
+      EndpointGroup group = HttpHealthCheckedEndpointGroup.of(
+        new StaticEndpointGroup(endpoints), "/_cluster/health");
+      EndpointGroupRegistry.register("elasticsearch", group, EndpointSelectionStrategy.ROUND_ROBIN);
+      clientUrl = urls.get(0).getProtocol() + "://group:elasticsearch" + urls.get(0).getPath();
```
codefromthecrypt, line 464: does this now work with round-robin across different https and port endpoints? if so it solves very requested issues. Ex #2340 #1422


cc @gquintana @maoling

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/PseudoAddressRecordSet.java
```diff
@@ -1,284 +0,0 @@
...
-import java.util.Set;
-import okhttp3.Dns;
-import okhttp3.HttpUrl;
-import zipkin2.internal.Nullable;
-
-/**
- * This returns a Dns provider that combines the IPv4 or IPv6 addresses from a supplied list of
- * urls, provided they are all http and share the same port.
- */
-final class PseudoAddressRecordSet {
```
codefromthecrypt, line 31: wow forgot how much code this was..

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
```diff
@@ -13,17 +13,20 @@
  */
 package zipkin2.elasticsearch.internal.client;
 
-import java.io.Closeable;
+import com.fasterxml.jackson.databind.util.ByteBufferBackedInputStream;
```
codefromthecrypt: can we delay jackson to a later pull request with some of the okio utilities?

ex does one or the other here help?
https://github.com/square/okio/pull/346/files
https://github.com/square/okio/blob/d7fcee8db94177ac4d02e205f928d8425f352d52/okio/src/jvmTest/java/okio/NioTest.java

if not, add a TODO: to follow-up to swap out okio/moshi as we shouldn't really have multiple things doing the same if we can avoid it.

codefromthecrypt (commented):
Thanks for all the hard yards here. 

Main last mile will be making sure zipkin-aws elastcisearch is possible to refactor over the work here
https://github.com/openzipkin/zipkin-aws/tree/master/autoconfigure/storage-elasticsearch-aws/src/main/java/zipkin/autoconfigure/storage/elasticsearch/aws

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
maoling: Yes it is Literally.

anuraaga: Yeah it'll round robin across different ports no problem and will work with https. It's health-checked, so if elasticsearch is RED, requests won't be sent to those hosts even if the connection succeeds.

I realized from reading the README that we want to support domains that return multiple IP records. So I've gone ahead and added `DnsAddressEndpointGroup` as well so A records for a domain lookup will all be rotated over. To do it generically, it added quite a bit of complexity here.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
anuraaga: Yeah without customization, we use the default `ClientFactory`, and it's a no-op to close it so no need for us to worry.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
anuraaga: Sure - fixed the bug in `ReadBuffer` and used it

anuraaga (commented):
First Kotlin weirdness, now weirdest Armeria issue ever https://github.com/line/armeria/issues/1895

I've learned my lesson to not call issues straightforward without care ;)

codefromthecrypt:
build is green apart from being too long which made travis kill the job

codefromthecrypt:
Thanks for all the work on this, Rag.

I started ES 6.8 then ran zipkin off this branch like so:

```bash
$ curl -sSL https://jitpack.io/com/github/anuraaga/zipkin/zipkin-server/elasticsearch-armeria-SNAPSHOT/zipkin-server-elasticsearch-armeria-SNAPSHOT-exec.jar > zipkin.jar
$ STORAGE_TYPE=elasticsearch ES_HTTP_LOGGING=BODY java -jar zipkin.jar
```

I then hit the health check endpoint like this:
```bash
$ curl -s localhost:9411/health
```

The health check hasn't returned a value (hung). I looked at the server logs and see the requests made (though it isn't unrolling the gzipped response into the console which iirc the old code did)

```
2019-07-12 09:31:11.045  INFO 34688 --- [-worker-nio-2-5] c.l.a.c.l.LoggingClient                  : Request: {startTime=2019-07-12T00:31:11.035Z(1562891471035865), length=0B, duration=7805µs(7805101ns), scheme=none+h1c, headers=[:method=GET, :path=/_cluster/health/zipkin*span-*, :authority=localhost:9200, :scheme=http, accept-encoding=gzip,deflate, user-agent=armeria/0.88.0]}
2019-07-12 09:31:11.071  INFO 34688 --- [-worker-nio-2-5] c.l.a.c.l.LoggingClient                  : Response: {startTime=2019-07-12T00:31:11.070Z(1562891471070979), length=234B, duration=372µs(372805ns), headers=[:status=200, content-type=application/json; charset=UTF-8, content-encoding=gzip, content-length=234]}
```

anuraaga:
Thanks - fixed the logging and found that zipkin-server itself had a somewhat severe bug since our Armeria migration. The health endpoint returned `HttpResponse` instead of `AggregatedHttpResponse` - this was causing it to run the service on the event loop, which is a big no-no since the health check is a blocking call. It means that users currently see poor latency for 1 / 2 * # of CPUs of requests during the processing of a health check.

And the reason this caused the health request to hang is that the event loop was trying to aggregate the client response on itself, though since it is blocked, this never happens and causes it to deadlock.

anuraaga (commented):
@trustin @minwoox you may be interested in skimming through this PR. A couple of observations I had

- The annotated service documentation probably needs to be updated to call out that you cannot do blocking calls if returning `HttpResponse` or `CompletionStage`

- I'm wondering if the functionality of `RawContentLoggingClient` should be in upstream somehow.

codefromthecrypt:
I am impressed with how quick you sorted this. thanks a lot! I will give a
spin again tomorrow.

anuraaga:
I find sometimes endpoint resolution takes a couple seconds on startup when recreating storage a lot - I think it's related to https://github.com/line/armeria/pull/1899 but I noticed there doesn't seem to be a real need to recreate it everytime since we clear the indexes before each test anyways. The integration tests go much faster now. I wonder if the same applied to ITCassandraStorage would also make that go faster. In any event, not closing the storage during integration tests seems like it can clog things up in general.

codefromthecrypt:
On Fri, Jul 12, 2019 at 9:50 PM Anuraag Agrawal <notifications@github.com>
wrote:

> I find sometimes endpoint resolution takes a couple seconds on startup
> when recreating storage a lot - I think it's related to line/armeria#1899
> <https://github.com/line/armeria/pull/1899> but I noticed there doesn't
> seem to be a real need to recreate it everytime since we clear the indexes
> before each test anyways. The integration tests go much faster now. I
> wonder if the same applied to ITCassandraStorage would also make that go
> faster
>

Thanks! yeah it could be the underlying concern with the cassandra tests
also. and we'd get double speedup if we improved them :P

anuraaga:
Tried doing similar for Cassandra, but it seems to not be as well behaved - I think it deletes tables asychronously, meaning it can happen in the middle of a test, so those tests seem to depend on the per-test keyspace behavior, so the storage also has to be per-test.

codefromthecrypt:
PS sorry I got pulled away as someone found a massive bug around local
roots in brave https://github.com/openzipkin/brave/pull/945

I'll check out your awesome work again after dinner!

codefromthecrypt:
We have a crappy toString which is not your fault, as it preceded this change.

```bash
curl -s localhost:9411/health|jq .
{
  "status": "UP",
  "zipkin": {
    "status": "UP",
    "details": {
      "ElasticsearchStorage{clientCustomizer=zipkin2.server.internal.elasticsearch.ZipkinElasticsearchStorageAutoConfiguration$$Lambda$318/0x0000000801468c40@19f040ba, clientFactoryCustomizer=zipkin2.server.internal.elasticsearch.ZipkinElasticsearchStorageAutoConfiguration$$Lambda$319/0x0000000801469040@72ab05ed, hostsSupplier=[http://localhost:9200], pipeline=null, flushOnWrites=false, maxRequests=64, strictTraceId=true, searchEnabled=true, autocompleteKeys=[], autocompleteTtl=3600000, autocompleteCardinality=20000, indexShards=5, indexReplicas=1, indexNameFormatter=IndexNameFormatter{index=zipkin, dateSeparator=-, dateFormat=zipkin2.elasticsearch.internal.IndexNameFormatter$Builder$1@27e32fe4}, namesLookback=86400000}": {
        "status": "UP"
      }
    }
  }
}
```

We should probably override this with a couple key fields similarly to how we do in the zipkin reporter project ex.

```java
  @Override public final String toString() {
    return "KafkaSender{"
        + "bootstrapServers=" + properties.get(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG)
        + ", topic=" + topic
        + "}";
  }
```

codefromthecrypt:
so I'm using wrk inside docker to toy around. https://github.com/openzipkin/docker-zipkin/pull/212

what I do is..

```bash
docker-compose -f docker-compose.yml -f docker-compose-elasticsearch.yml build --no-cache
docker-compose -f docker-compose.yml -f docker-compose-elasticsearch.yml up
```

Then, I run the frontend and backend apps of brave-webmvc-example (though any example would work).

And I do `curl -s localhost:8081` against the frontend, which a bit later results in a report of span from each of them.
```bash
zipkin                      | 2019-07-13 10:14:11.348  INFO 1 --- [search:9200/...] z.e.ElasticsearchStorage                 : <-- 200 OK http://elasticsearch:9200/_bulk (675ms, 259-byte body)
zipkin                      | 2019-07-13 10:14:11.353  INFO 1 --- [search:9200/...] z.e.ElasticsearchStorage                 : <-- 200 OK http://elasticsearch:9200/_bulk (669ms, 227-byte body)
```

Then, I escalate to using `wrk -t4 -c128 -d30s http://localhost:8081 --latency` and that would create a ton of spans.. there are timeouts every time.

```bash
zipkin                      | 2019-07-13 10:14:25.468  INFO 1 --- [search:9200/...] z.e.ElasticsearchStorage                 : --> POST http://elasticsearch:9200/_bulk http/1.1 (-1-byte body)
zipkin                      | 2019-07-13 10:14:25.725  INFO 1 --- [search:9200/...] z.e.ElasticsearchStorage                 : --> POST http://elasticsearch:9200/_bulk http/1.1 (-1-byte body)
zipkin                      | 2019-07-13 10:14:27.493  INFO 1 --- [search:9200/...] z.e.ElasticsearchStorage                 : --> POST http://elasticsearch:9200/_bulk http/1.1 (-1-byte body)
zipkin                      | 2019-07-13 10:14:29.851  INFO 1 --- [search:9200/...] z.e.ElasticsearchStorage                 : --> POST http://elasticsearch:9200/_bulk http/1.1 (-1-byte body)
zipkin                      | 2019-07-13 10:14:29.965  INFO 1 --- [search:9200/...] z.e.ElasticsearchStorage                 : --> POST http://elasticsearch:9200/_bulk http/1.1 (-1-byte body)
zipkin                      | 2019-07-13 10:14:34.004  INFO 1 --- [search:9200/...] z.e.ElasticsearchStorage                 : --> POST http://elasticsearch:9200/_bulk http/1.1 (-1-byte body)
zipkin                      | 2019-07-13 10:14:34.542  INFO 1 --- [search:9200/...] z.e.ElasticsearchStorage                 : --> POST http://elasticsearch:9200/_bulk http/1.1 (-1-byte body)
zipkin                      | 2019-07-13 10:14:36.025  INFO 1 --- [search:9200/...] z.e.ElasticsearchStorage                 : --> POST http://elasticsearch:9200/_bulk http/1.1 (-1-byte body)
zipkin                      | 2019-07-13 10:14:36.618  INFO 1 --- [search:9200/...] z.e.ElasticsearchStorage                 : <-- HTTP FAILED: java.net.SocketTimeoutException: timeout
zipkin                      | 2019-07-13 10:14:37.215  INFO 1 --- [search:9200/...] z.e.ElasticsearchStorage                 : --> POST http://elasticsearch:9200/_bulk http/1.1 (-1-byte body)
zipkin                      | 2019-07-13 10:14:38.720  INFO 1 --- [search:9200/...] z.e.ElasticsearchStorage                 : --> POST http://elasticsearch:9200/_bulk http/1.1 (-1-byte body)
zipkin                      | 2019-07-13 10:14:38.784  INFO 1 --- [search:9200/...] z.e.ElasticsearchStorage                 : <-- HTTP FAILED: java.net.SocketTimeoutException: timeout
zipkin                      | 2019-07-13 10:14:39.799  INFO 1 --- [search:9200/...] z.e.ElasticsearchStorage                 : --> POST http://elasticsearch:9200/_bulk http/1.1 (-1-byte body)
zipkin                      | 2019-07-13 10:14:40.440  INFO 1 --- [search:9200/...] z.e.ElasticsearchStorage                 
--snip--
```

I'm wondering if there's some issue related to bulk requests or possibly concurrent requests that's causing this.

anuraaga:
Those logs don't look like the output of Armeria's request logs which look more like what you posted a couple days ago (notably logger name is `c.l.a.c.l.LoggingClient` - can you confirm you're using the branch version and not master with that test?

codefromthecrypt:
I added `--no-cache` to the docker-compose instructions just in case.. I will do it with the exact commit id if we think somehow jitpack is stale..

codefromthecrypt:
sorry will check!

anuraaga:
Thanks :) FWIW, ran a simple stress test via `ITElasticStorage` (hence log4j2.properties is different from normal server) on master and compared the logs look closer to those ones which must be OKHTTP logs

```
INFO: --> POST http://localhost:32800/_bulk?refresh=wait_for (-1-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: <-- 200 OK http://localhost:32800/_bulk?refresh=wait_for (3970ms, 249038-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: <-- 200 OK http://localhost:32800/_bulk?refresh=wait_for (3977ms, 248851-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: --> POST http://localhost:32800/_bulk?refresh=wait_for (-1-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: --> POST http://localhost:32800/_bulk?refresh=wait_for (-1-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: --> POST http://localhost:32800/_bulk?refresh=wait_for (-1-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: --> POST http://localhost:32800/_bulk?refresh=wait_for (-1-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: <-- 200 OK http://localhost:32800/_bulk?refresh=wait_for (4031ms, 248631-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: <-- 200 OK http://localhost:32800/_bulk?refresh=wait_for (3956ms, 248455-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: --> POST http://localhost:32800/_bulk?refresh=wait_for (-1-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: <-- 200 OK http://localhost:32800/_bulk?refresh=wait_for (4043ms, 248522-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: --> POST http://localhost:32800/_bulk?refresh=wait_for (-1-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: <-- 200 OK http://localhost:32800/_bulk?refresh=wait_for (4064ms, 249038-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: --> POST http://localhost:32800/_bulk?refresh=wait_for (-1-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: <-- 200 OK http://localhost:32800/_bulk?refresh=wait_for (4028ms, 249038-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: --> POST http://localhost:32800/_bulk?refresh=wait_for (-1-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: <-- 200 OK http://localhost:32800/_bulk?refresh=wait_for (4072ms, 249038-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: --> POST http://localhost:32800/_bulk?refresh=wait_for (-1-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: --> POST http://localhost:32800/_bulk?refresh=wait_for (-1-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: <-- 200 OK http://localhost:32800/_bulk?refresh=wait_for (4138ms, 249038-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: --> POST http://localhost:32800/_bulk?refresh=wait_for (-1-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: <-- 200 OK http://localhost:32800/_bulk?refresh=wait_for (4297ms, 248647-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: --> POST http://localhost:32800/_bulk?refresh=wait_for (-1-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: <-- 200 OK http://localhost:32800/_bulk?refresh=wait_for (4180ms, 248849-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: <-- 200 OK http://localhost:32800/_bulk?refresh=wait_for (4334ms, 248827-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: --> POST http://localhost:32800/_bulk?refresh=wait_for (-1-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: <-- 200 OK http://localhost:32800/_bulk?refresh=wait_for (4341ms, 248816-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: --> POST http://localhost:32800/_bulk?refresh=wait_for (-1-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: --> POST http://localhost:32800/_bulk?refresh=wait_for (-1-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: <-- 200 OK http://localhost:32800/_bulk?refresh=wait_for (4246ms, 248712-byte body)
Jul 13, 2019 8:21:42 PM okhttp3.internal.platform.Platform log
INFO: --> POST http://localhost:32800/_bulk?refresh=wait_for (-1-byte body)
```

codefromthecrypt:
the problem was that I looked at stats and 100% of the bulk requests failed. I'll lock in the commit and try again though.

codefromthecrypt:
ok I rebuilt using the commit tag

```diff
--- a/zipkin/Dockerfile
+++ b/zipkin/Dockerfile
@@ -16,7 +16,7 @@ FROM alpine
 
 ENV USER anuraaga
 ENV BRANCH elasticsearch-armeria
-ENV VERSION SNAPSHOT
+ENV VERSION 76263f9bd0-1
 
 # Add environment settings for supported storage types
 COPY zipkin/ /zipkin/

```

from here https://jitpack.io/com/github/anuraaga/zipkin/elasticsearch-armeria-76263f9bd0-1/build.log

I'll do the test again now..

codefromthecrypt:
definitely I had an old version before.. maybe because I didn't do `--no-cache` at first, but anyway I will only test with commit IDs now to reduce distraction.

I'm sure you are right by the way, they were okhttp logs!

Definitely the logs look different
```
zipkin                      | 2019-07-13 11:34:10.080  INFO 1 --- [orker-epoll-4-6] c.l.a.c.l.LoggingClient                  : Request: {startTime=2019-07-13T11:34:10.074Z(1563017650074736), length=3739KiB(3829669B), duration=5635?s(5635700ns), scheme=none+h1c, headers=[]}
zipkin                      | 2019-07-13 11:34:10.313  INFO 1 --- [orker-epoll-4-7] c.l.a.c.l.LoggingClient                  : Request: {startTime=2019-07-13T11:34:10.311Z(1563017650311478), length=3397KiB(3479449B), duration=1783?s(1783500ns), scheme=none+h1c, headers=[]}
zipkin                      | 2019-07-13 11:34:12.416  WARN 1 --- [orker-epoll-4-1] c.l.a.c.l.LoggingClient                  : Response: {startTime=2019-07-13T11:34:12.416Z(1563017652416394), length=0B, duration=0ns, cause=com.linecorp.armeria.client.ResponseTimeoutException, headers=[]}
zipkin                      | 
zipkin                      | com.linecorp.armeria.client.ResponseTimeoutException: null
zipkin                      | 
zipkin                      | 2019-07-13 11:34:13.197  WARN 1 --- [orker-epoll-4-8] c.l.a.c.l.LoggingClient                  : Response: {startTime=2019-07-13T11:34:13.197Z(1563017653197137), length=0B, duration=0ns, cause=com.linecorp.armeria.client.ResponseTimeoutException, headers=[]}
zipkin                      | 
zipkin                      | com.linecorp.armeria.client.ResponseTimeoutException: null
zipkin                      | 
zipkin                      | 2019-07-13 11:34:15.379  WARN 1 --- [orker-epoll-4-2] c.l.a.c.l.LoggingClient                  : Response: {startTime=2019-07-13T11:34:15.378Z(1563017655378782), length=0B, duration=0ns, cause=com.linecorp.armeria.client.ResponseTimeoutException, headers=[]}
zipkin                      | 
zipkin                      | com.linecorp.armeria.client.ResponseTimeoutException: null
zipkin                      | 
zipkin                      | 2019-07-13 11:34:17.530  WARN 1 --- [orker-epoll-4-3] c.l.a.c.l.LoggingClient                  : Response: {startTime=2019-07-13T11:34:17.530Z(1563017657530053), length=0B, duration=0ns, cause=com.linecorp.armeria.client.ResponseTimeoutException, headers=[]}
zipkin                      | 
zipkin                      | com.linecorp.armeria.client.ResponseTimeoutException: null
zipkin                      | 
zipkin                      | 2019-07-13 11:34:18.613  WARN 1 --- [orker-epoll-4-4] c.l.a.c.l.LoggingClient                  : Response: {startTime=2019-07-13T11:34:18.613Z(1563017658613401), length=0B, duration=0ns, cause=com.linecorp.armeria.client.ResponseTimeoutException, headers=[]}
zipkin                      | 
zipkin                      | com.linecorp.armeria.client.ResponseTimeoutException: null
zipkin                      | 
zipkin                      | 2019-07-13 11:34:18.815  WARN 1 --- [orker-epoll-4-5] c.l.a.c.l.LoggingClient                  : Response: {startTime=2019-07-13T11:34:18.815Z(1563017658815107), length=0B, duration=0ns, cause=com.linecorp.armeria.client.ResponseTimeoutException, headers=[]}
zipkin                      | 
zipkin                      | com.linecorp.armeria.client.ResponseTimeoutException: null
zipkin                      | 
zipkin                      | 2019-07-13 11:34:20.228  WARN 1 --- [orker-epoll-4-6] c.l.a.c.l.LoggingClient                  : Response: {startTime=2019-07-13T11:34:20.228Z(1563017660228235), length=0B, duration=0ns, cause=com.linecorp.armeria.client.ResponseTimeoutException, headers=[]}
zipkin                      | 
zipkin                      | com.linecorp.armeria.client.ResponseTimeoutException: null
zipkin                      | 
zipkin                      | 2019-07-13 11:34:20.304  WARN 1 --- [orker-epoll-4-7] c.l.a.c.l.LoggingClient                  : Response: {startTime=2019-07-13T11:34:20.304Z(1563017660304298), length=0B, duration=0ns, cause=com.linecorp.armeria.client.ResponseTimeoutException, headers=[]}
zipkin                      | 
zipkin                      | com.linecorp.armeria.client.ResponseTimeoutException: null
zipkin                      | 
```

codefromthecrypt:
so basically it is still dropping almost every span before and after I guess..

codefromthecrypt:
```bash
{
  "counter.zipkin_collector.messages.http": 46,
  "counter.zipkin_collector.spans_dropped.http": 253822,
  "gauge.zipkin_collector.message_bytes.http": 1857454,
  "counter.zipkin_collector.bytes.http": 92070197,
  "gauge.zipkin_collector.message_spans.http": 5877,
  "counter.zipkin_collector.spans.http": 258104,
}
```

codefromthecrypt:
I'm running against last release to check also..

codefromthecrypt:
to make docker not fall over, I needed to change the sender configuration to send less data at a time. This is before and after the change. I just got a clean run of before.. will do one of the after now.

```diff
diff --git a/webmvc4/pom.xml b/webmvc4/pom.xml
index af9f420..79d2a2c 100755
--- a/webmvc4/pom.xml
+++ b/webmvc4/pom.xml
@@ -18,7 +18,7 @@
 
     <spring.version>4.3.24.RELEASE</spring.version>
     <log4j.version>2.11.2</log4j.version>
-    <brave.version>5.6.6</brave.version>
+    <brave.version>5.6.7</brave.version>
   </properties>
 
   <dependencyManagement>
diff --git a/webmvc4/src/main/java/brave/webmvc/TracingConfiguration.java b/webmvc4/src/main/java/brave/webmvc/TracingConfiguration.java
index 095c043..38722b2 100644
--- a/webmvc4/src/main/java/brave/webmvc/TracingConfiguration.java
+++ b/webmvc4/src/main/java/brave/webmvc/TracingConfiguration.java
@@ -20,6 +20,7 @@ import org.springframework.context.annotation.Import;
 import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter;
 import zipkin2.Span;
+import zipkin2.codec.Encoding;
 import zipkin2.reporter.AsyncReporter;
 import zipkin2.reporter.Sender;
 import zipkin2.reporter.okhttp3.OkHttpSender;
@@ -38,7 +39,10 @@ public class TracingConfiguration extends WebMvcConfigurerAdapter {
 
   /** Configuration for how to send spans to Zipkin */
   @Bean Sender sender() {
-    return OkHttpSender.create("http://127.0.0.1:9411/api/v2/spans");
+    return OkHttpSender.newBuilder()
+        .endpoint("http://127.0.0.1:9411/api/v2/spans")
+        .messageMaxBytes(4096)
+        .encoding(Encoding.PROTO3).build();
   }
 
   /** Configuration for how to buffer spans into messages for Zipkin */

```

anuraaga:
Tweaked `toString` to get this now
```
{"status":"UP","zipkin":{"status":"UP","details":{"ElasticsearchStorage{clientCustomizer=[TimeoutCustomizer{timeout=10000ms}, LoggingCustomizer{httpLogging=BODY}], clientFactoryCustomizer=[TimeoutCustomizer{timeout=10000ms}], hostsSupplier=[http://localhost:9200], pipeline=null, flushOnWrites=false, maxRequests=64, strictTraceId=true, searchEnabled=true, autocompleteKeys=[], autocompleteTtl=3600000, autocompleteCardinality=20000, indexShards=5, indexReplicas=1, indexNameFormatter=IndexNameFormatter{index=zipkin, dateSeparator=-, dateFormat=yyyy-MM-dd}, namesLookback=86400000}":{"status":"UP"}}}}
```

codefromthecrypt:
cool... I did a run with the prior commit, but metrics weren't reported at all for spans. Some unrelated glitch I'm sure..

```bash
curl -s localhost:9411/metrics|jq .
{
  "counter.zipkin_collector.messages.http": 17424,
  "counter.zipkin_collector.spans_dropped.http": 0,
  "gauge.zipkin_collector.message_bytes.http": 2736,
  "counter.zipkin_collector.bytes.http": 69106481,
  "gauge.zipkin_collector.message_spans.http": 0,
  "counter.zipkin_collector.spans.http": 0,
}
```

Anyway, I'm running off `ENV VERSION c1d3eac893-1` and will try again

codefromthecrypt:
metrics aren't coming through, but it is only the span count.. so weird.

codefromthecrypt:
I will build local as I can't see easily why the metrics are zero for spans. debugging the client shows the http response was 202

codefromthecrypt:
the problem in span stats is kindof making it hard for me to tell what's going on. it is possible something subtle during the bump to latest micrometer, but it is so weird. Ex I can manually send traffic, then a few times it will increase the number but after a while it doesn't. meanwhile the other stats seem fine.

codefromthecrypt:
Screenshot 2019-07-13 at 9 41 10 PM

ok something's odd

codefromthecrypt:
the above thing has to do with proto encoding. I'm switching tests back to json, just smaller messages, until it is figured out.

codefromthecrypt:
ok sorry about all the chatter. a successful test scenario where any load occurs requires storage throttling as we know ES falls over especially on localhost.

Ex if directly like this..
```bash
$ STORAGE_THROTTLE_ENABLED=true STORAGE_TYPE=elasticsearch java -jar ./zipkin-server/target/zipkin-server-*exec.jar
```

Then, something unrelated to this is busted in proto3 encoding. To avoid this, don't use it :P Ex. use the following sender setup in brave-webmvc-example. It is important to override the max bytes to something less than default 5MiB, if using laptop etc, but I don't know what that value is.

```java
  /** Configuration for how to send spans to Zipkin */
  @Bean Sender sender() {
    return OkHttpSender.newBuilder()
        .endpoint("http://127.0.0.1:9411/api/v2/spans")
        .messageMaxBytes(4096)
        .build();
  }
```

Anyway, with storage throttling enabled and the above running directly off this branch I got a clean bench against brave-webmvc-example. zero spans dropped. I haven't tried to setup this in docker, yet but I don't have it in me to do that tonight :P

anuraaga:
Well that's enough for me to move this off of draft :)

zipkin-server/src/main/java/zipkin2/server/internal/MetricsHealthController.java
```diff
@@ -85,7 +85,7 @@ public ObjectNode fetchMetricsFromMicrometer() {
   // Delegates the health endpoint from the Actuator to the root context path and can be deprecated
   // in future in favour of Actuator endpoints
   @Get("/health")
-  public HttpResponse getHealth() throws JsonProcessingException {
+  public AggregatedHttpResponse getHealth() throws JsonProcessingException {
```
codefromthecrypt, line 88: side q: will any of the other read methods end up blocking the event loop? ex getTraces etc?

zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/ZipkinElasticsearchStorageAutoConfiguration.java
```diff
@@ -59,126 +48,116 @@
...
-
-      @Override protected <C> Callable<C> wrap(Callable<C> task) {
-        return RequestContext.current().makeContextAware(traceCtx.wrap(task));
-      }
-
-      @Override protected Runnable wrap(Runnable task) {
-        return RequestContext.current().makeContextAware(traceCtx.wrap(task));
-      }
+  @Bean @Qualifier(QUALIFIER) @ConditionalOnSelfTracing Consumer<HttpClientBuilder>
+  elasticsearchTracing(Optional<Tracing> tracing) {
```
codefromthecrypt, line 156: trying self-tracing seems to have a glitch as I'm only seeing spans for static assets.. looking into it now

zipkin-storage/elasticsearch/README.md
```diff
@@ -1,22 +1,21 @@
...
-scheme must be http, not https.
+lookups to get the list of IP addresses, though you can alternatively supply 
+a list of http base urls. In either case, all of the resolved IP addresses
+from all provided hosts will be iterated over round-robin, with requests made
+only to healthy addresses.
 
 Here are some examples:
 
-* http://1.1.1.1:9200,http://2.2.2.2:9200
+* http://1.1.1.1:9200,http://2.2.2.2:19200
```
codefromthecrypt, line 18: great!

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -56,40 +77,41 @@
...
   abstract Builder toBuilder();
 
   @AutoValue.Builder
   public abstract static class Builder extends StorageComponent.Builder {
-    abstract Builder client(OkHttpClient client);
+    /**
+     * Customizes the {@link HttpClientBuilder} used when connecting to ElasticSearch. Mostly for
+     * testing.
+     */
+    public abstract Builder clientCustomizer(Consumer<HttpClientBuilder> clientCustomizer);
```
codefromthecrypt, line 107: if only for testing, we should make it package private like the above. the more we expose the more we break on next refactor :P

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -341,41 +370,141 @@ IndexTemplates ensureIndexTemplates() {
...
+    final String clientUrl;
+    if (endpointGroup != null) {
+      HttpHealthCheckedEndpointGroup healthChecked = new HttpHealthCheckedEndpointGroupBuilder(
+        endpointGroup, "/_cluster/health")
+        .clientFactory(clientFactory())
+        .build();
+      EndpointGroupRegistry.register(
+        "elasticsearch", healthChecked, EndpointSelectionStrategy.ROUND_ROBIN);
+      try {
+        healthChecked.awaitInitialEndpoints(30, TimeUnit.SECONDS);
```
codefromthecrypt, line 459: 30 seconds is probably excessive btw. usually we allow services to come up later (we don't fail fast rather defer folks to using health check). For example it is a bug for the collector to break startup.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/IndexNameFormatter.java
```diff
@@ -93,17 +93,21 @@ public static Builder newBuilder() {
     abstract char dateSeparator();
 
     public final IndexNameFormatter build() {
+      char separator = dateSeparator();
+      String format = separator == 0 ? "yyyyMMdd" : "yyyy-MM-dd".replace('-', separator);
```
codefromthecrypt, line 97: much better!

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -341,41 +370,141 @@ IndexTemplates ensureIndexTemplates() {
...
+          throw new IllegalArgumentException("Invalid host: " + host, e);
+        }
+      })
+      .collect(Collectors.toList());
+
+    final EndpointGroup endpointGroup;
+    if (urls.size() == 1) {
+      URL url = urls.get(0);
+      if (isIpAddress(url.getHost())) {
+        endpointGroup = null;
```
codefromthecrypt, line 417: logic like this needs a comment. my guess is that a hostname could be potentially more than one IPs and that's why. Just a few word trailing comment in the main branches here would be handy.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/RawContentLoggingClient.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+public class RawContentLoggingClient extends SimpleDecoratingClient<HttpRequest, HttpResponse> {
+
+  public static Function<Client<HttpRequest, HttpResponse>, Client<HttpRequest, HttpResponse>>
+  newDecorator() {
+    return RawContentLoggingClient::new;
+  }
+
+  /**
+   * Creates a new instance that decorates the specified {@link Client}.
```
codefromthecrypt, line 37: I'd love to see these few lines swapped with a comment below mentioning the importance of pooled objects stuff.

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/ElasticsearchSpanConsumerTest.java
```diff
@@ -41,64 +49,101 @@
...
+    AggregatedHttpResponse.of(ResponseHeaders.of(HttpStatus.OK), HttpData.EMPTY_DATA);
+
+  @Rule public ServerRule server = new ServerRule() {
+    @Override protected void configure(ServerBuilder sb) throws Exception {
+      sb.serviceUnder("/", (ctx, req) -> {
+        CompletableFuture<HttpResponse> responseFuture = new CompletableFuture<>();
+        req.aggregate().thenAccept(agg -> {
+          CAPTURED_REQUESTS.add(agg);
+          AggregatedHttpResponse response = MOCK_RESPONSES.remove();
+          if (response.headers().contains("delay-one-second")) {
```
codefromthecrypt, line 64: neat

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
```diff
@@ -133,6 +133,7 @@ public static ReadBuffer wrap(byte[] bytes, int pos, int limit) {
     }
 
     @Override public int read(byte[] dst, int offset, int length) {
+      if (available() == 0) return -1;
```
codefromthecrypt, line 136: nice

codefromthecrypt (approved):
This was a great amount of work. thank you! While not critical for this PR, I want to look into the self-tracing and see why it isn't working. Otherwise when you feel ready to merge, go ahead!

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
codefromthecrypt: I suspect one or two of these are exposed for zipkin-aws, if so a comment under the javadoc saying literally what is using it will help us from accidentally breaking.

codefromthecrypt:
self-tracing seems it was busted prior to this. we can follow-up with the small things in future pull requests. Great job @anuraaga!

codefromthecrypt:
self-tracing follow-up https://github.com/openzipkin/zipkin/pull/2663

zipkin-server/src/main/java/zipkin2/server/internal/MetricsHealthController.java
anuraaga: Nope they all return `AggregatedHttpResponse` so don't run on the event loop. Though I'm definitely interested in eventually migrating them to use `enqueue` instead of `execute` to be fully async :)

codefromthecrypt:
thx for the loopback

PR #2654, opened 2019-07-09
Add unit tests for zipkin2.Endpoint

JohnLBergqvist:
I've analysed your codebase and noticed that `zipkin2.Endpoint` is not fully tested.
I've written some tests for the methods in this class with the help of [Diffblue Cover](https://www.diffblue.com/opensource).

Hopefully, these tests will help you detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other classes that you consider important in a subsequent PR.

### DIFF
--- a/zipkin/pom.xml
+++ b/zipkin/pom.xml
@@ -73,6 +73,12 @@
       <artifactId>kryo</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.diffblue</groupId>
+      <artifactId>deeptestutils</artifactId>
+      <version>1.9.0</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
appreciate the offer, but I think the style of these generated tests is not very good. They feel cluttered and distracting. Also, this adds a dependency for routine state checking which I don't think we require. 

Considering the endpoint class is before this change at 95% methods and 90% lines, it is probably better to look at the parts not covered and create tests based on the semantics of the type. There isn't much left to do on this one.

Screenshot 2019-07-09 at 9 57 51 PM

If you are looking to add pull requests based on generated code, probably a better start next time would be asking us. We're usually on gitter. Thanks anyway, though!

https://gitter.im/openzipkin/zipkin

codefromthecrypt:
ps I do think what you are doing is neat btw.. it can be difficult to identify corner cases in legacy code and I think the utility you are working on has promise, even if not a great fit for us.

JohnLBergqvist:
@adriancole 
Many thanks for your feedback. We're constantly improving our product, so this kind of feedback is very valuable to us.

PR #2656, opened 2019-07-10
Redesign trace search page

### LINKED ISSUE #2635: Use friendly timestamp like old UI
**Feature:**
Labels like `less than a minute ago` were good to catch when something was recent over some trace that was old (mostly interesting when you have recent traces you wanna see). Also having labels for `x minutes ago` feels more natural to read.

**Rational**
It was in old UI and I think it was good.
image (2)

**Prior Art**

image (1)

Ping @kirkov

### LINKED ISSUE #2685: Remove trace preview feature
Many may recall we had a trace preview feature in the search results screen. This I believe was an idea borrowed from jaeger. As discussed with @tacigar this is probably not a worthwhile feature.

For example, it appears to solve the trace comparison use case but it doesn't. For example, it is unlikely you will have two traces near each-other in the results and even if you did, the scrolling up and down wildly through expanded traces thing isn't that effective.

Moreover, there's recent work in the UI to share data between the trace list and the trace view page, which obviates network calls. In other words, as long as navigation such as "back" works properly, there's no reason to overload the screen, especially for performance reasons.

This should buy back code and complexity for better purposes. For example, right now no screen is ideal for comparing multiple traces. However, we could build a compare screen and let people select into it.

cc @openzipkin/ui @narayaruna

tacigar:
Redesign trace search page by using Material-ui components.

* Remove trace preview feature (please see #2685)
* Support the feature of filtering by service name
* Improve whole design

## OLD DESIGN:

old-design

## NEW DESIGN:

new-design-1

new-design-2

fixes #2685 
fixes #2635


tacigar:
Added service filter.

ServiceFilter

tacigar:
Added result table.

Table

codefromthecrypt:
Is this still WIP?

tacigar:
Yes, this is still WIP.
I noticed that I should do #2677 first...

tacigar:
I'll add unit tests.

tacigar:
Current state:

search-page

codefromthecrypt:
cool client side service filter!

tacigar:
@adriancole @zeagord 
Please review 🙇

zipkin-lens/src/components/DiscoverPage/TracesTab/TracesTab.jsx
```diff
@@ -0,0 +1,97 @@
...
+  );
+};
+
+TracesTab.propTypes = propTypes;
+
+const mapStateToProps = state => ({
+  traceSummaries: state.traces.traceSummaries,
+});
+
+// React-Redux already provides two hooks API, useSelector and useDispatch.
```
zeagord, line 91: Makes sense

zeagord (approved):
Looks good.

codefromthecrypt:
@tacigar are the screen shots in the description latest?

tacigar:
@adriancole yes

codefromthecrypt:
can you show trace detail screen? I wonder if root span name is there also

codefromthecrypt:
epic thanks. cant wait to play

tacigar:
> can you show trace detail screen? I wonder if root span name is there also

Sorry, not yet 😭

codefromthecrypt:
no problem. as @zeagord advises best to do small changes when we can. This is a good foundation for small change later. excellent work

PR #2657, opened 2019-07-12
Fix bugs that fetching traces with old conditions

tacigar:
## Summary

After adding conditions in search bar and pushing enter key, I noticed that the conditions were not applied to parameters of the fetching.
This PR fixes this.

### Before

ServiceName condition is not added to the address bar of browser after pushing enter key.

before

### After

ServiceName condition is added.

after

## Content

* Fix the above bug.
* Remove unnecessary statements.
* Optimize a bit (use useCallback to handleKeyDown).

### DIFF
--- a/zipkin-lens/src/components/DiscoverPage/DiscoverPage.jsx
+++ b/zipkin-lens/src/components/DiscoverPage/DiscoverPage.jsx
@@ -12,7 +12,7 @@
  * the License.
  */
 import PropTypes from 'prop-types';
-import React, { useCallback } from 'react';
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
WHAT IS THIS MAGIC? HOW CAN I MOVIE WITH MY KEYBOARD?

codefromthecrypt (approved):
lovely

jcchavezs:
This is great Igarashi, any way we can cover this with a test to avoid this
to happen again?

lør. 13. jul. 2019, 10:20 skrev Adrian Cole <notifications@github.com>:

PR #2658, opened 2019-07-13
Fix an bug that can not scroll on the Trace page

tacigar:
This is a little bug.

Before

before-scroll

After

after-scroll


codefromthecrypt (approved):
thanks!

PR #2663, opened 2019-07-14
Prunes single-span traces after the fact until we can control http sampling

codefromthecrypt:
This helps ensure self-tracing isn't overrun with static assets.

Example that shows it works:
Screenshot 2019-07-14 at 9 31 08 AM

from
```bash
$ SELF_TRACING_ENABLED=true STORAGE_THROTTLE_ENABLED=true STORAGE_TYPE=elasticsearch java -jar ./zipkin-server/target/zipkin-server-*exec.jar
```

See https://github.com/line/armeria/issues/1223

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,7 +56,7 @@
     <wire.version>3.0.0-alpha01</wire.version>
     <unpack-proto.directory>${project.build.directory}/test/proto</unpack-proto.directory>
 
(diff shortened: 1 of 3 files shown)


anuraaga:
I made a different, potentially simpler alternative in #2668 in case that seems like a better approach.

codefromthecrypt:
obviated for now thanks @anuraaga

PR #2665, opened 2019-07-14
Accept many spans at once instead of in a loop in service name ITs.

anuraaga:
We've been hovering at Travis's limit for some time, but luckily found some low hanging mangos.


codefromthecrypt (approved):
thanks for the help. you are right we have for several months been battling near the limit

anuraaga:
I was hoping for some improvement in the cassandra test too, but didn't realize it doesn't batch at all!

PR #2666, opened 2019-07-14
Close Cassandra storage during integration tests.

anuraaga:
Don't expect a huge difference in test time or anything but seems like a good idea not to leak sessions if nothing else than as principle.


codefromthecrypt (approved):
agree

anuraaga:
By the way, added some null checks to the business logic too, it seems like any method that is clearing a field that's nullable should do the null check all the time.

codefromthecrypt:
good idea

PR #2667, opened 2019-07-14
Await endpoints in check, not httpClient

anuraaga:
Noticed that some non-integration tests were slow waiting and failing on the initial endpoints because they don't have real ones. It was dubious to be waiting in the method that creates an object anyways, so I moved it to `check` which seems like the right spot. Reduced the timeout to use the configured one too.

Also apply some cleanups mentioned in #2653

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
@@ -101,14 +101,14 @@ public abstract class ElasticsearchStorage extends zipkin2.storage.StorageCompon
   @AutoValue.Builder
   public abstract static class Builder extends StorageComponent.Builder {
     /**
-     * Customizes the {@link HttpClientBuilder} used when connecting to ElasticSearch. Mostly for
-     * testing.
+     * Customizes the {@link HttpClientBuilder} used when connecting to ElasticSearch. This is used
+     * by the server and tests to enable detailed logging and tweak timeouts.
      */
     public abstract Builder clientCustomizer(Consumer<HttpClientBuilder> clientCustomizer);
 
     /**
-     * Customizes the {@link ClientFactoryBuilder} used when connecting to ElasticSearch. Mostly for
-     * testing.
+     * Customizes the {@link ClientFactoryBuilder} used when connecting to ElasticSearch. This is
+     * used by the server and tests to tweak timeouts.
      */
     public abstract Builder clientFactoryCustomizer(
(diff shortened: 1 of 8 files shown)


zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -101,14 +101,14 @@ public static Builder newBuilder() {
   @AutoValue.Builder
   public abstract static class Builder extends StorageComponent.Builder {
     /**
-     * Customizes the {@link HttpClientBuilder} used when connecting to ElasticSearch. Mostly for
-     * testing.
+     * Customizes the {@link HttpClientBuilder} used when connecting to ElasticSearch. This is used
+     * by the server and tests to enable detailed logging and tweak timeouts.
```
codefromthecrypt, line 105: this is good but doesnt explain why it needs to be public. usually for tests we use internal package accessors which allows us to remove or change the method.  we dont expose public methods unless there is non test code using it in other words

codefromthecrypt: ex below you could add exactly why like // specifically this is used for self tracing

or whatever it is used for.. reason is, that sometimes we have added methidd for gcp or aws and forgot and then broke them

codefromthecrypt (approved):
good work

codefromthecrypt:
ps my note on docs is a nit as they are already better than probably everything. just didn't merge as it is red and haven't looked into why

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
anuraaga: The main reason is `the server` - it calls these from `zipkin2.server.internal.elasticsearch`. Actually don't see usage of these in `zipkin-aws`.

I'll leave the comments like this for now since not sure of a better alternative.

anuraaga (commented):
Yeah found the lifecycle during autocomplete integration tests was tripping up so made it more explicit. I'll take a scalpel to these integration tests soon to make them clearer / faster.

codefromthecrypt:
heh we will have to port zipkin-aws soon (probably today would be good) as it is not using this because it hasn't migrated off okhttp yet https://github.com/openzipkin/zipkin-aws/blob/master/autoconfigure/storage-elasticsearch-aws/src/main/java/zipkin/autoconfigure/storage/elasticsearch/aws/ZipkinElasticsearchAwsStorageAutoConfiguration.java#L53

PR #2668, opened 2019-07-14
Only enable tracing for query and collector API.

### LINKED ISSUE #2660: Switch self-tracing to rate limited so we can enable for storage
Right now, we don't self-trace storage because it can amplify writes. However, this won't be the case if we sample at a rate, say 1 per second. Let's use brave's new sampler because particularly new storage impls are very hard to troubleshoot without tracing.

cc @anuraaga

anuraaga:
- Applies self-tracing only to API endpoints (both GET and POST)
- Allows `zipkin.self-tracing.message-timeout` to accept any type of duration (e.g., `100ms`). The default behavior is preserved, where `1` will be 1 second
- Allows configuring `zipkin.self-tracing.traces-per-second` to enable QPS limited sampling instead of percentage sampling. For backwards compatibility, still respect the old flag

Fixes #2660


anuraaga:
Well maybe not since the test that I thought would fail for POST to v2 spans is succeeding, probably due to an Armeria bug will check that too.

anuraaga:
Nah it was a bug in the test. I also went ahead and enabled tracing of the collector too, opting to trust the integration test instead of the `method is GET` currently unused logic as the desired.

codefromthecrypt:
thanks  this should work but there is a small matter which was the sampler also disabled client originated traces (hence coupling with #2660 could be a good idea ex setting default to 1 trace per second regardless of endpoint)

one way to tell if there is a problem or not would be to improve our tests a bit to hit post endpoint and static asset and check to see if after the reporting interval how many traces exist.

ps we can also add a test scoped config to shorten the interval

anuraaga:
Thanks - added the rate limiting sampling support (made it backwards compatible but let me know if we don't need to) and some tests for ensuring static resources aren't traced.

I've introduced `awaitility` since it's an easy way to allow tests to be as fast as possible without being flaky. Using `Thread.sleep(1)` with a `1ms` timeout will often run into flaky behavior, while using a larger sleep is often a waste of time. Tests went from taking about 10s to 2s (initial test takes 1s, additional tests just 100ms).

zipkin-server/src/main/java/zipkin2/server/internal/brave/SelfTracingProperties.java
```diff
@@ -0,0 +1,62 @@
...
+   * the value of {@link #getTracesPerSecond()} will be used for sampling traces. Defaults to
+   * {@code 1.0}, sampling all traces.
+   */
+  private float sampleRate = 1.0f;
+  /**
+   * The number of traces per second to retain. If 0, an unlimited number of traces will be
+   * retained. This value has no effect if {@link #getSampleRate()} is set to something other than
+   * {@code 1.0}. Defaults to 1 trace per second.
+   */
+  private int tracesPerSecond = 1;
```
codefromthecrypt, line 24: fyi @marcingrzejszczak @devinsba @shakuzen @llinder @zeagord @jeqo we probably want a similar thing in sleuth to default to a certain traces per second instead of 100%

zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
```diff
@@ -85,34 +82,27 @@
 
   /** Controls aspects of tracing such as the name that shows up in the UI */
   @Bean Tracing tracing(
-      @Lazy Reporter<Span> reporter, @Value("${zipkin.self-tracing.sample-rate:1.0}") float rate) {
+      @Lazy Reporter<Span> reporter, SelfTracingProperties config) {
+    final Sampler sampler;
+    if (config.getSampleRate() != 1.0) {
```
codefromthecrypt, line 87: good idea. unbounded 100% sampling will always be wrong when we are tracing POST requests

zipkin-server/src/test/kotlin/zipkin2/server/internal/brave/ITZipkinSelfTracing.kt
```diff
@@ -28,37 +30,80 @@ import zipkin2.storage.InMemoryStorage
...
+    assertThat(traces[0]).anyMatch {
+      it.name() == "post" && it.tags()["http.path"] == "/api/v2/spans"
+    }
   }
 
   private fun assertServerTraced() {
-    Thread.sleep(1500) // wait for reporting interval
+    // wait for spans
+    await().untilAsserted {
+      assertThat(inMemoryStorage.acceptedSpanCount()).isGreaterThanOrEqualTo(1)
```
codefromthecrypt, line 104: specifically we should check for recursion.. like await until quiet (to ensure it will ever be quiet), but we've also never done that. So this is good in other words :)

zipkin-server/src/test/kotlin/zipkin2/server/internal/brave/ITZipkinSelfTracing.kt
```diff
@@ -28,37 +30,80 @@ import zipkin2.storage.InMemoryStorage
...
+    }
 
-    assertThat(Http.getAsString(server, "/api/v2/services"))
-      .isEqualTo("[\"zipkin-server\"]")
+    assertThat(inMemoryStorage.serviceNames.execute()).containsExactly("zipkin-server")
```
codefromthecrypt, line 107: better!

codefromthecrypt (approved):
good stuff. will try locally then merge

codefromthecrypt:
while master seems ok, This is arising from a POST request (from the sample app to the collector)

```
2019-07-16 11:30:09.033  WARN 21674 --- [-worker-nio-2-2] z.s.i.BodyIsExceptionMessage             : Unexpected error handling request.

com.linecorp.armeria.client.endpoint.EndpointGroupException: HealthCheckedEndpointGroup(all:], healthy:]) is empty
	at com.linecorp.armeria.client.endpoint.RoundRobinStrategy$RoundRobinSelector.select(RoundRobinStrategy.java:65) ~[armeria-0.88.0.jar!/:?]
	at com.linecorp.armeria.client.endpoint.EndpointGroupRegistry.selectNode(EndpointGroupRegistry.java:120) ~[armeria-0.88.0.jar!/:?]
	at com.linecorp.armeria.client.Endpoint.resolve(Endpoint.java:215) ~[armeria-0.88.0.jar!/:?]
	at com.linecorp.armeria.client.HttpClientDelegate.execute(HttpClientDelegate.java:61) ~[armeria-0.88.0.jar!/:?]
	at com.linecorp.armeria.client.HttpClientDelegate.execute(HttpClientDelegate.java:42) ~[armeria-0.88.0.jar!/:?]
	at com.linecorp.armeria.client.encoding.HttpDecodingClient.execute(HttpDecodingClient.java:92) ~[armeria-0.88.0.jar!/:?]
	at com.linecorp.armeria.client.encoding.HttpDecodingClient.execute(HttpDecodingClient.java:39) ~[armeria-0.88.0.jar!/:?]
	at com.linecorp.armeria.internal.ClientUtil.executeWithFallback(ClientUtil.java:42) ~[armeria-0.88.0.jar!/:?]
	at com.linecorp.armeria.client.UserClient.execute(UserClient.java:159) ~[armeria-0.88.0.jar!/:?]
	at com.linecorp.armeria.client.DefaultHttpClient.execute(DefaultHttpClient.java:61) ~[armeria-0.88.0.jar!/:?]
	at com.linecorp.armeria.client.DefaultHttpClient.execute(DefaultHttpClient.java:78) ~[armeria-0.88.0.jar!/:?]
	at com.linecorp.armeria.client.DefaultHttpClient.execute(DefaultHttpClient.java:74) ~[armeria-0.88.0.jar!/:?]
	at zipkin2.elasticsearch.internal.client.HttpCall.sendRequest(HttpCall.java:145) ~[zipkin-storage-elasticsearch-2.15.1-SNAPSHOT.jar!/:?]
	at zipkin2.elasticsearch.internal.client.HttpCall.doExecute(HttpCall.java:101) ~[zipkin-storage-elasticsearch-2.15.1-SNAPSHOT.jar!/:?]
	at zipkin2.Call$Base.execute(Call.java:379) ~[zipkin-2.15.1-SNAPSHOT.jar!/:?]
	at zipkin2.elasticsearch.VersionSpecificTemplates.getVersion(VersionSpecificTemplates.java:223) ~[zipkin-storage-elasticsearch-2.15.1-SNAPSHOT.jar!/:?]
	at zipkin2.elasticsearch.VersionSpecificTemplates.get(VersionSpecificTemplates.java:189) ~[zipkin-storage-elasticsearch-2.15.1-SNAPSHOT.jar!/:?]
	at zipkin2.elasticsearch.ElasticsearchStorage.ensureIndexTemplates(ElasticsearchStorage.java:381) ~[zipkin-storage-elasticsearch-2.15.1-SNAPSHOT.jar!/:?]
	at zipkin2.elasticsearch.AutoValue_ElasticsearchStorage.ensureIndexTemplates(AutoValue_ElasticsearchStorage.java:38) ~[zipkin-storage-elasticsearch-2.15.1-SNAPSHOT.jar!/:?]
```

codefromthecrypt:
I mean master without self-tracing

zipkin-server/src/main/java/zipkin2/server/internal/brave/SelfTracingProperties.java
marcingrzejszczak: Can you create an issue for that in Sleuth?

shakuzen: Sleuth defaults to 10% sampling. Do you mean changing that default sampling from 10% to 1 trace per second?

codefromthecrypt:
> Sleuth defaults to 10% sampling. Do you mean changing that default
> sampling from 10% to 1 trace per second?
>

sleuth defaults aren't really important to us. yes, this is 1 trace per
second for self-tracing zipkin itself. percentage is a bad choice because
storage requests can amplify. we need a rate limit approach for
self-tracing when also tracing writes.

PR #2669, opened 2019-07-15
Fix zipkin-lens build on Windows by separating path out of regex.

anuraaga:
Found this while trying to run `mvn package` on Windows. The regex unfortunately only matches unix paths when using `/`.

Also adds a missing license header I found while building and uses jest-standard-reporter so jest output goes to STDOUT instead of STDERR.

### DIFF
--- a/zipkin-lens/babel.config.js
+++ b/zipkin-lens/babel.config.js
@@ -1,3 +1,16 @@
+/*
+ * Copyright 2015-2019 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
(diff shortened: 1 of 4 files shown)


anuraaga:
@tacigar I'm wondering why the `package-lock.json` has such a huge diff - do you use a different command than `npm install` during development? If so, we should reflect that to `pom.xml`

anuraaga:
The missing license header is also causing jitpack to fail, so let me go ahead and merge this, let's follow up if needed (mostly about the package-lock.json)

codefromthecrypt:
thanks. we need appveyor I suspect

PR #2670, opened 2019-07-15
Add git commit to package to show in actuator.

### LINKED ISSUE #2661: Add git hash to /actuator/info
Right now, we have the version in /actuator/info, built from property substitution in maven iirc. We should see about git hash because especially testing jitpack from within docker it is annoying as heck to try to heuristically tell what you are actually running.

cc @anuraaga @abesto

anuraaga:
Fixes #2661

### DIFF
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -459,9 +459,6 @@
             </goals>
           </execution>
         </executions>
-        <configuration>
-          <generateGitPropertiesFile>true</generateGitPropertiesFile>
-        </configuration>
       </plugin>
 
       <plugin>
@@ -471,7 +468,6 @@
         <executions>
           <execution>
             <goals>
-              <goal>build-info</goal>
               <goal>repackage</goal>
             </goals>
           </execution>
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -448,12 +448,30 @@
       </plugin>
 
       <plugin>
+        <groupId>pl.project13.maven</groupId>
+        <artifactId>git-commit-id-plugin</artifactId>
+        <version>${git-commit-id.version}</version>
+        <executions>
+          <execution>
+            <id>extract-git-info</id>
+            <goals>
+              <goal>revision</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <generateGitPropertiesFile>true</generateGitPropertiesFile>
+        </configuration>
+      </plugin>
+
+      <plugin>
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-maven-plugin</artifactId>
         <version>${spring-boot.version}</version>
         <executions>
           <execution>
             <goals>
+              <goal>build-info</goal>
               <goal>repackage</goal>
             </goals>
           </execution>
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
taking a look. thanks!

codefromthecrypt:
I'm glad we have this, but it is a bit verbose.

Ex. before:
```bash
curl -s localhost:9411/actuator/info|jq .
{
  "zipkin": {
    "version": "2.15.0"
  }
}
```

After:
```bash
$ curl -s localhost:9411/actuator/info|jq .
{
  "zipkin": {
    "version": "2.15.1-SNAPSHOT"
  },
  "git": {
    "branch": "anuraaga-git-commit-id",
    "commit": {
      "id": "0479fd4",
      "time": {
        "nano": 0,
        "epochSecond": 1563234745
      }
    }
  },
  "build": {
    "artifact": "zipkin-server",
    "name": "Zipkin Server",
    "time": {
      "nano": 286000000,
      "epochSecond": 1563234902
    },
    "version": "2.15.1-SNAPSHOT",
    "group": "io.zipkin"
  }
}
```

Maybe we can prune it down to just adding the commit? Like:
```bash
curl -s localhost:9411/actuator/info|jq .
{
  "zipkin": {
    "version": "2.15.0",
    "commit": "0479fd4"
  }
}
```

I think this is enough to solve the use case (knowing on snapshot if it is stale, and also identifying a custom build in general for support).

Maybe we can do this without a dependency instead with property replacement such as exists in `zipkin-server-shared.yml`

```yaml
info:
  zipkin:
    version: "@project.version@"
```

zipkin-server/pom.xml
```diff
@@ -447,13 +447,31 @@
...
+        <executions>
+          <execution>
+            <id>extract-git-info</id>
+            <goals>
+              <goal>revision</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <generateGitPropertiesFile>true</generateGitPropertiesFile>
```
codefromthecrypt: ah I see. it I suppose wasn't a runtime dep responsible for the verbosity, but rather spring boot detecting this file. As format of things like this have drifted on us, probably another good reason to opt out. I do think this is the right plugin, though.

codefromthecrypt:
so if something I found here works..

```
    "git.commit.id" : "6f592254e2e08d99a8145f1295d4ba3042310848",
    "git.commit.id.abbrev" : "6f59225",
```

assuming the properties are activated at build time, then ideally we could add properties like this:

```yaml
info:
  zipkin:
    version: "@project.version@"
    commit: "@git.commit.id.abbrev@"
```

taking care to add `<injectAllReactorProjects>true</injectAllReactorProjects>`

per https://github.com/git-commit-id/maven-git-commit-id-plugin/blob/master/docs/faq.md#generated-properties-are-not-usable-inside-the-pom--properties-dont-get-exposed-by-the-plugin

codefromthecrypt:
ok it took less settings than I thought.

```bash
$ git rev-parse HEAD
f43d9d148ff91461b533d3e5c56a76fafef65754

$ curl -s localhost:9411/actuator/info|jq .
{
  "zipkin": {
    "version": "2.15.1-SNAPSHOT",
    "commit": "f43d9d1"
  }
}
```

anuraaga:
Thanks

PR #2671, opened 2019-07-15
Fix ReadBuffer when wrapping an array at an offset.

### LINKED ISSUE #2662: investigate proto3 reporting concern
using brave-webmvc-example and the following config, manually hit or `wrk` generated load of requests go through fine.

```java
  /** Configuration for how to send spans to Zipkin */
  @Bean Sender sender() {
    return OkHttpSender.newBuilder()
        .endpoint("http://127.0.0.1:9411/api/v2/spans")
        .messageMaxBytes(4096)
        .build();
  }
```

If you switch to PROTO3 encoding, and manually hit the example, the server here stores the spans. However, if you generate load with `wrk`, requests get to here and are empty!

Screenshot 2019-07-13 at 9 41 10 PM

If the payload was empty it would have kicked out before here. I don't know what's going on, but there is something that works with 2 or 3 spans in proto format, but not larger. This needs to be followed up on regardless of if the problem ends up something here or in the other repo (we can always move the issue if the latter)

cc @anuraaga

anuraaga:
Hmm - this is really broken.

The issue only happens with gzipped requests, where unzipping moves the content from direct memory to pooled heap buffers. Presumably its triggered by larger requests since the sender only gzips past some threshold though I didn't verify that behavior.

The problem came by mixing limit and length in the code. When wrapping in Java, it's more conventional to use length, so I made the APIs consistently use length.

Fixes #2662

### DIFF
--- a/zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
+++ b/zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
@@ -27,8 +27,7 @@ public abstract class ReadBuffer extends InputStream {
   public static ReadBuffer wrapUnsafe(ByteBuffer buffer) {
     if (buffer.hasArray()) {
       int offset = buffer.arrayOffset() + buffer.position();
-      int limit = offset + buffer.remaining();
-      return wrap(buffer.array(), offset, limit);
+      return wrap(buffer.array(), offset, buffer.remaining());
     }
     return buffer.order() == ByteOrder.BIG_ENDIAN
       ? new BigEndianByteBuffer(buffer)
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
I wonder if related to #2662

codefromthecrypt:
oh heh you said that :P

zipkin/src/main/java/zipkin2/internal/ReadBuffer.java
```diff
@@ -27,8 +27,7 @@
   public static ReadBuffer wrapUnsafe(ByteBuffer buffer) {
     if (buffer.hasArray()) {
       int offset = buffer.arrayOffset() + buffer.position();
-      int limit = offset + buffer.remaining();
-      return wrap(buffer.array(), offset, limit);
+      return wrap(buffer.array(), offset, buffer.remaining());
```
codefromthecrypt, line 30: eek.. well it did say unsafe :P

codefromthecrypt (approved):

PR #2675, opened 2019-07-16
Makes toString of components sane and fixes self-tracing initialization

codefromthecrypt:
This fixes toString used in health check and some wonky wiring that
didn't work.

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
@@ -94,7 +94,7 @@ public class TracingConfiguration {
    */
   static final class LocalSender extends Sender {
     final BeanFactory factory;
-    StorageComponent delegate;
+    volatile StorageComponent delegate; // volatile to prevent stale reads
 
     LocalSender(BeanFactory factory) {
       this.factory = factory;
@@ -138,6 +138,7 @@ public class TracingConfiguration {
     StorageComponent delegate() {
       StorageComponent result = delegate;
       if (result != null) return delegate;
+      // synchronization is not needed as redundant calls have no ill effects
       result = factory.getBean(StorageComponent.class);
       if (result instanceof TracingStorageComponent) {
         result = ((TracingStorageComponent) result).delegate;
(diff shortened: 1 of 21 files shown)


codefromthecrypt:
still looking at some nuance when self-tracing and throttling is going on

codefromthecrypt:
The kindof strange looking work in the tracing component is around that we accidentally had 2 storage components instantiated due to loading nuance. Testing the toString helps keep this fixed!

Ex. now with tracing and throttling

```bash
curl -s localhost:9411/health|jq .
{
  "status": "UP",
  "zipkin": {
    "status": "UP",
    "details": {
      "Traced{Throttled{ElasticsearchStorage{hosts=[http://localhost:9200], index=zipkin}}}": {
        "status": "UP"
      },
      "AsyncReporter{StorageComponent}": {
        "status": "UP"
      }
    }
  }
}
```

zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
```diff
@@ -131,52 +118,76 @@
...
 
     @Override public void close() {
       // don't close delegate as we didn't open it!
     }
+
+    /** Lazy lookup to avoid proxying */
+    StorageComponent delegate() {
+      StorageComponent result = delegate;
+      if (result != null) return delegate;
+      result = factory.getBean(StorageComponent.class);
```
anuraaga, line 142: Should this be synchronized i.e. double check?

anuraaga (approved):

zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
codefromthecrypt: I've noticed a lot of similar code not synchronizing on `factory.getBean` and my guess is that it should be non-'side effect causing' to call this due to a stale field read. In other words it could possibly cost less to occasionally redo factory.getBean vs synchronizing.

@wilkinsona did I reverse-engineer intent right on this pattern? just using you as a spring set of eyes, not because you established this pattern :P

wilkinsona: Yeah, I think you have. FWIW, when using this pattern I would declare the field `volatile` to eliminate the possibility of a stale read of the field. There's still a potential race between multiple threads, but duplicate calls to `factory.getBean(StorageComponent.class)` should not have any ill effects.

codefromthecrypt: thanks for the feedback folks!

PR #2678, opened 2019-07-16
Raise default timeout for server to 11s…

anuraaga:
… and handle health check timeouts.

The default server timeout is 10s, and it's good for the client to be lower so the server can e.g., return a nice error message.

The entire problem isn't really solved though as we noticed that the storage is registered twice for some rean, so if both checks, which are run sequentially, hit the connect timeout, then it's still much larger than 10s, so I added a timeout handler as well for that case.


codefromthecrypt (approved):
only food for thought is if we shouldn't flip it and make the post side of it 11 seconds default. 10 is a lucky number so probably some other clients will default to that.

codefromthecrypt (approved):

PR #2679, opened 2019-07-17
Reverts usage of Kotlin in tests

### LINKED ISSUE #2655: unkotlin the tests
we kindof rushed a decision in #2570 to kotlin the tests. This was from an understanding that folks wanted to get more familiarity with kotlin as we were being asked about it in Brave for example. In hind sight, I don't think those who were interested in the idea were impacted much, for better or for worse. On the other hand @anuraaga has been stifled in attempts to refactor elasticsearch.

We can still have a kotlin module somewhere, but I think this may have done more bad than good in hindsight.

I offer labor to undo the kotlinification.. whether that's before or after @anuraaga's in flight change. up to him

cc @huydx @shakuzen

codefromthecrypt:
We ported tests from Java to Kotlin in #2570. However, this created some
friction especially in integration tests. This reverts the change and
ports tests added or changed since then.

Fixes #2655
reverts f3ef3f4a2555ea8e35f4d9579f7144f707fa0d07


zipkin-server/src/test/java/zipkin2/server/internal/ITZipkinGrpcCollector.java
```diff
@@ -0,0 +1,111 @@
...
+    assertThat(storage.getTraces())
+      .containsExactly(TestObjects.TRACE);
+  }
+
+  @Test public void report_emptyIsOk() throws IOException {
+
+    callReport(new ListOfSpans.Builder().build());
+  }
+
+  ReportResponse callReport(ListOfSpans spans) throws IOException {
```
codefromthecrypt, line 85: I'd suggest we do porting of this to something else in a different PR as this is huge

anuraaga (approved):

PR #2687, opened 2019-07-19
Cleans up some Elasticsearch configuration prior to release

codefromthecrypt:
Notably `ES_MAX_REQUESTS` is unusable in real life, so this removes
support as we've changed other things anyway. This also exposes the
`HostsSupplier` so that zipkin-aws can use it in case of static urls.

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/ZipkinElasticsearchStorageProperties.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/ZipkinElasticsearchStorageProperties.java
@@ -14,12 +14,7 @@
 package zipkin2.server.internal.elasticsearch;
 
 import java.io.Serializable;
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.util.ArrayList;
-import java.util.List;
 import java.util.logging.Logger;
-import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import zipkin2.elasticsearch.ElasticsearchStorage;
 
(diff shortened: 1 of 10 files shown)


codefromthecrypt:
ps there's a lot of history about ES_MAX_REQUESTS and why we moved to throttling instead. the short story is that the backlog in ES is unrelated to the concurrent connection count. It is the backlog that's the problem and you can overrun that with very few connections.

zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/ZipkinElasticsearchStorageProperties.java
```diff
@@ -183,23 +138,32 @@ public void setHttpLogging(HttpLoggingLevel httpLogging) {
...
-        .dateSeparator(dateSeparator.isEmpty() ? 0 : dateSeparator.charAt(0))
-        .pipeline(pipeline)
-        .maxRequests(throttleMaxConcurrency == null ? maxRequests : throttleMaxConcurrency)
-        .indexShards(indexShards)
-        .indexReplicas(indexReplicas);
+    if (index != null) builder.index(index);
+    if (dateSeparator != null) {
+      builder.dateSeparator(dateSeparator.isEmpty() ? 0 : dateSeparator.charAt(0));
+    }
+    // TODO if (timeout != null) builder.timeout(timeout);
```
anuraaga, line 155: This means having the storage receive timeout and setup the client appropriately I think - agree it would be cleaner than current where server passes in customizers to set the timeout.

anuraaga (approved):

shakuzen:
Just leaving a note that we should mention the removal of ES_MAX_REQUESTS in the release notes so users of this have more visibility.

codefromthecrypt:
agreed. I'll own this part

PR #2688, opened 2019-07-19
Migrate from moshi/okio to jackson

anuraaga:
Removes some dependencies from the AWS module, and probably a little bit faster as our benchmarks have shown previously.

I have moved the current moshi code into the benchmarks folder.


benchmarks/src/main/java/zipkin2/elasticsearch/internal/BulkRequestBenchmarks.java
```diff
@@ -55,7 +56,7 @@
     BulkCallBuilder.newIndexEntry(spanIndex, "span", CLIENT_SPAN, BulkIndexWriter.SPAN);
 
   @Benchmark public void writeRequest_singleSpan() throws IOException {
-    BulkCallBuilder.write(Okio.buffer(Okio.blackhole()), entry, true);
+    BulkCallBuilder.write(Unpooled.compositeBuffer(Integer.MAX_VALUE), entry, true);
```
codefromthecrypt, line 58: only this one?

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/BasicAuthInterceptor.java
```diff
@@ -65,24 +64,23 @@ public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Ex
...
-            if (message != null) throw new IllegalStateException(message.nextString());
+            JsonParser message = enterPath(JsonSerializers.jsonParser(stream), "message");
+            if (message != null) throw new IllegalStateException(message.getValueAsString());
           } catch (IOException e) {
             Exceptions.throwUnsafely(e);
             throw new UncheckedIOException(e);  // unreachable
           }
           throw new IllegalStateException(msg.toString());
-        } finally {
-          ReferenceCountUtil.safeRelease(content);
```
codefromthecrypt, line 85: no need to release anymore?

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/BulkCallBuilder.java
```diff
@@ -78,12 +83,19 @@ public BulkCallBuilder(ElasticsearchStorage es, float esVersion, String tag) {
...
     if (waitForRefresh) urlBuilder.addParam("refresh", "wait_for");
 
-    Buffer okioBuffer = new Buffer();
-    for (IndexEntry<?> entry : entries) {
-      write(okioBuffer, entry, shouldAddType);
-    }
+    final HttpData body;
 
-    HttpData body = HttpData.wrap(okioBuffer.readByteArray());
+    CompositeByteBuf sink = RequestContext.mapCurrent(
```
codefromthecrypt, line 95: some comment for those less familiar with composite heap buffer etc? basically what does this buy us?

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/BulkCallBuilder.java
```diff
@@ -93,32 +105,35 @@ public BulkCallBuilder(ElasticsearchStorage es, float esVersion, String tag) {
...
 
-  static void write(BufferedSink sink, IndexEntry entry, boolean shouldAddType) {
-    Buffer document = new Buffer();
-    String id = entry.writer().writeDocument(entry.input(), document);
-    writeIndexMetadata(sink, entry, id, shouldAddType);
+  static void write(CompositeByteBuf sink, IndexEntry entry,
+    boolean shouldAddType) {
+    // Fuzzily assume a general small span is 600 bytes to reduce resizing while building up the
+    // JSON. Any extra bytes will be released back after serializing all the documents.
+    ByteBuf document = sink.alloc().heapBuffer(600).writeByte('\n');
```
codefromthecrypt, line 119: good comment

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/BulkCallBuilder.java
```diff
@@ -93,32 +105,35 @@ public BulkCallBuilder(ElasticsearchStorage es, float esVersion, String tag) {
...
-      sink.writeByte('\n');
-    } catch (IOException e) {
-      throw new AssertionError(e); // No I/O writing to a Buffer.
+      String id = entry.writer().writeDocument(entry.input(), new ByteBufOutputStream(document));
+      document.writeByte('\n');
+      writeIndexMetadata(new ByteBufOutputStream(metadata), entry, id, shouldAddType);
+    } catch (Throwable t) {
+      document.release();
+      metadata.release();
+      throw t;
```
codefromthecrypt, line 128: sneaky :P

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/BulkIndexWriter.java
```diff
@@ -73,103 +79,113 @@
...
+      if (Boolean.TRUE.equals(span.shared())) writer.writeBooleanField("shared", true);
+      writer.writeEndObject();
     } catch (IOException e) {
       throw new AssertionError(e); // No I/O writing to a Buffer.
     }
-    return new Buffer()
-      .writeUtf8(span.traceId()).writeByte('-').writeUtf8(hashingSink.hash().hex())
-      .readUtf8();
+
+    ByteBuf slice = sink.buffer().slice(startIndex, sink.buffer().writerIndex() - startIndex);
```
codefromthecrypt, line 125:
```suggestion
    // get a slice representing the document we just wrote so that we can make a content hash
    ByteBuf slice = sink.buffer().slice(startIndex, sink.buffer().writerIndex() - startIndex);
```

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonReaders.java
```diff
@@ -25,85 +25,77 @@
...
-    while (reader.hasNext()) {
-      if (reader.nextName().equals(path) && reader.peek() != JsonReader.Token.NULL) {
-        return reader;
+  public static JsonParser enterPath(JsonParser parser, String path) throws IOException {
+    if (!parser.isExpectedStartObjectToken()) return null;
+    JsonToken value;
+    while ((value = parser.nextValue()) != JsonToken.END_OBJECT) {
+      if (value == null) {
+        // End of input.
+        throw new IOException("End of input while parsing object.");
```
codefromthecrypt: our old code didn't throw I think, rather returned null

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonReaders.java
```diff
@@ -25,85 +25,77 @@
...
-      if (reader.nextName().equals(name)) {
-        result.add(reader.nextString());
+  static void visitObject(JsonParser parser, String name, Set<String> result) throws IOException {
+    if (!parser.isExpectedStartObjectToken()) {
+      throw new IOException("Expecting object start, got " + parser.currentToken());
+    }
+    JsonToken value;
+    while ((value = parser.nextValue()) != JsonToken.END_OBJECT) {
+      if (value == null) {
+        throw new IOException("End of input while parsing object.");
```
codefromthecrypt: I think the IOExceptions here could be turned to null returns possibly log.fine guarded if important. Starting to throw when wasn't before could break things.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonSerializers.java
```diff
@@ -0,0 +1,256 @@
...
+        case "port":
+          port = parser.getIntValue();
+          break;
+        default:
+          // Skip
+      }
+    }
+
+    if (serviceName == null && ipv4 == null && ipv6 == null && port == 0) return null;
+    return Endpoint.newBuilder()
```
codefromthecrypt, line 186: I presume this part was a straight port of the old one.. I forget the precedence of ipv4 vs ipv6

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonSerializers.java
```diff
@@ -0,0 +1,256 @@
...
+        case "value":
+          value = parser.getValueAsString();
+          break;
+        default:
+          // Skip
+      }
+    }
+
+    if (timestamp == 0 || value == null) {
+      throw new IllegalStateException("Incomplete annotation at " + parser.currentToken());
```
codefromthecrypt: I know we have been inconsistent, but usually use IllegalArgumentException on parsing as State is something that could resolve, and this won't.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonSerializers.java
```diff
@@ -0,0 +1,256 @@
...
+    }
+    return Annotation.create(timestamp, value);
+  }
+
+  public static final ObjectParser<DependencyLink> DEPENDENCY_LINK_PARSER =
+    new ObjectParser<DependencyLink>() {
+      @Override public DependencyLink parse(JsonParser parser) throws IOException {
+        if (!parser.isExpectedStartObjectToken()) {
+          throw new IOException("Expected start of dependency link object but was "
```
codefromthecrypt: similarly unless it is actually an IOException probably better to use IAE

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
```diff
@@ -171,7 +168,7 @@ public String toString() {
         } else {
           buf = ByteBuffer.wrap(content.array());
         }
-        return bodyConverter.convert(Okio.buffer(Okio.source(ReadBuffer.wrapUnsafe(buf))));
```
codefromthecrypt, line 174: haha yeah ...

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/SearchCallFactory.java
```diff
@@ -13,33 +13,40 @@
...
 import com.linecorp.armeria.common.RequestHeaders;
-import com.squareup.moshi.JsonAdapter;
-import com.squareup.moshi.Moshi;
 import java.util.List;
 import zipkin2.internal.Nullable;
 
 public class SearchCallFactory {
 
+  static final ObjectMapper OBJECT_MAPPER = new ObjectMapper()
+    .setSerializationInclusion(JsonInclude.Include.NON_NULL);
```
codefromthecrypt, line 31: good. also not empty is good (ex don't serialize [] or {})

codefromthecrypt (approved):
Good work. only some nit things that were probably not better before your change.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/BasicAuthInterceptor.java
anuraaga: Yeah, we pass `releaseOnClose = true` to `ByteBufInputStream`'s constructor instead.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/BulkCallBuilder.java
anuraaga: Good idea.

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/internal/BulkIndexWriterTest.java
```diff
@@ -30,24 +30,31 @@
 
 public class BulkIndexWriterTest {
 
+  // Our usual test span depends on currentTime for testing span stores with TTL, but we'd prefer
+  // to have a fixed span here to avoid depending on business logic in test assertions.
+  static final Span STABLE_SPAN = CLIENT_SPAN.toBuilder()
+    .timestamp(100)
```
codefromthecrypt, line 36: fyi we have a TestObjects.TODAY, though not important to use it

codefromthecrypt: oh heh I see what you are doing. :)

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonReaders.java
anuraaga: I believe this essentially corresponds to `reader.endObject()` - Moshi will generally throw JsonDataException runtime exceptions on malformed JSON or EOF within one of its methods. I stuck with throwing IOException since all our methods declare throwing IOException but can change to IllegalArgumentException if it makes more sense. Returning `null` will probably not reproduce the old behavior.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonReaders.java
anuraaga: See above - moshi should have been throwing JsonDataException in the old code - I'm surprised that exception is not an IOException though since moshi's methods declare `throws IOException` too

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonSerializers.java
anuraaga: Yup - by the way you can still see the old code in the benchmarks folder. It calls `.ip` with ipv4 before ipv6, meaning ipv6 has precedence.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonSerializers.java
anuraaga: I went ahead and changed the "expected blah" to `IllegalArgumentException` and kept the end-of-input as IOException which feels write, though it's not clear.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonReaders.java
codefromthecrypt: anyway I wrote this code :) it wasn't planned to have multiple layers of catch. it would be a heck of a lot nicer to not have to eyeball if correct or not and instead just return null.

codefromthecrypt: ex if you look it does things like this

```
    return enterPath(reader, path1) != null ? enterPath(reader, path2) : null;
```

if you feel strongly about making it throw that's ok just it is tiring to fix little exceptions here and there.

anuraaga: Ah ok it does make more sense here. It means we're returning `null` where we used to throw, seems safe here though.

PR #2691, opened 2019-07-19
Add rootServiceAndSpanName function

tacigar:
I want to show root span information in search result list in Discover page (in #2656).
So I need a function which calculate the root span from span node tree.

I tried to implement it.
But I do not know if this implementation is correct...

WDYT? @adriancole

### DIFF
--- a/zipkin-lens/src/zipkin/trace.js
+++ b/zipkin-lens/src/zipkin/trace.js
@@ -14,7 +14,7 @@
 import _ from 'lodash';
 import moment from 'moment';
 import { compare } from './span-cleaner';
-import { getErrorType, newSpanRow } from './span-row';
+import { getErrorType, newSpanRow, getServiceName } from './span-row';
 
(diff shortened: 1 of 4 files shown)


tacigar:
> I want to show root span information in search result list in Discover page (in #2656).

Like this...

root

zipkin-lens/src/zipkin/trace.js
```diff
@@ -266,6 +266,14 @@ function addLayoutDetails(
   }
 }
 
+export function rootServiceAndSpanName(root) {
+  const { span } = root.queueRootMostSpans().shift();
```
codefromthecrypt: I think maybe we should be less precise on a headless span, and not name the detail pane.

so, use `root.span()` and if undefined, don't name this trace.

tacigar:
PTAL

codefromthecrypt:
looks good. if we want to later change the headless/incomplete root span name, we have a test to change!

PR #2692, opened 2019-07-19
Remove span details table if empty

### LINKED ISSUE #2617: lens: consider removing span details table if empty
When a span does not have details, the table with 'No rows found' takes up a lot of screen estate giving a noisy impression. Maybe the table could be omitted alltogether when empty ?
Selection_040

tacigar:
#fixes #2617 

Thank you @jorgheymans !


codefromthecrypt:
probably hard to start a test if one doesn't exist, but now would be a good time to add a test. our old codebase zipkin-ui got into a critical spot once where we had to do a lot of backfilling of tests and still never caught up.

tacigar:
OK, I'll add tests.

PR #2693, opened 2019-07-20
Performs cleanups in Elasticsearch code, seeking better test coverage

codefromthecrypt:
the coverage issues predated the recent move to armeria.


zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/BasicAuthInterceptor.java
```diff
@@ -57,31 +46,6 @@ public static BasicAuthInterceptor create(
   @Override
   public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Exception {
     ctx.addAdditionalRequestHeader(HttpHeaderNames.AUTHORIZATION, basicCredentials);
-    return HttpResponse.from(delegate().execute(ctx, req)
```
codefromthecrypt, line 60: the responsibility here was misplaced. we don't need to duplicate response parsing

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/SearchResultConverter.java
```diff
@@ -15,41 +15,36 @@
...
+    JsonParser hits = enterPath(JSON_FACTORY.createParser(toInputStream(content)), "hits", "hits");
+    if (hits == null || !hits.isExpectedStartArrayToken()) return Collections.emptyList();
 
     List<T> result = new ArrayList<>();
     while (hits.nextToken() != JsonToken.END_ARRAY) {
       JsonParser source = enterPath(hits, "_source");
-      if (source != null) {
-        result.add(adapter.parse(source));
-      }
-      hits.nextToken();
```
codefromthecrypt, line 51: this line `hits.nextToken();` was a bug, which could cause a spin cc @anuraaga

codefromthecrypt:
I plan to merge this on green because the code around logging may change a bit once I refactor some complexity in http call cloning

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/BodyConverters.java
```diff
@@ -13,42 +13,37 @@
...
-      SearchResultConverter.create(JsonSerializers.SPAN_PARSER);
-  static final HttpCall.BodyConverter<List<DependencyLink>> DEPENDENCY_LINKS =
-      new SearchResultConverter<DependencyLink>(JsonSerializers.DEPENDENCY_LINK_PARSER) {
-        @Override
-        public List<DependencyLink> convert(ByteBuffer content) throws IOException {
-          List<DependencyLink> result = super.convert(content);
-          return result.isEmpty() ? result : DependencyLinker.merge(result);
-        }
-      };
+  static final BodyConverter<Object> NULL = new BodyConverter<Object>() {
```
anuraaga: Just wondering, what do you think about using lambdas?

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/SearchResultConverter.java
anuraaga: Ah yup

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonReaders.java
```diff
@@ -101,5 +99,24 @@ static void visitNextOrSkip(JsonParser parser, String name, Set<String> result)
...
+      // The parser may not be at a token, yet. If that's the case advance.
+      if (currentToken == null) currentToken = parser.nextToken();
+
+      // If we are still not at the expected token, we could be an another or an empty body.
+      if (currentToken == JsonToken.START_OBJECT) return true;
+      if (shouldThrow) {
+        throw new IllegalArgumentException("Expected start object, was " + currentToken);
+      }
+      return false;
+    } catch (JsonParseException e) { // likely not json
```
anuraaga: catch `Throwable`? When methods shouldn't throw, like where `shouldThrow` is `false` here, it's easy to end up with a hang because an exception type changed.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -522,4 +499,22 @@ static boolean isIpAddress(String address) {
...
+      if (status == null) {
+        throw new IllegalArgumentException("Health status couldn't be read " + result);
+      }
+      if ("RED".equalsIgnoreCase(status.getText())) {
+        return CheckResult.failed(new IllegalStateException("Health status is RED"));
+      }
+      return CheckResult.OK;
+    }
+
+    @Override  public String toString() {
```
anuraaga: Extra space

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -522,4 +499,22 @@ static boolean isIpAddress(String address) {
 
   ElasticsearchStorage() {
   }
+
+  static final BodyConverter<CheckResult> READ_STATUS = new BodyConverter<CheckResult> (){
+    @Override public CheckResult convert(HttpData body) throws IOException {
+      String result = body.toStringUtf8();
```
anuraaga, line 514: I sort of wanted to untie these parsers from armeria or netty by using `ByteBuffer` but I guess it's ok. In that case, I'd pass in `ByteBuf` instead of `HttpData` since it's more flexible. e.g., a little better to do something like this to minimize copies for the success case.

```java
try (ByteBufInputStream stream = new ByteBufInputStream(buf.slice(), true)) {
  JsonParser status = enterPath(JSON_FACTORY.createParser(result), "status");
  if (status == null) {
    throw new IllegalArgumentException("Health status couldn't be read " + buf.toString(UTF_8));
  }
}
```

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonReaders.java
```diff
@@ -101,5 +99,24 @@ static void visitNextOrSkip(JsonParser parser, String name, Set<String> result)
     }
   }
 
-  JsonReaders() {}
+  static boolean checkStartObject(JsonParser parser, boolean shouldThrow) throws IOException {
+    try {
+      JsonToken currentToken = parser.currentToken();
+      // The parser may not be at a token, yet. If that's the case advance.
+      if (currentToken == null) currentToken = parser.nextToken();
```
anuraaga, line 106: Yeah was thinking about this - this is good

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
```diff
@@ -13,27 +13,44 @@
...
+
 public final class HttpCall<V> extends Call.Base<V> {
 
-  public interface BodyConverter<V> {
-    V convert(ByteBuffer content) throws IOException;
+  public static abstract class BodyConverter<V> {
+    /** Most convert with {@link HttpData#toStringUtf8()} or {@link #toInputStream(HttpData)} */
+    public abstract V convert(HttpData content) throws IOException;
+
+    // TODO: once https://github.com/line/armeria/issues/1918 is done, switch back to an interface
```
anuraaga: Since this is java 8, stick with being an interface and use a default method?

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
```diff
@@ -13,27 +13,44 @@
...
+    /** Use this when you don't need a string or only need to read the response once. */
+    protected InputStream toInputStream(HttpData content) {
+      if (content instanceof ByteBufHolder) {
+        return new ByteBufferBackedInputStream(((ByteBufHolder) content).content().nioBuffer());
```
anuraaga: `return new ByteBufInputStream(content.content())`

Without setting `true` it won't release, and it's the same semantics with slightly better performance.

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/HttpCallTest.java
```diff
@@ -120,16 +134,58 @@
...
+  }
+
+  @Test public void executionException_404() throws Exception {
+    MOCK_RESPONSE.set(AggregatedHttpResponse.of(HttpStatus.NOT_FOUND));
+
+    Call<?> call = http.newCall(REQUEST, BodyConverters.NULL);
+
+    try {
+      call.execute();
+      failBecauseExceptionWasNotThrown(FileNotFoundException.class);
```
anuraaga, line 160: `assertThatThrownBy` more readable

codefromthecrypt:
thanks for all the hints. one type isnt an interface so it cant be a
lambda. I can try to refactor it back but it was awaiting the input stream
change on httpdata

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonReaders.java
adriancole: yep!

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
codefromthecrypt: hmm not completely on board with this as we have so much complex code as it is. The cases we use string are not called often or in performance sensitive places. do you think calls to health check are going to significantly warrant this?

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
codefromthecrypt: yeah doing something like that now.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
codefromthecrypt: k

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/HttpCallTest.java
codefromthecrypt: let's do this in another change because we use this pattern maybe a hundred different places.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
codefromthecrypt: I will change this one anyway :P

anuraaga: Yeah it's probably not a big deal for these so don't necessarily need to change it (I was a bit more concerned about the index template call but realized it's memoized). I might experiment with the interface some more in another PR, and it definitely wouldn't be release-bocking. I tend to have some OCD about failure cases slowing down success cases ;)

codefromthecrypt: ugh. it far from pretty.

codefromthecrypt: yeah just thinking about the actual length of health check responses which is is small and how often we end up converting things to byte arrays anyway. I don't really like so much complex code which ends up being allocatey afterall.

codefromthecrypt: here's where I got stuck on this btw. One thing to keep in mind is that I barely understand the code now, and I suspect few in zIpkin do anymore. we do need to be careful how crazy netty we get especially in simple paths.

```
    default ByteBuf toByteBuf(HttpData content) {
      if (content instanceof ByteBufHolder) {
        return ((ByteBufHolder) content).content();
      } else {
        // what to do here?
      }
    }
```

anuraaga: Just for reference, it'd be `Unpooled.wrappedBuffer(content.array())` but yeah no worries. I agree we should keep the netty in one place as much as possible, have some ideas on achieving both will explore them later.

codefromthecrypt:
thanks

PR #2694, opened 2019-07-20
Always make requests even if health check is all down and catch all e…

anuraaga:
…xceptions, not just IOException

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
@@ -87,8 +87,7 @@ public final class HttpCall<V> extends Call.Base<V> {
       } else {
         try {
           callback.onSuccess(parseResponse(response, bodyConverter));
-        } catch (Throwable e) {
-          Call.propagateIfFatal(e);
+        } catch (IOException e) {
           callback.onError(e);
         }
       }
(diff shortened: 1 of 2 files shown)


anuraaga:
I guess since with this change it will always fallback when there are no health checked endpoints, maybe we don't need an extra flag.

Also we might be seeing some non-IOException causing hangs

codefromthecrypt (approved):

codefromthecrypt:
> I guess since with this change it will always fallback when there are no
> health checked endpoints, maybe we don't need an extra flag.
>
> Also we might be seeing some non-IOException causing hangs
>

I'm happy if hangs go away. Main thing is master isn't shippable until we
can get hang-free.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
```diff
@@ -88,7 +88,7 @@ public Factory(HttpClient httpClient) {
       } else {
         try {
           callback.onSuccess(parseResponse(response, bodyConverter));
-        } catch (IOException e) {
+        } catch (Throwable e) {
```
codefromthecrypt: usually I do either ` IOException|RuntimeException` or if doing Throwable, use propagateIfFatal

codefromthecrypt (approved):
thanks for the continued help, rag!

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
codefromthecrypt: I will scrub this in another pull request

codefromthecrypt: if you don't mind reverting this line, it won't drift :P

anuraaga: Makes sense - going to stick with `Throwable` since I've seen a lot of our code, I think following okio's example, use `AssertionError` which would hang I think.

anuraaga: Ok reverted

codefromthecrypt:
merge when ready.

PR #2696, opened 2019-07-20
Switches to always issue onError with throwable

codefromthecrypt:
This helps prevent subtle bugs. It will take a bit to write tests for these.

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
@@ -86,11 +86,14 @@ public final class HttpCall<V> extends Call.Base<V> {
       if (t != null) {
         callback.onError(t);
       } else {
+        V value = null;
         try {
-          callback.onSuccess(parseResponse(response, bodyConverter));
-        } catch (IOException e) {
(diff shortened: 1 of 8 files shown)


zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
```diff
@@ -86,11 +86,14 @@ public Factory(HttpClient httpClient) {
...
         try {
-          callback.onSuccess(parseResponse(response, bodyConverter));
-        } catch (IOException e) {
-          callback.onError(e);
+          value = parseResponse(response, bodyConverter);
+        } catch (Throwable t1) {
+          propagateIfFatal(t1);
+          callback.onError(t1);
         }
+        if (value != null) callback.onSuccess(value);
```
anuraaga: Think this needs to go right below line 91. `parseResponse` can return `null` (line 133), which will cause this future to never complete.

codefromthecrypt:
thanks.. good eye. I will also make a test to prove it.

anuraaga:
Took care of it so it can get back to the CI

codefromthecrypt:
thanks. actually I need to pause until monday so if you want to merge
without tests I can backfill monday

anuraaga:
There are tests :)

PR #2703, opened 2019-07-22
Run span accepting logic on blocking executor.

anuraaga:
Our storage components expose asynchronous APIs, but this is somewhat deceptive since some of their parts run blocking logic in constructors which isn't allowed. We currently are not using event loops properly during the time the server gets ready but it's ok since the blocking calls eventually complete and aren't ever made again. Switching to Armeria client for ES though means that the blocking call itself also needs to use the event loop, and if a server and client call get assigned to the same event loop, the thread hangs forever.

This has the span accepting logic run on a blocking executor thread. It's a bit more overhead, but shouldn't be too bad.

I also found a trick in the cassandra client library for preventing accidentally deadlocking like this.

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinGrpcCollector.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinGrpcCollector.java
@@ -13,7 +13,9 @@
  */
 package zipkin2.server.internal;
 
+import com.linecorp.armeria.common.CommonPools;
 import com.linecorp.armeria.common.grpc.protocol.AbstractUnsafeUnaryGrpcService;
+import com.linecorp.armeria.server.ServiceRequestContext;
 import com.linecorp.armeria.spring.ArmeriaServerConfigurator;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
@@ -67,7 +69,11 @@ final class ZipkinGrpcCollector {
       try {
         CompletableFutureCallback result = new CompletableFutureCallback();
         List<Span> spans = SpanBytesDecoder.PROTO3.decodeList(bytes.nioBuffer());
-        collector.accept(spans, result);
+
+        ServiceRequestContext.mapCurrent(
+          ctx -> ctx.makeContextAware(ctx.blockingTaskExecutor()),
+          CommonPools::blockingTaskExecutor).execute(() -> collector.accept(spans, result));
+
         return result;
       } finally {
         bytes.release();
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinGrpcCollector.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinGrpcCollector.java
@@ -70,6 +70,8 @@ final class ZipkinGrpcCollector {
         CompletableFutureCallback result = new CompletableFutureCallback();
         List<Span> spans = SpanBytesDecoder.PROTO3.decodeList(bytes.nioBuffer());
 
+        // collector.accept might block so need to move off the event loop. We make sure the
+        // callback is context aware to continue the trace.
         ServiceRequestContext.mapCurrent(
           ctx -> ctx.makeContextAware(ctx.blockingTaskExecutor()),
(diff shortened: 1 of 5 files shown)


anuraaga:
I initially tried an approach of eagerly initializing storage to get the blocking out of the way, but it means storage has to be up before zipkin server, which I guess we don't require.

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinGrpcCollector.java
```diff
@@ -67,7 +69,11 @@
       try {
         CompletableFutureCallback result = new CompletableFutureCallback();
         List<Span> spans = SpanBytesDecoder.PROTO3.decodeList(bytes.nioBuffer());
-        collector.accept(spans, result);
+
+        ServiceRequestContext.mapCurrent(
```
codefromthecrypt, line 75: maybe copy same comment from collector here about rationale and blocking

contextaware for tracing and blocking because we dont yet have a way to tell if call.enqueue is actually non blocking

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -263,6 +263,7 @@ public final Builder dateSeparator(char dateSeparator) {
   public abstract int namesLookback();
 
   @Override
+  @Memoized // Memoized to reduce garbage as this is effectively a singleton.
```
codefromthecrypt: why. is this called per request? if so this also adds a lock internally. can we not add this at the moment until analyzed as this assumption isnt carried to the several other impls. the only other consumer of this is the http api and that isnt hit often enough for specializations.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
```diff
@@ -93,6 +94,12 @@ public Factory(HttpClient httpClient) {
   }
 
   @Override protected V doExecute() throws IOException {
+    for (EventExecutor eventLoop : httpClient.factory().eventLoopGroup()) {
```
codefromthecrypt, line 98: very good idea though ISE is usually state that can resolve and I doubt this would. maybe just runtimeexception? ps I dont mind backfilling tests but I am losing track. can you add TODO testme on things like this?

codefromthecrypt (approved):
thanks for the help.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
anuraaga: It's called per request. I noticed Cassandra memoizes these too, so figured doing similar would be ok. Was hoping to memoize at this point, since the constructors access multiple memoized fields themselves - but it's not a big deal so reverted.

codefromthecrypt (approved):
cool beans

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
codefromthecrypt: cassandra is different reason

```
  /** {@inheritDoc} Memoized in order to avoid re-preparing statements */
```

Right now the main attention is ideally drawn to important things like data or setup dependencies. I'm trying to help keep focus until things are working. I think this is really not going to affect performance in a holistic way to memoize because it is focusing on object allocation. Let's keep focus where it is important. this part has different goals than the core library which results in measurable app overhead. otoh, If we change code here to save 2 of 4000 allocations, probably more like racing stripes on a yugo.

make sense?

codefromthecrypt: ps the annoying dodging or re-preparing statements is something Cassandra 4 driver gets rid of. (you get warning in log now if you accidentally do it). otoh I dont have stomache for another big refactor at the moment because api 4 is very different than 3. good news though is when that occurs we can remove that memorization or have the option to

PR #2704, opened 2019-07-22
Avoid fetching traces when the search criteria is not changed

tacigar:
In the previous implementation, fetch processing was performed each time the search page was opened, even if the search criteria were the same.
This implementation degrades UX significantly because fetch processing is slow.
We should avoid fetching traces when the search criteria is not changed.
This PR do so.

Related: #2685 
If we do not avoid unnecessary fetching, the page transition of `search page` → `see details` → `back button` → `search page` UX will be the worst....


zipkin-lens/src/actions/traces-action.test.js
```diff
@@ -85,6 +76,10 @@ describe('traces async actions', () => {
         traces: rawTraces,
         correctedTraceMap,
         traceSummaries,
+        lastFetchingParams: {
```
codefromthecrypt: maybe lastQueryParams? if the relates to the query api (/api/v2/traces) https://zipkin.io/zipkin-api/#/default/get_traces

codefromthecrypt (commented):
only thing curious about is if a threshold is useful. Ex if 5 or 10 seconds later, it will do a refresh when clicking back. Maybe that isn't useful? not sure.

tacigar:
I think a threshold is not useful...
If the user want to refresh search results, the user only push Enter key.

tacigar:
PTAL 🙇

codefromthecrypt:
agree about press enter is intuitive. thanks for thinking about it!

PR #2700, opened 2019-07-21
Moves to Armeria with HttpTracing support

codefromthecrypt:
This is using Armeria 0.89 which has new tracing support from @kojilin https://github.com/line/armeria/pull/1906

The following is from self-tracing elasticsearch. 
Screenshot 2019-07-26 at 5 21 47 PM

PR #2706, opened 2019-07-24
Move logging configuration to ES storage component

### LINKED ISSUE #2695: don't ES_HTTP_LOGGING elasticsearch healthcheck
it is super difficult to see what is happening in requests when ES_HTTP_LOGGING is set because there are so many health checks going on. I know we talked about making this not happen on health check but raising an issue so that we don't lose track.

anuraaga:
…ontrol how it's applied (not applied to health check).

Fixes #2695


codefromthecrypt:
travis should check this 10-15m faster than before!

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -207,6 +220,22 @@ public final Builder dateSeparator(char dateSeparator) {
...
+     * <p>This is only for use from tests and ZipkinElasticsearchStorageConfiguration.
+     */
+    public abstract Builder httpLogging(HttpLoggingLevel httpLoggingLevel);
+
+    /**
+     * Sets the {@link Tracing} to use to trace HTTP requests made by the Elasticsearch client. If
+     * not set, tracing will be disabled.
+     *
+     * <p>This is only for use from tests and ZipkinElasticsearchStorageConfiguration.
+     */
```
codefromthecrypt: hmm currently our only api exposure to brave is in zipkin-server. is it possible to keep it there? none of the other libraries have a direct dep on something besides what they use directly

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -462,6 +495,33 @@ public HttpClient httpClient() {
...
+      }
+    }
+
+    if (tracing() != null) {
+      options.decorator(BraveClient.newDecorator(tracing(), "elasticsearch"));
+    }
+
     clientCustomizer().accept(options);
     HttpClientBuilder client = new HttpClientBuilder(clientUrl)
       .factory(clientFactory())
```
codefromthecrypt, line 511: I wonder if we can somehow bind auth to the client factory.  if so it makes the rest a little easier I think.

codefromthecrypt: thinking being that somehow we can get the health checks to apply the most basic and required config. otherwise we could expose a separate  config for health checks and in worst case pass same interceptor to both places.

codefromthecrypt:
If it isn't practical to decouple the health check http config from the real http config, probably the better choice would be to have the health check config use tracing, but still unsampled. We can switch the headers to b3 single to lower the impact also. I want to be conscious of the amount of energy spent on health checking elasticsearch, which is more than some of our storage impls get at all :P

codefromthecrypt:
basically the most annoying thing is health checks polluting the output of ES_HTTP_LOGGING. by reverting the tracing back where it was, we can still accomplish the more important and visible change.

codefromthecrypt:
The combo of what you did here moving logging, and being less verbose in b3 headers via https://github.com/openzipkin/zipkin/pull/2707 should be enough I think. Let's revert the tracing part and get this merged?

anuraaga:
Yup will clean it up by tomorrow. Also realized that IIUC https://github.com/openzipkin/zipkin/pull/2700 should prevent the health check from being sampled since health check isn't in the context of a server request

codefromthecrypt:
Yup will clean it up by tomorrow. Also realized that IIUC #2700
> <https://github.com/openzipkin/zipkin/pull/2700> should prevent the
> health check from being sampled since health check isn't in the context of
> a server request
>

right it will prevent it from being sampled, but for consistency we still
propagate sampled=0 downstream :P Thanks for plugging along, Rag!

anuraaga (commented):
Reverted tracing change

codefromthecrypt:
so happy to have this change in.. thanks @anuraaga

PR #2707, opened 2019-07-24
Leak less stuff into Elasticsearch requests with b3 single

codefromthecrypt:
Currently, when self-tracing is enabled, all requests, traced or not,
end up with B3 headers. This reduces the load on Elasticsearch by using
b3 single, which is also less cluttered when viewing api requests.


codefromthecrypt:
https://github.com/openzipkin/brave/issues/949 is related

anuraaga (approved):

PR #2708, opened 2019-07-26
Fallback to static endpoints instead of possibly DNS-resolved ones si…

anuraaga:
…nce DNS may still take some time to resolve on first request.


codefromthecrypt:
thanks as this is very repeatable I can try it locally!

anuraaga:
It repros locally too? I'm quite impressed by how slow DNS is in practice...

codefromthecrypt:
things are good!

PR #2709, opened 2019-07-26
Set custom name for elasticsearch requests for nicer traces.

anuraaga:
The default Armeria parser uses `RpcRequest.method()` as the span name if it's available - normally is for RPC like Thrift or gRPC but we can do it too.


codefromthecrypt:
neat. will test it locally in a bit

codefromthecrypt (approved):
I added a commit because it seemed easier to rendezvous this way cc @trustin @kojilin 

Screenshot 2019-07-26 at 9 50 57 PM

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/HttpCallTest.java
```diff
@@ -192,6 +197,24 @@
...
+    http = new HttpCall.Factory(new HttpClientBuilder(server.httpUri("/"))
+      .decorator((client, ctx, req) -> {
+        ctx.log().addListener(log::set, RequestLogAvailability.COMPLETE);
+        return client.execute(ctx, req);
+      })
+      .build());
+
+    http.newCall(REQUEST, BodyConverters.NULL, "custom-name").execute();
+
+    await().untilAsserted(() -> assertThat(log).doesNotHaveValue(null));
```
codefromthecrypt, line 212: I suspect I broke this test :P

zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/ZipkinElasticsearchStorageConfiguration.java
```diff
@@ -161,8 +164,20 @@
...
       };
     }
+
     HttpTracing httpTracing = maybeHttpTracing.get().clientOf("elasticsearch");
-    return client -> client.decorator(BraveClient.newDecorator(httpTracing));
+    SpanCustomizer spanCustomizer = CurrentSpanCustomizer.create(httpTracing.tracing());
+
+    return client -> {
+      client.decorator((delegate, ctx, req) -> {
+        if (ctx.hasAttr(HttpCall.NAME)) { // override the span name if set
```
anuraaga, line 173: Yeah this is similar to my first commit on this branch before thinking there's no need for a decorator.

I would say this decorator helps decouple zipkin-server's span names from Armeria's logging-related implementation details. If that's the goal of the commit, this is good and ready to merge. As an Armeria advocate though, I'd personally rather have zipkin and users both use similar best practices where possible - having logs and span names the same seemed nice and I'm leaning towards not adding a similar extension point (XXX.name) upstream when I realized setting `RpcRequest.method()` covers both. Admittedly, the API for setting a custom name is not so easy to use so upstream probably could improve there - I filed https://github.com/line/armeria/issues/1931 for that.

What do you think?

codefromthecrypt: as long as the name ends up the same as it does now (ex not prefixed with a pseudo class name etc) I am fine if you want to go back later and change things once the other thing is merged. I think the mechanics here dont bleed to others as our modules are only used by us. so easy to change later

codefromthecrypt:
thanks for the help here.. will look at other details to see if we can release in the next few days or not!

PR #2710, opened 2019-07-29
Uses more things from Armeria 0.89 and more polish

codefromthecrypt:


zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
```diff
@@ -46,16 +43,19 @@
...
-    // TODO: once https://github.com/line/armeria/issues/1918 is done, switch back to an interface
-    default InputStream toInputStream(HttpData content) {
-      if (content instanceof ByteBufHolder) {
-        return new ByteBufInputStream(((ByteBufHolder) content).content());
-      } else {
-        return content.toInputStream();
+  public interface InputStreamConverter<V> extends BodyConverter<V> {
+    V convert(InputStream content) throws IOException;
+
+    @Override default V convert(HttpData content) throws IOException {
```
anuraaga, line 53: Wondering, what do you think about this

```java
V convert(JsonParser parser, Supplier<String> contentString);

@Override default V convert(HttpData content) throws IOException {
  try (InputStream stream = content.toInputStream()) {
    return convert(JSON_FACTORY.createParser(stream), content::toStringUtf8);
  } finally {
    ReferenceCountUtil.safeRelease(content);
  }
}
```

I think all our conversions are of JSON so can pass `JsonParser` (or leave it as `InputStream` anyways), and it also allows the converters to have access to the full content string if they want it for an error message or something. Maybe it's too complex (a bit hesitant to make suggestions that add any complexity at this point...).

anuraaga (approved):

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
codefromthecrypt: will think about it. right now I need to unwind the client which is unrelated and get back to this after.. thanks for the suggestion!

codefromthecrypt: I think I get this.. I've a nag in head to help weave this in after getting aws working. mainly an issue of drift control at the moment

PR #2713, opened 2019-07-30
Revive scatter chart

### LINKED ISSUE #2636: Revive the scatter graph of searching trace page in Lens
This is a proposal.
I want to get feedbacks.

Early Lens had a scatter graph in the searching trace page like the below.

old-lens

But it was removed because trace data is sampled data so the scatter graph can show only the part of all traces.
I agreed with that opinion at that time.

But after seeing Lens page thousand times, I have became to think this scatter graph is very useful now.

For example, when user's service get a latency problem, if there is not the scatter graph, I think the user operates like the following in UI.

1. Set criterion for searching traces and click the search button.
1. Sort traces using dropdown menu.
1. Check the durations of the trace list in order from the top.

If there is the scatter graph, we can find the desired trace faster.
We just look at the scatter graph and click on the traces with the characteristic duration in the graph.

I think this is more intuitive and has higher usability.
What do you think?

tacigar:
Related & fixes #2636 

I know there are the pros and cons about Scatter Chart.
I want to get feedback.

scatter

cc: @narayaruna

### DIFF
--- a/zipkin-lens/package-lock.json
+++ b/zipkin-lens/package-lock.json
@@ -6367,6 +6367,73 @@
         "type": "^1.0.1"
       }
     },
+    "d3-array": {
+      "version": "1.2.4",
+      "resolved": "https://registry.npmjs.org/d3-array/-/d3-array-1.2.4.tgz",
+      "integrity": "sha512-KHW6M86R+FUPYGb3R5XiYjXPq7VzwxZ22buHhAEVG5ztoEcZZMLov530mmccaqA1GghZArjQV46fuc8kUqhhHw=="
+    },
+    "d3-collection": {
+      "version": "1.0.7",
+      "resolved": "https://registry.npmjs.org/d3-collection/-/d3-collection-1.0.7.tgz",
+      "integrity": "sha512-ii0/r5f4sjKNTfh84Di+DpztYwqKhEyUlKoPrzUFfeSkWxjW49xU2QzO9qrPrNkpdI0XJkfzvmTu8V2Zylln6A=="
+    },
+    "d3-color": {
+      "version": "1.2.8",
+      "resolved": "https://registry.npmjs.org/d3-color/-/d3-color-1.2.8.tgz",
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
cc @openzipkin/core indicate your opinion on this.. we do want to get out of a thrash on the topic. on one hand, yes it isn't statistically significant. on the other hand, it is a different way to navigate to a trace ID.

codefromthecrypt:
one thing this does is similar to the new twitter UI. it takes a large amount of space away from viewing results, pushing the eyes down and center.

I'm wondering if we can revisit this after we add a topological view also simultaneously considered.

Considering that is also a view based on aspects including latency I'm wondering if it is really necessary to also add more here. The other problem with adding this is it temps a lot of small adjustments and feature requests like zoom in the scatter etc.

tacigar:
Thank you for your feed back @adriancole.

> one thing this does is similar to the new twitter UI. it takes a large amount of space away from viewing results, pushing the eyes down and center.

Yes I totally agree on this.
But I think we can set the toggle button as follows.

scatter-toggle

The reason I want a scatter chart is that if we look at the scatter chart we can see at a glance which traces are causing problems with latency.

Ofcourse, I think we can revisit this after adding a topological view.

codefromthecrypt:
the toggle if collapsed by default could be a way out.


speaking from experience on another topic you should bear in mind. The more
features we add, the more reliance we are on more hands to do anything. For
example, consider if you switched careers because of your love of ramen.
Whoever maintains the code next needs to maintain every feature we have
added. This means a hard choice on what features to add, are they worth it?
are they worth the next person's time? this is why we try the "rule of
three" to make sure that others buy-in not just to the idea of if the
feature seems nice, but they personally would help with maintainance or
questions about it. This paragraph is not saying scatterplot would break
the camel's back.. just I'm in a position personally right now of
reflection and I think we add things faster than we can support them
sometimes. it is worth considering where we spend effort and how much value
we get from unit of maintenance.

tacigar:
Oops!!
I accidentally deleted the branch ... 😭 
Sorry!!

codefromthecrypt:
heh yikes

PR #2718, opened 2019-07-31
Enlists load related errors from cassandra in storage throttling

codefromthecrypt:
This classifies the following exception trees as possibly recoverable.
The storage throttling feature should back off when encountered.

* BusyPoolException
* BusyConnectionException
* QueryConsistencyException

cc @openzipkin/cassandra @Logic-32

### DIFF
--- a/zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/internal/call/ResultSetFutureCall.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/internal/call/ResultSetFutureCall.java
@@ -15,11 +15,15 @@ package zipkin2.storage.cassandra.internal.call;
 
 import com.datastax.driver.core.ResultSet;
 import com.datastax.driver.core.ResultSetFuture;
+import com.datastax.driver.core.exceptions.BusyConnectionException;
+import com.datastax.driver.core.exceptions.BusyPoolException;
 import com.datastax.driver.core.exceptions.DriverException;
 import com.datastax.driver.core.exceptions.DriverInternalError;
+import com.datastax.driver.core.exceptions.QueryConsistencyException;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.Uninterruptibles;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.RejectedExecutionException;
 import zipkin2.Call;
 import zipkin2.Callback;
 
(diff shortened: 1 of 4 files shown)


zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/internal/call/ResultSetFutureCall.java
```diff
@@ -76,16 +80,21 @@ protected final boolean doIsCanceled() {
...
+        return Uninterruptibles.getUninterruptibly(future);
+      } catch (ExecutionException e) {
+        Throwable cause = e.getCause();
+        if (cause instanceof Error) throw ((Error) cause);
+        if (cause instanceof DriverException) throw ((DriverException) cause).copy();
+        throw new DriverInternalError("Unexpected exception thrown", cause);
+      }
+    } catch (QueryConsistencyException | BusyConnectionException | BusyPoolException e) {
+      // Our throttling function relies on rejected execution when encountering load-specific errors
+      throw new RejectedExecutionException(e.getMessage(), e);
```
Logic-32, line 97: Mentioned this in a related review, but rather than stutter the exception message can we put something more descriptive in it?  E.g. "Cassandra overloaded; please enable storage throttling so we can attempt to slow request rate."

codefromthecrypt: on the line of stuttering :) this ends up in heath checks and 500 errors on read. having a relevant top-level message is important. Errors are not isolated to writes even if the code you are thinking about is.

codefromthecrypt: also storage throttling isn't the only way. for example, cassandra has its own backoff queue local to handle things like this. it is just allowing those who alternatively want to use this. I really don't think we should be advertizing storage throttling in expection messages as an answer as depending on the product different tools can be used, including better tuning.

codefromthecrypt: if we are going to monkey with messages, make them misleading, obfuscated etc, I'd rather refactor out the RejectedExecutionException signal. so that we don't have to

codefromthecrypt: just right now there's tons of work and little hands to help with it, it is much easier to use the source data as we do with elasticsearch already.

Ex below we can see the actual result in health checks or 500 errors resulting from browser etc.
```
        if (status != null && status.intValue() == 429) {
          toThrow = new RejectedExecutionException(tree.toString());
```

To change that here would just make troubleshooting harder

codefromthecrypt:
raised #2721 about nervousness about use of wrappers. someone who feels strongly can pony time to solve it and then revert this.

PR #2719, opened 2019-07-31
Translate UnprocessedRequestException to RejectedExecutionException.

### LINKED ISSUE #2714: enlist empty endpoint group exceptions as RejectedExecution
With the new elasticsearch code, we will have dynamic endpoints. This is good because it can shrink and grow due to health. It is also possible to get an empty endpoint group exception invoking a request. This needs to integrate with the storage throttling code, and I think the easiest way is to throw a RejectedExecutionException when the endpoints are empty

cc @Logic-32

anuraaga:
Also unwrap `CompletionException`

Fixes #2714 

Log message looks like

```
2019-07-31 15:43:17.433  WARN 18272 --- [cking-tasks-1-3] c.l.a.i.a.DefaultExceptionHandler        : [id: 0x124a5aa0, L:/0:0:0:0:0:0:0:1:9411 - R:/0:0:0:0:0:0:0:1:51877][h1c://desktop-6rdo32o:9411/zipkin/api/v2/services#GET] Unhandled exception from an annotated service:

java.util.concurrent.RejectedExecutionException: Could not process request.
	at zipkin2.elasticsearch.internal.client.HttpCall.lambda$sendRequest$3(HttpCall.java:168) ~[classes/:?]
	at java.util.concurrent.CompletableFuture.uniExceptionally(CompletableFuture.java:986) ~[?:?]
	at java.util.concurrent.CompletableFuture$UniExceptionally.tryFire(CompletableFuture.java:970) ~[?:?]
	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506) ~[?:?]
	at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2088) ~[?:?]
	at com.linecorp.armeria.common.HttpMessageAggregator.fail(HttpMessageAggregator.java:146) ~[armeria-0.89.0.jar:?]
	at com.linecorp.armeria.common.HttpMessageAggregator.onError(HttpMessageAggregator.java:60) ~[armeria-0.89.0.jar:?]
	at com.linecorp.armeria.common.stream.AbstractStreamMessage$CloseEvent.notifySubscriber(AbstractStreamMessage.java:370) ~[armeria-0.89.0.jar:?]
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.notifySubscriberOfCloseEvent(DefaultStreamMessage.java:194) ~[armeria-0.89.0.jar:?]
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.handleCloseEvent(DefaultStreamMessage.java:363) ~[armeria-0.89.0.jar:?]
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.notifySubscriber0(DefaultStreamMessage.java:306) ~[armeria-0.89.0.jar:?]
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.doRequest(DefaultStreamMessage.java:181) ~[armeria-0.89.0.jar:?]
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.request(DefaultStreamMessage.java:166) ~[armeria-0.89.0.jar:?]
	at com.linecorp.armeria.common.stream.AbstractStreamMessage$SubscriptionImpl.request(AbstractStreamMessage.java:309) ~[armeria-0.89.0.jar:?]
	at com.linecorp.armeria.common.HttpMessageAggregator.onSubscribe(HttpMessageAggregator.java:55) ~[armeria-0.89.0.jar:?]
	at com.linecorp.armeria.common.stream.DefaultStreamMessage.lambda$subscribe$0(DefaultStreamMessage.java:124) ~[armeria-0.89.0.jar:?]
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163) ~[netty-common-4.1.38.Final.jar:4.1.38.Final]
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:416) ~[netty-common-4.1.38.Final.jar:4.1.38.Final]
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:515) ~[netty-transport-4.1.38.Final.jar:4.1.38.Final]
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:918) ~[netty-common-4.1.38.Final.jar:4.1.38.Final]
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74) ~[netty-common-4.1.38.Final.jar:4.1.38.Final]
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) [netty-common-4.1.38.Final.jar:4.1.38.Final]
	at java.lang.Thread.run(Thread.java:834) [?:?]
Caused by: com.linecorp.armeria.client.endpoint.EndpointGroupException: com.linecorp.armeria.client.endpoint.dns.DnsAddressEndpointGroup@12c75a76 is empty
```

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
@@ -172,6 +172,8 @@ public final class HttpCall<V> extends Call.Base<V> {
     responseFuture = responseFuture.exceptionally(t -> {
       if (t instanceof UnprocessedRequestException) {
         Throwable cause = t.getCause();
+        // Go ahead and reduce the output in logs since this is usually a configuration or
+        // infrastructure issue and the Armeria stack trace won't help debugging that.
         Exceptions.clearTrace(cause);
         throw new RejectedExecutionException("Rejected execution: " + cause.getMessage(), cause);
       } else {
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
@@ -35,6 +35,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.RejectedExecutionException;
 import zipkin2.Call;
 import zipkin2.Callback;
@@ -115,7 +116,14 @@ public final class HttpCall<V> extends Call.Base<V> {
           + "Either use doEnqueue() or run this in a separate thread.");
       }
     }
-    AggregatedHttpResponse response = sendRequest().join();
+    final AggregatedHttpResponse response;
+    try {
+      response = sendRequest().join();
+    } catch (CompletionException e) {
+      propagateIfFatal(e);
+      Exceptions.throwUnsafely(e.getCause());
+      return null;  // Unreachable
+    }
     return parseResponse(response, bodyConverter);
   }
 
@@ -165,7 +173,7 @@ public final class HttpCall<V> extends Call.Base<V> {
       if (t instanceof UnprocessedRequestException) {
         Throwable cause = t.getCause();
         Exceptions.clearTrace(cause);
-        throw new RejectedExecutionException("Could not process request.", cause);
+        throw new RejectedExecutionException("Rejected execution: " + cause.getMessage(), cause);
       } else {
         Exceptions.throwUnsafely(t);
       }
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
@@ -16,6 +16,7 @@ package zipkin2.elasticsearch.internal.client;
 import com.fasterxml.jackson.core.JsonParser;
 import com.linecorp.armeria.client.Clients;
 import com.linecorp.armeria.client.HttpClient;
+import com.linecorp.armeria.client.UnprocessedRequestException;
 import com.linecorp.armeria.common.AggregatedHttpRequest;
 import com.linecorp.armeria.common.AggregatedHttpResponse;
 import com.linecorp.armeria.common.HttpData;
@@ -23,6 +24,7 @@ import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.HttpStatus;
 import com.linecorp.armeria.common.HttpStatusClass;
 import com.linecorp.armeria.common.RequestContext;
+import com.linecorp.armeria.common.util.Exceptions;
 import com.linecorp.armeria.common.util.SafeCloseable;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufHolder;
@@ -33,6 +35,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.RejectedExecutionException;
 import zipkin2.Call;
 import zipkin2.Callback;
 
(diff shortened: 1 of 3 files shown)


zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
```diff
@@ -158,6 +161,16 @@ public Factory(HttpClient httpClient) {
         ctx -> response.aggregateWithPooledObjects(ctx.eventLoop(), ctx.alloc()),
         // This should never be used in practice since the module runs in an Armeria server.
         response::aggregate);
+    responseFuture = responseFuture.exceptionally(t -> {
+      if (t instanceof UnprocessedRequestException) {
+        Throwable cause = t.getCause();
+        Exceptions.clearTrace(cause);
+        throw new RejectedExecutionException("Could not process request.", cause);
```
codefromthecrypt: would be nice to have the message in the exception here because this will also show up on health check potentially, and that only shows the direct message iirc. IOTW I don't think "Could not process request." is more telling than the cause.getMessage()

Logic-32: I agree that the current message isn't super useful but stuttering an exception message is generally considered bad form and just makes stack traces redundant/noisy.  Is there a more verbose message we can put here that illustrates the point without being repetitive?  I'm not sure what causes an UnprocessedRequestException exactly so I can't immediately offer any suggestions.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
```diff
@@ -158,6 +161,16 @@ public Factory(HttpClient httpClient) {
         ctx -> response.aggregateWithPooledObjects(ctx.eventLoop(), ctx.alloc()),
         // This should never be used in practice since the module runs in an Armeria server.
         response::aggregate);
+    responseFuture = responseFuture.exceptionally(t -> {
+      if (t instanceof UnprocessedRequestException) {
+        Throwable cause = t.getCause();
+        Exceptions.clearTrace(cause);
```
Logic-32, line 177: What purpose does this serve?  Might be my own personal opinion, but I generally like to see all the causes when possible.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
codefromthecrypt: the reason I say to replace the message is that it is artificial that we are wrapping at all, that it makes the health check unreadable, and also it will be hard to guess the format of the upstream message in a way that doesn't change. This would be obvious with a unit test.

In other words, I think it serves us better to make things like health checks usable vs spend time on an artificial thing only needed because we aren't making an exception registry for the throttling system

codefromthecrypt:
I would rather not merge this vs add a wrapped message as where I sit on gitter support I really have no desire to have to translate this from memory into the actual problem #2721

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
anuraaga: I noticed in another change that we don't like to have verbose exception output when it's not really an issue of debugging code (e.g. DNS resolution failure). This is similar, even with a stacktrace for this exception, users wouldn't be able to explore much.

anuraaga (commented):
Reflected the cause message into the top-level exception and also removed `CompletionException` wrapping when calling `execute`

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/HttpCallTest.java
```diff
@@ -214,6 +218,22 @@
...
+    http = new HttpCall.Factory(new HttpClientBuilder(server.httpUri("/"))
+      .decorator((client, ctx, req) -> {
+        throw new UnprocessedRequestException("Could not process request.",
+          new EndpointGroupException("No endpoints"));
+      })
+      .build());
+
+    assertThatThrownBy(() -> http.newCall(REQUEST, BodyConverters.NULL, "test").execute())
+      .isInstanceOf(RejectedExecutionException.class)
+      .hasMessageContaining("No endpoints");
```
codefromthecrypt: better to do equals as it can show potentially horrible messages :P

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
codefromthecrypt: makes sense. comment?

anuraaga: Sure

codefromthecrypt:
cool thanks! merging and will add a low priority but merge any time change to allow use of exception types other than Rejected

codefromthecrypt:
cool thanks! merging and will add a low priority but merge any time change to allow use of exception types other than Rejected

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
Logic-32: Works for me :)

PR #2722, opened 2019-08-01
Add support for health-checking ES client.

### LINKED ISSUE #2720: Implement client-side check for multiple elasticsearch servers
This is a backported rationale which may affect the design already in master. TL;DR; is we should tread carefully when adding health checks so that we don't shoot ourselves in the foot or end up with emergency bug fixes to turn them off.

# Background

We raised a #1611 a while back for the ability to handle multiple TLS endpoints for elasticsearch in at least the same way we do for non TLS (connection based round-robin)
https://github.com/square/okhttp/issues/4530

Now that we are using Armeria, we have multiple options for health checking and should be very careful about what mechanisms we use and that it can be turned off.

## On what mechanism to use?
If we do health checking at all, minimally we could use TCP/TLS connection level. This is cheap and unaffected by what Elasticsearch may return in an http request. It might be advantageous to use this verses ES health endpoint because of a few factors:

### Reasons to use http level checks
  * ES may have a status where a node is technically available, but out of service

### Reasons to not use http level checks
  * People upgrading may not be prepared to add the additional load or latency implied by ES http-level health check
  * ES itself has been known to have unpredictable behavior at times, hanging or otherwise having load problems. Adding an http layer check can add risk.
  * ES themselves do not do any active checking using their own apis https://github.com/elastic/elasticsearch/blob/master/client/rest/src/main/java/org/elasticsearch/client/RestClient.java#L178-L194
  * Flurry of requests can hide real problems when debugging ( ex literally debugging, wireshark or `--logging.level.com.linecorp.armeria.logging.traffic.client=trace`)

## Health checks should be disablable
As the code we are using in armeria is new, that we don't have experience in how this will affect the cluster itself, and also that debugging is much harder with another system involved, we should allow disabling health checks. For example, setting zipkin.storage.elasticsearch.health-check.enabled=false

anuraaga:
By default, health-check is enabled with a check interval of 3 seconds. This allows the client to skip nodes that do not return a successful health check. When health-check is disabled, no check requests will be made, but if a node becomes unusable it will still have requests sent to it as part of the round-robin.

Renamed a test named health check since it didn't seem to be testing health check, but I needed to :P
Also raised the timeout on `defersIOUntilFirstUse`, on my computer I couldn't get it to pass on master either.

Fixes #2720

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/LazyHttpClientImpl.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/LazyHttpClientImpl.java
@@ -15,11 +15,13 @@ package zipkin2.server.internal.elasticsearch;
 
 import com.linecorp.armeria.client.Endpoint;
 import com.linecorp.armeria.client.HttpClient;
+import com.linecorp.armeria.client.endpoint.DynamicEndpointGroup;
 import com.linecorp.armeria.client.endpoint.EndpointGroup;
 import com.linecorp.armeria.client.endpoint.EndpointGroupRegistry;
 import com.linecorp.armeria.client.endpoint.StaticEndpointGroup;
 import com.linecorp.armeria.client.endpoint.healthcheck.HttpHealthCheckedEndpointGroup;
 import com.linecorp.armeria.client.endpoint.healthcheck.HttpHealthCheckedEndpointGroupBuilder;
+import com.linecorp.armeria.client.retry.Backoff;
 import com.linecorp.armeria.common.SessionProtocol;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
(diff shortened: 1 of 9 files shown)


zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/LazyHttpClientImpl.java
```diff
@@ -67,28 +73,47 @@ Endpoint getEndpoint() {
...
         .withClientOptions(factory::configureOptionsExceptLogging)
+        .retryBackoff(Backoff.fixed(healthCheck.getInterval().toMillis()))
         .build();
 
     // Since we aren't holding up server startup, or sitting on the event loop, it is ok to block.
     // The alternative is round-robin, which could be unlucky and hit a bad node first.
     //
-    // We are blocking a second as this should be enough time for a health check to respond
-    // TODO: I don't know if this is enough time considering amazon and it is really too fragile
+    // We are blocking up to the connection timeout which should be enough time for a health check
```
codefromthecrypt, line 113: agree. smart idea re-using timeout for this

zipkin-server/src/test/java/zipkin2/server/internal/elasticsearch/ITElasticsearchHealthCheck.java
```diff
@@ -31,90 +15,144 @@
...
+  @Test public void healthCheckDisabled() throws Exception {
+    AnnotationConfigApplicationContext context = new AnnotationConfigApplicationContext();
+
     TestPropertyValues.of(
       "spring.config.name=zipkin-server",
       "zipkin.storage.type:elasticsearch",
-      "zipkin.storage.elasticsearch.hosts:127.0.0.1:1234,127.0.0.1:5678")
+      "zipkin.storage.elasticsearch.timeout:200",
+      "zipkin.storage.elasticsearch.health-check.enabled:false",
+      "zipkin.storage.elasticsearch.health-check.interval:100ms",
```
codefromthecrypt, line 139: handy

codefromthecrypt (approved):
I don't have any change requests on this. if you are keen, click merge!

PR #2723, opened 2019-08-01
Allows throttling to re-use 3rd party exceptions as opposed to wrapping

### LINKED ISSUE #2721: Consider not using RejectedExecutionException
Right now, we are in a position where we use `RejectedExecutionException` as a signal to back off. This was simple in elasticsearch as we originated all exceptions directly.  When we are using another library, many have exception types that are indicative of a rejection, either client side (some health check went red) or server side (a push back based on a pool or queue that's too busy).

Not everyone knows the use cases of exceptions, for example, it might be easy to miss that an exception's direct message results in 500 errors and our own health checks. Wrapping can obfuscate that if we meddle with the original message, rendering it difficult to help third parties who come to us for support. Moreover, it may not be obvious that when a run queue is full it affects also the read side. This means browser requests and the messages we see from chrome debug still need to be useful.

When we consider common text of messages about a rejection, we can neither guess a format, nor a remedy. For example, if a health check is temporarily down, it isn't necessarily the case that enabling throttling will resolve it. However, to the degree it buffers and the problem is temporary, it can prevent more spans from dropping than necessary.

OTOH, throttling isn't a universal answer either. While elastic is famous for falling over during a flurry of requests, many sites are using kafka which doesn't necessarily imply a flurry of requests. Moreover drivers like cassandra have internal queueing mechanisms. This doesn't mean our throttling feature is harmful, just that it isn't a universal answer, and certainly would be disingenuous to wildcard suggest in in error messages instead of saying what exactly happened.

People who are literally viewing zipkin's logs, and are sensitive to seeing something twice might take note and feel the need to somehow change the message, but I personally think that is working against ourselves. For example, the people who have to answer support questions may not even have access to the stack trace which might if you saw it be one layer bigger than before. By preventing us from seeing the actual message we can't help. Let's not do that.

If there is a fundamental feeling about how a wrapped message must be different, then it cannot hurt our team in order to satisfy the itch. My suggestion then would be to change the exception mechanism still allow `RejectedExecutionException` when originating messages, but rather than waste time arguing formats of wrapped messages possibly hurting ourselves in the process..

change to a registry. For example, the storage throttling component could source a`List<RejectionClassifier>` injection supplied by each storage configuration module giving a bean of type `RejectionClassifier`. This has one disadvantage in so far that the spring config lives farther away than the library doing the storage. However, it could be a good approach to handling the classification concern of retryable exceptions, and in a way that doesn't damage the volunteers ability to support the server.

codefromthecrypt:
While re-wrapping exceptions can still make sense (ex to trim huge stack
traces), re-wrapping only to fit `RejectedExecutionException` can be
counter-productive and result in debates about naming that don't need to
be debated.

This adds `StorageComponent.isOvercapacity(Throwable)` to make
classification more flexible.

Fixes #2721

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
@@ -21,6 +21,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.function.Predicate;
 import zipkin2.Call;
 import zipkin2.Callback;
 
@@ -39,16 +40,20 @@ import zipkin2.Callback;
 final class ThrottledCall<V> extends Call.Base<V> {
   final ExecutorService executor;
   final Limiter<Void> limiter;
+  final Predicate<Throwable> isOverCapacity;
   final Call<V> delegate;
 
-  ThrottledCall(ExecutorService executor, Limiter<Void> limiter, Call<V> delegate) {
+  ThrottledCall(ExecutorService executor, Limiter<Void> limiter,
+    Predicate<Throwable> isOverCapacity, Call<V> delegate) {
     this.executor = executor;
     this.limiter = limiter;
+    this.isOverCapacity = isOverCapacity;
     this.delegate = delegate;
   }
 
   @Override protected V doExecute() throws IOException {
-    Listener limitListener = limiter.acquire(null).orElseThrow(RejectedExecutionException::new);
+    Listener limitListener = limiter.acquire(null)
+      .orElseThrow(RejectedExecutionException::new); // TODO: make an exception message
 
     try {
       // Make sure we throttle
@@ -66,7 +71,7 @@ final class ThrottledCall<V> extends Call.Base<V> {
       return result;
     } catch (ExecutionException e) {
       Throwable cause = e.getCause();
-      if (cause instanceof RejectedExecutionException) {
+      if (isOverCapacity.test(cause)) {
         // Storage rejected us, throttle back
         limitListener.onDropped();
       } else {
@@ -93,10 +98,11 @@ final class ThrottledCall<V> extends Call.Base<V> {
   }
 
   @Override protected void doEnqueue(Callback<V> callback) {
-    Listener limitListener = limiter.acquire(null).orElseThrow(RejectedExecutionException::new);
+    Listener limitListener = limiter.acquire(null)
+      .orElseThrow(RejectedExecutionException::new); // TODO: make an exception message
 
     try {
-      executor.execute(new QueuedCall<>(delegate, callback, limitListener));
+      executor.execute(new QueuedCall<>(this, callback, limitListener));
     } catch (RuntimeException | Error e) {
       propagateIfFatal(e);
       // Ignoring in all cases here because storage itself isn't saying we need to throttle.  Though, we may still be
@@ -107,7 +113,7 @@ final class ThrottledCall<V> extends Call.Base<V> {
   }
 
   @Override public Call<V> clone() {
-    return new ThrottledCall<>(executor, limiter, delegate.clone());
+    return new ThrottledCall<>(executor, limiter, isOverCapacity, delegate.clone());
   }
 
   @Override public String toString() {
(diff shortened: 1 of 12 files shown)


zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/internal/call/ResultSetFutureCall.java
```diff
@@ -79,22 +78,24 @@ protected final boolean doIsCanceled() {
...
-      try { // emulate ResultSetFuture.getUninterruptibly
-        return Uninterruptibles.getUninterruptibly(future);
-      } catch (ExecutionException e) {
-        Throwable cause = e.getCause();
-        if (cause instanceof Error) throw ((Error) cause);
-        if (cause instanceof DriverException) throw ((DriverException) cause).copy();
-        throw new DriverInternalError("Unexpected exception thrown", cause);
-      }
-    } catch (QueryConsistencyException | BusyConnectionException | BusyPoolException e) {
-      // Our throttling function relies on rejected execution when encountering load-specific errors
```
codefromthecrypt, line 96: Unlike in the elasticsearch code, this wrapping added no value.

zipkin/src/main/java/zipkin2/storage/StorageComponent.java
```diff
@@ -68,6 +70,33 @@ public ServiceAndSpanNames serviceAndSpanNames() { // delegates to deprecated me
 
   public abstract SpanConsumer spanConsumer();
 
+  /**
```
codefromthecrypt, line 73: Here's the actual text to read

codefromthecrypt:
as this is public api affecting, will let this sit a bit

zipkin/src/main/java/zipkin2/storage/StorageComponent.java
```diff
@@ -68,6 +70,33 @@ public ServiceAndSpanNames serviceAndSpanNames() { // delegates to deprecated me
...
+   * Implementations of throttling can use this signal to differentiate between failures, for
+   * example to reduce traffic.
+   *
+   * <p>Callers of this method will submit an exception raised by {@link Call#execute()} or on the
+   * error callback of {@link Call#enqueue(Callback)}.
+   *
+   * <p>By default, this returns true if the input is a {@link RejectedExecutionException}. When
+   * originating exceptions, use this type to indicate a load related failure.
+   *
+   * <p>Given a 3rd party exception, re-wrapping to force {@link RejectedExecutionException} seems
```
anuraaga: I was a bit confused what automatic handling meant here since this code itself is a bit far from the actual handling. Maybe can be more opinionated:

```java
/**
  * It is generally preferred to specialize this method to handle relevant exceptions for the particular
  * storage rather than wrapping them in {@link RejectedExecutionException} at call sites. Extra 
  * wrapping can make errors harder to read, for example, by making it harder to "google" a solution 
  * for a well known error message for the storage client, instead thinking the error is in Zipkin code 
  * itself.
  */
```

codefromthecrypt: agree this is is better text!

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
```diff
@@ -39,16 +40,20 @@
...
     this.executor = executor;
     this.limiter = limiter;
+    this.isOverCapacity = isOverCapacity;
     this.delegate = delegate;
   }
 
   @Override protected V doExecute() throws IOException {
-    Listener limitListener = limiter.acquire(null).orElseThrow(RejectedExecutionException::new);
+    Listener limitListener = limiter.acquire(null)
+      .orElseThrow(RejectedExecutionException::new); // TODO: make an exception message
```
Logic-32, line 56: FWIW, this is more of a safety check than anything.  The Limiter I implemented should never cause this exception to be thrown.

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledStorageComponent.java
```diff
@@ -102,18 +103,20 @@ public ThrottledStorageComponent(StorageComponent delegate, MeterRegistry regist
...
+    final SpanConsumer delegate;
 
-    ThrottledSpanConsumer(SpanConsumer delegate, Limiter<Void> limiter, ExecutorService executor) {
-      this.delegate = delegate;
-      this.limiter = limiter;
-      this.executor = executor;
+    ThrottledSpanConsumer(ThrottledStorageComponent throttledStorage) {
+      this.executor = throttledStorage.executor;
+      this.limiter = throttledStorage.limiter;
+      this.isOverCapacity = throttledStorage::isOverCapacity;
```
Logic-32: This evaluates to creating a new Predicate.  Is that desirable given that it would just result in the `isOverCapacity` check simply having an ever increasing stack depth?

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
codefromthecrypt: if you recall the reason why it is unreachable mention. otherwise no biggie thx

Logic-32: https://github.com/openzipkin/zipkin/blob/31338b52714c2dbb989b07beaa58cf7957d12e8c/zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledStorageComponent.java#L188-L193

That is why :)

PR #2724, opened 2019-08-01
Use unique attribute key.

anuraaga:
Sorry for not noticing earlier - this will drift https://github.com/openzipkin/zipkin-aws/pull/135

An `AttributeKey` needs to be globally unique or it could possibly get clobbered.

While I could have made the `String` of the name longer and unique enough, I figured exposing the `AttributeKey` itself with documentation would be OK too but can tweak it.

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
@@ -28,7 +28,6 @@ import com.linecorp.armeria.common.util.Exceptions;
 import com.linecorp.armeria.common.util.SafeCloseable;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufHolder;
-import io.netty.util.AttributeKey;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.concurrent.EventExecutor;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
hmm weird it worked before

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
```diff
@@ -39,12 +38,12 @@
 import java.util.concurrent.RejectedExecutionException;
 import zipkin2.Call;
 import zipkin2.Callback;
+import zipkin2.elasticsearch.ElasticsearchStorage;
 
 import static zipkin2.elasticsearch.internal.JsonReaders.enterPath;
 import static zipkin2.elasticsearch.internal.JsonSerializers.JSON_FACTORY;
 
 public final class HttpCall<V> extends Call.Base<V> {
-  public static final AttributeKey<String> NAME = AttributeKey.valueOf("name");
```
codefromthecrypt, line 47: are you planning to use the word 'name' in armeria? right now, if we change this here we also need to change in the autoconfig that uses it. This seems unnecessary as we've no current problem with this working.

anuraaga: Hmm - I guess in Armeria we'd always use a fully qualified key so it's probably ok if users want to stick to non-qualified ones.

codefromthecrypt: cool.. I feel this is a temporary part of code anyway. Mainly I don't want to have to manually regression check everything again, so definitely happy if we can keep this for now

PR #2725, opened 2019-08-02
Adds ForwardingStorageComponent to end little bugs

codefromthecrypt:
We've had numerous places where people wrap and forget to forward
`.check()` or aren't aware when they forgot to forward a new method.

This removes that pattern of maintenance for a forwarding decorator.


codefromthecrypt:
cc @anuraaga @Logic-32 @meltsufin I think maybe you've seen some of this in the past.

PR #2726, opened 2019-08-02
Redesign Trace page & TraceViewer page

### LINKED ISSUE #2673: Add traceId search box to lens individual trace page
I really like the addition of the search by traceId and upload buttons. Directly navigating to a trace via a traceId is a workflow that I use a lot.
image

One thing that I do miss is being able to switch to another trace once I've already loaded a trace. Can we include the above options on the individual trace page?

tacigar:
This pr changes several things.

* Integrate TraceViewer page component and Trace page component.
* Display root span information on the top left of the trace page & traceViewer page.
* Improve Error message design.
* fixes #2673 

## Integrate TraceViewer page component and Trace page component

Trace page components and TraceViewer page components were previously separate components.
But these two components have many things in common, so I integrated them.

## Display root span information on the top left of the trace page & traceViewer page.

Previously, Trace ID was displayed at the top left of the page.
But root span information more important to humans than Trace ID.

#### Before

top-left-before

#### After

top-left-after

#### Before

error-message-before

#### After

error-message-after


codefromthecrypt:
I like this so much better.

I've never been a fan of taking screen real-estate to show the trace ID which is already in the window url, but I know some have asked for it in the past. Is the trace ID displayed on screen anywhere else?

tacigar:
No, it is not displayed now.
I will once again consider the header design of the Trace page.
Please wait a little.

zeagord (approved):

tacigar:
PTAL 🙇

codefromthecrypt (approved):
I agree completely with the design. the trace ID should be visible but not the face of the trace. This is excellent

PR #2728, opened 2019-08-02
Changes elasticsearch parsing to only lazy allocate strings

### LINKED ISSUE #2717: reduce api exposure to HttpData
Right now, `HttpData` is a dangerous api as it requires reference counting. Calling release twice will screw up everything. We should not expose this anywhere we can avoid it.

@anuraaga mentioned that we only really use this in two cases, parsing json and when there is an unknown error getting a full string of the body. By exposing a different api, we can reduce fragility.

Ex
```java
V convert(JsonParser parser, Supplier<String> contentString);
```

These would be provided underneath by something like this which ensures reference counting happens correctly.
```java
@Override default V convert(HttpData content) throws IOException {
  try (InputStream stream = content.toInputStream()) {
    return convert(JSON_FACTORY.createParser(stream), content::toStringUtf8);
  } finally {
    ReferenceCountUtil.safeRelease(content);
  }
}
```

See https://github.com/line/armeria/pull/1938

codefromthecrypt:
This also hides where reference counting is used. Thanks @anuraaga for
the advice. cc @trustin

Fixes #2717

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
@@ -19,7 +19,6 @@ import com.google.auto.value.extension.memoized.Memoized;
 import com.linecorp.armeria.client.HttpClient;
 import com.linecorp.armeria.client.endpoint.EndpointGroup;
 import com.linecorp.armeria.common.AggregatedHttpRequest;
-import com.linecorp.armeria.common.HttpData;
 import com.linecorp.armeria.common.HttpMethod;
 import java.io.Closeable;
 import java.io.IOException;
@@ -47,7 +46,6 @@ import static zipkin2.elasticsearch.ElasticsearchSpanStore.DEPENDENCY;
 import static zipkin2.elasticsearch.ElasticsearchSpanStore.SPAN;
 import static zipkin2.elasticsearch.EnsureIndexTemplate.ensureIndexTemplate;
 import static zipkin2.elasticsearch.internal.JsonReaders.enterPath;
-import static zipkin2.elasticsearch.internal.JsonSerializers.JSON_FACTORY;
 
 @AutoValue
 public abstract class ElasticsearchStorage extends zipkin2.storage.StorageComponent {
@@ -59,8 +57,8 @@ public abstract class ElasticsearchStorage extends zipkin2.storage.StorageCompon
    */
   public interface LazyHttpClient extends Supplier<HttpClient>, Closeable {
     /**
-     * Lazily creates an instance of the http client configured to the correct elasticsearch host
-     * or cluster. The same value should always be returned.
+     * Lazily creates an instance of the http client configured to the correct elasticsearch host or
+     * cluster. The same value should always be returned.
      */
     @Override HttpClient get();
 
@@ -296,7 +294,7 @@ public abstract class ElasticsearchStorage extends zipkin2.storage.StorageCompon
     return new HttpCall.Factory(lazyHttpClient().get());
   }
 
-  @Override public void close() throws IOException {
+  @Override public void close() {
     lazyHttpClient().close();
   }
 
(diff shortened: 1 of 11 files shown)


zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/BulkCallBuilder.java
```diff
@@ -30,29 +31,29 @@
...
+    @Override public Void convert(JsonParser parser, Supplier<String> contentString) {
       RuntimeException toThrow = null;
       try {
-        JsonNode tree = OBJECT_MAPPER.readTree(content);
+        JsonNode tree = OBJECT_MAPPER.readTree(parser);
         Number status = tree.findPath("status").numberValue();
         if (status != null && status.intValue() == 429) {
           toThrow = new RejectedExecutionException(tree.toString());
         } else if (tree.path("/errors").booleanValue()) {
-          toThrow = new RuntimeException(content.toString());
```
codefromthecrypt, line 55: actually before this would have been a bad message :P

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -296,21 +294,21 @@ String buildUrl(IndexTemplates templates, String type) {
...
-      // The health check is only invoked periodically. Hence, there is less impact to allocating
-      // strings. We retain the string so that it can be logged if the ES response is malformed.
-      String result = body.toStringUtf8();
-      JsonParser status = enterPath(JSON_FACTORY.createParser(result), "status");
+    @Override public CheckResult convert(JsonParser parser, Supplier<String> contentString)
+      throws IOException {
+      JsonParser status = enterPath(parser, "status");
       if (status == null) {
-        throw new IllegalArgumentException("Health status couldn't be read " + result);
+        // The health check is only invoked periodically. Hence, there is less impact to allocating
```
anuraaga: Don't think you need this comment anymore.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/VersionSpecificTemplates.java
```diff
@@ -225,10 +226,10 @@ static float getVersion(HttpCall.Factory callFactory) throws IOException {
   enum ReadVersionNumber implements HttpCall.BodyConverter<Float> {
     INSTANCE;
 
-    @Override public Float convert(HttpData content) {
+    @Override public Float convert(JsonParser parser, Supplier<String> contentString) {
       // The version number is read only once per startup. Hence, there is less impact to allocating
       // strings. We retain the string so that it can be logged if the ES response is malformed.
-      String body = content.toStringUtf8();
+      String body = contentString.get();
```
anuraaga: No big deal but think readTree accepts JsonParser too.

anuraaga (approved):

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
codefromthecrypt: agreed

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/VersionSpecificTemplates.java
codefromthecrypt: it broke the api chain but I'll try to figure out why

codefromthecrypt: annoyingly :)

Screenshot 2019-08-02 at 12 28 05 PM

codefromthecrypt: I avoided that api..

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
```diff
@@ -194,9 +195,15 @@ public Factory(HttpClient httpClient) {
...
       bodyConverter = (parser, contentString) -> {
-        parser = enterPath(parser, "message");
-        throw new RuntimeException(parser != null
-          ? parser.getValueAsString()
+        String message = null;
+        try {
+          JsonNode root = OBJECT_MAPPER.readTree(parser);
+          message = root.findPath("reason").textValue();
+          if (message == null) message = root.at("/message").textValue();
+          if (message == null) message = root.at("/Message").textValue();
```
anuraaga, line 203: Nice detective work

anuraaga (approved):

PR #2729, opened 2019-08-02
Unhooks protocols that are usually synchronous from storage commands

### LINKED ISSUE #2715: release post request sooner
The below trace you can see we could release the request sooner on worst case first request. While this is a first request problem, if there are storage errors, delay could be routine until they resolve.

* implicitly lookup of #elasticsearch endpoint from #aws
* lazy index template where all templates weren't there
* finally doing the post

Screenshot 2019-07-31 at 11 57 37 AM

Looking at Collector.java there's a little bit of lying when we have a first request. It is true that we are swapping callbacks. However, the method `store` blocks when lazy instantiating. We could probably add an executor to Collector that can be used to avoid this blocking, so that the function can return sooner, releasing the http post from the client.

```
    // In order to ensure callers are not blocked, we swap callbacks when we get to the storage
    // phase of this process. Here, we create a callback whose sole purpose is classifying later
    // errors on this bundle of spans in the same log category. This allows people to only turn on
    // debug logging in one place.
    Callback<Void> logOnErrorCallback = storeSpansCallback(sampledSpans);
    try {
      store(sampledSpans, logOnErrorCallback);
      callback.onSuccess(null); // release the callback given to the collector
    } catch (RuntimeException | Error e) {
      // While unexpected, invoking the storage command could raise an error synchronously. When
      // that's the case, we wouldn't have invoked callback.onSuccess, so we need to handle the
      // error here.
      handleStorageError(spans, e, callback);
    }
```

codefromthecrypt:
http and gRPC protocols usually hold clients. We want to hold them until
messages are parsed, but not later. This employs an executor to ensure
any blocking commands needed to get the storage command enqueued do not
block the request.

This also fixes some lack of logging where nio is used.

Fixes #2715

### DIFF
--- a/zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
+++ b/zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
@@ -132,8 +132,13 @@ public class Collector { // not final for mock
     // phase of this process. Here, we create a callback whose sole purpose is classifying later
     // errors on this bundle of spans in the same log category. This allows people to only turn on
     // debug logging in one place.
-    executor.execute(new StoreSpans(sampledSpans));
-    callback.onSuccess(null);
+    try {
+      executor.execute(new StoreSpans(sampledSpans));
+      callback.onSuccess(null);
+    } catch (Throwable unexpected) { // ensure if a future is supplied we always set value or error
+      callback.onError(unexpected);
+      throw unexpected;
+    }
   }
 
   /** Like {@link #acceptSpans(byte[], BytesDecoder, Callback)}, except using a byte buffer. */
--- a/zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
+++ b/zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
@@ -13,16 +13,19 @@
  */
 package zipkin2.collector;
 
+import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.concurrent.Executor;
 import java.util.function.Supplier;
 import java.util.logging.Logger;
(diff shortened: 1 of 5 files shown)


zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
```diff
@@ -195,19 +213,34 @@ String idString(Span span) {
...
+        onError(e);
       }
-    };
+    }
+
+    @Override public void onSuccess(Void value) {
+    }
+
+    @Override public void onError(Throwable t) {
+      handleStorageError(spans, t, NOOP_CALLBACK);
```
codefromthecrypt, line 243: the intent here is to only log/increment metrics on error, and that's why we use noop-callback as opposed to passing in the callback from the transport. Otherwise, we would hold up the request.

zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
```diff
@@ -103,6 +106,16 @@ public Collector build() {
   }
 
   public void accept(List<Span> spans, Callback<Void> callback) {
+    accept(spans, callback, Runnable::run);
+  }
+
+  /**
+   * @param executor Returns the executor used to enqueue the storage request.
```
anuraaga: Remove `Returns `

anuraaga (approved):

zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
codefromthecrypt: 👍

codefromthecrypt:
tried with amazon and it looks like a win

slow one

Screenshot 2019-08-02 at 6 47 24 PM

fast one

Screenshot 2019-08-02 at 6 49 09 PM

Issue #2701, opened 2019-07-22
Topology view of query results


codefromthecrypt:
This is about an alternate query screen that is topology based as opposed to table based, but still using the query api and universal search paramters from the existing search screen. Other projects like pinpoint and skywalking seem to use topology instead of query interface.

The result of a query would be an active image which could help people isolate the trace they want.

<img width="125" alt="Screenshot-2019-07-18-at-11 53 30-AM" src="https://user-images.githubusercontent.com/64215/61605608-4af77200-ac81-11e9-984c-1dc6c64fa33e.png">

image from https://pdfs.semanticscholar.org/94c8/17e196e71c03b3425f905ebd1793dc6469c2.pdf

idea is if the data under the trace list screen could be repurposed for a scoped aggregation
where the circles could help group nodes by a selected site-tag and the fibres represent traces, which you click to go to the trace view.

All this in javascript would require first porting the DependencyLinker code to javascript then doing normal dependency graph (to prove data is ok). then fancy things possibly like above possibly not based on the assumption we can aggregate any way we want if done client side with small data.

This is related to https://github.com/openzipkin/zipkin/issues/2622 as both parts need a javascript dependency linker to produce the view.

https://nivo.rocks/network/ could be used for the presentation possibly

tacigar:
I wonder if all users are using zipkin-dependencies spark jobs...

codefromthecrypt:
for this I mean aggregate the traces returned by /api/v2/traces in
javascript, via a port of DependencyLinker. In this way spark doesn't
matter.

tacigar:
Ah, I see!

codefromthecrypt:
https://github.com/openzipkin/zipkin/pull/2731 is a first step as it ports the basic dependency linker used in spark jobs to javascript

Issue #2622, opened 2019-06-03
Service Topology view for a single trace


connectwithnara:
The waterfall diagram in Zipkin UI is useful to understand the latency distribution for a trace. However the view is not very friendly to perceive the services involved in that trace. It will be useful to have another tab 'Topology' and render the topology of the services involved in the trace.

At Netflix we trace 100% of the requests that is enabled for failure injection. The users will find it useful to understand what services were involved in a failure injected request.

codefromthecrypt:
Thanks for raising this, Nara. I'm positive it has been mentioned at least for a year (including by me!), but never formally in an issue.

So the idea is to take the trace we already have (in fact it is already assembled into a tree internally), then run a linker to aggregate the service dependencies. At that point, it is only changing the data sent to the dependencies page, possibly with labeling of the trace ID, similar to our "show trace" functionality.

I'd be happy to port the dependency linker logic to javascript. @tacigar @zeagord could one of you take a stab at UI/UX on this?

cc also @bulicekj as you may have similar requests in haystack.

codefromthecrypt:
naver pinpoint also has this feature.

codefromthecrypt:
Some notes from discussion with @tacigar 

I have service graph now, but I cannot tell if there is a problem with one node behavior
if I want to understand behavior of a single machine in a service, I cannot do this today.
this is because the edges are aggregatate (parent, child, count), based on service name, not any other information in the endpoint such as IP address.

This can make certain problems difficult to understand, such as if one node in a service is running the wrong code, or if a new version of code only deployed to one node is causing a problem.

So, if I have a way to classify by another means, I can identify this type of behavior. For example, IP address. {(parent, ip), (child, ip), count)  the dependency link has one more qualifier than before, for this example IP address.

There are now many more nodes.. because what was before just service-service is now (service,ip) -> (service, ip). Inside one trace, this could be fine because maybe not that many combinations


IP is just an example, user may want to explore by site tag like cluster or department or something else. The difference between this and normal dependency graph is we generate the links in javascript. This means we can aggregate by anything, not just service but also custom tag (like http.route).

RestfulBlue:
i also hope such functionality will appear, 
for example jaeger also has that feature
https://miro.medium.com/max/2625/1*W6OGeCA1unSqQfPIZq7VGg.png

codefromthecrypt:
[...]

Issue #2230, opened 2018-10-31
Visualize large swaths of traces


codefromthecrypt:
From https://twitter.com/soini/status/1057642318365306880
> For zipkin, if I had one wish, it'd be native visualization options for large swaths of traces.

**Rational**
Currently, our only aggregate tool is the service dependency diagram. While we've discussed doing single trace aggregate (for a stable shape regardless of call count to redis for example), We've not discussed how to do other ways of visualizing large amounts of traces.

**Example Scenario**
Hoping to hear about this from Joe so that we make the right thing!

**Prior Art**
Joe is building a multi-level sankey diagram w/ Vega in Kibana in the interim.
* Please note other prior art. *

codefromthecrypt:
also related https://engineering.salesforce.com/anomaly-detection-in-zipkin-trace-data-87c8a2ded8a1

codefromthecrypt:
because we currently need to understand if this is visualizing for the purpose of detecting anomalies or traffic flow or shape of call graphs, we'll want to see more input. Especially in netflix we discussed that people may want something like a service graph with more facts (such as cluster and zone), but we don't know exactly what's in mind without hearing more from Jon or others.

codefromthecrypt:
From Jon on Twitter: https://twitter.com/soini/status/1058078887488315392

> here's one of the viz's that's been kicking around in my head put (quickly) to paper. Being able to aggregate the node<>node flows, and then break them down by terms (error codes) or duration ranges would be cool, I think.

![dq8myiaucaixl7a](https://user-images.githubusercontent.com/64215/47949638-83371300-df81-11e8-981f-e3394f59ebb1.png)

codefromthecrypt:
Chat with @narayuna offline I think is clarifying to me about how to phrase this effort.

So notes here aren't to say sanke is the way, as there are lots of details that need to be elaborated, not only why should something be built, but also who is it for, what affects the feature (related features, data things etc): it isn't a goal to implement an image, our goal is to serve users

Jon is a user who asked about this, and he has some experience to share, so we are welcome to take it. This is awesome as not everyone in Zipkin are experienced in the same ways. In OSS, some are happy to help with work needed to implement things they don't fully understand. We love experience dumps for this reason.. that's also why we hold workshops
https://cwiki.apache.org/confluence/display/ZIPKIN/Workshops

[...]

PR #2731, opened 2019-08-03
ports dependency linker to javascript

codefromthecrypt:
This should allow integrations including:

* single trace aggregates
* aggregates across the results of the current query

Right now, this is a direct and complete port of the Java code. However,
we could do more later once this is integrated. Here are some options:

* include references to each span in the dependency link
  * Ex. trace/span id key so that you can jump directly to the row.
* full path tracing (future)
  * Ex. instead of parent/child, complete path from root to each leaf.

See #2701
See #2622
See #2230


zeagord (approved):
Looks good to me and will be helpful for the topology view

PR #2732, opened 2019-08-04
Slight server cleanups around /health and self-tracing

codefromthecrypt:
Before, we were ignoring the sampling rate during self-tracing by
returning a constant (true) as opposed to deferring to the rate-limited
global sampler. Also, we put variables in the toString for
InMemoryStorage which isn't helpful as spring caches it in the /health
endpoint.

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
@@ -73,20 +73,21 @@ public class TracingConfiguration {
     return ThreadLocalSpan.create(tracing.tracer());
   }
 
-  /** Controls aspects of tracing such as the name that shows up in the UI */
-  @Bean Tracing tracing(Reporter<Span> reporter, SelfTracingProperties config) {
-    final Sampler sampler;
+  @Bean Sampler sampler(SelfTracingProperties config) {
     if (config.getSampleRate() != 1.0) {
       if (config.getSampleRate() < 0.01) {
-        sampler = BoundarySampler.create(config.getSampleRate());
+        return BoundarySampler.create(config.getSampleRate());
       } else {
-        sampler = Sampler.create(config.getSampleRate());
+        return Sampler.create(config.getSampleRate());
       }
     } else if (config.getTracesPerSecond() != 0) {
-      sampler = RateLimitingSampler.create(config.getTracesPerSecond());
-    } else {
-      sampler = Sampler.ALWAYS_SAMPLE;
+      return RateLimitingSampler.create(config.getTracesPerSecond());
     }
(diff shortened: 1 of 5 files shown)


zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
```diff
@@ -73,20 +73,21 @@
     return ThreadLocalSpan.create(tracing.tracer());
   }
 
-  /** Controls aspects of tracing such as the name that shows up in the UI */
-  @Bean Tracing tracing(Reporter<Span> reporter, SelfTracingProperties config) {
-    final Sampler sampler;
+  @Bean Sampler sampler(SelfTracingProperties config) {
```
codefromthecrypt, line 76: exposed this for sanity reasons (I can look at the /actuate/beans endpoint to see which sampler is in use

zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
```diff
@@ -103,7 +104,10 @@
       .serverSampler(new HttpSampler() {
         @Override public <Req> Boolean trySample(HttpAdapter<Req, ?> adapter, Req request) {
           String path = adapter.path(request);
-          return path.startsWith("/api") || path.startsWith("/zipkin/api");
+          if (path.startsWith("/api") || path.startsWith("/zipkin/api")) {
```
codefromthecrypt, line 107: Subtle, but the contract for the HttpSampler is to do your rate here (with a true or false) or return null to use the global rate. We were accidentally returning true which meant the global sampler was entirely ignored.

cc just as a lesson @anuraaga @kojilin @marcingrzejszczak

zipkin-server/src/test/java/zipkin2/server/internal/brave/ITZipkinSelfTracing.java
```diff
@@ -106,7 +106,7 @@ InMemoryStorage inMemoryStorage() {
    * information.
    */
   @Test public void toStringContainsOnlySummaryInformation() {
-    assertThat(storage).hasToString("Traced{InMemoryStorage{traceCount=0}}");
```
codefromthecrypt, line 109: if you look at health endpoint it would always say traceCount=0 :)

codefromthecrypt:
Results make sense now. The whole point of the pending change to use rate-limited sampler was to be able to safely trace post requests. Now, you can see that the local (self-tracing) metrics match expectations. In this case a 100s burst of traces and we end up with 204 spans (102 traces).

Screenshot 2019-08-04 at 1 46 30 PM
Screenshot 2019-08-04 at 1 48 06 PM

PR #2735, opened 2019-08-05
Instruments when STORAGE_THROTTLE_ENABLED=true

codefromthecrypt:
Before, we couldn't see the effects of self-tracing when throttling
was enabled. This change simplifies some of the code in the throttling
feature such that it uses a simpler to instrument `Executor` to invoke
commands. This also makes explicit some tracing we are trying to avoid,
notably anything spawned by the tracer itself.

cc @Logic-32

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
@@ -51,9 +51,16 @@ import zipkin2.storage.StorageComponent;
 @EnableConfigurationProperties(SelfTracingProperties.class)
 @ConditionalOnSelfTracing
 public class TracingConfiguration {
+  static volatile Thread reporterThread;
+
+  public static boolean isSpanReporterThread() {
+    return Thread.currentThread() == reporterThread;
+  }
+
   /** Configuration for how to buffer spans into messages for Zipkin */
   @Bean Reporter<Span> reporter(BeanFactory factory, SelfTracingProperties config) {
     return AsyncReporter.builder(new LocalSender(factory))
+      .threadFactory(r -> reporterThread = new Thread(r))
       .messageTimeout(config.getMessageTimeout().toNanos(), TimeUnit.NANOSECONDS)
       .metrics(new ReporterMetricsAdapter(factory))
       .build();
--- a/zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
@@ -51,16 +51,9 @@ import zipkin2.storage.StorageComponent;
 @EnableConfigurationProperties(SelfTracingProperties.class)
 @ConditionalOnSelfTracing
 public class TracingConfiguration {
-  static volatile Thread reporterThread;
-
-  public static boolean isSpanReporterThread() {
-    return Thread.currentThread() == reporterThread;
-  }
-
   /** Configuration for how to buffer spans into messages for Zipkin */
   @Bean Reporter<Span> reporter(BeanFactory factory, SelfTracingProperties config) {
     return AsyncReporter.builder(new LocalSender(factory))
-      .threadFactory(r -> reporterThread = new Thread(r))
       .messageTimeout(config.getMessageTimeout().toNanos(), TimeUnit.NANOSECONDS)
       .metrics(new ReporterMetricsAdapter(factory))
       .build();
@@ -80,6 +73,14 @@ public class TracingConfiguration {
     return ThreadLocalSpan.create(tracing.tracer());
   }
 
+  /**
+   * This controls the general rate. In order to not accidentally start traces started from the
+   * tracer itself, this isn't used as {@link Tracing.Builder#sampler(Sampler)}. The impact of this
+   * is that we can't currently start traces from Kafka or Rabbit (until we use a messaging
+   * sampler).
+   *
+   * See https://github.com/openzipkin/brave/pull/914 for the messaging abstraction
+   */
   @Bean Sampler sampler(SelfTracingProperties config) {
     if (config.getSampleRate() != 1.0) {
       if (config.getSampleRate() < 0.01) {
@@ -94,10 +95,10 @@ public class TracingConfiguration {
   }
 
   /** Controls aspects of tracing such as the name that shows up in the UI */
-  @Bean Tracing tracing(Reporter<Span> reporter, Sampler sampler) {
+  @Bean Tracing tracing(Reporter<Span> reporter) {
     return Tracing.newBuilder()
       .localServiceName("zipkin-server")
-      .sampler(sampler)
+      .sampler(Sampler.NEVER_SAMPLE) // don't sample traces at this abstraction
       .currentTraceContext(currentTraceContext())
       // Reduce the impact on untraced downstream http services such as Elasticsearch
       .propagationFactory(B3SinglePropagation.FACTORY)
@@ -105,20 +106,18 @@ public class TracingConfiguration {
       .build();
   }
 
-  @Bean HttpTracing httpTracing(Tracing tracing) {
+  @Bean HttpTracing httpTracing(Tracing tracing, Sampler sampler) {
     return HttpTracing.newBuilder(tracing)
       // server starts traces for read requests under the path /api
       .serverSampler(new HttpSampler() {
         @Override public <Req> Boolean trySample(HttpAdapter<Req, ?> adapter, Req request) {
           String path = adapter.path(request);
           if (path.startsWith("/api") || path.startsWith("/zipkin/api")) {
-            return null; // use the global rate limit
+            return sampler.isSampled(0L); // use the global rate limit
           }
           return false;
         }
       })
-      // client doesn't start new traces
-      .clientSampler(HttpSampler.NEVER_SAMPLE)
       .build();
   }
 
--- a/zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingStorageComponent.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingStorageComponent.java
@@ -29,8 +29,6 @@ import zipkin2.storage.SpanConsumer;
 import zipkin2.storage.SpanStore;
 import zipkin2.storage.StorageComponent;
 
-import static zipkin2.server.internal.brave.TracingConfiguration.isSpanReporterThread;
-
 // public for use in ZipkinServerConfiguration
 public final class TracingStorageComponent extends ForwardingStorageComponent {
   final Tracing tracing;
@@ -167,8 +165,6 @@ public final class TracingStorageComponent extends ForwardingStorageComponent {
     }
 
     @Override public Call<Void> accept(List<Span> spans) {
-      Call<Void> call = delegate.accept(spans);
-      if (isSpanReporterThread()) return call;
       return new TracedCall<>(tracer, delegate.accept(spans), "accept-spans");
     }
 
--- a/zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingStorageComponent.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingStorageComponent.java
@@ -29,6 +29,8 @@ import zipkin2.storage.SpanConsumer;
 import zipkin2.storage.SpanStore;
 import zipkin2.storage.StorageComponent;
 
+import static zipkin2.server.internal.brave.TracingConfiguration.isSpanReporterThread;
+
 // public for use in ZipkinServerConfiguration
 public final class TracingStorageComponent extends ForwardingStorageComponent {
   final Tracing tracing;
@@ -165,6 +167,8 @@ public final class TracingStorageComponent extends ForwardingStorageComponent {
     }
 
     @Override public Call<Void> accept(List<Span> spans) {
+      Call<Void> call = delegate.accept(spans);
+      if (isSpanReporterThread()) return call;
       return new TracedCall<>(tracer, delegate.accept(spans), "accept-spans");
     }
 
--- a/zipkin-server/src/main/java/zipkin2/server/internal/throttle/LimiterMetrics.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/throttle/LimiterMetrics.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2015-2019 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin2.server.internal.throttle;
+
+import com.netflix.concurrency.limits.Limiter.Listener;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import zipkin2.collector.CollectorMetrics;
+
+/** Follows the same naming convention as {@link CollectorMetrics} */
+final class LimiterMetrics {
+  final Counter requests, requestsSucceeded, requestsIgnored, requestsDropped;
+
+  final Listener wrap(Listener delegate) {
+    return new Listener() {
+      @Override public void onSuccess() {
+        // usually we don't add metrics like this,
+        // but for now it is helpful to sanity check acquired vs erred.
+        requestsSucceeded.increment();
+        delegate.onSuccess();
+      }
+
+      @Override public void onIgnore() {
+        requestsIgnored.increment();
+        delegate.onIgnore();
+      }
+
+      @Override public void onDropped() {
+        requestsDropped.increment();
+        delegate.onDropped();
+      }
+    };
+  }
+
+  LimiterMetrics(MeterRegistry registry) {
+    requests = Counter.builder("zipkin_storage.throttle.requests")
+      .description("cumulative amount of limiter requests acquired")
+      .register(registry);
+    requestsSucceeded = Counter.builder("zipkin_storage.throttle.requests_succeeded")
+      .description("cumulative amount of limiter requests acquired that later succeeded")
+      .register(registry);
+    requestsDropped =
+      Counter.builder("zipkin_storage.throttle.requests_dropped")
+        .description(
+          "cumulative amount of limiter requests acquired that later dropped due to capacity")
+        .register(registry);
+    requestsIgnored =
+      Counter.builder("zipkin_storage.throttle.requests_ignored")
+        .description(
+          "cumulative amount of limiter requests acquired that later dropped not due to capacity")
+        .register(registry);
+  }
+}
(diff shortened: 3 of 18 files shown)


codefromthecrypt:
PS the reason I want this is to be able to run full self-tracing under load. This will help identify what exactly caused errors that drop spans due to rejection. This is really hard to figure out without tracing.

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
```diff
@@ -37,183 +38,181 @@
...
-    return new ThrottledCall<>(executor, limiter, isOverCapacity, delegate.clone());
+  @Override public Call<Void> clone() {
+    return new ThrottledCall(delegate.clone(), executor, limiter, isOverCapacity);
   }
 
   @Override public String toString() {
     return "Throttled(" + delegate + ")";
   }
 
-  static String setCurrentThreadName(String name) {
```
codefromthecrypt, line 123: I removed this (eventhough I asked for it) because actually it didn't help at all and so was a bit distracting, and also untested. I want to focus lines on tested code that helps troubleshooting

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledStorageComponent.java
```diff
@@ -102,21 +109,47 @@ public ThrottledStorageComponent(StorageComponent delegate, MeterRegistry regist
     return "Throttled{" + delegate.toString() + "}";
   }
 
+  /**
+   * Lazy accesses the request context as we cannot scope the executor used by storage commands per
+   * request.
+   */
+  static final class RequestContextInstrumentedExecutor implements Executor {
```
codefromthecrypt, line 118: @trustin @kojilin this is what I was asking about in slack.. Basically, I'd like to be able to say `RequestContext.contextAwareExecutor()` (static not instance!)

codefromthecrypt:
here's a trace where you can see blocking at play:

Screenshot 2019-08-05 at 12 25 32 PM

zoomed to the place where you can see the issue
Screenshot 2019-08-05 at 12 25 48 PM

Note @shakuzen @tacigar this is a good example of a trace that has very few spans, but still needs zoom. I take back <10 being good for removing mini-map, if it also means removing zoom. That said, if I could re-root the throttled span, it would be fine in this case!

codefromthecrypt:
very interesting that the self-tracing of the "accept-spans" command takes so long to complete... doesn't quite make sense to me at the moment, but it is a lot later than the storage call..

codefromthecrypt:
the long time for "accept-spans" to return is a different topic I think, possibly it is because of reading the body. depends on how the client trace is instrumented. Ex this is without throttling..

Screenshot 2019-08-05 at 12 36 42 PM

codefromthecrypt:
looks like here blocking could be an intended side-effect of the limiter listener. adding a commit about it, and also adding metrics because it is currently invisible when any capacity errors occur.

codefromthecrypt:
ahh much better

```bash
$ curl -sSL localhost:9411/prometheus|grep throttle|grep 0
zipkin_storage_throttle_requests_succeeded_total 16224.0
zipkin_storage_throttle_requests_dropped_total 0.0
zipkin_storage_throttle_requests_ignored_total 0.0
zipkin_storage_throttle_in_flight_requests 4.0
zipkin_storage_throttle_queue_size 0.0
zipkin_storage_throttle_requests_total 16228.0
zipkin_storage_throttle_concurrency 10.0
```

codefromthecrypt:
and finally have a problem to investigate!

Screenshot 2019-08-05 at 2 19 45 PM


but metrics show no drops
```
curl -sSL localhost:9411/prometheus|grep throttle|grep 0
zipkin_storage_throttle_in_flight_requests 0.0
zipkin_storage_throttle_requests_dropped_total 0.0
zipkin_storage_throttle_requests_ignored_total 3825.0
zipkin_storage_throttle_concurrency 10.0
zipkin_storage_throttle_requests_total 37026.0
zipkin_storage_throttle_requests_succeeded_total 33201.0
zipkin_storage_throttle_queue_size 0.0
```

which tells me there's a classification problem somewhere..

codefromthecrypt:
I added a custom error parser to figure out what the exception was about..

```
+      .errorParser(new ErrorParser() {
+                     @Override protected void error(Throwable error, Object span) {
+                       String className = error.getClass().getSimpleName();
+                       String message = error.getMessage();
+                       if (message == null) {
+                         message = className + "(" + message + ")";
+                       } else {
+                         message = className;
+                       }
+                       StringWriter sw = new StringWriter();
+                       PrintWriter pw = new PrintWriter(sw);
+                       error.printStackTrace(pw);
+                       tag(span, "error", message);
+                       tag(span, "error.stack", sw.toString());
+                     }
+                   }
+      )
```

Looks like were hitting an exception at the storage layer, just not a remote one. I'll patch the code to also consider load errors one enqueue:

Screenshot 2019-08-05 at 2 54 25 PM

codefromthecrypt:
nevermind, there's an existing note about this.. while the name armeria is there, the actual thread pool was the one from the throttle component itself.

codefromthecrypt:
I think the best way to ensure this is working is to change elasticsearch to have a much shorter run queue

codefromthecrypt:
putting this down a bit, but feeling positive!

zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
```diff
@@ -51,9 +51,16 @@
 @EnableConfigurationProperties(SelfTracingProperties.class)
 @ConditionalOnSelfTracing
 public class TracingConfiguration {
+  static volatile Thread reporterThread;
```
anuraaga: While bean's are generally singleton, it seems highly unusual to save into a static variable from a bean constructor.

I think I'd at least go ahead and make another bean holder, something like

```java
static class ReporterThreadFactory implements ThreadFactory {
  volatile Thread reporterThread;

  public Thread newThread(Runnable r) {
     if (reporterThread != null) throw?
     return reporterThread = new Thread(r);
  }

  public boolean isSpanReporterThread() {
    return Thread.currentThread() == reporterThread;
  }
}
```

Then areas that want to know if it's a span reporter thread can inject this bean and call the method.

zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingStorageComponent.java
```diff
@@ -165,6 +167,8 @@ public TracingStorageComponent(Tracing tracing, StorageComponent delegate) {
     }
 
     @Override public Call<Void> accept(List<Span> spans) {
+      Call<Void> call = delegate.accept(spans);
+      if (isSpanReporterThread()) return call;
```
anuraaga: I'm probably wrong but I thought since we don't sample client spans, calls from the span reporter thread wouldn't be sampled anyways, they originate from the flush thread not server requests. Why is this check necessary?

zipkin-server/src/main/java/zipkin2/server/internal/throttle/LimiterMetrics.java
```diff
@@ -0,0 +1,64 @@
...
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import zipkin2.collector.CollectorMetrics;
+
+/** Follows the same naming convention as {@link CollectorMetrics} */
+final class LimiterMetrics {
+  final Counter requests, requestsSucceeded, requestsIgnored, requestsDropped;
+
+  final Listener wrap(Listener delegate) {
+    return new Listener() {
```
anuraaga, line 26: Maybe want to override toString

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
```diff
@@ -37,183 +38,187 @@
...
-    QueuedCall(ThrottledCall<V> throttledCall, Callback<V> callback, Listener limitListener) {
+    EnqueueAndAwait(ThrottledCall throttledCall, LimiterReleasingCallback callback) {
       this.delegate = throttledCall.delegate;
       this.isOverCapacity = throttledCall.isOverCapacity;
       this.callback = callback;
-      this.limitListener = limitListener;
     }
 
+    /**
+     * This awaits callback completion in order to slow down (throttle) calls.
```
anuraaga: It took me a while to follow how the throtlling works - to confirm, is it because the callbacks are run on a fixed number of threads? And the reason to use an `Executor` is to have both a limit on concurrent calls and control on the size of queued tasks, whereas a semaphore like the old ES implementation only handles concurrent? I think this makes sense just confirming my understanding.

zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
anuraaga: Also realized it might be better to have a `TracedCall` factory instead of calling the constructor directly - it could return a `Call` when something shouldn't be traced.

codefromthecrypt:
more later but a couple thoughts.. thanks for the feedback.


 The problem in the async reporter stating traces is presence of the global
sampler. unless the global sampler is set to always deny, a call starting
from a reporter thread will sample subject to the rate limit. this is
because it is not a client span rather a local one. we can change this by
moving the rate limiter to inside the http sampler and set the global one
to deny. this will rid need of that thread check stuff in our case.

regardless some of this defensiveness is to also rid that warning message
from armeria's currenttracecontext instance. There will be request scoped
stuff happening alongside not request scoped. any case of the latter.. even
unsampled will trigger that darned message. haven't thought about how to
get rid of it in a better way that literally removing instrumentation.

codefromthecrypt:
I guess one way is to disable the logger but still I would like to figure
out a way to have things that aren't request originated not trigger the
"contextAware" warning. something as simple as checking if a span exists
will trip it.

zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
codefromthecrypt: generally I feel both approaches are a bit crap from an abstraction pov, but I like your first workaround better as it feels less dodgy. Ex traced call shouldn't need special knowledge anyway. the main issue is around instrumentation policy on the edge case of self-tracing inside the tracer though.. and so yeah.. anyway this is good input. I'm hoping moving the sampler around can help.

zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingStorageComponent.java
codefromthecrypt: separate thread answered, but !server != client. the tracer itself lives at an abstraction below http (or messaging) and this call is considered a local span which is not a client.

anyway the defense was about preventing an infinite loop where the span reporter thread creates a span about reporting spans, which then creates a span about storage of that, and looping.

zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
codefromthecrypt: PS both of these are good suggestions.. the main problem is that the call makes a downstream call, and that would still hit a path that trips the request context based implementation (even if not sampled). Ex there's currently no thread-local means to say not just don't sample, but also don't instrument at all. Dancing around the log warning is probably an important pain to feel, even if it is super tempting to just quiet the message  heh.

zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingStorageComponent.java
codefromthecrypt: https://github.com/line/armeria/issues/1964

zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
codefromthecrypt: moved the root concern here https://github.com/line/armeria/issues/1964

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
codefromthecrypt: @Logic-32 can you validate this? I really do want to have the better docs here without getting too long winded. It isn't obvious at the moment.

zipkin-server/src/main/java/zipkin2/server/internal/brave/TracingConfiguration.java
codefromthecrypt: meanwhile one or both of these approaches is a lot of spaghetti drawing attention to something really unfortunate. I like the former is less code less beans less distraction as ideally we'll never have to expose this stuff... ex I don't get insight looking through /actuator/beans and now have 2 more beans just because of this crap logging concern. Meanwhile, I'm just going to kill the log message with a ref to the issue.

codefromthecrypt:
disabled the logger vs keep digging a hole and piling up beans around the log reporter thread. https://github.com/line/armeria/issues/1964 should help us discuss what to do. I don't think zipkin server will be the only app using both request-originated traffic and also not. We need to work out a good solution, but for now disabling the log vs adding spaghetti.

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
```diff
@@ -37,183 +38,181 @@
...
   }
 
-  static String setCurrentThreadName(String name) {
-    Thread thread = Thread.currentThread();
-    String originalName = thread.getName();
-    thread.setName(name);
-    return originalName;
+  static final class AwaitableCallback implements Callback<Void> {
+    final CountDownLatch countDown = new CountDownLatch(1);
+    final AtomicReference<Throwable> throwable = new AtomicReference<>();
```
Logic-32: I'm not immediately seeing the need for this to be an AtomicReference.  You create a new instance per-Call and only one Thread should execute that Call.  So you can save some time/space by just leaving this as a normal reference.

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
Logic-32: I'm sure I could go through the original PR and other supporting docs we made and find an answer for this.  But in the interest of not going cross-eyed, the short answer is that @anuraaga is correct.

A Semaphore is insufficient for the throttling goals as it would aggressively drop Spans.  The Executor offers us both the flexibility of rate-limiting requests to storage as well as the buffer (queue) so we don't have to aggressively drop Spans while storage is being sluggish.  It does all this in a nice, standardized bundle.

The control on the buffer size, btw, is so we don't reintroduce the original issue that the Semaphore was introduced to try and prevent.  Namely, OutOfMemoryErrors.  If a user desired, they could totally put an unbounded queue on the Executor but they'd be doing so at their own risk.

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
codefromthecrypt: ah yeah. I should add a comment. true that only one thread will set this. but the problem is that the thread invoking the callback isn't the same as the thread reading the exception. we could consider volatile though I suppose. would that be better?

Logic-32: You can consider `volatile` though I personally wouldn't.  I believe the semantics around Threads and variable assignment are safe enough that the assignment of that value should resolve before attempting to read it.  Simply because the Thread doing the read is waiting for the entire writing Thread to finish before progressing.  So the write implictly has to succeed before the read will even be attempted.  Assuming of course that this is only used in `doExecute()`.

Normally, `volatile` variables are used where there is a high likelihood of near-concurrent read/write.  Which again, I don't think is the case here.

codefromthecrypt:
hmm I have never taken on faith that a one thread writing a value (ex
Armeria's blocking thread pool) will be available for reading by a
different thread (ex this different thread pool) without synchronization
volatile or otherwise.

@trustin @anuraaga would you be comfortable with this? I can learn just
this is new to me.. I would be tempted to say which JVMs provide flush
guarantees as currently we work with jre 1.8 plus and this is the first
time I am hearing about this shortcut...

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
anuraaga: You would generally need `volatile` when writes and reads happen in different threads just to make sure the writes were already flushed to memory, it's less about the concurrency of the read/write.

But here it is actually clear that we don't need to use `volatile` since `CountDownLatch` guarantees writes are flushed before `await()` returns

https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CountDownLatch.html
`Memory consistency effects: Until the count reaches zero, actions in a thread prior to calling countDown() happen-before actions following a successful return from a corresponding await() in another thread.`

anuraaga:
As my comment, I'm comfortable with it. It's well documented by `CountDownLatch`. Without such documentation, I would generally err on safety though.

codefromthecrypt:
good point on happens before, and sorry to hijack this as a programming
lesson for me but in the name of me spreading to others.. does happens
before imply the same as flush? closest trick I have seen to a cheap
rendezvous was a single length array used with the same.. just never yet
heard that timing implies flush to other threads yet

anuraaga:
Yup - happens-before is JDK jargon for implying memory is flushed so subsequent reads definitely see it.

More details are in this well written summary, probably a good read - https://docs.oracle.com/javase/6/docs/api/java/util/concurrent/package-summary.html#MemoryVisibility

codefromthecrypt:
thanks for bearing with me guys. I now know more than before and will
happily clean out other overly defensive things

zipkin-server/src/main/java/zipkin2/server/internal/throttle/LimiterMetrics.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+package zipkin2.server.internal.throttle;
+
+import com.netflix.concurrency.limits.Limiter.Listener;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import zipkin2.collector.CollectorMetrics;
+
+/** Follows the same naming convention as {@link CollectorMetrics} */
+final class LimiterMetrics {
```
anuraaga, line 22: I don't know if this would improve anything, but just as a hint if this implemented `MeterBinder` and was a bean, spring would automatically register it.

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
anuraaga: Thanks for confirming! I think the words "slow down" confused me the most as I started looking for introducing artificial delay or something. Maybe just `waits until completion to ensure the number of executing calls doesn't surpass the concurrency limit of the executor`.

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
```diff
@@ -37,183 +37,187 @@
...
   }
 
-  static String setCurrentThreadName(String name) {
-    Thread thread = Thread.currentThread();
-    String originalName = thread.getName();
-    thread.setName(name);
-    return originalName;
+  static final class AwaitableCallback implements Callback<Void> {
+    final CountDownLatch countDown = new CountDownLatch(1);
+    Throwable throwable; // thread visibility guaranteed by the countdown latch
```
anuraaga, line 107: Good comment, it'd be unclear without I think.

Incidentally, realized it didn't matter here so much - a `volatile` variable has equivalent performance to a synchronized one if it's only ever written / read once ;)

anuraaga (approved):

zipkin-server/src/main/java/zipkin2/server/internal/throttle/LimiterMetrics.java
codefromthecrypt: ic. thx

codefromthecrypt:
cheers for the hard look folks. I'm mostly offline tomorrow but I think we're close to a release. Maybe we can wait for armeria 0.90 if it is soon.

Issue #2737, opened 2019-08-07
re-logo the things


codefromthecrypt:
for @syleeeee 

we need to update
./zipkin-lens/src/img/zipkin-sm-logo.svg
./zipkin-lens/src/img/zipkin-logo.svg

Also the banner, which was originally made by @making in #1612
./zipkin-server/src/main/resources/banner.txt

you can verify afterwards by starting the server using the instructions here:
https://github.com/openzipkin/zipkin/blob/master/zipkin-server/README.md#building-locally

maybe during this or after, the color scheme should change for lens. cc @openzipkin/ui

making:
I'll contribute banner.txt ;)

jorgheymans:
Things were re-logo'd, thanks !

jcchavezs:
Did we relogoed for attic and contrib too?

codefromthecrypt:
Did we relogoed for attic and contrib too?
>
we could use the old logo for attic, for the sake of irony :)

jorgheymans:
Attic still has the old logo +1 to keep it, i wanted to re-logo contrib but i think it requires more rights than I currently have. After that we can close this issue.

codefromthecrypt:
heh.. try now?

jorgheymans:
Yep, new logo applied to contrib thanks,

making:
https://twitter.com/zipkinproject is still using old one 🤔

codefromthecrypt:
unfortunately we have no access to the twitter account to change that, and
have failed for years getting someone to help us do that. I'll try again..

PR #2738, opened 2019-08-07
Replace banner.txt with new logo

making:
#2737

image


zipkin-server/src/main/resources/banner.txt
```diff
@@ -1,25 +1,28 @@
...
-                                       **
-                                       **
-
-
-             *****      **     *****     ** **       **     **   **
-               **       **     **  *     ***         **     **** **
-              **        **     *****     ****        **     **  ***
-             ******     **     **        **  **      **     **   **
+${AnsiColor.RED}
+                           o
```
jeqo: nit: I feel is missing one dot on the left

```suggestion
                          oo
```

making: Fixed :)

shakuzen:
We might want to wait on merging this for the outcome of the discussion on openzipkin vs zipkin in the logo.

codefromthecrypt:
actually definitely the server should just be zipkin. I think it would be
certainly incorrect here.

making:
Updated

image

codefromthecrypt (approved):

codefromthecrypt:
ahead of the curve huh? :D Screenshot 2019-08-08 at 1 21 31 PM

making:
I tested `banner.txt` in a latest spring boot app.

tacigar (approved):

codefromthecrypt:
if you can verify this is correct still with the new logo content: https://github.com/openzipkin/openzipkin.github.io/pull/142 then later or now we still need to update lens images

codefromthecrypt:
only difference is if we think red or orange, I guess

jeqo (approved):
sadly, seems that orange is not available https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/ansi/AnsiColor.html but I will let @making confirm if possible.
we might fallback to white as well.

making:
@jeqo @adriancole as far as I know, orange is not supposed.

codefromthecrypt:
I think black/white/grey is better than red when orange unsupported. Red is a more aggressive color.

can you raise an issue to support orange :)

making:
I'll try to implement a new `org.springframework.boot.ansi.AnsiElement` so that it supports 256 colors.

Which color do you prefer? :)
image

You can get this table by

```
for i in $(seq 1 255);do printf "%03d \033[38;5;${i}mZipkin\033[0;00m\t" $i; if [ $(( $i % 10 )) -eq 0 ];then echo;fi ;done;echo
```

making:
I think I managed to change the color :)

image

Let me update the source code later

jorgheymans:
> I think I managed to change the color :)
> 
> image
> 
> Let me update the source code later

Cool ! Will it work in a windows shell as well ?

making:
I've just pushed the updated banner in 4ca0ae3e5dcc14b420e5424bccf8b82fa91cd6b6.

image


@jorgheymans I don't have a Windows environment. Could you please give it a try?

making:
In 4ca0ae3e5dcc14b420e5424bccf8b82fa91cd6b6
I added `ZipkinAnsi256Color`, `ZipkinAnsi256PropertySource` and `ZipkinBanner`, then configured the banner programmatically as described in the boot [doc](https://docs.spring.io/spring-boot/docs/2.1.8.RELEASE/reference/html/boot-features-spring-application.html#boot-features-banner) in order to support 256 colors that is not available in Spring Boot out of the box now.
I'm thinking to contribute to Spring Boot but it wouldn't come to Spring Boot 2.1.x. So Zipkin needs to have these classes until a feature version.

zipkin-server/src/main/java/zipkin2/server/internal/banner/ZipkinAnsi256Color.java
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin2.server.internal.banner;
+
+import org.springframework.boot.ansi.AnsiElement;
+
+public class ZipkinAnsi256Color implements AnsiElement {
```
codefromthecrypt, line 23: add a TODO to delete on spring boot 2.2 if that's when this is scheduled to be normally possible.

making: Updated in edebc31428c9a3551181aa053ecb008b9dd2666e

codefromthecrypt:
Will merge this on green.

Added issue to update lens https://github.com/openzipkin/zipkin/issues/2806

codefromthecrypt (approved):
thanks again!

making:
It's been merged in Spring Boot 2.2 :)
https://github.com/spring-projects/spring-boot/pull/18264

codefromthecrypt:
nice to change your shoes on occasion!

PR #2740, opened 2019-08-07
Simplifies ThrottledCall to reduce GC overhead

codefromthecrypt:
@Logic-32 in one review noticed we could re-use a countdown latch when
doing execute. IIRC, he mentioned `Call` is one-shot anyway, which is
true. This was good advice. Flattening `ThrottledCall` made measurably
less GC in micro benchmarks.

Before
```
Benchmark                                                           Mode     Cnt     Score     Error   Units
ThrottledCallBenchmarks.execute                                   sample  774208    17.379 ±   0.077   us/op
ThrottledCallBenchmarks.execute:execute·p0.00                     sample             1.696             us/op
ThrottledCallBenchmarks.execute:execute·p0.50                     sample            15.376             us/op
ThrottledCallBenchmarks.execute:execute·p0.90                     sample            23.712             us/op
ThrottledCallBenchmarks.execute:execute·p0.95                     sample            29.664             us/op
ThrottledCallBenchmarks.execute:execute·p0.99                     sample            54.208             us/op
ThrottledCallBenchmarks.execute:execute·p0.999                    sample            95.616             us/op
ThrottledCallBenchmarks.execute:execute·p0.9999                   sample           185.829             us/op
ThrottledCallBenchmarks.execute:execute·p1.00                     sample          5464.064             us/op
ThrottledCallBenchmarks.execute:·gc.alloc.rate                    sample      15    40.681 ±   5.778  MB/sec
ThrottledCallBenchmarks.execute:·gc.alloc.rate.norm               sample      15   552.430 ±  77.257    B/op
ThrottledCallBenchmarks.execute:·gc.churn.G1_Eden_Space           sample      15    44.803 ±  53.001  MB/sec
ThrottledCallBenchmarks.execute:·gc.churn.G1_Eden_Space.norm      sample      15   612.823 ± 727.247    B/op
ThrottledCallBenchmarks.execute:·gc.churn.G1_Old_Gen              sample      15     0.052 ±   0.110  MB/sec
ThrottledCallBenchmarks.execute:·gc.churn.G1_Old_Gen.norm         sample      15     0.704 ±   1.488    B/op
ThrottledCallBenchmarks.execute:·gc.churn.G1_Survivor_Space       sample      15     0.044 ±   0.184  MB/sec
ThrottledCallBenchmarks.execute:·gc.churn.G1_Survivor_Space.norm  sample      15     0.588 ±   2.434    B/op
ThrottledCallBenchmarks.execute:·gc.count                         sample      15     7.000            counts
ThrottledCallBenchmarks.execute:·gc.time                          sample      15    29.000                ms
```

After
```
Benchmark                                                     Mode       Cnti       Score   Error   Units
ThrottledCallBenchmarks.execute                               sample  606001     17.678 ±   0.429   us/op
ThrottledCallBenchmarks.execute:execute·p0.00                 sample              9.152             us/op
ThrottledCallBenchmarks.execute:execute·p0.50                 sample             14.368             us/op
ThrottledCallBenchmarks.execute:execute·p0.90                 sample             23.360             us/op
ThrottledCallBenchmarks.execute:execute·p0.95                 sample             27.904             us/op
ThrottledCallBenchmarks.execute:execute·p0.99                 sample             42.368             us/op
ThrottledCallBenchmarks.execute:execute·p0.999                sample             80.768             us/op
ThrottledCallBenchmarks.execute:execute·p0.9999               sample           3725.718             us/op
ThrottledCallBenchmarks.execute:execute·p1.00                 sample          30277.632             us/op
ThrottledCallBenchmarks.execute:·gc.alloc.rate                sample      15     30.837 ±   6.145  MB/sec
ThrottledCallBenchmarks.execute:·gc.alloc.rate.norm           sample      15    401.481 ±  60.868    B/op
ThrottledCallBenchmarks.execute:·gc.churn.G1_Eden_Space       sample      15     32.435 ±  50.762  MB/sec
ThrottledCallBenchmarks.execute:·gc.churn.G1_Eden_Space.norm  sample      15    442.977 ± 708.124    B/op
ThrottledCallBenchmarks.execute:·gc.churn.G1_Old_Gen          sample      15      0.046 ±   0.130  MB/sec
ThrottledCallBenchmarks.execute:·gc.churn.G1_Old_Gen.norm     sample      15      0.567 ±   1.615    B/op
ThrottledCallBenchmarks.execute:·gc.count                     sample      15      5.000            counts
ThrottledCallBenchmarks.execute:·gc.time                      sample      15     18.000                ms
```

### DIFF
--- a/benchmarks/src/main/java/zipkin2/server/internal/throttle/ThrottledCallBenchmarks.java
+++ b/benchmarks/src/main/java/zipkin2/server/internal/throttle/ThrottledCallBenchmarks.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2015-2019 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin2.server.internal.throttle;
+
+import com.linecorp.armeria.common.metric.NoopMeterRegistry;
+import com.netflix.concurrency.limits.limit.FixedLimit;
+import com.netflix.concurrency.limits.limiter.SimpleLimiter;
+import java.io.IOException;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Predicate;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import zipkin2.Call;
+import zipkin2.Callback;
+
+@Measurement(iterations = 5, time = 1)
+@Warmup(iterations = 10, time = 1)
+@Fork(3)
+@BenchmarkMode(Mode.SampleTime)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@State(Scope.Thread)
+@Threads(2)
+public class ThrottledCallBenchmarks {
+  ExecutorService fakeCallExecutor = Executors.newSingleThreadExecutor();
+  ExecutorService executor = Executors.newSingleThreadExecutor();
+  ThrottledCall call;
+
+  @Setup public void setup() {
+    executor = Executors.newSingleThreadExecutor();
+    fakeCallExecutor = Executors.newSingleThreadExecutor();
+    SimpleLimiter<Void> limiter = SimpleLimiter.newBuilder().limit(FixedLimit.of(1)).build();
+    LimiterMetrics metrics = new LimiterMetrics(NoopMeterRegistry.get());
+    Predicate<Throwable> isOverCapacity = RejectedExecutionException.class::isInstance;
+    call =
+      new ThrottledCall(new FakeCall(fakeCallExecutor), executor, limiter, metrics, isOverCapacity);
+  }
+
+  @TearDown public void tearDown() {
+    executor.shutdown();
+    fakeCallExecutor.shutdown();
+  }
+
+  @Benchmark public Void execute() throws IOException {
+    return call.clone().execute();
+  }
+
+  static final class FakeCall extends Call.Base<Void> {
+    final Executor executor;
+    boolean overCapacity = false;
+
+    FakeCall(Executor executor) {
+      this.executor = executor;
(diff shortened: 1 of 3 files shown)


anuraaga (approved):

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
```diff
@@ -59,38 +69,38 @@
...
+    if (t instanceof RuntimeException) throw (RuntimeException) t;
+    throw new RuntimeException(t);
   }
 
+  // When handling enqueue, we don't block the calling thread. Any exception goes to the callback.
   @Override protected void doEnqueue(Callback<Void> callback) {
     Listener limiterListener = limiter.acquire(null)
       .orElseThrow(RejectedExecutionException::new); // TODO: make an exception message
-    limiterMetrics.requests.increment();
-    limiterListener = limiterMetrics.wrap(limiterListener);
```
Logic-32, line 80: I see no equivalent for this in the new code which could throw off GC metrics.

benchmarks/src/main/java/zipkin2/server/internal/throttle/ThrottledCallBenchmarks.java
```diff
@@ -0,0 +1,112 @@
...
+    call =
+      new ThrottledCall(new FakeCall(fakeCallExecutor), executor, limiter, metrics, isOverCapacity);
+  }
+
+  @TearDown public void tearDown() {
+    executor.shutdown();
+    fakeCallExecutor.shutdown();
+  }
+
+  @Benchmark public Void execute() throws IOException {
```
Logic-32, line 72: In order to avoid oddities that occur between JVM boots you should add a second `@Benchmark` which would test the second implementation under the same circumstances.  Basically, do a ThrottledCall and ThrottledCall2 impl.

Also, and I never did figure out how to do to this properly so please let me know if you figure it out, but you should test this with the G1GC compiler as well since that is the default in more modern JVMs.  I think I succeeded at doing that once by simply switching to the JRE11.

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
```diff
@@ -38,11 +38,21 @@
  * @see ThrottledStorageComponent
  */
 final class ThrottledCall extends Call.Base<Void> {
```
Logic-32, line 40: An interesting, GC related, aspect of this change is that you technically may have increased the memory footprint of this class while keeping the EnqueueAndAwait class the same (removed a member variable but made it non-static).  Technically speaking, there is a chance that would make GC worse but the benchmarks are reasonably clear at the moment.

Either way, one thing that would be interesting to see is if you ran [JoL](https://openjdk.java.net/projects/code-tools/jol/) against the two instances (look at the samples, it's pretty easy to use) and see if any concerning differences pop up.

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
```diff
@@ -59,38 +69,38 @@
...
-    Throwable t = awaitableCallback.throwable;
-    if (t != null) {
-      if (t instanceof Error) throw (Error) t;
-      if (t instanceof IOException) throw (IOException) t;
-      if (t instanceof RuntimeException) throw (RuntimeException) t;
-      throw new RuntimeException(t);
-    }
-    return null; // Void
+    // Enqueue the call invocation on the executor and block until it completes.
+    doEnqueue(NOOP_CALLBACK);
```
Logic-32, line 73: This is another thing that can affect the GC metrics: constant callback instead of a new one.

I'm thinking the real gains you got from this change is less related to CountDownLatch and more to some of the tertiary changes.

zipkin-server/src/main/java/zipkin2/server/internal/throttle/ThrottledCall.java
```diff
@@ -38,11 +38,21 @@
...
+    @Override public void onError(Throwable t) {
+    }
+  };
+
   final Call<Void> delegate;
   final Executor executor;
   final Limiter<Void> limiter;
   final LimiterMetrics limiterMetrics;
   final Predicate<Throwable> isOverCapacity;
+  final CountDownLatch latch = new CountDownLatch(1);
```
Logic-32, line 54: Would you be opposed to moving this instantiation into the existing constructor?  By leaving it up here you litterally end up with two constructors in your class which can unnecessarily increase class size and slown down instantiation times.  Unless the HotSpot optimizer can detect and fix this of course ;)

Logic-32:
Per some of my comments/concerns in the review, you may want to un-merge this for just a minute while we hash some things out :)

codefromthecrypt:
feel free to raise a different pr to improve but please use jmh or other
objective way to measure whatever it is you are concerned about

PR #2741, opened 2019-08-08
Enhance unit tests for App components

tacigar:
## Changes

* Add `react-testing-library`(https://github.com/testing-library/react-testing-library) for testing.
* Update unit tests for components which are under `/src/components/App` directory.
* Refactor some components just a little.
    * Add `classnames`(https://github.com/JedWatson/classnames) library to write className concisely.

---

#### Before

Before

#### After

After

---

So far we have only used Enzyme for testing framework.
But there is little documentation on testing components with React Hooks + Enzyme.
And I've had several experiences where the Enzyme + React Hooks test didn't work as expected.

On the other hand, as far as I know, there are many materials using [react-testing-library](https://github.com/testing-library/react-testing-library) + Hooks.
And react-testing-library is simple and does not have so many features, but seems to work good.

So I came to the conclusion that it would be best to use both Enzyme and react-testing-library in hybrid.
AFAIK, it seems that material-ui is taking the same method.
This PR is the beginning of that attempt...

### DIFF
--- a/zipkin-lens/src/components/App/App.jsx
+++ b/zipkin-lens/src/components/App/App.jsx
@@ -23,15 +23,10 @@ import DiscoverPage from '../DiscoverPage';
 import TracePage from '../TracePage';
 import configureStore from '../../store/configure-store';
 import { theme } from '../../colors';
-import { useMount } from '../../hooks';
-
-const applicationTitle = 'Zipkin';
+import { useDocumentTitle } from '../../hooks';
 
 const App = () => {
-  useMount(() => {
-    document.title = applicationTitle;
-  });
-
+  useDocumentTitle('Zipkin');
   return (
     <MuiPickersUtilsProvider utils={MomentUtils}>
       <ThemeProvider theme={theme}>
--- a/zipkin-lens/src/components/App/App.test.js
+++ b/zipkin-lens/src/components/App/App.test.js
@@ -11,54 +11,61 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  */
-import { shallow } from 'enzyme';
 import React from 'react';
 import { Provider } from 'react-redux';
 import { BrowserRouter, Route } from 'react-router-dom';
 import { ThemeProvider } from '@material-ui/styles';
 import { MuiPickersUtilsProvider } from '@material-ui/pickers';
+import { createMount, createShallow } from '@material-ui/core/test-utils';
 
 import App from './App';
 import Layout from './Layout';
+import DiscoverPage from '../DiscoverPage';
+import TracePage from '../TracePage';
+import { theme } from '../../colors';
 
 describe('<App />', () => {
-  let wrapper;
+  let mount;
+  let shallow;
 
   beforeEach(() => {
-    wrapper = shallow(<App />);
+    mount = createMount({ strict: true });
+    shallow = createShallow();
+  });
+
+  afterEach(() => {
+    mount.cleanUp();
+  });
+
+  it('should update document title', () => {
+    mount(<App />);
+    expect(document.title).toBe('Zipkin');
+  });
+
+  it('should render providers', () => {
+    const wrapper = shallow(<App />);
+    expect(wrapper.find(MuiPickersUtilsProvider).length).toBe(1);
+    expect(wrapper.find(ThemeProvider).length).toBe(1);
+    expect(wrapper.find(ThemeProvider).props().theme).toEqual(theme);
+    expect(wrapper.find(Provider).length).toBe(1);
+  });
+
+  it('should render Router as a parent of Layout', () => {
+    const wrapper = shallow(<App />);
+    expect(wrapper.find(Layout).parent().type()).toEqual(BrowserRouter);
   });
 
   it('should render Layout', () => {
-    const items = wrapper.find(Layout);
-    expect(items.length).toBe(1);
+    const wrapper = shallow(<App />);
+    expect(wrapper.find(Layout).length).toBe(1);
   });
 
-  describe('should render router', () => {
-    it('should render BrowserRouter', () => {
-      const items = wrapper.find(BrowserRouter);
-      expect(items.length).toBe(1);
-    });
-
-    it('should render 2 routes', () => {
-      const items = wrapper.find(Route);
-      expect(items.length).toBe(2);
-    });
-  });
-
-  describe('should render providers', () => {
-    it('should render MuiPickersUtilsProvider', () => {
-      const items = wrapper.find(MuiPickersUtilsProvider);
-      expect(items.length).toBe(1);
-    });
-
-    it('should render ThemeProvider', () => {
-      const items = wrapper.find(ThemeProvider);
-      expect(items.length).toBe(1);
-    });
-
-    it('should render redux Provider', () => {
-      const items = wrapper.find(Provider);
-      expect(items.length).toBe(1);
-    });
+  it('should render Route', () => {
+    const wrapper = shallow(<App />);
+    const routes = wrapper.find(Route);
+    expect(routes.at(0).props().path).toEqual(['/zipkin', '/zipkin/dependency']);
+    expect(routes.at(0).props().component).toEqual(DiscoverPage);
+    expect(routes.at(1).props().path).toEqual(['/zipkin/traces/:traceId', '/zipkin/traceViewer']);
+    expect(routes.at(1).props().component).toEqual(TracePage);
   });
 });
--- a/zipkin-lens/src/components/App/App.test.js
+++ b/zipkin-lens/src/components/App/App.test.js
@@ -17,6 +17,7 @@ import { BrowserRouter, Route } from 'react-router-dom';
 import { ThemeProvider } from '@material-ui/styles';
 import { MuiPickersUtilsProvider } from '@material-ui/pickers';
 import { createMount, createShallow } from '@material-ui/core/test-utils';
+import MomentUtils from '@date-io/moment';
 
 import App from './App';
 import Layout from './Layout';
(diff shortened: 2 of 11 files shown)


codefromthecrypt:
is there some way to quiet the advertising here?

```
[INFO] > node scripts/postinstall || echo "ignore"
[INFO] 
[INFO] Thank you for using core-js ( https://github.com/zloirock/core-js ) for polyfilling JavaScript standard library!
[INFO] 
[INFO] The project needs your help! Please consider supporting of core-js on Open Collective or Patreon: 
[INFO] > https://opencollective.com/core-js 
[INFO] > https://www.patreon.com/zloirock 
[INFO] 
[INFO] Also, the author of core-js ( https://github.com/zloirock ) is looking for a good job -)
[INFO] 
[INFO] 
```

codefromthecrypt:
can you add instructions to get coverage report to the README? Also, a separate change, but we should have debugger instructions like we do in zipkin-js also

zipkin-lens/src/components/App/App.jsx
```diff
@@ -23,15 +23,10 @@ import DiscoverPage from '../DiscoverPage';
...
-
-const applicationTitle = 'Zipkin';
+import { useDocumentTitle } from '../../hooks';
 
 const App = () => {
-  useMount(() => {
-    document.title = applicationTitle;
-  });
-
+  useDocumentTitle('Zipkin');
```
codefromthecrypt, line 29: simpler I like

zipkin-lens/src/components/App/App.test.js
```diff
@@ -11,54 +11,61 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  */
-import { shallow } from 'enzyme';
```
codefromthecrypt, line 14: is enzyme still used directly? if not, do we need to change package.json?

zipkin-lens/src/components/App/App.test.js
```diff
@@ -11,54 +11,61 @@
...
+    mount.cleanUp();
   });
 
-  describe('should render router', () => {
-    it('should render BrowserRouter', () => {
-      const items = wrapper.find(BrowserRouter);
-      expect(items.length).toBe(1);
-    });
+  it('should update document title', () => {
+    mount(<App />);
```
codefromthecrypt, line 44: for my learning.. mount() seems to be documented for things that need access to whole dom. Can you briefly mention why we need whole dom for this test? I guess shallow doesn't set the title? https://material-ui.com/guides/testing/#full-dom-rendering

zipkin-lens/src/components/App/App.test.js
```diff
@@ -11,54 +11,61 @@
...
+    expect(document.title).toBe('Zipkin');
+  });
 
-    it('should render 2 routes', () => {
-      const items = wrapper.find(Route);
-      expect(items.length).toBe(2);
-    });
+  it('should render providers', () => {
```
codefromthecrypt: this test says what not why. providers are a generic term. Ex. 'should provide only one theme' or something.

then, below, have a comment for each and what they control. This will give the why about this test. In other words, I think there are separate goals.. coverage is one and why we code is another. The latter is more thought even if the former more work.

zipkin-lens/src/components/App/App.test.js
```diff
@@ -11,54 +11,61 @@
...
+    expect(wrapper.find(ThemeProvider).props().theme).toEqual(theme);
+    expect(wrapper.find(Provider).length).toBe(1);
   });
 
-  describe('should render providers', () => {
-    it('should render MuiPickersUtilsProvider', () => {
-      const items = wrapper.find(MuiPickersUtilsProvider);
-      expect(items.length).toBe(1);
-    });
+  it('should render Router as a parent of Layout', () => {
```
codefromthecrypt: why do I care? ex. what would happen if Layout was not the parent?

zipkin-lens/src/components/App/App.test.js
```diff
@@ -11,54 +11,61 @@
...
+  it('should render Layout', () => {
+    const wrapper = shallow(<App />);
+    expect(wrapper.find(Layout).length).toBe(1);
+  });
 
-    it('should render redux Provider', () => {
-      const items = wrapper.find(Provider);
-      expect(items.length).toBe(1);
-    });
+  it('should render Route', () => {
```
codefromthecrypt: check length of routes.

zipkin-lens/src/components/App/App.test.js
```diff
@@ -11,54 +11,61 @@
...
+    const wrapper = shallow(<App />);
+    const routes = wrapper.find(Route);
+    expect(routes.at(0).props().path).toEqual(['/zipkin', '/zipkin/dependency']);
```
codefromthecrypt, line 80:
```suggestion
    // Check routes that use the same global search bar
    expect(routes.at(0).props().path).toEqual(['/zipkin', '/zipkin/dependency']);
```

zipkin-lens/src/components/App/App.test.js
```diff
@@ -11,54 +11,61 @@
...
+    expect(routes.at(0).props().component).toEqual(DiscoverPage);
+    expect(routes.at(1).props().path).toEqual(['/zipkin/traces/:traceId', '/zipkin/traceViewer']);
```
codefromthecrypt, line 83:
```suggestion
    // check routes that use a single trace as input
    expect(routes.at(1).props().path).toEqual(['/zipkin/traces/:traceId', '/zipkin/traceViewer']);
```

zipkin-lens/src/components/App/Sidebar.test.js
```diff
@@ -11,63 +11,40 @@
...
-    });
-
-    it('should not pass isExternalLink', () => {
-      internalLinks.forEach((internalLink) => {
-        expect(internalLink.props().isExternalLink).toBeUndefined();
-      });
-    });
+  it('should render internal links', () => {
+    const wrapper = shallow(<Sidebar />);
+    const list = wrapper.find('[data-testid="internal-links"]');
```
codefromthecrypt, line 41: add comment reminding difference between internal and external links

zipkin-lens/src/components/App/SidebarMenuItem.test.js
```diff
@@ -11,128 +11,98 @@
...
+      expect(logo.props().component).toBe('span');
+      expect(logo.props().className).toBe('fab fa-home');
     });
 
-    it('should not call push when clicked', () => {
-      const wrapper = shallow(<SidebarMenuItem.WrappedComponent {...commonProps} />);
-      wrapper.find(ListItem).simulate('click');
-      expect(pushMock.mock.calls.length).toBe(0);
+    it('should call history.push when ListItem is clicked', () => {
+      const { getByTestId } = render(<SidebarMenuItemImpl {...props} />);
```
codefromthecrypt, line 102: add comment why history.push is important (I presume so you can hit back button)

codefromthecrypt (commented):
Thanks so much. As you are increasing rigor, I thought it is a good time to make tests more meaningful. I sampled some tests and added commentary, but it will be the same commentary for a lot of tests. If you have time, please adjust some of these so that others jumping in need less guessing to understand why something exists or what would be broken if the test fails.

codefromthecrypt:
ps not a big deal as many of the integration tests we have take forever, but I noticed this which seems long for a javascript test

```
 PASS  src/components/App/App.test.js (7.643s)
```

tacigar:
Thank you for your review!
I'll update test code.
Just a moment, please 🙇

zipkin-lens/src/components/App/App.test.js
tacigar: Yes Enzyme is still used directly in other components.

tacigar:
PTAL 🙇

zipkin-lens/src/components/App/App.test.js
```diff
@@ -37,34 +38,48 @@ describe('<App />', () => {
...
-  it('should update document title', () => {
+  it('should update document title when mounted', () => {
+    // Since it is not possible to verify that document.title has
+    // been changed using shallow rendering, use mount.
     mount(<App />);
     expect(document.title).toBe('Zipkin');
   });
 
-  it('should render providers', () => {
+  it('should provide React Context for MuiPickers', () => {
```
codefromthecrypt: close.. I would ask why do I care about MuiPicker? Which function of the app does MuiPicker provide?

Ex 'should provide React Context for time range selection' is more a why MuiPicker is a what

zipkin-lens/src/components/App/App.test.js
```diff
@@ -37,34 +38,48 @@ describe('<App />', () => {
...
+    expect(wrapper.find(Provider).length).toBe(1);
   });
 
-  it('should render Layout', () => {
+  it('should share Layout among all page components', () => {
     const wrapper = shallow(<App />);
-    expect(wrapper.find(Layout).length).toBe(1);
+    const layout = wrapper.find(Layout);
+    expect(layout.length).toBe(1);
+    // In order to share Layout, the Route components must be defined under Layout.
```
codefromthecrypt, line 69: good comments

zipkin-lens/src/components/App/App.test.js
```diff
@@ -37,34 +38,48 @@ describe('<App />', () => {
...
+    // Since the Layout component is wrapped with withRouter, it needs to be defined
+    // under BrowserRouter.
+    expect(layout.parent().type()).toEqual(BrowserRouter);
   });
 
-  it('should render Route', () => {
+  it('should render 2 Route', () => {
     const wrapper = shallow(<App />);
     const routes = wrapper.find(Route);
+    // Check routes that use a single trace as input.
```
codefromthecrypt: I think this should say `// Check routes that use the same global search bar` or similar right? these are about multiple traces

zipkin-lens/src/components/App/Layout.test.js
```diff
@@ -25,7 +25,7 @@ describe('<Layout />', () => {
     shallow = createShallow();
   });
 
-  it('should render CssBaseline', () => {
+  it('should render CssBaseline to use Material UI\'s base styles', () => {
```
codefromthecrypt, line 28: cool

zipkin-lens/src/components/App/Sidebar.test.js
```diff
@@ -39,12 +39,14 @@ describe('<Sidebar />', () => {
   it('should render internal links', () => {
     const wrapper = shallow(<Sidebar />);
     const list = wrapper.find('[data-testid="internal-links"]');
+    // Only discover page is the internal link.
```
codefromthecrypt: I guess I understand now that "internal link" means link served by the application here (routed in javascript) as opposed to "external link" meaning http request.. is that right?

codefromthecrypt (approved):
thanks much better!

zipkin-lens/src/components/App/App.test.js
tacigar: Oops!

tacigar:
Thank you for your feedback!!
PTAL 🙇

zipkin-lens/src/components/App/App.test.js
```diff
@@ -11,54 +11,76 @@
...
-    it('should render ThemeProvider', () => {
-      const items = wrapper.find(ThemeProvider);
-      expect(items.length).toBe(1);
-    });
+  it('should provide redux store', () => {
+    const wrapper = shallow(<App />);
+    expect(wrapper.find(Provider).length).toBe(1);
+  });
+
+  it('should share Layout among all page components', () => {
```
codefromthecrypt, line 65: good name

codefromthecrypt (approved):
🙇 excellent

PR #2744, opened 2019-08-08
Tweaks throttled benchmarks so it is easier to compare older commits

codefromthecrypt:
Whenever I did JMH stuff, I kept updated the benchmark and improved the
code, later reverting the main code and running again. I often chose to
combine multiple small improvements in the same PR. I noticed others
due so too.

Anyway, @Logic-32 has questioned if changes made things worse despite me
pasting in before and after runs using JRE 11 (which is what our docker
image uses).

This change makes it easier for someone else to run the same benchmark
file against older commits so that any further improvements can be
assessed independently all the way back to vs the last release.

For example with the below diff you can run against the tag 2.15.0.

```diff
diff --git a/benchmarks/src/main/java/zipkin2/server/internal/throttle/ThrottledCallBenchmarks.java b/benchmarks/src/main/java/zipkin2/server/internal/throttle/ThrottledCallBenchmarks.java
index 5e9ce58be..8d05230f0 100644
--- a/benchmarks/src/main/java/zipkin2/server/internal/throttle/ThrottledCallBenchmarks.java
+++ b/benchmarks/src/main/java/zipkin2/server/internal/throttle/ThrottledCallBenchmarks.java
@@ -13,7 +13,6 @@
  */
 package zipkin2.server.internal.throttle;

-import com.linecorp.armeria.common.metric.NoopMeterRegistry;
 import com.netflix.concurrency.limits.limit.FixedLimit;
 import com.netflix.concurrency.limits.limiter.SimpleLimiter;
 import java.io.IOException;
@@ -22,7 +21,6 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Predicate;
 import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.BenchmarkMode;
 import org.openjdk.jmh.annotations.Fork;
@@ -58,10 +56,8 @@ public class ThrottledCallBenchmarks {
     executor = Executors.newSingleThreadExecutor();
     fakeCallExecutor = Executors.newSingleThreadExecutor();
     SimpleLimiter<Void> limiter = SimpleLimiter.newBuilder().limit(FixedLimit.of(1)).build();
-    LimiterMetrics metrics = new LimiterMetrics(NoopMeterRegistry.get());
-    Predicate<Throwable> isOverCapacity = RejectedExecutionException.class::isInstance;
     call =
-      new ThrottledCall(new FakeCall(fakeCallExecutor), executor, limiter, metrics, isOverCapacity);
+      new ThrottledCall(executor, limiter, new FakeCall(fakeCallExecutor));
   }

   @TearDown public void tearDown() {

```

### DIFF
--- a/benchmarks/src/main/java/zipkin2/server/internal/throttle/ThrottledCallBenchmarks.java
+++ b/benchmarks/src/main/java/zipkin2/server/internal/throttle/ThrottledCallBenchmarks.java
@@ -42,8 +42,6 @@ import org.openjdk.jmh.runner.options.OptionsBuilder;
 import zipkin2.Call;
 import zipkin2.Callback;
 
-import static com.linecorp.armeria.common.util.Exceptions.clearTrace;
-
 @Measurement(iterations = 5, time = 1)
 @Warmup(iterations = 10, time = 1)
 @Fork(3)
@@ -71,7 +69,7 @@ public class ThrottledCallBenchmarks {
     fakeCallExecutor.shutdown();
   }
 
-  @Benchmark public Void execute() throws IOException {
+  @Benchmark public Object execute() throws IOException {
     return call.clone().execute();
   }
 
@@ -88,14 +86,10 @@ public class ThrottledCallBenchmarks {
 
   @Benchmark public void execute_throttled() throws IOException {
     call.limiter.acquire(null); // capacity is 1, so this will overdo it.
-    try {
-      call.clone().execute();
-    } catch (RejectedExecutionException e) {
-      assert e == ThrottledCall.STORAGE_THROTTLE_MAX_CONCURRENCY;
-    }
+    call.clone().execute();
   }
 
-  static final Exception OVER_CAPACITY = clearTrace(new RejectedExecutionException("overCapacity"));
+  static final RejectedExecutionException OVER_CAPACITY = new RejectedExecutionException();
 
   static final class FakeCall extends Call.Base<Void> {
     final Executor executor;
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
warning I'm out of time on the throttled code, which has gotten a massive amount of my time lately. The last changes I did were mainly to prove I could understand the code, which I didn't at first. I wanted to follow-up because I was stuck on a flight with no internet and had some focus time to look deeply. Right now, though I really can't sink another week into saving a few hundred more bytes, but we certainly have saved a good bit here in a provable way.

codefromthecrypt:
folks really interested in the overall performance, I dare to look at the elasticsearch part of this :) we're looking at the tip of the iceberg considering this code is only used for storage and there is so much that happens besides what we've been gazing at.

luckily @anuraaga started focusing on that by recently introducing `BulkRequestBenchmarks`. IOTW, I'd recommend saving an order of magnitude more bytes with the same time by placing effort in a file that hasn't been looked at yet vs casting doubt on one we've really poured over.

anuraaga (approved):
Thanks for all the changes. From the diff, I'm not so sure what changed here but am guessing old semantics have been brought back to life for easier testing against old code.

codefromthecrypt:
before the type was generic, that was one change, also it didn't have the same metrics, that was the other thing api affecting (granted this is a hidden api). The biggest thing on this code path was that in the recent code, we reuse the "enqueue" path even for "execute"

Intuitively, that might imply performance worse due to being less direct. I was most worried about hurting perf in this move.

The main reason for moving the code around was about it being easier to see what's going on, less copy/paste and less entrypoints to scrutinize. Now, the code has a single entry for the more concerning paths, which really helps testing as we can borrow the same unit tests and metrics for any path through. Basically, I think the goal of making this file as boring as possible is mostly there. 

Not this release, but I think after the next ideally we can pull the "experimental status" from the throttling code. We do have some backlog prior.. we can ignore mysql, but we've not looked at if this helps with cassandra at all, and also certainly we should try this with GCP as any SaaS should be throttled I think. Anyway, this brain dump is here because soon everything in my mind will dump to make room for the UI work that's next :)

Thanks again @Logic-32 and @anuraaga

PR #2746, opened 2019-08-09
Classifies Elasticsearch ResponseTimeoutException as capacity related

codefromthecrypt:
@anuraaga and I noticed that `ResponseTimeoutException` is raised when
Elasticsearch is unresponsive. Eventhough it can be raised for other
reasons, it seems a better choice to classify this as capacity related
vs not.

You'll notice the upstream project also classifies timeout as a signal
of capacity with the same result (drop)

https://github.com/Netflix/concurrency-limits/pull/132


anuraaga:
Is this a good time to remove this wrapping?

https://github.com/openzipkin/zipkin/blob/master/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java#L165

codefromthecrypt:
the re-wrapping has the secondary effect of trimming the cause back. so, as long as we are ok with that, we can leave it and then classify the unprocessed exception naturally. wdyt?

anuraaga:
Makes sense let's leave it as is

PR #2750, opened 2019-08-10
Addresses most errorprone feedback so we have less compile warnings

codefromthecrypt:
This makes a couple small dependency bumps and clears almost all of the
errorprone warnings.


zipkin-collector/scribe/src/main/java/zipkin2/collector/scribe/NettyScribeServer.java
```diff
@@ -61,6 +60,7 @@ void start() {
 
   void close() {
     if (channel == null) return;
+    // TODO: chain these futures, and probably block a bit
```
codefromthecrypt, line 63: this is the only one I left as a compile warning, as I think it is legit. help wanted on the most elegant way to chain this granted the return val is void

codefromthecrypt:
added commit from https://github.com/openzipkin/zipkin/pull/2751 as the underlying syntax glitch isn't worth messing with

PR #2751, opened 2019-08-10
Restricts build to JDK 11

codefromthecrypt:
Before, we were pinned to the EOL JDK 1.8 because we started Kafka 0.8. We forgot to remove the pinning part. More importantly, moving off 1.8 fixes build problems related to errorprone.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,7 @@ cache:
 
 language: java
 
-jdk:
-  # needs to be JDK 1.8 as long as we start Kafka 0.8
-  - oraclejdk8
+jdk: openjdk11
 
 services:
   - docker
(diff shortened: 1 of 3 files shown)


shakuzen:
We probably will want to mention in the README that the project requires JDK 11 to build now.

shakuzen:
Also, I know we use animalsniffer, but upgrading to JDK9+ presents the opportunity to use the new `release` compiler arg instead, where the compiler can enforce that Java API is not used beyond the specified version. This doesn't need to happen in this pull request, though.
ref: http://openjdk.java.net/jeps/247

codefromthecrypt:
https://github.com/openzipkin/zipkin/issues/2752 on the --release thingy thanks

codefromthecrypt:
good catch on the README @shakuzen

shakuzen (approved):

codefromthecrypt:
merging as the only change after a green was to README and the build takes ages.

PR #2761, opened 2019-08-19
Define an interface for supplying a request to give call sites more c…

anuraaga:
…ontrol over how the request is created (e.g., using pooled buffers or streaming).

I think this was your idea :)

I have a lot of personal work today and tomorrow so don't know if I can get to adding tests / cleaning up that this needs - but wanted to get it out there in case you want to play with it. I only ran the integration tests and didn't run any benchmarks either, but in theory it should reduce memory usage by 1) not copying into a heap byte[] and 2) writing entries to the request while it's executing so they're streamed onto the connection. A future extension would be to use a supplier tied to json writing instead of `AggregatedHttpRequest` for the search requests so those also can use pooled buffers, but it's much lower priority.

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/BulkCallBuilder.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/BulkCallBuilder.java
@@ -17,17 +17,18 @@ import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.google.auto.value.AutoValue;
-import com.linecorp.armeria.common.AggregatedHttpRequest;
 import com.linecorp.armeria.common.HttpData;
 import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.common.HttpMethod;
+import com.linecorp.armeria.common.HttpRequest;
+import com.linecorp.armeria.common.HttpRequestWriter;
 import com.linecorp.armeria.common.MediaType;
 import com.linecorp.armeria.common.RequestContext;
 import com.linecorp.armeria.common.RequestHeaders;
+import com.linecorp.armeria.common.util.Exceptions;
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.ByteBufOutputStream;
-import io.netty.buffer.ByteBufUtil;
-import io.netty.buffer.CompositeByteBuf;
 import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.handler.codec.http.QueryStringEncoder;
 import java.io.IOException;
@@ -39,6 +40,7 @@ import zipkin2.elasticsearch.ElasticsearchStorage;
 import zipkin2.elasticsearch.internal.client.HttpCall;
 import zipkin2.elasticsearch.internal.client.HttpCall.BodyConverter;
 
+import static zipkin2.Call.propagateIfFatal;
 import static zipkin2.elasticsearch.internal.JsonSerializers.OBJECT_MAPPER;
 
 // See https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html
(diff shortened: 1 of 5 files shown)


zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
```diff
@@ -56,6 +57,60 @@
     V convert(JsonParser parser, Supplier<String> contentString) throws IOException;
   }
 
+  public interface RequestSupplier {
+    /**
+     * Creates a new {@link HttpRequest} to send. If this {@link HttpRequest} is not closed, it
```
codefromthecrypt: can you add javadoc link on closed? sorry does request have a close function? I don't quite get this part.

codefromthecrypt (commented):
I think I get most of this. What I don't get is why we don't fully abstract the creation of the http request.

This feels like half-preparing one, which has one side-effect of putting low-level code in two types: the call site and also the implementation of HttpCall.

What if similar to cassandra ResultSetFutureCall, our HttpCall was abstract. there'd be no split duties or need to document cloning behavior per-call because nothing would be written at all until execute/enqueue anyway. Each of the less than 10 commands would be a subclass which then can internally deal with if the resultant one-shot request is immutable or not.. it wouldn't matter.

Meanwhile if you you think this particular avenue is worth persuing let me know. I have a couple questions, but if duties were centralized by subtype I think many of these questions would disappear.

anuraaga:
Sorry this was just put together stream of thought, I agree the API is junk. Was hoping either you, or me once I finish moving preparations, first benchmark the difference before cleaning it up since the abstractions won't affect the performance :)

codefromthecrypt:
ah ok I get it now. yeah makes sense. sure I can fire up the engine and see

anuraaga:
Thanks! Realized I forgot to mark this as WIP which added confusion

codefromthecrypt:
I did a quick run just to check. I can't see a significant drop in memory usage, but then again I assume the prometheus endpoint is troubled as in the last two runs each lost metrics during peak load (gaps in the graph)

```bash
curl -s localhost:9411/actuator/info
{"zipkin":{"version":"2.16.2-SNAPSHOT","commit":"2ea9c0a"}}
```

I'm gonna try to sort out something else unrelated tonight

codefromthecrypt:
ps I use the jitpack branch of docker-zipkin repo then start like..

```
docker-compose -f docker-compose.yml -f docker-compose-elasticsearch.yml build
docker-compose -f docker-compose.yml -f docker-compose-elasticsearch.yml up
```

then the `wrk -t4 -c128 -d100s http://localhost:8081 --latency` against it

the other thing as mentioned offline, is that now we are able to kill ES 7 :P not sure if that's because of load or more memory displacement or what.

codefromthecrypt:
will give another go, thanks

codefromthecrypt:
I will switch to stock sleuth-webmvc-example because it is slower and results in less things falling over so fast.

anuraaga:
Thanks, if you have some time
 Didn't add any comment yet since not sure if the change helps but it is definitely important. I'm not able to reproduce metrics endpoint falling over on my computer even without it so still going to continue investigation. I'm thinking it's important to look at spans accepted too since without the semaphore, more spans will be accepted I believe, so that will use more memory.

codefromthecrypt:
ok good news.. heap usage looks ok. I do need to rebase this vs 2.15 as I also changed the load. Also, I want to make sure some amount of load actually succeed :D right now almost all spans dropped.

Screenshot 2019-08-21 at 10 32 48 AM

codefromthecrypt:
ex in this run, max json payload was 600k, max spans/second 1.2k

146675 spans accepted
132041 spans dropped

I know this is elasticsearch and all, but for example our in-memory defaults to 500k spans before expiring. It seems like it should be possible to swallow this many but anyway..

codefromthecrypt:
I will turn on the throttling, also, but we also have an existing bug where the stats end up messed up..

codefromthecrypt:
I think for storage throttling I will just count documents in elasticsearch until stats are fixed

codefromthecrypt:
now trying with throttling enabled.. and will check the document count directly after to see if the metrics messup was spurious or not.

```diff
--- a/docker-compose-elasticsearch.yml
+++ b/docker-compose-elasticsearch.yml
@@ -12,8 +12,8 @@ services:
     image: openzipkin/zipkin-elasticsearch7
     container_name: elasticsearch
     # Uncomment to expose the storage port for testing
-    # ports:
-    #   - 9200:9200
+    ports:
+      - 9200:9200
 
   # Switch storage type to Elasticsearch
   zipkin:
@@ -22,6 +22,7 @@ services:
       # Point the zipkin at the storage backend
       - ES_HOSTS=elasticsearch
       # Uncomment to see requests to and from elasticsearch
+      - STORAGE_THROTTLE_ENABLED=true
       # - ES_HTTP_LOGGING=BODY
 
   dependencies:
```

codefromthecrypt:
ok only about half dropped now with throttling on according to stats.. :)

92458/165822

but.. the actually succeeded according to ES
```
curl -X POST  -s localhost:9200/zipkin-span-2019-08-21/_count|jq .count
165822
```

I wonder if these timeouts are actually succeeding in the end

codefromthecrypt:
same fail pattern on 2.15.0 vs this branch with throttling on (throttling on disables the semaphore in 2.15). What's different is 2.15.0 uses half the heap for what's within spitting distance of the same load.

Screenshot 2019-08-21 at 1 42 52 PM

I think what I need to do next, is compare this branch against master (the commit before) vs comparing against 2.15.0.

I do believe this is better at latency on the post endpoint. I'm not sure if memory is different or not vs last commit. Looking into this now.

codefromthecrypt:
hah lol.. jitpack hates master right now :)

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  02:49 min
[INFO] Finished at: 2019-08-21T01:10:00Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal com.mycila:license-maven-plugin:3.0:check (default) on project zipkin-server: Some files do not have the expected license header -> [Help 1]
```

lemme fix that

codefromthecrypt:
here's running off latest master (I verified the commit)

Similar numbers considering the curiousity of throttling vs timeouts..

```bash
$ curl -X POST  -s localhost:9200/zipkin-span-2019-08-21/_count|jq .count
163769

{
  "counter.zipkin_collector.messages.http": 138,
  "counter.zipkin_collector.spans_dropped.http": 86797,
  "gauge.zipkin_collector.message_bytes.http": 94395,
  "counter.zipkin_collector.bytes.http": 59788050,
  "gauge.zipkin_collector.message_spans.http": 221,
  "counter.zipkin_collector.spans.http": 163769,
}
```

Screenshot 2019-08-21 at 1 22 09 PM

codefromthecrypt:
I realize when I last ran off this branch maybe I didn't select the max heap used. I'll run again..

codefromthecrypt:
@anuraaga I can say confidently this improves measurably vs last commit under the same load. Please proceed! I'll open other issues about the curiosities

Screenshot 2019-08-21 at 1 36 41 PM

codefromthecrypt:
meanwhile I'll run against 2.15.0 to correct any misstatement about max heap above

codefromthecrypt:
updated https://github.com/openzipkin/zipkin/pull/2761#issuecomment-523277546 

so yeah in summary. this improves the current code a lot, and worth landing so we can reduce impact to those who already upgraded to 2.16

I think it is fair to concede we are still consuming a lot more memory, and causing a lot more GC than 2.15 for the same load. Many sites won't notice this I suspect, and we can continue to look for ways to get nearer to 2.15 levels of overhead.

anuraaga:
I have reworked the interface to remove a lot of the weirdness like the semantics of `close()` and `clone()`. I didn't think too much on what it would look like with abstract Call objects but feel fairly satisfied with the API now that those difficulties are gone.

codefromthecrypt:
taking a look. thanks!

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/BulkCallBuilder.java
```diff
@@ -117,51 +117,70 @@ public BulkCallBuilder(ElasticsearchStorage es, float esVersion, String tag) {
...
+    }
+
+    @Override public RequestHeaders headers() {
+      return headers;
+    }
+
+    @Override public void writeBody(HttpCall.RequestStream requestStream) {
+      for (IndexEntry<?> entry : entries) {
+        if (!requestStream.tryWrite(HttpData.wrap(serialize(alloc, entry, shouldAddType)))) {
+          // Stream aborted, no need to serialize anymore.
```
codefromthecrypt, line 154: great comment

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
```diff
@@ -56,6 +61,65 @@
...
+     * aborted (e.g., the request timed out while writing), or {@code true} otherwise.
+     */
+    boolean tryWrite(HttpData obj);
+  }
+
+  /**
+   * A supplier of {@linkplain HttpHeaders headers} and {@linkplain HttpData body} of a request
+   * to Elasticsearch.
+   */
+  public interface RequestSupplier {
```
codefromthecrypt, line 78: Napoleon's Corporal understands what these interfaces are for. thanks!

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
```diff
@@ -188,10 +248,10 @@ V parseResponse(AggregatedHttpResponse response, BodyConverter<V> bodyConverter)
       if (status.codeClass().equals(HttpStatusClass.SUCCESS)) {
         return null;
       } else if (status.code() == 404) {
-        throw new FileNotFoundException(request.path());
+        throw new FileNotFoundException(request.headers().path());
```
codefromthecrypt, line 250: the fact that armeria headers includes path always confuses me. I wonder if `requestLine()` would have been in hindsight a better term

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/internal/client/HttpCallTest.java
```diff
@@ -58,41 +57,36 @@
...
-    new AtomicReference<>();
   private static final AggregatedHttpResponse SUCCESS_RESPONSE =
     AggregatedHttpResponse.of(HttpStatus.OK);
 
-  @ClassRule public static ServerRule server = new ServerRule() {
-    @Override protected void configure(ServerBuilder sb) {
-      sb.service("/", ((ctx, req) -> HttpResponse.of(MOCK_RESPONSE.get())));
-    }
-  };
+  @RegisterExtension static MockWebServerExtension server = new MockWebServerExtension();
```
codefromthecrypt, line 66: thanks for changing this

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/internal/client/HttpCallTest.java
```diff
@@ -270,6 +264,25 @@
         "Fielddata is disabled on text fields by default. Set fielddata=true on [spanName] in order to load fielddata in memory by uninverting the inverted index. Note that this can however use significant memory. Alternatively use a keyword field instead.");
   }
 
+  @Test void streamingContent() throws Exception {
```
codefromthecrypt, line 267: I love this test

codefromthecrypt (approved):
I think this is merge-town

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/internal/client/HttpCallTest.java
anuraaga: I don't - added checks to make sure `headers()` is asserted too so maybe now love-worthy :)

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/client/HttpCall.java
anuraaga: Request line is something like `POST /`. I'm not sure if users usually deal with the combined request line like that, method and path are often separate, like in the servlet API - https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletRequest.html

Is the confusion because you would expect headers to only be headers, without the request line? That makes some sense, but not sure I want to spend brainpower thinking too much about it ;)

PR #2765, opened 2019-08-20
make storage links point to server storage configuration

jorgheymans:
The links to the storage components on https://github.com/openzipkin/zipkin should not point to the respective modules in zipkin-storage, but rather to the sections in https://github.com/openzipkin/zipkin/tree/master/zipkin-server that explain how to configure said storage components. And from there on there is the link to the storage modules. That way you go from zipkin high level storage options, then how-to-configure said storage options, then finally how does the storage option work technically. Most ppl don't care about the last bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ is for testing, for example starting a server on your laptop without any
 database needed.
 
 ### Cassandra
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
excellent choice

PR #2766, opened 2019-08-20
make in-memory storage component more visible

jorgheymans:
... it's mentioned on the main README.md but not here. Text mostly grabbed from https://github.com/openzipkin/zipkin/blob/b84bb3261d8402507a2d54f3b55e017484c50a85/zipkin-server/src/main/resources/zipkin-server-shared.yml#L81

### DIFF
--- a/zipkin-server/README.md
+++ b/zipkin-server/README.md
@@ -155,7 +155,23 @@ Defaults to true
 * `STORAGE_TYPE`: SpanStore implementation: one of `mem`, `mysql`, `cassandra`, `elasticsearch`
 * `COLLECTOR_SAMPLE_RATE`: Percentage of traces to retain, defaults to always sample (1.0).
 * `AUTOCOMPLETE_KEYS`: list of span tag keys which will be returned by the `/api/v2/autocompleteTags` endpoint; Tag keys should be comma separated e.g. "instance_id,user_id,env"
-* `AUTOCOMPLETE_TTL`: How long in milliseconds to suppress calls to write the same autocomplete key/value pair. Default 3600000 (1 hr)
+* `AUTOCOMPLETE_TTL`: How long in milliseconds to suppress calls to write the same autocomplete key/value pair. Default 3600000 (1 hr) 
+
+### In-Memory Storage
+Zipkin's In-Memory Storage is the default storage component that is used when no other storage type is configured. By default it stores a maximum of 500000 spans. Oldest traces (and their spans) will be purged first when this limit is exceeded. 
+
+Memory wise, a safe estimate is to count with 1K of memory per span (assuming 2 annotations + 1 binary annotation), plus 100 MB for a safety buffer. These numbers are by no means fixed, you'll need to verify in your own environment and increase memory (-Xmx) if needed.
+
+Example usage:
+```bash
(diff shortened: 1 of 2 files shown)


zipkin-server/README.md
```diff
@@ -155,7 +155,23 @@ Defaults to true
 * `STORAGE_TYPE`: SpanStore implementation: one of `mem`, `mysql`, `cassandra`, `elasticsearch`
 * `COLLECTOR_SAMPLE_RATE`: Percentage of traces to retain, defaults to always sample (1.0).
 * `AUTOCOMPLETE_KEYS`: list of span tag keys which will be returned by the `/api/v2/autocompleteTags` endpoint; Tag keys should be comma separated e.g. "instance_id,user_id,env"
-* `AUTOCOMPLETE_TTL`: How long in milliseconds to suppress calls to write the same autocomplete key/value pair. Default 3600000 (1 hr)
+* `AUTOCOMPLETE_TTL`: How long in milliseconds to suppress calls to write the same autocomplete key/value pair. Default 3600000 (1 hr) 
+
+### In-Memory Storage
+Zipkin's In-Memory Storage is the default storage component that is used when no other storage type is configured. By default it stores a maximum of 500000 spans. Oldest traces (and their spans) will be purged first when this limit is exceeded. 
+
+Memory wise, a safe estimate is to count with 1K of memory per span (assuming 2 annotations + 1 binary annotation), plus 100 MB for a safety buffer. These numbers are by no means fixed, you'll need to verify in your own environment and increase memory (-Xmx) if needed.
```
```suggestion
Memory wise, a safe estimate is to count with 1K of memory per span (assuming one tag and no annotations), plus 100 MB for a safety buffer. These numbers are by no means fixed, you'll need to verify in your own environment and increase memory (-Xmx) if needed.
```

codefromthecrypt: the code has drifted.. we don't use v1 format internally and "2 annotations + 1 binary annotation" loosly associates to one normal span.. in fact maybe we should remove this section as the memory usage is probably wrong.

zipkin-server/README.md
```diff
@@ -155,7 +155,23 @@ Defaults to true
...
+
+Example usage:
+```bash
+$ java -jar zipkin.jar
+```
+You can override the maximum number of spans stored using the `--max-spans` application parameter:
```
codefromthecrypt, line 167: instead of the lengthy and probably incorrect above paragraph on memory, you could also just add a sentence like..

If you see out-of-memory errors, you may need to increase your heap size as well.

jorgheymans: done.

jorgheymans:
note detailed memory requirements were taken from the max-spans parameter documentation in application.yml, should i remove it there as well ?

https://github.com/openzipkin/zipkin/blob/b84bb3261d8402507a2d54f3b55e017484c50a85/zipkin-server/src/main/resources/zipkin-server-shared.yml#L81

codefromthecrypt:
note detailed memory requirements were taken from the max-spans parameter
> documentation in application.yml, should i remove it there as well ?
>

yeah better to be honest that with this impl we don't know sizing details
until someone recollects the info

codefromthecrypt:
thanks much

PR #2770, opened 2019-08-21
Add enabled variable for collectors

### LINKED ISSUE #2762: Add `zipkin.collector.abc.enabled` conditional property to enable collectors
**Feature:**

Add conditional property to enable collectors, similar to Scribe, HTTP or Grpc

**Rational**

Currently, most messaging collectors are enabled based on env. variables defined (e.g. Kafka collector enabled when KAFKA_BOOTSTRAP_SERVERS is defined).
 

**Example Scenario**

While developing a [Kafka-based storage](https://github.com/jeqo/zipkin-storage-kafka) I would like to reuse variable names (KAFKA_xxx) without forcing users to enable Kafka Collector. 
Is not probable for this to happen when RabbitMQ or ActiveMQ are used as collector, but might be good to standardize the same property to enable collectors, as we do with `zipkin.storage.type`. 

**Prior Art**
 
* HTTP: <https://github.com/openzipkin/zipkin/blob/b84bb3261d8402507a2d54f3b55e017484c50a85/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java#L59>
* Scribe: <https://github.com/openzipkin/zipkin/blob/b84bb3261d8402507a2d54f3b55e017484c50a85/zipkin-server/src/main/java/zipkin2/server/internal/scribe/ZipkinScribeCollectorConfiguration.java#L31>

jeqo:
Fixes #2762 

Changes:

- Add enabled variable for messaging collectors.
- Standardize env variables to enable collectors: `COLLECTOR_XXX_ENABLED` (this require changing variables for SCRIBE and HTTP)

### DIFF
--- a/zipkin-server/README.md
+++ b/zipkin-server/README.md
@@ -329,7 +329,7 @@ correspond to "New Consumer Configs" in [Kafka documentation](https://kafka.apac
 
 Variable | New Consumer Config | Description
 --- | --- | ---
-`COLLECTOR_KAFKA_ENABLED` | N\A | `true` enables the Kafka collector. Defaults to `false`.
+`COLLECTOR_KAFKA_ENABLED` | N/A | `true` enables the Kafka collector. Defaults to `false`.
 `KAFKA_BOOTSTRAP_SERVERS` | bootstrap.servers | Comma-separated list of brokers, ex. 127.0.0.1:9092. No default
 `KAFKA_GROUP_ID` | group.id | The consumer group this process is consuming on behalf of. Defaults to `zipkin`
 `KAFKA_TOPIC` | N/A | Comma-separated list of topics that zipkin spans will be consumed from. Defaults to `zipkin`
--- a/zipkin-server/README.md
+++ b/zipkin-server/README.md
@@ -288,10 +288,10 @@ The HTTP collector supports the following configuration:
 
 Property | Environment Variable | Description
 --- | --- | ---
-`zipkin.collector.http.enabled` | `HTTP_COLLECTOR_ENABLED` | `false` disables the HTTP collector. Defaults to `true`.
+`zipkin.collector.http.enabled` | `COLLECTOR_HTTP_ENABLED` | `false` disables the HTTP collector. Defaults to `true`.
 
 ### Scribe (Legacy) Collector
-A collector supporting Scribe is enabled when `SCRIBE_ENABLED=true`. New
+A collector supporting Scribe is enabled when `COLLECTOR_SCRIBE_ENABLED=true`. New
 sites are discouraged from using this collector as Scribe is an archived
 technology.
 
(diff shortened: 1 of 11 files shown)


zipkin-server/src/main/resources/zipkin-server-shared.yml
```diff
@@ -26,11 +27,12 @@ zipkin:
       password: ${ACTIVEMQ_PASSWORD:}
     http:
       # Set to false to disable creation of spans via HTTP collector API
-      enabled: ${HTTP_COLLECTOR_ENABLED:true}
+      enabled: ${COLLECTOR_HTTP_ENABLED:true}
```
jeqo: This is part of a proposal to have enabled variables under `COLLECTOR_XXX_ENABLED`

zipkin-server/src/main/resources/zipkin-server-shared.yml
```diff
@@ -53,7 +56,7 @@ zipkin:
       useSsl: ${RABBIT_USE_SSL:false}
       uri: ${RABBIT_URI:}
     scribe:
-      enabled: ${SCRIBE_ENABLED:false}
+      enabled: ${COLLECTOR_SCRIBE_ENABLED:false}
```
jeqo: Same as HTTP

zipkin-server/src/main/resources/zipkin-server-shared.yml
codefromthecrypt: sadly we'll need to nest the old variable. I forget the exact syntax, but it can be done. Then update the READMEs

shakuzen: Not sure which we want to give precedent to if both exist, but the `:` is the "else" operator here, so giving precedent to the old one should be like `${HTTP_COLLECTOR_ENABLED:${COLLECTOR_HTTP_ENABLED:true}}`

jeqo: Thanks! Adding that now.

zipkin-server/README.md
```diff
@@ -302,11 +302,12 @@ Environment Variable | Property | Description
...
 
 ### ActiveMQ Collector
-The [ActiveMQ Collector](../zipkin-collector/activemq) is enabled when `ACTIVEMQ_URL` is set to a v5.x broker. The following
-settings apply in this case.
+The [ActiveMQ Collector](../zipkin-collector/activemq) is enabled when `ACTIVEMQ_URL` is set to a v5.x broker and 
+`COLLECTOR_ACTIVEMQ_ENABLED` is true. The following settings apply in this case.
 
 Environment Variable | Property | Description
 --- | --- | ---
+`COLLECTOR_ACTIVEMQ_ENABLED` | `zipkin.collector.activemq.enabled` | `true` enables the ActiveMQ collector. Defaults to `false`.
```
codefromthecrypt: hmmm I don't think we want to do this :) this means you have to set 2 properties where before only 1.

I think we want something like XXX will start collecting from blah, unless XXXX_ENABLED is set to false.

jeqo: Rolled back :) and added `enabled` to conditional evaluation.

All collectors will be "enabled=true", but actually enabled as before when url/bootstrap-servers is defined. Then if `enabled=false`, it will be disabled.

codefromthecrypt:
great. I noticed only some formatting fuzz. will merge after I scrub.

PR #2771, opened 2019-08-21
Bumps version of jackson-databind lib to tackle CVE-2019-14379

gregoriomelo:
Similar to https://github.com/openzipkin/zipkin-dependencies/pull/145

The version of jackson-databind in use, 2.9.9, is said to be vulnerable. The version 2.9.9.2 has fixed this issue. I've picked 2.9.9.3, the latest patch in the minor version. Here's the description of the CVE: https://nvd.nist.gov/vuln/detail/CVE-2019-14379

As I'm new to the project, is there a quick way we can assure this doesn't break core? Otherwise, please let me know how else I can manually integration-test this patch.

Cheers,
Greg

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -100,6 +100,7 @@
     <maven-failsafe-plugin.version>3.0.0-M3</maven-failsafe-plugin.version>
     <maven-enforcer-plugin.version>3.0.0-M2</maven-enforcer-plugin.version>
     <git-commit-id.version>3.0.0</git-commit-id.version>
+    <jackson-databind.version>2.9.9.3</jackson-databind.version>
   </properties>
 
   <name>Zipkin (Parent)</name>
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
thanks! similar comments to zipkin-dependencies.

Here are the three things that use it
* kafka has this change, just not released yet. https://github.com/apache/kafka/pull/7125
* I presume the next patch of spring boot has this due to a custom jackson version https://github.com/spring-projects/spring-boot/blob/2.1.x/spring-boot-project/spring-boot-dependencies/pom.xml#L84
* armeria doesn't seem to quite yet have this https://github.com/line/armeria/blob/master/dependencies.yml#L20

Can you help raise a ticket for armeria? Meanwhile I can also see if we can remove what may be an accidental indirect dep on spring-boot-starter-json

We can override the version anyway, but we should revert this when our indirect deps resolve at or past this.

PR #2777, opened 2019-08-24
Falls back to simple name and stops prefixing rejected ES exceptions

codefromthecrypt:
Armeria has a number of constant throwables, which neither have a stack
trace, nor a message. This switches to use these while we figure out if
there is a better way to unveil an underlying http connection problem
the health-check sees, but a health-checked client would not.

The approach is used all over in zipkin, most notably Brave.

This also stops prefixing RejectionException messages also with the word
as it is annoying to read it twice in toString.

Most importantly, this unwraps RejectionException in health check as it
doesn't effectively describe the cause of the problem, rather obfuscates
it.

Before:
```bash
curl -s localhost:9411/health
{"status":"DOWN","zipkin":{"status":"DOWN","details":{"ElasticsearchStorage{initialEndpoints=https://localhost:9200, index=zipkin}":{"status":"DOWN","details":{"error":"java.util.concurrent.RejectedExecutionException: Rejected execution: null"}}}}}
```

Now:
```bash
curl -s localhost:9411/health
{"status":"DOWN","zipkin":{"status":"DOWN","details":{"ElasticsearchStorage{initialEndpoints=https://localhost:9200, index=zipkin}":{"status":"DOWN","details":{"error":"com.linecorp.armeria.common.ClosedSessionException"}}}}}
```

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHealthIndicator.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHealthIndicator.java
@@ -41,7 +41,13 @@ final class ZipkinHealthIndicator extends CompositeHealthIndicator {
     /** synchronized to prevent overlapping requests to a storage backend */
     @Override public synchronized Health health() {
       CheckResult result = component.check();
-      return result.ok() ? Health.up().build() : Health.down((Exception) result.error()).build();
+      if (result.ok()) return Health.up().build();
+      Throwable ex = result.error();
+      // Like withException, but without the distracting ": null" when there is no message.
+      String message = ex.getMessage();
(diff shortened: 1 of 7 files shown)


peacejj:
This is an exception after requesting the webpage. Do you know how to solve it??

2020-02-06 14:21:54.094 [armeria-common-blocking-tasks-1-9023] WARN  zipkin2.server.internal.BodyIsExceptionMessage - Unexpected error handling request.
com.linecorp.armeria.common.ClosedSessionException: null

codefromthecrypt:
@peacejj this is usually when there is a connection or latency problem. please use gitter for troubleshooting https://gitter.im/openzipkin/zipkin

PR #2779, opened 2019-08-24
Fixes thread safety bug on in-memory storage regarding accept count

codefromthecrypt:
Before, we were using an unsafe means to increment a counter for the
in-memory storage. This contributed to a test flake. This fixes the
state bug and also tightens up the flakey test.

### DIFF
--- a/zipkin-server/src/test/java/zipkin2/server/internal/brave/ITZipkinSelfTracing.java
+++ b/zipkin-server/src/test/java/zipkin2/server/internal/brave/ITZipkinSelfTracing.java
@@ -22,6 +22,7 @@ import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.Response;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


anuraaga (approved):
Code and test code LGTM

codefromthecrypt:
so still flaking on one endpoint which isn't returning a span for the storage layer..

```json
  [
    {
      "traceId": "7dddc296c13a5448",
      "id": "7dddc296c13a5448",
      "kind": "SERVER",
      "name": "post /api/v1/spans",
      "timestamp": 1566632878396438,
      "duration": 14240,
      "localEndpoint": {
        "ipv4": "172.17.0.1"
      },
      "remoteEndpoint": {
        "ipv4": "127.0.0.1",
        "port": 57474
      },
      "annotations": [
        {
          "timestamp": 1566632878396451,
          "value": "wr"
        },
        {
          "timestamp": 1566632878410664,
          "value": "ws"
        }
      ],
      "tags": {
        "http.method": "POST",
        "http.path": "/api/v1/spans"
      }
    }
  ]
```

PR #2782, opened 2019-08-24
Catch and return all possible errors in collector.

anuraaga:
Found this because lately my IDE-run zipkin server doesn't accept JSON spans due to gson classload error for some reason but this would happen silently. But anyways, we need to make sure we propagate any error into the callback.


codefromthecrypt (approved):
good idea

PR #2786, opened 2019-08-29
Add an integrated benchmark runner.

anuraaga:
The benchmark starts up containers similar to docker-compose, along with a container for wrk, and runs requests. At the end, it dumps out some prometheus values and can optionally keep the containers alive until manual termination to take a look at Grafana by setting the environment variable `ZIPKIN_BENCHMARK_WAIT`.

By default, the zipkin server is the one for the current active code and run in-process. An environment variable `ZIPKIN_VERSION` can be set to use a released binary instead, in which case zipkin server will run in a docker container too. 

Final output looks like

```
Benchmark started.
Container openzipkin/example-sleuth-webmvc:latest ports exposed at {9000=http://localhost:33103}
Container openzipkin/example-sleuth-webmvc:latest ports exposed at {8081=http://localhost:33106}
Container prom/prometheus:latest ports exposed at {9090=http://localhost:33107}
Container grafana/grafana:latest ports exposed at {3000=http://localhost:33108}
Benchmark complete, wrk output:
Running 2m test @ http://frontend:8081
  4 threads and 128 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   604.32ms  305.09ms   1.99s    67.53%
    Req/Sec    54.02     25.81   161.00     66.88%
  Latency Distribution
     50%  598.18ms
     75%  800.31ms
     90%    1.00s 
     99%    1.37s 
  21190 requests in 1.67m, 2.87MB read
  Socket errors: connect 0, read 0, write 0, timeout 27
Requests/sec:    211.73
Transfer/sec:     29.38KB

WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by io.netty.resolver.dns.DnsNameResolver (file:/C:/Users/Anuraag/.m2/repository/io/netty/netty-resolver-dns/4.1.39.Final/netty-resolver-dns-4.1.39.Final.jar) to method sun.net.dns.ResolverConfiguration.open()
WARNING: Please consider reporting this to the maintainers of io.netty.resolver.dns.DnsNameResolver
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
Messages received: 173
Spans received: 63954
Spans dropped: 0
Memory quantiles:
jvm_memory_used_bytes{area="heap"}[0.0] = 6291456
jvm_memory_used_bytes{area="heap"}[0.25] = 8388608
jvm_memory_used_bytes{area="heap"}[0.5] = 10485760
jvm_memory_used_bytes{area="heap"}[0.75] = 66060288
jvm_memory_used_bytes{area="heap"}[1.0] = 121634816
jvm_memory_used_bytes{area="nonheap"}[0.0] = 1359232
jvm_memory_used_bytes{area="nonheap"}[0.25] = 4740736
jvm_memory_used_bytes{area="nonheap"}[0.5] = 8130368
jvm_memory_used_bytes{area="nonheap"}[0.75] = 17668608
jvm_memory_used_bytes{area="nonheap"}[1.0] = 67858656
Total GC time (s): 0.14300000000000002
Number of GCs: 5
POST Spans latency (s)
http_server_requests_seconds_bucket{method="POST",status="202",uri="/api/v2/spans"}[0.5] = 0.0005029069767441861
http_server_requests_seconds_bucket{method="POST",status="202",uri="/api/v2/spans"}[0.9] = 0.000905232558139535
http_server_requests_seconds_bucket{method="POST",status="202",uri="/api/v2/spans"}[0.99] = 0.0009957558139534885
```


benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
```diff
@@ -0,0 +1,301 @@
...
+    LOGGER.info("Memory quantiles:");
+    printQuartiles(prometheusClient, "jvm_memory_used_bytes{area=\"heap\"}");
+    printQuartiles(prometheusClient, "jvm_memory_used_bytes{area=\"nonheap\"}");
+
+    LOGGER.info("Total GC time (s): {}",
+      prometheusValue(prometheusClient, "sum(jvm_gc_pause_seconds_sum)"));
+    LOGGER.info("Number of GCs: {}",
+      prometheusValue(prometheusClient, "sum(jvm_gc_pause_seconds_count)"));
+
+    LOGGER.info("POST Spans latency (s)");
```
codefromthecrypt: is the logger helping us? seems to add a bunch of redundant formatting :)

codefromthecrypt (commented):
looks neat. haven't tried yet

@marcingrzejszczak I think you had a jenkins job with perf overhead numbers.. was there anything keeping track of those over time? or was it just something to look at

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
anuraaga: That's true - switched to println

codefromthecrypt:
cc @bsideup I think you might be interested

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
```diff
@@ -0,0 +1,302 @@
...
+    startContainer(backend);
+    startContainer(frontend);
+
+    String prometheusResource = RELEASED_ZIPKIN_VERSION == null
+      ? "prometheus-local.yml" : "prometheus-container.yml";
+    GenericContainer<?> prometheus = new GenericContainer<>("prom/prometheus")
+      .withNetwork(Network.SHARED)
+      .withNetworkAliases("prometheus")
+      .withExposedPorts(9090)
+      .withCopyFileToContainer(
```
bsideup: `withClasspathResourceMapping` should work too

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
```diff
@@ -0,0 +1,302 @@
...
+      .waitingFor(new HttpWaitStrategy().forPath("/actuator/health"));
+
+    if (RELEASED_ZIPKIN_VERSION == null) {
+      Testcontainers.exposeHostPorts(9411);
+      String zipkinArg = "-Dspring.zipkin.baseUrl=http://host.testcontainers.internal:9411";
+      backend.withEnv("JAVA_OPTS", zipkinArg);
+      frontend.withEnv("JAVA_OPTS", zipkinArg);
+    }
+
+    startContainer(backend);
```
bsideup: FYI in Testcontainers 1.12.x, there is a helper method to start containers in parallel (including their transitive dependencies):
```
Startables.deepStart(Stream.of(container1, container2)).join()
```

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
```diff
@@ -0,0 +1,302 @@
...
+
+    GenericContainer<?> frontend = new GenericContainer<>("openzipkin/example-sleuth-webmvc")
+      .withNetwork(Network.SHARED)
+      .withNetworkAliases("frontend")
+      .withCommand("frontend")
+      .withExposedPorts(8081)
+      .waitingFor(new HttpWaitStrategy().forPath("/actuator/health"));
+
+    if (RELEASED_ZIPKIN_VERSION == null) {
+      Testcontainers.exposeHostPorts(9411);
```
bsideup: keep in mind that this feature starts an SSH port forwarding session and may affect the results of the benchmark (since traffic will be proxied, in not ultra performant way :D)

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
```diff
@@ -0,0 +1,302 @@
...
+
+  void runBenchmark(@Nullable GenericContainer<?> storage) throws Exception {
+    GenericContainer<?> zipkin = startZipkin(storage);
+
+    GenericContainer<?> backend = new GenericContainer<>("openzipkin/example-sleuth-webmvc")
+      .withNetwork(Network.SHARED)
+      .withNetworkAliases("backend")
+      .withCommand("backend")
+      .withExposedPorts(9000)
+      .waitingFor(new HttpWaitStrategy().forPath("/actuator/health"));
```
bsideup: Just FYI: `Wait.forHttp("/actuator/health")`

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
```diff
@@ -0,0 +1,302 @@
...
+
+  static final boolean WAIT_AFTER_BENCHMARK = "true".equals(System.getenv("ZIPKIN_BENCHMARK_WAIT"));
+
+  Closer closer;
+
+  @BeforeEach void setUp() {
+    closer = Closer.create();
+  }
+
+  @AfterEach void tearDown() throws Exception {
```
bsideup: I would highly recommend to explicitly destroy the containers after the run. There is also an undocumented way to trigger the cleanup:
```java
ResourceReaper.instance().performCleanup();
```

Although Testcontainers will cleanup them later on, the deferred cleanup may affect the results of the next iteration

bsideup:
@adriancole nice! Commented a few things, but I like the idea 👍
Next step - [chaos testing](https://www.testcontainers.org/modules/toxiproxy/) of Zipkin? ;)

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
anuraaga: The closer here is used to close the containers. Is `GenericContainer.stop` not enough to clean up?

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
anuraaga: Yeah it's definitely something to keep in mind, especially when comparing a docker version to in-process one. The numbers I'm seeing don't bother me too much.

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
anuraaga: Yeah I initially used that but had an error like "already mounted" or something once on Windows. Since these files are static copying seems less tricky than binding.

anuraaga (commented):
Thanks for the tips will apply them!

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
bsideup: My bad, I didn't catch that the closer is for containers :) Yes, `stop`/`close` is enough 👍 

Although, the `ResourceReaper`-based suggestion still applies, and can simplify the thing (no need to register containers to `Closer`), but that's up to you, shouldn't make a lot of difference

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
bsideup: Ouch.... Actually, I just recall that I submitted https://github.com/testcontainers/testcontainers-java/issues/1163 but we never did the change, but we should :)

Yes, copy is better here :) Meanwhile, I will make sure that https://github.com/testcontainers/testcontainers-java/issues/1163 is fixed soon :)

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
anuraaga: Realized I could copy the classpath into a Java ontainer, so now even for in-development code, the server is run in docker and the issue goes away. Having a method in testcontainers to start a container with the current classpath might be a very nice feature :)

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
```diff
@@ -0,0 +1,315 @@
...
+      .withExposedPorts(9200)
+      .waitingFor(new HttpWaitStrategy().forPath("/_cluster/health"));
+    startContainer(elasticsearch);
+    runBenchmark(elasticsearch);
+  }
+
+  @Test void cassandra() throws Exception {
+    GenericContainer<?> cassandra = new GenericContainer<>("openzipkin/zipkin-cassandra")
+      .withNetwork(Network.SHARED)
+      .withNetworkAliases("cassandra3")
```
codefromthecrypt: note: we use "cassandra" in docker compose because this works for any of our cassandra storage types. prefer parity with the docker-compose files when naming etc.

codefromthecrypt (approved):
I presume my fan will stop and all work.

only suggested change is to parameterize cassandra into 2 test runs, by an optional parameter which is then used to override the container name. This would only be used to run "cassandra3" against "cassandra"


background: we still have sites that run "cassandra" storage type because "cassandra3" requires the latest or next to latest patch of cassandra to work. This implies a lot of commerical variants or clones of cassandra don't work with zipkin unless using "cassandra" type.

By adding another parameter that in simple case falls back to the container name, you can bench all of our supported types.

benchmarks/src/main/resources/log4j2.properties
```diff
@@ -0,0 +1,22 @@
...
+# or implied. See the License for the specific language governing permissions and limitations under
+# the License.
+#
+
+appenders=console
+appender.console.type=Console
+appender.console.name=STDOUT
+appender.console.layout.type=PatternLayout
+appender.console.layout.pattern=%d{ABSOLUTE} %-5p [%t] %C{2} (%F:%L) - %m%n
+rootLogger.level=info
```
codefromthecrypt: now that we are using normal stdout, we don't need all the spam that comes with info

```suggestion
rootLogger.level=warn
```

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
```diff
@@ -0,0 +1,315 @@
...
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.Network;
+import org.testcontainers.containers.wait.strategy.HttpWaitStrategy;
+import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
+import org.testcontainers.utility.MountableFile;
+
+class ServerIntegratedBenchmark {
+
+  static final Logger LOGGER = LoggerFactory.getLogger(ServerIntegratedBenchmark.class);
```
codefromthecrypt: no longer used/

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
```diff
@@ -0,0 +1,315 @@
...
+import org.junit.jupiter.api.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.Network;
+import org.testcontainers.containers.wait.strategy.HttpWaitStrategy;
+import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
+import org.testcontainers.utility.MountableFile;
+
+class ServerIntegratedBenchmark {
```
codefromthecrypt: maybe add a comment on top why this is in src/main I presume you are trying to ensure it isn't executed during the normal test runs. It might be nicer on the benchmark classpath to move it and its deps to src/test and just have it not run by default. just a thought.

codefromthecrypt:
sorry was going to address own feedback but got distracted..

anuraaga:
Think I addressed all the comments. I changed the log level back to `info` if it's ok - I find the `info` messages showing container pulling / startup to be fairly useful. On the flip side, at least on my computer it spams an `error` log about not finding a docker credential helper and falling back to docker-java, which shows even when that's set to `warn`.

anuraaga:
By he way, it sounds like you tried it out - I'm guessing on a Mac OS X? Any gotchas? I have to restart docker before a run of benchmarks or I get no prometheus metrics since docker time gets out of sync when my laptop sleeps. I'm wondering if this only affects Windows

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
anuraaga: Thanks for this tip - the containers start up way faster thanks to this :D

codefromthecrypt:
building was fine on os/x

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
bsideup: This sounds more like Arquillian :)

I am not sure such module will be easy to maintain because I expect a lot of hidden gems (module path alone makes me feel scared), but sounds like a nice idea for a library **on top** of Testcontainers 👍

codefromthecrypt:
as this seems to crash when run in sequence one way or another (timeout, never respond etc) I think either we get to the bottom of it, or don't offer to run sequentially multiple configurations.

The first passes, the next dies either just spinning or a timeout.

anuraaga:
Will check it out think I repro that too. It didn't happen yesterday so something in the meantime made it worse so just need to find it.

benchmarks/src/main/java/zipkin2/it/ServerIntegratedBenchmark.java
anuraaga: I tried using this but unfortnately it didn't work  -`ResourceReaper.instance().registeredContainers` is empty. Not sure if it's because of using `deepStart` or not but this would have been a nice shortcut given the parallel containers, sticking with manual for now though.

anuraaga:
When migrating to parallel container startup, I lost the code to close the containers. With it back, the benchmarks can run consecutively without crashing.

codefromthecrypt:
good hunting!

marcingrzejszczak:
> @marcingrzejszczak I think you had a jenkins job with perf overhead numbers.. was there anything keeping track of those over time? or was it just something to look at

There's a JMH plugin that can store previous build information and then it can break a build if your performance has dropped in comparison to that build.

Issue #2788, opened 2019-08-30
Directly support endpoints we use in support vs implicit redirects to Spring Boot Actuator


codefromthecrypt:
In former versions we mounted /health and /prometheus at root. I think starting with spring boot 2 or armeria, I forget which, we redirected instead of forwarding (dual-mounting) these paths.

I think this dependency is unnecessary and requires us to use one of the heftiest parts of spring boot for routine work (actuator) actuator has a lot of slow paths and is itself bulky and ultimately built on servlet layers. If instead we mount the required urls /prometheus /info and /health directly, we have flexibility in terms of certain dependencies and to increase our startup time as well memory usage.

So, I prefer we directly mount the following

* /health
* /info
* /prometheus

cc @openzipkin/core

codefromthecrypt:
the other thing to consider is that our http runtime doesn't support all the things actuator expects anyway (ex spring-security), by directly mounting these, we have a much more honest relationship, too. Finally anyone wanting to clone the server can just implement these and env variables  vs trying to re-implement all of spring boot. Even if we eventually unhook actuator, we can always add an optional module to add it back for sites who really want it.

codefromthecrypt:
here's some examples of investigations opting out of actuator

https://github.com/dsyer/spring-boot-memory-blog#summary-of-data
https://github.com/dsyer/spring-boot-startup-bench#actuator

shakuzen:
I'm not sure I'm understanding the problem or proposed solution. We explicitly enable all Actuator endpoints because this was the behavior with Spring Boot 1 and we thought some users would expect these endpoints or find them helpful (even if their path changed). I know we have used the env/configprops endpoints for troubleshooting user-reported issues in the past, at least.

https://github.com/openzipkin/zipkin/blob/9885c8cc6ec302f7689047d17833d7c362f6dbd8/zipkin-server/src/main/resources/zipkin-server-shared.yml#L237-L241

We can limit that to a subset of endpoints if we don't want other endpoints because of some perceived cost.

Actuator does not require Spring Security - it's an optional dependency if you want to add security features it provides.

[...]

PR #2789, opened 2019-08-30
Use Armeria to expose prometheus instead of actuator.

anuraaga:
This doesn't remove the fact that actuator sets up the prometheus collector, but it's a quick start.

For #2788

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
@@ -26,6 +26,7 @@ import io.micrometer.core.instrument.config.MeterFilter;
 import io.prometheus.client.CollectorRegistry;
 import java.time.Duration;
 import java.util.List;
+import java.util.Optional;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanFactory;
 import org.springframework.beans.factory.BeanFactoryAware;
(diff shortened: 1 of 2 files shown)


zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
```diff
@@ -75,7 +77,7 @@
       if (httpCollector != null) sb.annotatedService(httpCollector);
       if (healthController != null) sb.annotatedService(healthController);
       // Redirects the prometheus scrape endpoint for backward compatibility
-      sb.service("/prometheus", new RedirectService("/actuator/prometheus"));
+      sb.service("/prometheus", new PrometheusExpositionService(prometheusRegistry));
```
codefromthecrypt: If I understand correctly, we just have now 2 implementations of /prometheus. actuator is handling the other one

anuraaga: Yeah - I think that's what #2788 meant but maybe I misunderstood :)

anuraaga: Ah think I missed latest comment - so you think we should use this service on `/actuator/prometheus` too?

codefromthecrypt: well right now we aren't yanking actuator, but when we get around to it, we should mount it. that or disable the autoconfig and mount it now.

codefromthecrypt:
ps build broke didn't look at it. will be offline mostly next 36hrs

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
anuraaga: Went ahead and disabled the actuator/prometheus endpoint and mount it now.

codefromthecrypt:
great. thanks rag!

PR #2790, opened 2019-08-30
Add a Dockerfile for building a master image of zipkin-server.

anuraaga:
This will be referenced from docker hub to begin automated building of a master docker image of zipkin-server. This solves https://github.com/openzipkin/docker-zipkin/issues/170 which also contains a good summary of the rationale - it's great for previewing development features and testing new infra to be able to just start up a docker image with the latest code, no `mvn` or jitpack required.

This copies configuration from https://github.com/openzipkin/docker-zipkin - these don't change much so I think copying temporarily should be fine. The intent is to first have master images for all the components that use the latest code, then later gradually migrate release builds off of quay to use these Dockerfiles.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,13 @@
+**/\.*
+!.git
+!**/.eslintrc
+
+**/target
+**/node_modules
+**/*.iml
+
+*.md
+
+**/Dockerfile
+travis
+mvnw*
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,41 @@
+#
+# Copyright 2015-2019 The OpenZipkin Authors
+#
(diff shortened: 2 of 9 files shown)


codefromthecrypt:
I think maybe add a README in the docker directory about migration in
progress and that this is a copy of the contents of the other repo. then
sgtm

devinsba:
I'm OK with this, if I had time I wonder if there is a way to make this happen from the other repo

anuraaga:
I think there's always a way with writing code :) But having the Dockerfile in the same repo is least indirect I think - you can always build a docker image as long as you have the code - currently we often push a branch, rewrite the Dockerfile to jitpack, and build from there just to have an image of development code.

We might run into different issues that the separate repo helps with, so we're currently going to copy configuration rather than moving things right away. If we directly compare both worlds, things will probably come into clarity.

PR #2791, opened 2019-08-30
custom servers are not supported !

jorgheymans:


README.md
```diff
@@ -73,8 +73,7 @@ Note: The above is just an example, most likely you'll want to use an existing t
 
 ## Storage Component
 Zipkin includes a [StorageComponent](zipkin/src/main/java/zipkin2/storage/StorageComponent.java), used to store and query spans and
-dependency links. This is used by the server and those making custom
-servers, collectors, or span reporters. For this reason, storage
+dependency links. This is used by the server and those making collectors, or span reporters. For this reason, storage
```
codefromthecrypt, line 76: "span reporters" is quite the edge case for the storage component. in hind sight I think we should be more cautious about this in general, perhaps remove it from the top-level readme.

I'm thinking move section about storage component to readme under zipkin-storage, saying the only use case is server-development and while we keep dependencies down etc, the storage and collector apis primarily are in service to server extensions and are not intended for people to nest into traced applications

jorgheymans:
Fully agree moving the storage component info out of the top level readme.
This is not what a first user is interested in.

README.md
shakuzen: Adrian's suggestion here sounds good to me. Better to get this out of the top-level README so people aren't tempted to try this kind of thing.

codefromthecrypt:
you have time to land this? @fbeaufume had some comments that make this such a relevant pull request. I suspect even after we land this folks will still do custom servers and not look at our docs that ask please do not. 

https://github.com/openzipkin/zipkin/blob/master/zipkin-server/src/main/java/zipkin/server/EnableZipkinServer.java#L25-L33

codefromthecrypt:
I think maybe we need to go from clear to aggressively clear

for example

"That zipkin uses spring boot at all is an implementation detail that can change at any time. We do not support custom servers outside our org. Anything else is completely use at your own risk"

cc @openzipkin/core

basvanbeek:
Suggesting minor changes...

"That Zipkin uses spring boot at all is solely an internal implementation detail that can change at any time. We do NOT support custom servers outside our org! Anything else is completely use at your own risk!!!"

jeqo:
I'd also suggest to add a header with anchor to link directly to.

jorgheymans:
@adriancole you can go ahead and apply this already. Rest of the doc patches will follow the next couple of days.

codefromthecrypt:
thanks @jorgheymans

PR #2794, opened 2019-09-01
Add a builder Docker image to use when building other images.

anuraaga:
Docker hub uses standard Docker layer caching to cache artifacts between builds. This is different from e.g., Travis, which mounts home directory folders from a cache it downloads / uploads out of line.

Currently, all of our builds start with a fresh `maven` image which has no cache at all, so every build downloads the world. We can prepare a `zipkin-builder` image which contains a package cache to speed up our downstream builds. After the first version of this image is pushed, I'll point the Dockerfiles to use it instead of `maven`. This builder image would have no value if code changes caused it's Docker cache key to be invalidated (the Docker cache key is the file hashes computed for a given COPY command, etc), so we use a special `.dockerignore` file to exclude as much as possible. Some of the linked issues in the comments provide more background on this approach (ideally COPY would accept inclusions / exclusions as arguments but there seems to be no desire to implement that in Dockerfile).

In this PR, I have added a `skipLens` property that can be used in the build to not build lens. It's used here, but I think it could be useful in general development since I know I've found myself waiting for a zipkin-lens build to finish despite not needing lens for the particular dev task.

The `zipkin-builder` image has a special `dockerignore` file that attempts to only copy in pom.xml and package.json files into the build context. This means that the result of a maven build is only invalidated when these change, which is rare. This means that there would be two scenarios:

1) A master commit that only contains code changes (no changes to pom or package.json). The zipkin-builder and zipkin builds would be started together. zipkin-builder would end without doing anything since it's cache key, which does not include code, is unaffected. zipkin will build and not download the world since it uses zipkin-builder's cached repositories.

2) A master commit that does change pom or package.json. The zipkin-builder and zipkin builds would be started together. zipkin-builder would start from scratch, downloading the world. zipkin will start and will use the latest, now outdated, zipkin-builder when building. This means it won't download the world, but only the new dependencies since the previous one. This is pretty fast, since I notice that the majority of the world is maven itself with all its plugins, and those don't change so often. After these builds finish, following commits would go back to scenario 1 until a next set of dependency updates.

Downstream projects zipkin-aws and zipkin-gcp will download more stuff on scenario 1. While it'd be possible to prepare builders for each of them, I think the delta in dependencies is small enough that we don't need to worry about that (the prime culprits, maven and spring-boot will be in zipkin-builder for use by all the buildz).

While it's a bit annoying having this extra image, once it's set up it should all manage itself automatically so hopefully it is not too complex.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -9,5 +9,6 @@
 *.md
 
 **/Dockerfile
-travis
+hooks/**
+travis/**
 mvnw*
--- a/docker/builder/Dockerfile
+++ b/docker/builder/Dockerfile
@@ -16,7 +16,7 @@ FROM maven:3-jdk-11-slim
 
 COPY . /code/
 
-# We build the project so all its dependencies get downloaded. There doesn't seem to be a good way
-# of only downloading dependencies without building too or we get unresolved dependency failures for
-# the local modules, but since we don't copy any code in the compilation itself is fast.
-RUN cd /code && mvn -B package -DskipTests=true && cd / && rm -rf /code
+RUN cd /code && \
+    mvn -B package -Dlicense.skip=true -DskipTests=true -DskipLens=true && \
+    mvn -B generate-resources -pl zipkin-lens -Dlicense.skip=true -DskipTests=true && \
+    cd / && rm -rf /code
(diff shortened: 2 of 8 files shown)


codefromthecrypt (approved):
As far as I understand this, it is mechanics to make the build/publish/depend flow cleaner and for end users there is no difference in images produced. Only thing and end user sees is an addition image in our namespace. I'm good with that.

PR #2800, opened 2019-09-13
Deletes code no longer needed regarding state checking of ES hosts

codefromthecrypt:
We no longer need two-tiered checking of ES addresses because single
endpoint is special cased to not cause notable overhead.

See https://github.com/line/armeria/issues/2071#issuecomment-531139475

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/LazyHttpClientImpl.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/LazyHttpClientImpl.java
@@ -70,32 +70,37 @@ final class LazyHttpClientImpl implements LazyHttpClient {
   }
 
   Endpoint getEndpoint() {
-    EndpointGroup rawEndpointGroup = initialEndpoints.get();
+    EndpointGroup initial = initialEndpoints.get();
     // Only health-check when there are alternative endpoints. There aren't when instanceof Endpoint
-    if (rawEndpointGroup instanceof Endpoint) return (Endpoint) rawEndpointGroup;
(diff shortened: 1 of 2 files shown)


zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/LazyHttpClientImpl.java
```diff
@@ -71,30 +70,30 @@
...
+      // block. The alternative is round-robin, which could be unlucky and hit a bad node first.
+      //
+      // We are blocking up to the connection timeout which should be enough time for any DNS
+      // resolution that hasn't happened yet to finish.
+      //
+      // TODO: When using health checks, if all hosts fail the health check, this will throw TOE
+      // See https://github.com/line/armeria/issues/2075
+      endpointGroup.awaitInitialEndpoints(timeoutMillis, TimeUnit.MILLISECONDS);
+    } catch (Exception e) {
+      // We'll try again next time around.
```
anuraaga: `endpointGroup.close()` - no-op if it's not a health checked group.

codefromthecrypt: thanks

anuraaga (approved):

codefromthecrypt:
@anuraaga I added a commit as we weren't looking for empty results before. This should make the change compatible with https://github.com/line/armeria/pull/2074

anuraaga (commented):
LGTM

codefromthecrypt:
Thanks @anuraaga!

PR #2803, opened 2019-09-16
Adds Traces api for retrieving multiple traces at the same time

codefromthecrypt:
In #2116, we added the ability to perform multi-get by trace ID in support of collecting trace IDs based on out-of-band discovery in one trip.

This resurrects work formerly completed by @zeagord and @llinder in such a way that it doesn't break compatibility in any way.

Instead of adding `getTraces(List)` to the `SpanStore` api, this adds `Traces` with a backported bridge to the `SpanStore` api.

### DIFF
--- a/zipkin/src/main/java/zipkin2/internal/TracesAdapter.java
+++ b/zipkin/src/main/java/zipkin2/internal/TracesAdapter.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2015-2018 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin2.internal;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import zipkin2.Call;
+import zipkin2.Span;
+import zipkin2.storage.SpanStore;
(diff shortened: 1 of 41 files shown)


codefromthecrypt:
Here's the api on this https://github.com/openzipkin/zipkin-api/pull/82

zipkin/src/main/java/zipkin2/storage/Traces.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Allows readback of traces by ID, as written by a {@link SpanConsumer}.
+ *
+ * <p>Specifically, this provides apis present when {@link StorageComponent.Builder#searchEnabled(boolean)
+ * search is disabled}.
+ *
+ * <p>Note: This is not considered a user-level Api, rather an Spi that can be used to bind
+ * user-level abstractions such as futures or observables.
+ */
```
codefromthecrypt, line 30: TODO: add since tag if we end up merging this.

zipkin/src/main/java/zipkin2/internal/TracesAdapter.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override public List<List<Span>> map(List<Span> input) {
+      return Collections.singletonList(input);
+    }
+
+    @Override public String toString() {
+      return "Collections.singletonList()";
+    }
+  }
+
+  static final class ScatterGather extends AggregateCall<List<Span>, List<List<Span>>> {
```
jeqo, line 60: looking forward to use this one :)

jeqo (approved):
This looks very good. I'd assume at some point we will have `Dependencies` also as an independent module, and we would need a way to have different back-end supporting different parts of the storage; am I following correctly?

codefromthecrypt:
@jeqo thanks for the feedback

> This looks very good. I'd assume at some point we will have Dependencies also as an independent module, and we would need a way to have different back-end supporting different parts of the storage; am I following correctly?

We can pull out Dependencies as another module, in fact we started towards that with links query. However, we don't have a complete use case likely to be merged about that. Anyway, I think it will eventually happen and can follow similarly to this.

Issue #2802, opened 2019-09-16
Pluggable span/trace handlers


codefromthecrypt:
from https://github.com/openzipkin/zipkin/pull/2072#issuecomment-474802294

> I'm inclined to pick this back up with a zipkin2.Span variant of https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/handler/FinishedSpanHandler.java#L30
> 
> Naming is a funny one, but I like CollectedSpanHandler at the moment. I would add a lot of javadoc mentioning limitations I said including that a "collected span" isn't necessarily a finished one (partial data can be here, such as a span holding nothing but a single tag). Later, you could imagine a CollectedTraceHandler. regardless, these types would be in the zipkin-collector jar.

So, this would result in `CollectedSpanHandler` and `CollectedTraceHandler`. As discussed also with @jeqo the latter we'd want an in-memory mechanism to trigger, which could be swapped out in integrations that have partitioned triggers like kafka-streaming, flink and voltdb.

cc @benjefferies @afalko

codefromthecrypt:
https://github.com/openzipkin/zipkin/pull/2807 single span version

PR #2807, opened 2019-09-18
Adds CollectedSpanHandler

codefromthecrypt:
This starts #2802 by implementing low-hanging fruit: single span handler.

This one is easy because it fits directly into the existing collector infrastructure. To show how nicely it does, this backports `CollectorSampler` as a `CollectedSpanHandler`

### DIFF
--- a/zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
+++ b/zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
@@ -15,16 +15,19 @@ package zipkin2.collector;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.Executor;
 import java.util.function.Supplier;
 import java.util.logging.Logger;
+import zipkin2.Call;
 import zipkin2.Callback;
 import zipkin2.Span;
 import zipkin2.SpanBytesDecoderDetector;
 import zipkin2.codec.BytesDecoder;
 import zipkin2.codec.SpanBytesDecoder;
+import zipkin2.collector.handler.CollectedSpanHandler;
 import zipkin2.storage.StorageComponent;
 
 import static java.lang.String.format;
@@ -33,21 +36,13 @@ import static zipkin2.Call.propagateIfFatal;
 
 /**
  * This component takes action on spans received from a transport. This includes deserializing,
- * sampling and scheduling for storage.
+ * sampling, invoking any handlers, and scheduling for storage.
  *
  * <p>Callbacks passed do not propagate to the storage layer. They only return success or failures
  * before storage is attempted. This ensures that calling threads are disconnected from storage
  * threads.
  */
 public class Collector { // not final for mock
-  static final Callback<Void> NOOP_CALLBACK = new Callback<Void>() {
-    @Override public void onSuccess(Void value) {
-    }
-
-    @Override public void onError(Throwable t) {
-    }
-  };
-
   /** Needed to scope this to the correct logging category */
   public static Builder newBuilder(Class<?> loggingClass) {
     if (loggingClass == null) throw new NullPointerException("loggingClass == null");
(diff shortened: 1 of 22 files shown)


zipkin-collector/core/src/main/java/zipkin2/collector/handler/CollectedSpanHandler.java
```diff
@@ -0,0 +1,62 @@
...
+   * approach to this mechanism as it results in less overhead.
+   *
+   * <p>It is ok to handle spans in a way that incurs no transformation, such as aggregating
+   * metrics. However it is inappropriate to raise any kind of exception from the handle method.
+   * Please code defensively.
+   *
+   * @param span model as {@link SpanBytesDecoder decoded} or handled upstream.
+   * @return the input if unmodified, a derived value, or null to drop this span.
+   */
+  @Nullable Span handle(Span span);
```
codefromthecrypt, line 61: here is the important type

zipkin-server/src/test/java/zipkin2/server/internal/Access.java
```diff
@@ -0,0 +1,23 @@
...
+package zipkin2.server.internal;
+
+import zipkin2.collector.Collector;
+
+/** opens package access for testing */
+public final class Access {
+  public static Collector collector(ZipkinHttpCollector httpCollector) {
+    return httpCollector.collector;
+  }
+}
```
devinsba, line 23: Interesting pattern. I think I like this more than marking something "VisibleForTesting"

zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
```diff
@@ -56,14 +51,35 @@ public static Builder newBuilder(Class<?> loggingClass) {
...
+    /**
+     * @see {@link CollectorComponent.Builder#addCollectedSpanHandler(CollectedSpanHandler)}
+     * @since 2.17
+     */
+    public Builder addCollectedSpanHandler(CollectedSpanHandler collectedSpanHandler) {
+      if (collectedSpanHandler == null) {
+        throw new NullPointerException("collectedSpanHandler == null");
+      }
+      if (collectedSpanHandler != CollectedSpanHandler.NOOP) { // lenient on config bug
+        this.collectedSpanHandlers.add(collectedSpanHandler);
```
jeqo, line 78: should we allow adding same handler many times?

jeqo (commented):
I could be missing something but should we also add how these handlers will be mounted? (e.g. via classpath, similar to storages)

zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
codefromthecrypt: I really hate the idea of duplicates, but there was a change in brave where @anuraaga felt strongly about letting duplicates exist. I can move the "this will cause bugs" disclaimer here also unless rag can hopefully change his mind https://github.com/openzipkin/brave/pull/961

codefromthecrypt: my mistake, he revised his opinion. I will revert the other change, but drop silently.

codefromthecrypt: https://github.com/openzipkin/brave/pull/1001

codefromthecrypt:
@jeqo yeah it would be via adding to the classpath using "normal spring" but it is true most probably don't know how to do that. I have a local project I'll resume tonight or tomorrow.

codefromthecrypt:
PS occurred to me that we could use a collected span handler to clean up data which results in needless index load, such as the '0' and '-' tags envoy adds. At some point these could be clean from the source, but it is a common problem, and in fact the first streaming project's goal was to clean up bad finagle data!

Ex here there are dubious tags from envoy and probably the shared flag is wrong also. cc @basvanbeek @dio
```json
  {
    "traceId": "978883983d506fa5",
    "id": "978883983d506fa5",
    "kind": "SERVER",
    "name": "localhost:10000",
    "timestamp": 1570523661059232,
    "duration": 127115,
    "localEndpoint": {
      "ipv4": "169.254.65.45"
    },
    "tags": {
      "component": "proxy",
      "downstream_cluster": "-",
      "guid:x-request-id": "d709b78f-583e-988e-b581-25b72c63852d",
      "http.method": "GET",
      "http.protocol": "HTTP/1.1",
      "http.status_code": "200",
      "http.url": "http://www.google.com/",
      "request_size": "0",
      "response_flags": "-",
      "response_size": "67342",
      "upstream_cluster": "service_google",
      "user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/77.0.3865.90 Safari/537.36"
    },
    "shared": true
  }
```

codefromthecrypt:
I'm abandoning this as MutableSpan in brave is possibly a better choice to use directly vs us copy similar types back and forth between here and brave. Since collectors can use Java 8, and different streaming engines have different transform functions.. possibly better to let experience accrue as localized plugins to projects like kafka-storage vs generalize at this point.

PR #2808, opened 2019-09-18
Adds a docker-compose.test.yml to do a health check of built container.

anuraaga:
Also adds a health check command to our docker image for being able to see health with `docker ps`.

Docker Hub automatically runs this file after building an image. Example build at https://cloud.docker.com/repository/registry-1.docker.io/anuraaga/docker-hub-test/builds/7f0936ce-74e2-4614-815c-ef57daa7db42

Currently, the test just curl's the health endpoint, without jq. This doesn't actually test until https://github.com/line/armeria/pull/2088 - I figured since this is a new test, it's not worth getting too complex finding a reputable image with curl and jq and can wait until that's in to finish it.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,7 @@
 **/\.*
 !.git
 !**/.eslintrc
+**/*.test.yml
 
 **/target
 **/node_modules
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -38,4 +38,6 @@ USER zipkin
 
 EXPOSE 9410 9411
 
+HEALTHCHECK --interval=10s --start-period=30s --timeout=5s CMD wget --quiet http://localhost:9411/health || exit 1
+
 ENTRYPOINT ["/busybox/sh", "run.sh"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -38,6 +38,7 @@ USER zipkin
 
 EXPOSE 9410 9411
 
-HEALTHCHECK --interval=10s --start-period=30s --timeout=5s CMD wget --quiet http://localhost:9411/health || exit 1
+# https://docs.docker.com/engine/reference/builder/#healthcheck
+HEALTHCHECK --interval=5s --start-period=30s --timeout=5s CMD wget --quiet http://localhost:9411/health || exit 1
 
 ENTRYPOINT ["/busybox/sh", "run.sh"]
--- a/docker/docker-compose.test.yml
+++ b/docker/docker-compose.test.yml
@@ -21,6 +21,7 @@ services:
       default:
         aliases:
         - zipkin
+  # Docker Hub automated test service - https://docs.docker.com/docker-hub/builds/automated-testing/
   sut:
     image: appropriate/curl
     networks:
--- a/docker/docker-compose.test.yml
+++ b/docker/docker-compose.test.yml
@@ -1,0 +1,30 @@
+#
+# Copyright 2015-2019 The OpenZipkin Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+# in compliance with the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License
+# is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+# or implied. See the License for the specific language governing permissions and limitations under
+# the License.
+#
+version: "3.7"
+services:
+  zipkin:
+    build:
+      context: ..
+      dockerfile: docker/Dockerfile
+    networks:
+      default:
+        aliases:
+        - zipkin
+  sut:
+    image: appropriate/curl
+    networks:
+    - default
+    command: -v --connect-timeout 5 --retry 30 --retry-connrefused --retry-delay 5 http://zipkin:9411/health
+    depends_on:
+    - zipkin
--- a/docker/docker-compose.test.yml
+++ b/docker/docker-compose.test.yml
@@ -11,7 +11,7 @@
 # or implied. See the License for the specific language governing permissions and limitations under
 # the License.
 #
-version: "3.7"
+version: "3.6"
 services:
   zipkin:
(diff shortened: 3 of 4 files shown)


docker/Dockerfile
```diff
@@ -38,4 +38,6 @@ USER zipkin
 
 EXPOSE 9410 9411
 
+HEALTHCHECK --interval=10s --start-period=30s --timeout=5s CMD wget --quiet http://localhost:9411/health || exit 1
```
devinsba: Cool, I didn't realize we were using the debug image

docker/docker-compose.test.yml
```diff
@@ -0,0 +1,30 @@
...
+services:
+  zipkin:
+    build:
+      context: ..
+      dockerfile: docker/Dockerfile
+    networks:
+      default:
+        aliases:
+        - zipkin
+  sut:
```
devinsba, line 25: perhaps `healthchecker` would be a better name?

devinsba (approved):

docker/docker-compose.test.yml
anuraaga: That name is hard coded as the name that docker hub checks for the test status - added a link to the docs so it's easier to follow.

devinsba: :+1: seems like an odd name but works for me

docker/Dockerfile
codefromthecrypt: does this imply any blocking? the start period is really long

codefromthecrypt: cc also @jcarres-mdsol in case you are using health checking in your deployment. This isn't the release dockerfile, but I suspect it will soon converge.

.dockerignore
```diff
@@ -1,6 +1,7 @@
 **/\.*
 !.git
 !**/.eslintrc
+**/*.test.yml
```
codefromthecrypt, line 4: fuzz?

codefromthecrypt (commented):
I'm interested.. I think we should document the parameters used in a comment, maybe get a few people to ack though I understand it is deployment specific.

My main concern would be blocking for an unnecessarily long time (like tools that depend on healthcheck to pass). I know we've had complaints and at least on large site dropping zipkin due to a 30s startup time.

However, I'd like to think we can improve that and would prefer a few more health checks invoked vs lock in a big delay which ideally we can improve. See https://github.com/openzipkin/zipkin/issues/2788
cc also @jeqo @eddumelendez

docker/Dockerfile
anuraaga: Doesn't block, it probes every interval regardless - went ahead and lowered interval to 5s. Any failures during the start-period aren't counted as "unhealthy", they're just skipped, but they still happen.

.dockerignore
anuraaga: This is since building a docker image doesn't require the compsoe file in the build context.

anuraaga:
Going to go ahead and merge this in to get it in action, happy to follow up with any more tweaks. The healthcheck is just a default if a user didn't specify anything, and I think it's a reasonable default for those that don't have stricter needs in their deployments.

codefromthecrypt:
Going to go ahead and merge this in to get it in action, happy to follow up
> with any more tweaks. The healthcheck is just a default if a user didn't
> specify anything, and I think it's a reasonable default for those that
> don't have stricter needs in their deployments.
>
Thanks I agree. Appreciate your answering the questions!

codefromthecrypt:
PS I don't agree that a start period of 30s is reasonable. We've intentionally overridden defaults with a very long start period. I just want to clarify that I still thing this is excessive.

anuraaga:
I think your concern was whether it blocks, but it

> Doesn't block, it probes every interval regardless - went ahead and lowered interval to 5s. Any failures during the start-period aren't counted as "unhealthy", they're just skipped, but they still happen.

A shorter start period has a higher chance of waiting for startup since if the server starts up just after the start period, it will take 3 healthy checks before the server is treated as healthy (any health check failures during the start period are ignored rather than setting the server as explicitly unhealthy, an unhealthy server requires multiple healthy checks to come back unlike a starting one that just takes one).

Let me know if there's still a problem with that behavior.

codefromthecrypt:
When I used to run systems, it was often the case (perhaps excessive) that polling intervals were one second. I have seen 5 before, though. I know at twitter it was a real problem to have long delays before health was marked as such, some of this was caused by having to readback metrics that were themselves delayed.

What you are saying is that service will receive traffic until such time as a health check is bad (ex not until such time it is good). That should be in a comment because otherwise people will read this and say "zipkin takes a half a minute to start" which echos and causes more damage and time loss than most bugs do.

codefromthecrypt:
I would suggest a comment like

```
# We put an excessive 30s initial delay to avoid spurious health check failures from delaying tests.
# The health check config here is not a recommendation for production. Do not copy/paste
```

codefromthecrypt:
For example, I know @basvanbeek had a startup on his laptop that took almost 30s. He rebooted his laptop and then it took less than 10. It is super important that we discuss the difference between things overly defensive due to laptops etc and things for production. We are in a copy/paste culture.

A health check of 1s if bad, we should also highlight that. Literally we are overriding all the defaults and the values used should be transparently explained. If 1s is triggering a bug or something, because health checks are not synchronized properly etc, we should say why the duration is so long.

This whole issue would have been easier on me at least if you were transparent about the values used in the first place. Every time we override defaults we all should know why.. that's how you can take holidays etc.

codefromthecrypt:
Anyway I can't explain the 5s part, I would like to know why we are overriding it.. if it is random, based on personal experience, or walking around a problem. If the latter, I'd like to make an issue about it.

Meanwhile I'll make a comment about the initial duration in a separate PR to try to avoid FUD from spreading.

codefromthecrypt:
Here's an attempt to get off this topic and back to work :) https://github.com/openzipkin/zipkin/pull/2812

docker/Dockerfile
codefromthecrypt: @devinsba I didn't notice the debug image until now also. In fact my hotel can't even pull the zipkin image it is so big. I opened this up for follow-up in case there's a way to reduce size again https://github.com/openzipkin/docker-zipkin/issues/226

PR #2810, opened 2019-09-19
ignore docker yml files in license plugin

jorgheymans:


codefromthecrypt:
patch-5 ... keep'em coming!

PR #2816, opened 2019-09-23
Removes spring-web dependency

codefromthecrypt:
We don't use any Spring infrastructure for serving requests or making
them: Everything is Armeria. This removes a couple megs of dependencies
we don't use.

```before
du -k ./zipkin-server/target/zipkin-server-*exec.jar
57372	./zipkin-server/target/zipkin-server-2.16.3-SNAPSHOT-exec.jar
```

```after
du -k ./zipkin-server/target/zipkin-server-*exec.jar
55100	./zipkin-server/target/zipkin-server-2.16.3-SNAPSHOT-exec.jar
```

cc @trustin @anuraaga @minwoox

### DIFF
--- a/zipkin-server/src/it/minimal-dependencies/src/test/java/zipkin/minimal/ZipkinServerTest.java
+++ b/zipkin-server/src/it/minimal-dependencies/src/test/java/zipkin/minimal/ZipkinServerTest.java
@@ -34,8 +34,11 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 @SpringBootTest(
   classes = ZipkinServer.class,
-  webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
-  properties = {"spring.config.name=zipkin-server", "server.port=0"}
+  webEnvironment = SpringBootTest.WebEnvironment.NONE, // RANDOM_PORT requires spring-web
+  properties = {
+    "server.port=0",
+    "spring.config.name=zipkin-server"
+  }
 )
 @RunWith(SpringRunner.class)
 public class ZipkinServerTest {
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
@@ -14,6 +14,7 @@
 package zipkin2.server.internal;
 
 import brave.Tracing;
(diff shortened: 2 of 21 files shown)


zipkin-server/src/it/minimal-dependencies/src/test/java/zipkin/minimal/ZipkinServerTest.java
```diff
@@ -34,8 +34,11 @@
 
 @SpringBootTest(
   classes = ZipkinServer.class,
-  webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
-  properties = {"spring.config.name=zipkin-server", "server.port=0"}
+  webEnvironment = SpringBootTest.WebEnvironment.NONE, // RANDOM_PORT requires spring-web
```
codefromthecrypt, line 37: incidentally tests run *much* faster now, as they don't re-initialize RestTemplate proxies, which we never used either.

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
```diff
@@ -68,6 +68,10 @@
   @Autowired(required = false)
   ZipkinMetricsHealthController healthController;
 
+  @Bean @ConditionalOnMissingBean ObjectMapper objectMapper() {
+    return new ObjectMapper();
```
anuraaga, line 72: Does the current server use this or is it just for safety? Wondering if the default object mapper should ignore missing properties (good for JSON APIs where fields get added or ignored, which might be true for our ES client).

codefromthecrypt: `ZipkinMetricsHealthController` uses this, though databind is indeed used primarily by ES. I'll scrub it out as better to be aware of where we are using databind (one of our largest dependencies)

codefromthecrypt: (scrub it from the server, to avoid this bean, not from ES at the moment)

codefromthecrypt: https://github.com/openzipkin/zipkin/pull/2819 fixes this

codefromthecrypt: nice catch btw

PR #2818, opened 2019-09-23
Introduces a second "slim" build of the server

codefromthecrypt:
This introduces a second classifier, slim, which includes Armeria-native functionality. This would include all HTTP ingress, including gRPC, and HTTP egress (ex ElasticSearch storage), but not other storage or transports.

Also, slim disables the Actuator functionality, while still retaining support for endpoints we've previously documented as supported. (In #2817, we discovered even a pared down actuator contributes ~12% to our startup time).

The main goal is to allow a quicker start for those piloting Zipkin or making test environments. This quicker start is a function of dramatically faster boostrap and also image size.

Best of 5, slim starts in 2.3s where exec (including actuator) starts in 3.1s

57% smaller than including all the dependencies for Actuator, Cassandra, MySQL and messaging transports (RabbitMQ, Kafka, ActiveMQ):

```bash
$ du -k ./zipkin-server/target/zipkin-server-2.16.3-*jar
54712	./zipkin-server/target/zipkin-server-2.16.3-SNAPSHOT-exec.jar
23596	./zipkin-server/target/zipkin-server-2.16.3-SNAPSHOT-slim.jar
```

### DIFF
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -493,6 +493,111 @@
               <classifier>exec</classifier>
             </configuration>
           </execution>
+
+          <!-- Make a jar less than half the size, which only supports armeria-native features -->
+          <execution>
+            <id>slim</id>
+            <goals>
+              <goal>repackage</goal>
+            </goals>
+            <configuration>
+              <classifier>slim</classifier>
+              <!-- https://github.com/spring-projects/spring-boot/issues/3426 transitive exclude doesn't work -->
+              <excludeGroupIds>
+                io.zipkin.java,com.google.auto.value,com.google.guava,io.dropwizard.metrics,com.datastax.cassandra,com.github.jnr,org.ow2.asm,org.jooq,javax.xml.bind,org.mariadb.jdbc,com.zaxxer,org.apache.activemq,org.apache.geronimo.specs,org.fusesource.hawtbuf,org.apache.kafka,com.github.luben,org.lz4,org.xerial.snappy,com.rabbitmq,jakarta.annotation,org.apache.thrift,
+              </excludeGroupIds>
+              <excludes>
+                <!-- Actuator -->
+                <dependency>
+                  <groupId>${armeria.groupId}</groupId>
+                  <artifactId>armeria-spring-boot-actuator-autoconfigure</artifactId>
+                </dependency>
+                <dependency>
+                  <groupId>org.springframework.boot</groupId>
+                  <artifactId>spring-boot-actuator-autoconfigure</artifactId>
+                </dependency>
+                <dependency>
+                  <groupId>org.springframework.boot</groupId>
+                  <artifactId>spring-boot-actuator</artifactId>
+                </dependency>
+                <dependency>
+                  <groupId>com.fasterxml.jackson.datatype</groupId>
+                  <artifactId>jackson-datatype-jsr310:jar</artifactId>
+                </dependency>
+
+                <!-- Unnecessary netty deps -->
+                <dependency>
+                  <groupId>io.netty</groupId>
+                  <artifactId>netty-tcnative-boringssl-static</artifactId>
+                </dependency>
+                <dependency>
+                  <groupId>io.netty</groupId>
+                  <artifactId>netty-codec-haproxy</artifactId>
+                </dependency>
+
+                <!-- Unnecessary micrometer deps -->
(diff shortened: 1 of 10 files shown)


codefromthecrypt:
cc @openzipkin/devops-tooling 

I rebuilt a docker image, only changing the exec jar to this slim jar. Best of 5 "startups inside docker" went down from 5.7 to 4.1s. I suspect this will be linear with slower hosts also.

While I tried changing to the JRE 1.8 image, I don't recommend doing that. It is much smaller (139M instead of 221M), but starts up slower. Ideally, we can recover docker image size in other ways like  https://github.com/openzipkin/docker-zipkin/issues/226

zipkin-server/src/main/java/zipkin2/server/internal/ui/ZipkinUiConfiguration.java
```diff
@@ -36,12 +37,10 @@
 import io.netty.handler.codec.http.cookie.ServerCookieDecoder;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.UncheckedIOException;
+import java.io.InputStreamReader;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
-import org.jsoup.Jsoup;
```
codefromthecrypt: jsoup was a large dependency, and we are only replacing a href tag in staticly defined index.html under our control. regex will do the trick.

zipkin-server/src/main/java/zipkin2/server/internal/ui/ZipkinUiConfiguration.java
```diff
@@ -137,6 +137,9 @@ String processedIndexHtml(Resource indexHtml) {
 
   @Bean @Lazy ArmeriaServerConfigurator uiServerConfigurator(
     IndexSwitchingService indexSwitchingService) throws IOException {
+    HttpService indexService =
```
codefromthecrypt: slim doesn't include zipkin-classic, so we need this to tolerate its absence

codefromthecrypt:
TODO: add note prior to merge that "slim" is not a good base image to layer on. The reason is that even mild configuration mess-ups are hard to diagnose unless you have actuator. IOTW, folks using slim shouldn't monkey with it at all, and pay even stricter attention to the supported variables in our README.

codefromthecrypt:
I've extracted #2820 to buy us time to make the decision (as it makes it possible to build slim variant externally)

codefromthecrypt:
verified https://github.com/line/armeria/pull/2107 via the following. If that PR changes, just do the below again until merged. cc @trustin @anuraaga

```
$ ./mvnw -Dlicense.skip=true -DskipTests --also-make -pl zipkin-server clean install -U
$ SELF_TRACING_ENABLED=true STORAGE_TYPE=elasticsearch java -jar ./zipkin-server/target/zipkin-server-*slim.jar
```

codefromthecrypt:
I've decided I will temporarily break this in order to pick up some quickness in master branch. will comment upon rebase

codefromthecrypt:
wait.. maybe I can get around it :P

codefromthecrypt:
In docker, this starts up in 3.5s on the current base image. This is .5s faster than not slim same base commit.

codefromthecrypt:
ok slim now no longer uses log4j2 due to work by @jorgheymans .. just need to get it to use slf4j-simple :P

zipkin-server/pom.xml
```diff
@@ -134,6 +134,18 @@
       <artifactId>jul-to-slf4j</artifactId>
     </dependency>
 
+    <!-- the "exec" jar will exclude these -->
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-slf4j</artifactId>
```
jorgheymans: https://github.com/spring-projects/spring-boot/tree/master/spring-boot-project/spring-boot-starters does not list this as a starter, confused where this comes from ?

codefromthecrypt: sorry I will fix this

codefromthecrypt:
-Dorg.slf4j.simpleLogger.log.zipkin2=debug totally works

codefromthecrypt:
We've done a lot of work since this started. Now, slim starts in less than 2s

PR #2819, opened 2019-09-24
Decouples health and metrics controllers, backfills tests for health

codefromthecrypt:
Before, we had health and metrics code mixed in the same controller.
This decouples that and ensures not only that health json is pretty by
default, but that it is also verified. For example, we had an encoding
bug on an error path we didn't notice.


zipkin-server/src/main/java/zipkin2/server/internal/health/ComponentHealth.java
```diff
@@ -41,25 +36,13 @@ public static ComponentHealth ofComponent(Component component) {
     return new ComponentHealth(component.toString(), STATUS_DOWN, error);
   }
 
-  @JsonIgnore final String name;
+  final String name;
   final String status;
-  @JsonProperty("details.error") final String error;
```
codefromthecrypt, line 46: this was a bug not noticed

zipkin-server/src/test/java/zipkin2/server/internal/health/ZipkinHealthControllerTest.java
```diff
@@ -0,0 +1,56 @@
...
+      + "  \"status\" : \"DOWN\",\n"
+      + "  \"zipkin\" : {\n"
+      + "    \"status\" : \"DOWN\",\n"
+      + "    \"details\" : {\n"
+      + "      \"foo\" : {\n"
+      + "        \"status\" : \"UP\"\n"
+      + "      },\n"
+      + "      \"bar\" : {\n"
+      + "        \"status\" : \"DOWN\",\n"
+      + "        \"details\" : {\n"
```
codefromthecrypt, line 47: now, we check proper nesting

zipkin-server/src/main/java/zipkin2/server/internal/health/ZipkinHealthController.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    String healthJson = writeJson(overallStatus, healths);
+    return newHealthResponse(overallStatus, mediaType, healthJson);
+  }
+
+  static HttpResponse newHealthResponse(String overallStatus, MediaType mediaType,
+    String healthJson) {
+    byte[] body = healthJson.getBytes(UTF_8);
+    int code = overallStatus.equals(STATUS_UP) ? 200 : 503;
```
anuraaga, line 101: More readable to use `HttpStatus` then a number.

zipkin-server/src/main/java/zipkin2/server/internal/health/ZipkinHealthController.java
```diff
@@ -0,0 +1,153 @@
...
+    ResponseHeaders headers = ResponseHeaders.builder(code)
+      .contentType(mediaType)
+      .setInt(CONTENT_LENGTH, body.length).build();
+    return HttpResponse.of(AggregatedHttpResponse.of(headers, HttpData.wrap(body)));
```
anuraaga, line 105: Think this method can just be `HttpResponse.of(status, mediaType, healthJson)` headers are prepared automatically.

zipkin-server/src/main/java/zipkin2/server/internal/health/ZipkinHealthController.java
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  static String writeJsonError(String error) throws IOException {
+    StringWriter writer = new StringWriter();
```
anuraaga, line 109: Could consider `ByteArrayOutputStream` since it immediately becomes bytes anyways

zipkin-server/src/main/java/zipkin2/server/internal/health/ZipkinHealthController.java
```diff
@@ -0,0 +1,153 @@
...
+    generator.writeEndObject(); // .zipkin
+    generator.writeEndObject(); // .
+    generator.flush();
+    return writer.toString();
+  }
+
+  static String writeJson(String overallStatus, List<ComponentHealth> healths) throws IOException {
+    StringWriter writer = new StringWriter();
+    JsonGenerator generator = JSON_FACTORY.createGenerator(writer);
+    generator.useDefaultPrettyPrinter();
```
anuraaga, line 127: Just wondering why we use pretty print here and not above.

zipkin-server/src/main/java/zipkin2/server/internal/prometheus/ZipkinMetricsController.java
```diff
@@ -0,0 +1,57 @@
...
+    this.meterRegistry = meterRegistry;
+    this.collectorRegistry = collectorRegistry;
+  }
+
+  // Extracts Zipkin metrics to provide backward compatibility
+  @Get("/metrics")
+  @ProducesJson
+  public ObjectNode fetchMetricsFromMicrometer() {
+    ObjectNode metricsJson = factory.objectNode();
+    // Get the Zipkin Custom meters for constructing the Metrics endpoint
```
anuraaga, line 41: Since we rewrote a lot using `JsonGenerator`, maybe a good time to port this snippet too.

zipkin-server/src/main/java/zipkin2/server/internal/health/ZipkinHealthController.java
codefromthecrypt: fixing in another PR

zipkin-server/src/main/java/zipkin2/server/internal/health/ZipkinHealthController.java
codefromthecrypt: thx I had no idea why we were double-wrapping earlier

zipkin-server/src/main/java/zipkin2/server/internal/prometheus/ZipkinMetricsController.java
codefromthecrypt: https://github.com/openzipkin/zipkin/pull/2821

codefromthecrypt:
thanks for the feedback @anuraaga!

PR #2821, opened 2019-09-24
Removes remaining use of jackson-databind from zipkin-server

codefromthecrypt:
This is still in use in elasticsearch, but the server no longer uses it.
Also addresses feedback from @anuraaga in https://github.com/openzipkin/zipkin/pull/2819

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/health/ZipkinHealthController.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/health/ZipkinHealthController.java
@@ -15,11 +15,9 @@ package zipkin2.server.internal.health;
 
 import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonGenerator;
-import com.linecorp.armeria.common.AggregatedHttpResponse;
-import com.linecorp.armeria.common.HttpData;
 import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
 import com.linecorp.armeria.common.MediaType;
-import com.linecorp.armeria.common.ResponseHeaders;
 import com.linecorp.armeria.server.ServiceRequestContext;
 import com.linecorp.armeria.server.annotation.Get;
 import java.io.IOException;
@@ -29,8 +27,6 @@ import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
 import zipkin2.Component;
 
-import static com.linecorp.armeria.common.HttpHeaderNames.CONTENT_LENGTH;
-import static java.nio.charset.StandardCharsets.UTF_8;
 import static zipkin2.server.internal.ZipkinServerConfiguration.MEDIA_TYPE_ACTUATOR;
 import static zipkin2.server.internal.health.ComponentHealth.STATUS_DOWN;
 import static zipkin2.server.internal.health.ComponentHealth.STATUS_UP;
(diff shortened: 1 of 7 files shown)


codefromthecrypt:
I started to look into the scope of dropping databind from ES.. mainly it is the marshaling the search request. It is really odd how huge that jackson-databind is compared to what we had before. moshi is <200k. Anyway it is feasible given a long flight to drop our jar size over a meg once we axe databind from ES.

zipkin-server/src/main/java/zipkin2/server/internal/health/ZipkinHealthController.java
```diff
@@ -95,19 +91,15 @@ static HttpResponse newHealthResponse(List<ComponentHealth> healths, MediaType m
...
-    return HttpResponse.of(AggregatedHttpResponse.of(headers, HttpData.wrap(body)));
+  static HttpResponse newHealthResponse(String status, MediaType mediaType, String healthJson) {
+    HttpStatus code = status.equals(STATUS_UP) ? HttpStatus.OK : HttpStatus.SERVICE_UNAVAILABLE;
+    return HttpResponse.of(code, mediaType, healthJson);
   }
 
   static String writeJsonError(String error) throws IOException {
     StringWriter writer = new StringWriter();
     JsonGenerator generator = JSON_FACTORY.createGenerator(writer);
+    generator.useDefaultPrettyPrinter();
```
anuraaga: I believe calling this every time is equivalent to this on the factory

`.enable(SerializationFeature.INDENT_OUTPUT);`

zipkin-server/src/main/java/zipkin2/server/internal/prometheus/ZipkinMetricsController.java
```diff
@@ -47,11 +54,13 @@ public ObjectNode fetchMetricsFromMicrometer() {
...
-        metricsJson.put("counter." + name + "." + transport, ((Counter) meter).count());
+        generator.writeNumberField("counter." + name + "." + transport, ((Counter) meter).count());
       } else if (type == Meter.Type.GAUGE) {
-        metricsJson.put("gauge." + name + "." + transport, ((Gauge) meter).value());
+        generator.writeNumberField("gauge." + name + "." + transport, ((Gauge) meter).value());
       } // We only use counters and gauges
     }
-    return metricsJson;
+    generator.writeEndObject();
+    generator.flush();
```
anuraaga: Ah hadn't noticed this on the other snippets - I'd usually use try/resources on the generator instead of manually flushing

zipkin-server/src/test/java/zipkin2/server/internal/prometheus/ITZipkinMetrics.java
```diff
@@ -200,11 +201,14 @@ private Response post(String path, byte[] body) throws IOException {
...
   }
 
-  static List readJson(String json) throws Exception {
-    ObjectMapper mapper = new ObjectMapper();
-    JsonNode jsonNode = mapper.readTree(json);
-    List<String> fieldsList = new ArrayList<>();
-    jsonNode.fieldNames().forEachRemaining(fieldsList::add);
-    return fieldsList;
+  static Map<String, Integer> readJson(String json) throws Exception {
+    Map<String, Integer> result = new LinkedHashMap<>();
```
anuraaga, line 205: Maybe it's fine to parse straight into a map for tests

zipkin-server/src/main/java/zipkin2/server/internal/ui/ZipkinUiConfiguration.java
```diff
@@ -146,9 +149,9 @@ String processedIndexHtml(Resource indexHtml) throws IOException {
         .orElse(
           HttpFileServiceBuilder.forClassPath("zipkin-lens").cacheControl(maxAgeYear).build());
 
-    byte[] config = new ObjectMapper().writeValueAsBytes(ui);
+    String config = writeConfig(ui);
     return sb -> {
-      sb.service("/zipkin/config.json", HttpFileBuilder.of(HttpData.wrap(config))
+      sb.service("/zipkin/config.json", HttpFileBuilder.of(HttpData.of(UTF_8, config))
```
anuraaga: Think there's an `ofUtf8` method

anuraaga (approved):

zipkin-server/src/test/java/zipkin2/server/internal/prometheus/ITZipkinMetrics.java
codefromthecrypt: this was mostly to get rid of databind api usage (eventhough it is an implicit dep)

codefromthecrypt:
outside the server, the only thing we use databind for is elasticsearch read requests which is not performance bound (as ES results are way slower than making requests). I only now noticed this from a @bsideup tweet. Seems an easier way to kill a meg https://github.com/FasterXML/jackson-jr

zipkin-server/src/main/java/zipkin2/server/internal/health/ZipkinHealthController.java
codefromthecrypt: actually that is in the databind package.

zipkin-server/src/main/java/zipkin2/server/internal/prometheus/ZipkinMetricsController.java
codefromthecrypt: in some cases this helps, though not 100% it is that much more clear. I changed to try/finally where the extra indent level didn't cause more pain

zipkin-server/src/main/java/zipkin2/server/internal/ui/ZipkinUiConfiguration.java
codefromthecrypt: nice!

zipkin-server/src/test/java/zipkin2/server/internal/prometheus/ITZipkinMetrics.java
codefromthecrypt: I'll leave this as plan to remove databind and prefer to not re-introduce it just to remove it

anuraaga (approved):

PR #2822, opened 2019-09-24
downgrade surefire and failsafe

jorgheymans:
zipkin core tests are not being run due to maven plugin conflicts
https://travis-ci.org/openzipkin/zipkin/jobs/589012904#L1708
and 
https://springframework.guru/why-your-junit-5-tests-are-not-running-under-maven/

downgrading to 2.21.0 made the tests run again, the rest of the modules did not seem to care either way.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -95,7 +95,11 @@
     <license-maven-plugin.version>3.0</license-maven-plugin.version>
     <maven-jar-plugin.version>3.1.2</maven-jar-plugin.version>
     <maven-shade-plugin.version>3.2.1</maven-shade-plugin.version>
-    <maven-failsafe-plugin.version>3.0.0-M3</maven-failsafe-plugin.version>
+    <!--
+      See https://springframework.guru/why-your-junit-5-tests-are-not-running-under-maven/
+      but even though it says to use 2.22.0 this did not solve the issue either.
+    -->
+    <maven-failsafe-plugin.version>2.21.0</maven-failsafe-plugin.version>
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
Thanks Jorg. I think we need to look closer.. at least fail safe is
working else build wouldnt take so long. I would prefer to cite an actual
issue vs a blog as not always blogs are correct and at least issues can go
somewhere

cc @anuraaga

codefromthecrypt:
I think there's something else going on. For example, stackdriver has mixed junit classic and jupiter tests in one module, and I notice they run: `./mvnw clean verify -pl  :zipkin-autoconfigure-storage-stackdriver`. Maybe there's something @saturnism fixed we need to carry here?

https://github.com/openzipkin/zipkin-gcp/commit/32c479d16f209f49e51e63f6acb32a1107ed5ee3

jorgheymans:
It's entirely possible something else is going on but i'm pretty sure zipkin-core tests were being run in Travis at some point, and most likely an upgrade of either maven or a plugin made it crash. zipkin-core is really trivial in terms of tests, yet surefire is unable to run them. I'll raise an issue to track, you can either downgrade and let this work out itself - or get out the flashlight and enter the rabbit hole ;-)

codefromthecrypt:
let me take this on on. thanks @jorgheymans!

codefromthecrypt:
ps I didn't get to this. if someone wants to yank please do

codefromthecrypt:
fixed by @anuraaga #2828

PR #2827, opened 2019-10-01
Ensures index templates are verified before passing health.

### LINKED ISSUE #2825: ES health check should block on schema creation
There's a race condition where a POST could leak prior to elasticsearch storage's schema being fully initialized. It leads to query errors like this:

```
java.lang.IllegalStateException: response for aggregation failed: {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Fielddata is disabled on text fields by default. Set fielddata=true on [traceId] in order to load fielddata in memory by uninverting the inverted index. Note that this can however use significant memory. Alternatively use a keyword field instead."}],"type":"search_phase_execution_exception","reason":"all shards failed","phase":"query","grouped":true,"failed_shards":[{"shard":0,"index":"zipkin:span-2019-09-26","node":"Htub5Z_6Rp2gARM-ckCj6Q","reason":{"type":"illegal_argument_exception","reason":"Fielddata is disabled on text fields by default. Set fielddata=true on [traceId] in order to load fielddata in memory by uninverting the inverted index. Note that this can however use significant memory. Alternatively use a keyword field instead."}}],"caused_by":{"type":"illegal_argument_exception","reason":"Fielddata is disabled on text fields by default. Set fielddata=true on [traceId] in order to load fielddata in memory by uninverting the inverted index. Note that this can however use significant memory. Alternatively use a keyword field instead.","caused_by":{"type":"illegal_argument_exception","reason":"Fielddata is disabled on text fields by default. Set fielddata=true on [traceId] in order to load fielddata in memory by uninverting the inverted index. Note that this can however use significant memory. Alternatively use a keyword field instead."}}},"status":400}
    at zipkin2.elasticsearch.internal.client.HttpCall.parseResponse(HttpCall.java:151) ~[zipkin-storage-elasticsearch-2.14.2.jar!/:?]
    at zipkin2.elasticsearch.internal.client.HttpCall.doExecute(HttpCall.java:78) ~[zipkin-storage-elasticsearch-2.14.2.jar!/:?]
```

Notably, this is possible when there are multiple collector nodes.. basically the index template could be in the process of being applied on one node while the other checks, and that lets a POST through which is now not indexed properly.

The reason is that the health check doesn't initialize the index templates, such as we do in cassandra. The fix is easy, but for those who need a workaround now, hit `/api/v2/autocompleteKeys` instead as eventhough it returns a constant (no perf impact), it does the init logic needed first (ensures index templates).

Thanks @basvanbeek for noticing this. cc @openzipkin/elasticsearch

codefromthecrypt:
Currently, we pass health checks which can hide errors provisioning
index templates, and also worsen template provisioning race conditions.

This makes sure index templates are at least checked once per startup.

This also refactors the code so that someone can later do offline
template installation.

Fixes #2825

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
@@ -23,6 +23,7 @@ import com.linecorp.armeria.common.AggregatedHttpRequest;
 import com.linecorp.armeria.common.HttpMethod;
 import java.io.Closeable;
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -37,7 +38,6 @@ import zipkin2.elasticsearch.internal.Internal;
 import zipkin2.elasticsearch.internal.client.HttpCall;
 import zipkin2.elasticsearch.internal.client.HttpCall.BodyConverter;
 import zipkin2.internal.Nullable;
-import zipkin2.internal.Platform;
 import zipkin2.storage.AutocompleteTags;
 import zipkin2.storage.ServiceAndSpanNames;
 import zipkin2.storage.SpanConsumer;
@@ -45,10 +45,11 @@ import zipkin2.storage.SpanStore;
 import zipkin2.storage.StorageComponent;
 import zipkin2.storage.Traces;
 
-import static zipkin2.elasticsearch.ElasticsearchAutocompleteTags.AUTOCOMPLETE;
-import static zipkin2.elasticsearch.ElasticsearchSpanStore.DEPENDENCY;
-import static zipkin2.elasticsearch.ElasticsearchSpanStore.SPAN;
+import static com.linecorp.armeria.common.HttpMethod.GET;
 import static zipkin2.elasticsearch.EnsureIndexTemplate.ensureIndexTemplate;
+import static zipkin2.elasticsearch.VersionSpecificTemplates.TYPE_AUTOCOMPLETE;
+import static zipkin2.elasticsearch.VersionSpecificTemplates.TYPE_DEPENDENCY;
+import static zipkin2.elasticsearch.VersionSpecificTemplates.TYPE_SPAN;
 import static zipkin2.elasticsearch.internal.JsonReaders.enterPath;
 
 @AutoValue
@@ -242,8 +243,8 @@ public abstract class ElasticsearchStorage extends zipkin2.storage.StorageCompon
   /** This is an internal blocking call, only used in tests. */
   public void clear() throws IOException {
     Set<String> toClear = new LinkedHashSet<>();
-    toClear.add(indexNameFormatter().formatType(SPAN));
-    toClear.add(indexNameFormatter().formatType(DEPENDENCY));
+    toClear.add(indexNameFormatter().formatType(TYPE_SPAN));
+    toClear.add(indexNameFormatter().formatType(TYPE_DEPENDENCY));
     for (String index : toClear) clear(index);
   }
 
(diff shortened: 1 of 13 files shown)


zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -265,37 +266,71 @@ void clear(String index) throws IOException {
...
+  CheckResult ensureIndexTemplatesAndClusterReady(String index) {
     try {
-      HttpCall.Factory http = http();
-      AggregatedHttpRequest request = AggregatedHttpRequest.of(
-        HttpMethod.GET, "/_cluster/health/" + index);
-      return http.newCall(request, READ_STATUS, "get-cluster-health").execute();
+      ensureIndexTemplates(); // called only once, so we have to double-check health
+      AggregatedHttpRequest request = AggregatedHttpRequest.of(GET, "/_cluster/health/" + index);
+      CheckResult result = http().newCall(request, READ_STATUS, "get-cluster-health").execute();
+      if (result == null) throw new IllegalArgumentException("No content reading cluster health");
```
codefromthecrypt, line 283: ps accidentally found a couple NPE doing this :P

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -265,37 +266,71 @@ void clear(String index) throws IOException {
...
+      return CheckResult.failed(e);
     }
   }
 
-  @Memoized // since we don't want overlapping calls to apply the index templates
+  volatile IndexTemplates indexTemplates; // visible for testing
+
+  // Memoized since we don't want overlapping calls to apply the index templates
   IndexTemplates ensureIndexTemplates() {
+    if (indexTemplates == null) {
```
codefromthecrypt, line 302: this is the same thing that auto-value would generate (we do this all over in zipkin-reporter). plus on my list to un-auto-value anyway.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -265,37 +266,71 @@ void clear(String index) throws IOException {
...
-      ensureIndexTemplate(http, buildUrl(templates, SPAN), templates.span());
-      ensureIndexTemplate(http, buildUrl(templates, DEPENDENCY), templates.dependency());
-      ensureIndexTemplate(http, buildUrl(templates, AUTOCOMPLETE), templates.autocomplete());
+      IndexTemplates templates = versionSpecificTemplates(http);
+      ensureIndexTemplate(http, buildUrl(templates, TYPE_SPAN), templates.span());
+      ensureIndexTemplate(http, buildUrl(templates, TYPE_DEPENDENCY), templates.dependency());
+      ensureIndexTemplate(http, buildUrl(templates, TYPE_AUTOCOMPLETE), templates.autocomplete());
       return templates;
     } catch (IOException e) {
-      throw Platform.get().uncheckedIOException(e);
```
codefromthecrypt, line 295: guard not needed as storage components are all java 1.8+

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -265,37 +266,71 @@ void clear(String index) throws IOException {
...
+      throw new UncheckedIOException(e);
     }
   }
 
+  IndexTemplates versionSpecificTemplates(HttpCall.Factory http) throws IOException {
+    Float version = ElasticsearchVersion.INSTANCE.get(http);
```
codefromthecrypt: fixed this in the ElasticsearchVersion type

```suggestion
    float version = ElasticsearchVersion.INSTANCE.get(http);
```

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/VersionSpecificTemplates.java
```diff
@@ -13,51 +13,54 @@
...
    * In Zipkin search, we do exact match only (keyword). Norms is about scoring. We don't use that
    * in our API, and disable it to reduce disk storage needed.
    */
   static final String KEYWORD = "{ \"type\": \"keyword\", \"norms\": false }";
 
-  final ElasticsearchStorage es;
-
-  VersionSpecificTemplates(ElasticsearchStorage es) {
-    this.es = es;
+  //final ElasticsearchStorage es;
```
codefromthecrypt:
```suggestion
```

anuraaga (approved):

PR #2830, opened 2019-10-02
move to slf4j-api and slf4j-simple

jorgheymans:


jorgheymans:
before:
image
after:
image

Compared to 2.17 the server jar is about 2MB smaller  (1988790 bytes)

zipkin-collector/core/src/test/java/zipkin2/collector/CollectorTest.java
```diff
@@ -45,22 +45,13 @@
...
-    {
-      setLevel(Level.ALL);
-    }
-
-    @Override public void log(Level level, String msg, Throwable thrown) {
-      assertThat(level).isEqualTo(Level.FINE);
-      messages.add(unprefixIdString(msg));
-    }
-  };
+  private TestLogger testLogger = TestLoggerFactory.getTestLogger("");
```
jorgheymans, line 48: jaeger also uses this in their tests so i figured it was ok :-)

zipkin-collector/core/pom.xml
```diff
@@ -31,6 +31,16 @@
   </properties>
 
   <dependencies>
+    <dependency>
+      <groupId>org.slf4j</groupId>
```
codefromthecrypt, line 35: this is ok with me because we talk about collector being only for server now.

pom.xml
```diff
@@ -372,6 +360,18 @@
   </dependencyManagement>
 
   <dependencies>
+    <!-- all modules use slf4j-api -->
+    <dependency>
```
codefromthecrypt: not all. we can't do this here.. this will add the dependency to zipkin's core jar which will end up everywhere in brave etc. I think you'll want to move this to dependency management

zipkin-junit/pom.xml
```diff
@@ -55,6 +55,14 @@
       <artifactId>junit</artifactId>
     </dependency>
 
+    <dependency>
+      <!-- okhttp uses jul -->
+      <groupId>org.slf4j</groupId>
```
codefromthecrypt, line 60: so do we (in the core jar), so .. we'll need this in the server. Also we need to make sure we didn't accidentally add a strict dep here on slf4j either as it is unnecessary.

zipkin-junit/src/test/java/zipkin2/junit/ZipkinRuleTest.java
```diff
@@ -38,6 +39,12 @@
 
 public class ZipkinRuleTest {
 
+  static {
+    // ensure jul-to-slf4j works
```
codefromthecrypt, line 43: cool

zipkin-server/README.md
```diff
@@ -45,12 +45,14 @@ ZIPKIN_QUERY_ALLOWED_ORIGINS=http://foo.bar.com
...
-By default, zipkin writes log messages to the console at INFO level and above. You can adjust categories using the `--logging.level.XXX` parameter, a `-Dlogging.level.XXX` system property, or by adjusting [yaml configuration](src/main/resources/zipkin-server-shared.yml).
+By default, zipkin writes log messages to the console at INFO level and above. A limited set of 
+[configuration options](https://www.slf4j.org/api/org/slf4j/impl/SimpleLogger.html) is available 
+to control the logging output. 
 
-For example, if you want to enable debug logging for all zipkin categories, you can start the server like so:
+For example, if you want to enable debug logging for all zipkin categories:
 
 ```bash
-$ java -jar zipkin.jar --logging.level.zipkin2=DEBUG
```
codefromthecrypt: I think maybe spring-boot-starter log4j should obviate this?

zipkin-server/src/main/java/zipkin/server/ZipkinServer.java
```diff
@@ -21,11 +21,6 @@
 @SpringBootApplication
 @EnableZipkinServer
 public class ZipkinServer {
-  static {
```
codefromthecrypt, line 24: think maybe need to install the logging bridge (you can try hitting the dependencies endpoint on trace level when you have spans loaded)

zipkin/pom.xml
```diff
@@ -42,6 +42,13 @@
       <scope>provided</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.slf4j</groupId>
```
codefromthecrypt: yeah sadly you'll have to do the opposite... explicitly add to collector and storage main poms, not at parent as it leaks here

zipkin/src/main/java/zipkin2/Span.java
```diff
@@ -25,14 +25,13 @@
 import java.util.Locale;
 import java.util.Map;
 import java.util.TreeMap;
-import java.util.logging.Logger;
```
codefromthecrypt: please revert all of this stuff in core.. I'll take another pass after. cheers and thanks for the help!

codefromthecrypt (commented):
made some notes and happy to do another pass!

anuraaga (commented):
I think what we get from this is a size savings in the server jar. Could you list up what we lose in exchange? We should make sure it's all stuff we're not interested in supporting. 

Things that come to mind without thinking much

- performance (? not sure of the real-world impact). By default there aren't many logs, but enabling access logs of server and/or client could have lots and are possibly useful for certain sites.

- Site overrides of log4j.properties. I guess it's fairly simple to override the logging settings at a site, one useful override I can imagine is switching to JSON layout to integrate with JSON-based logging collectors. Is this something we don't support? (just confirming, want to avoid accidentally losing something we do support)

- log4j plugins. Similar to above - like how we encourage users to plugin for something like secondary sampling, log4j is nice and pluggable too. Is this something we don't support?

- stability? No clue on this, but I imagine the number of servers that use slf4j-simple in the wild in production is close to zero, wonder if we'll run into something unexpected

Just want to vet the impact of the change

codefromthecrypt:
For the record, I think I was the one that pushed for log4j2 in the first
place. no one asked for it. It was slf4j before as default in spring boot.

We don't formally support custom servers anyway, so I wouldn't wander down
that path of plugins etc. OTOH, it is fair to ask what we ourselves want to
use or support.

bear in mind all someone has to do in normal slf4j is to make a layer that
deletes the simple jar and put what they want in. of course switching
completely off slf4j would be more work.

anuraaga:
That's a good point - migrating to slf4j-api seems like it has no drawbacks since it's more extensible. Actually it made me realize that by doing that, we'd also have the option of log4j2 backend in the normal image and simple backend in the proposed `slim` distribution - some food for thought :)

codefromthecrypt:
yeah provided our deps haven't moved to log4j2, which I don't think
any have.. we can keep watch of that.

codefromthecrypt:
@jorgheymans considering what Rag suggested.. I agree. I think also to respect your time most, we should focus this change on moving to slf4j api (including spring boot integration of it), and keeping the old logging backend if possible. We can do a separate pull request to then swap log4j2 the backend out.. taking time with that but without holding the slf4j-api stuff hostage. Make sense?

of course the tests can still use simple etc

jorgheymans:
> * performance (? not sure of the real-world impact). By default there aren't many logs, but enabling access logs of server and/or client could have lots and are possibly useful for certain sites.

Honestly i can't tell. Gut feeling says 'simple' should be 'faster' but yeah not always true. It's literally just 5 classes of which 4 dealing with config. https://github.com/qos-ch/slf4j/tree/master/slf4j-simple/src/main/java/org/slf4j/simple . The logging method itself is nothing exotic https://github.com/qos-ch/slf4j/blob/233803688a809fa8ea599306a058351d4c4246a4/slf4j-simple/src/main/java/org/slf4j/simple/SimpleLogger.java#L362 . I'll ask on their user list for perf numbers.

> * Site overrides of log4j.properties. I guess it's fairly simple to override the logging settings at a site, one useful override I can imagine is switching to JSON layout to integrate with JSON-based logging collectors. Is this something we don't support? (just confirming, want to avoid accidentally losing something we do support)

slf4j-simple would not support this indeed.

> * log4j plugins. Similar to above - like how we encourage users to plugin for something like secondary sampling, log4j is nice and pluggable too. Is this something we don't support?

no plugins in slf4j-simple.

> * stability? No clue on this, but I imagine the number of servers that use slf4j-simple in the wild in production is close to zero, wonder if we'll run into something unexpected

that might be a bit speculative, simple logging needs in a more constrained environment (rasp-pi etc) would be a good usecase for this.

> 
> Just want to vet the impact of the change

and thanks for that !

codefromthecrypt:
@jorgheymans no pressure, but what's the status on this?

jorgheymans:
Managed to do most of it locally, then ran out of spare time this week. So
95% were just needed to do zipkin-server to bridge slf4j back to log4j...

Op do 10 okt. 2019 20:36 schreef Adrian Cole <notifications@github.com>:

zipkin/pom.xml
jorgheymans: done

jorgheymans:
@adriancole ready for another review, sorry for the delay. Let me know if anything is not clear or uncertain. Basically everything except core depends on sfl4j-api now, and in zipkin-server log4j-starter is still used. The goal of reducing server size is no longer relevant in this case. But it should be easier to bridge everything to a new logging impl, though it's not a simple matter of swapping out dependencies.

PR #2831, opened 2019-10-02
Migrates zipkin docker image building to a single multi stage build instead of multiple

anuraaga:
Dockerfiles.

Currently, we have three very similar builds running to create docker images, zipkin, zipkin-ui, zipkin-builder. They all more or less run `mvn package`, possibly with some tweaks. Even with good caching, these have a lot of overlap and waste free build resources. As we're also looking at a possible zipkin:slim variant to push, this would increase to four and is just not a good use of build resources.

Now, instead of a build triggered for each of these images, there is only one build in the zipkin repo - it uses a Dockerfile that is configured for multiple images and builds and pushes them all. Using a single Dockerfile is still convenient for users to build images locally. And adding a slim variant will be quite simple, just picking up the `-slim` classifier jar as we do `-exec` right now in its own stage.


docker/hooks/post_build
```diff
@@ -0,0 +1,15 @@
...
+
+# This hook is called with the current directory set to the same as the Dockerfile, so we go back
+# to top level.
+cd ..
+
+# We always build an image containing nginx and zipkin-lens in addition to the default image of zipkin-server
+echo Building zipkin-ui
+docker build -f $DOCKERFILE_PATH -t openzipkin/zipkin-ui:$DOCKER_TAG --target zipkin-ui .
+
+if [[ "$DOCKER_TAG" == "master" ]]; then
```
devinsba, line 11: Should we move this to the build hook before the image build? That way the build always starts with an up to date image

devinsba (commented):
I haven't gotten to play with their build system too much so I might be completely off base

docker/hooks/post_build
anuraaga: The first build command goes through all steps in order, so it will have built this and the lens image as intermediate steps. We have to run build again for them so they get tagged, but as the build has happened already these docker build commands in the `post_build` are all served from cache and don't take any time.

So the order of building these doesn't matter so much, but I think it's a bit clearer that both of these are secondary artifacts by having them in `post_build`

codefromthecrypt (approved):
LGTM these are the "master" tags so in any case progress is good and small adjustments can occur later.

Thanks for the refactoring!

Issue #2829, opened 2019-10-02
Blank page when click detail trace at Lens Zipkin


thanhdongnguyen:
I use Zipkin Plugin at Kong Gateway to send data to Zipkin server.
But when i click to detail a trace, it is show blank page and i open Inspect -> Console, it have error: `Uncaught (in promise) TypeError: Cannot read property 'length' of undefined`

This is my json detail of a trace:
```json
[...]

PR #2834, opened 2019-10-07
Makes Lens more tolerant of missing localEndpoint.serviceName or .name

codefromthecrypt:
See https://github.com/openzipkin/zipkin/issues/2829

### DIFF
--- a/zipkin-lens/src/test/data/malformed.json
+++ b/zipkin-lens/src/test/data/malformed.json
@@ -1,0 +1,82 @@
+[
+  {
+    "traceId": "6cf5624e3452d1bc202a57dad35921c0",
+    "parentId": "d65495313f944801",
+    "id": "d5cb70daead478dc",
+    "name": "kong.header_filter",
+    "timestamp": 1570005203864000
+  },
+  {
+    "traceId": "6cf5624e3452d1bc202a57dad35921c0",
+    "parentId": "d65495313f944801",
+    "id": "80b8efd52d66af27",
+    "name": "kong.body_filter",
+    "timestamp": 1570005203864000
+  },
+  {
+    "traceId": "6cf5624e3452d1bc202a57dad35921c0",
+    "parentId": "ed395cb866d02665",
+    "id": "2a49858b1f1d506b",
+    "name": "kong.rewrite",
+    "timestamp": 1570005203829000
+  },
+  {
+    "traceId": "6cf5624e3452d1bc202a57dad35921c0",
+    "parentId": "d65495313f944801",
+    "id": "3e53779bdde7c111",
+    "name": "kong.access",
+    "timestamp": 1570005203829000,
+    "duration": 16000
+  },
+  {
+    "traceId": "6cf5624e3452d1bc202a57dad35921c0",
+    "parentId": "ed395cb866d02665",
+    "id": "d65495313f944801",
+    "kind": "CLIENT",
+    "name": "kong.proxy",
+    "timestamp": 1570005203829000,
+    "duration": 35000,
+    "localEndpoint": {
+      "serviceName": "rewards"
+    },
+    "remoteEndpoint": {
+      "port": 80
+    },
+    "tags": {
+      "kong.route": "aaa20d62-813a-428f-aa40-447324cd30af",
+      "kong.service": "872eae11-4571-490b-86d2-59834b93f870",
+      "peer.hostname": ""
+    }
+  },
+  {
+    "traceId": "6cf5624e3452d1bc202a57dad35921c0",
+    "id": "ed395cb866d02665",
+    "kind": "SERVER",
+    "name": "kong.request",
+    "timestamp": 1570005203829000,
+    "duration": 35000,
+    "remoteEndpoint": {
+      "port": 3771
+    },
+    "tags": {
+      "component": "kong",
+      "http.method": "GET",
+      "http.status_code": "200",
+      "http.url": "",
+      "kong.node.id": "37714786-d34b-4513-a459-8020a1f917de"
+    }
+  },
+  {
+    "traceId": "6cf5624e3452d1bc202a57dad35921c0",
+    "parentId": "d65495313f944801",
+    "id": "862d0f978033cc53",
+    "name": "kong.balancer",
+    "timestamp": 1570005203845000,
+    "remoteEndpoint": {
+      "port": 80
+    },
+    "tags": {
+      "kong.balancer.try": "1"
+    }
+  }
+]
--- a/zipkin-lens/src/components/DiscoverPage/TracesTab/TracesTableRow.jsx
+++ b/zipkin-lens/src/components/DiscoverPage/TracesTab/TracesTableRow.jsx
@@ -23,9 +23,24 @@ import grey from '@material-ui/core/colors/grey';
 import { fade } from '@material-ui/core/styles/colorManipulator';
 
 import ServiceBadge from '../../Common/ServiceBadge';
-import { rootServiceAndSpanName } from '../../../zipkin';
+import { getServiceName } from '../../../zipkin';
 import { traceSummaryPropTypes } from '../../../prop-types';
 import { theme } from '../../../colors';
+
+export function rootServiceAndSpanName(root) {
+  const { span } = root;
+  if (span) {
+    const serviceName = getServiceName(span.localEndpoint) || getServiceName(span.remoteEndpoint);
+    return {
+      serviceName: serviceName || 'unknown',
+      spanName: span.name || 'unknown',
+    };
+  }
+  return {
+    serviceName: 'unknown',
+    spanName: 'unknown',
+  };
+}
 
 const propTypes = {
   traceSummary: traceSummaryPropTypes.isRequired,
--- a/zipkin-lens/src/components/DiscoverPage/TracesTab/TracesTableRow.test.js
+++ b/zipkin-lens/src/components/DiscoverPage/TracesTab/TracesTableRow.test.js
@@ -14,7 +14,70 @@
 import React from 'react';
 import { mount } from 'enzyme';
 
-import { TracesTableRowImpl } from './TracesTableRow';
+import { TracesTableRowImpl, rootServiceAndSpanName } from './TracesTableRow';
+import { SpanNode } from '../../../zipkin/span-node';
+
+describe('rootServiceAndSpanName', () => {
+  it('should return serviceName and spanName of the span', () => {
+    const root = new SpanNode({
+      traceId: '1',
+      id: '1',
+      name: 'get',
+      localEndpoint: { serviceName: 'frontend' },
+    });
+
+    expect(rootServiceAndSpanName(root)).toEqual({
+      serviceName: 'frontend',
+      spanName: 'get',
+    });
+  });
+
+  it('should return unknown serviceName when missing localEndpoint', () => {
+    const root = new SpanNode({ traceId: '1', id: '1', name: 'get' });
+
+    expect(rootServiceAndSpanName(root)).toEqual({
+      serviceName: 'unknown',
+      spanName: 'get',
+    });
+  });
+
+  it('should return unknown serviceName when missing localEndpoint.serviceName', () => {
+    const root = new SpanNode({
+      traceId: '1', id: '1', name: 'get', localEndpoint: { },
+    });
+
+    expect(rootServiceAndSpanName(root)).toEqual({
+      serviceName: 'unknown',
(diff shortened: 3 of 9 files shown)


devinsba:
Should we add the offending trace to the folder of sample data?

codefromthecrypt:
horrible-trace.json? :D

codefromthecrypt:
Now unknown instead of undefined in trace list

Screenshot 2019-10-07 at 11 35 06 AM

Now, no more white screen rendering this really bad trace:

Screenshot 2019-10-07 at 11 34 43 AM

codefromthecrypt:
PS it is tempting to make the span-cleaner fill empty names and such, but it actually would spread debt into other things like clock skew and dependency linker which looks for presence of service names. Instead, I made it a goal to make the white screen go away.

codefromthecrypt:
PPS here, Kong should have used an annotation, not an empty span except for timestamp and name. While definitely bad data, I wasn't able to figure out how to get the warning about "no rows found" to go away.

Screenshot 2019-10-07 at 11 39 38 AM

zipkin-lens/src/test/data/malformed.json
```diff
@@ -0,0 +1,82 @@
+[
+  {
+    "traceId": "6cf5624e3452d1bc202a57dad35921c0",
+    "parentId": "d65495313f944801",
+    "id": "d5cb70daead478dc",
+    "name": "kong.header_filter",
+    "timestamp": 1570005203864000
```
codefromthecrypt, line 7: this should have been an annotation

zipkin-lens/src/test/data/malformed.json
```diff
@@ -0,0 +1,82 @@
...
+  },
+  {
+    "traceId": "6cf5624e3452d1bc202a57dad35921c0",
+    "parentId": "d65495313f944801",
+    "id": "80b8efd52d66af27",
+    "name": "kong.body_filter",
+    "timestamp": 1570005203864000
```
codefromthecrypt, line 14: this should have been an annotation

zipkin-lens/src/test/data/malformed.json
```diff
@@ -0,0 +1,82 @@
...
+  },
+  {
+    "traceId": "6cf5624e3452d1bc202a57dad35921c0",
+    "parentId": "ed395cb866d02665",
+    "id": "2a49858b1f1d506b",
+    "name": "kong.rewrite",
+    "timestamp": 1570005203829000
```
codefromthecrypt, line 21: this should have been an annotation

zipkin-lens/src/test/data/malformed.json
```diff
@@ -0,0 +1,82 @@
...
+  },
+  {
+    "traceId": "6cf5624e3452d1bc202a57dad35921c0",
+    "parentId": "d65495313f944801",
+    "id": "3e53779bdde7c111",
+    "name": "kong.access",
```
codefromthecrypt, line 27: this is missing localEndpoint.serviceName or this should be an annotation, and the duration moved to where it is missing..

zipkin-lens/src/test/data/malformed.json
```diff
@@ -0,0 +1,82 @@
...
+    "localEndpoint": {
+      "serviceName": "rewards"
+    },
+    "remoteEndpoint": {
+      "port": 80
+    },
+    "tags": {
+      "kong.route": "aaa20d62-813a-428f-aa40-447324cd30af",
+      "kong.service": "872eae11-4571-490b-86d2-59834b93f870",
+      "peer.hostname": ""
```
codefromthecrypt, line 48: don't add empty hostname

zipkin-lens/src/test/data/malformed.json
```diff
@@ -0,0 +1,82 @@
...
+    "parentId": "ed395cb866d02665",
+    "id": "d65495313f944801",
+    "kind": "CLIENT",
+    "name": "kong.proxy",
+    "timestamp": 1570005203829000,
+    "duration": 35000,
+    "localEndpoint": {
+      "serviceName": "rewards"
+    },
+    "remoteEndpoint": {
```
codefromthecrypt, line 42: suspiciously missing ip address

zipkin-lens/src/test/data/malformed.json
```diff
@@ -0,0 +1,82 @@
...
+    }
+  },
+  {
+    "traceId": "6cf5624e3452d1bc202a57dad35921c0",
+    "id": "ed395cb866d02665",
+    "kind": "SERVER",
+    "name": "kong.request",
+    "timestamp": 1570005203829000,
+    "duration": 35000,
+    "remoteEndpoint": {
```
codefromthecrypt, line 58: suspiciously missing IP address

zipkin-lens/src/test/data/malformed.json
```diff
@@ -0,0 +1,82 @@
...
+      "kong.route": "aaa20d62-813a-428f-aa40-447324cd30af",
+      "kong.service": "872eae11-4571-490b-86d2-59834b93f870",
+      "peer.hostname": ""
+    }
+  },
+  {
+    "traceId": "6cf5624e3452d1bc202a57dad35921c0",
+    "id": "ed395cb866d02665",
+    "kind": "SERVER",
+    "name": "kong.request",
```
codefromthecrypt, line 55: bad span name. prefer route-based ex `get /users/{userId}` or even `get` as kong is already tagged

zipkin-lens/src/test/data/malformed.json
```diff
@@ -0,0 +1,82 @@
...
+    "name": "kong.access",
+    "timestamp": 1570005203829000,
+    "duration": 16000
+  },
+  {
+    "traceId": "6cf5624e3452d1bc202a57dad35921c0",
+    "parentId": "ed395cb866d02665",
+    "id": "d65495313f944801",
+    "kind": "CLIENT",
+    "name": "kong.proxy",
```
codefromthecrypt, line 36: bad span name. prefer route-based ex get /users/{userId} or even get as kong isn't helpful

what you are describing here, seems to be better as a tag

zipkin-lens/src/test/data/malformed.json
```diff
@@ -0,0 +1,82 @@
...
+    "timestamp": 1570005203829000,
+    "duration": 35000,
+    "remoteEndpoint": {
+      "port": 3771
+    },
+    "tags": {
+      "component": "kong",
+      "http.method": "GET",
+      "http.status_code": "200",
+      "http.url": "",
```
codefromthecrypt, line 65: don't add empty url

zipkin-lens/src/test/data/malformed.json
```diff
@@ -0,0 +1,82 @@
...
+    "id": "ed395cb866d02665",
+    "kind": "SERVER",
+    "name": "kong.request",
+    "timestamp": 1570005203829000,
+    "duration": 35000,
+    "remoteEndpoint": {
+      "port": 3771
+    },
+    "tags": {
+      "component": "kong",
```
codefromthecrypt, line 62: conventionally zipkin uses "lc" for "local component"

zipkin-lens/src/test/data/malformed.json
```diff
@@ -0,0 +1,82 @@
...
+      "http.url": "",
+      "kong.node.id": "37714786-d34b-4513-a459-8020a1f917de"
+    }
+  },
+  {
+    "traceId": "6cf5624e3452d1bc202a57dad35921c0",
+    "parentId": "d65495313f944801",
+    "id": "862d0f978033cc53",
+    "name": "kong.balancer",
+    "timestamp": 1570005203845000,
```
codefromthecrypt, line 74: missing duration. probably better as an annotation named `kong.balancer.try1` as that should be bounded, unless there's a chance of infinite tries. That or make this the client span if this is the actual wire request.. hard to tell what's going on (might need a wiki)

codefromthecrypt (commented):
@bungle @jeremyjpj0916 the comments on this part of the review are some hints about data problems for https://github.com/Kong/kong-plugin-zipkin/issues/49 Hope they help.

anuraaga (approved):

codefromthecrypt:
verified one of the envoy traces renders properly eventhough it is missing the local service name:

Screenshot 2019-10-08 at 8 01 50 AM

zipkin-lens/src/test/data/malformed.json
kikito: Could you expand on this a little bit? The [Zipkin docs](https://zipkin.io/pages/instrumenting#timestamps-and-durations) make it very clear that the `timestamp` field exists, and also say:

> The bottom-line is that choosing not to record Span.timestamp and duration will result in less accurate data and less functionality. Since it is very easy to record these authoritatively before reporting, all Zipkin instrumentation should do it or ask someone to help them do it.

Wouldn't removing the `timestamp` field making Zipkin less accurate?

Also: can I name an annotation `timestamp`?

zipkin-lens/src/test/data/malformed.json
kikito: By looking at the examples in Zipkin's source code, I think this means that the span should look like so:

```
  {
    "traceId": "6cf5624e3452d1bc202a57dad35921c0",
    "parentId": "d65495313f944801",
    "id": "3e53779bdde7c111",
    "name": "kong.access",
    "localEndpoint": {
      "serviceName": "<name of the Kong Service matched for this Request>",
    },
    "timestamp": 1570005203829000,
    "duration": 16000
  },
```

I think I can try to add that. But I have questions:

* Is `localEndpoint.serviceName` also needed in the other spans, or just on this one?
* What does "the duration moved to where it is missing" mean? That the other spans should also have a duration? That the `duration` field on this particular span should be put in a different place on this span (where)? Or is it both?

zipkin-lens/src/test/data/malformed.json
kikito: Does this apply to the other span names as well? (`kong.access`, `kong.header_filter`, `kong.body_filter`, `kong.rewrite`, `kong.access`) or is it just `kong.proxy`?

kikito (commented):
Hi @adriancole, I am trying to fix the issues reported on the Kong plugin. Thanks for reporting the problems to us. I left some questions on this PR, do you think you could answer those?

zipkin-lens/src/test/data/malformed.json
james-callahan: These are spans in Kong; I recall that when they execute in under a millisecond then they appear as a 0-length span in zipkin.

zipkin-lens/src/test/data/malformed.json
codefromthecrypt: I think what's missing here is that a span can include annotations.. annotations have a required timestamp. So instead of making a sub-span that only includes a name and a timestamp, make an annotation with that name and timestamp.

ex 
```json
{
    "traceId": "6cf5624e3452d1bc202a57dad35921c0",
    "parentId": "ed395cb866d02665",
    "id": "d65495313f944801",
    "kind": "CLIENT",
    "name": "get /users/{userId}",
    "timestamp": 1570005203829000,
    "duration": 35000,
--snip--
    "annotations": [
       {"name": "kong.header_filter", "timestamp": 1570005203864000}
     ]
```

zipkin-lens/src/test/data/malformed.json
codefromthecrypt: minimum duration is 1us, so prefer to use that vs rounding down to zero, especially if precision is millis (1000us) on the kong side. Zero implies incomplete.

The main thing is to differentiate annotations (timestamped events on a timeline) from spans (some wrapped thing with a duration).

zipkin-lens/src/test/data/malformed.json
codefromthecrypt: the client span should be complete, that's the most important one, but all spans should be complete. The subspans, some of them have duration. It was unclear to me why these are modeled as spans vs events.

all spans should have the exact same localEndpoint value if coming from the same node. Most of the time folks just take the link local IP, the advertised port and service name and treat that as a constant.

zipkin-lens/src/test/data/malformed.json
codefromthecrypt: the client span should be semantically correct for http. It is better to have a coherent client span than any of the other spans mentioned. As you noticed, they didn't render and actually made the diagram more confusing vs if they didn't exist at all. My advice is probably use one client span and render the rest with annotations if it is commonly the case you cannot record a duration of precision less than 1ms.

PR #2839, opened 2019-10-12
Changes the banner to include the Zipkin version and hash

codefromthecrypt:
This removes property replacement infrastructure to make the banner
"pop" faster. Along the way, it removes the mostly irrelevant Spring
Boot version in favor of the Zipkin version and hash from the /info
endpoint.

Screenshot 2019-10-13 at 3 38 48 AM

cc @making


zipkin-server/src/main/java/zipkin2/server/internal/banner/ZipkinBanner.java
```diff
@@ -13,35 +13,44 @@
...
+  static final AnsiElement ZIPKIN_ORANGE = new AnsiElement() {
+    @Override public String toString() {
+      return "38;5;208"; // Ansi 256 color code 208 (orange)
+    }
+  };
 
-  public ZipkinBanner() {
-    super(new ClassPathResource("zipkin.txt"));
-  }
+  final Resource resource = new ClassPathResource("zipkin.txt");
```
anuraaga: Can the banner be printed twice? I guess it can be a local. Or otherwise static

anuraaga (approved):

zipkin-server/src/main/java/zipkin2/server/internal/banner/ZipkinBanner.java
codefromthecrypt: sgtm

PR #2840, opened 2019-10-12
Disables auto-configuration by default to save 10% startup time

codefromthecrypt:
This disables auto-configuration, which implied unwrapping some Lazy
wiring.

As actuator doesn't have an entrypoint to load its classes when auto-
configuration is disabled, this cherry-picks ones we use. Unfortunately,
this creates a compile time dependency which we didn't have before on
actuator.. this trips up one of our tests. Maybe @dsyer knows a way
around this.

This achieved 10% quicker start using a relatively complex set of config. I intentionally used the most complex startup parameters to ensure everything works.

```bash
SELF_TRACING_ENABLED=true STORAGE_THROTTLE_ENABLED=true STORAGE_TYPE=elasticsearch java -jar zipkin.jar
```

Before, the best JVM time to ready was 3.162s and now it is 2.801.


codefromthecrypt:
PS I integration tested this in zipkin-gcp as mentioned.

zipkin-server/src/main/java/zipkin2/server/internal/ActuatorConfiguration.java
```diff
@@ -0,0 +1,54 @@
...
+ * using {@link EnableAutoConfiguration}.
+ *
+ * <p>This cherry-picks the most relevant endpoints for when actuator is in the classpath because
+ * we disable scanning in {@link ZipkinServer} by default.
+ */
+@Configuration
+@ConditionalOnClass({
+  ArmeriaSpringActuatorAutoConfiguration.class
+})
+@ImportAutoConfiguration({
```
codefromthecrypt: unfortunately I see no out-of-box way to specify the type names so as to avoid a compile dep. I suppose we could write one..

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
```diff
@@ -102,18 +94,18 @@
       // and default to a slightly longer timeout on the server to be able to handle these with
       // better error messages where possible.
       sb.requestTimeout(Duration.ofSeconds(11));
-    };
-  }
 
-  @Bean Consumer<MeterRegistry.Config> noAdminMetrics() {
```
codefromthecrypt, line 108: these consumers created a cycle which we didn't notice before because bytecode magic walked it out. I undid this as at the end of the day the config is mutable anyway cc @anuraaga @trustin

zipkin-server/src/test/java/zipkin2/server/internal/ui/ZipkinUiConfigurationTest.java
```diff
@@ -51,14 +48,11 @@ public void close() {
...
 
-  @Rule
-  public ExpectedException thrown = ExpectedException.none();
-
   @Test
   public void indexHtmlFromClasspath() {
     context = createContext();
 
-    assertThat(context.getBean(ZipkinUiConfiguration.class).indexHtml)
+    assertThat(context.getBean(ZipkinUiConfiguration.class))
```
codefromthecrypt: TODO: delete this as the below test covers it now

zipkin-server/src/main/java/zipkin/server/ZipkinServer.java
```diff
@@ -28,8 +45,23 @@
...
-    new SpringApplicationBuilder(ZipkinServer.class)
+    SpringApplicationBuilder builder = new SpringApplicationBuilder(ZipkinServer.class)
       .banner(new ZipkinBanner())
-      .properties("spring.config.name=zipkin-server").run(args);
+      .properties(
+        EnableAutoConfiguration.ENABLED_OVERRIDE_PROPERTY + "=false",
+        "spring.config.name=zipkin-server");
+
+    try {
+      Class<?> actuatorClass = Class.forName("zipkin.server.ActuatorConfiguration");
```
codefromthecrypt: the next step would be to switch the maven profile that excludes actuator to opt-out of compiling this

codefromthecrypt:
rewrote the approach cc @anuraaga

TODO: unit tests including that the normal spring conditions still hold (ex disabling an endpoint by config eventhough we enabled it)

zipkin-server/src/main/java/zipkin/server/ConditionallyImportActuator.java
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Actuator relies on classpath scanning, so we cannot choose one configuration entrypoint unless
+ * using {@link EnableAutoConfiguration}. We disable scanning in {@link ZipkinServer} by default so
+ * this conditionally loads actuator a different way.
+ *
+ * <p>This queries the property {@link #PROPERTY_NAME_ACTUATOR_INCLUDE} which cherry-picks the most
+ * relevant configuration for when actuator is in the classpath.
+ */
+final class ConditionallyImportActuator
```
codefromthecrypt: this is the new fancy-pants one

anuraaga: Maybe more conventional to name as noun `ActuatorConditionalImporter`.

zipkin-server/src/main/java/zipkin/server/ConditionallyImportActuator.java
```diff
@@ -0,0 +1,66 @@
...
+
+  @Override public void initialize(GenericApplicationContext context) {
+    ConfigurableEnvironment env = context.getEnvironment();
+    if ("false".equalsIgnoreCase(env.getProperty(PROPERTY_NAME_ACTUATOR_ENABLED))) {
+      LOG.debug("skipping actuator as it is disabled");
+      return;
+    }
+
+    // We are in a configurable environment, so bind the include property so that we can use it.
+    String[] includes = Binder.get(env).bind(PROPERTY_NAME_ACTUATOR_INCLUDE, String[].class).get();
```
anuraaga: Why do we `getProperty` above but `bind` here? Maybe a bit more comment.

zipkin-server/src/main/java/zipkin/server/ConditionallyImportActuator.java
```diff
@@ -0,0 +1,66 @@
...
+    if (includes == null) return;
+
+    try {
+      context.registerBean(Class.forName(ACTUATOR_IMPL_CLASS));
+    } catch (Exception e) {
+      LOG.debug("skipping actuator as implementation is not available", e);
```
anuraaga: Should we skip the below for loop in this case too?

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinServerConfiguration.java
anuraaga: Not using `Consumer` seems fine but for enlightenment what's the cycle? All it does is call `deny` which is a static method and seems non-cyclical.

anuraaga (approved):

zipkin-server/src/main/java/zipkin/server/ConditionallyImportActuator.java
codefromthecrypt: yep. good catch

codefromthecrypt:
> Not using Consumer seems fine but for enlightenment what's the cycle? All it does is call deny which is a static method and seems non-cyclical.

The cycle was in where we provided PrometheusMeterRegistry. The registry type was needed in call sites that themselves defined consumers. Consumers being required to configure the `PrometheusMeterRegistry` was the cycle.

codefromthecrypt:
yay it is green, which means our invoker tests work. This proves that actuator is not compile required anymore. I still have some tests to write and also need to double-check performance.

zipkin-server/src/main/java/zipkin/server/ConditionallyImportActuator.java
codefromthecrypt: Good point, it was not obvious.. maybe still isn't. PTAL now

anuraaga: LGTM

zipkin-server/src/main/java/zipkin/server/ActuatorConditionalImporter.java
```diff
@@ -0,0 +1,73 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.context.properties.bind.Binder;
+import org.springframework.context.ApplicationContextInitializer;
+import org.springframework.context.support.GenericApplicationContext;
+import org.springframework.core.env.ConfigurableEnvironment;
+
+/**
+ * Actuator relies on classpath scanning, so we cannot choose one configuration entrypoint unless
```
dsyer: Nitpick. There is no classpath scanning involved in autoconfiguration. I would like to see the wording changed because enough people are freaked out by the concept of classpath scanning, and we wouldn't like inaccurate descriptions to get attached. I'm not sure what to suggest because I don't know what the intent of this paragraph is really. Is it just autoconfiguration that is disabled and relied on?

zipkin-server/src/main/java/zipkin/server/ActuatorConditionalImporter.java
```diff
@@ -0,0 +1,73 @@
...
+ * using {@link EnableAutoConfiguration}. We disable scanning in {@link ZipkinServer} by default so
+ * this conditionally loads actuator a different way.
+ *
+ * <p>This queries the property {@link #PROPERTY_NAME_ACTUATOR_INCLUDE} which cherry-picks the most
+ * relevant configuration for when actuator is in the classpath.
+ */
+final class ActuatorConditionalImporter
+  implements ApplicationContextInitializer<GenericApplicationContext> {
```
dsyer: This feels like an overly complex solution when we are already using `@Configuration` and `@Conditional` is available. Couldn't you just write a `@Configuration` that did this?

zipkin-server/src/main/java/zipkin/server/ZipkinServer.java
```diff
@@ -13,12 +13,27 @@
...
+import org.springframework.boot.SpringBootConfiguration;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.builder.SpringApplicationBuilder;
 import zipkin2.server.internal.EnableZipkinServer;
 import zipkin2.server.internal.banner.ZipkinBanner;
 
-@SpringBootApplication
+/**
+ * This adds the {@link EnableAutoConfiguration} annotation, but disables it by default to save
+ * startup time. A supported integration will need to explicitly enable auto-configuration like so,
```
dsyer: This seems fragile to me, and might break unexpectedly if the main app comes to rely on the fact that it has no autoconfiguration. Maybe there is a better way using a custom autoconfiguration annotation (so that supported integrations only need to add that annotation).

zipkin-server/src/main/java/zipkin/server/ActuatorConditionalImporter.java
codefromthecrypt: I think it does rely on it, but it could be changed to not. For example `BeansEndpointAutoConfiguration` is only found through META-INF spring factories, which implies scanning, right?

zipkin-server/src/main/java/zipkin/server/ZipkinServer.java
codefromthecrypt: the implementation has changed. our modules no longer rely on autoconfiguration at all. most of the time I wouldn't expect anyone to enable auto-configuration.

zipkin-server/src/main/java/zipkin/server/ActuatorConditionalImporter.java
codefromthecrypt: The main problem in actuator is the types are needed. we have another pull request that makes it possible to remove actuator (at compile time meaning actuator is not present)

Like mentioned before there's no way I know of now anyway to get actuator to load when you turn off autoconfiguration.

zipkin-server/src/main/java/zipkin/server/ActuatorConditionalImporter.java
dsyer: No. Not at all. The use of `spring.factories` is more or less entirely to prevent classpath scanning. You don't need to scan if you have a list of classes to import via a properties file.

zipkin-server/src/main/java/zipkin/server/ActuatorConditionalImporter.java
dsyer: I expect you would have to explicitly list all the configurations you need. But that's a sacrifice you already made, and you could do it in a class that was marked as `@Conditional` right?

zipkin-server/src/main/java/zipkin/server/ActuatorConditionalImporter.java
codefromthecrypt: by changed not to, I mean actuator could have an entrypoint type that includes imports of all the autoconfiguration it manages instead of having each loaded by meta-inf. As each of the endpoints is already subject to secondary config based on enabled properties this could work. Ex this would change this to a much more manageable thing where we look for a class name once vs N* endpoints defined. OTOH, I don't think upstream probably cares to optimize for this.

Meanwhile it is important to keep actuator optional, and this is basically an alternative to looking up the spring factories that have package name prefix related to actuator.

codefromthecrypt: I'll remove the comment about "spring.factories"

meanwhile we get constant complaints about startup time. team members here's teams complain, windows starts in 10s and I noticed as you mentioned actuator takes a notable percentage off the startup. We can pay attention to what we want but meanwhile some mass of effort is spent, maybe not on classpath scanning, but whatever that effort is, less of it is occurring now.

dsyer: Boot 2.2 enables only 2 actuators by default. It's much faster. Did you fold that in yet? Users would have to explicitly enable all the others that they want to expose. Seems like a legitimate task for a Zipkin server operator.

codefromthecrypt:
@dsyer I rewrote the rationale in the types to try and avoid saying that things are slow etc.

zipkin-server/src/main/java/zipkin/server/ActuatorConditionalImporter.java
codefromthecrypt: If I had a type that was conditional, I'd still need a way to load the types mentioned. For example, bean methods are not used here, we are doing things like imports, and imports requires strict types, not class names. ex  I could if there was somethign like `@Import(name="fully.qualified.type")` but I scrubbed and couldn't find one. Looking at your bench repo literally was how I discovered this other way. I'm glad I found that as before it was a hard stop.

dsyer: I think what you are looking for is `ImportSelector` or `ImportBeanDefinitionRegistrar` (both mentioned in the javadocs for `@Import`).

codefromthecrypt:
ps I would like to be honest while avoiding loaded terms. otherwise someone could blindly undo work as they dont understand the underlying problem. in hindsight, "classpath scanning" seems to be emotive. The main thing happening here is disabling autoconfiguration, but that is a what not a why. I suspect there is a way to phrase the why.. for example this stops having to weed through things discovered implicitly via the classpath. for example the factories are primarily discovered via the classpath which is why I used the term scanning though I admit I forgot that was a historical term for something more deep. at any rate some suggestion of a non scary way and also honest would be appreciated.

zipkin-server/src/main/java/zipkin/server/ActuatorConditionalImporter.java
codefromthecrypt: we haven't folded 2.2 in.

Custom servers are unsupported and we define custom as anything not defined in env variables. I would consider expectations around autoconfiguration tuning in the realm of that unsupported area though folks obviously can self support that. In other words, our operations are not defined as "a usual spring boot app" as that implies a support load we cannot afford.

zipkin-server/src/main/java/zipkin/server/ActuatorConditionalImporter.java
codefromthecrypt: I'll make a note that someone can figure out how to use those things and refactor

codefromthecrypt:
added tests to 100% coverage and moved rationale around with a note that someone can refactor later as they wish. Since tests are there, it should be possible to do the refactoring safely should one desire. As for me, I'm over budget on this so calling it a day.

dsyer:
> some suggestion of a non scary way and also honest would be appreciated

I think the current wording ("disabling auto-configuration") works.

It might be a good idea to quantify the "significant improvements" though. The Github issue has the figure 10% in the title, and I think that's reasonable. Why not put it in the readme/rationale as well?

codefromthecrypt:
loaded a complex setup from zipkin-aws (which is customized to use this) and checked the logs and also the health endpoint.

complex config as it loads both a storage and a collector, using modules and properties loader.
```bash
SQS_QUEUE_URL=https://ap-southeast-1.queue.amazonaws.com/012345678901/zipkin SELF_TRACING_ENABLED=true ES_AWS_DOMAIN=zipkin-ha STORAGE_TYPE=elasticsearch    java  -Dloader.path='elasticsearch-aws.jar,elasticsearch-aws.jar!/lib,sqs.jar,sqs.jar!/lib'  -Dspring.profiles.active=sqs,elasticsearch-aws     -cp zipkin.jar     org.springframework.boot.loader.PropertiesLauncher  --logging.level.zipkin2=DEBUG
```

log shows the modules and also autoconfigure loaded, and show no errors trying to. I also verified this in the health endpoint.
```
2019-10-14 22:06:28.512 DEBUG 99268 --- [           main] z.s.i.ZipkinModuleImporter               : attempting to load modules: [elasticsearch-aws, sqs]
2019-10-14 22:06:28.652 DEBUG 99268 --- [           main] z.s.i.ZipkinActuatorImporter             : attempting to load actuator configuration: [org.springframework.boot.actuate.autoconfigure.beans.BeansEndpointAutoConfiguration, org.springframework.boot.actuate.autoconfigure.condition.ConditionsReportEndpointAutoConfiguration, org.springframework.boot.actuate.autoconfigure.context.properties.ConfigurationPropertiesReportEndpointAutoConfiguration, org.springframework.boot.actuate.autoconfigure.endpoint.EndpointAutoConfiguration, org.springframework.boot.actuate.autoconfigure.env.EnvironmentEndpointAutoConfiguration, org.springframework.boot.actuate.autoconfigure.management.HeapDumpWebEndpointAutoConfiguration, org.springframework.boot.actuate.autoconfigure.logging.LoggersEndpointAutoConfiguration, org.springframework.boot.actuate.autoconfigure.management.ThreadDumpEndpointAutoConfiguration]
```

dsyer:
Also noteworthy: benchmark results indicate that Spring Boot 2.2 is not significantly faster for this "stripped down" version of the app, but only if launched from the main class. The `JarLauncher` (and hence probably the `PropertiesLauncher` that is used in the docker container for zipkin-server) is *much* faster in Spring Boot 2.2. If we are going to continue to use that, it would be worthwhile to upgrade Spring Boot Maven plugin, even if the underlying library is held back to 2.1 for now. (BTW I tested with 2.2 as a library as well - had to explicitly pin the version for actuator jars, but it worked.)

dsyer:
See #2841

codefromthecrypt:
since we don't support custom servers we can upgrade anything at any time. indeed worthwhile to check into the property loading.

PS The amount of time savings depends greatly on which configurations are used. Ad-hoc, the more the more savings, but I don't have a matrix. I will make a comment to such effect.

If only using the stock zipkin in-memory, you won't see anything much unless you set `--zipkin.internal.actuator.enabled=false`. Here's a best of 10 with it disabled:

Best of 10 this commit just `java -jar zipkin.jar`
```
1.854 seconds (JVM running for 2.578)
```

Best of 10 this commit `java -jar zipkin.jar --zipkin.internal.actuator.enabled=false`
```
1.681 seconds (JVM running for 2.434)
```

codefromthecrypt:
best in several for the elasticsearch use case mentioned in the description is about 120ms savings even with actuator left on.

It is still on plate to look at docker setup to see how much impact, but as there are still a lot of things from the startup bench talk still not implemented, will sync this for a bit. We can revisit things decoupled as we go along.

codefromthecrypt:
PPS the build from the skipActuator profile runs the same speed as skipping via the internal runtim parameter (Eventhough skipping compilation and inclusion reduces the size of the distribution)

codefromthecrypt:
going to merge this as there's much left to do and I think these lines of code now rank as some of the most analyzed in the repo :P

PR #2841, opened 2019-10-14
Use Boot 2.2 for building the exec jar (but not as a dependency)

dsyer:
Startup time is down from 2.2 to 1.6 seconds in my benchmarks
(nearly 25% faster).

Spring Boot 2.2 is only available as a RC currently so that's what I used here. I guess we might want to wait for the GA release.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,6 +61,7 @@
     <jooq.version>3.12.1</jooq.version>
     <micrometer.version>1.3.0</micrometer.version>
     <spring-boot.version>2.1.9.RELEASE</spring-boot.version>
+    <spring-boot-plugin.version>2.2.0.RC1</spring-boot-plugin.version>
     <!-- MySQL connector is GPL, even if it has an OSS exception.
          https://www.mysql.com/about/legal/licensing/foss-exception/
 
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -469,10 +469,11 @@
       <plugin>
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-maven-plugin</artifactId>
-        <version>${spring-boot.version}</version>
+        <version>${spring-boot-plugin.version}</version>
         <configuration>
           <mainClass>zipkin.server.ZipkinServer</mainClass>
           <executable>true</executable>
+		  <layout>ZIP</layout>
         </configuration>
         <executions>
           <execution>


codefromthecrypt:
thanks, sounds promising.

as it is going on a 15hr day so far, I'm gonna look more tomorrow. this sort of work is very time consuming.

codefromthecrypt:
I tried this after adding the repo. I restarted it many times and couldn't find any difference between before and after. I triple checked the console output etc, including looking inside the jar.

Note that when you have network on (wifi enabled) something inside the app is more volatile in startup times. It could possibly be unrelated to your experience. I noticed yesterday that starting could take 7+ seconds on a bad wifi and on my list to trace the bootstrap to see if any dependencies are doing nasty things like making outgoing connections.

codefromthecrypt:
closing this out, though thanks for the hint.

Our most analyzed start times come from docker, and there's no notable difference when launching from the main class in our setup of 2.1 vs 2.2. We've already pared down so much that this isn't terribly surprising.

once boot 2.2 is out, we'll surely upgrade.

dsyer:
Just checking, you were starting the app from the exploded archive? (I'm guessing not, otherwise your results are too different than mine.) Maybe I'm reading it wrong and that isn't how we start the docker container. The exploded archive case really is 25% faster with Spring Boot 2.2. Really. But not relevant if no-one is actually doing that.

codefromthecrypt:
our docker images was exploded formerly. you can verify, and at any rate the only change is to switch the version. just I spent a lot of time trying to prove 2.2 was faster and failed.

codefromthecrypt:
the main thing is that PropertiesLauncher is indeed a dog. we have no need to run that in our main image. So main class vs main class performance is near identical if you start several times. at least on my host.  changing other variables have decreased startup, just not switching this one.

dsyer:
PropertiesLauncher is not a dog in 2.2. That was the point. But if we really aren't using it, then it doesn't matter (I thought I saw it, but maybe looking in the wrong place).

codefromthecrypt:
2.1 I meant, but right we only need that when doing layered images. We don't need to in the base image.

codefromthecrypt:
indeed we accidentally used it in our docker image on the base image, but this is reverted. IOTW, the main thing was avoiding unnecessary use of `PropertiesLauncher` not switching versions. A bump in perf for the layers which may use `PropertiesLauncher` will be welcome, but that's not what folks are watching. They are watching our base image perf.

codefromthecrypt:
here's the related thing btw https://github.com/openzipkin/docker-zipkin/pull/229

PR #2844, opened 2019-10-15
enforce the fact that core should not have compile time dependencies

jorgheymans:
It's not so much to catch unsuspecting devs adding a dependency in core pom.xml itself, but rather to the parent pom dependencies section where it's less obvious that core will inherit these as well.

### DIFF
--- a/zipkin/pom.xml
+++ b/zipkin/pom.xml
@@ -46,7 +46,7 @@
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
       <!-- this dependency is shaded out -->
-      <optional>true</optional>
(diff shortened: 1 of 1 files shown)


jorgheymans:
Nope won't work, maven-shade requires dependencies to have runtime scope (hence gson being set to optional and not provided, and the enforcer plugin cannot target the optional dependencies because that's not even a scope. Closing this worthwhile idea :/

codefromthecrypt:
actually the shading of gson is something that has been problematic (it stains the classpath of tests which you'll notice in the other maven poms). It is on my list to remove anyway.

jorgheymans:
ok i will revive this then if needed.

PR #2845, opened 2019-10-16
Build master images of slim.

anuraaga:


docker/Dockerfile
```diff
@@ -43,6 +52,65 @@ EXPOSE 80
 
 CMD ["/usr/local/bin/nginx.sh"]
 
+#####
+# zipkin-server-slim - An image containing the slim distribution of Zipkin server.
```
codefromthecrypt: I would suggest zipkin-slim instead of adding server

codefromthecrypt (approved):
nice. thanks!

PR #2846, opened 2019-10-16
Add release dockerfile.

anuraaga:
Since we want our release images to use the same artifacts as published to Maven, we need to have a separate `Dockerfile` for it from the master build. Luckily it's content isn't all that different.

I'll reconfigure Docker Hub to use this dockerfile for tag pushes. Then we should be ready to finish the migration.

### DIFF
--- a/docker/Dockerfile.release
+++ b/docker/Dockerfile.release
@@ -9,10 +9,10 @@ ARG version
 # since it's where we publish to.
 
 RUN mvn org.apache.maven.plugins:maven-dependency-plugin:get \
-  -DremoteRepositories=bintray::::http://dl.bintray.com/openzipkin/maven -Dtransitive=false \
+  -DremoteRepositories=bintray::::https://dl.bintray.com/openzipkin/maven -Dtransitive=false \
   -Dartifact=io.zipkin:zipkin-server:${version}:jar:exec
 RUN mvn org.apache.maven.plugins:maven-dependency-plugin:get \
-  -DremoteRepositories=bintray::::http://dl.bintray.com/openzipkin/maven -Dtransitive=false \
+  -DremoteRepositories=bintray::::https://dl.bintray.com/openzipkin/maven -Dtransitive=false \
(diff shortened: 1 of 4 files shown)


docker/Dockerfile.release
```diff
@@ -0,0 +1,98 @@
...
+# Add environment settings for supported storage types
+COPY --from=built --chown=zipkin /zipkin-slim/ /zipkin/
+COPY --chown=zipkin docker/zipkin/ /zipkin/
+WORKDIR /zipkin
+
+RUN ln -s /busybox/* /bin
+
+USER zipkin
+
+EXPOSE 9410 9411
```
codefromthecrypt: I think we actually removed thrift from slim

```suggestion
EXPOSE 9411
```

docker/Dockerfile.release
```diff
@@ -0,0 +1,98 @@
...
+#####
+# zipkin-slim - An image containing the slim distribution of Zipkin server.
+#####
+
+FROM openzipkin/jre-full:11.0.4-11.33 as zipkin-slim
+LABEL MAINTAINER Zipkin "https://zipkin.io/"
+
+# Use to set heap, trust store or other system properties.
+ENV JAVA_OPTS -Djava.security.egd=file:/dev/./urandom
+# 3rd party modules like zipkin-aws will apply profile settings with this
```
codefromthecrypt: remove this as slim doesn't support it. I know I forgot to remove it upstream :P

docker/Dockerfile.release
```diff
@@ -0,0 +1,98 @@
...
+
+FROM openzipkin/zipkin-builder as built
+
+ARG version
+
+# Download jars using Maven. It will try to resolve the artifact from Maven Central, which might
+# not work right away if the sync is taking time, followed by bintray, which should always work
+# since it's where we publish to.
+
+RUN mvn org.apache.maven.plugins:maven-dependency-plugin:get \
```
codefromthecrypt, line 11: nifty!

codefromthecrypt (approved):
looks good (to the degree I understand :) )

PR #2847, opened 2019-10-16
Default to 9200 when host but no port specified for elasticsearch.

### LINKED ISSUE #2843: Possible bug in ES when connecting DNS
In zipkin 2.17.2 running in docker, looks like it is connecting to the wrong port. Maybe we are defaulting wrong? It should default to 9200, not 80. User shouldn't have to specify 9200

```java
 curl -s localhost:9411/health
{
  "status" : "DOWN",
  "zipkin" : {
    "status" : "DOWN",
    "details" : {
      "ElasticsearchStorage{initialEndpoints=elasticsearch, index=zipkin}" : {
        "status" : "DOWN",
        "details" : {
          "error" : "java.lang.IllegalStateException: couldn't connect any of [Endpoint{elasticsearch:80, ipAddr=172.19.0.2, weight=1000}]"
        }
      }
    }
  }
```

anuraaga:
Fixes #2843

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/InitialEndpointSupplier.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/InitialEndpointSupplier.java
@@ -71,7 +71,14 @@ final class InitialEndpointSupplier implements Supplier<EndpointGroup> {
 
   int getPort(URI url) {
     int port = url.getPort();
-    if (port == -1) port = 9200 /* default Elasticsearch port */;
+    if (port == -1) {
+      if (sessionProtocol.isTls()) {
+        port = 443;
+      } else {
+        // Elasticsearch default plain-text port
+        port = 9200;
+      }
+    }
     return port;
   }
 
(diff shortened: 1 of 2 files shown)


codefromthecrypt (approved):
thanks!

anuraaga:
Realized we had a test explicitly verifying it defaulted to 80 or 443 - assuming it's fine to delete the test

zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/InitialEndpointSupplier.java
```diff
@@ -71,7 +71,7 @@
 
   int getPort(URI url) {
     int port = url.getPort();
-    if (port == -1) port = sessionProtocol.defaultPort();
+    if (port == -1) port = 9200 /* default Elasticsearch port */;
```
codefromthecrypt: ohh.. actually we should default to 443 if https!

codefromthecrypt (approved):
thanks!

codefromthecrypt:
it occurred to me maybe this is not precise? the problem is when a blank host is present. if it is http://host it should still default to port 80. I was wondering if maybe this change could possibly coerce http://host into http://host:9200 when the intent was coercing a plain hostname into http://hostname:9200

codefromthecrypt:
follow-up https://github.com/openzipkin/zipkin/pull/2857

PR #2853, opened 2019-10-17
Adds elasticsearch6 and elasticsearch7 images

codefromthecrypt:
These are used in our integration tests for zipkin-storage/elasticsearch

See https://github.com/openzipkin/docker-zipkin/issues/86

Tested with
```bash
$ cd docker
$ DOCKERFILE_PATH=docker/Dockerfile DOCKER_TAG=foo hooks/post_build
```


codefromthecrypt:
verified all the things

anuraaga:
Does this mean when we attic `docker-zipkin` elasticsearch2 and 5 will stop being updated forever? Makes sense to me just want to confirm.

codefromthecrypt:
yep let them die

codefromthecrypt:
I'm only adding in images we use or suggest for testing. we no longer reference 2 and 5 in our integration tests, which is the longer rationale.

we should bring the instructive docker-compose files over, including the prometheus demo, but yeah the old ES images can die

PR #2856, opened 2019-10-18
WIP: document new docker release process

codefromthecrypt:
cc @openzipkin/devops-tooling @saturnism

See https://github.com/openzipkin/docker-zipkin/issues/86

### DIFF
--- a/docker/RELEASE.md
+++ b/docker/RELEASE.md
@@ -1,0 +1,13 @@
+The files in the [hooks/] directory control the sequence of image build and push.
+Invocation of these files is based on webhooks defined in [dockerhub](https://cloud.docker.com/u/openzipkin/repository/docker/openzipkin/zipkin/builds).
+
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
cc @making as I know you have magical build things also

codefromthecrypt:
this will need to be integrated in https://github.com/openzipkin/zipkin/pull/2868

codefromthecrypt:
TODO: add tip that those wanting to be notified on docker build failures, you'll have to enable the notification in Docker Hub settings (notify on failures).

codefromthecrypt:
@anuraaga so to clarify as I think I know the answer. When webhook failed or the build failed for reasons like maven sync timeout.. we push docker-N.N.N from anywhere on master, correct? It doesn't have to be the same as the tag N.N.N

codefromthecrypt:
this is out-of-date now and covered in build-bin/README.md

PR #2860, opened 2019-10-18
Optimize image size

anuraaga:
`RUN chown` touches every file effectively doubling the size of the directory in the Docker image it seems. So we do need to make sure to use docker's native chowning.

Also moves classpath computation to build step since no need to do it every time the container starts

### DIFF
--- a/docker/Dockerfile.release
+++ b/docker/Dockerfile.release
@@ -35,6 +35,10 @@ RUN mvn org.apache.maven.plugins:maven-dependency-plugin:get \
 RUN mkdir -p /zipkin && cp ~/.m2/repository/io/zipkin/zipkin-server/${version}/zipkin-server-${version}-exec.jar /zipkin && cd /zipkin && jar xf *.jar && rm *.jar
 RUN mkdir -p /zipkin-slim && cp ~/.m2/repository/io/zipkin/zipkin-server/${version}/zipkin-server-${version}-slim.jar /zipkin-slim && cd /zipkin-slim && jar xf *.jar && rm *.jar
 
+# statically evaluate classpath to avoid https://github.com/docker/for-mac/issues/3643
+RUN cd /zipkin && echo .:$(ls ${PWD}/BOOT-INF/lib/*.jar|tr '\n' ':')${PWD}/BOOT-INF/classes > classpath
+RUN cd /zipkin-slim && echo .:$(ls ${PWD}/BOOT-INF/lib/*.jar|tr '\n' ':')${PWD}/BOOT-INF/classes > classpath
+
 # Extract zipkin-lens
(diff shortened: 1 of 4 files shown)


docker/Dockerfile.release
```diff
@@ -35,6 +35,10 @@ RUN mvn org.apache.maven.plugins:maven-dependency-plugin:get \
 RUN mkdir -p /zipkin && cp ~/.m2/repository/io/zipkin/zipkin-server/${version}/zipkin-server-${version}-exec.jar /zipkin && cd /zipkin && jar xf *.jar && rm *.jar
 RUN mkdir -p /zipkin-slim && cp ~/.m2/repository/io/zipkin/zipkin-server/${version}/zipkin-server-${version}-slim.jar /zipkin-slim && cd /zipkin-slim && jar xf *.jar && rm *.jar
 
+# statically evaluate classpath to avoid https://github.com/docker/for-mac/issues/3643
+RUN cd /zipkin && echo .:$(ls ${PWD}/BOOT-INF/lib/*.jar|tr '\n' ':')${PWD}/BOOT-INF/classes > classpath
+RUN cd /zipkin-slim && echo .:$(ls ${PWD}/BOOT-INF/lib/*.jar|tr '\n' ':')${PWD}/BOOT-INF/classes > classpath
```
codefromthecrypt: hmm I think this won't be correct, because PWD will be /zipkin-slim where we want it to end up /zipkin later..

anuraaga: Oops - thanks for noticing I only tested fat version ;) Fixed to prepend `/zipkin/` to relative paths

anuraaga:
Actually I found out how to use wildcard - perhaps when you tried before you added `.jar` which doesn't seem to be allowed. How about leaving it to Java?

docker/Dockerfile.release
codefromthecrypt: hey hey hey... not slim != fat.. more like slim++

docker/zipkin/run.sh
```diff
@@ -20,7 +20,7 @@ fi
 # Use main class directly if there are no modules, as it measured 14% faster from JVM running to available
 # verses PropertiesLauncher when using Zipkin was based on Spring Boot 2.1
 if [[ -z "$MODULE_OPTS" ]]; then
-  exec java ${JAVA_OPTS} -cp $(cat /zipkin/classpath) zipkin.server.ZipkinServer
+  exec java ${JAVA_OPTS} -cp '.:BOOT-INF/lib/*:BOOT-INF/classes' zipkin.server.ZipkinServer
```
codefromthecrypt, line 23: this actually works? wow.. must have fat-fingered something (pun intended)

codefromthecrypt (approved):

PR #2861, opened 2019-10-18
Squashes mysql things together to simplify apk

codefromthecrypt:
somewhat meaningless but why not


docker/storage/mysql/Dockerfile
```diff
@@ -16,13 +16,11 @@ FROM alpine
 LABEL MAINTAINER Zipkin "https://zipkin.io/"
 
 WORKDIR /mysql
-ADD docker/storage/mysql/install.sh /mysql/install
-ADD docker/storage/mysql/configure.sh /mysql/configure
+ADD docker/storage/mysql/install.sh /mysql/install.sh
```
anuraaga: Bonus points might be `ADD docker/storage/mysql/*.sh /mysql/`

anuraaga (approved):

docker/storage/mysql/Dockerfile
codefromthecrypt: oh .. didn't know that worked

PR #2866, opened 2019-10-19
Fix UI release build not copying from correct directory and add docke…

anuraaga:
…r-compose.*.test.yml files to check all our images


docker/Dockerfile.release
```diff
@@ -47,7 +47,7 @@ LABEL MAINTAINER Zipkin "https://zipkin.io/"
 
 ENV ZIPKIN_BASE_URL=http://zipkin:9411
 
-COPY --from=built /zipkin-lens /var/www/html/zipkin
+COPY --from=built /zipkin-lens/zipkin-lens /var/www/html/zipkin
```
codefromthecrypt, line 50: heh subtle.. nice find! thanks!

PR #2864, opened 2019-10-18
Updates to Spring Boot 2.2 and stops using proxyBeanMethods

codefromthecrypt:
`proxyBeanMethods` is disablable in Spring 5.2, shipped with Spring Boot 2.2

Fixes https://github.com/openzipkin/zipkin/issues/2863

PR #2867, opened 2019-10-19
Remove invocation of Bean methods.

anuraaga:
Not allowed to invoke Bean methods when turning off proxying (being allowed to invoke them in the first place has always been a terrible feature of Spring anyways...).

I didn't take a detailed look at whether this fixes the failure in #2864 but not surprised if it does. Instead of examining and debugging failures, I just went through every reference in `@InternalZipkinConfiguration` and ctrl-clicked on `@Bean` methods to see if they have references.

Also overrides the pretty printer for Jackson so the output always uses `\n` so tests pass on Windows.


codefromthecrypt:
good. thanks!

PR #2868, opened 2019-10-19
Allow the built tag to be prefixed with docker- which can be used to …

anuraaga:
…rebuild images without Zipkin server. Docker Hub will autobuild the `docker-*` tags. When rebuilding a Docker image for a given Zipkin version, can just retag with the same `docker-`

Also combine `post_build` into `build` - `post_build` is only really helpful when using Docker Hub's default `build` script but since we override, may as well have one file.


.travis.yml
```diff
@@ -41,11 +41,14 @@ before_install:
...
-# Don't build release tags. This avoids publish conflicts because the version commit exists both on master and the release tag.
-# See https://github.com/travis-ci/travis-ci/issues/1532
 branches:
   except:
+    # Don't build release tags. This avoids publish conflicts because the version commit exists both on master and the release tag.
+    # See https://github.com/travis-ci/travis-ci/issues/1532
     - /^[0-9]/
+    # Don't build docker- tags with Travis. Docker Hub will pick up and run the build. docker- tags are used to recreate
+    # docker images without rebuilding Zipkin so we don't need to run here.
+    - /^docker-/
```
codefromthecrypt, line 51: cool

PR #2871, opened 2019-10-20
Don't use Stream.parallel which moves tasks to another threadpool whi…

anuraaga:
…ch can break tracing.


codefromthecrypt:
thanks again!

codefromthecrypt:
confirmed!

PR #2873, opened 2019-10-21
Move source_branch parsing into Dockerfile to allow it to work with b…

anuraaga:
…oth docker build and docker-compose

There isn't really a way to specify the variables available to the docker-compose tests, short of overriding the `test` hook completely. I think it's fine to move the logic for determining the version instead. This is so the `docker-*` tags can pass tests.

### DIFF
--- a/docker/Dockerfile.release
+++ b/docker/Dockerfile.release
@@ -12,31 +12,35 @@
 # the License.
 #
 
-
 # Download artifact from bintray, where the artifact should have been published by now.
 
 FROM openzipkin/zipkin-builder as built
 
-ARG version
+ARG source_branch
 
+# source_branch is something like
+#
+#   master       - Building the master branch. This cut command will return 'master', and Dockerfiles will ignore it
+#   1.0.1        - Building a release image along with a new Zipkin server version. This cut command will return
+#                  '1.0.1' and Dockerfiles will use it to fetch the correct version of Zipkin.
+#   docker-1.0.1 - Building a release image, but not a new Zipkin server. This cut command will return
+#                  '1.0.1' and Dockerfiles will use it to fetch the correct version of Zipkin.
+
+RUN export version=$(echo "${source_branch}" | cut -d '-' -f 2) && \
 # Download jars using Maven. It will try to resolve the artifact from Maven Central, which might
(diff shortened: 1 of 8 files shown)


anuraaga:
By the way, this is only if we need to push the docker images (but not zipkin artifact) again, it doesn't affect the current release of `2.18.3`

codefromthecrypt:
ps so I know urgency, is anything broke at the moment until this? :D

docker/Dockerfile.release
```diff
@@ -12,31 +12,35 @@
...
 
-
 # Download artifact from bintray, where the artifact should have been published by now.
 
 FROM openzipkin/zipkin-builder as built
 
-ARG version
+ARG source_branch
+
+# source_branch is something like
```
codefromthecrypt, line 21: good docs

anuraaga:
The `docker-` tag, which would be used for repushing the docker image without a new java version is broken. But we don't currently need that (we'll know after checking the released docker images ;) )

docker/Dockerfile.release
```diff
@@ -12,31 +12,35 @@
...
 # since it's where we publish to.
-
-RUN mvn org.apache.maven.plugins:maven-dependency-plugin:get \
-  -DremoteRepositories=bintray::::https://dl.bintray.com/openzipkin/maven -Dtransitive=false \
-  -Dartifact=io.zipkin:zipkin-server:${version}:jar:exec
-RUN mvn org.apache.maven.plugins:maven-dependency-plugin:get \
-  -DremoteRepositories=bintray::::https://dl.bintray.com/openzipkin/maven -Dtransitive=false \
-  -Dartifact=io.zipkin:zipkin-server:${version}:jar:slim
-
+  mvn org.apache.maven.plugins:maven-dependency-plugin:get \
```
codefromthecrypt, line 33: worthwhile to use the wrapper?

codefromthecrypt (approved):
I think I understand this. thanks!

docker/Dockerfile.release
anuraaga: Probably is for consistency - will do it in separate PR since we should do it in the other files too.

PR #2880, opened 2019-10-24
Bump span name and autocomplete limit to 10k

drolando:
The main monolith at Yelp has more than 1k span names. This is due to
several reasons, main one being that all mysql queries also have
"yelp-main" as localEndpoint.serviceName. Since we use the query
fingerprint as name, this causes us to have around 6k span names.

### DIFF
--- a/zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/SelectRemoteServiceNames.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/SelectRemoteServiceNames.java
@@ -61,7 +61,7 @@ final class SelectRemoteServiceNames extends ResultSetFutureCall<ResultSet> {
     return factory.session.executeAsync(factory.preparedStatement
       .bind()
       .setString("service", service)
-      .setInt("limit_", 1000)); // no one is ever going to browse so many span names
+      .setInt("limit_", 1000)); // no one is ever going to browse so many service names
   }
 
   @Override public ResultSet map(ResultSet input) {
--- a/zipkin-storage/cassandra/src/test/resources/autocomplete_tags-stress.yaml
+++ b/zipkin-storage/cassandra/src/test/resources/autocomplete_tags-stress.yaml
@@ -45,5 +45,5 @@ queries:
     cql: SELECT DISTINCT key FROM autocomplete_tags
     fields: samerow
    select_values:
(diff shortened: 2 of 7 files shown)


zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/SelectRemoteServiceNames.java
```diff
@@ -61,7 +61,7 @@ protected ResultSetFuture newFuture() {
     return factory.session.executeAsync(factory.preparedStatement
       .bind()
       .setString("service", service)
-      .setInt("limit_", 1000)); // no one is ever going to browse so many span names
+      .setInt("limit_", 10000));
```
codefromthecrypt: this one is not about span names

```suggestion
      .setInt("limit_", 1000)); // no one is ever going to browse so many service names
```

zipkin-storage/cassandra/src/test/resources/autocomplete_tags-stress.yaml
```diff
@@ -45,5 +45,5 @@ queries:
     cql: SELECT DISTINCT key FROM autocomplete_tags
```
codefromthecrypt, line 45: I *think* in these files we also want to update `select:  fixed(1)/1000`

drolando: Yeah I saw those but I've no idea of what they mean. We can wait for the tests and see if they break

drolando: Travis seems happy without that change

codefromthecrypt: travis doesn't execute these files.

codefromthecrypt: the stress tests are invoked manually I mean.

drolando: How do I run them to test this?

codefromthecrypt: it says how at the top of the file.

codefromthecrypt:
thanks @drolando!

PR #2882, opened 2019-10-24
Redesign dependencies page

tacigar:
## Summary

Mainly I do the following two things in this PR.

* Detach Dependencies from DiscoverPage.
* Refactor Dependencies code.

### Detach Dependencies from DiscoverPage

Dependencies are currently embedded in the Discover Page as follows:

スクリーンショット 2019-10-25 8 14 58

At first glance, this design seems to share GlobalSearch between traces searching and dependencies searching.
Actually, however, Dependencies search uses only Lookback search criteria.

スクリーンショット 2019-10-25 8 19 49

I thought this would be very confusing to users.
And, I think that the topology view considered here(https://github.com/openzipkin/zipkin/issues/2701) should be placed where Dependencies is now.

Therefore I want to detach dependencies from DiscoverPage.

### Refactor Dependencies code

Currently, the code of Dependencies is the oldest and worst code in Lens.
I want to refactor it.

## Screenshots

スクリーンショット 2019-10-25 8 35 19

スクリーンショット 2019-10-24 18 19 59

スクリーンショット 2019-10-24 18 20 09

スクリーンショット 2019-10-24 18 20 41


tacigar:
Current State:

スクリーンショット 2019-10-24 18 19 59
スクリーンショット 2019-10-24 18 20 09
スクリーンショット 2019-10-24 18 20 41

codefromthecrypt:
@tacigar so let's add a description. say what you intend to do, why you are doing this, and why you love it!

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
```diff
@@ -0,0 +1,147 @@
...
+import _ from 'lodash';
+import Box from '@material-ui/core/Box';
+import { makeStyles } from '@material-ui/styles';
+
+import VizceralExt from './VizceralExt';
+import { theme } from '../../colors';
+
+const useStyles = makeStyles({
+  root: {
+    // If height is 100%, scroll bar appears.
```
codefromthecrypt, line 50: almost haiku

codefromthecrypt:
@openzipkin/ui another epic awesome on the way

tacigar:
@adriancole Sorry, I added description!

codefromthecrypt:
Thanks for the summary. The only thing I might suggest is that #2537 needs to fit somewhere.. would the filtering by service be a part of a separate screen? If not, wouldn't we want to share the context here?

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
```diff
@@ -0,0 +1,147 @@
...
+  nodes: PropTypes.arrayOf(PropTypes.shape({})).isRequired,
+};
+
+const DependenciesGraph = ({
+  selectedNodeName,
+  onClickNode,
+  edges,
+  nodes,
+}) => {
+  const classes = useStyles();
```
anuraaga, line 107: I remember previously you avoided `useStyles` to make it easier to write tests. Is that problem gone?

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
```diff
@@ -0,0 +1,147 @@
...
+  colorLabelBorder: theme.palette.primary.dark,
+  colorLabelText: theme.palette.primary.contrastText,
+  colorTrafficHighlighted: {
+    normal: theme.palette.primary.main,
+  },
+};
+
+const propTypes = {
+  selectedNodeName: PropTypes.string.isRequired,
+  onClickNode: PropTypes.func.isRequired,
```
anuraaga: Not sure if this pattern is prevalent in the code but I'd usually expect something like `onNodeClick`

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
```diff
@@ -0,0 +1,147 @@
...
+  selectedNodeName,
+  onClickNode,
+  edges,
+  nodes,
+}) => {
+  const classes = useStyles();
+  const [filter, setFilter] = useState('');
+
+  const handleObjectHighlight = useCallback((highlightedObject) => {
+    if (typeof highlightedObject === 'undefined') {
```
anuraaga: `!highlightedObject` is probably fine since there doesn't seem to be a need to treat `undefined` and `null` differently

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
```diff
@@ -0,0 +1,147 @@
...
+      return;
+    }
+    if (highlightedObject.type === 'node' && highlightedObject.getName() !== selectedNodeName) {
+      onClickNode(highlightedObject.getName());
+    }
+  }, [onClickNode, selectedNodeName]);
+
+  const maxVolume = useMemo(() => {
+    if (edges.length > 0) {
+      const maxVolumeEdge = _.maxBy(
```
anuraaga: Don't think you need lodash (if you did, better to import the method itself).
```ts
return edges.map(edge => edge.metrics.normal + edge.metrics.danger).reduce((a, b) => Math.max(a, b));
```

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
```diff
@@ -0,0 +1,147 @@
...
+    }
+    return 0;
+  }, [edges]);
+
+  const filterOptions = useMemo(() => nodes.map(node => ({
+    value: node.name,
+    label: node.name,
+  })), [nodes]);
+
+  const handleFilterChange = useCallback((selected) => {
```
anuraaga: Think you can combine `useCallback` and `useState` using `useReducer`

`const [filter, selectFilter] = useReducer((_, selected) => selected ? selected.value : ''), '')`

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
```diff
@@ -0,0 +1,147 @@
...
+          styles={reactSelectStyles}
+        />
+      </Box>
+    </Box>
+  );
+};
+
+DependenciesGraph.propTypes = propTypes;
+
+export default DependenciesGraph;
```
anuraaga, line 189: `React.memo`?

zipkin-lens/src/components/DependenciesPage/DependenciesPage.jsx
```diff
@@ -0,0 +1,235 @@
...
+  isLoading,
+  dependencies,
+  fetchDependencies,
+  clearDependencies,
+  location,
+  history,
+}) => {
+  const classes = useStyles();
+  const graph = useMemo(() => new Graph(dependencies), [dependencies]);
+  const isGraphExists = useMemo(() => graph.allNodes().length !== 0, [graph]);
```
anuraaga: `allNodes().length !== 0` is very cheap, this doesn't need to be memoized.

zipkin-lens/src/components/DependenciesPage/DependenciesPage.jsx
```diff
@@ -0,0 +1,235 @@
...
+  const [nodeName, setNodeName] = useState(null);
+  const [timeRange, setTimeRange] = useState({
+    startTime: moment().subtract(1, 'days'),
```
anuraaga, line 106: Probably not for this change but recommend using `dayjs` instead of `moment` - mutable date values in state always cause a bug at some point.

zipkin-lens/src/components/DependenciesPage/DependenciesPage.jsx
```diff
@@ -0,0 +1,235 @@
...
+      pathname: '/zipkin/dependency',
+      search: buildQueryParameters({ startTs, endTs }),
+    });
+  }, [fetchDependencies, timeRange, history]);
+
+  const handleClickNode = useCallback((newNodeName) => {
+    setNodeName(newNodeName);
+  }, []);
+
+  useMount(() => {
```
anuraaga: Personally I don't find the `useMount` / `useUnmount` helpers as readable as simple `useEffect`

- The code savings is four characters `, []`
- Anyone familiar with react standard library will know immediately what `useEffect` does
- Having `fetchDependencies` and `return clearDependencies` in the same `useEffect` makes very clear the lifecycle relationship between the two.

zipkin-lens/src/components/DependenciesPage/DependenciesPage.jsx
```diff
@@ -0,0 +1,235 @@
...
+    const queryParams = queryString.parse(location.search);
```
anuraaga: Consider just using standard library

https://caniuse.com/#feat=urlsearchparams (seems to match the targets in our babel config)

zipkin-lens/src/components/DependenciesPage/DependenciesPage.jsx
```diff
@@ -0,0 +1,235 @@
...
+          nodeName ? (
+            <Box className={classes.nodeDetailWrapper}>
+              <AutoSizer>
+                {
+                  ({ width, height }) => (
+                    <Box width={width} height={height} overflow="auto">
+                      <NodeDetail
+                        serviceName={nodeName}
+                        minHeight={height}
+                        targetEdges={graph.getTargetEdges(nodeName)}
```
anuraaga: Think these edges should be memoized

zipkin-lens/src/components/DependenciesPage/NodeDetail.jsx
```diff
@@ -0,0 +1,167 @@
...
+
+const NodeDetail = ({
+  serviceName,
+  targetEdges,
+  sourceEdges,
+  minHeight,
+}) => {
+  const classes = useStyles();
+
+  const usesEdgeNames = useMemo(
```
anuraaga: Unless uses / used are very well written in zipkin vocabulary, it might be worth considering renaming since they are very similar to react's `use*` hooks idiom. `input` / `output`?

bulicekj:
In a disconnected view like this, it would seem useful to me to have a way to link off to the traces view from a selected service, retaining the used from/to time parameters. Maybe inside the sidebar that opens when selecting a node?

bulicekj:
> Thanks for the summary. The only thing I might suggest is that #2537 needs to fit somewhere.. would the filtering by service be a part of a separate screen? If not, wouldn't we want to share the context here?

+1, these changes are great but doesn't solve the issues when there are a huge cluster of services in one graph

tacigar:
@anuraaga Thank you for your kind reviews!

tacigar:
> In a disconnected view like this, it would seem useful to me to have a way to link off to the traces view from a selected service, retaining the used from/to time parameters. Maybe inside the sidebar that opens when selecting a node?

@bulicekj Thank you for your feedback!
I tried to implement that as follows.

dependencies-page-1

tacigar:
And I tried to do the same hack as https://github.com/openzipkin/zipkin/pull/2537 in the new dependencies page.

dependencies-page-2

@bulicekj Thank you so much!

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
tacigar: Yes, If I don't use `shallow` rendering, `useStyles` is not a problem.
I felt that hooks and shallow rendering were incompatible, and in this PR I tried to write tests without shallow rendering.
So I could use useStyles.

tacigar:
PTAL 🙇

zipkin-lens/src/components/DependenciesPage/ExplainBox.jsx
```diff
@@ -0,0 +1,56 @@
...
+        <Box className="fas fa-search" />
+      </Box>
+      <Typography variant="h4">
+        Search Dependencies
+      </Typography>
+      <Typography variant="body1" className={classes.description}>
+        Please select the start time and the end time,
+        <br />
+        and click the search button.
+      </Typography>
```
tacigar, line 51: I want to be reviewed these messages.
Please tell me more natural English 🙇

zipkin-lens/src/components/App/Sidebar.test.js
```diff
@@ -39,8 +39,8 @@ describe('<Sidebar />', () => {
   it('should render internal links', () => {
     const wrapper = shallow(<Sidebar />);
     const list = wrapper.find('[data-testid="internal-links"]');
-    // Only discover page is the internal link (links routed in JavaScript).
-    expect(list.find(SidebarMenuItem).length).toBe(1);
+    // Only discover page and dependencies page are the internal link (links routed in JavaScript).
```
codefromthecrypt:
```suggestion
    // Only discover page and dependencies page are internal links (routed in JavaScript).
```

zipkin-lens/src/components/DependenciesPage/DependenciesPage.test.js
```diff
@@ -0,0 +1,146 @@
...
+    cleanup();
+  });
+
+  it('should render a loading indicator when isLoading is true', () => {
+    const { queryAllByTestId } = render(<DependenciesPageImpl {...commonProps} isLoading />);
+    const loadingIndicators = queryAllByTestId('loading-indicator');
+    expect(loadingIndicators.length).toBe(1);
+  });
+
+  it('should render a explain box if there is not graph yet', () => {
```
codefromthecrypt: an: https://www.grammar.com/a-vs-an-when-to-use/
no instead of not: because 'not' negates, and we want to say it is absent

```suggestion
  it('should render an explain box if there is no graph, yet', () => {
```

zipkin-lens/src/components/DependenciesPage/DependenciesPage.test.js
```diff
@@ -0,0 +1,146 @@
...
+    const { queryAllByTestId } = render(
+      <DependenciesPageImpl {...commonProps} dependencies={[]} />,
+    );
+    const explainBoxes = queryAllByTestId('explain-box');
+    expect(explainBoxes.length).toBe(1);
+  });
+
+  it('should render a dependencies graph when isLoading is false and there is a graph data', () => {
```
codefromthecrypt: dependency is a type of graph that shows dependencies.

the word data is plural (datum is singular, but we don't care how many data(): so 'a data' is incorrect. Easiest change is to remove the 'a'

```suggestion
  it('should render a dependency graph when isLoading is false and there is graph data', () => {
```

zipkin-lens/src/components/DependenciesPage/DependenciesPage.test.js
```diff
@@ -0,0 +1,146 @@
...
+      <DependenciesPageImpl
+        {...commonProps}
+        dependencies={exampleDependencies}
+      />,
+    );
+    const dependenciesGraphs = queryAllByTestId('dependencies-graph');
+    expect(dependenciesGraphs.length).toBe(1);
+  });
+
+  it('should render a node detail when nodes are clicked', () => {
```
codefromthecrypt: 'a' implies single, but single or many isn't important. We want to emphasize on the details.

```suggestion
  it('should render node details when nodes are clicked', () => {
```

zipkin-lens/src/components/DependenciesPage/DependenciesPage.test.js
```diff
@@ -0,0 +1,146 @@
...
+        {...commonProps}
+        dependencies={exampleDependencies}
+      />,
+    );
+    const serviceA = getByTestId('serviceA-button');
+    fireEvent.click(serviceA);
+    let nodeDetails = queryAllByTestId('node-detail');
+    expect(nodeDetails.length).toBe(1);
+
+    // And remove the node detail when graph's background is clicked.
```
codefromthecrypt: Changing the sentence order to match the code..

```suggestion
    // When the graph's background is clicked, the node detail should be removed.
```

zipkin-lens/src/components/DependenciesPage/DependenciesPage.test.js
```diff
@@ -0,0 +1,146 @@
...
+    const graphBackground = getByTestId('graph-background');
+    fireEvent.click(graphBackground);
+    nodeDetails = queryAllByTestId('node-detail');
+    expect(nodeDetails.length).toBe(0);
+  });
+
+  it('should fetch dependencies when mounted by using query parameters', () => {
```
codefromthecrypt: by is redundant with using in this case

```suggestion
  it('should fetch dependencies when mounted using query parameters', () => {
```

zipkin-lens/src/components/DependenciesPage/DependenciesPage.test.js
```diff
@@ -0,0 +1,146 @@
...
+    render(
+      <DependenciesPageImpl
+        {...commonProps}
+        fetchDependencies={fetchDependencies}
+      />,
+    );
+    expect(fetchDependencies.mock.calls.length).toBe(1);
+  });
+
+  it('should not fetch dependencies when query parameters is not set', () => {
```
codefromthecrypt: avoid double-negative (multiple nots). it is technically correct here, but can be distracting https://www.grammar-monster.com/glossary/double_negative.htm

```suggestion
  it('should not fetch dependencies when query parameters are missing', () => {
```

zipkin-lens/src/components/DependenciesPage/ExplainBox.jsx
```diff
@@ -0,0 +1,56 @@
...
+  return (
+    <Box className={classes.root}>
+      <Box className={classes.iconWrapper}>
+        <Box className="fas fa-search" />
+      </Box>
+      <Typography variant="h4">
+        Search Dependencies
+      </Typography>
+      <Typography variant="body1" className={classes.description}>
+        Please select the start time and the end time,
```
codefromthecrypt: less words same meaning. Notice I end the sentence because using 'and' in different ways can be tricky. I instead later use 'Then,'

https://guidetogrammar.org/grammar/conjunctions.htm

```suggestion
        Please select the start and end time.
```

zipkin-lens/src/components/DependenciesPage/ExplainBox.jsx
```diff
@@ -0,0 +1,56 @@
...
+        <br />
+        and click the search button.
```
codefromthecrypt:
```suggestion
        Then, click the search button.
```

zipkin-lens/src/components/DependenciesPage/NodeDetail.test.js
```diff
@@ -0,0 +1,66 @@
...
+
+  it('should render "Not found" message when sourceEdges is empty', () => {
+    const { queryAllByText } = render(
+      <NodeDetail {...commonProps} sourceEdges={[]} />,
+    );
+    const notFoundMessages = queryAllByText('Not found...');
+    expect(notFoundMessages.length).toBe(1);
+  });
+
+  it('should not render "Not found" message when targetEdges and sourceEdges are not empty', () => {
```
codefromthecrypt: reducing not count

```suggestion
  it('should render "Not found" message only when targetEdges or sourceEdges are present', () => {
```

codefromthecrypt (commented):
OK you asked for english clarifications :D

tacigar:
@adriancole Thank you so much!
You are my teacher! :)

zipkin-lens/src/components/DependenciesPage/DependenciesPage.jsx
```diff
@@ -0,0 +1,245 @@
...
+  const handleStartTimeChange = useCallback((startTime) => {
+    setTimeRange({ ...timeRange, startTime });
+  }, [timeRange]);
+
+  const handleEndTimeChange = useCallback((endTime) => {
+    setTimeRange({ ...timeRange, endTime });
+  }, [timeRange]);
+
+  const handleFindButtonClick = useCallback(() => {
+    const startTs = timeRange.startTime.valueOf();
```
codefromthecrypt, line 127: the parameters are endTs and lookback :) https://zipkin.io/zipkin-api/#/default/get_dependencies

tacigar: oops!!

tacigar:
PTAL 🙇

codefromthecrypt:
using yelp.json I cannot read the labels.. they look obfuscated (firefox)

Screenshot 2019-10-29 at 11 43 18 AM

codefromthecrypt:
same in safari. so it seems zooming in doesn't change the size of the labels it only makes the text blurry. I wonder if the font is not scaling for some reason. Possibly the font size for the dependency graph should be a function of how many nodes exist?

Oct-29-2019 11-52-24

tacigar:
Thank you for your report.
I don't know the reason, but I think the root cause is in third party libraries (Vizceral? Three.js?)
Anyway, I don't know the solution at all, so I want to solve it later.

If the browser's zoomup is between 80% and 120%, we can read the label correctly.
So I don't think this is basically a big problem...
WDYT?

codefromthecrypt:
if the label size was larger, I'd be ok. right now I can barely read them

codefromthecrypt:
I can also run the previous just I don't recall the label text being so small before.. maybe it was..

tacigar:
@adriancole Please tell me your browser's zoom percentage?

codefromthecrypt:
The animated gif started with no zoom level. I will try to post a higher resolution screenshot of zero zoom level.

codefromthecrypt:
reason I try zoom was the labels were small and I couldn't read :P

tacigar:
I see. thank you.
I investigated it, but I could not found way to adjust the font size.
My best guess is the font size is fixed....
https://github.com/Netflix/vizceral/blob/45c8c49c7210e883d357ffb28b03de46132da6f3/src/base/nodeNameView.js#L84

codefromthecrypt:
I see. anyway I can read as long as I don't attempt zoom as it makes them worse.

codefromthecrypt (approved):
works for me using the yelp data.

I think we should track separately the usability of zoom with labels as different than this change

codefromthecrypt:
I haven't tracked the comments from @anuraaga, but if they are considered, I'm fine to merge.

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
```diff
@@ -0,0 +1,172 @@
...
+const DependenciesGraph = React.memo(({
+  selectedNodeName,
+  onNodeClick,
+  edges,
+  nodes,
+}) => {
+  const classes = useStyles();
+  const [filter, selectFilter] = useReducer((_, selected) => (selected ? selected.value : ''), '');
+
+  // A hack that forces to reload vizceral graph...
```
anuraaga: Is there an issue we can link to here? This hack is very confusing so I think we need to make sure it has plenty of background.

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
```diff
@@ -0,0 +1,172 @@
...
+  return (
+    <Box className={classes.root}>
+      <VizceralExt
+        allowDraggingOfNodes
+        targetFramerate={30}
+        traffic={{
+          renderer: 'region',
+          layout: 'ltrTree',
+          name: 'dependencies-graph',
+          updated: new Date().getTime(),
```
anuraaga: Looks like vizceral-react uses current time if it's not passed so we shouldn't need to pass anything.

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
```diff
@@ -0,0 +1,172 @@
...
+    label: node.name,
+  })), [nodes]);
+
+  if (reloading) {
+    return null; // Now reloading Vizceral...
+  }
+
+  return (
+    <Box className={classes.root}>
+      <VizceralExt
```
anuraaga, line 135: This is a wild guess but if it works, I think it's still easier to understand than the current reload dance - can we try setting `key` to a value that changes when the filter changes? I think in react, components with different key are considered different, so it should result in a new component being constructed which is what we want IIUC.

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
```diff
@@ -0,0 +1,172 @@
...
+}) => {
+  const classes = useStyles();
+  const [filter, selectFilter] = useReducer((_, selected) => (selected ? selected.value : ''), '');
+
+  // A hack that forces to reload vizceral graph...
+  const [reloading, setReloading] = useState(false);
+  useEffect(() => { setReloading(true); }, [filter]);
+  useEffect(() => { setReloading(false); }, [reloading]);
+
+  const shownConnections = useMemo(() => {
```
anuraaga: Also I found vizceral has its own API for `filters` different from `match` - have we tried using it?

https://github.com/Netflix/Vizceral/wiki/Configuration#filters

I think this filter would work

```js
[
  { 
    name: 'shownConnections',
    type: 'connection',
    passes: (object, value) =>(object.source === value || object.target === value)
    value: filter
  },
  { 
    name: 'shownNodes',
    type: 'node',
    // I can't figure out a difference between this and the current logic of searching in shownConnections
    passes: (object, value) =>(object.source === value || object.target === value)
    value: filter
  },
]
```

Note, I couldn't find the reason that `shownNodes` currently has a dependency on `shownConnections`. If I'm missing something, it should also still be possible to compute the filters in our javascript like we do now, but continue to pass allNodes, allConnections to vizceral and filter like

```
   { 
    name: 'shownConnections',
    type: 'connection',
    passes: (object, value) => shownConnections.has(object)
    value: filter
  },
```

If doing this, we should also make sure to set `traffic.updated` to something that is only updated when `allNodes` and `allConnections` changes as there's no need to update the data anymore to update filters.

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
tacigar: Your guess is perfectly correct!
If I set key props, graph was reloaded correctly.

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
tacigar: I haven't noticed about `filters` until now...
thank you!

tacigar:
@anuraaga thank you for your review!
PTAL 🙇

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
```diff
@@ -0,0 +1,180 @@
...
+const DependenciesGraph = React.memo(({
+  selectedNodeName,
+  onNodeClick,
+  edges,
+  nodes,
+}) => {
+  const classes = useStyles();
+  const [filter, selectFilter] = useReducer((_, selected) => (selected ? selected.value : ''), '');
+
+  const filterConnections = useCallback((object, value) => {
```
anuraaga: These can be defined at the top level instead of using `useCallback`

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
```diff
@@ -0,0 +1,180 @@
...
+      return true;
+    }
+    return object.source.name === value || object.target.name === value;
+  }, []);
+
+  const filterNodes = useCallback((object, value) => {
+    if (!value) {
+      return true;
+    }
+    if (object.name === value) {
```
anuraaga: Ah cool, this filter code is easier to understand for me :)

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
```diff
@@ -0,0 +1,180 @@
...
+    value: node.name,
+    label: node.name,
+  })), [nodes]);
+
+  return (
+    <Box className={classes.root}>
+      <VizceralExt
+        allowDraggingOfNodes
+        targetFramerate={30}
+        traffic={{
```
anuraaga, line 138: If possible, we should set `updated` to a value that reflects when `nodes` and `edges` are actually updated to avoid `Vizceral.updateData` on every rerender. Do you think we can compute `currentTime` in the graph constructor and pass it through to here?

https://github.com/openzipkin/zipkin/blob/master/zipkin-lens/src/util/dependencies-graph.js#L15

For reference, the logic that refers to `updated` is this one

https://github.com/Netflix/vizceral-react/blob/master/src/vizceral.jsx#L130

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
```diff
@@ -0,0 +1,180 @@
...
+          renderer: 'region',
+          layout: 'ltrTree',
+          name: 'dependencies-graph',
+          maxVolume: maxVolume * 50,
+          nodes,
+          connections: edges,
+        }}
+        objectHighlighted={handleObjectHighlight}
+        styles={vizStyle}
+        key={filter}
```
anuraaga: Ah do we need this anymore? I'm hoping that using vizceral's `filters` feature fixes the issue we were seeing. If not, we can keep it.

anuraaga: If we do need to keep it, we should still add the comment right here
```
// For some reason, when updating `filters`, vizceral does not rerender the graph correctly in place. So instead of rerendering in place, we replace the `key`, which causes react to completely destroy and reconstruct the component, effectively rerendering the graph from scratch.
```

tacigar: Yes, we need this yet...
Without this, the layout will not be reset...

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
```diff
@@ -0,0 +1,187 @@
...
+          layout: 'ltrTree',
+          name: 'dependencies-graph',
+          maxVolume: maxVolume * 50,
+          nodes,
+          connections: edges,
+          updated,
+        }}
+        objectHighlighted={handleObjectHighlight}
+        styles={vizStyle}
+        key={
```
anuraaga, line 149: Thanks! I think we're ready to merge. If possible, can you try just one more time to see if it's still broken without this `key=`? I want to hope that using both `filters` and `updated` allows visczeral to manage its own state correctly and we might not have the problem anymore (filtering seems quite core to visceral so this breakage is surprising). And if we do need it, that's fine.

anuraaga (approved):

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
tacigar: The layout is not still reset.
I think this this is not a bug, this is a Vizceral specification...

If click serviceE in the dependency graph below...

スクリーンショット 2019-10-29 17 41 10

#### If `key` prop is not set...

スクリーンショット 2019-10-29 17 41 29

#### If `key` prop is set... (Layout is reset)

スクリーンショット 2019-10-29 17 41 57

anuraaga: Oh is that what we were working around? Personally not relayouting seems reasonable too for less brain context switch, but I can see us keeping our workaround then.

I thought in https://github.com/openzipkin/zipkin/pull/2537#issuecomment-487688553 there was actually a bug where two nodes got rendered on top of each other - if that bug is fixed, then whether we prefer relayout or not is a design call (and my comment doesn't make sense anymore...)

anuraaga: If we want to force relayout, then the comment is better like `Normally, when updating filters, Vizsceral will show and hide nodes without relaying them out. For a large dependency graph, this often won't let us see well the information about the filtered nodes, so we prefer to force a relayout any time we update the filter. Changing the key based on the filter like this causes react to destroy and reconstruct the component from scratch, which will have a layout zooming in on the filtered nodes.`

tacigar: This bug (https://github.com/openzipkin/zipkin/pull/2537#issuecomment-487688553) is fixed now.
Even if without `key` prop, the filter feature works well. (Maybe due to using `target.filters` prop of Vizceral).

As you say, relayouting forces brain context switch...
However, for complex dependency graphs with a large number of nodes, the graph relayouted is overwhelmingly easy to see.
And I think many users have such complex dependency graphs.
So personally I want to relayout when the filter is changed.

anuraaga: Yup that makes sense - can you go ahead and just update the comment with my second suggestion? And then we're good

codefromthecrypt:
great team work!

PR #2887, opened 2019-10-25
Tweak the build and reduce lodash import.

anuraaga:
Figured it'd be nice to add some features to minimize JS build size and keep it in check. In practice, the bundle is only affected ~10K since the main culprit of lens size is fontawesome, which we can improve by migrating to `react-fontawesome` or material-ui icons so we only bundle what we use - @tacigar do you happen to have any plan on migrating icons?

- Clean `target/classes` before compiling. Otherwise JS builds up in there which all get packed into the JAR - not an issue in our production builds since they start clean but makes debugging bundle size harder locally
- We're currently targeting a fixed version of modern browsers that gets older over time, but we may as well target the most recent ones automatically
- Enable `@babel/plugin-transform-runtime` to optimize code generation of polyfills. The impact isn't so large though since we only target modern browsers
- Have webpack dump an HTML with its stats to `target/stats.html` to be able to see how the ES6 module size is doing
- Replace lodash full import with method imports. Before, all of lodash was being loaded but now only the used methods are

Biggest dependency is visceral, followed by moment since we load but don't use all locales, and then chart.js

- while there is a moment webpack plugin to not load locales, it's probably better to just migrate to day.js
- I could only find one usage of chartjs, for a `Doughnot` on the dependencies page. Wonder if there's a more optimized way of rendering that

image

### DIFF
--- a/zipkin-lens/babel.config.js
+++ b/zipkin-lens/babel.config.js
@@ -11,21 +11,39 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  */
-const presets = [
-  [
-    '@babel/env',
-    {
-      targets: {
-        edge: '17',
-        firefox: '60',
-        chrome: '67',
-        safari: '11.1',
-      },
-      useBuiltIns: 'usage',
-      corejs: 3,
-    },
-  ],
-  '@babel/react',
-];
 
-module.exports = { presets };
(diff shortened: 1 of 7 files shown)


codefromthecrypt (approved):
approving on the idea :P

PR #2889, opened 2019-10-28
Update mockwebserver

adriangonz:
## Changelog
- Move `okhttp.version` to `zipkin-parent` so that all remaining usages of `okhttp3` (mostly on tests) follow the same version.
- Update `mockwebserver` on `zipkin-junit` to `v4.2.2` (same as `zipkin-server`).

## Context
`zipkin-junit` currently points to `com.squareup.okhttp3:mockwebserver` `v3.14.4`. This makes it tricky to update `okhttp3` on packages which use `zipkin-junit`. 

Since `zipkin` is moving away from `okhttp` to `armeria`, a cleaner solution would probably be to remove completely `okhttp` from `zipkin-junit` and `zipkin-server`, but that seems a far larger change.

While I was there, I also thought it would make sense to ensure that all the remaining usages of `okhttp3` follow the same versions. That's why I moved the dependencies to the `dependencyManagement` section of the `zipkin-parent`'s `pom.xml`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,7 +82,6 @@
     <awaitility.version>4.0.1</awaitility.version>
     <hamcrest.version>1.3</hamcrest.version>
     <testcontainers.version>1.12.2</testcontainers.version>
-    <okhttp.version>4.2.2</okhttp.version>
 
     <auto-value.version>1.7</auto-value.version>
     <animal-sniffer-maven-plugin.version>1.18</animal-sniffer-maven-plugin.version>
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
note: this will bump anyone actively not interested in moving to okhttp v4 with the reverse problem. I think recently @vietj had a concern of some kind about okhttp version, presumably due to zipkin-junit.

There's no plan to remove MWS from zipkin-junit as this hasn't caused any problems and it is unlikely armeria will cause less in a testing tool.

In other words even if we use armeria directly in zipkin-server, it doesn't affect the decision in zipkin-junit. For this reason, I don't think it should be joined in any way to the dependency tree in zipkin-server. please keep this a discussion of which version zipkin-junit should use.

adriangonz:
Got it. Thanks for the feedback @adriancole! 

Based on that, it makes sense to revert the change on `zipkin-parent` and `zipkin-server` and just change the version directly on `zipkin-junit`. I've just pushed a couple commits with this.

codefromthecrypt:
PS fwiw I'm cool with moving this up to v4

@vietj @liubao68 @objectiser @nicmunroe any of you not ok?

only issue you might care about is okhttp 4 has a dependency on kotlin

codefromthecrypt:
Thanks @adriangonz

adriangonz:
Thanks for merging @adriancole!

PR #2892, opened 2019-10-29
Update dependencies

jeqo:
Update deps, mainly Armeria, in order to fix [this](https://github.com/openzipkin-contrib/zipkin-storage-kafka/pull/58#issuecomment-545752973)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,9 +74,6 @@
     <junit.version>4.12</junit.version>
     <junit.jupiter.version>5.5.2</junit.jupiter.version>
     <powermock.version>2.0.4</powermock.version>
-    <!-- Up to v2.27.0 of mockito has a conflict https://github.com/mockito/mockito/issues/1606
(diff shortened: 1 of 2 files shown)


anuraaga:
There's a major breaking change in 0.95.0 for how annotated services are run https://github.com/line/armeria/releases/tag/armeria-0.95.0

We need to add `@Blocking` to all annotated services that don't return `HttpResponse` or `CompletionStage` - I think in this codebase, it's just the query API that returns `AggregatedHttpResponse`, also can't forget in the kafka code to add it to all the methods returning a JSON object.

pom.xml
```diff
@@ -73,15 +73,15 @@
 
     <junit.version>4.12</junit.version>
     <junit.jupiter.version>5.5.2</junit.jupiter.version>
-    <powermock.version>2.0.2</powermock.version>
+    <powermock.version>2.0.4</powermock.version>
     <!-- Up to v2.27.0 of mockito has a conflict https://github.com/mockito/mockito/issues/1606
```
codefromthecrypt: remove this comment as you just invalidated it!

codefromthecrypt (approved):

PR #2897, opened 2019-10-30
Loading indicators to discover page

tacigar:
## Summary

If users store a large amount of data, the data fetch time will be longer.
In this case, if the loading indicator is not displayed, the user may simply think that there is no data on zipkin-server.
This is a very big UX issue. we should fix it as soon as possible.

## Screenshots

loading-services

loading-traces


tacigar:
Because there is no uni tests for the DiscoverPage component now, I have to write them.
So WIP.

drolando:
+1
Whenever zipkin takes a while to return a result I'm always unsure whether it found nothing or it's simply still waiting for a response

drolando:
It'd be great to add the same to the trace view page, especially when querying by a specific traceId. But feel free to postpone that if it requires a lot of extra changes

codefromthecrypt:
Thank you!

PR #2899, opened 2019-10-30
Blot AUTHORIZATION header from HTTP logging.

anuraaga:
Don't think there's ever a good reason to log AUTHORIZATION header.

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/HttpClientFactory.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/HttpClientFactory.java
@@ -21,6 +21,7 @@ import com.linecorp.armeria.client.HttpClient;
 import com.linecorp.armeria.client.encoding.HttpDecodingClient;
 import com.linecorp.armeria.client.logging.LoggingClientBuilder;
 import com.linecorp.armeria.client.metric.MetricCollectingClient;
+import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.common.HttpHeaders;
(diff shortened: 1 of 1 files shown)


codefromthecrypt (commented):
looks like okhttp blots them out.. wdyt? https://github.com/square/okhttp/blob/1e3762ecb73ff896807253a10013fd0a9b4bd868/okhttp-logging-interceptor/src/test/java/okhttp3/logging/HttpLoggingInterceptorTest.java#L777

anuraaga:
Sorry by scrub I meant blot, which is what I'm doing. Difference is using asterisks instead of the weird character. Should we match the okhttp blotting?

codefromthecrypt:
I was just looking at the exact thing in okhttp, since multiple people on it.

but yeah it is more a style thing than anything else. The redact as in how humans redact with sharpee is more obviously redacted. otoh it is probably more cute than it is important. I'll leave it with you!

https://github.com/square/okhttp/pull/4287#discussion_r219717377

codefromthecrypt (approved):
probably deserves a test, if convenient

anuraaga:
Yeah unfortunately the logs go to a static logger so not really testable - will add later if possible.

codefromthecrypt:
thanks much!

anuraaga:
Forgot to mention, just in case it's an interesting cultural point - first thing I thought when opening that okhttp link was my fonts or GitHub's fonts are messed up. I think for many Asian users that block character would look a lot like [tofu](https://en.m.wikipedia.org/wiki/Noto_fonts?wprov=sfla1)

PR #2901, opened 2019-10-31
Updates all the versions

codefromthecrypt:


zipkin-collector/kafka/pom.xml
```diff
@@ -47,7 +47,7 @@
 
     <dependency>
       <groupId>com.github.charithe</groupId>
-      <artifactId>kafka-junit</artifactId>
+        <artifactId>kafka-junit</artifactId>
```
codefromthecrypt:
```suggestion
      <artifactId>kafka-junit</artifactId>
```

codefromthecrypt:
@jeqo FYI I tried to update to kafka 2.3.1 but the junit thing fails on method missing.. I guess signatures can change at any time huh :D

PR #2905, opened 2019-11-01
fix: add log4j2 metrics binder

jeqo:
To complete restoring metrics missing in `/prometheus` endpoint, this PR adds Log4j2 metrics.

### DIFF
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -515,7 +515,7 @@
               <classifier>slim</classifier>
               <!-- https://github.com/spring-projects/spring-boot/issues/3426 transitive exclude doesn't work -->
               <excludeGroupIds>
-                com.google.auto.value,com.google.guava,io.dropwizard.metrics,com.datastax.cassandra,com.github.jnr,org.ow2.asm,org.jooq,javax.xml.bind,org.mariadb.jdbc,com.zaxxer,org.apache.activemq,org.apache.geronimo.specs,org.fusesource.hawtbuf,org.apache.kafka,com.github.luben,org.lz4,org.xerial.snappy,com.rabbitmq,jakarta.annotation,org.apache.thrift,org.apache.logging.log4j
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
hmm we changed the server to use slf4j not log4j2. won't this bust slim which doesn't use log4j2? 

if you look at the pom, you'll notice we use log4j2 in the big dist and simple logging in slim. An annotation approach might not work properly for this reason.

cc @jorgheymans

zipkin-server/pom.xml
```diff
@@ -515,7 +515,7 @@
               <classifier>slim</classifier>
               <!-- https://github.com/spring-projects/spring-boot/issues/3426 transitive exclude doesn't work -->
               <excludeGroupIds>
-                com.google.auto.value,com.google.guava,io.dropwizard.metrics,com.datastax.cassandra,com.github.jnr,org.ow2.asm,org.jooq,javax.xml.bind,org.mariadb.jdbc,com.zaxxer,org.apache.activemq,org.apache.geronimo.specs,org.fusesource.hawtbuf,org.apache.kafka,com.github.luben,org.lz4,org.xerial.snappy,com.rabbitmq,jakarta.annotation,org.apache.thrift,org.apache.logging.log4j
```
codefromthecrypt, line 518: we can't do this. we spent a lot of effort trying to get rid of log4j2 in slim as it is huge.

jeqo: yeap, was about to comment that.

codefromthecrypt:
an idea from offline which is that the log counter is barely, if relevant in our dashboard as we always log to debug or trace. We swallow exceptions into logs yes, but at a low level. The indicator that there is a problem is drop metrics going up. In other words a counter on severe messages isn't likely to add signal and maybe we shouldn't have added that by default either https://github.com/openzipkin-attic/docker-zipkin/issues/213

So, another fix is to just delete that part of the dashboard and re-upload to grafana.

codefromthecrypt:
another solution is to make a custom condition for the log annotation so that it doesn't trip up the slim build

jeqo:
closing this in favor of updating grafana dashboard.

PR #2908, opened 2019-11-05
Add query timeout knob.

### LINKED ISSUE #2809: Align server timeout with ES_TIMEOUT
The server times out before the ES client times out, when the ES_TIMEOUT is set to a high value like 60s.

Align the server timeout to match with the ES_TIMEOUT value either automatically or add necessary documentation to avoid such conflicts.

anuraaga:
Currently, there is a hard server timeout of 11s. For queries, requests can be very long, especially when using Elasticsearch. While we can configure the Elasticsearch client timeout, we also need to make sure the query APIs themselves don't timeout.

Fixes #2809


zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpConfiguration.java
```diff
@@ -44,11 +48,17 @@
...
-    Optional<CollectorRegistry> collectorRegistry) {
+    Optional<CollectorRegistry> collectorRegistry,
+    @Value("${zipkin.query.timeout:11s}") Duration queryTimeout) {
     return sb -> {
       httpQuery.ifPresent(h -> {
-        sb.annotatedService(httpQuery.get());
-        sb.annotatedService("/zipkin", httpQuery.get()); // For UI.
+        Function<Service<HttpRequest, HttpResponse>, Service<HttpRequest, HttpResponse>>
+          timeoutDecorator = service -> (ctx, req) -> {
+          ctx.setRequestTimeout(queryTimeout);
```
jcchavezs, line 57: Is it possible one can pass timeout=(0s|0ms) for some reason? maybe ignore it and drop a log or is that case covered? Not sure about this edge case.

jcchavezs:
I guess the piece where the timeout is used is already in place and this is just making it an option a user can configure, isn't?

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpConfiguration.java
anuraaga: Thanks - `0` means disabling the timeout which is probably fine to leave exposed. I updated the `README` to explain that.

anuraaga:
Yeah there is a `11s` timeout currently in place, and this just adds an option so users can configure it.

jcchavezs (approved):

PR #2912, opened 2019-11-07
Armeria Server Configurator to Consumer Server Builder

jeqo:
Replacing Armeria Server configurator with ServerBuilder consumer to add services.

@anuraaga this PR is to showcase that replacing configurator with consumers is not working out-of-the-box. I did this to check that is not something related to [1], and might be related with how Zipkin server is configured. 

[1] https://github.com/openzipkin-contrib/zipkin-storage-kafka/issues/61

Anyhow, I'm fine keeping configurator in both Zipkin and Kafka Storage if it's required; and we can close this if you agree.


anuraaga:
Hmm that's weird - but yeah I'm fine with closing this and leaving as is

PR #2919, opened 2019-11-14
Use Jackson to write names and write to pooled buffer to reduce garbage.

### LINKED ISSUE #2915: Tabs in span name breaks span name dropdown
We have some span names that contain tabs. This is breaking zipkin ui (both classic and lens) in recent versions of zipkin because the browser cannot parse the JSON returned from the /api/v2/spans endpoint. I get an error like "SyntaxError: JSON.parse: bad control character in string literal at line 1 column 916 of the JSON data". Looking at the raw data returned from this endpoint the "bad character" is a tab. In older versions of zipkin (confirmed in 2.11.12) this character would have been encoded as "\t".

image

## Steps to Reproduce (using zipkin version 2.19.1)
1. Save a span with a tab in the span name
2. Load the investigation page and enter the service name for the above span
- See error produced by classsic ui
  - Note that the lens-ui doesn't report any error, even though span names failed to load
- See JSON.parse error in inspector

anuraaga:
We have an optimization to write out names as a quoted list for names, but this doesn't really work due to JSON's escaping rules. So instead we should stick to using Jackson which escapes correctly. We can get some performance back in exchange though by writing into a pooled buffer to reduce garbage generation.

Fixes #2915

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java
@@ -13,24 +13,25 @@
  */
 package zipkin2.server.internal;
 
+import com.fasterxml.jackson.core.JsonGenerator;
 import com.linecorp.armeria.common.AggregatedHttpResponse;
 import com.linecorp.armeria.common.HttpData;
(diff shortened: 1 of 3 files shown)


jeqo (approved):
LGMT :+1:

PR #2926, opened 2019-11-19
Fix a bug that graph is not rendered when timestamp is not set in span

### LINKED ISSUE #2906: NaNs in LensUI 2.19 instead of timings
## Describe the Bug
There is a bug with trace rendering in LensUI 2.19. I see NaNs instead of timings and there are no spans displayed.
 
image

## Steps to Reproduce
Upload trace https://gist.github.com/f0y/1fd8461bd52c522acd71061ebeff965e

tacigar:
resolves #2906 

### Before

スクリーンショット 2019-11-19 15 11 46

### After

スクリーンショット 2019-11-19 15 11 28

### DIFF
--- a/zipkin-lens/src/components/TracePage/TraceSummary.jsx
+++ b/zipkin-lens/src/components/TracePage/TraceSummary.jsx
@@ -35,6 +35,12 @@ const TraceSummary = React.memo(({ traceSummary }) => {
   const [childrenHiddenSpanIds, setChildrenHiddenSpanIds] = useState({});
   const [isSpanDetailOpened, setIsSpanDetailOpened] = useState(true);
   const traceTimelineWidthPercent = isSpanDetailOpened ? 60 : 100;
+  const traceTimestamp = useMemo(() => {
+    if (traceSummary.spans && traceSummary.spans.length > 0 && traceSummary.spans[0].timestamp) {
+      return traceSummary.spans[0].timestamp;
+    }
+    return 0; // Unfortunately, valid timestamps are not set.
+  }, [traceSummary.spans]);
 
   const handleChildrenToggle = useCallback((spanId) => {
     setChildrenHiddenSpanIds(prevChildrenHiddenSpanIds => ({
@@ -49,6 +55,7 @@ const TraceSummary = React.memo(({ traceSummary }) => {
 
   const handleTimelineRowClick = useCallback((spanId) => {
     const idx = traceSummary.spans.findIndex(span => span.spanId === spanId);
+    // When the currently selected span is clicked again, set RootSpanIndex back to 0.
     if (isRootedTrace && currentSpanIndex === idx) {
       if (rootSpanIndex === idx) {
         setRootSpanIndex(0);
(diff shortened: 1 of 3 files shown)


zipkin-lens/src/components/TracePage/TraceSummary.jsx
```diff
@@ -94,15 +102,28 @@ const TraceSummary = React.memo(({ traceSummary }) => {
     return rerootedTree.filter(span => !allHiddenSpanIds[span.spanId]);
   }, [rerootedTree, childrenHiddenSpanIds]);
 
-  // Find the minumum and maximum timestamps in the shown spans.
-  const startTs = useMemo(() => minBy(rerootedTree, 'timestamp').timestamp, [rerootedTree]);
-  const endTs = useMemo(() => rerootedTree.map((span) => {
```
tacigar, line 99: If there are some spans which does not have timestamps, `endTs` will be `NaN`.
This is the cause of #2926 error.

zipkin-lens/src/components/TracePage/TraceSummary.jsx
```diff
@@ -94,15 +102,28 @@ const TraceSummary = React.memo(({ traceSummary }) => {
...
-    let ts = span.timestamp;
-    if (span.duration) {
-      ts += span.duration;
-    }
-    return ts;
-  }).reduce((a, b) => Math.max(a, b)), [rerootedTree]);
+  // Find the minumum and maximum timestamps in shown spans.
+  const { startTs, endTs } = useMemo(() => {
```
anuraaga, line 106: It would be nice if we didn't have to special-case for 0 timestamp any time we use a span in the UI. Do you think it could make sense to add a `normalizedTimestamp` to span in trace-cleaner.js that is set to the trace timestamp when timestamp isn't present? We could use that everywhere except in the detailed view of the span where we should indicate the timestamp wasn't present.

tacigar: I agree with you.

codefromthecrypt:
I would recommend seeing if this can be merged because we should release zipkin soon due to code drift between armeria here and what's used in zipkin-gcp (possibly also zipkin-aws). Maybe update to latest armeria after deciding what to do here then cut a minor.

codefromthecrypt:
sorry I meant cut a patch

tacigar:
I agree with this Tommy's opinion.(https://github.com/openzipkin/zipkin/issues/2906#issuecomment-556940485)
What do you think @adriancole ?

codefromthecrypt:
sorry about the late comment. @shakuzen's note was asking about if data would be valid if there's no timestamp based on the notes in our docs about it. It is still invalid, but the trace should render.

we've had users think that because our data format permits certain fields like localEndpoint and timestamp to be absent, that means the data is valid. It isn't valid, but there's a different question about whether the UI should crash (shouldn't) or what the UI should do about it (skip or backfill based on some heuristic)

The reason it can be confusing is that spans can be reported twice. ex `span.flush()` followed by `span.finish()`. There are 2 json in this case and span-cleaner in lens will merge those as one. This means that if one or the other has a timestamp, we are fine.

Spans that never started are a bug, in other words. Spans reported twice are not necessarily a bug. In any case a missing timestamp shouldn't crash the UI for the trace.

Given this, should we merge the change?

cc @jorgheymans as being the issue-master this topic is an important and frequently misunderstood one. I'm even thinking about finding a place in our openapi docs to double or triple underscore that nullability of a field does not imply it is good data to leave it out. Somehow we need to clearly say that "patching" a span can happen and in that case, consider the merged data when thinking of how to interpret fields.

jorgheymans:
Also sorry about being late, i'ld say merge this in pending this comment https://github.com/openzipkin/zipkin/pull/2926#discussion_r348915149 . 

About making this clear in the openapi docs, I don't think this will buy us much. Having a trace diagnostic tool somewhere that is able to tell if a trace is more or less as we expect it or not, what is wrong, what we can compensate in the UI etc would be more helpful. It's along these lines https://github.com/openzipkin/zipkin/issues/1484 , but i'm more thinking a separate endpoint where you would upload a trace or give it the id of a trace that isn't showing and it spits out the diagnostics.

zipkin-lens/src/components/TracePage/TraceSummary.jsx
codefromthecrypt: this will assume trace has a timestamp :D

PR #2929, opened 2019-11-25
Only collect metrics for API endpoints.

### LINKED ISSUE #2928: Too many metrics/timeseries in zipkin
## Describe the Bug
As of right now, there are way to many metric/timeseries in zipkin that prometheus is scraping. The reason for this seems to be that among other things, the server_requests creates timeseries for ALL urls that is accessed on the server. Some examples:

http_server_requests_seconds_bucket{method="GET",status="200",uri="/zipkin/webfonts/fa-brands-400-cccc9d29470e879e40eb70249d9a2705.woff2",le="0.003844776",} 8.0
http_server_requests_seconds_bucket{method="GET",status="200",uri="/zipkin/traces.properties",le="0.003495251",} 7.0
http_server_requests_seconds_bucket{method="GET",status="200",uri="/zipkin/main.5676091b5509c77eec8a.css",le="1.789569706",} 7.0

## Steps to Reproduce
Look at the metrics under /actuator/prometheus

## Expected Behaviour
It is very nice that we can monitor how long requests the server processes takes, but it is not interesting to see how long it uses to load .css and .js files. 

On some parts of the metrics, it looks like this: 
http_server_requests_seconds_bucket{method="GET",status="200",uri="/api/v2/trace/{traceId}",le="0.008388607",} 16.0

This ({traceid}) is a very nice way of sorting the metrics, either use that way or perhaps look into only creating metrics for /api/-uris would be nice.

anuraaga:
Currently we attach metric collection to the entire server, with a currently insufficient blacklist that allows static resources to be recorded. Instead, we should just go ahead and whitelist to the API endpoints as that's what really matters for metrics and are in our grafana.

Fixes #2928

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpConfiguration.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpConfiguration.java
@@ -25,7 +25,6 @@ import com.linecorp.armeria.server.file.HttpFileBuilder;
 import com.linecorp.armeria.server.metric.PrometheusExpositionService;
 import com.linecorp.armeria.spring.ArmeriaServerConfigurator;
 import io.micrometer.core.instrument.MeterRegistry;
-import io.micrometer.core.instrument.config.MeterFilter;
 import io.prometheus.client.CollectorRegistry;
 import java.time.Duration;
 import java.util.Optional;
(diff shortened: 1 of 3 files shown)


anuraaga:
I have gone ahead and removed tests that checked for metrics for not-found URLs and redirects - they don't seem that useful.

jeqo (approved):
LGTM. Like much better the inverse filter, thanks!

PR #2934, opened 2019-11-29
Add download JSON button

### LINKED ISSUE #2885: Add download/view json button in lens ui
**Feature:**
The regular UI has a download/view JSON button which shows you the JSON representation of the current trace. Lens doesn't have such thing.

**Rational**
It's super useful when debugging UI issues and it's also one of the only ways to share traces in a ticket (unless you want to manually curl the /trace endpoint)

tacigar:
resolves #2885 

download-json

### DIFF
--- a/zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
+++ b/zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
@@ -32,84 +32,102 @@ const defaultProps = {
 };
 
 const useStyles = makeStyles(theme => ({
-  headerTop: {
-    borderColor: theme.palette.grey[300],
+  root: {
+    paddingLeft: theme.spacing(3),
+    paddingRight: theme.spacing(3),
+  },
+  upperBox: {
+    width: '100%',
+    display: 'flex',
+    justifyContent: 'space-between',
+    borderBottom: `1px solid ${theme.palette.grey[300]}`,
+  },
+  serviceNameAndSpanName: {
+    display: 'flex',
+    alignItems: 'center',
   },
   serviceName: {
     textTransform: 'uppercase',
   },
   spanName: {
-    color: theme.palette.text.hint,
+    color: theme.palette.text.secondary,
+  },
+  jsonUploaderAndSearchInput: {
+    display: 'flex',
+    alignItems: 'center',
+    paddingRight: theme.spacing(4),
+  },
+  lowerBox: {
+    display: 'flex',
+    marginTop: theme.spacing(0.5),
+    marginBottom: theme.spacing(0.5),
+    alignItems: 'center',
+  },
+  traceInfoEntry: {
+    marginRight: theme.spacing(1),
+    display: 'flex',
+  },
+  traceInfoLabel: {
+    fontWeight: 'bold',
+    color: theme.palette.grey[600],
+  },
+  traceInfoValue: {
+    fontWeight: 'bold',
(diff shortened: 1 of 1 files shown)


tacigar:
cc @drolando @lambcode

zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
```diff
@@ -32,84 +35,135 @@ const defaultProps = {
...
 }));
 
 const TraceSummaryHeader = React.memo(({ traceSummary, rootSpanIndex }) => {
   const classes = useStyles();
 
+  const handleSaveButtonClick = useCallback(() => {
+    if (!traceSummary || !traceSummary.traceId) {
+      return;
+    }
+    fetch(`${api.TRACE}/${traceSummary.traceId}`)
```
anuraaga, line 101: I think it's fine to `fetch`, but don't we already have the trace's JSON in memory?

anuraaga (approved):

zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
tacigar: Strictly speaking, no.
Raw trace data will be converted in the action creator.

https://github.com/openzipkin/zipkin/blob/7e80e71bf682f74d900f866bc9ab1608ac01f6c8/zipkin-lens/src/actions/trace-action.js#L54-L55

And only the converted data are saved as the redux state.

tacigar: Conversion processes of the trace data often takes time, so I thought it should be done asynchronously within Action creator, and I did so.

tacigar: Of course, raw data can also be stored as a redux state.
However, in Trace page, lens currently re-uses data which is fetched in Search page.

https://github.com/openzipkin/zipkin/blob/7e80e71bf682f74d900f866bc9ab1608ac01f6c8/zipkin-lens/src/actions/trace-action.js#L43-L46

Because of this process, I thought that storing raw data in the redux state would make code more complex than necessary.
I thought the simplicity of the implementation is the most important, so I decided to fetch the JSON again when the button was pressed.

anuraaga: Sure it makes sense since exporting traces is uncommon anyways

zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
```diff
@@ -32,84 +35,135 @@ const defaultProps = {
...
-                </Box>
-              </Box>
-            ))
-          ) : null
-        }
+      <Box className={classes.lowerBox}>
+        {traceInfo}
+        <Button variant="outlined" className={classes.saveButton} onClick={handleSaveButtonClick}>
+          <Box component="span" className={`${classes.saveButtonIcon} fas fa-download`} />
+          Save JSON
```
shakuzen, line 165: Just a nit pick on the text of the button. I think "Download JSON" may be better because the trace is technically already saved in the backend. Maybe others won't have any confusion of these terms. Anyone else have an opinion?

jorgheymans:
Download seems better yes.

Op ma 2 dec. 2019 07:45 schreef Tommy Ludwig <notifications@github.com>:

PR #2936, opened 2019-11-29
Fix bug that webpack does not work on Windows

tacigar:
In Windows, path delimiters is `\`, not `/`... :sob:


tacigar:
https://github.com/openzipkin/zipkin/pull/2937

Instead of this PR, I created a new PR that solves more essential problems.
Therefore I close this PR.

PR #2937, opened 2019-11-30
Use react-fontawesome instead of using fontawesome directly

tacigar:
Before, Zipkin Lens used Font-Awesome directly, but I think this way has the following problems.

* Additional configs are required for Webpack (see: https://github.com/openzipkin/zipkin/blob/08cee5023bf6e24c277bce85839ad1c043230f63/zipkin-lens/webpack.dev.config.js#L51-L54)
* All icons need to be imported, so bundle size will increase (see: https://github.com/openzipkin/zipkin/blob/08cee5023bf6e24c277bce85839ad1c043230f63/zipkin-lens/scss/main.scss#L19-L22)

I think it is better to use Font-Awesome through React components provided by Fort-Awesome (https://github.com/FortAwesome/react-fontawesome).
This library seems to avoid such problems.


anuraaga (commented):
Yes!! Thanks a lot for this. I wasn't expecting big styling changes for this though. Is it possible to just migrate libraries with less code change? If not no worries will take a deeper look

tacigar:
Sorry for big changes.
In addition to library migration, I changed the Sidebar code a bit.
But all other changes are library migrations.

The SidebarMenuItem component received a FontAwesome icon string as props.

https://github.com/openzipkin/zipkin/blob/08cee5023bf6e24c277bce85839ad1c043230f63/zipkin-lens/src/components/App/Sidebar.jsx#L63-L66

So it was necessary to change SidebarMenuItem's props when migrating the library.
When making that change, I realized that SidebarMenuIcon has too verbose and meaningless logic.
So I refactored the Sidebar code a bit.

If possible, I would appreciate it if you could review this change within this PR... :sob: 
Sorry...

zipkin-lens/src/components/App/Layout.jsx
```diff
@@ -13,33 +13,90 @@
  */
 import PropTypes from 'prop-types';
 import React from 'react';
+import { faSearch, faProjectDiagram, faHome } from '@fortawesome/free-solid-svg-icons';
```
tacigar: I have integrated Sidebar component into Layout component, since they are very simple components,

zipkin-lens/src/components/App/Layout.jsx
```diff
@@ -13,33 +13,90 @@
...
-    <Box
-      component="main"
-      display="flex"
-      flexDirection="column"
-      height="100vh"
-      width="100%"
-      overflow="auto"
-    >
-      {children}
+const Layout = ({ children }) => {
```
tacigar, line 71: I have integrated Sidebar component into Layout component, since they are very simple components,

zipkin-lens/src/components/App/Layout.jsx
```diff
@@ -13,33 +13,90 @@
...
+import List from '@material-ui/core/List';
+import { makeStyles } from '@material-ui/styles';
 
-import Sidebar from './Sidebar';
+import SidebarMenu from './SidebarMenu';
+import Logo from '../../img/zipkin-logo.svg';
+
+const drawerWidth = '3.2rem';
+
+const useStyles = makeStyles(theme => ({
```
tacigar, line 29: This style is almost the same as the previous one.

tacigar: This combines the styles of the previous Sidebar component and the previous Layout component.

zipkin-lens/src/components/App/Layout.jsx
```diff
@@ -13,33 +13,90 @@
...
-      overflow="auto"
-    >
-      {children}
+const Layout = ({ children }) => {
+  const classes = useStyles();
+
+  return (
+    <Box className={classes.root}>
+      <CssBaseline />
+      <Drawer open variant="permanent" className={classes.drawer} classes={{ paper: classes.drawerPaper }}>
```
tacigar, line 77: This Drawer was previously in Sidebar component.

zipkin-lens/src/components/App/Layout.test.js
```diff
@@ -1,62 +0,0 @@
-/*
```
tacigar, line 1: I thought these unit tests were meaningless and would prevent us from changing structures of the DOM tree in the future.
So I deleted these unit tests.

zipkin-lens/src/components/App/SidebarMenu.jsx
```diff
@@ -0,0 +1,76 @@
...
+      backgroundColor: theme.palette.primary.dark,
+    },
+  },
+  'item--selected': {
+    color: theme.palette.common.white,
+    backgroundColor: theme.palette.primary.dark,
+  },
+}));
+
+export const SidebarMenuImpl = ({
```
tacigar, line 47: I renamed SidebarMenuItem to SidebarMenu.

anuraaga (approved):
Thanks!

zipkin-lens/src/components/App/Layout.test.js
abesto: Was gonna ask about this, thanks for explaining your reasoning.

abesto (approved):
+1 for more clearly separated changes, but the individual commits in here are actually doing an OK job of that I think. Methinks this is a good change 👍

PR #2938, opened 2019-12-01
Show client's remote service name in all cases. Fixes #2932

### LINKED ISSUE #2932: Lens UI: Not displaying inner CLIENT spans correctly
Version: 2.19.2

Attached trace is displayed as:

```
+ my-service (with dropdown button)
   + my-service (with dropdown button)
      - remote-service
```

The dropdown button always drops the main root.
In the discovery panel, it's displayed as 3x `my-service` and 2x `remote-service`

Expected:
- displayed:
```
+ my-service (with dropdown button)
   - remote-service
      - remote-service
```
- discovery panel count 1x `my-service` and 2x `remote-service`

```
[
   {
      "traceId":"73ace58cca1e3f85",
      "parentId":"73ace58cca1e3f85",
      "id":"036f42fcb6661c79",
      "kind":"CLIENT",
      "name":"get",
      "timestamp":1574947566084564,
      "duration":9126,
      "localEndpoint":{
         "serviceName":"my-service"
      },
      "remoteEndpoint":{
         "serviceName":"remote-service"
      }
   },
   {
      "traceId":"73ace58cca1e3f85",
      "parentId":"036f42fcb6661c79",
      "id":"2059bd44ae854a73",
      "kind":"CLIENT",
      "name":"get",
      "timestamp":1574947566095573,
      "duration":19880,
      "localEndpoint":{
         "serviceName":"my-service"
      },
      "remoteEndpoint":{
         "serviceName":"remote-service"
      }
   },
   {
      "traceId":"73ace58cca1e3f85",
      "id":"73ace58cca1e3f85",
      "kind":"SERVER",
      "name":"get",
      "timestamp":1574947566012997,
      "duration":103426,
      "localEndpoint":{
         "serviceName":"my-service"
      },
      "remoteEndpoint":{
         "ipv4":"127.0.0.1"
      }
   }
]
```

dstepanov:
The code had a check to only show the client's remote service name when the span doesn't have any children.

Problem is described in https://github.com/openzipkin/zipkin/issues/2932

Basically, when reactive HTTP client (Java in my case) is used, CLIENT spans are created in hierarchical order.

Use case:

```
Incoming request:
Two parallel operations:
1. Fetch user A, update user A
2. Fetch user B, update user B
Wait for completion...
```
Is producing CLIENT spans in hierarchical order:
```
-> Main span 
  -> "Fetch user A" CLIENT span -> `update user A` CLIENT span
  -> "Fetch user B" CLIENT span -> `update user B` CLIENT span
```
But UI doesn't support it.

### DIFF
--- a/zipkin-lens/src/zipkin/span-row.test.js
+++ b/zipkin-lens/src/zipkin/span-row.test.js
@@ -186,12 +186,12 @@ describe('SPAN v2 -> spanRow Conversion', () => {
           value: '192.168.99.101:9000 (backend)',
         },
       ],
-      serviceName: 'frontend', // prefer the local address vs remote
+      serviceName: 'backend', // prefer the local address vs remote
       serviceNames: ['backend', 'frontend'],
       errorType: 'none',
     };
 
-    expect(newSpanRow([v2], false)).toEqual(spanRow);
+    expect(newSpanRow([v2])).toEqual(spanRow);
   });
 
   it('should not duplicate service names', () => {
@@ -200,7 +200,7 @@ describe('SPAN v2 -> spanRow Conversion', () => {
       id: '3',
       localEndpoint: frontend,
       remoteEndpoint: frontend,
-    })], false);
+    })]);
 
     expect(converted.serviceNames).toEqual(['frontend']);
   });
@@ -244,7 +244,7 @@ describe('SPAN v2 -> spanRow Conversion', () => {
       errorType: 'none',
     };
 
-    expect(newSpanRow([v2], false)).toEqual(spanRow);
+    expect(newSpanRow([v2])).toEqual(spanRow);
   });
 
   // originally zipkin2.v1.SpanConverterTest.client_kindInferredFromAnnotation
@@ -286,7 +286,7 @@ describe('SPAN v2 -> spanRow Conversion', () => {
       errorType: 'none',
     };
 
-    expect(newSpanRow([v2], false)).toEqual(spanRow);
+    expect(newSpanRow([v2])).toEqual(spanRow);
   });
 
   // originally zipkin2.v1.SpanConverterTest.lateRemoteEndpoint_cr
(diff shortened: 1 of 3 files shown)


tacigar:
There seems to be a discussion here(https://github.com/openzipkin/zipkin/pull/2345#issuecomment-452942153 and https://github.com/openzipkin/zipkin/issues/2274) about whether to display remote services which are not leaf node...

What do you think? @openzipkin/ui @anuraaga @drolando

dstepanov:
Anyone else can comment on it? I'm using Brave as described here https://github.com/openzipkin/zipkin/issues/2932#issuecomment-561096546

zipkin-lens/src/zipkin/span-row.test.js
```diff
@@ -316,12 +316,12 @@ describe('SPAN v2 -> spanRow Conversion', () => {
         },
       ],
       tags: [{ key: 'Server Address', value: '192.168.99.101:9000 (backend)' }],
-      serviceName: 'frontend',
+      serviceName: 'backend',
```
codefromthecrypt, line 319: this is a good reason why we shouldn't merge this as is.

codefromthecrypt (commented):
a client calling a client, if we were to change to believe that was intentional as opposed to layered instrumentation, will have other affects also. For example, the dependency graph both in java and javascript also has assumptions around this. Changing hastily may cause a lot of other types of bugs.

For now, it is a better idea to change/fix the instrumentation to not make child spans of client spans.

dstepanov:
@adriancole Does it mean `CurrentTraceContext#wrap` should be fixed to attach the span to the last SERVER span in a case when the parent is CLIENT?

codefromthecrypt:
it could be a server span, or an intermediate (local) span.

the "folding" of seemingly duplicate clients is something that prevents, say duplicate http client middleware from appearing as multiple calls.

in any case we want to make sure that duplicate calls which are intentionally dupes are hierarchically sensible. callbacks certainly muddy this water. I think if you are referring to brave, which I think you are, there are some recent code merging to ensure that response callbacks reset to whatever the invocation span was.. regardless of whether that was a server, message consumer, or some intermediate span. https://github.com/openzipkin/brave/pull/1055

codefromthecrypt:
I appreciate you raising this, and one outcome was that we have better docs. That said, the change as proposed won't be merged, so let's close it and move on. Cheers

https://github.com/openzipkin/brave/blob/master/instrumentation/RATIONALE.md#why-does-the-client-response-callback-run-in-the-invocation-context

PR #2939, opened 2019-12-03
Release URL allocated with createObjectURL using revokeObjectURL

tacigar:
Probably, blob URI which is allocated with createObjectURL will not be collected by GC...
We should revoke this explicitly.


anuraaga (commented):
Found this - should we add to the DOM and use a small timeout for better compatibility?

https://stackoverflow.com/questions/30694453/blob-createobjecturl-download-not-working-in-firefox-but-works-when-debugging

tacigar:
@anuraaga Thank you for your investigation!
PTAL 🙇

anuraaga (approved):

PR #2949, opened 2019-12-28
Show ellipsis when the service name is too long in trace tree

### LINKED ISSUE #2948: Show ellipsis when the service name is too long in trace tree
**Feature:**

Currently, when service names are too long, service name components of the trace tree is displayed as follows:

current-long-service-name

This is very confusing, so we should show ellipsis as below:

long-service-name

And we should show tooltips that tell users the complete service name when hovered.

tacigar:
resolves #2948

### DIFF
--- a/zipkin-lens/src/components/TracePage/TraceTimeline/ServiceBadge.hook.js
+++ b/zipkin-lens/src/components/TracePage/TraceTimeline/ServiceBadge.hook.js
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2015-2019 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+import React from 'react';
+
+const useServiceBadge = (serviceName) => {
+  const rectEl = React.useRef();
+  const textEl = React.useRef();
+  const [maxWidth, setMaxWidth] = React.useState();
+
+  React.useEffect(() => {
+    // getBBox().width will be 0 unless the first rendering is completed.
+    // Therefore, use setTimeout.
+    setTimeout(() => setMaxWidth(rectEl.current.getBBox().width));
+
+    const observer = new ResizeObserver(() => {
+      setMaxWidth(rectEl.current.getBBox().width);
+    });
+
+    // If rectEl.current is directly used in the cleanup function, eslint issues a warning.
+    // To avoid this warning, assign rectEl.current into a local variable.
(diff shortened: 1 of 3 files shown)


tacigar:
show-ellipsis

I was able to display ellipsis in SVG, but for traces with about 1000 spans, the rendering speed is too slow. Probably this is caused by `ResizeObserver`.

I think virtualizing the trace graph may help avoid this. I will investigate it.

anuraaga (commented):
Does CSS not work?

https://www.w3schools.com/cssref/css3_pr_text-overflow.asp

tacigar:
Probably, not work.
In svg 1.1, `text` element does not seem to support `text-overflow` attribute.

This feature may be introduced in svg 2.
https://www.w3.org/TR/2018/CR-SVG2-20181004/text.html#TextOverflowProperty

tacigar:
Virtualizing the trace tree is a very high cost for implementing.
So I tried optimization by using Intersection Observer API.

Large traces are now drawn relatively smoothly.

zipkin-lens/src/components/TracePage/TraceTimeline/ServiceBadge.hook.js
```diff
@@ -0,0 +1,65 @@
...
+      resizeObserver.current.unobserve(rectEl.current);
+    }
+  }, []);
+
+  React.useEffect(() => {
+    const intersectionObserver = new IntersectionObserver(handleIntersect, {
+      rootMargin: '200px',
+    });
+
+    // If rectEl.current is directly used in the cleanup function, eslint issues a warning.
```
anuraaga, line 41: Better to suppress the warning with `eslint-disable-next-line` which is self-documenting and avoids code smells.

zipkin-lens/src/components/TracePage/TraceTimeline/ServiceBadge.hook.js
```diff
@@ -0,0 +1,65 @@
...
+    });
+    return () => intersectionObserver.unobserve(el);
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+
+  React.useEffect(() => {
+    let textContent = serviceName;
+    let textLength = textContent.length;
+    textEl.current.textContent = textContent;
+
+    while (textEl.current.getBBox().width > maxWidth && textLength >= 1) {
```
anuraaga, line 55: This seems to be the main point of slowness since it works with the real, not virtual dom.

How about computing the diff vs max width and using a reasonable definition of min char width to conservatively resize only once? Even if it's off by a char or two it should be fine in practice.

anuraaga:
Apologies if this is a double comment I thought I sent it but can't find it.

By the way, for a future PR, what do you think of migrating from SVG to normal DOM + CSS? In the past SVG or Canvas were fast but I find that currently, it's hard to beat just normal HTML since it's what browsers optimize for. With flexbox, a lot of the reason to use non-dom went away, and we're paying the price of not using modern CSS.

tacigar:
I currently agree with using normal DOM + CSS instead of SVG.
Drawing trees with DOM + CSS is a bit difficult, but it is possible.
And, there is a possibility that the number of DOM elements will be huge due to the drawing of the tree, but virtualizing the graph using react-virtualized etc. will get rid of this problem.
So I'll close this PR, and will try to re-implement the trace graph component with normal DOM + CSS.

PR #2950, opened 2020-01-02
Revert "Show ellipsis when the service name is too long in trace tree"

tacigar:
Reverts openzipkin/zipkin#2949


tacigar:
I'm really sorry 😭 
I made a mistake.

Issue #2922, opened 2019-11-14
Orphaned spans causing UI issues


lambcode:
## Describe the Bug
I'm not sure that orphaned spans is the correct term here, but I'm going to use it to describe a span that does not have a parent span but is still part of a complete trace. These orphaned spans are very difficult to differentiate from other spans, and they seem to cause problems with the current lens UI.

## Steps to Reproduce (zipkin 2.19.1)
1. Open this saved trace in lens UI [exampleTrace.json.txt](https://github.com/openzipkin/zipkin/files/3848574/exampleTrace.json.txt) (saved as txt for github compatibility)
2. Collapse the root span
- Note that the orphaned span does not get collapsed
- Note that the orphaned span appears like a child span of the root span

## Expected Behaviour
This is up for discussion, but what I'd like to see is orphaned spans to show up as siblings to the root span in the tree. I would also appreciate it if these spans were somehow marked as orphaned. As for the collapsing behavior, I would expect that collapsing the root span does not collapse these orphaned spans, but if the 'collapse all' button is used then all spans would be collapsed leaving just the root span and the orphaned span roots visible.

jcchavezs:
@tacigar do you have some time to check it?

tacigar:
Thank you for your report.
I think showing up orphaned spans as siblings to the root span is most reasonable.

tacigar:
I investigated code which builds trace tree.

https://github.com/openzipkin/zipkin/blob/8a5609a44ea52fa85d5309d7e4e46bc70a4c3ca8/zipkin-lens/src/zipkin/span-node.js#L272-L276

In the current implementation, all orphaned spans will be children of the root span...

tacigar:
I still don't know the best way to draw graphs which contains orphaned spans.
However, according to the code comment above, I think that orphaned spans should be drawn as children of the root span anyway.
Of course, the current behavior when collapsing orphaned spans is a bug, so I will fix it first.

jorgheymans:
@tacigar can this be closed since https://github.com/openzipkin/zipkin/issues/2922 was merged ?

jorgheymans:
Comparing latest 2.19.1 (left) with 2.22 (right):

Not collapsed:

![image](https://user-images.githubusercontent.com/193792/97023744-3c6bca80-1556-11eb-9860-abc5a812aa25.png)

Collapsed:

[...]

PR #2956, opened 2020-01-08
Fix orphaned spans bug

tacigar:
Fix an obvious bug pointed out in #2922 .

Mainly change the calculation method of `shownTree`.
Previously, **span ID** was used to calculate the span's parent-child relationship and Expand/Collapse buttons were implemented using that parent-child relationship.
This created the orphaned span problem pointed out in #2922 because orphaned spans don't have parent-child relationship using **span IDs**.

Please see here. Orphan spans are simply appended to the root span:
https://github.com/openzipkin/zipkin/blob/8a5609a44ea52fa85d5309d7e4e46bc70a4c3ca8/zipkin-lens/src/zipkin/span-node.js#L272-L276

This PR uses span indices to calculate parent-child relationship instead of span IDs.

p.s. TraceSummary component is fairly complex, so I wanted to create custom hooks and improve readability, but I'm not doing it now to minimize PR.

### Before

before-orphaned

### After

after-orphaned

### DIFF
--- a/zipkin-lens/src/components/TracePage/TraceSummary.jsx
+++ b/zipkin-lens/src/components/TracePage/TraceSummary.jsx
@@ -23,6 +23,8 @@ import SpanDetail from './SpanDetail';
 import { detailedTraceSummaryPropTypes } from '../../prop-types';
 import { hasRootSpan } from '../../util/trace';
 
+const findSpanIndex = (spans, spanId) => spans.findIndex(span => span.spanId === spanId);
+
 const propTypes = {
   traceSummary: detailedTraceSummaryPropTypes.isRequired,
 };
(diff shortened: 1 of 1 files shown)


tacigar:
cc @lambcode

zipkin-lens/src/components/TracePage/TraceSummary.jsx
```diff
@@ -80,19 +83,34 @@ const TraceSummary = React.memo(({ traceSummary }) => {
...
   const shownTree = useMemo(() => {
-    const allHiddenSpanIds = {};
-    rerootedTree.forEach((span) => {
-      if (childrenHiddenSpanIds[span.parentId]) {
-        allHiddenSpanIds[span.spanId] = true;
+    let depth = 0;
+    let skip = false;
+
+    return rerootedTree.reduce((acc, cur) => {
+      if (cur.depth > depth && skip) {
```
anuraaga: This logic is pretty complex, can you add some comments? In particular, this propagation of `skip` is confusing me.

zipkin-lens/src/components/TracePage/TraceSummary.jsx
```diff
@@ -80,19 +83,34 @@ const TraceSummary = React.memo(({ traceSummary }) => {
...
 
   const shownTree = useMemo(() => {
-    const allHiddenSpanIds = {};
-    rerootedTree.forEach((span) => {
-      if (childrenHiddenSpanIds[span.parentId]) {
-        allHiddenSpanIds[span.spanId] = true;
+    let depth = 0;
+    let skip = false;
+
+    return rerootedTree.reduce((acc, cur) => {
```
anuraaga: Let's use semantic names instead of `acc`, `cur` - I think it's `shownTraces`, `trace`?

zipkin-lens/src/components/TracePage/TraceSummary.jsx
tacigar: I simplified this logic.
As you said, `skip` is unnecessary... 😓

tacigar:
PTAL 🙇

anuraaga (approved):

codefromthecrypt:
🙇 on behalf of all

PR #2959, opened 2020-01-20
README.md update for activemq username

rossms4:
Modified username environment key for ACTIVEMQ to match zipkin-server-shared.yml


codefromthecrypt:
thanks for the fix!

PR #2960, opened 2020-01-20
Works around mysql_install_db assumption on PAM things

codefromthecrypt:
the ever-red build notifications were annoying me


anuraaga:
Looks like needs copyright header update otherwise LGTM

PR #2965, opened 2020-01-23
added support for RabbitMQ prefetch count setting

sufiannazim:
Prefetch count limits number of unacknowledged deliveries permitted on a channel.


codefromthecrypt:
thanks for the help. a few questions:

* what version constraints are there (if any) on prefetch
* what benefit do you expect to see
* what value would you use and when would you adjust it?

we will eventually need guidance in the javadoc and/or readme

sufiannazim:
I will open another pull request, because I found couple other performance related issues in Zipkin Server.

codefromthecrypt:
ok great. PS we are due to upgrade our rabbitmq driver version. If you are up to helping with that, much oblidged!

PR #2968, opened 2020-01-28
Addressed Zipkin server crashing issue

sufiannazim:
**Environment:** 
Platform: Pivotal Cloud Foundry
Memory: 1GB
Span Collector: RabbitMQ
Storage: MySQL

**Issue:**
Server crashes with resources exhausted error, when RabbitMQ queue has couple thousand messages.

**Cause:**
RabbitMQ broker delivers messages to consumer channel as soon as they are published to a queue. Since, Zipkin RabbitMQ collector utilizes auto acknowledgment feature, broker continuously deliver incoming messages through consumer channel, it doesn’t wait for an ack for a previously delivered message.

**Solution:**
Use channel prefetch count with manual acknowledgement to limit number of inflight/unacknowledged deliveries permitted on a consumer channel. Message consumption rate can be tuned through concurrency and synchronous span collection execution mode configurations.

### DIFF
--- a/zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
+++ b/zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
@@ -13,6 +13,12 @@
  */
 package zipkin2.collector;
 
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.concurrent.Executor;
+import java.util.function.Supplier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import zipkin2.Callback;
@@ -22,16 +28,8 @@ import zipkin2.codec.BytesDecoder;
 import zipkin2.codec.SpanBytesDecoder;
 import zipkin2.storage.StorageComponent;
 
-import java.io.IOException;
-import java.io.UncheckedIOException;
-import java.nio.ByteBuffer;
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
-import java.util.concurrent.Executor;
-import java.util.function.Supplier;
-
 import static java.lang.String.format;
+import static java.util.logging.Level.FINE;
 import static zipkin2.Call.propagateIfFatal;
 
 /**
@@ -108,18 +106,16 @@ public class Collector { // not final for mock
   }
 
   public void accept(List<Span> spans, Callback<Void> callback) {
-    accept(spans, callback, Runnable::run, true);
+    accept(spans, callback, Runnable::run);
   }
 
   /**
-   *  Based on the async indicator collect spans synchronously or asynchronously
-   *
-   * @param spans spans to store
-   * @param callback callback function to invoke on success or error
    * @param executor the executor used to enqueue the storage request.
-   * @param async indicator used to determine whether to collect spans async or sync
+   *
+   * <p>Calls to get the storage component could be blocking. This ensures requests that block
+   * callers (such as http or gRPC) do not add additional load during such events.
    */
-  public void accept(List<Span> spans, Callback<Void> callback, Executor executor, boolean async) {
+  public void accept(List<Span> spans, Callback<Void> callback, Executor executor) {
     if (spans.isEmpty()) {
       callback.onSuccess(null);
       return;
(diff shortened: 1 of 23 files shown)


codefromthecrypt:
please revert the changes to the collector component and focus on the rabbitmq module instead. this is way too much change for what is needed. if you want to change the blocking mechanism in presence of the prefetch count that is ok but this is too much as it breaks api of other collectors

codefromthecrypt:
few other notes of interest:
* mysql is not intended for production use, and high throughput problems are expected in this storage
* there is a storage throttling feature in the server which may also help

sufiannazim:
Thank you for feedback, I will update the pull requests based on the comments.

sufiannazim:
By the way, performance issue was resolved with that fix. Performance test was done with 100K messages on a queue.

codefromthecrypt:
the asyncExecution thing which is to block the sender has been visited in other issues. you can look at the history and issues log to find out where we got to. there's also a design doc around the various problem space https://github.com/openzipkin/openzipkin.github.io/wiki/Collector-surge-and-error-handling

What I'm hoping is that you can lower the scope of your change so that it can be done with less design revisiting as what this PR did was change a base type. Also, it would probably be better to use gitter to discuss large design changes vs open multiple pull requests as the latter can send a flurry of notifications to watchers, some of whom have seen the same discussion many times and might prefer those issues looked at first before seeing another storm of the same content. https://gitter.im/openzipkin/zipkin

sufiannazim:
@adriancole I have checked in the latest updates.

zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
```diff
@@ -106,16 +108,18 @@ public Collector build() {
...
    *
-   * <p>Calls to get the storage component could be blocking. This ensures requests that block
-   * callers (such as http or gRPC) do not add additional load during such events.
+   * @param spans spans to store
+   * @param callback callback function to invoke on success or error
+   * @param executor the executor used to enqueue the storage request.
+   * @param async indicator used to determine whether to collect spans async or sync
    */
-  public void accept(List<Span> spans, Callback<Void> callback, Executor executor) {
+  public void accept(List<Span> spans, Callback<Void> callback, Executor executor, boolean async) {
```
codefromthecrypt: this is still a low level change. please lets discuss on chat? https://gitter.im/openzipkin/zipkin

codefromthecrypt (changes requested):
I appreciate the iteration, but to proceed to merging, you'll really need to remove the changes completely from zipkin-collector as I requested.

If you look carefully, the default is to invoke the command synchronously (`Runnable::run` not an actual thread pool). The extra logic is confusing knowing this.

Please use unit tests or otherwise if you need to explore the behavior. We want to merge the least code necessary as all code is a liability for maintenance, especially if statements and signature changes. Only with unit tests would we consider changing a main type and they would need to show a behavior that's not already achievable now.

sufiannazim:
yes, Runnable::run would call span storage in the same thread. The 'async' parameter in accept method is used to determine whether to call enqueue (asynchronous) or execute (synchronous) on storage.

codefromthecrypt:
if your concern is storage throttling, there's already a solution for this. please see the server README on that. The change to the collector tier will not be accepted. You need to try the solutions that exist prior to new core level change. it took us months to work out the throttling thing.


particularly, note that you are using mysql which is specifically unsupported for prod anyway. I've asked a few times so far to keep this contained to the rabbitmq specific stuff. If you are able to do that, please revert the other things and we can focus on that. otherwise, if you are unwilling to you can always fork the project.

sufiannazim:
@adriancole Updated pull request as requested.

codefromthecrypt:
I'm no longer helping with this pull request. I've asked many times to keep the change contained to rabbitmq, and to continue looking at this work displaces time from other things.

someone else can help ask you to revert the other changes and the mysql change you recently introduced. Otherwise, we should close this PR

codefromthecrypt:
to whoever, if anyone, continues this.

new features need explanation, justification and relevance. As we've noted on our README, mysql is not a production setup, so we should not be adding awkward new flags like isAsync to it.

Focusing on the change here, we should have the same rigor as looking at change like the elasticsearch shard settings.. ask "when would a user override prefetch?" "what's the side effect of acking, particuarly on error vs not at all?" "Is this acking similar to other collectors such as SQS?"

Because the shared understanding will help with this, especially if the next version of the driver changes the behaviour. Ideally someone is able to confirm the intent is actually useful, who has RMQ experience vs me reading docs etc. But for me, I've failed to even coerce this change to become contained. Anyway, this is my 2p for the next person, if there is one.

For now, I'm ignoring this issue as it is irritating to continue to ask for the same thing and yet it not occur.

sufiannazim:
Manual ACK with prefetch count setting on a channel prevents RabbitMQ broker from flooding the consumer with messages. These settings will not make any difference if you are acknowledging messages that are't completely processed. Non blocking span storage does not make sense for collectors like RabbitMQ, because it is processing messages of a queue. Span sender isn't waiting for any response.

Solution that I proposed in this pull request did resolve the performance issue in our environment.

Anyway, thank you for your time. I'm closing the pull request.

Issue #2970, opened 2020-02-04
Add chinese language support to lens


codefromthecrypt:
The only language people maintained in the classic UI was chinese, so we can probably ignore the others.

anuraaga:
Let me look into adding the i18n stuff to lens

tacigar:
I think this issue can be close.

Issue #2969, opened 2020-02-04
Read config.json in lens


codefromthecrypt:
Needed for https://github.com/openzipkin/zipkin/issues/2491

Issue #2519, opened 2019-04-19
Validate running lens under a reverse proxy


codefromthecrypt:
https://github.com/apache/incubator-zipkin/pull/1946 implemented reverse proxy configuration, where zipkin was mounted alongside other services. This was addressed by an env variable `ZIPKIN_UI_BASEPATH` described below:

> path prefix placed into the tag in the UI HTML; useful when running behind a reverse proxy. Default "/zipkin"

We need to read that variable from /config.json in lens, then ping @wdittmer @abesto to verify later (unless one of them wants to help implement!)

jcchavezs:
Maybe a browser test would be good here? I think we have some expertise in
my company I could borrow.

shakuzen:
> I think we have some expertise in my company I could borrow.

@jcchavezs just checking back to see if you or someone from your company would be able to help resolve this (it's one of the issues blocking switching the default UI to Lens).

jorgheymans:
I tested this with mod_proxy and a local httpd running on 8080

```
ProxyPass "/myzipkin"  "http://localhost:9411/zipkin"
ProxyPassReverse "/myzipkin"  "http://localhost:9411/zipkin"
```

Started zipkin-server using `java -jar target/zipkin-server-2.16.3-SNAPSHOT-exec.jar --zipkin.ui.basepath=/myzipkin`

Then accessing `http://localhost:8080/myzipkin` ends up loading lens correctly but the browser url becomes `http://localhost:8080/zipkin/` which does not seem desired.

codefromthecrypt:
great. Thanks @jorgheymans. this is progress

anuraaga:
I'm looking at this issue. I know in #1946 we added a mechanism for injecting the base path from the server to read in the client. But wondering, does anyone remember if there was any gotcha from just trying to extract the base path from the current `window.location`?

I think a URL that renders zipkin-lens can only look like one of these patterns
- https://fooserver.com/mysite/myzipkin/
- https://fooserver.com/mysite/myzipkin/dependency
- https://fooserver.com/mysite/myzipkin/traces/39048234098234
- https://fooserver.com/mysite/myzipkin/traceViewer

So simple processing of `window.location.href` can find that the base path is `https://fooserver.com/mysite/myzipkin/`

One corner case is if a user enters there own 404 URL like this
- https://fooserver.com/mysite/myzipkin/foobar/

[...]

Issue #2491, opened 2019-04-09
Implement logs button in lens


codefromthecrypt:
This is the description of the zipkin-ui README. Note that in lens we could implement the pattern differently. I recall someone asking to have properties besides traceId available. For example, localEndpoint.serviceName

## How do I find logs associated with a particular trace

Since zipkin provides a correlation id (the trace id), it's a good pattern to add it in your logs.
If you use zipkin, it's very likely that you have distributed logging sytem also and a way to query your logs (e.g. ELK stack).

One convenient way to switch from a trace to its logs is to get a button which directly links a trace to its logs.

This feature can be activated by setting the property `zipkin.ui.logs-url` or its corresponding environment variable:

* Kibana 3: `ZIPKIN_UI_LOGS_URL=http://kibana.company.com/query={traceId}`
* Kibana 4 or newer: `ZIPKIN_UI_LOGS_URL=http://kibana.company.com/app/kibana#/discover?_a=(index:'filebeat-*',query:(language:lucene,query:'{traceId}'))`
  This assumes that the log data is stored under the `filebeat-*` index pattern (replace this with the correct pattern if needed) and it uses the default timerange of 15 minutes (change it to 24 hours by adding `&_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-24h,mode:quick,to:now))` to the URL for example).

where `{traceId}` will be contextually replaced by the trace id.

If this feature is activated, you'll see on the trace detail page an additional button named `logs`.

![Logs Button](https://cloud.githubusercontent.com/assets/9842366/20482538/6e35ca66-afed-11e6-90e9-1e28f66d985e.png)

codefromthecrypt:
this implies reading /config.json

codefromthecrypt:
@tacigar fyi @shsamkit mentioned he can use lens once this is closed and until then is using classic

aerfus:
Hello,
@adriancole are there plans to implement the feature
in the near future?

codefromthecrypt:
we need someone to volunteer. there is no bench of people waiting for
tasks.. volunteer based. are you up for this?

shsamkit:
I can help with it.

aerfus:
Hello,
@adriancole sorry for misunderstanding.
I just wanted to know if someone is already implementing it...
@shsamkit do you need help?

codefromthecrypt:
generally unless someone assigns themself or makes a comment that it is
being worked on, no one is on it.

[...]

PR #2971, opened 2020-02-04
Makes Lens default, removing the dependency on the classic UI

codefromthecrypt:
For well over a year, we've been working on Lens, most notably @tacigar.

A year ago we added the ability to transition to the new Lens UI: a
"Try Lens" button which resulted in a lot of addressed feedback,
much of it personally done by @tacigar. Even before that, we asked
others to contribute and while some tasks have happened, not all.

We can't be in perpetual transition, and Lens is a very good UI. By
removing the classic one, ideally those really concerned about the
remaining features can help to implement them. Meanwhile, we allow
Zipkin to live longer by stopping the large and self-imposed burden
that displaces the small amount of time we have.

I don't think everyone will be happy about this, but that doesn't mean
continue to do nothing except add work on ourselves either. For this
reason I think we should proceed and remove classic on the next
release even if some folks don't like that.

cc @openzipkin/core @openzipkin/ui


codefromthecrypt:
For context, one feedback we got last year from Haystack was that they too had a try button. So few people used it, that they had to pull the cord.

IMHO adding more "try" and "suggest" features is not an option anymore. We have to commit to the health of this project over a sense of perfection.

jorgheymans:
Also those really infatuated with the classic UI always have the option to
stick to the previous version for a bit longer. +1 on making Lens default.

Op di 4 feb. 2020 04:49 schreef Adrian Cole <notifications@github.com>:

codefromthecrypt:
also they can mount the web assets on a proxy and still benefit from the
core changes. remember the classic UI only uses the rest api and
config.json there's no buddy or anything.

basvanbeek (approved):
LGTM.

Happy to see the gaps also commented in code as TODO.

anuraaga (approved):
With some documentation about nginx + zipkin-classic in the release notes, seems like it's relatively safe.

jcchavezs:
+1

shakuzen (approved):

tacigar (approved):
LGTM

huydx (approved):

codefromthecrypt:
FYI I offer to add documentation in release notes about nginx with classic. sure. still would be archived, which can formalize that we aren't doing security patch management on it, etc.

anuraaga:
Ah I was expecting the release notes for this repo when updating - `zipkin-classic` doesn't need any more work other than archival

eirslett:
LGTM! Cheers to the classic UI, it was fun while it lasted, long live Lens! 🍻 🍾 🎉

zeagord (approved):
It makes sense to retire Classic UI and makes Lens default. Less maintenance burden and Lens is also in a good shape now. We can add features based on requests and focus on that. I am good with this.

codefromthecrypt:
merging due to unanimous consent!

codefromthecrypt:
PS all the big ticket items listed here were later solved by @anuraaga!

* ZIPKIN_UI_BASEPATH for reverse proxy https://github.com/openzipkin/zipkin/issues/2519
* ZIPKIN_UI_LOGS_URL https://github.com/openzipkin/zipkin/issues/2491
* /config.json config overrides https://github.com/openzipkin/zipkin/issues/2969
* Chinese language support https://github.com/openzipkin/zipkin/issues/2970

PR #2980, opened 2020-02-18
Extract more strings for translation from DiscoverPage

anuraaga:
For #2970


zipkin-lens/src/components/DiscoverPage/TracesTab/messages.js
```diff
@@ -0,0 +1,27 @@
...
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+import { defineMessages } from 'react-intl';
+
+export default defineMessages({
+  duration: 'Duration',
+  filter: 'Filter',
+  numResults: `{num, number} {num, plural,
```
codefromthecrypt, line 20: is this correct? looks awkward..

zipkin-lens/src/translations/zh-cn.json
```diff
@@ -1,7 +1,17 @@
 {
   "src.components.App.changeLanguage": "改变语言",
+  "src.components.DiscoverPage.TracesTab.duration": "持续时间",
+  "src.components.DiscoverPage.TracesTab.filter": "",
```
codefromthecrypt, line 4: can someone help provide text for the empty strings here?

codefromthecrypt (commented):
I presume that before and after for english would be exactly the same.. that only in chinese we'd see a difference? (ex more translated)

zipkin-lens/src/components/DiscoverPage/TracesTab/messages.js
anuraaga: It's ICU syntax for rendering plurals common to pretty much all i18n libraries I think. We could be lazy and render `1 Results` if we want, but may as well take advantage of the feature

anuraaga:
Almost all of them yes, except `1 Results` now renders as `1 Result` :)

codefromthecrypt:
is it possible to comment this eventhough folks familiar will know? ex that
this syntax comes from xxx and will render like yyy

anuraaga:
Yeah added a comment

codefromthecrypt (approved):
I'm happy even if this merges without complete chinese translation. If someone complains, hopefully that person will be able to give us text. it would cause a patch release, but an easy one.

tacigar (approved):
Cool!

PR #2982, opened 2020-02-22
Infer Zipkin base path based on current URL.

### LINKED ISSUE #2519: Validate running lens under a reverse proxy

anuraaga:
Will add tests after getting an OK that the approach seems ok.

We previously passed a base path for zipkin through the server via a property. But I think it's relatively robust to just infer the base path from the browser URL. 

There is a difference regards to find traces URL - now they render at e.g., `/zipkin/?endTs...`, previously it was `/zipkin?endTs...` (note no trailing slash in that one). I remember we had an issue where we removed the trailing slash for this page, but I think with the unlaunch of classic (hence, no switching between the two UIs), this isn't an issue anymore, but I can't remember it well enough to be sure. Note that the server will redirect `/zipkin?endTs...` to `/zipkin/?endTs...` so both URLs will still be accepted.

Unblocks #2519


codefromthecrypt:
@jorgheymans would you mind giving this a spin? I think you helped verify once upon a time..

zipkin-lens/src/constants/api.js
```diff
@@ -13,8 +13,30 @@
...
+    return pathname.substring(0, pathname.length - '/traceViewer'.length);
+  }
+  const tracesIndex = pathname.lastIndexOf('/traces/');
+  if (tracesIndex !== -1) {
+    return pathname.substring(0, tracesIndex);
+  }
+
+  // Zipkin server always redirects from /zipkin to /zipkin/ but handle the non-redirected path too
+  // just in case.
+  return pathname.endsWith('/') ? pathname.substring(0, pathname.length - 1) : pathname;
```
jcchavezs, line 35: I think it is healthy to just avoid weird routing issues by just removing all `/` at the end using something like https://stackoverflow.com/a/6680877 instead of only looking at the last one. What do you think?

anuraaga: Here we're only inferring the base path rather than a full routing - I think that means that less sanitization is better than more since such extraneous `/` removal won't happen within react-router, which essentially concatenates the base path and paths to do routing so I've stuck with the simple match. Added a comment about it too. Let me know if this doesn't make sense.

anuraaga (commented):
Added tests - while I've only tested by hand-rewriting `/zipkin` to `/myzipkin` in the zipkin server code, rather than setting up an entire reverse proxy, I think I'm happy with this.

codefromthecrypt (approved):
I'd like to get manual verification with a proxy as used before, but doesn't seem like that's in the cards.

So, verification will happen after release when someone tries to use it! Worst case we have a small patch, which is better than holding back change.

jorgheymans:
I'll try and give this a spin today.

Op di 25 feb. 2020 07:15 schreef Adrian Cole <notifications@github.com>:

anuraaga:
Thanks a lot @jorgheymans! I don't have a config on hand which is why setting that up is a bit low priority for me - as @adriancole says, I think we're still better with a chance of it working rather than no chance, even if we don't get to manually verify :)

anuraaga:
Found the reason to not infer base path :) Static resources are still in the HTML so need to tweak base tag so they're served properly. Going to send another PR

PR #2983, opened 2020-02-23
Remove unused scss definitions and convert renaming to normal css.

anuraaga:
lens used to be an SCSS-styled app, but was migrated to Material UI and its JS styling solution. Now, we have a lot of unused variable definitions along with just a few external _css_ imports and a few lines of styles. We won't be adding any more styles here because we are using Material UI, so it makes sense to improve build performance a little by simplifying it. In particular, this removes the use of node-sass, a native extension which can cause some development headaches when the NodeJS version in pom.xml is different than that of the developer's machine.


zipkin-lens/package.json
```diff
@@ -52,12 +52,10 @@
     "jest-standard-reporter": "^1.0.1",
     "mini-css-extract-plugin": "^0.7.0",
     "node-fetch": "^2.3.0",
-    "node-sass": "^4.11.0",
```
codefromthecrypt, line 55: node-sass is a quite common problem area when people are building.. are we no longer using this? or is it just no longer an explicit dep?

codefromthecrypt (approved):
SGTM cc also @abesto

zipkin-lens/package.json
anuraaga: We're no longer using it with this PR - I think there was a ton of SCSS a while ago but now none :)

PR #2985, opened 2020-02-25
Add View Logs link to lens.

### LINKED ISSUE #2491: Implement logs button in lens

anuraaga:
I've mentioned considering `Grid` for layout before and did it here - I find it much simpler to layout items on a page only with markup instead of context-switching with the CSS any time a new item is added.

Fixes #2491

With logsUrl set
image

Without logsUrl set (unchanged)
image

### DIFF
--- a/zipkin-lens/src/components/TracePage/TracePage.jsx
+++ b/zipkin-lens/src/components/TracePage/TracePage.jsx
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2019 The OpenZipkin Authors
+ * Copyright 2015-2020 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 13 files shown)


codefromthecrypt (approved):
thanks for taking time to do this!

PR #2986, opened 2020-02-26
Bumps all things versioned to latest release

codefromthecrypt:


zipkin-storage/mysql-v1/src/main/java/zipkin2/storage/mysql/v1/internal/generated/DefaultCatalog.java
```diff
@@ -4,40 +4,30 @@
...
-import javax.annotation.Generated;
-
 import org.jooq.Schema;
 import org.jooq.impl.CatalogImpl;
 
 
 /**
  * This class is generated by jOOQ.
  */
-@Generated(
```
codefromthecrypt, line 20: these changes happened via the mysql README which has a command to run when changing jooq versions

codefromthecrypt:
localdate problem.. the least bad api is java 9+.. I will sort it shortly

anuraaga (approved):

PR #2987, opened 2020-02-26
Read base tag in zipkin-lens to allow reverse proxies.

### LINKED ISSUE #2519: Validate running lens under a reverse proxy

anuraaga:
I was tricked that #2982 works because dev server returns bundles fine regardless of URL but realized that for the actual build the base tag is important for loading bundles.

This extracts base path from the base tag like before and sets up routing with it as the basename. This allows all links in the site to just have the subpath, without /zipkin.

I have verified with @jorgheymans apache configuration that it works.

Fixes #2519 

Note URL bar:

image

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ui/ZipkinUiConfiguration.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ui/ZipkinUiConfiguration.java
@@ -15,29 +15,21 @@ package zipkin2.server.internal.ui;
 
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.linecorp.armeria.common.HttpData;
-import com.linecorp.armeria.common.HttpHeaderNames;
-import com.linecorp.armeria.common.HttpRequest;
-import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.HttpStatus;
 import com.linecorp.armeria.common.MediaType;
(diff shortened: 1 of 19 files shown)


zipkin-server/src/main/java/zipkin2/server/internal/ui/ZipkinUiConfiguration.java
```diff
@@ -176,8 +168,10 @@ static String maybeResource(String basePath, Resource resource) throws IOExcepti
       if (DEFAULT_BASEPATH.equals(basePath)) return content;
 
       String baseTagValue = "/".equals(basePath) ? "/" : basePath + "/";
+      // html-webpack-plugin seems to strip out quotes from the base tag when compiling so be
```
codefromthecrypt, line 171: nice comment!

codefromthecrypt (approved):

codefromthecrypt:
thanks mucho

jcchavezs:
Thanks demasiado

PR #2990, opened 2020-02-27
Fix docker examples prevents creating zipkin dashboard on grafana

cmlonder:
docker examples fail to create datasource for zipkin on grafana. Wrong URL changed to correct one which fixes the problem

### DIFF
--- a/docker/examples/prometheus/create-datasource-and-dashboard.sh
+++ b/docker/examples/prometheus/create-datasource-and-dashboard.sh
@@ -15,7 +15,7 @@
 
 set -xeuo pipefail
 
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
odd... https://grafana.com/docs/grafana/latest/http_api/dashboard/

cmlonder:
May be I could not describe correctly, example tries to check if datasource exist and adss if not, but it makes request to the dashboard, or I get something wrong?

codefromthecrypt:
I just think it odd, doesn't mean it isn't correct. will verify locally. thanks for contributing!

codefromthecrypt:
works for me! thanks 

merged as 271ee49bbeb01d5aad7e6ab610b5117be8a6ec46

PR #2991, opened 2020-02-27
Remove flexbox from top level app container to prevent flickering.

### LINKED ISSUE #2973: zipkin-lens constantly flashes in and out a scroll bar
## Describe the Bug
On Chrome in Windows, after opening a trace, the vertical scroll bar flashes in and out at ~60FPS causing a dancing page.

Screen capture tools can't capture the behavior since it seems to flash between the frames of a 30FPS video perfectly.

My camera can barely capture it
https://drive.google.com/open?id=1I3iSPvYDLdDxsQQPdbFJAfvkgyGprX88

## Steps to Reproduce
Start current master zipkin server locally with SELF_TRACING_ENABLED
Open home page a couple of times
Search traces with no conditions
Click on first trace

## Expected Behaviour
The page doesn't dance.

I'm not sure if this will repro for anyone else. But maybe if @tacigar has any ideas on any panel that may be missing an `overflow: hidden` that could cause fix corner cases

anuraaga:
I found this mostly by opening up inspector and randomly disabling CSS styles. But intuitively it does seem incorrect to apply flexbox and overflow: auto to the same element - both automatically reflow the content (flexbox by nature, overflow to compensate for a scroll bar) and probably are fighting each other in our flickeroo scenario.

From an encapsulation standpoint, it also seems fishy to have flexbox in the top level of the app since children can't rely lay themselves out correctly in such a case unless they are flexbox-aware themselves, hence encapsulation gets broken.

And when I remove this and open discover page, search for traces, go to trace, and open dependency page, I don't see any layout issues so it seems the flexbox isn't being used.

Fixes #2973

### DIFF
--- a/zipkin-lens/src/components/App/Layout.jsx
+++ b/zipkin-lens/src/components/App/Layout.jsx
@@ -57,8 +57,6 @@ const useStyles = makeStyles(theme => ({
     },
   },
   childrenWrapper: {
-    display: 'flex',
-    flexDirection: 'column',
     width: '100%',
     height: '100vh',
     overflow: 'auto',
--- a/zipkin-lens/webpack.dev.config.js
+++ b/zipkin-lens/webpack.dev.config.js
@@ -22,7 +22,7 @@ module.exports = {
   entry: path.join(__dirname, './src/index.js'),
   output: {
     filename: 'bundle.js',
-    publicPath: '/zipkin',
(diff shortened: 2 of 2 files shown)


codefromthecrypt (approved):
tried locally and I can't see any flickers.. was easier before!

PR #3002, opened 2020-03-03
Migrate from react-intl to linguijs.

anuraaga:
I originally used `react-intl` for localization since I've used it before and it's fairly common. However, when looking into migrating to `create-react-app`, I found `react-intl` doesn't work very well with it. While not a blocker, I looked around for alternatives and found LinguiJS, and I find it is better designed by providing very powerful macros. Also, initializing the i18n provider is far more intuitive (I didn't understand why I have to be the one to find out what messages to provide to react-intl...)

A comparison with `react-intl` - https://lingui.js.org/misc/react-intl.html

One thing I particularly like is `linguijs` supports patterns of using the source text itself as a message ID in addition to explicit IDs.

https://lingui.js.org/tutorials/react.html#message-id

As it says, the decision for one or the other is case-by-case. I think Zipkin is not YouTube, we don't have that many strings and the chance of having two strings with the same English but different translations (due to context differences) is very low that we don't have to worry about it. By using source language, our markup becomes easier to read (no importing in from a separate messages module) and more concise. 

Aside: it's quite fun being able to make this change without updating any test files!


zipkin-lens/src/components/App/LanguageSelector.jsx
```diff
@@ -72,7 +71,7 @@ const LanguageSelector = () => {
   return (
     <>
       <SidebarMenu
-        title={intl.formatMessage(messages.changeLanguage)}
+        title={i18n._(t`Change Language`)}
```
codefromthecrypt, line 74: neat

zipkin-lens/src/components/DependenciesPage/DependenciesPageHeader.jsx
```diff
@@ -84,13 +84,13 @@ const DependenciesPageHeader = React.memo(({
...
   const classes = useStyles();
-  const intl = useIntl();
+  const { i18n } = useLingui();
 
   return (
     <Box className={classes.root}>
       <Box className={classes.upperBox}>
         <Typography variant="h5" className={classes.pageTitle}>
-          {intl.formatMessage(messages.dependencies)}
+          <Trans>Dependencies</Trans>
```
codefromthecrypt, line 93: this fake tag thing is neat

codefromthecrypt (approved):
looks intuitive. Thanks for the README update

PR #3003, opened 2020-03-03
Use react-router link in side menu for SPA-type navigation and clear …

anuraaga:
…discover page when navigating back to it from another page.

Also fixes the absolute URL issue in the sidebar.

This allows all the navigation to be in the SPA, currently only search results are.


codefromthecrypt (approved):
I think I understand all of this!

kubukoz:
Awesome, I was about to report/try to fix the absolute URL issue, glad it's fixed! (it's not in the docker image yet)

PR #3004, opened 2020-03-03
Allows lens to open traces in a new tab

### LINKED ISSUE #2998: Cannot open traces in a new tab
## Describe the Bug
In the latest version of lens, I'm unable to open the search results in a new tab. If I cmd + click on a trace it still opens it in the current window and if I right click there's no "open in new tab" option.

## Steps to Reproduce
search for anything and try to open any of the result traces in a new tab.

tacigar:
fixes #2998 

I think basically there are two ways to move pages, using `history.push` and using an `<a/>` tag.
However, when using `history.push`, the page cannot be displayed in a new tab as pointed out #2998 .

So this PR use `<a/>` tag (Strictly speaking `<Link/>` component of react-router-dom) instead of `history.push`.

### DIFF
--- a/zipkin-lens/src/components/DiscoverPage/TracesTab/TracesTableRow.jsx
+++ b/zipkin-lens/src/components/DiscoverPage/TracesTab/TracesTableRow.jsx
@@ -102,7 +102,7 @@ export const TracesTableRowImpl = ({
   const { spanName, serviceName } = rootServiceAndSpanName(correctedTrace);
 
   return (
-    <Box className={classes.root} data-test="root">
+    <Box className={classes.root}>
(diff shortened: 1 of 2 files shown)


tacigar:
cc @drolando

codefromthecrypt:
@drolando if you have a chance to try this, would be best!

codefromthecrypt (approved):
I don't fully understand because I don't know enough about the context (not very experienced). However, the changes look clear in code and I appreciate the easier to read unit test :D

drolando (approved):
It works now. Thanks for the quick fix!

codefromthecrypt:
Thanks @tacigar! awesome

PR #3006, opened 2020-03-05
Don't render search box when searchEnabled is false.

### LINKED ISSUE #3000: Lens does not respect SEARCH_ENABLED=false
## Describe the Bug
When I set `SEARCH_ENABLED=false` I expect lens search bar to be disabled. Instead right now I can still try to write a search, but all the dropdowns are empty since there are no indexes.
And ideally I'd like to have the same banner as in the old UI that says that search is disabled and to input the traceId in the top right textbox

## Steps to Reproduce
Run zipkin with SEARCH_ENABLED=false

anuraaga:
image

Fixes #3000


codefromthecrypt:
fixing our 3000th issue huh?

zipkin-lens/src/components/DiscoverPage/DiscoverPage.jsx
```diff
@@ -125,7 +126,23 @@ const propTypes = {
   setLimitCondition: PropTypes.func.isRequired,
 };
 
-const DiscoverPageImpl = ({
+const DiscoverPageHeader = () => {
```
tacigar, line 129: How about splitting files for each components?
This is not a strong opinion, but other components is in the individual file..

codefromthecrypt (approved):

codefromthecrypt:
will leave the code curation comment for post-merge, as our docker images are busted and having this in the next patch will be helpful!

Thanks @anuraaga and @tacigar!

anuraaga:
@tacigar Yeah I noticed we currently have only one component per file. I went with keeping in the file here since there's no package private in js - having this splitting which is more an implementation detail rather than an encapsulation of APIs seems good. Especially for pure wrappers like the one we export here.

TBH, I currently find it a bit hard to understand the layout of some components due to the sheer number of files I have to jump around and wish there were fewer files in general.

@tacigar What do you think?

PR #3009, opened 2020-03-06
Updates to latest Spring Boot and Armeria

codefromthecrypt:


pom.xml
```diff
@@ -75,13 +75,13 @@
     <slf4j.version>1.7.30</slf4j.version>
 
     <junit.version>4.13</junit.version>
-    <junit.jupiter.version>5.6.0</junit.jupiter.version>
+    <junit-jupiter.version>5.6.0</junit-jupiter.version>
```
codefromthecrypt, line 78: switched this as our other stuff are artifact name based, rather than group ID based. This makes searching mildly easier

PR #3013, opened 2020-03-09
Try disabling shallow clone to see if it fixes license plugin flakiness.

### LINKED ISSUE #1512: Maven license plugin fails in travis

anuraaga:
I removed the comments about license skipping since I think `skipTests` is essentially the same. I could add something like `Disable all checks since we do it in the mvn verify step of publish.sh` to them all instead though if that seems helpful though.

This causes `git clone` to take much longer than before - meaning a couple of seconds now takes a few seconds and has very little impact on the entire build so shouldn't be an issue.

Might fix #1512

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@
 sudo: required
 dist: trusty
 
+# Don't do a shallow clone to allow license plugin to correctly read git history.
+git:
+  depth: false
+
 cache:
   directories:
   # zipkin-lens gets dependencies via NPM
@@ -37,7 +41,7 @@ before_install:
   - git config credential.helper "store --file=.git/credentials"
   - echo "https://$GH_TOKEN:@github.com" > .git/credentials
 
-# Override default travis to use the maven wrapper; skip license on travis due to #1512
+# Override default travis to use the maven wrapper
 install: ./mvnw install -DskipTests=true -Dlicense.skip=true -Dmaven.javadoc.skip=true -B -V
 script: ./travis/publish.sh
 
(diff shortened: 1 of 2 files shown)


codefromthecrypt (approved):
later we can follow-up to stitch license check to not skipTests, but not critical

PR #3014, opened 2020-03-09
Reduce openzipkin links to just home page and allow a user-configurab…

anuraaga:
…le support link.

Discussion in #2375

After (with supportUrl defined)
image

After (without supportUrl defined)
image

Before
image


codefromthecrypt:
maybe before and after image to make it more clear?

zipkin-lens/src/translations/es/messages.json
```diff
@@ -12,6 +12,7 @@
   "End Time": "Tiempo de Fin",
   "Failed to load this file": "Fallo al cargar este archivo",
   "Filter": "Filtro",
+  "Help": "",
```
jeqo:
```suggestion
  "Help": "Ayuda",
```

shakuzen (approved):
Looks good to me. Thanks for making the changes! Could we try to get a review from the user who gave the initial feedback?

codefromthecrypt:
Should we use English "Help" or "Support" ? I tend to think of the word support, and whatever we put in here lives in config, so figured worth a short bikeshed.

As far as I can tell in the code, by default we don't add a default now. Maybe a screenshot of default (help/support defined) can ... help :)

@jcarres-mdsol does this make sense for configuring your slack url?
@Monkeyanator thoughts on the reduced UI clutter (for lack of better word)?

anuraaga:
Was hoping to leave the other screenshot to imagination ;) But added it.

Have no opinion at all on `help` or `support` - if no one else votes on it I'll rename to support.

codefromthecrypt:
I like the without visually as it balances the white icons top and bottom. Nice!

zipkin-lens/src/translations/es/messages.json
```diff
@@ -12,6 +12,7 @@
   "End Time": "Tiempo de Fin",
   "Failed to load this file": "Fallo al cargar este archivo",
   "Filter": "Filtro",
+  "Help": "Ayuda",
```
jeqo: if changing to "Support": 

```suggestion
  "Help": "Soporte",
```

jeqo (approved):

jcchavezs:
+1

anuraaga:
Renamed from help -> support

codefromthecrypt:
Good job team. thanks for driving this home @anuraaga!

PR #3016, opened 2020-03-11
Fix all lint errors in lens

tacigar:
Since my VSCode reported a lot of lint errors, I fixed them all.


zipkin-lens/package.json
```diff
@@ -9,7 +9,7 @@
     "test": "jest --ci --reporters=jest-standard-reporter",
     "test-dev": "jest",
     "coverage": "jest --coverage",
-    "lint": "eslint src",
+    "lint": "eslint 'src/**/*.{js,jsx}'",
```
tacigar, line 12: I'm sorry, I made a mistake setting the lint command... 😢

codefromthecrypt (approved):
this kind of thing is not worth rebasing.. let's merge!

codefromthecrypt:
I mean it quickly drifts not just here but anything in progress. thanks for the help!

zipkin-lens/package.json
codefromthecrypt: happens to us all!

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2019 The OpenZipkin Authors
+ * Copyright 2015-2020 The OpenZipkin Authors
```
anuraaga: Thanks for this PR - just checking is lint also checking license header? It looks like it may be doing so differently than maven - instead of worrying about consistency between two checkers it would be good to disable the lint one.

anuraaga (approved):

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.jsx
tacigar: Oh, sorry. eslint does not check license header.
However, now, some files are 2019 and some files are 2020. I think it is strange, so I changed all the files because I think it’s just a good opportunity...
If there are problems, I'll revert it :)

codefromthecrypt: I think it best to revert even though some files will change later. ex we don't change all headers when the year changes.

tacigar:
Revert license header changes! 🙇

anuraaga (approved):

codefromthecrypt:
thanks again @tacigar !

PR #3017, opened 2020-03-11
Only run license check when executing tests

codefromthecrypt:
Thanks to @anuraaga for the hint that this can improve performance. This
also reduces the commandline length.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -546,82 +546,6 @@
       </plugin>
 
       <plugin>
-        <groupId>com.mycila</groupId>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -676,6 +600,93 @@
         <module>zipkin-lens</module>
       </modules>
     </profile>
+    <profile>
+      <id>license-check</id>
+      <activation>
+        <property>
```
anuraaga, line 606: Ah my maven-fu is too weak - was expecting a new profile but this is ideal.

anuraaga (approved):

PR #3018, opened 2020-03-11
Add Archive Trace button

drolando:
This button lets you easily reupload the current trace to a different
server.

The main motivation for having this is that you can have the
archival server have a very long retention period and use it as very
long term storage for traces that you care about. For example when
sharing a trace in a jira ticket since otherwise the link would expire
after 1 week.

Design doc explaining the reason behind this in more details and why we
went with this implementation: https://github.com/openzipkin/openzipkin.github.io/wiki/Favorite-trace

### DIFF
--- a/zipkin-lens/src/components/App/App.jsx
+++ b/zipkin-lens/src/components/App/App.jsx
@@ -19,6 +19,7 @@ import CircularProgress from '@material-ui/core/CircularProgress';
 import { ThemeProvider } from '@material-ui/styles';
 import { MuiPickersUtilsProvider } from '@material-ui/pickers';
 import MomentUtils from '@date-io/moment';
+import { SnackbarProvider } from 'notistack';
 
 import Layout from './Layout';
 import DiscoverPage from '../DiscoverPage';
@@ -39,29 +40,32 @@ const App = () => {
       <UiConfig>
         <MuiPickersUtilsProvider utils={MomentUtils}>
           <ThemeProvider theme={theme}>
-            <UiConfigConsumer>
-              {(config) => (
-                <Provider store={configureStore(config)}>
-                  <I18nProvider i18n={i18n}>
-                    <BrowserRouter basename={BASE_PATH}>
-                      <Layout>
-                        <Route exact path="/" component={DiscoverPage} />
-                        <Route
-                          exact
-                          path="/dependency"
-                          component={DependenciesPage}
-                        />
-                        <Route
-                          exact
-                          path={['/traces/:traceId', '/traceViewer']}
-                          component={TracePage}
-                        />
-                      </Layout>
-                    </BrowserRouter>
-                  </I18nProvider>
-                </Provider>
-              )}
-            </UiConfigConsumer>
+            {/* Snackbar is used to provide popup alerts to the user */}
+            <SnackbarProvider>
+              <UiConfigConsumer>
+                {(config) => (
+                  <Provider store={configureStore(config)}>
+                    <I18nProvider i18n={i18n}>
+                      <BrowserRouter basename={BASE_PATH}>
+                        <Layout>
+                          <Route exact path="/" component={DiscoverPage} />
+                          <Route
+                            exact
+                            path="/dependency"
+                            component={DependenciesPage}
+                          />
+                          <Route
+                            exact
+                            path={['/traces/:traceId', '/traceViewer']}
+                            component={TracePage}
+                          />
+                        </Layout>
+                      </BrowserRouter>
+                    </I18nProvider>
+                  </Provider>
+                )}
+              </UiConfigConsumer>
+            </SnackbarProvider>
           </ThemeProvider>
         </MuiPickersUtilsProvider>
       </UiConfig>
--- a/zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
+++ b/zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
@@ -11,7 +11,6 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  */
-/* eslint-disable no-alert */
 import { t, Trans } from '@lingui/macro';
 import { useLingui } from '@lingui/react';
 import PropTypes from 'prop-types';
@@ -23,6 +22,7 @@ import Box from '@material-ui/core/Box';
 import Grid from '@material-ui/core/Grid';
 import Button from '@material-ui/core/Button';
 import Typography from '@material-ui/core/Typography';
+import { useSnackbar } from 'notistack';
 
 import { useUiConfig } from '../UiConfig';
 
@@ -114,7 +114,20 @@ const TraceSummaryHeader = React.memo(({ traceSummary, rootSpanIndex }) => {
       ? config.archiveUrl.replace('{traceId}', traceSummary.traceId)
       : undefined;
 
+  const { enqueueSnackbar } = useSnackbar();
+
   const archiveClick = useCallback(() => {
+    const notify = (message, variant) => {
+      enqueueSnackbar(message, {
+        variant,
+        anchorOrigin: {
+          vertical: 'top',
+          horizontal: 'center',
+        },
+        autoHideDuration: 10000, // 10 seconds
+      });
+    };
+
     // We don't store the raw json in the browser yet, so we need to make an
     // HTTP call to retrieve it again.
     fetch(`${api.TRACE}/${traceSummary.traceId}`)
@@ -135,37 +148,37 @@ const TraceSummaryHeader = React.memo(({ traceSummary, rootSpanIndex }) => {
             break;
           }
         }
-
-        fetch(archivePostUrl, {
+        return json;
+      })
+      .then((json) => {
+        return fetch(archivePostUrl, {
           method: 'POST',
           headers: {
             'Content-Type': 'application/json',
           },
           body: JSON.stringify(json),
-        })
-          .then((response) => {
-            if (
-              !response.ok ||
-              (response.status !== 202 && response.status === 200)
-            ) {
-              throw new Error('Failed to archive the trace');
-            }
-            if (archiveUrl) {
-              alert(
-                `Archive successful! This trace is now accessible at ${archiveUrl}`,
-              );
-            } else {
-              alert(`Archive successful!`);
-            }
-          })
-          .catch(() => {
-            alert('Failed to archive the trace');
-          });
+        });
+      })
+      .then((response) => {
+        if (
+          !response.ok ||
+          (response.status !== 202 && response.status === 200)
+        ) {
+          throw new Error('Failed to archive the trace');
+        }
+        if (archiveUrl) {
+          notify(
+            `Archive successful! This trace is now accessible at ${archiveUrl}`,
+            'success',
+          );
+        } else {
+          notify(`Archive successful!`, 'success');
+        }
       })
       .catch(() => {
-        alert('Failed to fetch trace from backend');
+        notify('Failed to archive the trace', 'error');
       });
-  }, [archivePostUrl, archiveUrl, traceSummary]);
+  }, [archivePostUrl, archiveUrl, traceSummary, enqueueSnackbar]);
 
   const handleSaveButtonClick = useCallback(() => {
     if (!traceSummary || !traceSummary.traceId) {
--- a/zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
+++ b/zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
@@ -11,6 +11,7 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  */
+/* eslint-disable no-alert */
 import { t, Trans } from '@lingui/macro';
 import { useLingui } from '@lingui/react';
 import PropTypes from 'prop-types';
@@ -104,6 +105,67 @@ const TraceSummaryHeader = React.memo(({ traceSummary, rootSpanIndex }) => {
     config.logsUrl && traceSummary
       ? config.logsUrl.replace(/{traceId}/g, traceSummary.traceId)
       : undefined;
+
+  const archivePostUrl =
+    config.archivePostUrl && traceSummary ? config.archivePostUrl : undefined;
+
+  const archiveUrl =
+    config.archiveUrl && traceSummary
+      ? config.archiveUrl.replace('{traceId}', traceSummary.traceId)
+      : undefined;
+
+  const archiveClick = useCallback(() => {
+    // We don't store the raw json in the browser yet, so we need to make an
+    // HTTP call to retrieve it again.
+    fetch(`${api.TRACE}/${traceSummary.traceId}`)
+      .then((response) => {
+        if (!response.ok) {
+          throw new Error('Failed to fetch trace from backend');
+        }
+        return response.json();
+      })
+      .then((json) => {
+        // Add zipkin.archived tag to root span
+        /* eslint-disable-next-line no-restricted-syntax */
+        for (const span of json) {
+          if ('parentId' in span === false) {
+            const tags = span.tags || {};
+            tags['zipkin.archived'] = 'true';
+            span.tags = tags;
+            break;
+          }
+        }
+
+        fetch(archivePostUrl, {
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify(json),
+        })
+          .then((response) => {
+            if (
+              !response.ok ||
+              (response.status !== 202 && response.status === 200)
+            ) {
+              throw new Error('Failed to archive the trace');
+            }
+            if (archiveUrl) {
+              alert(
+                `Archive successful! This trace is now accessible at ${archiveUrl}`,
+              );
+            } else {
+              alert(`Archive successful!`);
+            }
+          })
+          .catch(() => {
+            alert('Failed to archive the trace');
+          });
+      })
+      .catch(() => {
+        alert('Failed to fetch trace from backend');
+      });
+  }, [archivePostUrl, archiveUrl, traceSummary]);
 
   const handleSaveButtonClick = useCallback(() => {
     if (!traceSummary || !traceSummary.traceId) {
@@ -204,6 +266,24 @@ const TraceSummaryHeader = React.memo(({ traceSummary, rootSpanIndex }) => {
                   className={classes.actionButtonIcon}
                 />
                 <Trans>View Logs</Trans>
+              </Button>
+            </Grid>
+          )}
+          {archivePostUrl && (
+            <Grid item>
+              <Button
+                variant="outlined"
+                className={classes.actionButton}
+                target="_blank"
+                rel="noopener"
+                data-testid="archive-trace-link"
+                onClick={archiveClick}
+              >
+                <FontAwesomeIcon
+                  icon={faFileAlt}
+                  className={classes.actionButtonIcon}
+                />
+                <Trans>Archive Trace</Trans>
               </Button>
             </Grid>
           )}
--- a/zipkin-lens/src/translations/es/messages.json
+++ b/zipkin-lens/src/translations/es/messages.json
@@ -36,7 +36,7 @@
   "Trace ID": "ID de Traza",
   "Upload JSON": "Subir JSON",
   "View Logs": "Ver Logs",
-  "Archive Trace": "",
+  "Archive Trace": "Archivar Traza",
   "Zipkin Home": "Zipkin Inicio",
   "hide annotations": "ocultar anotaciones",
   "show all annotations": "mostrar todas las anotaciones",
--- a/zipkin-lens/src/translations/es/messages.json
+++ b/zipkin-lens/src/translations/es/messages.json
@@ -36,6 +36,7 @@
   "Trace ID": "ID de Traza",
   "Upload JSON": "Subir JSON",
   "View Logs": "Ver Logs",
+  "Archive Trace": "",
   "Zipkin Home": "Zipkin Inicio",
   "hide annotations": "ocultar anotaciones",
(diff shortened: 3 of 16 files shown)


drolando:
I still need to figure out how translations work, but at least this is now out for people to comment on.

zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
```diff
@@ -104,6 +104,51 @@ const TraceSummaryHeader = React.memo(({ traceSummary, rootSpanIndex }) => {
...
+      .then((json) => {
+        fetch(archiveWriteUrl, {
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify(json),
+        })
+        .then((response) => {
+          if (!response.ok || response.status != 202) {
```
codefromthecrypt: I think some clones incorrectly use 200

zipkin-server/README.md
```diff
@@ -172,6 +172,8 @@ searchEnabled | zipkin.ui.search-enabled | If the Find Traces screen is enabled.
 queryLimit | zipkin.ui.query-limit | Default limit for Find Traces. Defaults to 10.
 instrumented | zipkin.ui.instrumented | Which sites this Zipkin UI covers. Regex syntax. e.g. `http:\/\/example.com\/.*` Defaults to match all websites (`.*`).
 logsUrl | zipkin.ui.logs-url | Logs query service url pattern. If specified, a button will appear on the trace page and will replace {traceId} in the url by the traceId. Not required.
+archiveWriteUrl | zipkin.ui.archive-write-url | Url to archive the current trace. If specified, a button will appear on the trace page and will let you upload the current trace to the specified url. Not required.
```
```suggestion
archivePostUrl | zipkin.ui.archive-post-url | Url to POST the current trace in Zipkin v2 json format. e.g. 'https://longterm/api/v2/spans'. If specified, a button will appear on the trace page accordingly. Not required.
```

zipkin-server/README.md
```diff
@@ -172,6 +172,8 @@ searchEnabled | zipkin.ui.search-enabled | If the Find Traces screen is enabled.
...
+archiveReadUrl | zipkin.ui.archive-read-url | Url to the archived trace. If specified, we will show it in the confirmation message after archiving a trace and replace {traceId} with the current traceId. Not required.
```
```suggestion
archiveUrl | zipkin.ui.archive-url | Url to a web application serving an archived trace, templated by '{traceId}'. e.g. https://longterm/zipkin/trace/{traceId}'. This is shown in a confirmation message after a trace is successfully POSTed to the `archivePostUrl`. Not required.
```

codefromthecrypt (commented):
some bike shed a little..

archiveWriteUrl -> archivePostUrl (as it is a POST)
archiveReadUrl -> archiveUrl (shorter.. emphasizing it is a web link not a raw json url)

mchandramouli:
@drolando @adriancole - this might be a bit late. I read the need @ https://github.com/openzipkin/openzipkin.github.io/wiki/Favorite-trace.  Instead of a second cluster, one can have a second storage with higher retention and do scatter/gather on search. When someone does an explicit search for a traceid, move the trace to the second storage in the background. This allows the trace to still show up in the cluster and be available longer. Not sure if zipkin allows multiple storage DBs and scatter / gather on search - if it does, this might be easier than a second archive cluster from user perspective as saved links in jira or issues will continue to work as is. This is what was being done when I last worked with Haystack team

codefromthecrypt:
@mchandramouli thanks for the context. I think for some setups an internal tiering could work. 

There is additional configuration complexity to achieve that.. we don't have anything that could do that internally, especially as TTL is implemented differently (if defined at all). For example, in cassandra it is literally copying the rows again (to update the TTL). Also, much of the configuration would double to achieve the goal.. I think we've talked about this tension in the past. It gets more complex when storage is different schema or tech per tier. These decisions are mostly backend I think anyway, and this change is all frontend.

It is a very good point on readback/permalink, we do have a change here to allow a different URL. However, there's no requirement to use one. Ex yelp did what you said with scatter gather in the past https://github.com/drolando/zipkin-mux. We don't have to use a separate URL space for things explicitly saved, or otherwise tiered to a longer TTL. If we mount the UI assets so its /zipkin path is mounted against a proxy api, it will never know the difference.

Stepping back, this change is focused on is users being explicitly able to choose traces, which is possibly higher signal of interest than queries, albeit manually trained. Our UI saves data from the list screen to the trace detail. A trace id-specific api call never occurs in this flow, unless the user hit the trace link directly. In other words, a user who is discovering a trace from a list will show no side-effects to the backend that they are looking at one trace. Basically you'd have to assume that any queried data with no parameters except time imply interest, then push all those a tier down. I think this is something that could require tuning to avoid chattiness, because I would suspect many just hit refresh a lot. However, I agree even the wider net would be significantly smaller than all traces: time-based is better than pushing all traces a tier back.

Automatic tiering is nevertheless an interesting idea sites can consider... maybe this feature can help build tension needed to explore that. Meanwhile, they have a lot of options that are easy as pie.

In simplest case you can imagine someone who is already using our in-memory server, and wants to save off traces to ES or even a cloud provider (many accept zipkin format now). In the case of cassandra, re-posting to another cluster is similar impact to overwriting the same rows with a new TTL. TL;DR; In this design there's no constraint in which technologies are in use, hence very big bank for the buck.

I'm very happy to hear about the story you mentioned though.. it is no doubt helpful for people to know existing practice! thanks!

zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
```diff
@@ -104,6 +104,51 @@ const TraceSummaryHeader = React.memo(({ traceSummary, rootSpanIndex }) => {
...
+        }
+        return response.json();
+      })
+      .then((json) => {
+        fetch(archiveWriteUrl, {
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify(json),
```
codefromthecrypt, line 159: We once thought about adding a tag, so it is easier to search for traces manually chosen. Do we want to do this?

ex. add a tag to the root span "archived" -> "" would be simplest

cc @openzipkin/core

shakuzen: Sounds reasonable to me. I wonder if a user might have an `archived` tag for their own purposes separate from this use case. If so, perhaps we should prefix the tag with `zipkin` or such so it is clear this is a sort of internal tag not part of user data.

jcchavezs:
My 2p: So if I am not wrong, this traces will be uploaded in a different
instance of zipkin (that is always the case, otherwise no point) hence
there is no way to mix normal spans from the archived ones, isn't? That
said, is the tag still needed?

If the tag is needed for a visibility reason (e.g. lens is going to show an
archived label in the trace view to make sure you don't accidentally
confuse with a normal trace) then this could make sense. I would still go
for `zipkin.archived` trace to make sure we don't mess up with other traces.

Letting imagination flight, it could be interesting to have a
`zipkin.archive.reason` tag to allow the user to specify the reason for
this trace to be interesting.

Finally I would like to think loud, since the archived trace is just
another zipkin instance, that's it make sense to add a link to archived
zipkin UI? Or even beyond that, maybe we can just use the normal instance's
lens to call the archived API? Something like
myzipkin:9411/zipkin/search/srchived/?blah=blah.

In any case I am in for giving visibility to archived traces in the
day-to-day zipkin UI.

codefromthecrypt:
@jcchavezs I think we did discuss having a reason at some point.. ex someone can put a JIRA ticket in the tag value. Good idea. basically someone can enter a choice of that tag value or nothing to just save empty.

On the other topic, I wouldn't assume it is definitely a zipkin server that's being archived to. It could be any APM, or even haystack. We can't really assume we know the link syntax, but that's what the POST-back is there for if it can work. Same as logs-url, if someone can express a query, they can populate the URL.

I'd recommend stopping there, short of bookmark service though. We already overloaded screen issues (ex #3014) so suggest we don't do too much adding of more links until there's more practice to show they are helpful. my 2p

jorgheymans:
Mentioning https://github.com/openzipkin/zipkin/pull/1747 and https://github.com/openzipkin/zipkin/issues/1093 as prior discussions of this feature, there may be more.

codefromthecrypt:
a shower helped me recognize why we can easily branch discussion into automated escalation of traces to another tier, when looking at this. The reason I think is that this change can feel very much like user-assisted late (after-the-fact or tail) sampling.

Ex for the same reasons @mchandramouli mentioned. @jeqo and others have discussed how automated late sampling can occur, ex by storing queries to the api. This led to a design idea  that a collector tier component would actually need to consider the entire trace. We had a couple initiatives to play with this including the [VoltDB experiment](https://github.com/adriancole/zipkin-voltdb) and a later more fleshed out [Kafka Streams](https://github.com/openzipkin-contrib/zipkin-storage-kafka) approach.

What I'd recommend in this change is to be conscious that something like this could help train automation, the automation itself could be the POST endpoint here! That said, we can probably develop things around tail sampling separately, mentioning this for context and visa versa. I'm really happy that folks helped participate here.. I feel clearer about the boundaries and am glad we can have a tie-in for those who can setup a late sampling arch.

We do owe ourselves a combined wiki on late sampling though, as we do on other things. It is a bit difficult to hold the context of the various efforts together, so a volunteer to harvest some of this thread and others about late sampling into a design doc would be welcome https://github.com/openzipkin/openzipkin.github.io/wiki/Designs

mchandramouli:
> Automatic tiering is nevertheless an interesting idea sites can consider... maybe this feature can help build tension needed to explore that. Meanwhile, they have a lot of options that are easy as pie.
> What I'd recommend in this change is to be conscious that something like this could help train automation, the automation itself could be the POST endpoint here! 

+1

I think this can lead to an "adaptive tiering/sampling" - agree on taking that to a different discussion and limit this to the change @drolando has started

On the perma link - one other thought (again could be another discussion) is that have Zipkin lens search both live and archive cluster (beyond live cluster time range) so archived traces can appear in the same UI

codefromthecrypt:
thx @mchandramouli filed https://github.com/openzipkin/zipkin/issues/3023 for follow-up on client-side proxying (scatter gather)

It occurs to me it is important to do the tagging here on POST. We should also make the archive url conditional on no root span tag of the same. Ex if we use the tag zipkin.archive, do not offer to archive it again :)

zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
drolando: I added this check on the status code after I mistakenly set the url to `localhost:9412/zipkin`. The homepage happily returned 200 to my POST request and I spent a while trying to figure out why the trace wasn't saved.

I can remove it though

codefromthecrypt: we can wait until someone asks for 200 support. it is not following our spec to use 200 anyway.

I'm really surprised homepage accepted POST! open an issue?

drolando:
Tests are now passing! I've also added a `zipkin.archived=true` tag

drolando:
@tacigar @anuraaga I've tried using react-alerts to display nicer popups than just bare alerts with little effort. I saw that material-ui also has an Alert component but it seems much more involved to use. Let me know if you're fine with adding an extra JS dependency or not.

Screenshot 2020-03-15 15 52 15
Screenshot 2020-03-15 15 51 56

Unfortunately I couldn't convince the alert to properly show the link. Adding an `<a href="..."> ... </a>` just resulted in the `<a>` tag being printed as well as it's not recognized as html...

tacigar:
@drolando 
I know material-ui's snackbar is difficult to use by itself, but it is relatively easy when used with notistack (you no longer need to write Snackbar component directly! notistack uses material-ui's Snackbar internally).
I'm not familiar with react-alert, but as far as I see a diff in your code, the usage of react-alert and notistack seems to be similar.
https://material-ui.com/components/snackbars/#notistack

IMHO, basically, if the css framework provides components that provide the same function, I think that using it will enhance the unity of the design.
But if you choose react-alert for reasons other than complexity of usage, that's fine for me :)

drolando:
@tacigar updated PR to use notistack. It even works better then react-alerts because it properly resizes to fit the text without me having to do any hack.

The only thing that confuses me a bit is that `alert` function. If I define it outside of `archiveClick` and pass it as "argument" in the `[]` at the end of the callback, npm complains with
```
The 'alert' function makes the dependencies of useCallback Hook (at line 188) change on every render. Move it inside the useCallback callback. Alternatively, wrap the 'alert' definition into its own useCallback() Hook
```
so I just moved it inside...

drolando:
@anuraaga Rebased on master and moved the notistack provider to App.jsx

zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
```diff
@@ -105,6 +106,81 @@ const TraceSummaryHeader = React.memo(({ traceSummary, rootSpanIndex }) => {
...
+        }
+
+        fetch(archivePostUrl, {
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify(json),
+        })
+          .then((response) => {
```
anuraaga: Instead of this nested fetching which makes it hard to read how about return the value of `fetch` and call `.then` above?

zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
```diff
@@ -105,6 +106,81 @@ const TraceSummaryHeader = React.memo(({ traceSummary, rootSpanIndex }) => {
...
+            } else {
+              notify(`Archive successful!`, 'success');
+            }
+          })
+          .catch((error) => {
+            notify(`Failed to archive the trace: ${error}`, 'error');
+          });
+      })
+      .catch(() => {
+        notify('Failed to fetch trace from backend', 'error');
```
anuraaga: If you use the same pattern as above of catching the error and notifying in the notication, you should be able to have a single `catch` instead of the nested

anuraaga (commented):
Thanks! Just one small point

zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
```diff
@@ -114,7 +114,20 @@ const TraceSummaryHeader = React.memo(({ traceSummary, rootSpanIndex }) => {
       ? config.archiveUrl.replace('{traceId}', traceSummary.traceId)
       : undefined;
 
+  const { enqueueSnackbar } = useSnackbar();
+
   const archiveClick = useCallback(() => {
+    const notify = function notify(message, variant) {
```
tacigar: `const notify = (message, variant) => {...}` is more natural?

drolando:
New popup UI using notistack:

image

image

zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
```diff
@@ -105,6 +106,80 @@ const TraceSummaryHeader = React.memo(({ traceSummary, rootSpanIndex }) => {
...
+        if (archiveUrl) {
+          notify(
+            `Archive successful! This trace is now accessible at ${archiveUrl}`,
+            'success',
+          );
+        } else {
+          notify(`Archive successful!`, 'success');
+        }
+      })
+      .catch(() => {
```
anuraaga, line 178: Go ahead and catch the error here and include it in the message (your previous nested catch did that)

anuraaga (approved):

zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
drolando: I removed that because the error is usually `Failed to fetch` which only confuses me and doesn't really add anything useful. Maybe there's a better way to get the real http error message though

anuraaga: Ah ok makes sense :)

zipkin-lens/src/components/App/App.jsx
```diff
@@ -39,29 +40,31 @@ const App = () => {
...
-                          path={['/traces/:traceId', '/traceViewer']}
-                          component={TracePage}
-                        />
-                      </Layout>
-                    </BrowserRouter>
-                  </I18nProvider>
-                </Provider>
-              )}
-            </UiConfigConsumer>
+            <SnackbarProvider>
```
codefromthecrypt, line 44: do we want to comment here why SnackbarProvider? it might not be obvious that this helps with things like archive

zipkin-lens/src/translations/es/messages.json
```diff
@@ -36,6 +36,7 @@
   "Trace ID": "ID de Traza",
   "Upload JSON": "Subir JSON",
   "View Logs": "Ver Logs",
+  "Archive Trace": "",
```
codefromthecrypt: @jeqo @jcchavezs traducelo por favor!

zipkin-lens/src/translations/zh-cn/messages.json
```diff
@@ -36,6 +36,7 @@
   "Trace ID": "",
   "Upload JSON": "",
   "View Logs": "",
+  "Archive Trace": "",
```
codefromthecrypt, line 39: @uckyk do you mind offering some chinese translation text here?

zipkin-server/README.md
```diff
@@ -177,13 +177,23 @@ queryLimit | zipkin.ui.query-limit | Default limit for Find Traces. Defaults to
...
+archiveUrl | zipkin.ui.archive-url | Url to a web application serving an archived trace, templated by '{traceId}'. e.g. https://longterm/zipkin/trace/{traceId}'. This is shown in a confirmation message after a trace is successfully POSTed to the `archivePostUrl`. Not required.
 dependency.lowErrorRate | zipkin.ui.dependency.low-error-rate | The rate of error calls on a dependency link that turns it yellow. Defaults to 0.5 (50%) set to >1 to disable.
 dependency.highErrorRate | zipkin.ui.dependency.high-error-rate | The rate of error calls on a dependency link that turns it red. Defaults to 0.75 (75%) set to >1 to disable.
 basePath | zipkin.ui.basepath | path prefix placed into the <base> tag in the UI HTML; useful when running behind a reverse proxy. Default "/zipkin"
 
 To map properties to environment variables, change them to upper-underscore case format. For
 example, if using docker you can set `ZIPKIN_UI_QUERY_LIMIT=100` to affect `$.queryLimit` in `/config.json`.
 
+### Trace archival
```
codefromthecrypt, line 189: looks great

codefromthecrypt (approved):
small things, looks legit!

zipkin-lens/src/zipkin/trace.js
```diff
@@ -291,7 +291,12 @@ function addLayoutDetails(
   }
 }
 
-export function detailedTraceSummary(root, logsUrl) {
+export function detailedTraceSummary(
+  root,
+  logsUrl,
+  archivePostUrl,
+  archiveUrl,
+) {
```
tacigar: Are these changes still needed?

drolando: I think so, as we still need to pass along the config values.

zipkin-lens/src/components/App/App.jsx
drolando: Well we decided to move it here so that it could be reused in other parts of the UI and to decouple it from the trace page and archiving

codefromthecrypt: not arguing the rationale, just saying it is invisible unless looking at the comments. maybe a single-line comment about what this helps with?

zipkin-lens/src/components/App/App.jsx
```diff
@@ -39,29 +40,32 @@ const App = () => {
...
-                          path={['/traces/:traceId', '/traceViewer']}
-                          component={TracePage}
-                        />
-                      </Layout>
-                    </BrowserRouter>
-                  </I18nProvider>
-                </Provider>
-              )}
-            </UiConfigConsumer>
+            // Snackbar is used to provide popup alerts to the user
```
codefromthecrypt: sweet

zipkin-lens/src/zipkin/trace.js
tacigar: All configurations seems possible to be retrieved from `useUiConfig` hooks.
like:   https://github.com/openzipkin/zipkin/blob/74759c913c80fa07871a232aa4dc77fe0b510938/zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx#L101

And these parameters(`archivePostUrl` and `archiveUrl`) does not seem to be used anywhere...

zipkin-lens/src/components/App/App.jsx
tacigar: Should be `{/* Snackbar is used to provide popup alerts to the user */}`

zipkin-lens/src/zipkin/trace.js
drolando: should I delete also logsUrl then?

tacigar: maybe...

anuraaga: Nice detective work finding this file - for reference in #2985 I didn't touch this file. I guess it's copy-pasted from somewhere and indeed these URL variables aren't used, so let's go ahead and delete

zipkin-lens/src/translations/es/messages.json
jeqo:
```suggestion
  "Archive Trace": "Archivar Traza",
```

jeqo: a la orden :)

jcchavezs: Sale!

anuraaga:
Thanks a lot @drolando!

tacigar:
Thank you!

PR #3021, opened 2020-03-11
Fix quote bug in DiscoverPage

tacigar:
While fixing lint errors, I escaped double quotes.
But it did not seem to work correctly in React.

Before: 
スクリーンショット 2020-03-11 18 50 57

After:
スクリーンショット 2020-03-11 18 50 44

### DIFF
--- a/zipkin-lens/src/components/DiscoverPage/DiscoverPage.jsx
+++ b/zipkin-lens/src/components/DiscoverPage/DiscoverPage.jsx
@@ -11,6 +11,7 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  */
+/* eslint-disable react/no-unescaped-entities */
 import { Trans } from '@lingui/macro';
 import PropTypes from 'prop-types';
 import React, { useEffect, useCallback } from 'react';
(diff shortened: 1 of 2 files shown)


anuraaga:
Can you check if there is a diff with `yarn run extract`? I think `Trans` won't work properly without inlining the text, and we would want to replace it with `{ i18n._(t`

tacigar:
Oh, sorry. you are right. Trans does not work...
I disabled eslint instead of it.

zipkin-lens/src/components/DiscoverPage/DiscoverPage.jsx
```diff
@@ -11,6 +11,7 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  */
+/* eslint-disable react/no-unescaped-entities */
```
anuraaga: We can probably disable this in the config since all our strings will be in i18n blocks in practice.

anuraaga (approved):

zipkin-lens/src/components/DiscoverPage/DiscoverPage.jsx
tacigar: I see!

codefromthecrypt (approved):

codefromthecrypt:
thanks again @tacigar and @anuraaga

PR #3022, opened 2020-03-11
Set default configs for UiConfig

tacigar:
If the item is not set, a default value should be set.
zipkin-classic did so: https://github.com/openzipkin-attic/zipkin-classic/blob/405aaa140921753e7ca7aa34e26bd8859d7a52ff/js/config.js#L3-L25


codefromthecrypt (approved):

jcchavezs:
Is suggestLens a thing in lens?

tacigar:
Oops! Thank you! @jcchavezs

codefromthecrypt:
Thanks @tacigar and @jcchavezs!

PR #3031, opened 2020-03-13
Apply prettier to zipkin-lens for automatic whitespace formatting.

anuraaga:
Note, this is just a strawman's proposal. Formatting preferences are not set-in-stone and I'm fine with just closing this PR if others don't like it. But figured it's worth giving one think on it.

I would say that automatic formatting using prettier is very common in the JS world, unlike say using Google Java Format for Java. As an example, our two most important dependencies, React and Material UI, both use it. So I think it can be said that it's idiomatic to use it - it's nice that's it's also "free" in terms of tooling, since it just integrates with our existing eslint. My personal opinion is not liking to think about formatting if a tool can do it for me, but interested in what others think on this.


codefromthecrypt (approved):
I feel less strongly about this than Java, so up to you and @tacigar

tacigar:
Awesome!
I agree with using prettier.
But I'm wondering if it is possible to exactly match the rules of prettier and eslint.
Does not eslint show errors for code formatted by prettier?

anuraaga:
Nope - we extend the eslint config with prettier-specific eslint configs which reconfigure other rules to not conflict. I've used this in other projects, and here too, I've never seen them fighting each other.

tacigar (approved):
Got it! LGTM~

PR #3036, opened 2020-03-17
Remove API_BASE environment variable

tacigar:
react-script requires `REACT_APP_` prefix for environment variables.

See https://create-react-app.dev/docs/adding-custom-environment-variables/

### DIFF
--- a/zipkin-lens/src/constants/api.js
+++ b/zipkin-lens/src/constants/api.js
@@ -11,7 +11,7 @@
  * or implied. See the License for the specific language governing permissions and limitations under
  * the License.
  */
-const { API_BASE } = process.env;
+const { REACT_APP_API_BASE } = process.env;
 
 const extractBasePath = () => {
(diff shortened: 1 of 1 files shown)


anuraaga (commented):
Thanks! By the way, with `proxy` set up `API_BASE` isn't required for development anymore. Should we go ahead and just remove this variable?

tacigar:
I see...
I want to set `proxy` with environment variables, but couldn't find an easy way.
Do you have any ideas?

anuraaga (approved):
Don't think we would be able to use `proxy` and environment variables. The question is how common it is to use anything other than the default since then for the rare cases it's not so bad to edit `package.json`. But if you think it's common enough I think it's ok to keep the variable.

tacigar:
I've been developing with the `API_BASE` environment variable so far, but this may not be common.
OK, I'll remove API_BASE.

anuraaga:
Thanks!

PR #3037, opened 2020-03-17
Replace all instances of {traceId} in logs URL instead of just first.

### LINKED ISSUE #3033: Zipkin UI logsUrl does not support multiple occurrence of {traceId}
## Describe the Bug
I am trying to link Kibana in Zipkin UI (Lens). The button is shown correctly. `{traceId}` are not replaced in the URL.

I tried with simple URL with multiple occurence of `{traceId}`. It is able to replace the first occurence of {traceId}.

## Steps to Reproduce
`zipkin.ui.logs-url=https://kibana.company.com/app/kibana#/discover?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-7d,to:now))&_a=(columns:!(_source),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'ca487de6-0028-4e74-b14b-65eddbcf1f5a',key:Trace_ID,negate:!f,params:(query:'{traceId}'),type:phrase,value:'{traceId}'),query:(match:(Trace_ID:(query:'{traceId}',type:phrase))))),index:'ca487de6-0028-4e74-b14b-65eddbcf1f5a',interval:auto,query:(language:kuery,query:''),sort:!('@timestamp',desc))`

## Expected Behaviour
Zipkin UI should able to replace multiple occurence of `{traceId}`.

anuraaga:
I originally thought this is a bug in lens but realized classic only replaced one too

https://github.com/openzipkin-attic/zipkin-classic/blob/bad04931779d6486b4fb7efd64c1c90965f79fcd/js/component_data/trace.js#L9

But it seems simple enough to just replace them all if it helps.

Fixes #3033

### DIFF
--- a/zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
+++ b/zipkin-lens/src/components/TracePage/TraceSummaryHeader.jsx
@@ -102,7 +102,7 @@ const TraceSummaryHeader = React.memo(({ traceSummary, rootSpanIndex }) => {
 
   const logsUrl =
     config.logsUrl && traceSummary
-      ? config.logsUrl.replace('{traceId}', traceSummary.traceId)
+      ? config.logsUrl.replace(/{traceId}/g, traceSummary.traceId)
       : undefined;
 
   const handleSaveButtonClick = useCallback(() => {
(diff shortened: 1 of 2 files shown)


tacigar (approved):
LGTM!

PR #3038, opened 2020-03-17
Default unsupported locales to English.

### LINKED ISSUE #3034: Pagination on the result page broken
Hello,

I just updated to 2.20.2 and when doing a search, I get the following on the top instead of the number of results/pagination:

`{0, plural, one {# Result} other {# Results}}`

anuraaga:
Even with good hints on what causes the issue, I still can't reproduce locally. Have tried changing browser language and hacking the code to return an unsupported locale and things always work fine for me.

But anyways, by ensuring the app locale is only ever a supported locale, I'm guessing any issues will go away.

Fixes #3034 hopefully

### DIFF
--- a/zipkin-lens/src/util/locale.ts
+++ b/zipkin-lens/src/util/locale.ts
@@ -12,7 +12,7 @@
  * the License.
  */
 
-import { Catalogs, setupI18n } from '@lingui/core';
+import { setupI18n } from '@lingui/core';
 
 import enMessages from '../translations/en/messages';
 import esMessages from '../translations/es/messages';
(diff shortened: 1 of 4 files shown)


zipkin-lens/src/util/locale.ts
```diff
@@ -12,25 +12,49 @@
...
+  if (hyphenIndex >= 0) {
+    const stripped = browserLanguage.substring(0, hyphenIndex);
+    if (catalogs.hasOwnProperty(stripped)) {
+      return stripped;
+    }
   }
 
-  return browserLanguage;
+  // Default to English if we don't support the current browser language.
+  return 'en';
```
codefromthecrypt, line 50: guessing this is the key

codefromthecrypt (approved):

PR #3040, opened 2020-03-23
Change Download JSON button to be a normal link with href.

### LINKED ISSUE #3039: lens: hovering over [download json] button could show REST url to fetch trace
**Feature:**
When hovering over the download json button, it seems intuitive to show the full link to the trace (ie http://myhost/api/v2/trace/4f711e17e7e2290a) , in the browser status bar just as if you would hover over a normal href. 
 
**Rational**
I could not remember the location of the swagger api definition at https://zipkin.io/zipkin-api/ , i wanted to fetch a trace using curl | jq . So i looked at the download button hoping to find the REST url there.

anuraaga:
No need to create a magic anchor to do the download instead of making the original button an anchor.

Fixes #3039

### DIFF
--- a/zipkin-lens/README.md
+++ b/zipkin-lens/README.md
@@ -19,6 +19,9 @@ Open [http://localhost:3000](http://localhost:3000) to view it in the browser.
 The page will reload if you make edits.<br />
 You will also see any lint errors in the console.
 
+By default, API requests are proxied to `http://localhost:9411`. You can change this target using
+the `API_BASE` environment variable, e.g., `API_BASE=http://tracing.company.com npm start`.
+
 ### `npm test`
 
 Launches the test runner in the interactive watch mode.<br />
--- a/zipkin-lens/package.json
+++ b/zipkin-lens/package.json
@@ -51,6 +51,7 @@
     "eslint-plugin-react-hooks": "^1.6.1",
     "fetch-mock": "^7.3.0",
     "history": "^4.10.1",
+    "http-proxy-middleware": "^1.0.3",
     "lodash": "^4.17.14",
     "moment": "^2.24.0",
     "mvy": "0.2.1",
@@ -99,6 +100,5 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
-  },
-  "proxy": "http://localhost:9411"
+  }
 }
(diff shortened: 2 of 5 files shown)


tacigar:
hmm...
I cannot download JSON in chrome (but I can in firefox)...

anuraaga:
Ah I only checked on Firefox, let me check Chrome too

anuraaga:
Just tried on Chrome and it worked ok - I do need to fix it to work on dev server using https://create-react-app.dev/docs/proxying-api-requests-in-development/#configuring-the-proxy-manually it only really works on normal server. Just to check, did you try using `:9411` or `:3000`?

tacigar:
I tried using `:3000`.
`:9411` works ok.
The same problem seems to have existed before...

anuraaga:
Thanks for confirming, yeah it broke in the dev server with the react-scripts migrations so will fix!

anuraaga:
Fixed the proxy - went ahead and restored our API_BASE variable at the same time since it's easy

jorgheymans:
maybe this is a good moment to have a deep and meaningful discussion about the placement of the button as well :-P

image

I'm thinking that it could make sense to have the download button next to the upload button instead, and make it an icon  only (with tooltip) so they're nicely uniform ? @tacigar also

anuraaga:
I've been thinking the same thing - moving the button up seems good to me.

anuraaga:
One point we'll need to think of is the other buttons. If we move download, we'd presumably move view logs and archive trace too. I'd say the icon for view logs is pretty bad - I desperately searched for something not terrible on font awesome. So icon-only might not be so nice. If we can get an icon designed that looks clearer that'd be very cool :D

jorgheymans:
This may turn into a slightly bigger UI concept refactoring then, separate issue ? Thing is, there are 'global' widgets like the search trace and upload trace, then you have the trace 'local' stuff like download, view logs and archive. Is it ok mixing these into one "toolbar"?

anuraaga:
Yup true that this would make the global toolbar dynamic - lots of code change as well as a big design change, so good to think about it separately.

codefromthecrypt:
as we have some glitches in our last docker images, plus we've that favorite trace thing in master.. if this is good and mergable, let's do it and cut a minor?

PR #3053, opened 2020-04-03
this parameter was never documented, fixes #718

### LINKED ISSUE #718: Allow caching of getServices in the UI

jorgheymans:
Talking about `QUERY_NAMES_MAX_AGE`, not sure on the description though


codefromthecrypt (approved):
thanks for the help!

PR #3062, opened 2020-04-13
Fix #3052

### LINKED ISSUE #3052: editing date range manually in dependencies page broken
put the cursor in one of the date range fields, pressing pretty much any key makes it barf.

dependency-bug

tacigar:
Fix #3052

As pointed out #3052 , when something is put to the input of DateTimePicker, it will often barf.
So disable input and force users to enter the date time using a calendar dialog.
Probably this is better UX.

### DIFF
--- a/zipkin-lens/src/components/DependenciesPage/DependenciesPageHeader.jsx
+++ b/zipkin-lens/src/components/DependenciesPage/DependenciesPageHeader.jsx
@@ -90,6 +90,21 @@ const DependenciesPageHeader = React.memo(
     const classes = useStyles();
     const { i18n } = useLingui();
 
+    const dateTimePickerCommonProps = React.useMemo(
+      () => ({
+        inputVariant: 'outlined',
+        className: classes.dateTimePicker,
+        InputProps: {
+          // See: https://github.com/openzipkin/zipkin/issues/3052
+          // DateTimePicker's input is not very useful, so disable it and
+          // let users use a calendar dialog instead.
+          disabled: true,
+          classes: { input: classes.dateTimePickerInput },
+        },
+      }),
+      [classes.dateTimePicker, classes.dateTimePickerInput],
+    );
+
     return (
       <Box className={classes.root}>
         <Box className={classes.upperBox}>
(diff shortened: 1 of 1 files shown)


zipkin-lens/src/components/DependenciesPage/DependenciesPageHeader.jsx
```diff
@@ -105,7 +108,10 @@ const DependenciesPageHeader = React.memo(
             value={startTime}
             onChange={onStartTimeChange}
             className={classes.dateTimePicker}
-            InputProps={{ classes: { input: classes.dateTimePickerInput } }}
+            InputProps={{
+              disabled: true,
```
codefromthecrypt: same file, so make a constant and use in both places? then you have only one place to add comment why

codefromthecrypt (approved):

zipkin-lens/src/components/DependenciesPage/DependenciesPageHeader.jsx
```diff
@@ -87,6 +90,21 @@ const DependenciesPageHeader = React.memo(
     const classes = useStyles();
     const { i18n } = useLingui();
 
+    const dateTimePickerCommonProps = React.useMemo(
+      () => ({
+        inputVariant: 'outlined',
```
anuraaga: This object is very simple and won't force rerender, so better not to memo.

zipkin-lens/src/components/DependenciesPage/DependenciesPageHeader.jsx
```diff
@@ -87,6 +90,21 @@ const DependenciesPageHeader = React.memo(
...
+        className: classes.dateTimePicker,
+        InputProps: {
+          // See: https://github.com/openzipkin/zipkin/issues/3052
+          // DateTimePicker's input is not very useful, so disable it and
+          // let users use a calendar dialog instead.
+          disabled: true,
```
anuraaga: I think this is fine for now but I'll probably look into whether it's possible to make it not barf at some point. On desktop, I usually get frustrated when a date input doesn't let me type in the date since it's usually way faster than scrolling through the calendar.

tacigar:
I tried `format` prop instead of disabling input.
I think it's pretty easy to use.

datetimepicker

anuraaga (approved):
Sweet thanks!

PR #3063, opened 2020-04-14
Allows opting out via ES_ENSURE_TEMPLATES=false

### LINKED ISSUE #3025: Disable index template creation at startup
The ES index templates are created during server startup. Sites like Netflix have customized the template and prefer managing the templates outside the server runtime.

We would like if the template creation can be controlled via a property.

@adriancole

codefromthecrypt:
Users can now disable Elasticsearch index template creation at startup
via `ES_ENSURE_TEMPLATES=false`. This breaks the warranty of the install
and any upgrades when they do this.

Fixes #3025


codefromthecrypt:
@uckyk can you try?

```bash
$ curl -sSL https://jitpack.io/com/github/openzipkin/zipkin/zipkin-server/ES_ENSURE_TEMPLATES-SNAPSHOT/zipkin-server-ES_ENSURE_TEMPLATES-SNAPSHOT-exec.jar > zipkin.jar
```

zipkin-server/src/test/java/zipkin2/server/internal/elasticsearch/ITElasticsearchDynamicCredentials.java
```diff
@@ -80,9 +79,7 @@
     storage.close();
   }
 
-  @Test void healthcheck_usesDynamicCredentialsAndTls() throws IOException {
-    pretendIndexTemplatesExist(storage);
```
anuraaga, line 84: Test cleanup is awesome!

anuraaga (approved):

uckyk:
emm, I modified the source code myself, and the get permission has been added. I commented out  "the put method". cant try now, sorry...

PR #3065, opened 2020-04-15
Adds parameter to disable the dependencies screen

### LINKED ISSUE #3056: Allow config to disable the Lens UI dependencies page
The dependencies page only works in a production (Cassandra / ES) deployment when accompanied with a specific aggregation job. Since not all customers deploy such a job or even need it, it would be nice if we can disable the Lens UI dependencies page through Zipkin configuration.

codefromthecrypt:
zipkin.ui.dependency.enabled now controls whether or not the
Dependencies screen is visible. Like all other UI props, this can also
be set as an env variable.

```bash
ZIPKIN_UI_DEPENDENCY_ENABLED=false
```

The naming choice was made to match other UI config for Dependencies
vs creating a different convention just for this.

Fixes #3056

### DIFF
--- a/zipkin-lens/src/components/App/Layout.test.jsx
+++ b/zipkin-lens/src/components/App/Layout.test.jsx
@@ -31,7 +31,7 @@ describe('<Layout />', () => {
 
   it('does render help link when defined', () => {
     // children is required so avoid warning by passing dummy children
-    const { getByTitle } = render(
+    const { queryByTitle } = render(
       <Layout>
         <span>Test</span>
         <span>Test</span>
@@ -42,12 +42,12 @@ describe('<Layout />', () => {
         },
       },
     );
-    const helpLink = getByTitle('Support');
+    const helpLink = queryByTitle('Support');
     expect(helpLink).toBeInTheDocument();
     expect(helpLink.href).toEqual('https://gitter.im/openzipkin/zipkin');
   });
 
-  it('does render Dependencies with default config', () => {
+  it('does render Dependencies Page with default config', () => {
     // children is required so avoid warning by passing dummy children
     const { queryByTitle } = render(
       <Layout>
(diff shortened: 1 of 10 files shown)


codefromthecrypt:
I still need to check this actually works, just raising for early feedback @openzipkin/ui

codefromthecrypt:
works 
Screenshot 2020-04-15 at 10 12 41 AM

```bash
$ ZIPKIN_UI_DEPENDENCY_ENABLED=false java -jar ./zipkin-server/target/zipkin-server-*slim.jar
```

tacigar (approved):

zipkin-lens/src/components/App/Layout.test.jsx
```diff
@@ -42,8 +42,37 @@ describe('<Layout />', () => {
         },
       },
     );
-    const helpLink = getByTitle('Support');
+    const helpLink = queryByTitle('Support');
```
anuraaga, line 45: It might be a deficiency of the assertions framework but I found `getByTitle` to give a better error message when the element isn't found.

codefromthecrypt: thanks will change in the next PR if it works. just noticed things seemed arbitrarily different

codefromthecrypt: 74 |       },
      75 |     );
    > 76 |     expect(getByTitle('Dependencies Page')).not.toBeInTheDocument();
         |            ^
      77 |   });
      78 | });
      79 | 

      at Object.getElementError (node_modules/@testing-library/dom/dist/config.js:34:12)
      at args (node_modules/@testing-library/dom/dist/query-helpers.js:71:38)
      at getByTitle (node_modules/@testing-library/dom/dist/query-helpers.js:54:17)
      at Object.it (src/components/App/Layout.test.jsx:76:12)

codefromthecrypt: so we can either use them differently and comment why, or use queryByTitle consistently and not comment anything, or do some error handling?

codefromthecrypt: I'll leave this one to you. I don't really like unintuitive bombs in code, where we do one thing and then secret knowledge why we have to do it differently somewhere else.

anuraaga: Agree it's not intuitive and I don't like it, will try to think of a good way, may just a helper method we define, or a monkey patch. For reference, the big difference is `getBy` prints out a dump of the DOM if an element isn't found while `queryBy.toBeInDocument` just says something like `undefined not in document`. Though it does seem like a bug in `toBeInDocument` to me.

codefromthecrypt: yeah I didn't mean to shoot the messenger either, I know javascript in particular is rife with tribal knowledge as a way of doing things.. something I've never acclimated to

PR #3067, opened 2020-04-18
Enable dependency by default

tacigar:


codefromthecrypt:
oh no! does this mean what I fear it means?

or is this only a problem in dev mode...

tacigar:
Only in dev mode!
Maybe no problem in production mode :)

codefromthecrypt:
whew

PR #3071, opened 2020-04-20
Update web dependencies.

anuraaga:
Notably upgrades past some major version boundaries.

- `jest-environment-jsdom-sixteen` is to allow updating jsdom before create-react-app does (they're not sure about the change of default since it could break existing tests)
- prettier 2.0 changed some default settings and removed the ones we had that became redundant
- Found several libraries in package.json that aren't referenced in code. @tacigar Can you confirm the removals are OK? In particular, I found `rc-slider` and `react-table` in the CSS but not in JS so pretty sure they're not used but please confirm.

Doesn't update
- date-io, pinned by material-ui/pickers
- @types/jest, jest version not upgraded by create-react-app yet


tacigar:
Yes, no problem!
Thank you!

tacigar (approved):
LGTM!

codefromthecrypt:
thanks a lot for taking the time on this. I know managing deps is usually
far trickier than numerical increments!

PR #3077, opened 2020-05-08
add TLS note

jorgheymans:


zipkin-server/README.md
```diff
@@ -579,6 +579,16 @@ of known open source libraries on 128-bit trace identifiers.
 
 See `zipkin2.storage.StorageComponent.Builder` for even more details!
 
+## TLS/SSL
```
jorgheymans, line 582: added it at the end of the README since i don't think this is a prominent feature that everyone is looking for

zipkin-server/README.md
```diff
@@ -579,6 +579,16 @@ of known open source libraries on 128-bit trace identifiers.
...
+Zipkin-server can be made to run with TLS if needed:
+
+```bash
+# assuming you generate the key like this
+keytool -genkeypair -alias mysite -keyalg RSA -keysize 2048 -storetype PKCS12 -keystore zipkin.p12 -validity 3650
+
+java -jar zipkin.jar --armeria.ssl.key-store=zipkin.p12 --armeria.ssl.key-store-type=PKCS12 --armeria.ssl.key-store-password=123123 --armeria.ssl.key-alias=mysite  --armeria.ssl.enabled=true --armeria.ports[0].port=9411 --armeria.ports[0].protocols[0]=https
```
jorgheymans, line 589: @anuraaga i tried to find a pointer in the armeria documentation that explains these properties, but the closest i could find was here https://line.github.io/armeria/advanced-dropwizard-integration.html?highlight=ssl

codefromthecrypt: I think that they are conventional to spring not explicit in docs.. https://github.com/line/armeria/blob/70f4f60e6becc0322c0d0fd86a0b6ff0290800cf/spring/boot-autoconfigure/src/main/java/com/linecorp/armeria/spring/Ssl.java cc @trustin

codefromthecrypt (approved):

PR #3080, opened 2020-05-14
add trace.trim()

uckyk:
when use "@get("/api/v2/trace/{traceId}")"
here is the code traceId = Span.normalizeTraceId(traceId);
if trace with space , it will doesnt work.
1、 in SPAN.JAVA ,we add trace.trim();
2、in ZipkinQueryApiV2, we add this “ traceId = traceId != null ? traceId.trim() :null; ”

i choiced the first one

### DIFF
--- a/zipkin/src/main/java/zipkin2/Span.java
+++ b/zipkin/src/main/java/zipkin2/Span.java
@@ -641,7 +641,7 @@ public final class Span implements Serializable { // for Spark and Flink jobs
    */
   public static String normalizeTraceId(String traceId) {
     if (traceId == null) throw new NullPointerException("traceId == null");
-    int length = traceId.length();
+    int length = traceId.trim().length();
     if (length == 0) throw new IllegalArgumentException("traceId is empty");
     if (length > 32) throw new IllegalArgumentException("traceId.length > 32");
     int zeros = validateHexAndReturnZeroPrefix(traceId);


zipkin/src/main/java/zipkin2/Span.java
```diff
@@ -641,7 +641,7 @@ public Span build() {
    */
   public static String normalizeTraceId(String traceId) {
     if (traceId == null) throw new NullPointerException("traceId == null");
-    int length = traceId.length();
```
codefromthecrypt, line 644: trim has performance impact and this is the lowest level library... seems the trim if occurs should happen in the UI or Query controller instead..

codefromthecrypt: so I suggest if this is for query api then maybe add the trim there and in one of the api tests add a test case with the space?

uckyk: I think it's okay to add trim when querying, this performance is not affected at all

uckyk: 2、in ZipkinQueryApiV2, we add this “ traceId = traceId != null ? traceId.trim() :null; ”

codefromthecrypt: yeah if even possible for that to be null in ZipkinQueryApiV2 which shouldn't be possible. trimming is fine there vs here which is core library used by many instrumentation

uckyk: yes,  ill  pull request  again

PR #3081, opened 2020-05-15
Friendly advice about trim

uckyk:
Friendly advice
when use "@get("/api/v2/trace/{traceId}")"
here is the code traceId = Span.normalizeTraceId(traceId);
if trace with space , it will doesnt work.

[ in ZipkinQueryApiV2, we add this “ traceId = traceId != null ? traceId.trim() :null; ” ]

I think it's okay to add trim when querying, this performance is not affected at all.
Maybe this is not the best way to deal with it, you can also improve it according to the friendliness I mentioned in the follow-up, and you do n’t have to use mine completely.
I also mentioned that adding a common to SPAN may cause low performance problems, so set the trim here

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java
@@ -151,7 +151,7 @@ public class ZipkinQueryApiV2 {
   @Get("/api/v2/trace/{traceId}")
   @Blocking
   public AggregatedHttpResponse getTrace(@Param("traceId") String traceId) throws IOException {
-    traceId = traceId != null ? traceId.trim() :null;
+    traceId = traceId != null ? traceId.trim() : null;
     traceId = Span.normalizeTraceId(traceId);
     List<Span> trace = storage.traces().getTrace(traceId).execute();
     if (trace.isEmpty()) {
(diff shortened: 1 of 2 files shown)


zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java
```diff
@@ -151,6 +151,7 @@ public AggregatedHttpResponse getTraces(
   @Get("/api/v2/trace/{traceId}")
   @Blocking
   public AggregatedHttpResponse getTrace(@Param("traceId") String traceId) throws IOException {
+    traceId = traceId != null ? traceId.trim() :null;
```
codefromthecrypt:
```suggestion
    traceId = traceId != null ? traceId.trim() : null;
```

codefromthecrypt:
looks good. can you add a test to ITZipkinServer to make sure this isn't broken later?

```java
  @Test public void getTrace_spaceAfterTraceId() throws Exception {
    storage.accept(TRACE).execute();

    Response response = get("/api/v2/trace/" + TRACE.get(0).traceId() + " ");
    assertThat(response.isSuccessful()).isTrue();

    assertThat(response.body().bytes())
      .containsExactly(SpanBytesEncoder.JSON_V2.encodeList(TRACE));
  }
```

codefromthecrypt:
in fact best is to use the same spacing that triggered this.. not sure if your HTTP command had a space before or after the trace ID.. I assume after

uckyk:
> in fact best is to use the same spacing that triggered this.. not sure if your HTTP command had a space before or after the trace ID.. I assume after

done

codefromthecrypt:
thanks! will just wait for build green. no more feedback

PR #3083, opened 2020-05-17
Bump Kafka versions to 2.5.0

jeqo:
Btw, if https://github.com/openzipkin-contrib/zipkin-storage-kafka/pull/70 is successful, we could evaluate replacing kafka-junit with Testcontainers :)

### DIFF
--- a/docker/collector/kafka/install.sh
+++ b/docker/collector/kafka/install.sh
@@ -20,10 +20,6 @@ apk add --update --no-cache jq curl
 
 APACHE_MIRROR=$(curl --stderr /dev/null https://www.apache.org/dyn/closer.cgi\?as_json\=1 | jq -r '.preferred')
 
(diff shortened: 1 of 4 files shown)


docker/collector/kafka/install.sh
```diff
@@ -20,10 +20,6 @@ apk add --update --no-cache jq curl
 
 APACHE_MIRROR=$(curl --stderr /dev/null https://www.apache.org/dyn/closer.cgi\?as_json\=1 | jq -r '.preferred')
 
-curl -sSL $APACHE_MIRROR/zookeeper/zookeeper-$ZOOKEEPER_VERSION/apache-zookeeper-$ZOOKEEPER_VERSION.tar.gz | tar xz
```
jeqo, line 23: AFAIK, Zookeeper binaries are not required as they come as part of Kafka distribution (Unless until kip-500 ;) )

codefromthecrypt (approved):
thanks. I like the red

Issue #1646, opened 2017-07-05
Add unit tests to ensure component names aren't crap in health check


codefromthecrypt:
Right now, elasticsearch shows as `AutoValue_ElasticsearchHttpStorage` in the health check. This should be the simple name w/o AutoValue_. ZipkinHealthIndicator should scrub `AutoValue_` from the simple class name.

jorgheymans:
This is actually covered by specific tests in [CassandraStorageTest](https://github.com/openzipkin/zipkin/blob/f238246b30ffae0e52b8323e3547a0d97b9aef0c/zipkin-storage/cassandra/src/test/java/zipkin2/storage/cassandra/CassandraStorageTest.java#L39) , [CassandraStorageTest.v1](https://github.com/openzipkin/zipkin/blob/f238246b30ffae0e52b8323e3547a0d97b9aef0c/zipkin-storage/cassandra-v1/src/test/java/zipkin2/storage/cassandra/v1/CassandraStorageTest.java#L28), [ElasticsearchStorageTest](https://github.com/openzipkin/zipkin/blob/10c478ba049696e9ef32e8400a3d7209e3f31c5a/zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/ElasticsearchStorageTest.java#L196), and InMemoryStorageTest was aligned to this in https://github.com/openzipkin/zipkin/pull/3085 which will close this issue once merged.

codefromthecrypt:
oh wow github is smarter than I thought about auto closing!

PR #3085, opened 2020-05-19
completes (and thus fixes https://github.com/openzipkin/zipkin/issues/1646 )

jorgheymans:


codefromthecrypt (approved):
thanks!

codefromthecrypt:
protip: a single line in the description or commit "fixes #1646" would magically close it

PR #3093, opened 2020-05-26
Stops looking for local RabbitMQ connection

### LINKED ISSUE #1748: RabbitMQ integration test fails on Travis
We currently have a rabbitmq service added to travis config because the test-containers config fails with connect errors.

codefromthecrypt:
This improves the Docker based RabbitMQ tests in the same manner as
https://github.com/openzipkin/zipkin-reporter-java/pull/176

This notably stops looking for a local RabbitMQ install. It also reduces
flakes by using the CLI and not sharing static collector resources in
instance-level tests.

Fixes #1748

### DIFF
--- a/zipkin-collector/rabbitmq/src/test/java/zipkin2/collector/rabbitmq/ITRabbitMQCollector.java
+++ b/zipkin-collector/rabbitmq/src/test/java/zipkin2/collector/rabbitmq/ITRabbitMQCollector.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2019 The OpenZipkin Authors
+ * Copyright 2015-2020 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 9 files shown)


zipkin-collector/rabbitmq/src/test/java/zipkin2/collector/rabbitmq/ITRabbitMQCollector.java
```diff
@@ -15,151 +15,120 @@
...
 public class ITRabbitMQCollector {
   List<Span> spans = Arrays.asList(LOTS_OF_SPANS[0], LOTS_OF_SPANS[1]);
 
-  @ClassRule
-  public static RabbitMQCollectorRule rabbit = new RabbitMQCollectorRule("rabbitmq:3.7-alpine");
+  @ClassRule public static RabbitMQCollectorRule rabbit = new RabbitMQCollectorRule();
 
-  @After
-  public void clear() {
-    rabbit.metrics.clear();
```
codefromthecrypt, line 48: a lot of flakes are possible when static java connections and such are used for test methods.. It is very true that rabbit itself should be instantiated once per (suite if possible) as it is slow to start.

zipkin-collector/rabbitmq/src/test/java/zipkin2/collector/rabbitmq/ITRabbitMQCollector.java
```diff
@@ -15,151 +15,120 @@
...
-    rabbit.publish(malformed2);
-    rabbit.publish(THRIFT.encodeList(spans));
+    publish(THRIFT.encodeList(spans));
+    publish(new byte[0]);
+    publish(malformed1);
+    publish(malformed2);
+    publish(THRIFT.encodeList(spans));
 
-    Thread.sleep(1000);
+    Thread.sleep(200L);
```
codefromthecrypt, line 88: As far as I know rabbit is push not poll, so any delay here would be how long it can decode the message and synchronously add to our in-memory storage. 200L is a bit excessive but less slow than 1s

jorgheymans (approved):
Nice cleanup, a lot more involved than what i had in mind !

codefromthecrypt:
thanks for keeping the project eyes focused @jorgheymans !

Issue #1501, opened 2017-01-22
Instructions for building on windows


ahxm:
hello,
  I can run zipkin-server locally,but I cannot access http://localhost:9411/ ,the response is :There was an unexpected error (type=Internal Server Error, status=500).class path resource [zipkin-ui/index.html] cannot be resolved to URL because it does not exist。I don't know  how to start zipkin-ui at windows.

codefromthecrypt:
@ahxm do you have any errors from the console? I assume you are running `java jar zipkin-server.jar` as noted in the README?

@smartthought @gena01 @Horusiath @theoptimiste I think each of you have run zipkin locally. Did you have to do anything special in order to do so? We are sorely missing a windows section on readme.

@eirslett @abesto maybe we should setup a CI worker to use free windows VM to test zipkin can be started on windows (if not also that it can be built)

ahxm:
I think it is really necessary that detailed steps  to run zipkin in windows environment.
I am not running  java jar zipkin-server.jar 
I use IntelliJ IDEA to run ZipkinServer.main(). 
localhost:9411/config.json is OK
localhost:9411/health is OK
localhost:9411/info is OK
localhost:9411/metrics is OK
localhost:9411/ is error
there is the errors from the console:
2017-01-23 11:05:32.000 ERROR 4712 --- [io-9411-exec-10] o.a.c.c.C.[.[.[/].[dispatcherServlet]    : Servlet.service() for servlet dispatcherServlet threw exception

java.io.FileNotFoundException: class path resource [zipkin-ui/index.html] cannot be resolved to URL because it does not exist
	at org.springframework.core.io.ClassPathResource.getURL(ClassPathResource.java:187) ~[spring-core-4.3.5.RELEASE.jar:4.3.5.RELEASE]
	at org.springframework.core.io.AbstractFileResolvingResource.contentLength(AbstractFileResolvingResource.java:151) ~[spring-core-4.3.5.RELEASE.jar:4.3.5.RELEASE]
	at org.springframework.http.converter.ResourceHttpMessageConverter.getContentLength(ResourceHttpMessageConverter.java:99) ~[spring-web-4.3.5.RELEASE.jar:4.3.5.RELEASE]
	at org.springframework.http.converter.ResourceHttpMessageConverter.getContentLength(ResourceHttpMessageConverter.java:50) ~[spring-web-4.3.5.RELEASE.jar:4.3.5.RELEASE]
[...]

PR #3094, opened 2020-05-26
Tests zipkin-server/README.md instructions for building server on each pull request

codefromthecrypt:
We still get questions about windows not building properly. This
uses GitHub actions to help ensure we don't break by accident, or at least
so we automatically know if at least one windows environment works.

Thanks to curiostack for the reference! https://github.com/curioswitch/curiostack/blob/master/.github/workflows/continuous-build.yml

See #1501

### DIFF
--- a/.github/workflows/continuous-build.yml
+++ b/.github/workflows/continuous-build.yml
@@ -31,8 +31,8 @@ jobs:
         with:
          path: .mvn/wrapper/maven-wrapper.jar
          key: mvn-wrapper
-      - name: Execute Build Server
-        run: ./mvnw -DskipTests --also-make -pl zipkin-server clean install
+      - name: Execute Server Build
+        run: ./mvnw --batch-mode -DskipTests --also-make -pl zipkin-server clean install
         shell: bash
         env:
           CI: true
--- a/.github/workflows/continuous-build.yml
+++ b/.github/workflows/continuous-build.yml
@@ -20,17 +20,17 @@ jobs:
         uses: actions/cache@v1
         with:
           path: ~/.npm
-          key: npm-packages
+          key: npm-packages-${{ runner.os }}-${{ hashFiles('zipkin-lens/package-lock.json') }}
       - name: Cache Maven Modules
         uses: actions/cache@v1
         with:
           path: ~/.m2/repository
-          key: m2-repository
+          key: m2-repository-${{ hashFiles('**/pom.xml') }}
       - name: Cache Maven Wrapper
         uses: actions/cache@v1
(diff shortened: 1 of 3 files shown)


codefromthecrypt:
yaml somewhat from here https://github.com/openzipkin-attic/zipkin-azure/blob/master/appveyor.yml

anuraaga (approved):
Appveyor's cool too, but if I was starting a new Windows build I'd usually go with Actions now since can get a mac build easily too. Here's a reference one in case you feel like trying

https://github.com/curioswitch/curiostack/blob/master/.github/workflows/continuous-build.yml

codefromthecrypt:
hmm I like the idea of that better @anuraaga as I don't have to create a fake appveyor account for the org, share password etc..

codefromthecrypt:
ok PTAL! https://github.com/openzipkin/zipkin/actions/runs/115447278

.github/workflows/continuous-build.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: m2-repository
+      - name: Cache Maven Wrapper
+        uses: actions/cache@v1
+        with:
+         path: .mvn/wrapper/maven-wrapper.jar
+         key: mvn-wrapper
+      - name: Execute Build Server
```
anuraaga: Nit: either

- Execute Build: Server
- Execute Server Build

Was wondering what the cool build server is :)

.github/workflows/continuous-build.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: ./mvnw -DskipTests --also-make -pl zipkin-server clean install
```
anuraaga: Think want `--batch-mode` like in publish.sh to reduce tons of log spam

anuraaga (approved):

.github/workflows/continuous-build.yml
codefromthecrypt: heh yeah this is better thanks!

.github/workflows/continuous-build.yml
codefromthecrypt: agree. also fixed readme similarly

codefromthecrypt:
added hash keys, where only npm is conditional on OS as maven downloads should be well.. java write once run everywhere right?

anuraaga (approved):
Sweet

codefromthecrypt:
Thanks for the help @anuraaga and @chokoswitch

PR #3098, opened 2020-05-29
rebuilds lens package lock, fixes peer dependency and ts error

codefromthecrypt:
eslint-plugin-react-hooks 3 isn't compatible with the other things, yet.
```
npm WARN eslint-config-airbnb@18.1.0 requires a peer of eslint-plugin-react-hooks@^2.5.0 || ^1.7.0 but none is installed. You must install peer dependencies yourself.
npm WARN eslint-config-react-app@5.2.1 requires a peer of eslint-plugin-react-hooks@1.x || 2.x but none is installed. You must install peer dependencies yourself.
```

Also rebuilding the package lock showed a react-select TS problem,
though the workaround is inelegant. I spent a half hour looking on
GitHub and seems no one knows how to handle this correctly, most punt
to 'any'

https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/types.js

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -647,9 +647,13 @@
                 <exclude>**/nginx.conf</exclude>
                 <exclude>**/.idea/**</exclude>
                 <exclude>**/node_modules/**</exclude>
+                <exclude>**/build/**</exclude>
+                <exclude>**/coverage/**</exclude>
                 <exclude>**/.babelrc</exclude>
                 <exclude>**/.bowerrc</exclude>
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
ps I tried clicking around in the dependency service drop down and seems like all is ok

anuraaga:
Yeah looking forward to react-select going away with @tacigar's change in the meantime wouldn't worry about it

anuraaga (approved):

PR #3101, opened 2020-05-29
add note about generating jdiff, javadoc is TODO fixes #1221

### LINKED ISSUE #1221: publish javadoc and jdiff reports

jorgheymans:


codefromthecrypt (approved):
thanks!

PR #3121, opened 2020-06-18
Fixes glitch where root spans were ambiguous

### LINKED ISSUE #3001: Lens doesn't show the service name if any span in the trace is missing
## Describe the Bug
Some of my traces render just fine, but some of them render without service names. The span names and timing bars are fine, but the service names and the hierarchy doesn't show.

## Steps to Reproduce
Load this trace: https://i.fluffy.cc/wzGSB8HvZ83FHJ5RWk70cs8ZZwDSZ1TH.html

## Expected Behaviour
I can see the service names

codefromthecrypt:
There was a scenario where we thought we were removing accidental shared
flag on the root but, but weren't. cc @drolando 

Fixes #3001

Before
Screenshot 2020-06-18 at 4 18 42 PM

After
Screenshot 2020-06-18 at 9 10 16 PM

### DIFF
--- a/zipkin-lens/src/zipkin/span-cleaner.js
+++ b/zipkin-lens/src/zipkin/span-cleaner.js
@@ -288,8 +288,8 @@ export function mergeV2ById(spans) {
   // Look to see if there is a root span, in case we need to correct its shared flag
   if (sorted[0].parentId || !sorted[0].shared) return sorted;
 
-  // Sorting puts root spans first. If there's only one root, and it has shared flag, remove it.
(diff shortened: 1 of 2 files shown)


anuraaga (approved):

drolando (approved):

drolando:
I tried opening a bunch of traces and this issue doesn't seem to happen anymore! Thanks for looking into this

codefromthecrypt:
no problem @drolando!

PR #3130, opened 2020-06-30
Updates versions, most notably Armeria

codefromthecrypt:

PR #3131, opened 2020-07-01
Adds undocumented property to log Elasticsearch health check requests

codefromthecrypt:
This adds the below property, but doesn't assign an ENV variable as it
is mostly useful for our own tests:

```
zipkin.storage.elasticsearch.health-check.http-logging=headers
```

This also comments out in tests how to enable it temporarily. Ex to
troubleshoot #3130


zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/HttpClientFactory.java
```diff
@@ -97,7 +100,7 @@
   }
 
   /** This takes care to not expose health checks into wire level logging */
```
codefromthecrypt, line 102:
```suggestion
  /** We intentionally separate logging config of production requests from health checks */
```

anuraaga (approved):

Issue #3118, opened 2020-06-17
text-anchor center in trace view makes it hard to know service name


mrajah-twttr:
## Describe the Bug
When a trace is opened the first column shows the list of service names. these are set to text-anchor=middle with x=50% an y=50%.. in cases where the service names do not fit in the fixed width blocks this ends up showing the middle of the service name which is not ideal.
![image](https://user-images.githubusercontent.com/66497287/84953509-8294b700-b0a8-11ea-8dc8-7e7ab51e73b9.png)

## Steps to Reproduce
open a trace with long service name
[netflix-longsvcname.txt](https://github.com/openzipkin/zipkin/files/4795219/netflix-longsvcname.txt)

## Expected Behaviour
you can visually see the starting "n" characters of the service. Hover should show the full name.

PR #3133, opened 2020-07-03
[zipkin-lens] Add tooltip for service names

rahuliyer95:
Added a simple tooltip for service names.

Screenshot 2020-07-03 at 11 22 28 AM

### DIFF
--- a/zipkin-lens/src/components/TracePage/TraceTimeline/TraceTree.jsx
+++ b/zipkin-lens/src/components/TracePage/TraceTimeline/TraceTree.jsx
@@ -294,6 +294,8 @@ const TraceTree = React.memo(
                 className={classes.serviceBadgeText}
               >
                 {serviceName}
+                // This adds a tooltip to service name. This is useful for viewing large service names.
+                // @see - https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title
                 <title>{serviceName}</title>
               </text>
             </svg>
--- a/zipkin-lens/src/components/TracePage/TraceTimeline/TraceTree.jsx
+++ b/zipkin-lens/src/components/TracePage/TraceTimeline/TraceTree.jsx
@@ -294,9 +294,13 @@ const TraceTree = React.memo(
                 className={classes.serviceBadgeText}
               >
                 {serviceName}
-                // This adds a tooltip to service name. This is useful for viewing large service names.
-                // @see - https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title
-                <title>{serviceName}</title>
+                <title>
+                  {
+                    // This adds a tooltip to service name. This is useful for viewing large service
+                    // names. https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title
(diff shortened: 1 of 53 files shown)


codefromthecrypt:
Hi! thanks for helping. I wonder what is your motivation.
 is it a very long service name? if so can you show a screenshot of where this helps?

rahuliyer95:
Yes, should be useful for viewing the full name for services that have long names. An example screenshot 
Screenshot 2020-07-03 at 9 36 49 PM

codefromthecrypt:
makes sense. @openzipkin/ui anyone have other ideas on how to handle long service names? I admit I've also wondered about this.

zipkin-lens/src/components/TracePage/TraceTimeline/TraceTree.jsx
```diff
@@ -294,6 +294,7 @@ const TraceTree = React.memo(
                 className={classes.serviceBadgeText}
               >
                 {serviceName}
+                <title>{serviceName}</title>
```
codefromthecrypt: deserves a comment, unit test or both

tacigar:
I think tooltip is a good idea.

tacigar:
But I cannot see tooltip in my chrome(83.0.4103.116)... 🤔

jorgheymans:
Consider also #3118 and #2948

rahuliyer95:
I added a small comment about the tooltip and referenced it to the doc for the title element. I am not an expert here so I would love some guidance around the unit-test. I believe #3118 and #2948 are more involved and I would prefer tackling them separately and not as a part of this PR.

zipkin-lens/src/components/TracePage/TraceTimeline/TraceTree.jsx
```diff
@@ -294,6 +294,13 @@ const TraceTree = React.memo(
                 className={classes.serviceBadgeText}
               >
                 {serviceName}
+                <title>
```
codefromthecrypt, line 297: so this is all wrapped in `React.memo(` I suspect it is harder to test, does anyone know if we have an example that tests something like this?

we can merge, just so far one person said they cannot see tool tip as a result at the moment.. cc @openzipkin/ui

jorgheymans: Tooltips fine here on latest FF Ubuntu. 

@rahuliyer95 Wondering if this needs to be tested on a larger trace with a couple of thousand spans, would the implementation suffer from this? I don't have have such trace handy here but maybe it would be good to test this.

codefromthecrypt: @drolando mind testing this for impact ^^ I think we don't have a massive test trace

codefromthecrypt:
@hanahmily @beckje01 do you have large trace with large service names? We are looking for some feedback to make sure that the tooltip works, but also doesn't slow rendering etc.

PR #3136, opened 2020-07-05
Refactors DelayLimiter in preparation of re-use in cassandra v1

codefromthecrypt:
This refactors code relating to DelayLimiter so that it is easier to
reuse later. What will happen in a separate PR (or what will be
attempted) is change the limiter so that it can run a predicate instead
of an exact match. For example Cassandra v1's Indexer currently mutes
index insertions which are logically equivalent to data already written.

Concretely, a service name that appears 3 times in the same trace,
results in two writes (earliest and latest). If it appears again, and
between those timestamps, that write is muted as it is redundant.

### DIFF
--- a/zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CassandraSpanConsumer.java
+++ b/zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CassandraSpanConsumer.java
@@ -14,7 +14,6 @@
 package zipkin2.storage.cassandra.v1;
 
 import com.datastax.driver.core.Session;
-import com.google.common.cache.CacheBuilderSpec;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.LinkedHashSet;
@@ -43,7 +42,7 @@ final class CassandraSpanConsumer implements SpanConsumer {
   @Nullable final CompositeIndexer indexer;
   @Nullable final InsertAutocompleteValue.Factory insertAutocompleteValue;
 
-  CassandraSpanConsumer(CassandraStorage storage, CacheBuilderSpec indexCacheSpec) {
+  CassandraSpanConsumer(CassandraStorage storage) {
     Session session = storage.session();
     Schema.Metadata metadata = storage.metadata();
     searchEnabled = storage.searchEnabled;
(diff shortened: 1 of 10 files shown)


zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CassandraSpanConsumer.java
```diff
@@ -14,7 +14,6 @@
 package zipkin2.storage.cassandra.v1;
 
 import com.datastax.driver.core.Session;
-import com.google.common.cache.CacheBuilderSpec;
```
codefromthecrypt, line 17: changes like this, just strangle guava usages to be replaced by DelayLimiter into one place

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CompositeIndexer.java
```diff
@@ -27,19 +27,21 @@
   private final Set<Indexer> indexers;
   // Shared across all threads as updates can come from any thread.
   // Shared for all indexes to make data management easier (ex. maximumSize)
-  private final ConcurrentMap<PartitionKeyToTraceId, Pair> sharedState;
+  private final Map<PartitionKeyToTraceId, Pair> sharedState;
 
-  CompositeIndexer(CassandraStorage storage, CacheBuilderSpec spec, int indexTtl) {
-    this.sharedState = CacheBuilder.from(spec).<PartitionKeyToTraceId, Pair>build().asMap();
+  CompositeIndexer(CassandraStorage storage, int indexTtl) {
+    sharedState = CacheBuilder.newBuilder()
```
codefromthecrypt, line 33: later, this code will change to DelayLimiter (or an extension of it)

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchSpanConsumer.java
```diff
@@ -46,13 +46,13 @@
     this.indexTypeDelimiter = es.indexTypeDelimiter();
     this.searchEnabled = es.searchEnabled();
     this.delayLimiter = DelayLimiter.newBuilder()
-      .ttl(es.autocompleteTtl())
+      .ttl(es.autocompleteTtl(), TimeUnit.MILLISECONDS)
```
codefromthecrypt, line 49: this parameter changed because the units are different. Ex the index mutes are second granularity

zipkin/src/main/java/zipkin2/internal/DelayLimiter.java
```diff
@@ -21,67 +21,52 @@
...
-    return new Builder().build();
-  }
-
   public static Builder newBuilder() {
     return new Builder();
   }
 
   public static final class Builder {
-    Ticker ticker = new Ticker();
-    long ttlNanos = TimeUnit.HOURS.toNanos(1); // legacy default from cassandra
```
codefromthecrypt, line 35: these defaults and the factory method were not used except in the benchmarks.

anuraaga (approved):

PR #3137, opened 2020-07-06
Refactors Cassandra v1 indexing code and fixes some glitches

codefromthecrypt:
This change performs more cleanups in preparation of the move to Datastax
Java Driver 4, mostly consolidation in nature:

* Fixes bug when drop-down choices were added when span data wasn't indexed
* Uses less auto-value by choosing Map.Entry for pair/tuple data
* Simplify index state management by giving each indexer a separate cache
* Moves index clash test to CassandraSpanConsumerTest
* Adds a bunch of comments explaining loop logic of CassandraSpanConsumer


zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/IndexTraceId.java
```diff
@@ -0,0 +1,243 @@
...
+    final Map<Entry<String, Long>, Pair> sharedState;
+    final String table;
+    final int bucketCount;
+    final PreparedStatement preparedStatement;
+    final TimestampCodec timestampCodec;
+
+    Factory(CassandraStorage storage, String table, int indexTtl) {
+      super(TimeUnit.SECONDS.toMillis(storage.indexCacheTtl), storage.indexCacheTtl);
+      session = storage.session();
+      // TODO: this state should be merged with the DelayLimiter setup in the super class
```
codefromthecrypt: in reality this might not be possible to merge, as the data is used for different purposes. For example, here it is used to determine if an existing range will be extended.. not if an exact call was made prior.

As the main next goal is getting rid of guava, the follow-up PR may not actually merge these features into DelayLimiter, rather make a similar bounded expire-after-write map.

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/IndexTraceIdByServiceName.java
```diff
@@ -0,0 +1,39 @@
...
+
+  @Override public Insert declarePartitionKey(Insert insert) {
+    return insert
+      .value("service_name", QueryBuilder.bindMarker("service_name"))
+      .value("bucket", QueryBuilder.bindMarker("bucket"));
+  }
+
+  @Override public BoundStatement bindPartitionKey(BoundStatement bound, String partitionKey) {
+    return bound
+      .setInt("bucket", ThreadLocalRandom.current().nextInt(bucketCount))
```
codefromthecrypt: before, this saved a constant off for the value of `ThreadLocalRandom.current()`.. I moved it here, but possibly the idea was to ensure that there's fewer clashes between the small bucket choices... that it was `ThreadLocalRandom` made me think I should move it here.. but perhaps it could be reverted with a comment as to why we are saving a constant like that.

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/InsertTraceIdByAnnotation.java
```diff
@@ -1,58 +0,0 @@
...
-import java.util.Set;
-import java.util.concurrent.ThreadLocalRandom;
-import zipkin2.Span;
-
-import static zipkin2.storage.cassandra.v1.CassandraUtil.annotationKeys;
-import static zipkin2.storage.cassandra.v1.CassandraUtil.toByteBuffer;
-
-// QueryRequest.annotations/binaryAnnotations
-final class InsertTraceIdByAnnotation implements Indexer.IndexSupport {
-  private static final ThreadLocalRandom RAND = ThreadLocalRandom.current();
```
codefromthecrypt, line 28: specifically.. here is where we oddly saved off a "thread local random".. maybe the point was to save a random (regardless, it should be field not constant)

codefromthecrypt: @michaelsembwever if you happen to remember why we did this, let me know. I've left it in for now.

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CassandraSpanConsumer.java
```diff
@@ -80,74 +97,150 @@
...
-      // indexing occurs by timestamp, so derive one if not present.
-      long ts_micro = v2.timestampAsLong();
-      if (ts_micro == 0L) ts_micro = guessTimestamp(v2);
+    for (int i = 0; i < spanCount; i++) {
+      Span span = spans.get(i);
+      V1Span v1Span = converter.convert(span);
 
-      insertTraces.add(insertTrace.newInput(span, ByteBuffer.wrap(encoder.write(v2)), ts_micro));
+      // trace records need an insertion timestamp, so derive one if not present.
+      long ts_micro = span.timestampAsLong();
```
anuraaga, line 114: This is copied from before, but seems to be weird naming convention (unless it's referring to a virus scan)

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CassandraSpanConsumer.java
```diff
@@ -80,74 +97,150 @@
...
+      // parameters. For example, a search for serviceName=app&spanName=bar will look for a
+      // service_name_index row "app.foo" for the corresponding endTs and lookback. This has a few
+      // implications:
+      //
+      // * timestamps need only millis precision (as query precision is millis)
+      // * client-side indexing is pointless if there is no timestamp
+      // * populating choices for indexes never added is also pointless
+      //
+      // We floor the micros timestamp to millis and skip populating choices on zero.
+      long timestamp = 1000L * (span.timestampAsLong() / 1000L); // QueryRequest is precise to ms
```
anuraaga, line 158: `timestampMillis`

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/IndexTraceId.java
```diff
@@ -0,0 +1,247 @@
...
+  abstract static class Input {
+    static Input create(String partitionKey, long timestamp, long traceId) {
+      return new AutoValue_IndexTraceId_Input(partitionKey, timestamp, traceId);
+    }
+
+    abstract String partitionKey(); // ends up as a partition key, ignoring bucketing
+
+    abstract long ts(); // microseconds at millis precision
+
+    abstract long trace_id(); // clustering key
```
anuraaga, line 45: Ditto, do we usually name our methods like this even for json?

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/IndexTraceId.java
```diff
@@ -0,0 +1,247 @@
...
+    /**
+     * The input may include inserts that already occurred, or are redundant as they don't impact
+     * QueryRequest.endTs or QueryRequest.loopback. For example, a parsed timestamp could be between
+     * timestamps of rows that already exist for a particular trace. Optimized results will be
+     * smaller when the input includes traces with local spans, or when other threads indexed the
+     * same trace.
+     */
+    @Override public Iterator<Input> iterator() {
+      Set<IndexTraceId.Input> result = entriesThatIncreaseGap();
+      if (inputs.size() > result.size() && LOG.isDebugEnabled()) {
```
anuraaga: Guess we might check `isDebug` first

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/IndexTraceId.java
```diff
@@ -0,0 +1,247 @@
...
+        }
+        if (containsEntry(mappedInputs, needsUpdate, range.right)) {
+          result.add(Input.create(needsUpdate.getKey(), range.right, needsUpdate.getValue()));
+        }
+      }
+      return result;
+    }
+
+    static <K, V> void add(Map<K, Set<V>> multimap, K key, V value) {
+      Set<V> valueContainer = multimap.get(key);
```
anuraaga: Since this is storage, guess we can use Java 8 `computeIfAbsent(key, unused -> new LinkedHashSet<>())`

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/IndexTraceIdByServiceName.java
```diff
@@ -0,0 +1,41 @@
...
+package zipkin2.storage.cassandra.v1;
+
+import com.datastax.driver.core.BoundStatement;
+import com.datastax.driver.core.querybuilder.Insert;
+import com.datastax.driver.core.querybuilder.QueryBuilder;
+import java.util.concurrent.ThreadLocalRandom;
+
+// QueryRequest.serviceName
+final class IndexTraceIdByServiceName extends IndexTraceId.Factory {
+  // Legacy behaviour was to use a static singleton of ThreadLocalRandom for bucket selection
```
anuraaga, line 23: Do we need to preserve this?

anuraaga (approved):

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CassandraSpanConsumer.java
codefromthecrypt: :) these are the field names in the cassandra schema

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/IndexTraceIdByServiceName.java
codefromthecrypt: I punted to https://github.com/openzipkin/zipkin/pull/3137#discussion_r449952865 in case I'm missing something. let's carry on here?

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CassandraSpanConsumer.java
codefromthecrypt: this isn't timestamp in millis it is just floored to millis

`1467676800150134L -> 1467676800150000L`

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/IndexTraceId.java
codefromthecrypt: similar to above, these match the input variables in CQL. it isn't about json really, and is all over the cassandra v1 codebase.

codefromthecrypt:
OK I went ahead and finished the un-guava'ing which made the TraceIdIndexerTest a good bit larger than before. So, after this, we can move to Datastax 4 which can get rid of the last of it (ListenableFuture -> CompletableFuture which we can't do w/o v4).

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/TraceIdIndexer.java
```diff
@@ -0,0 +1,252 @@
...
+    long nanoTime() {
+      return System.nanoTime();
+    }
+
+    <K, V> Expiration<K, V> newExpiration(K key, V value) {
+      return new Expiration<>(this, key, value, nanoTime() + ttlNanos);
+    }
+
+    TraceIdIndexer newIndexer() {
+      trimCache(); // We let the cache overfill per call to SpanConsumer.accept. This trims it back
```
anuraaga: So IIUC, we clean up anytime we index a new batch of spans. May as well add a line of comment with that since it's not so obvious

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/TraceIdIndexer.java
```diff
@@ -0,0 +1,252 @@
...
+
+      // When the loop completes, we'll know one of our updates widened the interval of a trace, if
+      // it is the first or last timestamp. By ignoring those between an existing interval, we can
+      // end up with less Cassandra writes.
+      Set<Input> result = new LinkedHashSet<>();
+      for (Entry<String, Long> needsUpdate : toUpdate.keySet()) {
+        Expiration<Entry<String, Long>, Pair> existing = factory.cache.get(needsUpdate);
+
+        // unexpected, but possible. Something could have purged all our cached entries
+        Pair range = existing != null ? existing.getValue() : toUpdate.get(needsUpdate);
```
anuraaga, line 201: Do we need to check if `existing` is expired here?

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/TraceIdIndexer.java
codefromthecrypt: sure

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/TraceIdIndexer.java
codefromthecrypt: well since we just wrote it, it couldn't be purged due to expiry (expiration unit is seconds). the only thing possible is something else swamped the cache so it is not here due to the bounds of it (which is likely a tuning problem I suppose).

If it is expired we don't know if it is redundant or not is the main thing, so lack of knowledge means we should write it. I'll try to clarify in the comments.

codefromthecrypt:
@anuraaga I tried to fix the comments. PTAL

anuraaga (approved):

codefromthecrypt:
thanks for the look and advice @anuraaga!

PR #3143, opened 2020-07-11
Split dropped metrics to sampled out and storage error

huydx:
Motivation: Currently we don't know whether spans are dropped by error or sampled out

### DIFF
--- a/zipkin-collector/activemq/src/test/java/zipkin2/collector/activemq/ITActiveMQCollector.java
+++ b/zipkin-collector/activemq/src/test/java/zipkin2/collector/activemq/ITActiveMQCollector.java
@@ -149,6 +149,7 @@ public class ITActiveMQCollector {
       .isEqualTo(THRIFT.encodeList(spans).length * 2 + malformed1.length + malformed2.length);
     assertThat(activemqMetrics.spans()).isEqualTo(spans.size() * 2);
     assertThat(activemqMetrics.spansDropped()).isZero();
+    assertThat(activemqMetrics.spansSampledOut()).isZero();
   }
 
   /** Guards against errors that leak from storage, such as InvalidQueryException */
--- a/zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
+++ b/zipkin-collector/core/src/main/java/zipkin2/collector/Collector.java
@@ -214,7 +214,7 @@ public class Collector { // not final for mock
       }
     }
     int dropped = input.size() - sampled.size();
-    if (dropped > 0) metrics.incrementSpansDropped(dropped);
+    if (dropped > 0) metrics.incrementSpansSampledOut(dropped);
     return sampled;
   }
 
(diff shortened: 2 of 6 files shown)


codefromthecrypt:
are you actually using collector sample rate? it is a barely used feature and if it isn't set, they would never be dropped due to sampling..

huydx:
yes I recently introduce it due to storage overload, I thought most people use that feature..

codefromthecrypt:
the problem with it is that it requires uniform trace IDs or may not consistently drop a trace and isn't really surge protected (as it is an mask not a rate) 

Another way, if you are controlling the image, is meanwhile put a metric in CollectorSampler then you can see that and know how many are from this. If you find this is useful, we can pick this change back up?

codefromthecrypt:
ex you should be able to override this even if it isn't set ConditionalOnMissingBean possibly with the `@Primary` annotation.
```
  @Bean CollectorSampler traceIdSampler(@Value("${zipkin.collector.sample-rate:1.0}") float rate) {
    return CollectorSampler.create(rate);
  }
```

we can also change that bean to conditionalOnMissingBean

huydx:
> Another way, if you are controlling the image, is meanwhile put a metric in CollectorSampler then you can see that and know how many are from this. If you find this is useful, we can pick this change back up?

You mean implement CollectorSampler and add metrics into custom sampler?
It's a good idea, but I still think split storage error and sampled out metrics is reasonable. And for users are not setting sample-rate, they will still get the same metrics with same number with my change.

codefromthecrypt:
just this is a api break, so since almost no one uses this feature are breaking and it will add a zero count metric in prometheus because as far as I know micrometer will add metrics eagerly even if never incremented...

otoh custom collectors are also not common, and we can train people to ignore this.

for example, I don't think we'd ever even add collector sampling as it exists today if it was never there in the old thrift thing.

main issue is this is breaking api and adding distracting stuff over something that isn't used often and is also not a very good flow control strategy

codefromthecrypt:
for background, and really we should have never implemented what you are using right now.. this thing was only used by twitter and only for zookeeper managed rate.. a flat rate collector side doesn't do flow control in any helpful way. the sample rate in config was the "initial rate" which would change.  https://github.com/openzipkin-attic/zipkin-zookeeper

and that ZK thing had its own metrics also. It would be a lesser api break to make a factory for CollectorSampler like we do for CollectorMetrics, which would use the same keys... in fact it is probably not even an api break.

wdyt?

codefromthecrypt:
one neat thing if we did the factory approach is that we can add metrics only if someone actually set the rate to something also.. we just wrap it and make it an internal thing in the server config. Ex. CollectorSampler.Factory is injected.. by default it returns a NOOP.. no metrics no nothing. If the spring config is set, then the thing internally adds a metric (I assume you want prometheus anyway, not /metrics endpoint, but even that could be done)

That would make sure it doesn't drift our other collectors like in zipkin-aws or anything people have.

WDYT?

huydx:
i see, so for many reasons, I agree that the PR may not be exposed to normal end-user.

PR #3154, opened 2020-08-01
Cleans up Cassandra Session logic in preparation of Driver v4

codefromthecrypt:
The more notable change in Datastax Driver v4 is consolidation of
Cluster and Session types. This centralizes logic to create a session
regardless of test or mainline. It also ensures the cluster object is
closed during main and test code when storage is.

### DIFF
--- a/zipkin-storage/cassandra-v1/src/test/java/zipkin2/storage/cassandra/v1/CassandraStorageExtension.java
+++ b/zipkin-storage/cassandra-v1/src/test/java/zipkin2/storage/cassandra/v1/CassandraStorageExtension.java
@@ -13,12 +13,9 @@
  */
 package zipkin2.storage.cassandra.v1;
 
-import com.datastax.driver.core.Cluster;
-import com.datastax.driver.core.HostDistance;
-import com.datastax.driver.core.PoolingOptions;
 import com.datastax.driver.core.Session;
-import java.net.InetSocketAddress;
 import java.util.concurrent.TimeUnit;
+import org.junit.jupiter.api.TestInfo;
 import org.junit.jupiter.api.extension.AfterAllCallback;
 import org.junit.jupiter.api.extension.BeforeAllCallback;
 import org.junit.jupiter.api.extension.ExtensionContext;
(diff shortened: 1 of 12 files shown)


zipkin-storage/cassandra-v1/src/test/java/zipkin2/storage/cassandra/v1/CassandraStorageExtension.java
```diff
@@ -109,54 +93,27 @@ InetSocketAddress contactPoint() {
...
+        try (Session session = Access.tryToInitializeSession(contactPoint)) {
+          session.execute("SELECT now() FROM system.local");
+          logger().info("Obtained a connection to container ({})", contactPoint);
+          return null; // unused value
+        }
+      });
     }
   }
-
-  static Cluster getCluster(InetSocketAddress contactPoint) {
```
codefromthecrypt, line 154: might not be obvious, but these settings repeated in the default session factory, except partially. moving it up allows better control over what we configure and also things like the same logging config to work in test sessions as the main sessions

zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/DefaultSessionFactory.java
```diff
@@ -45,27 +47,23 @@
...
       if (cassandra.ensureSchema()) {
-        session = cluster.connect();
         Schema.ensureExists(keyspace, cassandra.searchEnabled(), session);
-        session.execute("USE " + keyspace);
       } else {
         LOG.debug("Skipping schema check on keyspace {} as ensureSchema was false", keyspace);
-        session = cluster.connect(keyspace);
       }
 
+      session.execute("USE " + keyspace);
```
anuraaga, line 61: Assuming connect / use is equivalent to `cluster.connect(keyspace)` from the old else branch

zipkin-storage/cassandra/src/test/java/zipkin2/storage/cassandra/CassandraStorageExtension.java
```diff
@@ -65,42 +57,49 @@ Session session() {
...
 
-  CassandraStorage.Builder computeStorageBuilder() {
-    InetSocketAddress contactPoint = contactPoint();
+  CassandraStorage.Builder newStorageBuilder(TestInfo testInfo) {
     return CassandraStorage.newBuilder()
-        .contactPoints(contactPoint.getHostString() + ":" + contactPoint.getPort())
-        .ensureSchema(true)
-        .keyspace("test_cassandra3");
+      .contactPoints(contactPoint())
+      .maxConnections(1)
```
codefromthecrypt, line 87: hostdistance local 1 is exactly the same as this in impl.

anuraaga (approved):

zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/DefaultSessionFactory.java
codefromthecrypt: yeah I looked into the code. basically cluster.connect(keyspace) is a composition of connect and use

codefromthecrypt:
thanks for having a look @anuraaga!

PR #3155, opened 2020-08-01
Check domain expiry

anuraaga:
Adds daily check of zipkin.io expiry and files an issue if less than 30 days left.

### DIFF
--- a/.github/workflows/check-domain-expiry.yml
+++ b/.github/workflows/check-domain-expiry.yml
@@ -3,8 +3,15 @@ on: [push]
 
 jobs:
   build:
-    name: Check zipkin.io
+    name: Check zipkin.io expiration
+    # MacOS comes with whois installed by default
     runs-on: macos-latest
     steps:
-      - name: whois
-        run:  exit $((($(date -j -f %Y-%m-%dT%H:%M:%SZ "$(whois zipkin.io | awk '/Registry Expiry Date:/ { print $4 }')" +%s) - $(date +%s)) / 3600 / 24 > 30 != 1))
+      # Exit code 0 is more than 30 days until expiration, 1 otherwise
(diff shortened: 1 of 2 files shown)


.github/workflows/check-domain-expiry.yml
```diff
@@ -0,0 +1,20 @@
...
+    - cron: '12 22 * * *'
+
+jobs:
+  build:
+    name: Check zipkin.io expiration
+    # MacOS comes with whois installed by default
+    runs-on: macos-latest
+    steps:
+      # Exit code 0 if more than 30 days until expiration, 1 otherwise
+      - run:  exit $((($(date -j -f %Y-%m-%dT%H:%M:%SZ "$(whois zipkin.io | awk '/Registry Expiry Date:/ { print $4 }')" +%s) - $(date +%s)) / 3600 / 24 > 30 == 0))
```
jorgheymans, line 13: command-line fu is :muscle:

jcchavezs:
I love this.

lør. 1. aug. 2020, 14:00 skrev Jorg Heymans <notifications@github.com>:

codefromthecrypt:
pretty slick. only q is if this shouldn't go on openzipkin website repo as that has the CNAME with zipkin.io in it. even using that could.be instructive and possibly reusable for others who have same problem more easily via copy/pasta

anuraaga:
Yeah I thought of that but was worried it would increase the chance of the issue being missed since it's not a repo we look at much. What do you think?

codefromthecrypt:
I think we can manually verify core members are watching. problem was no
one was watching before. even if of the 16 or so core members 12 dont pay
attention a few will. also it is so little volume on that repo it is less
likely core people will not watch it. that's my thinking..

anuraaga:
Sure - https://github.com/openzipkin/openzipkin.github.io/pull/156

PR #3156, opened 2020-08-02
Stops using prepared or query builder statements for simple queries

codefromthecrypt:
Before, we needlessly prepared service name query, which would have no
parameters. While, this is being followed up below, it is unlikely that
preparing is better than directly executing the statement:
https://groups.google.com/a/lists.datastax.com/forum/#!topic/java-driver-user/d6wLkH3xDLI

More importantly, there is a bug in the cassandra driver >v3.6 which
crashes on `QueryBuilder` statements that have no parameters. This
uses a simple statement for that instead (healthcheck). Moreover, the
log category that shows connection failures is now suggested in our
docs.

See https://github.com/datastax/java-driver/pull/1138 for the NPE

### DIFF
--- a/zipkin-storage/cassandra-v1/README.md
+++ b/zipkin-storage/cassandra-v1/README.md
@@ -11,8 +11,8 @@ The implementation uses the [Datastax Java Driver 3.x](https://github.com/datast
 operate against a local Cassandra installation.
 
 ## Logging
-Queries are logged to the category "com.datastax.driver.core.QueryLogger" when debug or trace is
-enabled via SLF4J. Trace level includes bound values.
+Queries are logged to the category "com.datastax.driver.core.QueryLogger.NORMAL" when debug or trace
+is enabled via SLF4J. Trace level includes bound values.
 
 See [Logging Query Latencies](http://docs.datastax.com/en/developer/java-driver/3.0/supplemental/manual/logging/#logging-query-latencies) for more details.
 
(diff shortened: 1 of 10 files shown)


zipkin-storage/cassandra-v1/README.md
```diff
@@ -11,8 +11,8 @@ The implementation uses the [Datastax Java Driver 3.x](https://github.com/datast
 operate against a local Cassandra installation.
 
 ## Logging
-Queries are logged to the category "com.datastax.driver.core.QueryLogger" when debug or trace is
-enabled via SLF4J. Trace level includes bound values.
+Queries are logged to the category "com.datastax.driver.core.QueryLogger.NORMAL" when debug or trace
+isenabled via SLF4J. Trace level includes bound values.
```
codefromthecrypt:
```suggestion
is enabled via SLF4J. Trace level includes bound values.
```

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CassandraStorage.java
```diff
@@ -384,7 +383,11 @@ public SpanConsumer spanConsumer() {
   public CheckResult check() {
     if (closeCalled) throw new IllegalStateException("closed");
     try {
-      session.get().execute(QueryBuilder.select("trace_id").from("traces").limit(1));
+      // Use direct CQL instead of query builder for simple statements.
```
anuraaga, line 386: Guess the world still builds SQL by concatenate strings :)

anuraaga (approved):

PR #3160, opened 2020-08-03
Bumps dependencies as high as we can go

codefromthecrypt:
Only one snagged and don't feel like pushing through yet.

Tested docker with local build including ui (lens)

### DIFF
--- a/docker/examples/docker-compose-ui.yml
+++ b/docker/examples/docker-compose-ui.yml
@@ -22,7 +22,7 @@ version: '2'
 
 services:
   zipkin-ui:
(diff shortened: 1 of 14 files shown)


docker/examples/docker-compose-ui.yml
```diff
@@ -22,7 +22,7 @@ version: '2'
 
 services:
   zipkin-ui:
-    image: openzipkin/zipkin-ui
+    image: openzipkin/zipkin-ui:test
```
codefromthecrypt: oops. will scrub out

anuraaga (approved):

PR #3161, opened 2020-08-03
Updates Cassandra docker to a newer JRE 1.8

codefromthecrypt:
Depends on https://github.com/openzipkin/docker-jre-full/pull/19

### DIFF
--- a/docker/storage/cassandra/install.sh
+++ b/docker/storage/cassandra/install.sh
@@ -16,8 +16,8 @@
 
 set -eu
 
-echo "*** Installing Python"
-apk add --update --no-cache python2
+echo "*** Temporarily installing Python and curl"
+apk add --update --no-cache python2 curl
 
 echo "*** Installing Cassandra"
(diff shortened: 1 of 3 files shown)


docker/storage/cassandra/install.sh
```diff
@@ -16,7 +16,7 @@
 
 set -eu
 
-echo "*** Installing Python and curl"
+echo "*** Temporarily installing Python and curl"
```
codefromthecrypt, line 19: just now noticed later we remove both of these regardless of if we installed them!

docker/storage/cassandra/run.sh
```diff
@@ -1,7 +1,7 @@
-#!/bin/bash
+#!/bin/sh
```
codefromthecrypt, line 1: I verified this manually also, there's no bash things we use here even when doing things like IP lookup

anuraaga (approved):

Issue #2921, opened 2019-11-14
Lens UI missing raw span and trace ids


lambcode:
Lens UI has lost the ability to show raw spanId, traceId, and parentSpanId in the recent UI overhaul. Reproduced in zipkin 2.19.1.

See #2507 and #1300 for when this was added the last two times

tacigar:
Thank you for your feedback!
I agree with you!

PR #3162, opened 2020-08-04
Add the spanId towards the end of the span bar in trace timeline view

mrajah-twttr:
Taking a stab at adding the span Id on the trace bars. this helps search for a spanId visually on the page without too much work as well.

Couple of call outs
1. Am super new to react so would love to know if the hard coding i have done of 0 and 80% are better replaced by something else
2. I wasnt able to quite understand what isTextLeft was about, flipping the condition to !isTextLeft made even existing span Name data look busted.

image

### DIFF
--- a/zipkin-lens/src/components/TracePage/TraceSummary.jsx
+++ b/zipkin-lens/src/components/TracePage/TraceSummary.jsx
@@ -34,7 +34,13 @@ const TraceSummary = React.memo(({ traceSummary }) => {
   const isRootedTrace = hasRootSpan(traceSummary.spans);
   const [rootSpanIndex, setRootSpanIndex] = useState(0);
   const isRerooted = rootSpanIndex !== 0;
-  const [currentSpanIndex, setCurrentSpanIndex] = useState(0);
+
+  
+  // If spanId is present as hash (subresource) make that the spanIndex to load
+  const getSpanIndexOnLoad = window.location.hash !== "" ?
+   findSpanIndex(traceSummary.spans, window.location.hash.substring(1)) : 0;
+
+  const [currentSpanIndex, setCurrentSpanIndex] = useState(getSpanIndexOnLoad);
   const [childrenHiddenSpanIndices, setChildrenHiddenSpanIndices] = useState(
     {},
   );


tacigar:
Thank you @mrajah-twttr !!

`isLeftText` means....

* `isLeftText === false`: the span's name and duration texts will be placed on the right side of the bar.
    * スクリーンショット 2020-08-05 18 13 04
* `isLeftText === true`: the span's name and duration text will be places on the left side of the bar.
    * スクリーンショット 2020-08-05 18 13 16

Now, when `isLeftText === true`, spanId is shown in here...

スクリーンショット 2020-08-05 18 20 13

mrajah-twttr:
thanks. any suggestion on how i can show the spanId for isLeftText == true or idea on how to fix the css to make it align correctly at start of box.. 

any other suggestions on the code itself?

codefromthecrypt:
I think adding the span ID to trace bar is cluttered. why would you add it to the bar? Are you trying to search for a span ID? Maybe a search would be better? If we know more maybe there's a cleaner way.

codefromthecrypt:
here's some background reading https://github.com/openzipkin/zipkin/issues/2921

codefromthecrypt:
in general when adding features added a few times one trick is to cc people who worked on them last time. It will get feedback, or has a chance to.

mrajah-twttr:
yes, its to allow ctrl-f search similar to how we search even service name currently. a search bar might be more powerful (search for anno or anything else in any span) but might be a much bigger change with complexity.. A small font spanId at end  seemed like a natural fit.

Given its a dev tool i felt its more valuable  to be functional than aesthetic but i understand if you feel its making it too cluttered.

codefromthecrypt:
Are you typically searching around for multiple span IDs in one trace or just one? this will steer things

Ex assuming one..

I think in dep graph we have a search by service name which is one option and far less cluttered than putting N possibly NNNN span IDs on the screen.

Another option is to allow pasting the span ID into the window which would scroll to it directly. ex trace/trace_id-span_id similarly support dash encoding to jump to span ID in the go to trace thing.

@drolando @jorgheymans @llinder @narayaruna thoughts on hopping to one span?

codefromthecrypt:
@bulicekj lazy github search.. does haystack-ui have a jump to span ID thing of any kind in trace view, or another way to get to a specific span?

codefromthecrypt:
ps if we go the current approach, assuming this is only for ctrl-f.. I would make the font almost transparent. It will work in find and not clutter as barely visible. I still think it will distract so prefer a different way, but this is less distracting.

mrajah-twttr:
i prefer keeping the spanId small instead of transparent so there is no confusion on where the cursor landed.. Typically just 1 span (given spanId are unique should have only one match)

FWIW i would love to have a search box but this is one of the most common features requested and would like a short term solution

codefromthecrypt:
I don't think search box, if that's the solution, is a long term change. I highlight this as there is existing code to search within a page (dependencies screen iirc). Maybe you can take a look at that?

Also we have to be aware of tech debt and not add things just to save some days of discussion which should be had.

Even doing this change at all before 3 different sites would use it is a stretch, asking to merge tech debt before even getting a response from other sites is premature.

We spent a lot of time to declutter the UI, lots of time in decisions last couple years. not only is embedding span IDs into bars clutter, but it is also uncommon in any UI. I don't think this is a good decision to do this especially if the main goal is to avoid a more complete feature. Users do not primarily think about the span IDs and detail is where they can be found when needed. anchors etc can also reveal (ex window can show a span searched for for bookmark purposes)

tacigar:
@mrajah-twttr 
I want to know the usecase (scenario) for using span ID to identify a span.
Do you think that usecase is common?

jorgheymans:
I can see the need for an easy way to jump to a span if the trace is very large or contains many similar looking spans (e.g. batch process messaging), but adding them visually into the UI is just not nice aesthetically. 

There might be a path towards this though. If we consider span being a 'subresource' of a trace, then URL wise it would need to show up as an [URI fragment](https://en.wikipedia.org/wiki/URI_fragment). If Lens were to support such fragments e.g. `http://localhost:9411/zipkin/traces/f2882c7bde1c028e#92168a494b0c68318`, we could load the trace detail screen with a specific span highlighted. UI wise, people could then just develop a bookmarklet that prompts for the span id and adds it to the URI. This would be beneficial for Lens in general (ability to 'deep link' into a span), and you could piggy back it for your search needs. WDOT ?

mrajah-twttr:
the primary usecase is when traces are discovered outside zipkin, we have a couple of such usecases
1. Our logs contain traceId and spanId
2. offline storage with larger retention where we perform more complex data mining and find some exmplar traces.

I love the idea of a subresource, i dont mind asking my customers to manually edit the URL to append the spanId for now. May be in the future we can have a full fledged search..

Any pointers on how to make this change ?

bulicekj:
> @bulicekj lazy github search.. does haystack-ui have a jump to span ID thing of any kind in trace view, or another way to get to a specific span?

We do not. We've got this in our backlog as a feature request but I have found just using the browser find on the raw trace is good enough for most situations. However as mentioned above the more spans that have similar service/operation names the worse that experience becomes. 

My opinion is the transparent text is an OK temporary solution but few people would think to search for invisible text. It would be much better to have a small searchbar that highlighted the matching span(s)

jcchavezs:
+1 on the jorgheymans suggestion. While I think it will be a half feature
ak think it is better than adding the span ID for EVERY span in the trace.
I am also concerned about edge cases, for example what happens if you
changed the local root and the span is not anymore in the screen? In that
sense I think the anchor approach is still good as it can point you
directly to that span by changing the local root.

tir. 11. aug. 2020, 06:50 skrev Jason Bulicek <notifications@github.com>:

mrajah-twttr:
closing this PR in favor of https://github.com/openzipkin/zipkin/pull/3184 @jorgheymans

PR #3163, opened 2020-08-05
Adds Docker HEALTHCHECK to Kafka and adjusts for upstream /bin/sh change

codefromthecrypt:
Before, we were linking individual images /bin/sh to BusyBox. This change
uses the upstream change here. Especially as it allows HEALTHCHECK to
work (HEALTHCHECK at Dockerfile level only works with /bin/sh)

https://github.com/openzipkin/docker-jre-full/pull/20

Using this, we add a HEALTHCHECK for our test Kafka image based on a Gitter
discussion.


docker/Dockerfile.release
```diff
@@ -108,4 +108,7 @@ USER zipkin
 
 EXPOSE 9410 9411
 
-ENTRYPOINT ["/busybox/sh", "run.sh"]
+# Same settings and rationale as Dockerfile (non-release)
+HEALTHCHECK --interval=5s --start-period=30s --timeout=5s CMD wget -qO- http://127.0.0.1:9411/health &> /dev/null || exit 1
```
codefromthecrypt, line 112: we forgot to add this to our main image!

docker/examples/docker-compose-kafka.yml
```diff
@@ -29,10 +29,7 @@ services:
     # environment:
       # - KAFKA_ADVERTISED_HOST_NAME=192.168.99.100
     ports:
-      - 2181:2181
```
codefromthecrypt, line 32: we don't need to bind host ports by default as it is especially confusing as only 19092 will work
the others are exposed directly by the Dockerfile so we don't need to re-declare expose here either.

docker/examples/docker-compose-mem.yml
```diff
@@ -20,7 +20,7 @@
 #
 # Note that this file is meant for learning Zipkin, not production deployments.
 
-version: '2'
+version: '2.4'
```
codefromthecrypt, line 23: this version is what allows health conditions (note: v3 does not!!)

anuraaga (approved):

codefromthecrypt:
verified

```bash
$ docker ps
CONTAINER ID        IMAGE                            COMMAND                  CREATED             STATUS                    PORTS                                          NAMES
a1183bf8dd51        grafana/grafana                  "/run.sh"                11 seconds ago      Up 10 seconds             0.0.0.0:3000->3000/tcp                         grafana
9a8aac993ff4        prom/prometheus                  "/bin/prometheus --c…"   12 seconds ago      Up 11 seconds             0.0.0.0:9090->9090/tcp                         prometheus
8d2c458100f6        openzipkin/zipkin-dependencies   "crond -f"               12 seconds ago      Up 11 seconds                                                            dependencies
a15e3d9d07f5        openzipkin/zipkin:test           "/bin/sh -c /zipkin/…"   12 seconds ago      Up 11 seconds (healthy)   9410/tcp, 0.0.0.0:9411->9411/tcp               zipkin
fd6755ad0c8b        openzipkin/zipkin-kafka:test     "/bin/sh -c /kafka/b…"   12 seconds ago      Up 11 seconds (healthy)   2181/tcp, 9092/tcp, 0.0.0.0:19092->19092/tcp   kafka-zookeeper
de98851f6db1        openzipkin/zipkin-mysql          "/bin/sh -c /mysql/r…"   12 seconds ago      Up 11 seconds             3306/tcp                                       mysql
```

docker/RATIONALE.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Why do we add HEALTHCHECK for test images?
+
+While most of our images are not production, we still add HEALTHCHECK for a better ad-hoc
+and automation experience. Many of our setups will not operate without a service
+dependency: having HEALTHCHECK present makes triage and scripting a bit easier.
+
+HEALTHCHECK on our test image serves primarily two purposes:
+ * ad-hoc or scripted status of health using docker ps instead of knowing Kafka commands
+ * to allow manual allows usage of the docker-compose v2 service_healthy condition
```
jeqo, line 11: nit: typo
```suggestion
 * to allow manual usage of the docker-compose v2 service_healthy condition
```

docker/examples/docker-compose-kafka.yml
jeqo: 👍

docker/examples/docker-compose-mem.yml
jeqo: just in case, is there any docs confirming this (v3 not supporting healthchecks)? I wasn't able to find this restriction.

jeqo (approved):

docker/examples/docker-compose-mem.yml
codefromthecrypt: I actually integrated this with a pinot image for test purposes, and it works in 2.x like so

```yaml
    depends_on:
      kafka-zookeeper:
        condition: service_healthy
```

The docs are very bad about this being removed, only mildly discusses it at the entrypoint of 'condition'
https://docs.docker.com/compose/compose-file/#depends_on
Screen Shot 2020-08-05 at 7 40 22 PM

Options for v3 are not incredible.. basically you have to use wait-for-it and duplicate the health check code of your upstream. I hope v2 lives forever

docker/RATIONALE.md
codefromthecrypt: thanks @jeqo! 421d8d682

docker/examples/docker-compose-mem.yml
jeqo: make sense. thanks @adriancole !

PR #3167, opened 2020-08-06
Update Armeria and Netty dependencies

minwoox:
Armeria 0.99.8 -> 0.99.9
Netty 4.1.50.Final -> 4.1.51.Final

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
@@ -26,7 +26,6 @@ import com.linecorp.armeria.server.annotation.Consumes;
 import com.linecorp.armeria.server.annotation.ConsumesJson;
 import com.linecorp.armeria.server.annotation.ExceptionHandler;
 import com.linecorp.armeria.server.annotation.Post;
-import com.linecorp.armeria.unsafe.PooledObjects;
 
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
@@ -114,16 +113,16 @@ public class ZipkinHttpCollector {
         return null;
       }
 
-      final HttpData content;
+      final HttpData requestContent;
       try {
-        content = UnzippingBytesRequestConverter.convertRequest(ctx, msg);
+        requestContent = UnzippingBytesRequestConverter.convertRequest(ctx, msg);
       } catch (Throwable t1) {
         propagateIfFatal(t1);
         result.onError(t1);
         return null;
       }
 
-      try {
+      try (HttpData content = requestContent) {
         // logging already handled upstream in UnzippingBytesRequestConverter where request context exists
         if (content.isEmpty()) {
           result.onSuccess(null);
@@ -158,8 +157,6 @@ public class ZipkinHttpCollector {
           result.onError(t1);
           return null;
         }
-      } finally {
-        PooledObjects.close(content);
       }
 
       return null;
@@ -247,7 +244,7 @@ final class UnzippingBytesRequestConverter {
       // The implementation of the armeria decoder is to return an empty body on failure
       if (content.isEmpty()) {
         ZipkinHttpCollector.maybeLog("Malformed gzip body", ctx, request);
-        PooledObjects.close(content);
+        content.close();
         throw new IllegalArgumentException("Cannot gunzip spans");
       }
     }
@@ -255,7 +252,7 @@ final class UnzippingBytesRequestConverter {
     if (content.isEmpty()) ZipkinHttpCollector.maybeLog("Empty POST body", ctx, request);
     if (content.length() == 2 && "[]".equals(content.toStringAscii())) {
       ZipkinHttpCollector.maybeLog("Empty JSON list POST body", ctx, request);
-      PooledObjects.close(content);
+      content.close();
       content = HttpData.empty();
     }
 
--- a/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
@@ -26,7 +26,6 @@ import com.linecorp.armeria.server.annotation.Consumes;
 import com.linecorp.armeria.server.annotation.ConsumesJson;
 import com.linecorp.armeria.server.annotation.ExceptionHandler;
 import com.linecorp.armeria.server.annotation.Post;
-
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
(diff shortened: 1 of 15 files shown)


zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
```diff
@@ -168,7 +159,7 @@ HttpResponse validateAndStoreSpans(SpanBytesDecoder decoder, ServiceRequestConte
           return null;
         }
       } finally {
-        ReferenceCountUtil.release(content);
+        PooledObjects.close(content);
```
anuraaga: Don't we have try / resources now?

anuraaga (approved):

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinHttpCollector.java
minwoox: Ah let me use that. Thanks!

codefromthecrypt (approved):
Amazing initiative. Thanks!

zipkin-server/src/test/java/zipkin2/server/internal/health/ITZipkinHealth.java
```diff
@@ -73,7 +73,7 @@
 
     // ensure we don't track health in prometheus
     assertThat(scrape())
-      .doesNotContain("health");
+      .doesNotContain("health_check");
```
minwoox: Is there any better string that only exposed by prometheus?
We have changed the meter tags to contain method and service name.
For example `scrape()` contains `method=getHealth` and `service=server.internal.health.ITzipkinHealth` which made this test failed. 😓

zipkin-server/src/test/java/zipkin2/server/internal/prometheus/ITZipkinMetrics.java
```diff
@@ -77,7 +77,7 @@
 
     // ensure we don't track prometheus, UI requests in prometheus
     assertThat(scrape())
-      .doesNotContain("prometheus")
+      .doesNotContain("prometheus_notifications")
```
minwoox: This one as well. Any better suggestion please?

zipkin-server/src/test/java/zipkin2/server/internal/health/ITZipkinHealth.java
codefromthecrypt: but isn't this possibly a correct positive? why would we be adding metrics for the scrape endpoint? Maybe you can dump the actual contents here.

zipkin-server/src/test/java/zipkin2/server/internal/prometheus/ITZipkinMetrics.java
codefromthecrypt: similarly we intentionally shouldn't see any meta stats about prometheus. Possibly this is a correct glitch and something else we need to disable now. Stats cost money (especially in SaaS) and we don't want to blanket add things especially monitoring the monitor without being conscious about it.

codefromthecrypt (commented):
Thanks for the help.. wondering where the new metrics are coming from, if there's a PR. Maybe it can help us hunt how to disable them.

zipkin-server/src/test/java/zipkin2/server/internal/prometheus/ITZipkinMetrics.java
minwoox: The metric contains `service="zipkin2.server.internal.prometheus.ZipkinMetricsController"` so it's the problem. 😄 

Here's the story.
We fixed to add FQCN of the service name for HTTP services as a metric tag in 0.99.7. https://github.com/line/armeria/pull/2780
So if the call is made to `ZipkinMetricsController.fetchMetricsFromMicrometer()`, 
one of the metric recorded is `...method="fetchMetricsFromMicrometer",service="zipkin2.server.internal.prometheus.ZipkinMetricsController",quantile="0.0"...`

But we had a bug in 0.99.8 that the `MetricCollectingService` is not automatically added.
If Armeria server is created using spring boot integration and `ArmeriaSettings.isEnableMetric()` returns `true`, `MetricCollectingService` is added by default.
https://github.com/line/armeria/pull/2898/files#diff-58a90bb70d1f88b9c2f45ee9bddd4468R126
https://github.com/line/armeria/pull/2898/files#diff-58a90bb70d1f88b9c2f45ee9bddd4468R148

So from 0.99.9 we add `MetricCollectingService` the `Server` so all the requests and responses are recorded by default.

Now, `metricsIsOK()` is executed before `prometheusIsOK()` is executed during the test which records the metric about `service="zipkin2.server.internal.prometheus.ZipkinMetricsController"` which contains `prometheus` that make the test fails.

minwoox: So it's better not to add `MetricCollectingService` by default using `ArmeriaSettings`? 🤔

codefromthecrypt:
ok I may not fully understand but it seems at least dirty stats could be at
play here. that and possibly a controller we have should be marked opt out
of metrics or both..

there's an expensive annotation for spring boot tests called
DirtiesApplicationContext or something for the former..

minwoox:
> that and possibly a controller we have should be marked opt out
of metrics or both..

Or we could just remove the `MetricCollectingService` that is added by Armeria spring boot autoconfiguration. It seems like Zipkin is collecting metrics of the requests under `/zipkin/api` and `/api`, we might not need the default `MetricCollectingService`. 😆 
https://github.com/openzipkin/zipkin/blob/master/zipkin-server/src/main/java/zipkin2/server/internal/prometheus/ZipkinPrometheusMetricsConfiguration.java#L91-L96

Let me just disable the `ArmeriaSettings.enableMetrics`.

codefromthecrypt:
Thanks @minwoox! I think maybe only main metrics we need to make sure work are the elasticsearch client ones particularly as we are using client-side loadbalancing. I'm not sure we don't already have IT for that.

minwoox:
> I think maybe only main metrics we need to make sure work are the elasticsearch client ones particularly as we are using client-side loadbalancing. I'm not sure we don't already have IT for that.

I counddn't find the IT for that. But `ArmeriaSettings.enableMetrics` is just about server-side, so no effect for the clients. 😉 

And the test failure seems flaky, do you have a chance to re-run it, please?

codefromthecrypt:
Excellent. Thanks for the help and followup @minwoox Thanks for the review feedback @anuraaga !

minwoox:
Thanks for reviewing! I hope I can have more chances to do more work on Zipkin. 😄

PR #3168, opened 2020-08-06
Migrate zipkin-junit from mock-web-server to armeria-junit

anuraaga:
I want to add a jupiter version of the zipkin rule, but I found mockwebserver doesn't seem to have a plan for jupiter support

https://github.com/square/okhttp/issues/4667

So first migrated to armeria-junit4. In a followup I'll extract common logic and add an extension as well - I think I'll use optional dependencies on junit4 / jupiter instead of splitting out an artifact.


zipkin-junit/src/main/java/zipkin2/junit/ZipkinRule.java
```diff
@@ -150,16 +148,106 @@ public ZipkinRule enqueueFailure(HttpFailure failure) {
...
+        return acceptSpans(req, decoder);
+      }
+    });
+    sb.decorator(((delegate, ctx, req) -> {
+      requestCount.incrementAndGet();
+
+      AggregatedHttpResponse maybeFailure = failureQueue.poll();
+      if (maybeFailure != null) {
+        if (maybeFailure == HttpFailure.DISCONNECT_DURING_REQUEST_BODY) {
+          return HttpResponse.ofFailure(new IllegalStateException("Closed"));
```
anuraaga, line 195: Not sure if this is the intent of this failure.

codefromthecrypt:
as armeria also implies a dependency on netty it might conflict things worse than junit does for those using this package. Maybe do a quick github search? I think vert.x uses it at least.

There's also https://github.com/square/okhttp/pull/6192 trying to address the problem from 16 days ago.. Is it possible we can keep this okhttp in other words, so I think while I understand the affection for armeria we should be fair about the problem also maybe?

specifically unless you are already using armeria would this change given that okhttp will split the classes up, be a better way out? Do we think the netty dep of armeria will be ok generally?

codefromthecrypt:
one thing I believe is prove or disprovable is where this lib is used.. if something else that uses netty like google's grpc is in place, knowing people tend to pin specific versions of netty, we'd have a problem that wasn't there before.

OTOH if no one is using it in such a scenario, we can rightly dismiss it. Even if there is, we can get an ack from folks about what impact they'd have on it, possibly even kicking this project to a contrib repo.

Before brave-tests we needed this a lot more than we need it nowadays, though it is true that projects who initialize statically need it. Ex I think Skywalking just recently started using this. Haven't looked if they use netty or not or are pinned to a version that armeria isn't compatible with.

anuraaga:
Yeah I only glanced at https://github.com/square/okhttp/pull/6192 - it seemed to not be moving but am ok with waiting for it. Good point about the potential for netty dependency hell.

yschimke:
Hoping to move forward with https://github.com/square/okhttp/pull/6317 to unblock our graal builds, probably in the order of 1-2 weeks.

yschimke:
It won't immediately give you a Junit 5 rule, but should in a follow up.  For now will just allow not bringing in a Junit 4 dependency.

codefromthecrypt:
@yschimke fantastic. thanks tons!

yschimke:
@adriancole any feedback on https://github.com/square/okhttp/pull/6330 - playing around with a Junit 5 Rule.  Not sure what is idiomatic.

PR #3171, opened 2020-08-08
es early deprecation fail

### LINKED ISSUE #2220: Fail integration tests if http warning headers are returned from ES
Based on advice from @jasontedor, we should be able to be more aware of things being deprecated if we crash our integration test requests:
https://github.com/elastic/elasticsearch/issues/23892#issuecomment-432921238

jorgheymans:
fixes https://github.com/openzipkin/zipkin/issues/2220


jorgheymans:
note i was unable to run the IT locally due to docker woes ...

@adriancole given that you logged https://github.com/openzipkin/zipkin/issues/2220 , do you remember which deprecated feature triggered you to create the issue ? Is there a scenario we could easily test this against ? Or a simple unit test outside of ES just for the decorator would suffice ?

codefromthecrypt:
I think it was : in index name. @xeraa do you know a simple "deprecated" api call we can run to test our.. failure thing works?

xeraa:
I think you'll need two things:

1. Check for "cluster, node, and index level settings that use deprecated features that will be removed or changed in the next major version" with the `GET _migration/deprecations` API: https://www.elastic.co/guide/en/elasticsearch/reference/current/migration-api-deprecation.html
  That's for any "static" settings and should require the free Basic license. IMO you could just run through your integration tests and check if the response of that API is empty.
2. For query deprecations there should be a custom header in the response. Example:

```sh
≻ curl -v -X GET "http://localhost:9200/users/_search" -H 'Content-Type: application/json' -d'{ "query": { "match": { "name": { "query": "philipp", "cutoff_frequency": 0.1 } } } }'
*   Trying ::1:9200...
* Connected to localhost (::1) port 9200 (#0)
> GET /users/_search HTTP/1.1
> Host: localhost:9200
> User-Agent: curl/7.71.1
> Accept: */*
> Content-Type: application/json
> Content-Length: 116
> 
* upload completely sent off: 116 out of 116 bytes
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Warning: 299 Elasticsearch-7.8.1-b5ca9c58fb664ca8bf9e4057fc229b3396bf3a89 "Deprecated field [cutoff_frequency] used, replaced by [you can omit this option, the [match] query can skip block of documents efficiently if the total number of hits is not tracked]"
< content-type: application/json; charset=UTF-8
< content-length: 262
< 
{"took":2,"timed_out":false,"_shards":{"total":1,"successful":1,"skipped":0,"failed":0},"hits":{"total":{"value":1,"relation":"eq"},"max_score":0.2876821,"hits":[{"_index":"users","_type":"_doc","_id":"1","_score":0.2876821,"_source":{
  "name": "Philipp"
}
}]}}
```

PS: Colon in the index name was already removed in 7.0, so you'd need to test that one against 6.x

```sh
curl -v -X PUT "http://localhost:9200/user:name"
*   Trying ::1:9200...
* Connected to localhost (::1) port 9200 (#0)
> PUT /user:name HTTP/1.1
> Host: localhost:9200
> User-Agent: curl/7.71.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 400 Bad Request
< content-type: application/json; charset=UTF-8
< content-length: 325
< 
{"error":{"root_cause":[{"type":"invalid_index_name_exception","reason":"Invalid index name [user:name], must not contain ':'","index_uuid":"_na_","index":"user:name"}],"type":"invalid_index_name_exception","reason":"Invalid index name [user:name], must not contain ':'","index_uuid":"_na_","index":"user:name"},"status":400}
```

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ElasticsearchStorageExtension.java
```diff
@@ -71,6 +87,7 @@
 
     try {
       tryToInitializeSession();
+      checkForDeprecatedFeatures();
```
codefromthecrypt: if this is about index templates, possibly we can make this a test instead. ex we don't have ITEnsureIndexTemplate like in cassandra ITEnsureSchema. My assumption is that this is independent of api calls and so only needs to run once, but it should run after we setup the templates. having it as a separate test file will make this part more explicit and close to what fails imho.

jorgheymans: makes sense, moved to a different test

jorgheymans:
This thing is paying off :-) Running the tests locally, it reveals some deprecated usage already:

```
java.lang.AssertionError: 
Expecting empty but was:<"{"cluster_settings":[],"node_settings":[{"level":"critical","message":"setting [xpack.monitoring.enabled] is deprecated and will be removed in the next major version","url":"https://www.elastic.co/guide/en/elasticsearch/reference/7.8/breaking-changes-7.8.html#deprecate-basic-license-feature-enabled","details":"the setting [xpack.monitoring.enabled] is currently set to [false], remove this setting (nodes impacted: [zipkin-elasticsearch])"}],"index_settings":{},"ml_settings":[]}">
```

And `ITElasticSearchStorageV6`  

```
Detected usage of deprecated API for request HeaderOverridingHttpRequest{[:method=GET, :path=/_template/zipkin-test:span_template, :scheme=http, :authority=localhost:32844]}:
299 Elasticsearch-6.8.10-537cb22 "[types removal] The parameter include_type_name should be explicitly specified in get template requests to prepare for 7.0. In 7.0 include_type_name will default to 'false', which means responses will omit the type name in mapping definitions."
```

Should i go ahead and attempt to fix these in the same PR ?

codefromthecrypt:
our image can disable `xpack.monitoring.enabled` here https://github.com/openzipkin/zipkin/blob/master/docker/storage/elasticsearch7/config/elasticsearch.yml wanna give a try?

On "The parameter include_type_name should be explicitly specified in get template requests to prepare for 7.0. In 7.0 include_type_name will default to 'false', which means responses will omit the type name in mapping definitions."

@xeraa do you happen to know if doing ^^ will break older versions of ES 6? If not, we could update that also..

xeraa:
1. `xpack.monitoring.enabled` will be removed. Binaries having a Basic licenses will just have the feature turned on, OSS ones don't have it: https://www.elastic.co/guide/en/elasticsearch/reference/current/monitoring-settings.html#general-monitoring-settings
2. `include_type_name` is like your escape hatch to keep mapping requests working like before the type removal. With 8.0 at the latest you will have to support calls without types though. https://www.elastic.co/guide/en/elasticsearch/reference/6.8/indices-put-mapping.html#_skipping_types_3 has a warning for indices created with 5.x — not sure if that is a concern for you. But for 6.x and 7.x this shouldn't be an issue (though anything before 6.8 is starting to be an old memory, so take any advice there with care).

jorgheymans:
Only warning left now is about the `catch-all` template using index-pattern `*`, can something be done about this ? @adriancole 

`
java.lang.IllegalArgumentException: Detected usage of deprecated API for request HeaderOverridingHttpRequest{[:method=PUT, :path=/_index_template/catch-all, :scheme=http, :authority=localhost:32798, content-type=application/json; charset=utf-8, content-length=186]}:
299 Elasticsearch-7.8.1-b5ca9c58fb664ca8bf9e4057fc229b3396bf3a89 "index template [catch-all] has index patterns [*] matching patterns from existing older templates [zipkin-test-span_template,.monitoring-es,.transform-internal-005,.monitoring-kibana,.logstash-management,.monitoring-logstash,ilm-history,zipkin-test-dependency_template,zipkin-test-autocomplete_template,.slm-history,.monitoring-beats,.monitoring-alerts-7,.transform-notifications-000002] with patterns (zipkin-test-span_template => [zipkin-test-span-*],.monitoring-es => [.monitoring-es-7-*],.transform-internal-005 => [.transform-internal-005],.monitoring-kibana => [.monitoring-kibana-7-*],.logstash-management => [.logstash],.monitoring-logstash => [.monitoring-logstash-7-*],ilm-history => [ilm-history-2*],zipkin-test-dependency_template => [zipkin-test-dependency-*],zipkin-test-autocomplete_template => [zipkin-test-autocomplete-*],.slm-history => [.slm-history-2*],.monitoring-beats => [.monitoring-beats-7-*],.monitoring-alerts-7 => [.monitoring-alerts-7],.transform-notifications-000002 => [.transform-notifications-*]); this template [catch-all] will take precedence during new index creation"
`

xeraa:
I think you are running into https://github.com/elastic/elasticsearch/pull/54367

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/VersionSpecificTemplates.java
```diff
@@ -63,9 +63,7 @@ String indexProperties(float version) {
       + "    \"index.number_of_shards\": " + indexShards + ",\n"
       + "    \"index.number_of_replicas\": " + indexReplicas + ",\n"
       + "    \"index.requests.cache.enable\": true";
-    // There is no explicit documentation of index.mapper.dynamic being removed in v7, but it was.
```
jorgheymans, line 66: because triggers a deprecation warning on ES6

jorgheymans:
@adriancole PTAL

codefromthecrypt (approved):
LGTM on green (we may need to kick travis a few times due to #3197)

Thanks for the help @jorgheymans and @xeraa!

jorgheymans:
Last 3 Travis builds error out here, can we merge this anyway ? It has nothing to do with my changes, locally the integration tests pass.


```
[INFO] Results:

[INFO] 

[ERROR] Errors: 

[ERROR] zipkin2.elasticsearch.integration.ITElasticsearchStorage$ITServiceAndSpanNames.getRemoteServiceNames

[ERROR]   Run 1: ITElasticsearchStorage$ITServiceAndSpanNames.getRemoteServiceNames » AbortedStream

[INFO]   Run 2: PASS
```

codefromthecrypt:
yep go for merge, these tests have been flakey since a recent armeria update and we've yet to get to the bottom of it #3197

PR #3173, opened 2020-08-10
Update Kafka to 2.6 which formally supports JRE 14

codefromthecrypt:
While we haven't had any problems, it is nice to use the version that
formally supports JRE 14 as that's our base image.

I cleaned up a bit to keep the overall image the same as before.

See https://downloads.apache.org/kafka/2.6.0/RELEASE_NOTES.html

### DIFF
--- a/docker/collector/kafka/install.sh
+++ b/docker/collector/kafka/install.sh
@@ -41,4 +41,10 @@ EOF
 
 mkdir /kafka/logs
 
+echo "*** Cleaning Up"
+apk del jq curl --purge
+# TODO: eventually cleanup irrelevant binaries from RocksDB
+# https://issues.apache.org/jira/browse/KAFKA-10380
+rm -rf kafka_$SCALA_VERSION-$KAFKA_VERSION site-docs bin/windows
+
 echo "*** Image build complete"
--- a/zipkin-collector/kafka/pom.xml
+++ b/zipkin-collector/kafka/pom.xml
@@ -29,11 +29,10 @@
   <properties>
     <main.basedir>${project.basedir}/../..</main.basedir>
     <!-- when changing this, a change to the version of kafka-junit is likely needed, too -->
-    <!-- TODO: update to 2.6.0 once https://github.com/charithe/kafka-junit/issues/54 is complete -->
-    <kafka.version>2.5.0</kafka.version>
+    <kafka.version>2.6.0</kafka.version>
     <!-- This version is tightly coupled to the version of kafka-clients being used.
      https://github.com/charithe/kafka-junit -->
-    <kafka-junit.version>4.1.9</kafka-junit.version>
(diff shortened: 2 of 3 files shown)


docker/collector/kafka/install.sh
```diff
@@ -41,4 +41,10 @@ EOF
 
 mkdir /kafka/logs
 
+echo "*** Cleaning Up"
+apk del jq curl --purge
```
codefromthecrypt, line 45: ack that we are tossing the builder image so this part doesn't matter

docker/collector/kafka/install.sh
```diff
@@ -41,4 +41,10 @@ EOF
...
+# TODO: eventually cleanup irrelevant binaries from RocksDB
+# https://issues.apache.org/jira/browse/KAFKA-10380
+rm -rf kafka_$SCALA_VERSION-$KAFKA_VERSION site-docs bin/windows
```
codefromthecrypt, line 48: deleting site-docs brought the size back to the same as 2.5+scala 12

docker/collector/kafka/install.sh
```diff
@@ -41,4 +41,10 @@ EOF
 
 mkdir /kafka/logs
 
+echo "*** Cleaning Up"
+apk del jq curl --purge
+# TODO: eventually cleanup irrelevant binaries from RocksDB
+# https://issues.apache.org/jira/browse/KAFKA-10380
```
codefromthecrypt, line 47: biggest single jar is the embedded RocksDB with several OS binaries inside: 15MB

anuraaga (approved):

jorgheymans (approved):

docker/collector/kafka/install.sh
jeqo: Good point, I don't think those binaries are required on the kafka package.

zipkin-collector/kafka/pom.xml
```diff
@@ -29,6 +29,7 @@
   <properties>
     <main.basedir>${project.basedir}/../..</main.basedir>
     <!-- when changing this, a change to the version of kafka-junit is likely needed, too -->
+    <!-- TODO: update to 2.6.0 once https://github.com/charithe/kafka-junit/issues/54 is complete -->
```
jeqo: Thanks for the reminder. I will check if testcontainers + zipkin kafka image can be used here to remove this dependency; similar to https://github.com/openzipkin-contrib/zipkin-storage-kafka/pull/70z

jeqo (approved):
LGTM

zipkin-collector/kafka/pom.xml
codefromthecrypt: yeah we can use our own "last image" so that the version bumping matches on things like our mysql tests.

codefromthecrypt:
awesome feedback folks!

PR #3176, opened 2020-08-17
Moves tests that read back prometheus metrics to ITZipkinMetricsDirty

codefromthecrypt:
Before, we moved most tests that read back metrics to `ITZipkinMetricsDirty`,
but "we missed a spot" and in doing so it causes flakes in Travis at the
moment on random change. This moves the last over and strengthens the
comments.

### DIFF
--- a/zipkin-server/src/test/java/zipkin2/server/internal/prometheus/ITZipkinMetricsDirty.java
+++ b/zipkin-server/src/test/java/zipkin2/server/internal/prometheus/ITZipkinMetricsDirty.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2019 The OpenZipkin Authors
+ * Copyright 2015-2020 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


anuraaga (approved):

zipkin-server/src/test/java/zipkin2/server/internal/prometheus/ITZipkinMetricsDirty.java
```diff
@@ -41,8 +39,14 @@
 import static zipkin2.server.internal.ITZipkinServer.url;
 
 /**
- * We cannot clear the micrometer registry easily, so we have recreate the spring context. This is
- * extremely slow, so please only add tests that require isolation here.
+ * Tests here look at values based on counter values, so need to run independently. It would seem
+ * correct to {@link PrometheusMeterRegistry#clear() clear the registry} to isolate counters
+ * incremented in one test from interfering with another. However, this clears the metrics
+ * themselves, resulting in an empty {@link PrometheusMeterRegistry#scrape()}.
```
shakuzen, line 45: This not working as hoped is unfortunate indeed. Increments to a `Counter` reference held that is registered to a meter registry that is `clear`ed effectively do nothing (from the view of the registry) since the counter is no longer registered to the `MeterRegistry`. Making a new Spring context will make a new registry and all the metrics will be registered to that again in the remade beans. If using the API that retrieves or creates a meter, this isn't an issue (e.g. `registry.counter("my.counter").increment()`) because the first time after `clear` it will create and register a Counter. Perhaps we can improve this situation or at least document it better in Micrometer.

codefromthecrypt: if "reset" lingo in prom matches setting back to zero, it could avail another method name if the feature ever existed...

PR #3178, opened 2020-08-19
Replace kafka-junit with testcontainer

jeqo:
Similar to rabbitmq integration test, this replaces usage of kafka-junit.

### DIFF
--- a/zipkin-collector/kafka/src/test/java/zipkin2/collector/kafka/KafkaCollectorRule.java
+++ b/zipkin-collector/kafka/src/test/java/zipkin2/collector/kafka/KafkaCollectorRule.java
@@ -31,7 +31,7 @@ import java.util.concurrent.ExecutionException;
 /** This should be used as a {@link ClassRule} as it takes a very long time to start-up. */
 class KafkaCollectorRule extends ExternalResource {
   static final Logger LOGGER = LoggerFactory.getLogger(KafkaCollectorRule.class);
-  static final String IMAGE = "openzipkin/zipkin-kafka:latest";
+  static final String IMAGE = "openzipkin/zipkin-kafka:2.21.5";
   static final int KAFKA_PORT = 19092;
   static final String KAFKA_BOOTSTRAP_SERVERS = "localhost:" + KAFKA_PORT;
   static final String KAFKA_TOPIC = "zipkin";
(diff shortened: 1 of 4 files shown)


zipkin-collector/kafka/src/test/java/zipkin2/collector/kafka/KafkaCollectorRule.java
```diff
@@ -0,0 +1,92 @@
...
+import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
+
+import java.util.Collections;
+import java.util.Properties;
+import java.util.concurrent.ExecutionException;
+
+/** This should be used as a {@link ClassRule} as it takes a very long time to start-up. */
+class KafkaCollectorRule extends ExternalResource {
+  static final Logger LOGGER = LoggerFactory.getLogger(KafkaCollectorRule.class);
+  static final String IMAGE = "openzipkin/zipkin-kafka";
```
codefromthecrypt: in others, we were specifying version tag. latest might be fine even if not necessarily repeatable... it would be less maintenance (and less thrashy than "master" tag) wdyt cc @anuraaga

anuraaga: Would recommend a tag for reproducible build, yeah

jeqo: Adding `latest`

zipkin-collector/kafka/src/test/java/zipkin2/collector/kafka/KafkaCollectorRule.java
```diff
@@ -31,7 +31,7 @@
 /** This should be used as a {@link ClassRule} as it takes a very long time to start-up. */
 class KafkaCollectorRule extends ExternalResource {
   static final Logger LOGGER = LoggerFactory.getLogger(KafkaCollectorRule.class);
-  static final String IMAGE = "openzipkin/zipkin-kafka";
+  static final String IMAGE = "openzipkin/zipkin-kafka:latest";
```
codefromthecrypt: I think @anuraaga probably meant reproducible as in using a specific version like the other ones. I've copied the version used by those which is easy enough to wildcard replace.

```suggestion
  static final String IMAGE = "openzipkin/zipkin-kafka:2.21.5";
```

codefromthecrypt:
Thanks @jeqo !

PR #3180, opened 2020-08-20
Improve search result table's design

tacigar:
The current search results table is cluttered.
And the code was more complicated than it needed to be.

So I've updated these components from scratch.

## **Before**

スクリーンショット 2020-08-31 8 24 14

## **After**

スクリーンショット 2020-09-04 0 44 07
スクリーンショット 2020-09-04 0 39 25

### Collapssible

feature-collapsible

### Filtering by service names

feature-filter

### Sortable

feature-sortable


tacigar:
I have a question.
The current result table has *filtering feature*.
Is this necessary?

スクリーンショット 2020-08-15 18 06 20

I haven't implemented this yet...

shakuzen:
> The current result table has _filtering feature_.
> Is this necessary?

Does this only allow filtering by service name? Is there a way to search for traces containing two or more different services? Maybe it is used for that, but if the search can do that maybe it isn't needed.

Unrelated and not critical to address in this pull request as I assume it is existing behavior, but what is the sorting of the service names? It made me wonder if it isn't more helpful to sort them by the highest number of spans if there isn't other meaning to the current order.

Someone who uses Zipkin on a regular basis should probably chime in on whether the service names being visible or not (without mousing over) affects their usual workflow (e.g. ctrl+F search in page for service... though I guess this is what the filter mentioned above is for).

tacigar:
Thank you for your comment!
I added service filter feature and sort service names by span count!

filter

スクリーンショット 2020-08-31 7 51 51

codefromthecrypt:
I raised this issue to correct the count so they aren't crazy inflated and also we can know what they mean cc @shakuzen https://github.com/openzipkin/zipkin/issues/3192

The UI presentation won't change, though this will help the numbers make sense.

The only impact this could have wrt presentations is if we want the charts to also show how many downstream/exit spans per service. I think we can wait for that, as it would layer on this if we did it as opposed to invalidate it.

zipkin-lens/src/components/DiscoverPage/DiscoverPageContent.tsx
```diff
@@ -399,29 +404,60 @@ const DiscoverPageContent: React.FC<DiscoverPageContentProps> = ({
...
 
-  const [traces, isLoadingTraces] = useSelector((state: RootState) => [
+  const [
+    traces,
+    isLoadingTraces,
+    traceSummaries,
+  ] = useSelector((state: RootState) => [
     state.traces.traces,
     state.traces.isLoading,
+    state.traces.traceSummaries as TraceSummary[],
```
anuraaga, line 490: Is it possible to avoid this cast?

zipkin-lens/src/components/DiscoverPage/DiscoverPageContent.tsx
```diff
@@ -399,29 +404,60 @@ const DiscoverPageContent: React.FC<DiscoverPageContentProps> = ({
...
   }, []);
 
-  let content: JSX.Element | undefined;
+  const [filters, setFilters] = useState<string[]>([]);
+  const filterOptions = useMemo(
+    () =>
+      Object.keys(
+        traceSummaries.reduce((acc, cur) => {
+          cur.serviceSummaries.forEach((serviceSummary) => {
+            acc[serviceSummary.serviceName] = true;
```
anuraaga: Can you use `Set` instead of object / keys?

zipkin-lens/src/components/DiscoverPage/DiscoverPageContent.tsx
```diff
@@ -483,14 +525,30 @@ const DiscoverPageContent: React.FC<DiscoverPageContentProps> = ({
...
+          <Container>
             <Box
               display="flex"
               alignItems="center"
               justifyContent="flex-end"
               mt={1.75}
             >
-              <Box mr={1} position="relative">
+              <TextField
+                label="Limit"
```
anuraaga: Can wait until all the UI feedback is finalized but let's not forget i18n

zipkin-lens/src/components/DiscoverPage/TraceSummaryRow.tsx
```diff
@@ -0,0 +1,231 @@
...
+};
+
+interface TraceSummaryRowProps {
+  traceSummary: TraceSummary;
+}
+
+const TraceSummaryRow: React.FC<TraceSummaryRowProps> = ({ traceSummary }) => {
+  const theme = useTheme();
+  const [open, setOpen] = useState(false);
+  const handleOpenButtonClick = useCallback(() => {
```
anuraaga: Believe you may be able to use `useReducer` instead of `useState` + `useCallback` since you only toggle.

codefromthecrypt (approved):
I like the design and regardless of service/span counts being not great, they weren't great before. I do think we should address that before release.

zipkin-lens/src/components/DiscoverPage/DiscoverPageContent.tsx
tacigar: Yes, but traceSummaries state is written in JavaScript, so this type is `any`..
So I want to cast until the type is attached to the state of traceSummaries..

zipkin-lens/src/components/DiscoverPage/DiscoverPageContent.tsx
anuraaga: Just confirming are you leaving this as is or considering Set?

tacigar: Oops! I was correcting a different code.
I'll use Set here.

jcchavezs:
I truly love this <3

tir. 1. sep. 2020, 07:20 skrev tacigar <notifications@github.com>:

zipkin-lens/src/components/DiscoverPage/DiscoverPageContent.tsx
tacigar: Done! 🙇

tacigar:
@jcchavezs Thank you so much!

shakuzen:
> Someone who uses Zipkin on a regular basis should probably chime in on whether the service names being visible or not (without mousing over) affects their usual workflow (e.g. ctrl+F search in page for service... though I guess this is what the filter mentioned above is for).

I'm still curious about this, but don't let my theoretical concern stop progress here if others are happy with the change. Compared to before, it is an extra click and mouseover away from seeing what the count is for a given service in a trace and if you are looking for the count of a particular service (theoretical use case: looking for traces where a service was called many times? or no times) you only have the first letter to figure out where to mouseover.

tacigar:
I see... 🤔 
I have two ideas..

* Add ~"Collapse All"~ "Expand All" button
* Add small tables under bar chart for showing complete service name and span count.

スクリーンショット 2020-09-01 19 10 48

## User scenario

1. Push "RUN QUERY" button
2. Filter services using Service Filter
3. If the user want to narrow down search more, push ~"COLLAPSE ALL"~ "Expand All" button.
4. All service name & span count chart and table will appear.
5. Then, the user can ctrl-F search (because the table displays complete service names).

What do you think? @shakuzen

codefromthecrypt:
PS the original design idea for the letters was that services that shared a same first letter would be grouped. The bar chart would not have X axis ordered by span count, rather by letters like books on a shelf. 

When you see a tall letter (high span count) you click into it and see which of the N services that might be there.

This is imperfect idea, as service names are neither always starting with a letter, nor are they distributed across the letters. However, it was an idea to allow you to browse a trace that has an extreme amount of service names in a way besides wrapping 20 times.

The design drifted a little from that which might not be bad, but it lost the impact of the grouping. You can see in the images the same letter repeated as it is now ordered by span count. If we go back to this then we can consider trying labels instead of the first letter or something else.

cc also @bulicekj 

PS anyone with a massive amount of (but valid) service name trace would be appreciated, cc @narayaruna @drolando @llinder also would be their feedback.

codefromthecrypt:
FWIW I think in practice this is showing that really large traces may have slightly different visual needs than smaller ones. If you have less than 10 service names, I think there's no point grouping at all, or even abbreviating. That said service names that are really long will always have a problem as their labels mess up things.

anuraaga (approved):

tacigar:
I think we need to discuss bar charts a bit more, so I decided to make it a separate PR and revive service name badges.

スクリーンショット 2020-09-04 0 39 25

new

codefromthecrypt:
I think screen shots are up to date correct?

tacigar:
yes!

codefromthecrypt:
EPIC!!!

codefromthecrypt:
Hi, I am looking closer now and feel expand/collapse in the index screen is clutter. I don't know why someone would click this also with the ability to filter by service..

Is this mainly to guess some people prefer to always see duration and other people not?

Screen Shot 2020-09-10 at 11 52 33 AM
Screen Shot 2020-09-10 at 11 52 20 AM

I personally always like less controls

PR #3182, opened 2020-08-21
Slims down Kafka by removing libs only needed for streams and connect

codefromthecrypt:
This takes another 33M off the image (which is half its size over our base image)

```
openzipkin/zipkin-kafka       test                    3513b24c41c6        6 minutes ago       133MB
openzipkin/zipkin-kafka       latest                  6bf053880509        2 weeks ago         166MB
```


docker/collector/kafka/install.sh
```diff
@@ -16,9 +16,7 @@
 set -eux
 
 echo "*** Installing Kafka and dependencies"
-apk add --update --no-cache jq curl
-
-APACHE_MIRROR=$(curl --stderr /dev/null https://www.apache.org/dyn/closer.cgi\?as_json\=1 | jq -r '.preferred')
+APACHE_MIRROR=$(curl --stderr /dev/null https://www.apache.org/dyn/closer.cgi\?as_json\=1 |sed -n '/preferred/s/.*"\(.*\)"/\1/gp')
 
 # download kafka binaries
 curl -sSL $APACHE_MIRROR/kafka/$KAFKA_VERSION/kafka_$SCALA_VERSION-$KAFKA_VERSION.tgz | tar xz
```
codefromthecrypt, line 22: we could simplify downloads further by inlining shell scripts and log config, but I think it is not an issue really

anuraaga (approved):

docker/collector/kafka/install.sh
```diff
@@ -16,9 +16,7 @@
 set -eux
 
 echo "*** Installing Kafka and dependencies"
-apk add --update --no-cache jq curl
-
-APACHE_MIRROR=$(curl --stderr /dev/null https://www.apache.org/dyn/closer.cgi\?as_json\=1 | jq -r '.preferred')
+APACHE_MIRROR=$(curl --stderr /dev/null https://www.apache.org/dyn/closer.cgi\?as_json\=1 |sed -n '/preferred/s/.*"\(.*\)"/\1/gp')
```
codefromthecrypt:
```suggestion
APACHE_MIRROR=$(curl --stderr /dev/null https://www.apache.org/dyn/closer.cgi\?as_json\=1 | sed -n '/preferred/s/.*"\(.*\)"/\1/gp')
```

codefromthecrypt:
@jeqo will this break https://github.com/openzipkin-contrib/zipkin-storage-kafka/blob/master/docker/docker-compose.test.yml?

I don't know if when using kafka streams what is needed broker side?

jeqo:
@adriancole broker is independent from both kafka-streams and connect modules. These libraries rely on common kafka protocol and client libraries. 

This libraries should be enough for broker to run:

https://github.com/apache/kafka/blob/8af7b96bfbffb2455191beac41ec8db4f6d24e3b/build.gradle#L719-L738

codefromthecrypt:
Thanks for the look @anuraaga and @jeqo!

PR #3184, opened 2020-08-21
Set default spanIndex based on spanId specified in URL hash

mrajah-twttr:
Set default spanIndex based on spanId specified in URL hash, this should scroll to the span, open up the span and highlight it

Need some guidance on how to make it scroll. the change currently works to highlight and open the details of the appropriate span @jorgheymans

### DIFF
--- a/zipkin-lens/src/components/TracePage/TraceSummary.jsx
+++ b/zipkin-lens/src/components/TracePage/TraceSummary.jsx
@@ -34,7 +34,14 @@ const TraceSummary = React.memo(({ traceSummary }) => {
   const isRootedTrace = hasRootSpan(traceSummary.spans);
   const [rootSpanIndex, setRootSpanIndex] = useState(0);
   const isRerooted = rootSpanIndex !== 0;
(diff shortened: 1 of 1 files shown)


jorgheymans:
Thanks, i have requested a review from our UI guru @tacigar as i'm not into the details at all on anything typescript related. 

Original comment this PR is based on: https://github.com/openzipkin/zipkin/pull/3162#issuecomment-670926360

jcchavezs:
I am curious about what happens when we have a local root that does not show the aimed span. Maybe we change the local root? cc @adriancole

tacigar:
Thanks, about for scrolling, an external library such as react-scroll may be helpful...
https://github.com/fisshy/react-scroll

tacigar:
Clicking a span twice will redisplay that span as the root.

reroot

Perhaps it would be easier to display this rerooted diagram than scrolling to the span...

codefromthecrypt:
very much agree that moving back from re-rooting should use the new anchor. I think we discussed moving back to an anchor in the first development of re-root, just we didn't have the fully formed idea for anchoring yet

Issue #3138, opened 2020-07-06
Special case Elasticsearch 7.8+ Composable index templates


codefromthecrypt:
**Feature:**
If we prefer composable templates or otherwise prioritize our index templates, we can avoid support problems like https://github.com/openzipkin/zipkin-support/issues/23 where the template doesn't kick in due to something else matching on '*'

**Rational**
[Composeable templates](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates.html) are both introduced and preferred in ES 7.8+, which refers to `_template` as legacy ones.

This would introduce an issue of composeable template priority, either it would have to determine automatically what value should be set or be parameterizable through env variable.

**Example Scenario**
Test would have a pre-existing composable template and ours still applies

Thanks @benqs for the hint. cc @openzipkin/elasticsearch

xeraa:
BTW this is part of a bigger change around time-based indices that will lead to [data streams](https://www.elastic.co/guide/en/elasticsearch/reference/master/data-streams.html); though that is still work in progress. So I would wait a little longer here and see if that might make sense for Zipkin as well.

PR #3185, opened 2020-08-22
Composable index templates for ES 7.8

ccharnkij:
# Description

For #3138, this PR lets Zipkin uses composable index templates for ES >= 7.8. Also, a new env variable is added for index template priority.

## Type of change

Please delete options that are not relevant.

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

# How Has This Been Tested?

Unit tests are added for the features. The unit tests passed when ran locally. Also, I have done E2E tests using both ES >=7.8 and ES < 7.8 using the brave sample project.

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/ZipkinElasticsearchStorageProperties.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/ZipkinElasticsearchStorageProperties.java
@@ -50,6 +50,7 @@ import zipkin2.elasticsearch.ElasticsearchStorage.LazyHttpClient;
  *     enabled: true
  *     http-logging: HEADERS
  *     interval: 3s
+ *   template-priority: 0
  * }</pre>
  */
 @ConfigurationProperties("zipkin.storage.elasticsearch")
@@ -212,6 +213,8 @@ class ZipkinElasticsearchStorageProperties implements Serializable { // for Spar
 
   private HealthCheck healthCheck = new HealthCheck();
 
+  private Integer templatePriority;
+
   public String getPipeline() {
     return pipeline;
   }
@@ -346,6 +349,10 @@ class ZipkinElasticsearchStorageProperties implements Serializable { // for Spar
     this.ssl = ssl;
   }
 
+  public Integer getTemplatePriority() { return templatePriority; }
+
+  public void setTemplatePriority(Integer templatePriority) { this.templatePriority = templatePriority; }
+
   public ElasticsearchStorage.Builder toBuilder(LazyHttpClient httpClient) {
     ElasticsearchStorage.Builder builder = ElasticsearchStorage.newBuilder(httpClient);
     if (index != null) builder.index(index);
@@ -360,6 +367,7 @@ class ZipkinElasticsearchStorageProperties implements Serializable { // for Spar
     if (maxRequests != null) {
       log.warning("ES_MAX_REQUESTS is no longer honored. Use STORAGE_THROTTLE_ENABLED instead");
     }
+    if (templatePriority != null) builder.templatePriority(templatePriority);
     return builder;
   }
 
--- a/zipkin-server/src/main/resources/zipkin-server-shared.yml
+++ b/zipkin-server/src/main/resources/zipkin-server-shared.yml
@@ -164,6 +164,7 @@ zipkin:
       health-check:
         enabled: ${ES_HEALTH_CHECK_ENABLED:true}
         interval: ${ES_HEALTH_CHECK_INTERVAL:3s}
+      template-priority: ${ES_TEMPLATE_PRIORITY:}
     mysql:
       jdbc-url: ${MYSQL_JDBC_URL:}
       host: ${MYSQL_HOST:localhost}
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
@@ -162,6 +162,15 @@ public abstract class ElasticsearchStorage extends zipkin2.storage.StorageCompon
     /** False disables automatic index template installation. */
     public abstract Builder ensureTemplates(boolean ensureTemplates);
 
+    /**
+     * Only valid when the destination is Elasticsearch >= 7.8. Indicates the index template
+     * priority in case of multiple matching templates. The template with highest priority is used.
+     * Default to 0.
+     *
+     * <p>See https://www.elastic.co/guide/en/elasticsearch/reference/7.8/_index_template_and_settings_priority.html
+     */
+    public abstract Builder templatePriority(@Nullable Integer templatePriority);
+
     /** {@inheritDoc} */
     @Override public abstract Builder strictTraceId(boolean strictTraceId);
 
@@ -212,6 +221,8 @@ public abstract class ElasticsearchStorage extends zipkin2.storage.StorageCompon
   abstract boolean ensureTemplates();
 
   public abstract int namesLookback();
+
+  @Nullable abstract Integer templatePriority();
 
   @Override public SpanStore spanStore() {
     ensureIndexTemplates();
@@ -337,12 +348,17 @@ public abstract class ElasticsearchStorage extends zipkin2.storage.StorageCompon
       indexReplicas(),
       indexShards(),
       searchEnabled(),
-      strictTraceId()
+      strictTraceId(),
+      templatePriority()
     ).get(version);
   }
 
   String buildUrl(IndexTemplates templates, String type) {
     String indexPrefix = indexNameFormatter().index() + templates.indexTypeDelimiter();
+
+    if (version() >= 7.8f && templatePriority() != null) {
+      return "/_index_template/" + indexPrefix + type + "_template";
+    }
     return "/_template/" + indexPrefix + type + "_template";
   }
 
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/VersionSpecificTemplates.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/VersionSpecificTemplates.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2019 The OpenZipkin Authors
+ * Copyright 2015-2020 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -33,14 +33,16 @@ final class VersionSpecificTemplates {
   final String indexPrefix;
   final int indexReplicas, indexShards;
   final boolean searchEnabled, strictTraceId;
+  final Integer templatePriority;
 
   VersionSpecificTemplates(String indexPrefix, int indexReplicas, int indexShards,
-    boolean searchEnabled, boolean strictTraceId) {
+    boolean searchEnabled, boolean strictTraceId, Integer templatePriority) {
     this.indexPrefix = indexPrefix;
     this.indexReplicas = indexReplicas;
     this.indexShards = indexShards;
     this.searchEnabled = searchEnabled;
     this.strictTraceId = strictTraceId;
+    this.templatePriority = templatePriority;
   }
 
   String indexPattern(String type, float version) {
@@ -66,12 +68,47 @@ final class VersionSpecificTemplates {
     return result + ",\n    \"index.mapper.dynamic\": false\n";
   }
 
-  /** Templatized due to version differences. Only fields used in search are declared */
-  String spanIndexTemplate(float version) {
-    String result = "{\n"
-      + "  " + indexPattern(TYPE_SPAN, version) + ",\n"
+  String indexTemplate(float version) {
+    if (useComposableTemplate(version)) {
+      return "\"template\": {\n";
+    }
+
+    return "";
+  }
+
+  String indexTemplateClosing(float version) {
+    if (useComposableTemplate(version)) {
+      return "},\n";
+    }
+
+    return "";
+  }
+
+  String templatePriority(float version) {
+    if (useComposableTemplate(version)) {
+      return "\"priority\": " + templatePriority + "\n";
+    }
+
+    return "";
+  }
+
+  String beginTemplate(String type, float version) {
+    return "{\n"
+      + "  " + indexPattern(type, version) + ",\n"
+      + indexTemplate(version)
       + "  \"settings\": {\n"
       + indexProperties(version);
+  }
+
+  String endTemplate(float version) {
+    return indexTemplateClosing(version)
+      + templatePriority(version)
+      + "}";
+  }
+
+  /** Templatized due to version differences. Only fields used in search are declared */
+  String spanIndexTemplate(float version) {
+    String result = beginTemplate(TYPE_SPAN, version);
 
     String traceIdMapping = KEYWORD;
     if (!strictTraceId) {
@@ -141,7 +178,7 @@ final class VersionSpecificTemplates {
         + "      \"_q\": " + KEYWORD + "\n"
         + "    }\n")
         + "  }\n"
-        + "}");
+        + endTemplate(version));
     }
     return result
       + ("  \"mappings\": {\n"
@@ -152,29 +189,23 @@ final class VersionSpecificTemplates {
       + "      \"tags\": { \"enabled\": false }\n"
       + "    }\n")
       + "  }\n"
-      + "}");
+      + endTemplate(version));
   }
 
   /** Templatized due to version differences. Only fields used in search are declared */
   String dependencyTemplate(float version) {
-    return "{\n"
-      + "  " + indexPattern(TYPE_DEPENDENCY, version) + ",\n"
-      + "  \"settings\": {\n"
-      + indexProperties(version)
+    return beginTemplate(TYPE_DEPENDENCY, version)
       + "  },\n"
       + "  \"mappings\": {\n"
       + maybeWrap(TYPE_DEPENDENCY, version, "    \"enabled\": false\n")
       + "  }\n"
-      + "}";
+      + endTemplate(version);
   }
 
   // The key filed of a autocompleteKeys is intentionally names as tagKey since it clashes with the
   // BodyConverters KEY
   String autocompleteTemplate(float version) {
-    return "{\n"
-      + "  " + indexPattern(TYPE_AUTOCOMPLETE, version) + ",\n"
-      + "  \"settings\": {\n"
-      + indexProperties(version)
+    return beginTemplate(TYPE_AUTOCOMPLETE, version)
       + "  },\n"
       + "  \"mappings\": {\n"
       + maybeWrap(TYPE_AUTOCOMPLETE, version, ""
@@ -184,7 +215,7 @@ final class VersionSpecificTemplates {
       + "      \"tagValue\": " + KEYWORD + "\n"
       + "    }\n")
       + "  }\n"
-      + "}";
+      + endTemplate(version);
   }
 
   IndexTemplates get(float version) {
@@ -199,6 +230,10 @@ final class VersionSpecificTemplates {
       .dependency(dependencyTemplate(version))
       .autocomplete(autocompleteTemplate(version))
       .build();
+  }
+
+  boolean useComposableTemplate(float version) {
+    return (version >= 7.8f && templatePriority != null);
   }
 
   /**
--- a/zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/ElasticsearchStorageTest.java
+++ b/zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/ElasticsearchStorageTest.java
@@ -199,6 +199,108 @@ class ElasticsearchStorageTest {
       String.format("ElasticsearchStorage{initialEndpoints=%s, index=zipkin}", server.httpUri()));
   }
 
+  /**
+   * Ensure that Zipkin uses the legacy resource path when priority is not set
+   */
+  @Test void check_create_legacy_indexTemplate_resourcePath_version78() throws Exception {
+    server.enqueue(AggregatedHttpResponse.of(
+      HttpStatus.OK, MediaType.JSON_UTF_8, "{\"version\":{\"number\":\"7.8.0\"}}"));
+    server.enqueue(SUCCESS_RESPONSE); // get span template
+    server.enqueue(SUCCESS_RESPONSE); // get dependency template
+    server.enqueue(SUCCESS_RESPONSE); // get autocomplete template
+    server.enqueue(SUCCESS_RESPONSE); // cluster health
+
+    storage.check();
+
+    server.takeRequest(); // get version
+
+    assertThat(server.takeRequest().request().path()) // get span template
+      .startsWith("/_template/zipkin-span_template");
+    assertThat(server.takeRequest().request().path()) // // get dependency template
+      .startsWith("/_template/zipkin-dependency_template");
+    assertThat(server.takeRequest().request().path()) // get autocomplete template
+      .startsWith("/_template/zipkin-autocomplete_template");
+  }
+
+  /**
+   * Ensure that Zipkin uses the correct resource path of /_index_template when creating index
+   * template for ES 7.8 when priority is set, as opposed to ES < 7.8 that uses /_template/
+   */
+  @Test void check_create_composable_indexTemplate_resourcePath_version78() throws Exception {
+    // Set up a new storage with priority
+    storage.close();
+    storage = newBuilder().templatePriority(0).build();
+
+    server.enqueue(AggregatedHttpResponse.of(
+      HttpStatus.OK, MediaType.JSON_UTF_8, "{\"version\":{\"number\":\"7.8.0\"}}"));
+    server.enqueue(SUCCESS_RESPONSE); // get span template
+    server.enqueue(SUCCESS_RESPONSE); // get dependency template
+    server.enqueue(SUCCESS_RESPONSE); // get autocomplete template
+    server.enqueue(SUCCESS_RESPONSE); // cluster health
+
+    storage.check();
+
+    server.takeRequest(); // get version
+
+    assertThat(server.takeRequest().request().path()) // get span template
+      .startsWith("/_index_template/zipkin-span_template");
+    assertThat(server.takeRequest().request().path()) // // get dependency template
+      .startsWith("/_index_template/zipkin-dependency_template");
+    assertThat(server.takeRequest().request().path()) // get autocomplete template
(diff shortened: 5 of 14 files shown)


zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/VersionSpecificTemplates.java
```diff
@@ -66,10 +67,35 @@ String indexProperties(float version) {
...
+    }
+
+    return "";
+  }
+
   /** Templatized due to version differences. Only fields used in search are declared */
   String spanIndexTemplate(float version) {
     String result = "{\n"
       + "  " + indexPattern(TYPE_SPAN, version) + ",\n"
+      + indexTemplate(version)
```
codefromthecrypt, line 98: beginTemplate

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/VersionSpecificTemplates.java
```diff
@@ -141,6 +167,8 @@ String spanIndexTemplate(float version) {
         + "      \"_q\": " + KEYWORD + "\n"
         + "    }\n")
         + "  }\n"
+        + indexTemplateClosing(version)
```
codefromthecrypt: squash this and below into endTemplate as it is distracting to add template priority here when sometimes there's no template anyway.

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/ElasticsearchStorageTest.java
```diff
@@ -199,6 +199,26 @@
...
+    server.enqueue(SUCCESS_RESPONSE); // get span template
+    server.enqueue(SUCCESS_RESPONSE); // get dependency template
+    server.enqueue(SUCCESS_RESPONSE); // get autocomplete template
+    server.enqueue(SUCCESS_RESPONSE); // cluster health
+
+    storage.check();
+
+    server.takeRequest(); // get version
+
+    assertThat(server.takeRequest().request().path()) // get span template
```
codefromthecrypt, line 245: sorry I don't understand what is special checked here, as unless blind, I would expect the same to pass before 7.8?

In any case, if we are verifying these are legacy, we should mention that in the test name or description, as currently the name doesn't seem to suggest what we are looking at (unless I miss something)

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/VersionSpecificTemplatesTest.java
```diff
@@ -107,6 +107,24 @@
...
 
+  @Test void version78() {
+    IndexTemplates template = storage.versionSpecificTemplates(7.8f);
+
+    assertThat(template.version()).isEqualTo(7.8f);
+    assertThat(template.autocomplete())
+      .withFailMessage("Starting at v7.x, we delimit index and type with hyphen")
+      .contains("\"index_patterns\": \"zipkin-autocomplete-*\"");
+    assertThat(template.span())
+      .contains("\"template\": {\n")
```
codefromthecrypt, line 139: nice. I suppose at some point we should switch this assert to json path, but works for me.

codefromthecrypt (commented):
This looks very good!

I would add something in the ITElasticsearch7 similar to ITEnsureSchema in cassandra, that sets up something that this priority would ensure doesn't mess up. Ex start ES docker and add the template prior to zipkin doing so. then make sure the things that failed before doesn't.

This way, we can automatically ensure something like @benqs comment doesn't hit us on 7.9 https://github.com/openzipkin/zipkin-support/issues/23#issuecomment-654066659

Also add a test for version 7.9 is same behavior as sometimes eyes cross on expressions. This happened to us in cassandra before.

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/ElasticsearchStorageTest.java
ccharnkij: Hi,

This test is for checking that Zipkin uses resource path of "/_index_template/....", as opposed to "/_template/...", for ES >= 7.8 when creating index templates. I can remove it if it's not needed. Or I can change test name (or add comment) for it to be more descriptive.

codefromthecrypt: yeah you can clarify with comment why it is important _index_template as I believe in some cases it worked before with _template. composable is a bit jargon to a passer-by.

ccharnkij:
@adriancole Please review again. I made begin/endTemplate and added commented on tests. I also added to ITElasticsearch7  as you suggested, I created a 'catch-all' template with pattern '*' and priority 0 prior to the tests, and the tests should proceed as normal.

I know that there are license fail checks, but since I didn't add any new file, I wasn't sure about updating the license header.

Also, I wasn't sure what you mean about tests for version 7.9. The code/tests that I added should apply to version 7.8 and above. Unless I'm missing something, please let me know.

codefromthecrypt:
FYI command to update license is here https://github.com/openzipkin/zipkin/blob/master/.github/CONTRIBUTING.md

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -88,7 +88,8 @@ public static Builder newBuilder(LazyHttpClient lazyHttpClient) {
       .flushOnWrites(false)
       .autocompleteKeys(Collections.emptyList())
       .autocompleteTtl((int) TimeUnit.HOURS.toMillis(1))
-      .autocompleteCardinality(5 * 4000); // Ex. 5 site tags with cardinality 4000 each
+      .autocompleteCardinality(5 * 4000)
+      .templatePriority("0"); // Ex. 5 site tags with cardinality 4000 each
```
codefromthecrypt: this comment doesn't belong. probably you want to say `// highest priority`

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/ElasticsearchStorageTest.java
```diff
@@ -199,6 +199,30 @@
       String.format("ElasticsearchStorage{initialEndpoints=%s, index=zipkin}", server.httpUri()));
   }
 
+  /**
+   * Ensure that Zipkin uses the correct resource path of /_index_template when creating index
+   * template for ES >= 7.8, as opposed to ES < 7.8 that uses /_template/
+   */
+  @Test void check_create_composable_indexTemplate_resourcePath() throws Exception {
+    server.enqueue(AggregatedHttpResponse.of(
+      HttpStatus.OK, MediaType.JSON_UTF_8, "{\"version\":{\"number\":\"7.8.0\"}}"));
```
codefromthecrypt, line 235: for example, eaisiest way is to change this to 7.9

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ITElasticsearchStorageV7.java
```diff
@@ -13,8 +13,14 @@
  */
 package zipkin2.elasticsearch.integration;
 
+import com.linecorp.armeria.common.*;
```
codefromthecrypt: if using square formatter it will expand wildcard imports

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ElasticsearchStorageExtension.java
```diff
@@ -116,6 +116,7 @@ Builder computeStorageBuilder() {
     WebClient client = builder.build();
     return ElasticsearchStorage.newBuilder(() -> client)
       .index("zipkin-test")
+      .templatePriority("10")
```
codefromthecrypt: maybe revert this as it effects all tests.. that or comment why.

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ITElasticsearchStorageV7.java
```diff
@@ -25,4 +31,40 @@
   @Override ElasticsearchStorageExtension backend() {
     return backend;
   }
+
+  /**
+   * Create a "catch-all" index template with the lowest priority prior to running tests to
+   * ensure that the index templates created during tests with higher priority function as
+   * designed. Only applicable for ES >= 7.8
+   */
+  @BeforeAll void setUpCatchAllTemplate() throws IOException {
```
codefromthecrypt: should be a separate test instead of a before for everything.

codefromthecrypt (commented):
thanks for the revision. Ideally the composite test is not implicit, rather a separate test. if the version is <7.8 you can use https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Assumptions.html#assumeTrue(java.util.function.BooleanSupplier)

ccharnkij:
@adriancole 3rd time's a charm?

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ITElasticsearchStorage.java
```diff
@@ -114,4 +114,12 @@
       storage.clear();
     }
   }
+
+  @Nested
+  class ITEnsureIndexTemplate extends zipkin2.elasticsearch.integration.ITEnsureIndexTemplate {
```
codefromthecrypt: I think you can move this to the ITEs7 class?

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ITEnsureIndexTemplate.java
```diff
@@ -0,0 +1,126 @@
...
+abstract class ITEnsureIndexTemplate {
+
+  ElasticsearchStorage storage;
+
+  /**
+   * Returns a new {@link ElasticsearchStorage.Builder} for connecting to the backend for the test.
+   */
+  protected abstract ElasticsearchStorage.Builder newStorageBuilder(TestInfo testInfo);
+
+  @BeforeAll void setUpCatchAllTemplate(TestInfo testInfo) throws IOException {
```
codefromthecrypt: with this pinned to the ES 7 IT, you don't need to check version as our image is already this I think right? In this case you can do this in the test method itself

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ITEnsureIndexTemplate.java
```diff
@@ -0,0 +1,126 @@
...
+    String url = "/_index_template/" + pattern;
+    AggregatedHttpRequest delete = AggregatedHttpRequest.of(HttpMethod.DELETE, url);
+    Internal.instance.http(storage)
+      .newCall(delete, (parser, contentString) -> null, "delete-index").execute();
+  }
+
+  /**
+   * Delete "catch-all" index template so it does not interfere with any other test
+   */
+  @AfterAll void deleteCatchAllTemplate() throws IOException {
```
codefromthecrypt: this should be  a "finally" of the catchAllTemplateTest as it would only need to be tested once anyway

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ITEnsureIndexTemplate.java
```diff
@@ -0,0 +1,126 @@
...
+    CheckResult check = storage.check();
+
+    assertThat(check.ok()).isTrue();
+
+    List<String> traceIds = asList(LOTS_OF_SPANS[0].traceId(), LOTS_OF_SPANS[1].traceId());
+    List<Span> spans = asList(LOTS_OF_SPANS[0], LOTS_OF_SPANS[1]);
+
+    storage.spanConsumer().accept(spans).execute();
+
+    assertThat(storage.traces().getTraces(traceIds).execute())
```
codefromthecrypt: verify the catch-all template works? you might want to make something more interesting so that you can use ES HTTP query on it. ex this one https://gist.github.com/adriancole/1af1259102e7a2da1b3c9103565165d7

codefromthecrypt (commented):
Thanks I think you did the hard work. Main thing is to isolate the catch-all to a single test and also re-scope it so that it would only ever invoke on ES 7.8+

ccharnkij:
@adriancole like that?

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ITElasticsearchStorageV7.java
```diff
@@ -25,4 +30,12 @@
   @Override ElasticsearchStorageExtension backend() {
     return backend;
   }
+
+  @Nested
+  class ITEnsureIndexTemplate extends zipkin2.elasticsearch.integration.ITEnsureIndexTemplate {
+    @Override protected ElasticsearchStorage.Builder newStorageBuilder(TestInfo testInfo) {
```
codefromthecrypt: as other tests we add to ITEnsureIndexTemplate won't necessarily effect templatePriority, I would move that setting into the test itself. Then, you can use the same setup as the other tests like this:

```java
    @Override protected ElasticsearchStorage.Builder newStorageBuilder(TestInfo testInfo) {
      return backend().computeStorageBuilder().index(index(testInfo));
    }

    @AfterEach public void clear() throws IOException {
      storage.clear();
    }
```

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ITEnsureIndexTemplate.java
```diff
@@ -0,0 +1,129 @@
...
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+abstract class ITEnsureIndexTemplate {
+
+  ElasticsearchStorage storage;
+
+  /**
+   * Returns a new {@link ElasticsearchStorage.Builder} for connecting to the backend for the test.
+   */
+  protected abstract ElasticsearchStorage.Builder newStorageBuilder(TestInfo testInfo);
```
codefromthecrypt: nit protected not needed in the same package

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ITEnsureIndexTemplate.java
```diff
@@ -0,0 +1,129 @@
...
+  String catchAllTemplate() {
+    return "{\n"
+      + "  \"index_patterns\" : [\"*\"],\n"
+      + "  \"priority\" : 0,\n"
+      + "  \"template\": {\n"
+      + "    \"settings\" : {\n"
+      + "      \"number_of_shards\" : 1\n"
+      + "    },\n"
+      + "    \"mappings\" : {\n"
+      + "      \"_source\" : { \"enabled\" : true }\n"
```
codefromthecrypt: if this is the catch all change you want to make, it would be useful to test that this applies. The test could pass for a subtle reason, then in ES 7.10 it might break and we wouldn't know what it was trying to do either.

For this reason, making a template that sets default which I think this does, maybe isn't a good choice. 
Instead try something with semantic value and also relevant in tracing? ex
```
  "mappings": {
    "properties": {
      "tags.http.status_code": {
        "norms": false,
        "type": "keyword"
      }
    }
  }
```

In this case, your test can switch from making a fake tag "queryTest"->"ok"  to a tag very commonly requested  like "http.status_code" "404"

Then, you can verify your catch-all worked by using HTTP to do the following (except in java):

```
curl -s 'localhost:9200/zipkin-span-*/_search?q=tags.http.status_code:404'
```

Ex what you are doing is literally verifying automatically this https://gist.github.com/adriancole/1af1259102e7a2da1b3c9103565165d7

codefromthecrypt (commented):
Almost there! let me know if you are out of steam, but otherwise it would be great if you can finish this. Thanks again

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ITEnsureIndexTemplate.java
ccharnkij: I don't think the suggested secondary template works anymore, given that ES no longer allows multiple matching index templates of the same priority. ES won't merge templates based on order anymore, and will only select 1 with the highest priority. Multiple index template with matching index patterns and priorities will result in error on PUT of the second index template. So in this case, it's either Zipkin's template or catch-all template, and I can't sneak in tags to zipkin-span-* like that.

I do see what you mean, though. And if the objective is to test the catch-all, I think I have to use a new index that doesn't match Zipkin's template at all. Thought?

codefromthecrypt: interesting.. thanks for the info! yeah a comment should say this is the case now ( I remember hints of this in the past from @xeraa ). I do think the default template should meddle with ours in a way that would be obvious then. For example, you could change something we don't set? then read back that it is unset (as opposed to inherited/merged) would that do it? 

meanwhile, using http in your test is still probably better as it is more realistic data, but yeah I can now expect people who setup multiple templates to have to do an offline merge of them instead of my gist now!

xeraa: 1. There are the [legacy / v1](https://www.elastic.co/guide/en/elasticsearch/reference/7.9/indices-templates-v1.html) templates that can be merged, but I would be careful (multiple templates with the same order will result in a non-deterministic merging order, merging valid templates might lead to an invalid mapping). Right now they are deprecated and we'll see when we'll remove them — you have at least until 8.0.
2. The new [component based approach](https://www.elastic.co/guide/en/elasticsearch/reference/7.9/index-templates.html) still allows you to combine multiple pieces, but you don't merge them (to avoid the two problems mentioned above). This feature was only added in 7.8, so I'm not sure what will be the sanest approach for you — support it in 7.x already or switch over for any (upcoming) 8.x cluster.

codefromthecrypt: ok so legacy/v1 would "for some time" retain the current behavior. I suppose in this case this means people in-place upgrading continue to work as this logic only applies when the templates don't already exist. cc also @basvanbeek for perspective on your schema tool. This chat will likely outgrow this PR :P

@ccharnkij for you and to not cage you to this decision tree, I would make a comment to the effect that @xeraa made in point 1. in your unit test and possibly also revise the javadoc/README if doesn't already say this. Basically, this relies on that we are not using legacy/v1 templates, so it is a win, not a merge.

hence we expect them to not merge. So, as mentioned earlier to show that actually happened (sanity check), have your template override something we don't and then magically notice that after the storage component initializes, whatever that setting is, is now ignored.

ccharnkij:
@adriancole how about this one? 

Regarding the legacy vs new template though, perhaps I could help more with it. I could allow both configuration based on environment variable, maybe ES_TEMPLATE_TYPE=legacy or new (or something like that). For ES >= 7.8, people can choose if they still want to use the legacy one or switch over to the one. Thought?

codefromthecrypt:
Thanks for the volunteer to move forward. I'm a fan of least config possible.

The template priority is a new config. Maybe we can use this in such a way that if you set it, you accept using "new templates" failing that defaults to legacy ones. If this is logical, it clears up the mixed state of 7.x and at any rate 8.0 won't support legacy ones. wdyt?

ccharnkij:
That works too. I just have to remove the implicit default priority=0 when not set. People that want to use new template just have to set priority all the time even if they want priority=0, until legacy template are completely removed.

codefromthecrypt:
yep sgtm!

ccharnkij:
@adriancole how about this one?

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -213,6 +222,8 @@ public final Builder dateSeparator(char dateSeparator) {
 
   public abstract int namesLookback();
 
+  @Nullable abstract String templatePriority();
```
codefromthecrypt: probably using Zipkin's Nullable for Integer is better, just to prevent a little brain break

zipkin-server/src/main/resources/zipkin-server-shared.yml
```diff
@@ -164,6 +164,7 @@ zipkin:
       health-check:
         enabled: ${ES_HEALTH_CHECK_ENABLED:true}
         interval: ${ES_HEALTH_CHECK_INTERVAL:3s}
+      priority: ${ES_TEMPLATE_PRIORITY:0}
```
codefromthecrypt: to match your docs..

```suggestion
      priority: ${ES_TEMPLATE_PRIORITY:}
```

zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/ZipkinElasticsearchStorageProperties.java
```diff
@@ -212,6 +213,8 @@ public void setInterval(Duration interval) {
 
   private HealthCheck healthCheck = new HealthCheck();
 
+  private String priority;
```
codefromthecrypt: make Integer

zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/ZipkinElasticsearchStorageProperties.java
```diff
@@ -346,6 +349,10 @@ public void setSsl(Ssl ssl) {
     this.ssl = ssl;
   }
 
+  public String getPriority() { return priority; }
+
+  public void setPriority(String priority) { this.priority = priority; }
```
codefromthecrypt: If this is a string to coerce empty to null, make a comment (also coerce empty to null :P) but still use an integer in the builder. I'm not actually sure what Spring does to an integer field set to ""

codefromthecrypt (commented):
Looks good. one minor thing and I think we're done

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -162,6 +162,15 @@ public final Builder dateSeparator(char dateSeparator) {
...
     public abstract Builder ensureTemplates(boolean ensureTemplates);
 
+    /**
+     * Only valid when the destination is Elasticsearch >= 7.8. Indicates the index template
+     * priority in case of multiple matching templates. The template with highest priority is used.
+     * Default to 0.
+     *
+     * <p>See https://www.elastic.co/guide/en/elasticsearch/reference/7.8/_index_template_and_settings_priority.html
+     */
+    @Nullable public abstract Builder templatePriority(Integer priority);
```
codefromthecrypt:
```suggestion
    public abstract Builder templatePriority(@Nullable Integer priority);
```

zipkin-server/src/main/resources/zipkin-server-shared.yml
```diff
@@ -164,6 +164,7 @@ zipkin:
       health-check:
         enabled: ${ES_HEALTH_CHECK_ENABLED:true}
         interval: ${ES_HEALTH_CHECK_INTERVAL:3s}
+      priority: ${ES_TEMPLATE_PRIORITY:}
```
codefromthecrypt: sorry I didn't notice this earlier: rename to template-priority

add a test in ZipkinElasticsearchStorageConfigurationTest to ensure that `zipkin.storage.elasticsearch.template-priority:` coerces to null not zero

zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/ZipkinElasticsearchStorageProperties.java
```diff
@@ -212,6 +213,8 @@ public void setInterval(Duration interval) {
 
   private HealthCheck healthCheck = new HealthCheck();
 
+  private Integer priority;
```
codefromthecrypt: rename to templatePriority

zipkin-server/src/main/java/zipkin2/server/internal/elasticsearch/ZipkinElasticsearchStorageProperties.java
```diff
@@ -50,6 +50,7 @@
  *     enabled: true
  *     http-logging: HEADERS
  *     interval: 3s
+ *   priority: 0
```
codefromthecrypt: rename to template-priority

codefromthecrypt (commented):
sorry I didn't notice the naming thing earlier

zipkin-server/src/main/resources/zipkin-server-shared.yml
ccharnkij: Would you prefer ES_TEMPLATE_PRIORITY= <--- (aka null)  to default to 0, instead of using legacy template? With the type set to nullable Integer, if the ES_TEMPLATE_PRIORITY is not integer (ES_TEMPLATE_PRIORITY=asfsd, for instance), Spring will throw error and fail start up. Andif ES_TEMPLATE_PRIORITY= (null), then legacy template will be used.

codefromthecrypt:
> Would you prefer ES_TEMPLATE_PRIORITY= <--- (aka null) to default to 0,
> instead of using legacy template? With the type set to nullable Integer, if
> the ES_TEMPLATE_PRIORITY is not integer (ES_TEMPLATE_PRIORITY=asfsd, for
> instance), Spring will throw error and fail start up. Andif
> ES_TEMPLATE_PRIORITY= (null), then legacy template will be used.
>
The main impact is the yml file which defaults this as then a user doesn't
need to know how it becomes null just how to make it not null :)

If spring has a way to represent a null integer, then I would recommend
using that in the template-priority ${ES_TEMPLATE_PRIORITY:XXXXXX} part for
XXXXXX
In any case a spring properties test can ensure however we are getting to a
null templatePriority works, in case spring boot changes their mind next
version about how to default things.. it will be caught.

make sense?

codefromthecrypt (approved):
I'm so impressed with what you've done, especially a first time committer to here. Thank you, Chanatan!

ccharnkij:
You're welcome. Glad I can be of help.

codefromthecrypt:
the build fail is unrelated flakey since armeria 1.0 update #3197

PR #3187, opened 2020-08-24
Updates Travis to latest Ubuntu

codefromthecrypt:


.travis.yml
```diff
@@ -1,10 +1,8 @@
 # Run `travis lint` when changing this file to avoid breaking the build.
 
-
-# Use a larger VM as we run a lot of services
```
codefromthecrypt, line 4: this point isn't one really as sudo has no special powers anymore afaik

codefromthecrypt:
I think there's some weird thing cache wise as exact same build on zipkin-aws doesn't fail to find JDK 14. I'm going to merge this and will own responsibility for fallout if it still fails after master eats it.

PR #3191, opened 2020-08-31
Use react-use instead of own hooks

tacigar:
Some hooks have been defined, but they are rarely used.
The only one used is `useDocumentTitle`, but the same hooks is available in react-use...
So this pr removes `/hooks/` dir and use `react-use` instead of it.


codefromthecrypt (approved):
great thanks!

tacigar:
It's a very trivial PR, so I'll merge it 🙇

PR #3193, opened 2020-08-31
Change directory structure

tacigar:
the directory structure is cluttered, so clean it up a bit.

* `src/dependencies` -> `src/components/DependenciesPage` as same as other page components (`src/components/DiscoverPage`, `src/components/TracePage` ...)
* `src/components/Common` -> `src/components/common`
    * There is no reason for this directory name to be PascalCase.
* `src/color/index.ts` -> `src/constants/color.ts`

PR #3194, opened 2020-08-31
Revert "Change directory structure"

tacigar:
Reverts openzipkin/zipkin#3193

in #3193, I renamed `components/Common` directory to `components/common` (`Common` -> `common`).
I think this case-sensitive change will probably have a bad effect on forks..
Perhaps you can't automatically merge the diffs, or if you can merge, lens won't work..
So revert this PR and I'll think of another method for this.


codefromthecrypt:
we can revert this, but we have also done case format correcting work in the past. We don't warranty forks.

anuraaga (commented):
Agree I wouldn't revert

tacigar:
Oh.. I see. Thanks.
I'll close this PR 🙇

Issue #3195, opened 2020-09-01
Show new lines in UI


fredy-virguez-adl:
**Feature:**
Add 'white-space: pre-line' or equivalent style to show new lines in tagged information.

**Rational**
In previous version new lines were shown, this allow users to read easily the multiline tagged information.

**Example Scenario**
Reading multiline tagged information

**Prior Art**
![image](https://user-images.githubusercontent.com/45631796/91884813-57077e80-ec4c-11ea-988c-54728feba27d.png)

![image](https://user-images.githubusercontent.com/45631796/91884649-0e4fc580-ec4c-11ea-93f1-2d8e3bf12443.png)

codefromthecrypt:
do you want to go ahead and make this change? seems legit

fredy-virguez-adl:
Ok, can you add me as a contributor? please

jcchavezs:
The usual workflow is that you fork the repo and then submit a PR. Could
you do that please?

fredy-virguez-adl:
Done. This is the PR https://github.com/openzipkin/zipkin/pull/3196

But travis-ci check is failing for not related reason.

PR #3196, opened 2020-09-02
Show multiline tags (#3195)

fredy-virguez-adl:


anuraaga (approved):
Nice, thanks!

codefromthecrypt (approved):
I went ahead and added a commit to address the license linter. While in our CONTRIBUTING file, it is something often missed: `./mvnw com.mycila:license-maven-plugin:format`

Thanks for the initiative also to make a unit test. will merge on green

PR #3199, opened 2020-09-06
Remove unnecessary css

tacigar:
There is `css` directory for setting up styles.
But these are no longer needed since material-ui overwrites more appropriate style values.
I want to remove them.


anuraaga (approved):

tacigar:
Thank you for review!

PR #3201, opened 2020-09-06
[Feedback Needed] Add service scatter chart

tacigar:
This is a chart showing the distribution of the duration, span count and start time of the trace search results.
It also indicates by color whether the trace contains error spans or not.

スクリーンショット 2020-09-06 11 43 59

service-scatter-chart

I like this chart because it allows me to understand the search results at a glance.
But I know that some people feel this complicated.

I have created a PR for more discussion.
I'd appreciate it if you could try it out and give me feedback.

Related: #2636

### DIFF
--- a/zipkin-lens/src/components/DiscoverPage/DiscoverPageContent.tsx
+++ b/zipkin-lens/src/components/DiscoverPage/DiscoverPageContent.tsx
@@ -703,7 +703,7 @@ const DiscoverPageContent: React.FC<DiscoverPageContentProps> = ({
                 </Box>
               </Box>
               <Collapse in={openServiceScatterChart}>
-                <Box width="100%" height={250} pt={1} pb={1} pl={4} pr={4}>
+                <Box width="100%" height={180} pt={1} pb={1} pl={4} pr={4}>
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
I think we have discussed this in a few threads so far. I would like to hear from an end user why this is important enough to add. You have recently bought simplicity back, but we should choose where to spend it.

@openzipkin/core I think we should stop adding new features until we get feedback from other sites and end users. Please help curate feedback as it is often only @tacigar and folks who formerly ran sites, but don't anymore. While glad folks not running sites are happy to help, and we should also do this, it is troubling we are working with less feedback from end users recently as this is a bad pattern.

FWIW on this, I think the red button is too loud and something that hints visually what this does would be better, such as a graph. 💹

jorgheymans:
I think Lens should excel in 2 parts: 1) finding a trace, and 2) making sense of said trace. Lets focus on this and make these bits really great. 

A graphical representation of a collection of traces could be useful but is hard to get right and does not always lead to finding a trace quicker. Also it gives the false impression (to me at least) we're venturing into metrics land.

PR #3205, opened 2020-09-11
Removes dead code from Lens

codefromthecrypt:
Found some of this with IntelliJ

### DIFF
--- a/zipkin-lens/src/components/DependenciesPage/DependenciesGraph.tsx
+++ b/zipkin-lens/src/components/DependenciesPage/DependenciesGraph.tsx
@@ -85,12 +85,18 @@ export const getNodesAndEdges = (dependencies: Dependencies) => {
   return { nodes, edges };
 };
 
+// IntelliJ may miss on "Unused property" analysis. Double-check here as we don't test CSS:
+// https://github.com/JedWatson/react-select/blob/cba15309c4d7523ab6a785c8d5c0c7ec1048e22f/packages/react-select/src/styles.js#L38-L61
 const reactSelectStyles = {
   control: (base: CSSProperties) => ({
     ...base,
     width: '15rem',
   }),
   option: (base: CSSProperties) => ({
+    ...base,
+    cursor: 'pointer',
+  }),
+  clearIndicator: (base: CSSProperties) => ({
     ...base,
     cursor: 'pointer',
   }),
--- a/zipkin-lens/src/components/DependenciesPage/DependenciesGraph.tsx
+++ b/zipkin-lens/src/components/DependenciesPage/DependenciesGraph.tsx
@@ -94,10 +94,6 @@ const reactSelectStyles = {
     ...base,
     cursor: 'pointer',
   }),
-  clearIndicator: (base: CSSProperties) => ({
-    ...base,
-    cursor: 'pointer',
-  }),
   dropdownIndicator: (base: CSSProperties) => ({
     ...base,
     cursor: 'pointer',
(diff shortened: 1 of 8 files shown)


zipkin-lens/src/components/DependenciesPage/DependenciesGraph.tsx
```diff
@@ -94,10 +94,6 @@ const reactSelectStyles = {
     ...base,
     cursor: 'pointer',
   }),
-  clearIndicator: (base: CSSProperties) => ({
-    ...base,
-    cursor: 'pointer',
-  }),
```
tacigar: I think this code is alive.

Without this code the cursor does not change when hovering over the clear button.

without

With this code: 

with

zipkin-lens/src/slices/remoteServicesSlice.ts
```diff
@@ -77,6 +77,4 @@ const remoteServicesSlice = createSlice({
   },
 });
 
-export const { clearRemoteServices } = remoteServicesSlice.actions;
```
tacigar, line 80: Thanks!
It looks like these codes can be removed as well :)

https://github.com/openzipkin/zipkin/blob/5562dd5b3677944027450f546fe4b9cc1b2ca5bd/zipkin-lens/src/slices/remoteServicesSlice.ts#L55-L59

https://github.com/openzipkin/zipkin/blob/5562dd5b3677944027450f546fe4b9cc1b2ca5bd/zipkin-lens/src/slices/spansSlice.ts#L55-L59

https://github.com/openzipkin/zipkin/blob/5562dd5b3677944027450f546fe4b9cc1b2ca5bd/zipkin-lens/src/slices/spansSlice.ts#L80

zipkin-lens/src/components/DependenciesPage/DependenciesGraph.tsx
codefromthecrypt: intellij analysis is imperfect :) I will put a comment that this is "not quite dead"

zipkin-lens/src/slices/remoteServicesSlice.ts
codefromthecrypt: great

codefromthecrypt: I am curious on these.. why can I remove them? Maybe you can tell me for each below:

servicesSlice, autocompleteKeys - only loaded ever once, so no need to clear
autocompleteValues - currently no reducer; data is partitioned by 'autocompleteKey'
remoteServices, spans - reducer to remove; data partitioned by 'service'
dependencies - contains a clear reducer, the data is not really cache-able except what's defined in HTTP

Later I would like to refactor based on this conversation. I would like to delete code, just I don't understand when we need a reducer to clear and not. cc @anuraaga

tacigar: https://github.com/openzipkin/zipkin/blob/5562dd5b3677944027450f546fe4b9cc1b2ca5bd/zipkin-lens/src/components/DiscoverPage/SearchBar/SearchBar.tsx#L138-L149

As soon as service criterion is changed, the spans and remoteServices fetching will begin, and the previous spans and remoteServices will be cleared.
So I think there is no need to explicitly clear them.

codefromthecrypt: thanks!

codefromthecrypt:
ok let me know anything else on this round

tacigar (approved):
Thank you so much! 😃

anuraaga (approved):

PR #3207, opened 2020-09-15
Bumps all versions and gets rid of NPM audit problems

codefromthecrypt:
Besides bumping versions, this moves the defunct `mvy` command to maven,
so it can still work in windows but not rely on something that trips
audits.


zipkin-lens/package.json
```diff
@@ -91,7 +90,6 @@
     "lint": "eslint \"src/**/*.{js,jsx,ts,tsx}\"",
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "postbuild": "mvy -f build target/classes/zipkin-lens",
```
codefromthecrypt, line 94: this was easier than the maven alternative, but it is abandoned so nagging npm audit

tacigar (approved):
Awesome!

anuraaga (approved):

PR #3208, opened 2020-09-17
Migrates Elasticsearch demo images to OSS license

codefromthecrypt:
This uses OSS license for Elasticsearch 6 and 7, reducing the size of
the images. Image size is further reduced in 7 by using a recent change
to our JRE that allows Thai segmentation support (module jdk.localedata)

This also reduces test execution time by only executing the 7.8+ schema
test vs the entire suite again.

Finally, this enables HEALTHCHECK on the images, so that we can later
automatically test them. Even though these are demo images, if they
break, it can causes time loss.


docker/storage/elasticsearch6/Dockerfile
```diff
@@ -12,41 +12,39 @@
...
 
-ENV ELASTICSEARCH_VERSION 6.8.11
+ENV ELASTICSEARCH_VERSION 6.8.12
 
-RUN apk add --update curl
+WORKDIR /install
 
-WORKDIR /elasticsearch
+# We don't download bin scripts as we customize for reasons including BusyBox problems
+RUN curl -sSL https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-oss-$ELASTICSEARCH_VERSION.tar.gz| tar xz \
```
codefromthecrypt, line 22: the `-oss` part removes the need to disable x-pack

docker/storage/elasticsearch6/Dockerfile
```diff
@@ -12,41 +12,39 @@
...
 
 FROM openzipkin/jre-full:14.0.2-14.29.23
 LABEL MAINTAINER Zipkin "https://zipkin.io/"
 
-# https://github.com/elastic/elasticsearch/pull/31003 was closed won't fix
-ENV ES_TMPDIR /tmp
-
-RUN adduser -g '' -h /elasticsearch -D elasticsearch
+# Add HEALTHCHECK and ENTRYPOINT scripts into the default search path
+COPY docker/storage/elasticsearch7/docker-bin/* /usr/local/bin/
```
codefromthecrypt, line 31: intentionally using the same scripts. didn't extract a common "elasticsearch" directory over this, as it would likely be more distracting.

docker/storage/elasticsearch7/Dockerfile
```diff
@@ -12,43 +12,40 @@
...
-ENV ELASTICSEARCH_VERSION 7.8.1
+ENV ELASTICSEARCH_VERSION 7.9.1
 
-RUN apk add --update curl
+WORKDIR /install
 
-WORKDIR /elasticsearch
+# Download only the OSS distribution (lacks X-Pack)
+# We don't download bin scripts as we customize for reasons including BusyBox problems
+RUN curl -sSL https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-oss-$ELASTICSEARCH_VERSION-no-jdk-linux-x86_64.tar.gz| tar xz \
```
codefromthecrypt, line 23: The `-no-jdk` part is needed in ES 7+ to avoid downloading a bundled JDK

docker/storage/elasticsearch7/Dockerfile
```diff
@@ -12,43 +12,40 @@
...
 
-WORKDIR /elasticsearch
+# Ensure the process doesn't run as root
+RUN adduser -g '' -h ${PWD} -D ${USER}
+USER ${USER}
 
-# elasticsearch complains if run as root
-USER elasticsearch
-
-EXPOSE 9200 9300
```
codefromthecrypt, line 50: there's no need to expose 9300

docker/storage/elasticsearch7/docker-bin/docker-healthcheck
```diff
@@ -0,0 +1,31 @@
...
+
+# HEALTHCHECK for use in `docker ps` or `docker-compose ps`.
+# It can also be used as a readiness probe in k8s
+
+# Fail on unset variables, but don't quit on rc!=0, so we can log what happened
+set -u +e
+
+IP="$(hostname -i || echo '127.0.0.1')"
+
+HEALTH_OUTPUT=$(wget -qO- http://${IP}:9200/_cluster/health 2>&1)
```
codefromthecrypt, line 24: not scraping the output as don't feel like it, but if we later feel like it, it would be easy to add to this script

docker/storage/elasticsearch7/docker-bin/start-elasticsearch
```diff
@@ -0,0 +1,39 @@
...
+# Notable settings:
+# * lower heap size
+# * tmpdir manual as https://github.com/elastic/elasticsearch/pull/31003 was closed won't fix
+# * disable log4j JMX not just because we don't use it...
+#  * ES enables security manager https://github.com/elastic/elasticsearch/issues/21932#issuecomment-264435034
+
+exec java \
+$ES_JAVA_OPTS \
+  -verbose:gc \
+  -Xms256m -Xmx256m -XX:+ExitOnOutOfMemoryError \
```
codefromthecrypt, line 33: ES will crash if you don't match heap sizes. During our integration tests, this was never above 120m. ExitOnOutOfMemoryError is to help things crash nicer vs have a neatly healthchecked, but broken container.

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ElasticsearchStorageExtension.java
```diff
@@ -36,10 +35,16 @@
 
 class ElasticsearchStorageExtension implements BeforeAllCallback, AfterAllCallback {
   static final Logger LOGGER = LoggerFactory.getLogger(ElasticsearchStorageExtension.class);
+
+  static {
+    // Gives better context when there's an exception such as AbortedStreamException
+    System.setProperty("com.linecorp.armeria.verboseExceptions", "always");
```
codefromthecrypt, line 41: this has so far not helped, but maybe it will here?

codefromthecrypt (commented):
notes

codefromthecrypt:
ES 6:

healthcheck works
```
072369f98f89        openzipkin/zipkin-elasticsearch6:test   "start-elasticsearch"   13 seconds ago      Up 12 seconds (healthy)   9200/tcp            dreamy_chandrasekhar
```

about 108M (was 223MB) above our base layer
```
d16ba2c42e5b        5 minutes ago       /bin/sh -c #(nop) COPY --chown=elasticsearch…   108MB               
```

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ElasticsearchStorageExtension.java
anuraaga: Dunno if this static block runs early enough for Armeria to pick it up, armeria reads the property only once at initialization

anuraaga (approved):
Nice improvement

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ElasticsearchStorageExtension.java
codefromthecrypt: yeah I was hoping to not rely on build config, so will move it up farther if this one doesn't work :P

codefromthecrypt: @minwoox @ikhoon @trustin ps any way we can do this without system properties?

minwoox: I think using system properties is the only way to change the value. 🤔

anuraaga: I think the easy verification is to add a debug statement and see if Armeria already logged it's flags-related stuff, if so too late. If not, it should work :)

Agree with the sentiment of not mucking with build, but for five lines maybe it's still worth not worrying about ordering issues

http://maven.apache.org/surefire/maven-surefire-plugin/examples/system-properties.html
```
<plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-surefire-plugin</artifactId>
        <version>3.0.0-M5</version>
        <configuration>
          <systemPropertyVariables>
            <propertyName>propertyValue</propertyName>
```

codefromthecrypt:
ES 7

```
CONTAINER ID        IMAGE                                   COMMAND                 CREATED             STATUS                   PORTS               NAMES
420bface9369        openzipkin/zipkin-elasticsearch7:test   "start-elasticsearch"   9 seconds ago       Up 8 seconds (healthy)   9200/tcp            gracious_wright
```

about 118MB (was 466MB) above our base layer
```
f310bfc14143        2 hours ago         /bin/sh -c #(nop) COPY --chown=elasticsearch…   118MB               
```

codefromthecrypt:
thanks for the look @anuraaga!

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ElasticsearchStorageExtension.java
codefromthecrypt: it is also that I often run in IDE and it wouldn't catch those lines, but then again it never fails in the IDE..

codefromthecrypt: 5eec6f3e6

Issue #3197, opened 2020-09-04
Flakey Test: ITElasticsearchHealthCheck.oneHealthy


codefromthecrypt:
```
[ERROR] zipkin2.server.internal.elasticsearch.ITElasticsearchHealthCheck.oneHealthy  Time elapsed: 0.078 s  <<< FAILURE!

org.opentest4j.AssertionFailedError: 

Expecting:

 <false>

to be equal to:

 <true>

but was not.

	at zipkin2.server.internal.elasticsearch.ITElasticsearchHealthCheck.oneHealthy(ITElasticsearchHealthCheck.java:106)
```

codefromthecrypt:
I added info to the assertion on fail. there are similar flakes recently in actual ES tests, with an AbortedStreamException also. I didn't notice these until 0.99.9 or 1.0.0 I forget... either these will go away magically or we will have to perhaps use tracing inside the IT to print a trace when there was an exception and/or use the `-Dcom.linecorp.armeria.verboseExceptions=always` or https://javadoc.io/doc/com.linecorp.armeria/armeria-javadoc/latest/com/linecorp/armeria/common/util/Sampler.html#of(java.lang.String)

thx for hints @trustin

codefromthecrypt:
Will apply the things as master build is erring

```
[ERROR] zipkin2.elasticsearch.integration.ITElasticsearchStorage$ITServiceAndSpanNames.getRemoteServiceNames  Time elapsed: 0.035 s  <<< ERROR!

com.linecorp.armeria.common.stream.AbortedStreamException
```

codefromthecrypt:
hmm even after changing to use system property, there's no causal chain here:

```
java.lang.AssertionError: Health check failed with message: null

	at zipkin2.server.internal.elasticsearch.ITElasticsearchHealthCheck.assertOk(ITElasticsearchHealthCheck.java:203)

	at zipkin2.server.internal.elasticsearch.ITElasticsearchHealthCheck.healthyThenNotHealthyThenHealthy(ITElasticsearchHealthCheck.java:141)

Caused by: com.linecorp.armeria.common.stream.AbortedStreamException
```
https://travis-ci.org/github/openzipkin/zipkin/builds/728217084

cc @ikhoon @minwoox @trustin

codefromthecrypt:
[...]

PR #3210, opened 2020-09-18
Ports MySQL test image to our conventions

codefromthecrypt:
Particularly, this image adds HEALTHCHECK and deletes 100MB of binaries not used at runtime, making it 170MB (compared to the official mariadb image which is 400MB)

Once released, this completes our storage images: we can add a health condition on any storage image or kafka:

Ex.
```
    depends_on:
      storage:
        condition: service_healthy
      kafka-zookeeper:
        condition: service_healthy
```

### DIFF
--- a/docker/storage/mysql/install.sh
+++ b/docker/storage/mysql/install.sh
@@ -35,7 +35,7 @@ mysqld ${MYSQL_OPTS} &
 
 # Excessively long timeout to avoid having to create an ENV variable, decide its name, etc.
 timeout=180
-echo "Will wait up to ${timeout} seconds for Cassandra to come up before installing Schema"
+echo "Will wait up to ${timeout} seconds for MySQL to come up before installing Schema"
 while [ "$timeout" -gt 0 ] && ! mysql --protocol=socket -uroot -e 'SELECT 1' > /dev/null 2>&1; do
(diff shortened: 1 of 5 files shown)


docker/storage/mysql/install.sh
```diff
@@ -19,22 +19,30 @@ echo "*** Installing MySQL"
...
+mysqld ${MYSQL_OPTS} &
 
-timeout=300
-while [[ "$timeout" -gt 0 ]] && ! mysql --user=mysql --protocol=socket -uroot -e 'SELECT 1' >/dev/null 2>/dev/null; do
-    echo "Waiting ${timeout} seconds for mysql to come up"
-    sleep 2
-    timeout=$(($timeout - 2))
+# Excessively long timeout to avoid having to create an ENV variable, decide its name, etc.
+timeout=180
+echo "Will wait up to ${timeout} seconds for Cassandra to come up before installing Schema"
```
codefromthecrypt:
```suggestion
echo "Will wait up to ${timeout} seconds for MySQL to come up before installing Schema"
```

anuraaga (approved):

codefromthecrypt:
flakes are the aborted stream exception #3197

Issue #3192, opened 2020-08-31
Make span name count by service in the UI semantic to local roots


codefromthecrypt:
**Feature:**
span count by service should only return remote links based on instrumented services (not remote/virtual ones). This is also the same as local root count, although we don't have a local root ID yet in our model.

With local root instead, we get the same count as using the dependency graph, except this is per trace, and currently excluding remote endpoints. When someone sees a high count, they can do a search for that service name, knowing it is a serviceName not a remoteServiceName!

**Implementation**
While I didn't implement it, fixing span count is using same logic as dependency linker on the spannode representing the root. instead of generate links (or after doing that), increase count only for each side corresponding to the localEndpoint.serviceName and span.kind

Ex. if links is span.kind=SERVER [{parent=foo child=bar},{parent=bar child=baz},{parent=bar child=baz}] we count the following in this trace:

foo=1
bar=2

iff foo and bar are localEndpoint.serviceName

There's some tricky when the links are messaging. This type of link is span.kind=CONSUMER would be {parent=broker, child=consumerService}, but a local root of the child, not the parent!

To summarize, the name to choose from the link will correspond to the localEndpoint.serviceName. In SERVER links this ends up the parent, but in CONSUMER, it is the child service

**Rational**
This allows us to avoid skewing the brain when buggy or fine grained instrumentation results inflates the count for a specific node. This is particularly important when we plot numbers after https://github.com/openzipkin/zipkin/pull/3180

Here are some examples that skew "span count"
* profiler style tracing could show a service only hit once having an order of magnitude more "spans"
* an edge service that calls MySQL 2000 times (yet MySQL is not a "service"!)
* buggy instrumentation that creates client spans when they meant to create local spans

**Example Scenario**
https://github.com/openzipkin/zipkin/pull/3180 will plot these numbers anyway, doing this will make the numbers semantically interpretable

**Prior Art**
dependency linker has most of this logic, though it is unaware of which are instrumented services, and it is also unaware of the span.kind of the link. This means you need a little more info to decide. Ex we may need to carry link.kind or just implement different than dependency linker

codefromthecrypt:
@tacigar I will be busy next couple days at least, possibly all week. I made a design here to correct the counts. Let me know if you have cycles to work on it, as I can definitely review even if I may not have cycles to perform the change.

codefromthecrypt:
[...]

PR #3213, opened 2020-09-20
[Feedback Needed] Show Pie charts and tables instead of badges for span counts

tacigar:
in #3180, we had a discussion about how to display span counts.
At that time I created a chart with the following design, but we found some problems for usability.

**rejected design**:
OLD CHART

So I created a new design with pie chart and table like the following:

**new design**:
スクリーンショット 2020-09-20 17 25 14

What do you think about this? Please give me your feedback.

cc @adriancole @shakuzen

### DIFF
--- a/zipkin-lens/package-lock.json
+++ b/zipkin-lens/package-lock.json
@@ -7071,6 +7071,30 @@
         "safer-buffer": "^2.1.0"
       }
     },
+    "echarts": {
+      "version": "4.9.0",
+      "resolved": "https://registry.npmjs.org/echarts/-/echarts-4.9.0.tgz",
+      "integrity": "sha512-+ugizgtJ+KmsJyyDPxaw2Br5FqzuBnyOWwcxPKO6y0gc5caYcfnEUIlNStx02necw8jmKmTafmpHhGo4XDtEIA==",
+      "requires": {
+        "zrender": "4.3.2"
+      }
+    },
+    "echarts-for-react": {
+      "version": "2.0.16",
+      "resolved": "https://registry.npmjs.org/echarts-for-react/-/echarts-for-react-2.0.16.tgz",
+      "integrity": "sha512-VmHCktay2qKt/+wpL/C7thbvIa7dYBEey0/U4Zaqo+qeA4wx+uiCd5NeCsPIhD/0Pv+2qqNswqiNiUCtcgccOw==",
+      "requires": {
+        "fast-deep-equal": "^2.0.1",
+        "size-sensor": "^1.0.0"
+      },
+      "dependencies": {
+        "fast-deep-equal": {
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
note: I still owe a change to make those numbers better (#3192).. we should be counting entry spans (local roots) not ad-hoc spans. That said, the UX wouldn't change

shakuzen:
Thanks for the continued effort and design work here, @tacigar. Great stuff! In a lot of ways, I like this much more. I worry, though, if we don't make it clear that the pie chart is showing span count, users will think it is a representation of relative duration. Even if we tried to make it clear, I worry if users will still jump to that conclusion. Likewise, the table with the span count almost aligns with the duration column from the search results, which again might lead people to think it is a duration value. I wonder if we need both the pie chart and the table, since the names of the services are clearly visible on the pie chart. Depending on how important seeing the exact numbers are, maybe just a pie chart is enough?

codefromthecrypt:
I'll also note that in design cases we aren't using a trace with a large amount of service names. The earlier design (bucketing by first letter) could equally be done in pie form, as could bars without bucketing by service name. I think we are still in need of a representative trace that has a lot of service names in it. @mrajah-twttr would you be up to contributing a realistic trace with a larger amount of service names in it?

mrajah-twttr:
typically we have large number of spans and not large number of services.. how many services are you looking at

codefromthecrypt:
> typically we have large number of spans and not large number of services.. how many services are you looking at

any trace with more than 20 service names would cause some design impact I think

mrajah-twttr:
i would not be able to give the trace without a bunch of manual scrubbing but i can try it out and post screenshot. stay tuned.

mrajah-twttr:
This looks pretty cool. 
image

Couple of points. 

1. Would be good to know the distinct service counts somewhere
2. When clicking on anyone color would help to hide the rest of the labels else its too cluttered sometimes (is it better to show names only for big pieces and show names for others only on hover?)
3. When selecting a particular service would help to highlight/scroll it on the right side table or move it to top.

codefromthecrypt:
It has been a long time since we cut a version and we need to..

We should decide soon if what's in master is ok, if this should be included, or if something should be reverted.

Those using the UI, do you have any concerns if master shipped as-is, or if this change as-is was merged?

jorgheymans:
For me master can ship as is, and we should not yet include this.

llinder:
I'm going to pull down this code to check out the UI changes later this evening.  I don't want to hold up decisions to release or not but if folks want I wouldn't mind setting up a 15 minute demo internally at work and gathering some additional input to share here.

codefromthecrypt:
thanks for the offer @llinder!

PR #3215, opened 2020-09-25
TypeScriptize TraceJsonUploader and TraceIdSearch

tacigar:


zipkin-lens/src/translations/en/messages.json
```diff
@@ -42,7 +42,6 @@
   "Zipkin Home": "",
   "hide annotations": "",
   "show all annotations": "",
-  "trace id...": "",
```
codefromthecrypt, line 45: I assume this is orphaned

codefromthecrypt (commented):
Can you mention motivation for this change? It seems like reduction of code is the intent.

Also, is there any visual difference as a result?

anuraaga (approved):
Probably mostly for consistent tooling / code. Didn't see any differences when reviewing seems nice.

codefromthecrypt:
Thanks!

PR #3216, opened 2020-09-27
Allow use of JDK 15+

codefromthecrypt:
This allows builds to work with JDK 15+ even if that JDK cannot compile
Java 1.6 anymore. Specifically, this removes Retrolambda, which formerly
rewrote our main tree to 1.6 bytecode. We had to do this as Retrolambda
no longer works in JDK 15.

In order to publish a JRE compatible 1.6 jar to Maven Central, we use
JDK during the release process (Travis). To ensure normal usage and
Docker can use JDK 15, we change our GitHub actions to use JDK 15.

This also adds a RATIONALE.md which covers the major aspects around
source, largely duplicated from Brave (though not exactly the same as
context slightly differs here.)


zipkin/src/main/java/zipkin2/storage/InMemoryStorage.java
```diff
@@ -77,15 +77,12 @@ public static Builder newBuilder() {
     int maxSpanCount = 500000;
     List<String> autocompleteKeys = Collections.emptyList();
 
-    /** {@inheritDoc} */
-    @Override
-    public Builder strictTraceId(boolean strictTraceId) {
+    @Override public Builder strictTraceId(boolean strictTraceId) {
```
anuraaga, line 80: This strategy seems good. Just wondering do you think this is a good time to move stuff out of core? For example, I guess storage, since it's for use by server and not client, doesn't need to be Java 6?

anuraaga: IIRC we might have some senders that delegate to storage classes (thinking zipkin-aws). I think it could be a net win to duplicate for this rare case and mark storage as server-only. It's the only way to get to `CompletableFuture` for example.

codefromthecrypt: yeah we currently only support java 1.6 on the core library. the storage classes are already 1.8 iirc. that zipkin-aws xray sender internally uses storage is something that we aren't optimizing for in a bytecode level..

typically, senders raise to the level of the library used, sometimes 1.7 (okhttp) or 1.8 (kafka)

codefromthecrypt:
ok ready!

zipkin/src/main/java/zipkin2/storage/InMemoryStorage.java
codefromthecrypt: oh I see what you are saying.. to move these classes out? I think it is a non-transparent change to do that where as the rest of this is transparent to end users. I think we could consider doing that separately, but there would be some transition work involved and so not entirely keen to take that on right now.

codefromthecrypt: https://github.com/openzipkin/zipkin/issues/3217

anuraaga (approved):

codefromthecrypt:
same elasticsearch flakes as usual

Issue #3218, opened 2020-09-30
Migrate our docker image consistently to alpine


codefromthecrypt:
openzipkin/zipkin-builder image is running a different distro, which is ok but it is a lot larger to download than necessary. One thing blocking this is https://github.com/eirslett/frontend-maven-plugin/pull/853 as we can't execute the build when node isn't working (required to be what's downloaded from the plugin.)

codefromthecrypt:
@anuraaga @zeagord @eirslett anyone with some time to help on this one? it would be neat and clean and quicker downloads

codefromthecrypt:
I think I got a workaround for now. Would certainly prefer the plugin to allow unofficial, but hey progress not perfection

```sh
#!/bin/sh
# This script downloads the nodejs archive into the Maven local repository so that
# frontend-maven-plugin can work on Alpine/musl. It mostly plays tricks to get rid
# of '-musl' in both the archive name and its contents.
#
# See https://github.com/eirslett/frontend-maven-plugin/pull/853
set -eux

# Get the version of node our build wants.
NODE_VERSION=$(mvn -pl zipkin-lens help:evaluate -Dexpression=node.version -q -DforceStdout)

# Get a local path corresponding to the Maven artifact the frontend-maven-plugin expects
NODE_MAVEN_PATH=~/.m2/repository/com/github/eirslett/node/${NODE_VERSION}/node-${NODE_VERSION}-linux-x64.tar.gz

if [ ! -f "${NODE_MAVEN_PATH}" ]; then
  # Get the URL of the unofficial build of musl
  NODE_DOWNLOAD_URL=https://unofficial-builds.nodejs.org/download/release/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-x64-musl.tar.gz
  echo "*** Downloading nodejs into the Maven local repository"

  # Strip '-musl' from the original archive's file paths
  mkdir /tmp/$$ && cd /tmp/$$
  wget -qO- ${NODE_DOWNLOAD_URL}| tar --transform 's/-musl//' -xz

  # Create a new archive where Maven would expect it
  mkdir -p $(dirname "${NODE_MAVEN_PATH}")
  tar -czf ${NODE_MAVEN_PATH} *
fi
```

codefromthecrypt:
https://github.com/eirslett/frontend-maven-plugin/pull/934 should make it nicer

PR #3219, opened 2020-09-30
Migrates builder to Alpine JDK 15

codefromthecrypt:
This makes things easier once #3218 is addressed

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -36,22 +36,16 @@ RUN mkdir -p /zipkin && cp zipkin-server/target/zipkin-server-*-exec.jar /zipkin
 RUN mkdir -p /zipkin-slim && cp zipkin-server/target/zipkin-server-*-slim.jar /zipkin-slim && cd /zipkin-slim && jar xf *.jar && rm *.jar
 
 # docker/hooks/post_push will republish what is otherwise built in docker/build/Dockerfile
-FROM azul/zulu-openjdk-alpine:15.0.0-15.27.17
+# This is a copy/paste from builder/Dockerfile:
+#
+# zipkin-builder is JDK + artifact caches because DockerHub doesn't support another way to update
+# cache between builds.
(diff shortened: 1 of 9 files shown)


docker/Dockerfile
```diff
@@ -35,11 +35,6 @@ RUN ./mvnw -q --batch-mode -DskipTests --also-make -pl zipkin-server install
 RUN mkdir -p /zipkin && cp zipkin-server/target/zipkin-server-*-exec.jar /zipkin && cd /zipkin && jar xf *.jar && rm *.jar
 RUN mkdir -p /zipkin-slim && cp zipkin-server/target/zipkin-server-*-slim.jar /zipkin-slim && cd /zipkin-slim && jar xf *.jar && rm *.jar
 
-FROM maven:3-openjdk-14-slim as zipkin-builder
```
codefromthecrypt, line 39: I couldn't see where this layer was used..

anuraaga: It's used here in a hook

https://github.com/openzipkin/zipkin/blob/206abbc26c333f974e09a08063995f4a23af0a39/docker/hooks/post_push#L38

It's the only way to have these cache across builds with docker hub that I could find.

codefromthecrypt: ic, I was actually wondering why we are explicitly caching in the image when we can do stuff like that in travis and github actions..

something to think about, and don't have a better plan at the moment. it is a pity to have copy/pasta..

codefromthecrypt:
Will leave this open a bit. It would be more coherent to depend on openzipkin/jdk which includes an alpine based JDK plus maven. that way "zipkin-builder" which is cache in nature can keep the same process with no more copy/pasta.

anuraaga (approved):
Cool

codefromthecrypt:
I will merge once local build finishes.

The first time I need to push the builder update to dockerhub kindof exactly when this merges :D

codefromthecrypt:
ok watching master build...

PR #3222, opened 2020-10-02
Reduces the size of intermediate layers when building images

codefromthecrypt:
According to `docker system df`, this change results in 1.2GB of image
size, prunable to 800MB after building all our base images (not Kafka or
storage).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,5 @@
 **/.*
 !.dockerignore
-!.git
 !.mvn
 !docker/zipkin/**
 !**/.eslintrc*
@@ -13,6 +12,7 @@
 **/node_modules
 **/*.iml
 
+**/.idea
 **/*.md
 
 **/Dockerfile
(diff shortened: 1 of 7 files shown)


.dockerignore
```diff
@@ -1,6 +1,5 @@
 **/.*
 !.dockerignore
-!.git
```
codefromthecrypt, line 3: this was 77 more MB of context, which isn't needed unless we run license check

zipkin-server/README.md
```diff
@@ -605,7 +605,7 @@ See [docker-zipkin](https://github.com/openzipkin/docker-zipkin) for details.
 To build and run the server from the currently checked out source, enter the following.
 ```bash
-$ ./mvnw -q --batch-mode -DskipTests --also-make -pl zipkin-server clean install
+$ ./mvnw -q --batch-mode -DskipTests --also-make -pl zipkin-server clean package
```
```suggestion
$ ./mvnw -q --batch-mode -DskipTests -Dlicense.skip=true --also-make -pl zipkin-server clean package
```

anuraaga (commented):
LGTM

PR #3223, opened 2020-10-03
Uses installed Maven to reduce image layer size

codefromthecrypt:
We can save 21MB of downloading and copying by not using the Maven
wrapper when in Docker, as we control the OS and Maven version
explicity here.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -29,8 +29,9 @@ COPY --from=scratch /code /code
 
 WORKDIR /code
 
-# Use the same command as we suggest in zipkin-server/README.md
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -29,8 +29,9 @@ COPY --from=scratch /code /code
 
 WORKDIR /code
 
-# Use the same command as we suggest in zipkin-server/README.md
-RUN ./mvnw -q --batch-mode -DskipTests -Dlicense.skip=true --also-make -pl zipkin-server package && \
+# Use the same command as we suggest in zipkin-server/README.md to hydrate the Maven and NPM cache
```
codefromthecrypt:
```suggestion
# Use the same command as we suggest in zipkin-server/README.md
```

anuraaga (approved):

PR #3226, opened 2020-10-04
Removes Alpine workaround for maven-frontend-plugin

### LINKED ISSUE #3218: Migrate our docker image consistently to alpine

codefromthecrypt:
per https://github.com/eirslett/frontend-maven-plugin/pull/934#issuecomment-703310430

Fixes #3218


anuraaga (approved):

codefromthecrypt:
one day I will look into the ES client flake deeply.. #3197

PR #3228, opened 2020-10-05
Add postgres storage

mrk-andreev:
- Add postgres storage based on mysql implementation
- Remove indexes for write speed needs

Issue: https://github.com/openzipkin/zipkin/issues/1232


codefromthecrypt:
Thanks for giving this a try, though probably could have saved you effort to [chat](https://gitter.im/openzipkin/zipkin) first.

TL'DR'  a straight port of the MySQL v1 schema won't be accepted in the main project.

It is enough burden to maintain that, and even that should be ported to a more current design if we had more time. Making a similar copy in a different database flavor adds to the tech debt burden while still not giving a reasonable schema to end users.

Some notes copied here for convenience on the topic:
> work here needs champions to run it through. If you like, you can look at
the work-on-progress voltdb experiment as an example, though in the schema
likely completing normalization of the data is preferred (unclear on this
really)
> https://github.com/adriancole/zipkin-voltdb

https://github.com/openzipkin/zipkin/issues/1232#issuecomment-462544544


>> Came across this :
>>
>>  https://github.com/IAMTJW/zipkin/tree/master/zipkin-storage/postgresql
>> 
>> Any chance of this getting merged(accepted) ?
>
> first, the author would need to volunteer to contribute the code.
> technically though, this looks like a port of the mysql code, which isn't a
> good schema to emulate. so, in its form I dont expect that specific code to
> be merged.
https://github.com/openzipkin/zipkin/issues/1232#issuecomment-467460732

mrk-andreev:
Anyway it was my needs :)

PR #3229, opened 2020-10-06
Makes docker-builder .git size agnostic

codefromthecrypt:
Thanks to @anuraaga for the nudge. By using another layer in the
builder dockerfile, we can avoid explaining how shallow clones
result in a smaller .git directory.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -12,6 +12,8 @@
 # the License.
 #
 
+ARG java_version=15.0.0-15.27.17
+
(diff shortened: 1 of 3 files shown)


anuraaga (approved):

codefromthecrypt:
On a non-shallow clone, this reduces the image size by a lot:

before
```
openzipkin/zipkin-builder   latest              29c52da94548        21 minutes ago      649MB
```

after
```
openzipkin/zipkin-builder   latest              50c1ed3cea36        2 hours ago         567MB
```

PR #3231, opened 2020-10-09
Check datasource url fix

arioko:


jorgheymans (commented):
Thanks ! Can you explain a bit more, the fix seems okay but i'm wondering if this was never blocking before, or there just happened to be a dashboard named `prom` as well.

arioko:
I can't give you too much information. I have tried testing yours code and got an error that made it impossible to create the "prom" datasource. Reviewing the script, I saw that it was trying to create the datasource if it did not exist, but that it was checking if there was a dashboard "prom" instead of a datasource "prom". I have not created any dashboard called "prom" at any time.

codefromthecrypt:
I don't think this is needed. If you try from docker/examples directory `docker-compose up` then go to http://localhost:3000 you can see the dashboard loaded fine. Note: this intends to load the dashboard the zipkin team created.

I just tried it with some example

Screen Shot 2020-10-13 at 7 26 24 PM
 traffic.

PR #3232, opened 2020-10-11
silence logging creep

jorgheymans:


jorgheymans:
For reference these lines were silenced:

image

codefromthecrypt:
thanks! there might be a different drift that caused this. I think we intended to disable the doc service https://github.com/openzipkin/zipkin/blob/master/zipkin-server/src/main/resources/zipkin-server-shared.yml#L271

maybe there's an auto-configuration (or property) we can disable and revert the log exclusion? cc @minwoox

jorgheymans:
`DocService` can be disabled selectively, will have a look at this later.

https://armeria.dev/docs/server-docservice#including-and-excluding-service-methods

jorgheymans:
In fact, `DocService` just logs the plugins it has scanned at loadtime, doesn't mean it is actually active https://github.com/line/armeria/blob/a0180359e92fb80101dc3d04e178db845077afc3/core/src/main/java/com/linecorp/armeria/server/docs/DocService.java#L100

Not much that can be done, except asking Armeria if that logging line is really needed ?

codefromthecrypt:
since they are already pinged here, I'll leave it. thanks for the close look

minwoox:
Sorry, I was leaving for my vacation. 😄 
Yeah, we recently added logging for SPI implementations because there was a case a user got in trouble with the SPI implementation in Fat JAR. 
Sorry about the inconvenience. 😅

codefromthecrypt:
@minwoox maybe this can be debug? usually lower level things like this aren't worth bothering everyone about

minwoox:
We usually use info level for the logs in static initializers but let me discuss with others. cc @trustin, @ikhoon

trustin:
Armeria logs the basic information required for understanding how it's configured at INFO level. I don't think we're going to lower to DEBUG because we log more dev-centric stuff at DEBUG level already. How about just setting the log level to WARN for `com.linecorp.armeria`?

codefromthecrypt:
do you consider the loading of a serialization provider "basic information required for understanding"?

it seems misplaced imho, meanwhile we've already silenced the logger or @jorgheymans did heh

trustin:
Ah yeah, we could log them at DEBUG level but I guess we still should log what modules were loaded, e.g. `Available modules: brave, grpc, ...`. /cc @minwoox

minwoox:
Yeah, sounds good to me too. 😄

PR #3234, opened 2020-10-14
Fix a bug in the unit of duration

### LINKED ISSUE #3233: total duration in overview wrongly says 'ms'
in the overview the total duration for a trace is shown as 15.714ms, but the total trace duration is 15.714s

image

image

tacigar:
resolve: https://github.com/openzipkin/zipkin/issues/3233

`formatDuration` requires a microsecond-value as its argument, but `trace.duration` is millisecond-value.
So we need to multiply `trace.duration` by 1000.

### DIFF
--- a/zipkin-lens/src/components/DiscoverPage/TraceSummaryRow.test.js
+++ b/zipkin-lens/src/components/DiscoverPage/TraceSummaryRow.test.js
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2015-2020 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+import React from 'react';
+
+import TraceSummaryRow from './TraceSummaryRow';
+import render from '../../test/util/render-with-default-settings';
+
+describe('<TraceSummaryRow />', () => {
+  it('should render timestamp and duration in correct unit', () => {
+    const { queryByTestId } = render(
+      <TraceSummaryRow
+        traceSummary={{
+          traceId: 'a03ee8fff1dcd9b9',
+          timestamp: 1571896375237354,
+          duration: 131848,
+          serviceSummaries: [],
+          spanCount: 10,
+          width: 10,
+          root: {
+            serviceName: 'routing',
+            spanName: 'post /location/update/v4',
+          },
+        }}
+      />,
+    );
+
+    const startTimeFormat = queryByTestId('TraceSummaryRow-startTimeFormat');
+    expect(startTimeFormat).toBeInTheDocument();
+    expect(startTimeFormat).toHaveTextContent('10/24 13:52:55:237');
+    // intentionally not asserting the relative time from now as it would drift tests
+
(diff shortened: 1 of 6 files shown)


codefromthecrypt (approved):
any test case :)

codefromthecrypt:
I will try to make a test

codefromthecrypt:
@tacigar I sketched a test.. feel free to make it nicer.

I notice this in the test output. maybe you recognize if this is a problem?

```
      Warning: validateDOMNesting(...): <tr> cannot appear as a child of <div>.
          in tr (created by ForwardRef(TableRow))
          in ForwardRef(TableRow) (created by WithStyles(ForwardRef(TableRow)))
          in WithStyles(ForwardRef(TableRow)) (at TraceSummaryRow.tsx:118)
          in TraceSummaryRow (at TraceSummaryRow.test.js:22)
          in ThemeProvider (at render-with-default-settings.tsx:73)
          in ThemeProvider (at render-with-default-settings.tsx:72)
          in Unknown (at render-with-default-settings.tsx:71)
          in Router (at render-with-default-settings.tsx:70)
          in I18nProvider (at render-with-default-settings.tsx:69)
          in Provider (at render-with-default-settings.tsx:68)
          in wrapper
```

zipkin-lens/src/zipkin/trace.js
```diff
@@ -171,7 +171,7 @@ export function traceSummaries(serviceName, summaries, utc = false) {
           (parseFloat(duration) / parseFloat(maxDuration)) * 100,
           10,
         );
-        res.duration = duration / 1000; // used only for client-side sort
```
codefromthecrypt, line 174: I fixed here instead!

zipkin-lens/src/models/TraceSummary.ts
```diff
@@ -20,6 +20,7 @@ export type ServiceNameAndSpanCount = {
 type TraceSummary = {
   traceId: string;
   timestamp: number;
+  // Unlike most places, this is milliseconds, not microseconds
```
codefromthecrypt:
```suggestion
```

tacigar:
> I notice this in the test output. maybe you recognize if this is a problem?

This problem occurs because `TableSummaryRow` (= `<tr>`) is not wrapped by `<table>` and `<tbody>` in the unit test.
So you can remove this problem by wrapping `TableSummaryRow` in `<table>` and `<tbody>` as follows in the unit test 👍 

```javascript
    const { queryByTestId } = render(
      <table>
        <tbody>
          <TraceSummaryRow
            traceSummary={{
              traceId: 'a03ee8fff1dcd9b9',
              timestamp: 1571896375237354,
              duration: 131848,
              serviceSummaries: [],
              spanCount: 10,
              width: 10,
              root: {
                serviceName: 'routing',
                spanName: 'post /location/update/v4',
              },
            }}
          />
        </tbody>
      </table>,
    );
```

zipkin-lens/src/zipkin/trace.js
tacigar: 👍

codefromthecrypt:
thanks @tacigar I think the test is ok now. let's see

zipkin-lens/src/components/DiscoverPage/TraceSummaryRow.test.js
```diff
@@ -0,0 +1,52 @@
...
+            }}
+          />
+        </tbody>
+      </table>,
+    );
+
+    const startTimeFormat = queryByTestId('TraceSummaryRow-startTimeFormat');
+    expect(startTimeFormat).toBeInTheDocument();
+    // Don't assert on hour as the timezone will be different in CI
+    expect(startTimeFormat).toHaveTextContent(/10\/2[34] [0-9][0-9]:52:55:237/);
```
codefromthecrypt, line 45: PS main reason I'm doing this at all is to make sure we are using the right unit in the raw date input. If we weren't, we'd not end up in the same month :D

PR #3239, opened 2020-10-15
disallow trace requests

### LINKED ISSUE #2286: Zipkin - Track and Trace method vulnerability
Hi all,

My company's security dept has notified me of a vulnerability in on one of our hosts. Specifically, it's on port 9411: our Zipkin container.

The indicated vulnerability is this one: [https://www.tenable.com/plugins/nessus/11213]

So, I have these 2 doubts:
- Is Zipkin functionality dependent on these 2 HTTP methods or can they be disabled?
- And if not, how do I disable those 2 HTTP methods? As Zipkin is a Spring Boot application, I don't know if this is externally configurable.

Thanks!

jorgheymans:
fixes #2286 (but mainly an excuse to dabble in Armeria a bit)


jorgheymans:
(also grateful if anyone can tell me how these first 8 commits ended up in my branch, it's not the first time and my local fork is `reset --hard` already )

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java
```diff
@@ -202,6 +204,12 @@ public AggregatedHttpResponse getAutocompleteValues(
     return maybeCacheNames(values.size() > 3, values, ctx.alloc());
   }
 
+  @Trace("regex:^.*$")
```
jorgheymans: couldn't find a global toggle to disallow it, hence this workaround. @anuraaga does this please you ?

codefromthecrypt: @ikhoon @minwoox ^^

minwoox: How about using route?
```java
in ZipkinHttpConfiguration.java

sb.route()
  .methods(HttpMethod.TRACE)
  .pathPrefix("/")
  .build((ctx, req) -> HttpResponse.of(HttpStatus.FORBIDDEN));
```

minwoox: @jorgheymans Created an issue for that. 😄 
https://github.com/line/armeria/issues/3114

ikhoon: I think a decorator would be better to forbid TRACE requests because it could have higher priority than HTTP services. 
```java
sb.routeDecorator()
  .methods(HttpMethod.TRACE)
  .pathPrefix("/")
  .build((delegate, ctx, req) -> HttpResponse.of(HttpStatus.FORBIDDEN));              
```

jorgheymans:
Using `routeDecorator` now which is more idiomatic Armeria code I agree.

zipkin-server/src/main/java/zipkin2/server/internal/ZipkinQueryApiV2.java
jorgheymans: thanks for the suggestions, i added the routeDecorator.

codefromthecrypt:
Thanks for taking this on @jorgheymans and thanks for the advice team Armeria!

codefromthecrypt:
Sorry. the flakey test had me merging on red, but this actually broke ITZipkinServerCORS.

I don't know why, but the decorator is being applied to OPTIONS methods, not just TRACE.

codefromthecrypt:
https://github.com/openzipkin/zipkin/pull/3245 to workaround until we find out why TRACE is also getting OPTIONS

minwoox:
@adriancole and @jorgheymans Sorry about the bug. 😅 
If should be fixed by https://github.com/line/armeria/pull/3120

PR #3236, opened 2020-10-15
elasticsearch: use backpressure to resolve streaming request flakes

minwoox:
The `HttpResponse` was sent before the request are fully sent.
So the request was aborted after getting the response. This changes
to send the response after the request is fully received.

Fixes #3197

PR #3243, opened 2020-10-17
Refactors Cassandra queries so they are cheaper and easier to migrate

codefromthecrypt:
Before, we used labels a lot, and generally a lot of query builder code.
This switches most to positional syntax, which is cheaper and easier to
migrate to cassandra driver 4.

This also reformats everything to make the migration feel like less
change.

Finally, this consolidates as much code as possible, notably repetitive
inserts and queries, again to reduce the heft needed to upgrade later.


codefromthecrypt:
did my best to axe a thousand lines

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/SelectRemoteServiceNames.java
```diff
@@ -17,33 +17,35 @@
...
     Factory(Session session) {
       this.session = session;
-      this.preparedStatement = session.prepare(QueryBuilder.select("remote_service_name")
-        .from(Tables.REMOTE_SERVICE_NAMES)
-        .where(QueryBuilder.eq("service_name", QueryBuilder.bindMarker("service_name")))
-        .limit(QueryBuilder.bindMarker("limit_")));
+      this.preparedStatement =
+        session.prepare(select("remote_service_name").from(REMOTE_SERVICE_NAMES)
+          .where(eq("service_name", bindMarker()))
+          .limit(1000));
```
codefromthecrypt, line 42: you'll notice constants like this around. we had assigned them via variables before, this is just a bit more obvious.

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CassandraSpanStore.java
```diff
@@ -67,8 +64,6 @@
     indexFetchMultiplier = storage.indexFetchMultiplier;
     strictTraceId = storage.strictTraceId;
     searchEnabled = storage.searchEnabled;
-    timestampCodec = new TimestampCodec(metadata.protocolVersion);
-    buckets = IntStream.range(0, storage.bucketCount).boxed().collect(Collectors.toSet());
```
codefromthecrypt, line 71: both of these complicated sharing and testing for things that were constant in practice. The timestamp codec has not difference based on version, and bucketCount was always set to 10 with no means to override it ever. since this is a legacy codebase I think being more obvious about constants is better than looking as if we had really parameterized them.

codefromthecrypt (commented):
couple notes

zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/internal/CassandraStorageBuilder.java
```diff
@@ -0,0 +1,166 @@
...
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import zipkin2.internal.Nullable;
+import zipkin2.storage.QueryRequest;
+import zipkin2.storage.StorageComponent;
+
+public abstract class CassandraStorageBuilder<B extends CassandraStorageBuilder<B>>
```
codefromthecrypt, line 27: this is mostly lifted from cassandra v1s (the common things)

zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/internal/SessionBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+import java.net.InetSocketAddress;
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+import zipkin2.internal.Nullable;
+
+import static zipkin2.Call.propagateIfFatal;
+
+public final class SessionBuilder {
```
codefromthecrypt, line 36: this was lifted to avoid copy/pasta between new and old cassandra. especially this code is different in v4 and annoying to do twice.

codefromthecrypt:
no tests were deleted as a part of this refactoring, though I moved some around. locally it is all green. hopefully same in CI

codefromthecrypt:
cassandra passed. the failures will be sorted by #3236

PR #3245, opened 2020-10-18
Adds workaround to missed decorator route

codefromthecrypt:
Temporary: see #3239


codefromthecrypt:
can't wait for the ES flakes to be gone..

PR #3246, opened 2020-10-18
Migrates to Datastax Driver v4

### LINKED ISSUE #1276: Get rid of zipkin3_udts keyspace
zipkin3_udts keyspace is needed because the current version of the datastax driver requires that UDT annotated classes have the same namespace as the annotation. The following solutions work:
- get rid of UDT classes and just read the UDT fields directly
- get rid of annotation parsing and write UDT codecs
- raise an issue on the datastax driver to allow configuration overrides when parsing annotations

### LINKED ISSUE #2405: Update to Datastax Cassandra Driver v4
We should probably try the beta of Datastax Cassandra Driver v4 in case there are problems around topics such as tracing, or cruft we can remove.

https://docs.datastax.com/en/developer/java-driver/4.0-rc/

cc @openzipkin/cassandra

See #952 for discussion about moving to v3

codefromthecrypt:
This migrates to the Netty-based Datastax Driver v4, which no longer
uses guava. This is very different library architecture, most obvious in
how configuration is managed. This migration tries to be as conventional
as possible while retaining defaults as close to as they were as
possible.

Special thanks to @adutra and @olim7t who helped with translation things on the https://groups.google.com/a/lists.datastax.com/g/java-driver-user list

cc @openzipkin/cassandra 

Fixes  #2405
Fixes #1276


docker/storage/cassandra/install.sh
```diff
@@ -26,7 +26,9 @@ wget -qO- $APACHE_MIRROR/cassandra/$CASSANDRA_VERSION/apache-cassandra-$CASSANDR
   --strip=1
 
 # Merge in our custom configuration
-sed -i '/enable_user_defined_functions: false/cenable_user_defined_functions: true' conf/cassandra.yaml
+for feature in enable_user_defined_functions enable_sasi_indexes; do
```
codefromthecrypt: while this enables it, heh we still get server side warnings 'SASI indexes are experimental and are not recommended for production use.'

zipkin-server/pom.xml
```diff
@@ -248,12 +235,12 @@
...
       <optional>true</optional>
     </dependency>
-    <dependency>
-      <groupId>io.zipkin.brave.cassandra</groupId>
-      <artifactId>brave-instrumentation-cassandra-driver</artifactId>
-      <version>0.10.3</version>
-      <optional>true</optional>
-    </dependency>
+<!--    <dependency>-->
+<!--      <groupId>io.zipkin.brave.cassandra</groupId>-->
```
codefromthecrypt, line 220: something else to upgrade now :)

zipkin-server/src/main/java/zipkin2/server/internal/cassandra3/ZipkinCassandra3StorageConfiguration.java
```diff
@@ -84,11 +82,11 @@ StorageComponent storage(
     }
 
     @Override public Object postProcessAfterInitialization(Object bean, String beanName) {
-      if (bean instanceof SessionFactory && beanFactory.containsBean("tracing")) {
-        SessionFactory delegate = (SessionFactory) bean;
-        Tracing tracing = beanFactory.getBean(Tracing.class);
-        return (SessionFactory) storage -> TracingSession.create(tracing, delegate.create(storage));
-      }
+      //if (bean instanceof SessionFactory && beanFactory.containsBean("tracing")) {
```
codefromthecrypt, line 85: @jeqo this is the style you can use for kafka (not the commenting out :P)

zipkin-storage/cassandra-v1/src/main/java/zipkin2/storage/cassandra/v1/CassandraUtil.java
```diff
@@ -144,10 +146,10 @@ static ByteBuffer toByteBuffer(String string) {
   }
 
   @SuppressWarnings("JdkObsolete")
-  static List<Date> getDays(long endTs, @Nullable Long lookback) {
-    List<Date> result = new ArrayList<>();
+  static List<Instant> getDays(long endTs, @Nullable Long lookback) {
+    List<Instant> result = new ArrayList<>();
```
codefromthecrypt, line 150: instant -> timestamp column

codefromthecrypt:
Special thanks to @adutra and @olim7t who helped with translation things on the https://groups.google.com/a/lists.datastax.com/g/java-driver-user list

michaelsembwever (approved):
PR looks good Adrian!

It's a bit big to go through line by line, but definitely the most of it looks good. I'm surprised you're still keeping the old cassandra storage.

codefromthecrypt:
NOTE: biggest thing here is that `CASSANDRA_LOCAL_DC` env must be set to a valid value as the driver now requires it. We fall back to Cassandra's default of "datacenter1" when unset.

codefromthecrypt:
> I'm surprised you're still keeping the old cassandra storage.

@michaelsembwever This is mainly because cassandra3 relies on SASI which is still marked experimental. We can decide to unhook from that I suppose, and just implement the same indexing style we used before (except more efficient as in v2 model now)...

codefromthecrypt:
"cassandra3" looks working so far. kinks in "cassandra" not many. we are sorely missing docker tests :P will get to that in a week across the board.

codefromthecrypt:
Unintuitively, even after trimming jars, this results in a larger server binary than when we didn't share netty. The reason is querybuilder includes a massive dep on shaded guava (2.8m). I will remove that in a later PR.

codefromthecrypt:
schema takes about 10s to do now (currently happens per test).. which is intensely slowing tests probably beyond what travis will allow.

codefromthecrypt:
yep.. "The job exceeded the maximum time limit for jobs, and has been terminated."

codefromthecrypt:
for cassandra pros, even locally it seems excessive time to install schema. I will try to re-use the schema as we can't afford a delay like this.

```
2020-10-19 19:06:50:811 [main] INFO zipkin2.storage.cassandra.Schema - Installing schema /zipkin2-schema.cql for keyspace skipsredundantindexinginatrace
2020-10-19 19:06:56:385 [main] INFO zipkin2.storage.cassandra.Schema - Installing indexes /zipkin2-schema-indexes.cql for keyspace skipsredundantindexinginatrace
2020-10-19 19:07:07:283 [main] DEBUG zipkin2.storage.cassandra.Schema - Registering endpoint and annotation UDTs to keyspace skipsredundantindexinginatrace
2020-10-19 19:07:07:285 [main] INFO zipkin2.storage.cassandra.Schema - done
```

codefromthecrypt:
Tests except ITEnsureSchema use the same keyspace now. should be a lot quicker

codefromthecrypt:
much faster, but cassandra-v1 has a few flakey tests, now. I'll look into them tommorrow

codefromthecrypt:
I have some refactoring to isolate heavy tests across the board. I'll move that to another PR.

codefromthecrypt:
I think I have things finally stable now and pulled out #3249 . once that's in I'll rebase this (removing the heft of the test infra changes)

codefromthecrypt:
yay green! Next PR will remove the querybuilder dep so that our server binary isn't larger.

codefromthecrypt:
https://github.com/openzipkin/zipkin/pull/3250 removes the querybuilder dep, but the core lib still seems to depend on a massive shaded guava lib https://groups.google.com/u/1/a/lists.datastax.com/g/java-driver-user/c/aN7NkxX_yQk

PR #3249, opened 2020-10-20
Refactors integration tests to be more isolated

codefromthecrypt:
This makes all integration tests use unique data. Notably, they no
longer use constant trace ID or service names, as this can lead to
difficulty when a keyspace is re-used. For example, service names and
trace IDs are often partitions. When debugging it is easier when test
data is isolated by a service name as it can be easily queried.

This also pulls the heaviest tests into their own classes so that they
don't overload the storage containers used by the bulk of our tests.

Finally, this fixes a few glitches in the v1 cassandra storage.


codefromthecrypt:
fwiw I don't expect anyone to go over these lines. the main thing here was rejigging tests so that cassandra can re-use a keyspace between methods, as in v4 of the driver there's a 10s overhead to install our schema. This is about the same as starting the whole container!

codefromthecrypt:
Flake is the usual ES one (after cassandra passed)

zipkin/src/main/java/zipkin2/Span.java
```diff
@@ -635,6 +635,7 @@ public static String normalizeTraceId(String traceId) {
     if (length > 32) throw new IllegalArgumentException("traceId.length > 32");
     int zeros = validateHexAndReturnZeroPrefix(traceId);
     if (zeros == length) throw new IllegalArgumentException("traceId is all zeros");
+    if (length == 15) throw new RuntimeException("WTF");
```
karkum, line 638: @adriancole can you provide us more context about this change? We were implicitly relying on the previous logic that left-padded 0s for trace ids generated by a legacy application. We have a workaround in place to fix the broken pipeline, but we wanted to hear why this explicitly checks for length of 15 (vs 14, 13 etc...)

rogierslag: @codefromthecrypt Why was this change added? Any other hex code (eg of length 14) will be nicely prepended with a leading zero, but why should the 15 throw an exception exactly?

codefromthecrypt: guess by the exception name that this was an accident especially as there's logic later to handle padding

rogierslag: I'll create a PR to remove this change then

rogierslag: https://github.com/openzipkin/zipkin/pull/3471 has been created. In case something is off with it, let's continue any discussion there.

Thanks for the quick replies!

PR #3252, opened 2020-10-22
Removes JMX dependency from our docker configuration

codefromthecrypt:
This buys back 800K plus startup time by disabling things that need JMX.
The 800K is not saved here, rather in our base docker image once this is
done.


codefromthecrypt:
@jeqo I noticed we use run-kafka-class in the zipkin-storage-kafka image to create topics, hence I added it here

anuraaga (approved):

codefromthecrypt:
For the curious, JMX has been disabled on our main server for a very long time. We use Prometheus for metrics export.

Spring activator endpoints are the unofficial way to change log categories at runtime (not JMX).

Anyway, folks who really want JMX can make own image or complain.

PR #3254, opened 2020-10-22
Deprecates Cassandra v1 schema for removal in Zipkin 2.23

codefromthecrypt:
The "cassandra3" storage type relies on Cassandra v3.11.3+. This has
been out for over 2 years. We will soon need to support v4 which is
coming out of beta soon. Maintenance on the v1 schema has been immense
and I don't think volunteers would do any more refactors of it. It is
time to deprecate and remove the v1 schema.

"cassandra3" uses UDTs and optimized SASI indexing.  While SASI is
marked experimental, it is enabled by default in recent Cassandra
versions. Some clones of Cassandra may choose to not support SASI, but
this does not mean we must maintain an ancient schema. Those who must
use a clone can disable search, which avoids SASI.

It is likely that when Cassandra 4 comes out, we will use SAI indexing
instead with some migration instructions. However, SASI as used today
has been used in production for 2 years. Even if the feature is marked
experimental, it is well practiced here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,18 +173,6 @@ This store does not require a [job to aggregate](https://github.com/openzipkin/z
 However, running the job will improve performance of dependencies
 queries.
 
-#### Cassandra
-The [Cassandra v1](zipkin-storage/cassandra-v1) component uses Cassandra
-2.2+ features, but is tested against the latest patch of Cassandra 3.11.
-
-The CQL was written in 2015, based on the original Cassandra schema from
-Twitter, and since been extended. Spans are stored as opaque thrifts,
-which means you cannot query fields in cqlsh. The schema was designed
-for scale, including manually implemented indexes to make querying
-larger data more performant.
-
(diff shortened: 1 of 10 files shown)


codefromthecrypt:
Quoting @michaelsembwever as this is following up from him :) cc @openzipkin/core 

> I would certainly favour SASI over the old schema. I would label that old schema as "broken" over "experimental" :-)
>
> And note sasi is enabled by default still in 3.11
https://github.com/apache/cassandra/blob/cassandra-3.11/conf/cassandra.yaml#L1279
> 
> I would stick with server-side indexes (i.e. SASI), switching to SAI when it becomes available.

michaelsembwever:
> It is likely that when Cassandra 4 comes out, we will use SAI indexing instead with some migration instructions.

fyi, it won't be in 4.0, but in something after that. (and we will be seeing a faster cadence of releases)
SAI is in CEP stage, see https://cwiki.apache.org/confluence/display/CASSANDRA/CEP-7%3A+Storage+Attached+Index

anuraaga (approved):
Nice!

PR #3261, opened 2020-10-25
Switches to run tests with release profile and using parallel builds

codefromthecrypt:
This makes travis run tests in the same profile as the release build, ensuring no surprises on release.

### DIFF
--- a/travis/publish.sh
+++ b/travis/publish.sh
@@ -167,8 +167,8 @@ fi
 if is_release_commit; then
   true
 else
-  # Ensure no tests rely on the actuator library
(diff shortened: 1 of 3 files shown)


travis/publish.sh
```diff
@@ -167,8 +167,8 @@ fi
 if is_release_commit; then
   true
 else
-  # Ensure no tests rely on the actuator library
-  MYSQL_USER=root ./mvnw verify -nsu -DskipActuator
```
codefromthecrypt, line 171: we don't use local mysql for a long time.

codefromthecrypt:
not such a good idea :P

PR #3267, opened 2020-10-28
Adds RELEASE_FROM_CONTEXT to allow Docker to re-use binaries just built

codefromthecrypt:
We currently have a race condition waiting for readback of remote repositories
containing zipkin and zipkin-slim jars. This introduces `RELEASE_FROM_CONTEXT`
which lets us re-use jars explictly allowed into the Docker context. The first
test is to re-use these on master build. After this, we can switch release
similarly.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
+# https://docs.docker.com/engine/reference/builder/#dockerignore-file
 **/.*
 !.dockerignore
 !.mvn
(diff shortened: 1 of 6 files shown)


.dockerignore
```diff
@@ -21,3 +22,7 @@
 **/Dockerfile
 docker/hooks/**
 travis/**
+
+# Allow re-use of built artifacts when building docker from master or a release tag
+!/zipkin-exec.jar
```
anuraaga, line 27: Cool

anuraaga (approved):
Way better than maven download

codefromthecrypt:
merging on winky build as this branch doesn't apply unless on master anyway. thanks for the look @anuraaga!

Issue #3166, opened 2020-08-06
Support building Zipkin on ARM64


odidev:
Hi Team,

I am interested in adding ARM64 support to Zipkin. 

There is an ‘elasticsearch’ binary for x86_64, getting downloaded in the RUN clause of the Dockerfile in zipkin/docker/storage/elasticsearch7/Dockerfile. The binary is available for aarch64 as well. I tried replacing the binary with aarch64, but build fails for zipkin-cassandra:2.21.5 docker image.

Please find my ARM64 job here : https://travis-ci.com/github/odidev/zipkin-1/jobs/368528378

Here is the failure logs : 
~~~
2020-08-05 06:53:00:414 [main] WARN zipkin2.storage.cassandra.v1.CassandraStorageExtension - Couldn't connect to docker image openzipkin/zipkin-cassandra:2.21.5: Assumption failed: All host(s) tried for query failed (tried: /127.0.0.1:9042 (com.datastax.driver.core.exceptions.TransportException: [/127.0.0.1:9042] Cannot connect))
~~~
I have checked in the docker hub <https://hub.docker.com/r/openzipkin/zipkin-cassandra/tags> that zipkin-cassandra is only available for LINUX/AMD64. 
I am suspecting that the further issues I am getting in the ARM64 build is due to zipkin-cassandra docker image unavailable for ARM64.
~~~
[ERROR]  HttpCallTest.wrongScheme:224
 
Expecting throwable message:
 <"not an SSL/TLS record: 485454502f312e31203430302042616420526571756573740d0a636f6e74656e742d747970653a20746578742f706c61696e3b20636861727365743d7574662d380d0a636f6e74656e742d6c656e6774683a2033310d0a7365727665723a2041726d657269612f302e39392e380d0a646174653a205468752c203330204a756c20323032302031353a35323a333720474d540d0a0d0a3430302042616420526571756573740a4465636f646572206661696c757265">
to contain:
 
 <"OPENSSL_internal">
 
but did not.
~~~
It would be really helpful if zipkin-cassandra docker image becomes available for LINUX/ARM64 platform as well. 
Please share your opinion on this.

codefromthecrypt:
hi. how are you using this? what you are mentioning is a test image. what is your tracing setup use as it is odd to try multiple storage as typically end users only use one.

regardless does the main Cassandra docker image work on arm64? if not then probably try that first or reference back here how it is working upstream as this is only a dev copy.

jorgheymans:
[...]

PR #3269, opened 2020-10-30
Updates so that lens can be built with arm64 (fixes skipTests on lens)

codefromthecrypt:
This notably changes the build so that when alpine+musl is in use,
frontend-maven-plugin isn't used. Instead, the same version it would
install is installed by the OS.

See #3166

### DIFF
--- a/build-bin/go_offline
+++ b/build-bin/go_offline
@@ -18,10 +18,13 @@ set -ue
 # This script hydrates the Maven and NPM cache to make later processes operate with less chance of
 # network problems.
 
+# Defensively avoid arm64+alpine problems with posix_spawn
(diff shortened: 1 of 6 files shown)


build-bin/go_offline
```diff
@@ -18,10 +18,13 @@ set -ue
 # This script hydrates the Maven and NPM cache to make later processes operate with less chance of
 # network problems.
 
+# Defensively avoid arm64+alpine problems with posix_spawn
+export MAVEN_OPTS="-Djdk.lang.Process.launchMechanism=vfork"
```
anuraaga, line 22: 🐉 🧙

anuraaga (approved):

PR #3270, opened 2020-10-30
Publishes multi-architecture Docker images

### LINKED ISSUE #3141: Add Docker builds for multi-arch (i.e amd64, arm64, arm/v7)
Please first, look at [existing issues](https://github.com/openzipkin/zipkin/issues) to see if the feature has been requested before. If you don't find anything tell us what problem you’re trying to solve. Often a solution already exists! Don’t send pull requests to implement new features without first getting our support. Sometimes we leave features out on purpose to keep the project small.

**Feature:**
The available deployment targets are changing. Low cost/low power ARM architectures are becoming more and more prevalent. Dockerhub supports these other architectures through different tags and build images. Multi-arch builds are also possible. An example Dockerfile is posted below as well as the command to build the various images with Docker Edge. This solution, unfortunately, requires the layering capabilities of Spring Boot 2.3 and might not be consistent with the zipkin server current build practices.

**Rational**
This is not in zipkin per se, however, it provides more flexible deployment targets for the server itself.

**Example Scenario**
AS an _Application Developer_
WHEN I want to build, deploy and monitor applications on an ARM architecture (i.e. rpi k8s cluster)
THEN I want to deploy the Zipkin Server image to k8s running on a rpi cluster

**Prior Art**
 * N/A

`Dockerfile` located in _/zipkin-server_
```txt
FROM adoptopenjdk/openjdk11 as builder
WORKDIR application
ARG JAR_FILE=target/zipkin-server-2.21.6-SNAPSHOT-exec.jar
COPY ${JAR_FILE} application.jar
RUN java -Djarmode=layertools -jar application.jar extract

FROM adoptopenjdk/openjdk11
WORKDIR application
COPY --from=builder application/dependencies/ ./
COPY --from=builder application/spring-boot-loader/ ./
COPY --from=builder application/snapshot-dependencies/ ./
COPY --from=builder application/application/ ./
ENTRYPOINT ["java", "org.springframework.boot.loader.JarLauncher"]
```

Build command
```bash
$ cd zipkin-server
$ docker buildx build -t openzipkin/zipkin-server:latest --platform linux/amd64,linux/arm64,linux/arm/v7 --push .
```

codefromthecrypt:
This runs tests with linux/amd64, but publishes linux/amd64,linux/arm64.

Fixes #3141

### DIFF
--- a/.github/workflows/docker-tests.yml
+++ b/.github/workflows/docker-tests.yml
@@ -46,7 +46,7 @@ jobs:
         with:
           fetch-depth: 1
       - name: Build Docker image openzipkin/${{ matrix.target }}:test
-        run:  docker/build_image ${{ matrix.target }}
+        run:  docker/build_image openzipkin/${{ matrix.target }}:test
       - name: Verify Docker image openzipkin/${{ matrix.target }}:test
         run: |
           # This just makes sure containers run and the HEALTHCHECK works (for now..)
--- a/docker/bin/install.sh
+++ b/docker/bin/install.sh
@@ -23,6 +23,10 @@ if [ "$RELEASE_FROM_CONTEXT" = "true" ]; then
   cp /code/zipkin-slim.jar zipkin-slim.jar
 elif [ "$RELEASE_VERSION" = "master" ]; then
   echo "*** Building from source..."
+
+  # Defensively avoid arm64+alpine problems with posix_spawn
+  export MAVEN_OPTS="-Djdk.lang.Process.launchMechanism=vfork"
+
   # Use the same command as we suggest in zipkin-server/README.md
   #  * Uses mvn not ./mvnw to reduce layer size: we control the Maven version in Docker
   (cd /code; mvn -T1C -q --batch-mode -DskipTests -Dlicense.skip=true --also-make -pl zipkin-server package)
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,8 @@ before_install:
   # Ensure Docker buildx is available and can build multi-architecture
   - |
     # Enable experimental features on the server (multi-arch)
-    echo '{"experimental":"enabled"}' > sudo tee /etc/docker/daemon.json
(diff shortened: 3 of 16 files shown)


.github/workflows/docker-tests.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         with:
           fetch-depth: 1
       - name: Build Docker image openzipkin/${{ matrix.target }}:test
-        run:  docker/build_image ${{ matrix.target }}
+        run:  docker/build_image openzipkin/${{ matrix.target }}:test
```
codefromthecrypt, line 49: @odidev when this PR merges, it should be ready to build on arm64. However, I don't know currently how to get the github action syntax to do a double-matrix (ex focal with amd64,arm64 against the docker image list)

docker/bin/install.sh
```diff
@@ -23,6 +23,10 @@ if [ "$RELEASE_FROM_CONTEXT" = "true" ]; then
   cp /code/zipkin-slim.jar zipkin-slim.jar
 elif [ "$RELEASE_VERSION" = "master" ]; then
   echo "*** Building from source..."
+
+  # Defensively avoid arm64+alpine problems with posix_spawn
+  export MAVEN_OPTS="-Djdk.lang.Process.launchMechanism=vfork"
```
codefromthecrypt, line 28: this cost a long time to figure out

.github/workflows/docker-tests.yml
odidev: @adriancole 
Thanks for the PR. Once this is merged we can add in Travis as it is faster for ARM64 with graviton2 in it. If you are ok with this, I can add and raise PR.

codefromthecrypt: at the moment we'll still need amd64 though at least on master branch, I think.. because that needs to do multi-arch push now. keep that in mind.

at any rate, yep help appreciated!

PR #3272, opened 2020-10-31
Stops using zipkin-builder Docker image

### LINKED ISSUE #3271: Replace zipkin-builder docker image with direct maven instead
We currently use openzipkin/zipkin-builder instead of normal jdk so that docker/bin/install.sh can build maven from within docker to produce a snapshot build. We no longer use this branch, instead using a flag `RELEASE_FROM_CONTEXT` which re-uses files from the docker context staged by the last mvn package invocation.

In doing the arch work I found it extremely slow and difficult to test zipkin-builder, because to build it requires downloading a lot of npm and maven artifacts. I think we should abandon this, as to my knowledge it was only made for docker hub builds and we no longer build master with docker hub (we do it in travis). even if we did it in another CI, it would be able to run java or our image in worst case.

so, the idea is to..

* remove the "Building from source" section from docker/bin/install.sh
* replace all usage of zipkin-builder with the normal jdk image (not jre classifier)
* in `docker/build_image`, if the tag is master and `RELEASE_FROM_CONTEXT` is not set, invoke maven like it says at the bottom of zipkin-server/README.md
* remove the job from .travis.yml which republishes zipkin-builder
* delete the zipkin-builder package from the github org.

codefromthecrypt:
This stops building a zipkin-builder image. This means that you can't
build Zipkin from source using only Docker.

Ex. a variant of the following won't work for source, only releases from
Maven central:
```bash
$ docker build https://github.com/openzipkin/zipkin
```

This implies Docker Hub auto-builds won't work for source releases,
which is not an issue as we don't use it anymore, and also Docker Hub
quota is problematic.

The pressure to remove source builds became strong when we started to
support multi-architecture builds. It is extremely time consuming to
attempt to run Maven and NPM inside of Docker, with an arm64 emulator.
This displaces effort that is sorely needed elsewhere and generally makes
the build unapproachable.

Since the Zipkin server itself is platform independent Java and client-
ide Javascript, there's no architectural impact to re-using the same
binaries for N arch builds. For example, this reworks the GitHub Action
validator to only build Zipkin once and use it for all the containers
to check.

Fixes #3271


.github/workflows/docker-tests.yml
```diff
@@ -6,16 +6,59 @@ on:
...
+        run: ./mvnw -T1C -q --batch-mode -DskipTests -Dlicense.skip=true --also-make -pl zipkin-server clean package
+        shell: bash
+        env:
+          CI: true
+      - uses: actions/upload-artifact@v2
+        with:
+          name: zipkin-server-jars
+          if: ${{ success() }}
+          if-no-files-found: error
+          path: |
```
codefromthecrypt, line 58: hoping this actually works, but can only test by watching :P

docker/bin/install.sh
```diff
@@ -15,33 +15,29 @@
...
-  cp /code/zipkin-slim.jar zipkin-slim.jar
-elif [ "$RELEASE_VERSION" = "master" ]; then
-  echo "*** Building from source..."
-
-  # Defensively avoid arm64+alpine problems with posix_spawn
-  export MAVEN_OPTS="-Djdk.lang.Process.launchMechanism=vfork"
-
-  # Use the same command as we suggest in zipkin-server/README.md
-  #  * Uses mvn not ./mvnw to reduce layer size: we control the Maven version in Docker
-  (cd /code; mvn -T1C -q --batch-mode -DskipTests -Dlicense.skip=true --also-make -pl zipkin-server package)
```
anuraaga, line 32: Even without the builder, does it hurt to just keep these three lines in the SNAPSHOT branch below instead of the warning message?

anuraaga (approved):

docker/bin/install.sh
codefromthecrypt: it hurts because it is never touched and can be wrong or drift and we wouldn't know it. Explicitly unsupporting nesting builds prevents the next person from losing ages trying to get it to work on multi-arch (also laying a bomb for the next maintainer if they succeed). They'll see where it was removed and why.

anuraaga: But it's not so bad if it doesn't support multiarch right? If it drifts then when someone notices they can fix it. There seems to me reasonable bang for the buck in these few commands. Up to you though :)

codefromthecrypt: you can add it back in a commit, but I think you will be the only user of this. I like the commit log shows how much horror this created.

PR #3276, opened 2020-10-31
Removes Docker Hub from integrated tests

codefromthecrypt:
Except ad-hoc tests that use Grafana and Prometheus, this rids remaining
use of Docker Hub registry images. This helps us avoid hitting rate
limits while debugging tests.


codefromthecrypt:
fyi @jeqo @adriancole @jcchavezs @jorgheymans @basvanbeek  this is the last step prior to releasing a patch to make it all official

PR #3278, opened 2020-11-01
Makes github actions re-use prior stage

codefromthecrypt:
Before, we didn't!


.github/workflows/docker-tests.yml
```diff
@@ -92,8 +91,9 @@ jobs:
       - uses: actions/download-artifact@v2
         with:
           name: zipkin-server-jars
+          path: zipkin-server/target/
```
codefromthecrypt, line 86: upload/download flattens the paths.. super weird.

anuraaga (approved):

PR #3280, opened 2020-11-01
silence most of Armeria except for the Server startup.

jorgheymans:


codefromthecrypt (approved):
yeah a log line snuck back in in testing zipkin-gcp

PR #3281, opened 2020-11-02
Tidies build and completes migration off docker.io images during build

### LINKED ISSUE #3279: Find a replacement for "rabbitmq:3.8-management-alpine"
"rabbitmq:3.8-management-alpine" used by RabbitMQCollectorRule is a docker hub registry image, which means each time a pull request runs potentially, this decrements a count and possibly can cause a build outage.

Can someone help find a replacement registry for it? ex in ghcr.io, gcr.io or quay? possibly we can layer it on our own image... not sure how complex it is.

cc @shakuzen @making @bsideup

codefromthecrypt:
This makes some editorial changes to make this project look the same as
zipkin-aws and zipkin-gcp. It also completes migration off docker.io
images so that routine test and development cannot cause a build outage.

Fixes #3279


benchmarks/src/test/resources/create-datasource-and-dashboard.sh
```diff
@@ -13,10 +13,9 @@
...
 #
 
-
 set -xeuo pipefail
 
-if ! curl --retry 5 --retry-connrefused --retry-delay 0 -sf http://grafana:3000/api/dashboards/name/prom; then
+if ! curl --retry 5 --retry-connrefused --retry-delay 0 -sf http://grafana:3000/api/datasources/name/prom; then
     curl -sf -X POST -H "Content-Type: application/json" \
          --data-binary '{"name":"prom","type":"prometheus","url":"http://prometheus:9090","access":"proxy","isDefault":true}' \
          http://grafana:3000/api/datasources
```
codefromthecrypt, line 21: this was a pre-flight check that checked the wrong condition, but wouldn't have mattered

codefromthecrypt:
fyi I verified grafana works in docker-compose, but for some reason, it starts, but doesn't import the dashboard in ServerIntegratedBenchmark. I'm leaving that as is for now as don't have time to figure out why.

benchmarks/src/test/java/zipkin2/server/ServerIntegratedBenchmark.java
```diff
@@ -151,46 +157,62 @@ void runBenchmark(@Nullable GenericContainer<?> storage) throws Exception {
...
-      .withCopyFileToContainer(
-        MountableFile.forClasspathResource("create-datasource-and-dashboard.sh"), "/create.sh");
+    // This is an arbitrary small image that has curl installed
+    // Use a quay.io mirror to prevent build outages due to Docker Hub pull quotas
+    // Use same version as in docker/examples/docker-compose-prometheus.yml
+    GenericContainer<?> grafanaDashboards =
+      new GenericContainer<>(parse("quay.io/rackspace/curl:7.70.0"))
+        .withNetwork(Network.SHARED)
+        .withWorkingDirectory("/tmp")
+        .withLogConsumer(new Slf4jLogConsumer(LOG))
```
codefromthecrypt, line 205: this is the container that I can't tell if even worked or not despite it being fine in docker-compose externally. Even it logged nothing though the logging properties here are set to info -> stdout.

anuraaga: If grafana has the zipkin dashboards, it should be working ok.

anuraaga (approved):

codefromthecrypt:
> If grafana has the zipkin dashboards, it should be working ok.

problem is that it doesn't :P not in the benchmark script anyway

codefromthecrypt:
we can debug the grafana dashboard lack in another PR. thanks for the look @anuraaga!

PR #3284, opened 2020-11-04
Removes root dependency management

codefromthecrypt:
Before, we imported BOMs at the root-most scope. This implies any
consumer of any library imports the same BOM. For example, I noticed
after clean installing brave-opentracing, that it was importing Netty's
BOM.

Root-level dependency management tempts us to accidentally add
dependencies to Zipkin's most core library. The convenience is not worth
the risk. Hence, this pushes out dependency management nearer to where
they are used.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,9 @@
     the License.
 
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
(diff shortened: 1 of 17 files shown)


codefromthecrypt:
cc @reta

pom.xml
```diff
@@ -155,233 +160,72 @@
...
-        <artifactId>zipkin-proto3</artifactId>
-        <version>0.2.2</version>
-      </dependency>
-      <dependency>
-        <groupId>com.squareup.wire</groupId>
-        <artifactId>wire-runtime</artifactId>
-        <version>${wire.version}</version>
-      </dependency>
-    </dependencies>
+    <!-- Be careful here, especially to not import BOMs as io.zipkin.zipkin2:zipkin has this parent.
```
codefromthecrypt, line 163: here's the important text

anuraaga (approved):

zipkin-collector/scribe/pom.xml
```diff
@@ -50,6 +50,7 @@
     <dependency>
       <groupId>javax.annotation</groupId>
       <artifactId>javax.annotation-api</artifactId>
+      <version>1.3.1</version>
```
reta, line 53: Sorry, a bit late with my comment, wouldn't it be better to define `javax.annotation-api` version in the parent `pom.xml`? I see it being used in a few places.

codefromthecrypt: sure. want to raise a PR to add that property? and use it in the few?

PR #3286, opened 2020-11-04
Fixes where benchmarks (40m jar) accidentally pushed to maven central

codefromthecrypt:
This ensures benchmarks doesn't end up deployed to maven central, which
both clutters and delays our deployment. Before, it didn't actually skip
nexus-maven-plugin. Rather that attempt to disable 3 plugins, this makes
the benchmarks module itself optional, disabling on deployment.

This also increase the sync timeout as when I tried to re-sync a failed
attempt it.. timed out. It seems most overriding this set the value to
20m which makes sense as likely double the time needed.


codefromthecrypt:
cc @reta @jeqo as this sort of thing has probably hit you before

reta (approved):
Yeah it did :)

PR #3288, opened 2020-11-05
Matches versions on Docker quick check and uses default Docker package

codefromthecrypt:
This uses the same Ubuntu and JDK version we use in the release process
for Docker "quick checks" which only build and verify containers start.

This style of check doesn't imply running integration tests, and can
identify issues far quicker as a result. Using the same OS dist and JDK
as Travis helps us share setup knowledge and reduce variables when
troubleshooting one vs another.

This also removes explicit version control of Docker, to prevent build outages
caused by rate limits. Instead, we use the implicit version of Docker, which
is new enough for our checks.

### DIFF
--- a/.github/workflows/docker-tests.yml
+++ b/.github/workflows/docker-tests.yml
@@ -8,7 +8,8 @@ on:
 jobs:
   prepare-docker-targets:
     name: Build Zipkin and Docker target matrix
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
@reta this should sort out the glitch

Thx @anuraaga for the advice to keep it simple

anuraaga (approved):

PR #3290, opened 2020-11-07
Avoids docker.io pulls from testcontainers (and removes a JDK install)

codefromthecrypt:
This avoids docker.io pulls implicitly done by testcontainers, in order
to prvent build outages due to pull limits.

This also removes a redundant JDK 11 install step, to speed up Travis.


codefromthecrypt:
verified visually that all integration tests that use docker in fact ran (weren't skipped)

PR #3294, opened 2020-11-10
Switch Travis-CI from 'arm64' to 'arm64-graviton2'

odidev:
Travis build environment is already set to 'lxd' by-default. 'virt: lxd' or 'virt: vm' is only required while working with 'arch: arm64-graviton2'

Signed-off-by: odidev <odidev@puresoftware.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 # Run `travis lint` when changing this file to avoid breaking the build.
 
 # See https://docs.travis-ci.com/user/reference/overview/#for-a-particular-travisyml-configuration
-virt: lxd   # LXD containers where possible as they boot much faster than full VMs
 os: linux   # required for arch different than amd64
 dist: focal # newest available distribution
 
@@ -79,7 +78,9 @@ jobs:
       if: 'NOT tag =~ /^docker-/'
     - stage: test
       name: "Run unit and integration tests including Docker [arm64]"
-      arch: arm64 # Not arm64-graviton2 as it cycles on boot for a long time before running [2020-11-02]
+      arch: arm64-graviton2 # Runs in AWS cloud and significantly faster than arm64
+      virt: lxd   # LXD containers where possible as they boot much faster than full VMs
+      group: edge
       # We run tests on non-tagged pushes to master that aren't commit made by the release plugin
       # We also run tests on pull requests targeted at the master branch.
       if: |


codefromthecrypt:
no thanks on this one. This calls attention to that the default VM isn't lxd which is the primary reason why we are using arm64 instead of amd64. Prefer to be explicit vs make assumptions that are only true when you hold arm64 constant.

https://docs.travis-ci.com/user/reference/overview/

codefromthecrypt:
ps if you prefer to move the rationale into the job definition, that's also fine. ex move lxd's comment into where arm64 is defined.

However we do things, it is important for folks who aren't familiar with this stuff to know quickly the tradeoffs made. I think almost no one has had time to track the arm64 effort so want their life easiest when they decide to jump in and do something, to understand what's going on and not accidentally subvert something. That's primarily the reason for all these comments which were explicitly written.. for the many folks who aren't able to watch 10 changes per day while kinks are sorted out.

odidev:
Setting ‘virt: lxd’ globally will also set the lxd environment for the AMD64 job, which is generally Full VM, as discussed here <https://docs.travis-ci.com/user/reference/overview/#which-one-do-i-use>. For Ubuntu Focal 20.04:  

arch: amd64: full VM only, default option
arch: arm64: LXD only
arch: arm64-graviton2: LXD and full VM
arch: ppc64le: LXD only
arch: s390x: LXD only

Hence, it is taking much time in the boot/queue state. 
We can mention ‘virt: lxd’ within the arm64 job itself. This would not affect the AMD64 build.

codefromthecrypt:
gotcha.. feel free to move that to the arm64 definition.

Also, if you want, give arm64 graviton another try. I just had a support email with travis about a different issue, but was plan related. My guess is graviton could have been affected by that, and so it might not cycle anymore. It would be cool for you to put  the PR in that actually puts graviton into use!

codefromthecrypt:
feel free to re-use this PR or make a new one. I'm watching something for a bit but will check back before I turn in for the night.

.travis.yml
```diff
@@ -79,7 +78,9 @@ jobs:
       if: 'NOT tag =~ /^docker-/'
     - stage: test
       name: "Run unit and integration tests including Docker [arm64]"
-      arch: arm64 # Not arm64-graviton2 as it cycles on boot for a long time before running [2020-11-02]
+      arch: arm64-graviton2
```
codefromthecrypt:
```suggestion
      arch: arm64-graviton2 # Runs in AWS cloud and significantly faster than arm64
```

codefromthecrypt (approved):
so far so good! Will merge in the morning if no one beats me to it.

codefromthecrypt:
raised a couple issues about the skipped tests

* #3295 kafka isn't running but I don't think it is related
* #3296 cassandra isn't running on arm64 - We need to fix the JRE if that's the problem. Then, we'll need to add details about the upstream issue in Cassandra if still relevant. This one is rather important to solve as this image is also used in zipkin-dependencies.. I'll look into it.

Thanks again for the help!

PR #3298, opened 2020-11-11
Makes zipkin-cassandra work with normal Java

codefromthecrypt:
This *should* avoid problems with Arm64, but needs to be tested post
merge. This does the following:

* significantly trim the image size by downloading from maven
* use less JVM-specific flags
* reduce memory usage for reasons including shutting off JMX
* dodge conflict during install by using a different port per arch

```
REPOSITORY                                      TAG                 IMAGE ID            CREATED             SIZE
openzipkin/zipkin-cassandra                     test                7b591f5d12cc        13 minutes ago      138MB
ghcr.io/openzipkin/zipkin-cassandra             2.22.2              ab5d087d0800        43 hours ago        262MB
```

### DIFF
--- a/docker/storage/cassandra/install.sh
+++ b/docker/storage/cassandra/install.sh
@@ -62,25 +62,6 @@ rm pom.xml
 # is decoupled from runtime
 mkdir -p conf data commitlog saved_caches hints triggers
 
-# Make a basic log4j conf which only logs warnings (to stderr)
-cat > conf/log4j.properties <<-'EOF'
-log4j.rootLogger=WARN, stderr
-
-log4j.appender.stderr=org.apache.log4j.ConsoleAppender
-log4j.appender.stderr.layout=org.apache.log4j.PatternLayout
-log4j.appender.stderr.layout.ConversionPattern=[%d] %p %m (%c)%n
-log4j.appender.stderr.Target=System.err
-
-# Ignore that we are using log4j as we aren't starting JMX anyway
-log4j.logger.org.apache.cassandra.utils.logging=ERROR
-# Ignore that we cannot increase RLIMIT_MEMLOCK or run Cassandra as root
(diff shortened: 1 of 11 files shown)


codefromthecrypt:
Note: I'm leaving JRE 1.8 as since after cassandra 3.11.3, you can't run it anymore due to use of sun.misc

docker/storage/cassandra/install.sh
```diff
@@ -18,63 +18,141 @@
...
     sleep 1
     timeout=$(($timeout - 1))
 done
 
 echo "*** Importing Scheme"
-cat zipkin-schemas/cassandra-schema.cql | bin/cqlsh --debug localhost
-cat zipkin-schemas/zipkin2-schema.cql | bin/cqlsh --debug localhost
-cat zipkin-schemas/zipkin2-schema-indexes.cql | bin/cqlsh --debug localhost
-
-echo "*** Adding custom UDFs to zipkin2 keyspace"
```
codefromthecrypt, line 65: removed UDFs as we didn't document them and they aren't used either

codefromthecrypt:
ran this with the docker-compose example to show that it is creating a working schema (ENSURE_SCHEMA is false there)

codefromthecrypt:
@openzipkin/core Travis is cut off due to a migration problem. I missed the fine print when migrated from travis-ci.org -> .com trying to get graviton up for @odidev... we have to explicitly apply to be OSS (similar to the application I filed the other day for Docker Hub). What happened was, I used up my personal credits which might replenish tomorrow I guess... I emailed their support to give us a big budget and sorry about the problem.

The build here was ironically only red on GH actions, which is now green, so I'll merge, though it won't cut an image because Travis is supposed to do that. I'll ping back here when travis gets back to me about the OSS credit thing.

PR #3302, opened 2020-11-12
Consolidates GH action build to a single file

codefromthecrypt:
This consolidates the two build processes into one flow, containing all
previous functionality, but less build time.

This builds the server on ubuntu, os/x and windows. If ubuntu, the flow
will continue into building and testing all our Docker images.

Something new is that this skips out if the change is documentation-only.


.github/workflows/continuous-build.yml
```diff
@@ -1,19 +1,33 @@
...
       - master
+    paths-ignore:
+      - '**/*.md'
   pull_request:
+    branches:
+      - master
+    paths-ignore:
+      - '**/*.md'
+
+# DEFAULT_OS is ubuntu-20.04 (focal) as that's the same as we use in Travis. This is find/replaced
```
anuraaga: Not sure, but are you able to use a single-element item in the matrix as a variable?

```
matrix:
  ubuntu-os:
  - ubuntu-20.04

...

if: matrix.os == ${{ matrix.ubuntu-os }}
```

anuraaga (approved):

.github/workflows/continuous-build.yml
codefromthecrypt: will try it! right now GH aren't charging us per run so why not give it a shot :D

codefromthecrypt: I think the last commit works. but GH actions is being annoying. will try to kick it with close/reopen

codefromthecrypt:
trying squash and force push as re-open had no effect on the stalled actions

codefromthecrypt:
last fail was legit at least parsing wise as it complained that yaml anchors are unsupported

codefromthecrypt:
ok I think this is as good as it gets. you can't use yaml anchors, you can't use env for runs-on, you also can't set default runs-on either.

Once you set a matrix somewhere, you can indeed use the variables so we can reduce, but not eliminate the copying.

PR #3306, opened 2020-11-16
Clarifies action.auto_create_index: false breaks Elasticsearch storage

### LINKED ISSUE #3300: Advice for action.auto_create_index
@varuna mentioned his site disabled `action.auto_create_index` and thought it would be helpful to document folks to check on this before using Zipkin in a shared cluster. There's some question of where to document that and what to document. For example, this seems to be a normal (non AWS) feature, but not sure the version range it applies to

https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#index-creation

Particularly, there's recent activity on this topic https://github.com/elastic/elasticsearch/pull/64208

any advice @openzipkin/elasticsearch?

jorgheymans:
fixes https://github.com/openzipkin/zipkin/issues/3300

### DIFF
--- a/zipkin-server/README.md
+++ b/zipkin-server/README.md
@@ -328,7 +328,7 @@ The above properties allow the most common SSL setup to work out of box. If you 
 customization, please make a comment in [this issue](https://github.com/openzipkin/zipkin/issues/2774).
 
 #### Automatic Index Creation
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
fantastic!

PR #3307, opened 2020-11-17
WIP: Support docker build for s390x

BbolroC:
As part of the work to support the [Knative](https://github.com/knative) pipeline for the s390x architecture, the dependency of the docker image `openzipkin/zipkin` was identified (https://github.com/knative/eventing/blob/master/test/config/monitoring.yaml#L46).

This PR is to resolve the dependency. 

The review can be started after the PR in `docker-java` (https://github.com/openzipkin/docker-java/pull/46) is merged. It will be on hold until then.

### DIFF
--- a/docker/build_image
+++ b/docker/build_image
@@ -42,7 +42,7 @@ RELEASE_FROM_MAVEN_BUILD=${RELEASE_FROM_MAVEN_BUILD:-false}
 JAVA_VERSION=${JAVA_VERSION:-15.0.1_p9}
 
 # Platforms to eventually push to the registry
-PLATFORMS="linux/amd64,linux/arm64"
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
hmm we should probably discuss this first, as especially aarch64 was painful as node.js doesn't support it. What's the underlying motivation here? Which end user is asking zipkin specifically to support this?

BbolroC:
> hmm we should probably discuss this first, as especially aarch64 was painful as node.js doesn't support it. What's the underlying motivation here? Which end user is asking zipkin specifically to support this?

Thanks for the comment. I will close this and https://github.com/openzipkin/docker-java/pull/46 first and open up an issue for discussion.

PR #3311, opened 2020-11-22
Moves to GitHub actions build

codefromthecrypt:
On CI:
  * Travis is still cutoff despite over a week of email.
  * We have to switch to GitHub actions to avoid paralyzing the project

On docker images:
  * zipkin-mysql and zipkin-ui now FROM openzipkin/alpine image, to ensure architectures like amd64 are uniform.
  * test images are reorganized under [docker/test-images] to help understand what's production or not.
  * [build-bin/docker_push] pushes test-images, but only to ghcr.io

On [build-bin]
  * [build-bin] largely reuses scripts used in other projects like zipkin-storage-kafka
  * This is discussed in [build-bin/README.md]

[zipkin-lens] is a node.js project, which has some additional concerns
* [build-bin/maybe_install_npm] is used to build on an unsupported node.js architecture.
* [build-bin/maven_go_offline] additionally seeds the NPM cache

On test:
* [build-bin/test], used by [.github/workflows/test.yml] runs Maven unit and integration tests.
* [.github/workflows/readme_test.yml] tests build commands in [zipkin-server] and [docker]
* zipkin, zipkin-lens and zipkin-slim Docker builds use `RELEASE_FROM_MAVEN_BUILD=true`
  * this avoids invoking Maven builds from within Docker, which are costly and fragile

On deploy:
* [build-bin/javadoc_to_gh_pages] pushes Javadoc to the gh-pages branch on N.M.L branch, but not master.


codefromthecrypt:
Main thing here is that the build-bin/*/** files will be common across projects. It was very hard to make them work for everything, but it is now the case. As mentioned in the README, files directly in build-bin/ are project specific, while directories like build-bin/docker are not.

hopefully the tests pass.. this is the first time we are using GH actions to actually run the docker storage tests. Before, we only built the server in GH actions, not ran heavy tests inside it.

codefromthecrypt:
ok all works. I'll refactor the tests slightly back as they were for merge.

PR #3315, opened 2020-11-23
Runs docker tests in parallel in CI

### LINKED ISSUE #3312: Split GitHub Actions test phase by docker
Most of our projects, except maybe a couple collectors, honor the "-Ddocker.skip=true" flag. It would be cool to split out the build so that the docker-related heavy tests run concurrently.

ex switch normal case to `build-bin/test -Ddocker.skip=true`, then add a matrix for the heavy things that use testcontainers

ex. `build-bin/test -pl zipkin-collector/rabbitmq` (rabbitmq might be one missing the skip thing though :P)

codefromthecrypt:
This splits normal tests from docker-based ones, doing the latter in
parallel as different jobs. The goal is to speed up the build which is
currently about 20 minutes.

Fixes #3312

### DIFF
--- a/.github/workflows/readme_test.yml
+++ b/.github/workflows/readme_test.yml
@@ -6,12 +6,8 @@ name: readme_test
 #
 # This doesn't literally scrape the README.md, so we don't test documentation-only commits.
 on:
-  # We run tests on non-tagged pushes to master that aren't a commit made by the release plugin
-  push:
-    tags: ''
-    branches: master
-    paths-ignore: '**/*.md'
-  # We also run tests on pull requests targeted at the master branch.
+  # We run tests on only on pull requests targeted at the master branch.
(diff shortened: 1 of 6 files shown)


codefromthecrypt:
ok folks I think we're finally under 20 minutes even with queuing delay

codefromthecrypt:
the tests that build our test images add a lot of github runner queuing delay. I've switched those to run sequentially instead.

codefromthecrypt:
last long pole was the windows build which held up docker tests until it completed. This decouples those by inlining the ubuntu server build into the docker image tests.

codefromthecrypt:
still have a slight queue delay as you can't do more than 8 concurrent jobs, and we have 9. anyway it is good enough I think

codefromthecrypt:
horray for 10minutes!

PR #3316, opened 2020-11-23
WIP eureka

### LINKED ISSUE #1870: Options for Service Discovery
We get occasionally asked about service discovery and registration etc. Please reply back with what you are using if anything to register zipkin in something like consul.

codefromthecrypt:
Don't think this is ready yet, as there are some gaps

@minwoox can you help champion this one?

Fixes #1870

### DIFF
--- a/zipkin-server/src/main/java/zipkin2/server/internal/eureka/ZipkinEurekaConfiguration.java
+++ b/zipkin-server/src/main/java/zipkin2/server/internal/eureka/ZipkinEurekaConfiguration.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2015-2020 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin2.server.internal.eureka;
+
+import com.linecorp.armeria.server.eureka.EurekaUpdatingListener;
+import com.linecorp.armeria.spring.ArmeriaServerConfigurator;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Condition;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.context.annotation.Conditional;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+/**
+ * Auto-configuration for {@link EurekaUpdatingListener}.
+ *
+ * <p>See https://armeria.dev/docs/server-service-registration#eureka-based-service-registration-with-eurekaupdatinglistener
+ */
+@ConditionalOnClass(EurekaUpdatingListener.class)
+@Conditional(ZipkinEurekaConfiguration.EurekaUrlSet.class)
+@Configuration(proxyBeanMethods = false)
+public class ZipkinEurekaConfiguration {
+  @Bean ArmeriaServerConfigurator eurekaListener(
+    @Value("${server.port:9411}") int port,
+    @Value("${zipkin.discovery.eureka.url}") String eurekaUrl,
+    @Value("${zipkin.discovery.eureka.appName:zipkin}") String appName,
+    @Value("${zipkin.discovery.eureka.instanceId:zipkin-server}") String instanceId) {
+    return sb ->
+      sb.serverListener(
+        EurekaUpdatingListener.builder(eurekaUrl)
+          .appName(appName)
+          // why can't this just take from the server?
+          .port(port)
+          .instanceId(instanceId)
+          // This will be wrong /health isn't allowed and our endpoint returns content, so can't
+          // implement HealthCheckService ex hostname and port are both likely wrong.
+          .healthCheckUrl("http://localhost:" + port + "/health")
+          .build());
+  }
+
+  /**
+   * This condition passes when "zipkin.discovery.eureka.url" is set to non-empty.
+   *
+   * <p>This is here because the yaml defaults this property to empty like this, and spring-boot
+   * doesn't have an option to treat empty properties as unset.
+   *
+   * <pre>{@code
+   * url: ${EUREKA_URL:}
+   * }</pre>
+   */
+  static final class EurekaUrlSet implements Condition {
+    @Override public boolean matches(ConditionContext context, AnnotatedTypeMetadata a) {
+      return !isEmpty(
+        context.getEnvironment().getProperty("zipkin.discovery.eureka.url")) &&
+        notFalse(context.getEnvironment().getProperty("zipkin.discovery.eureka.enabled"));
+    }
+
+    private static boolean isEmpty(String s) {
+      return s == null || s.isEmpty();
+    }
+
+    private static boolean notFalse(String s) {
+      return s == null || !s.equals("false");
+    }
+  }
+}
--- a/zipkin-server/src/test/java/zipkin2/server/internal/eureka/ITZipkinEureka.java
+++ b/zipkin-server/src/test/java/zipkin2/server/internal/eureka/ITZipkinEureka.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2015-2020 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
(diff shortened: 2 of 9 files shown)


zipkin-server/src/main/java/zipkin2/server/internal/eureka/ZipkinEurekaConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+@Configuration(proxyBeanMethods = false)
+public class ZipkinEurekaConfiguration {
+  @Bean ArmeriaServerConfigurator eurekaListener(
+    @Value("${server.port:9411}") int port,
+    @Value("${zipkin.discovery.eureka.url}") String eurekaUrl,
+    @Value("${zipkin.discovery.eureka.appName:zipkin}") String appName,
+    @Value("${zipkin.discovery.eureka.instanceId:zipkin-server}") String instanceId) {
+    return sb ->
+      sb.serverListener(
+        EurekaUpdatingListener.builder(eurekaUrl)
```
codefromthecrypt, line 43: questions here

zipkin-server/src/main/java/zipkin2/server/internal/eureka/ZipkinEurekaConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+          .appName(appName)
+          // why can't this just take from the server?
+          .port(port)
+          .instanceId(instanceId)
```
codefromthecrypt, line 47: ps this crashes unless you supply something

zipkin-server/src/test/java/zipkin2/server/internal/eureka/ITZipkinEureka.java
```diff
@@ -0,0 +1,128 @@
...
+
+  static final AtomicReference<HttpData> registerContentCaptor = new AtomicReference<>();
+
+  static final CompletableFuture<RequestHeaders> heartBeatHeadersCaptor =
+    new CompletableFuture<>();
+  static final CompletableFuture<RequestHeaders> deregisterHeadersCaptor =
+    new CompletableFuture<>();
+
+  @RegisterExtension static final ServerExtension eurekaServer = new ServerExtension() {
+    @Override protected void configure(ServerBuilder sb) {
```
codefromthecrypt, line 56: this is copy/pasta from armeria.. we don't need all of this I just had this half-way then noticed the health check was wrong. when done, all we need to do is check the json is sane.

zipkin-server/src/test/java/zipkin2/server/internal/eureka/ITZipkinEureka.java
```diff
@@ -0,0 +1,128 @@
...
+    );
+    context.refresh();
+
+    server = context.getBean(Server.class);
+  }
+
+  @Test void registersInEureka() {
+    await().until(() -> registerContentCaptor.get() != null);
+    assertThat(registerContentCaptor.get().toStringAscii())
+      .isEqualTo("sdd");
```
codefromthecrypt, line 116: this shows the json isn't sane

zipkin-server/src/main/java/zipkin2/server/internal/eureka/ZipkinEurekaConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+  @Bean ArmeriaServerConfigurator eurekaListener(
+    @Value("${server.port:9411}") int port,
+    @Value("${zipkin.discovery.eureka.url}") String eurekaUrl,
+    @Value("${zipkin.discovery.eureka.appName:zipkin}") String appName,
+    @Value("${zipkin.discovery.eureka.instanceId:zipkin-server}") String instanceId) {
+    return sb ->
+      sb.serverListener(
+        EurekaUpdatingListener.builder(eurekaUrl)
+          .appName(appName)
+          // why can't this just take from the server?
```
minwoox, line 45: Do you mean port?
If it so, it should take if from the server
https://github.com/line/armeria/blob/master/eureka/src/main/java/com/linecorp/armeria/server/eureka/EurekaUpdatingListener.java#L197-L200 🤔

zipkin-server/src/main/java/zipkin2/server/internal/eureka/ZipkinEurekaConfiguration.java
minwoox: Do you mean `instanceId`?
It should use `hostName` if it's null 🤔 
https://github.com/line/armeria/blob/master/eureka/src/main/java/com/linecorp/armeria/server/eureka/EurekaUpdatingListener.java#L191

zipkin-server/src/main/java/zipkin2/server/internal/eureka/ZipkinEurekaConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+          .port(port)
+          .instanceId(instanceId)
+          // This will be wrong /health isn't allowed and our endpoint returns content, so can't
+          // implement HealthCheckService ex hostname and port are both likely wrong.
```
minwoox, line 49: Do you explain more about this, please?

zipkin-server/src/test/java/zipkin2/server/internal/eureka/ITZipkinEureka.java
minwoox: https://github.com/openzipkin/zipkin/pull/3316/checks?check_run_id=1440795186#step:5:4120
So this is the content of the registering request. Which one is wrong?

minwoox (commented):
Thanks @adriancole! I left some questions. 😄

zipkin-server/src/main/java/zipkin2/server/internal/eureka/ZipkinEurekaConfiguration.java
codefromthecrypt: maybe it was just the unit test that NPE'd.. nevermind then!

zipkin-server/src/test/java/zipkin2/server/internal/eureka/ITZipkinEureka.java
codefromthecrypt: ah.. so I had to hard-code the port 9413 in order for it to show up here.. also I suspect the secure port 0 should not show up (not sure their protocol, but seems better to omit than say port 0)

the main issue is around default of the port and being able to indicate a healthcheck path as opposed to having to construct it.

```json
{"instance":{"instanceId":"zipkin-server","app":"zipkin","hostName":"fv-az75-872","ipAddr":"10.1.0.4","vipAddress":"fv-az75-872:9413","port":{"$":9413,"enabled":"true"},"securePort":{"$":0,"enabled":"false"},"status":"UP","healthCheckUrl":"http://localhost:9413/health","dataCenterInfo":{"@class":"com.netflix.appinfo.InstanceInfo$DefaultDataCenterInfo","name":"MyOwn","metadata":{}},"leaseInfo":{"renewalIntervalInSecs":30,"durationInSecs":90},"metadata":{},"lastUpdatedTimestamp":1606118031032,"lastDirtyTimestamp":1606118031032}}
```

zipkin-server/src/main/java/zipkin2/server/internal/eureka/ZipkinEurekaConfiguration.java
codefromthecrypt: when I ran without it, I saw 8080 showing up as the port.. maybe I'm blind.

but the idea would be that we don't indicate a port at all (ex in unit test server.port=0, which shoud become a real port), only a healthcheck path "/health" then it constructs from the actual server properties.

zipkin-server/src/main/java/zipkin2/server/internal/eureka/ZipkinEurekaConfiguration.java
codefromthecrypt: sure I basically would like to say healthcheckPath("/health") right now our healthcheck is an annotated service, not `HealthCheckService` which is a final class. `HealthCheckService` expects a lambda that returns boolean, whereas our health check endpoint returns more detail than that. Would be nice to be able to say the path without having to say the full url.

zipkin-server/src/main/java/zipkin2/server/internal/eureka/ZipkinEurekaConfiguration.java
minwoox: I have no idea where 8080 came from.
It should use the active port that the server finds:
https://github.com/line/armeria/blob/master/eureka/src/main/java/com/linecorp/armeria/server/eureka/EurekaUpdatingListener.java#L246
🤔

zipkin-server/src/main/java/zipkin2/server/internal/eureka/ZipkinEurekaConfiguration.java
minwoox: Ah, so you just want to specify the path, not the URL. I understood. 😄

zipkin-server/src/test/java/zipkin2/server/internal/eureka/ITZipkinEureka.java
minwoox: IIRC, it needs to guarantee compatibility with the legacy. 🤔 
https://github.com/Netflix/eureka/blob/07876a70a547ea52822c286b49c001971a1e6880/eureka-client/src/main/java/com/netflix/discovery/converters/jackson/serializer/InstanceInfoJsonBeanSerializer.java#L56

aegliv:
This MR seems dormant, is anyone still working on it?

jcchavezs:
Are you up to taking it over?

codefromthecrypt:
https://github.com/openzipkin/zipkin/pull/3671 finishes this

codefromthecrypt:
the problems here were mostly test lifecycle in nature (properties read at the wrong time), so https://github.com/openzipkin/zipkin/pull/3671 walked around them by using testcontainers+dynamic properties instead of nested initialization. https://github.com/openzipkin/zipkin/blob/master/zipkin-server/src/test/java/zipkin2/server/internal/eureka/ITZipkinEureka.java

PR #3318, opened 2020-11-27
Adjusts Cassandra image to work with JRE 11 and preps for Cassandra 4

codefromthecrypt:
This adds some workarounds for Cassandra when used on JRE 11, allowing
us a single base image in this repo (note zipkin-dependencies still
uses JRE 8 until Elasticsearch updates their spark support).

A single base image helps us as we don't need to download a couple
hundred meg only for one storage option. This also carves in Cassandra
4 support based on 4.0.0-beta-3 which passes almost all tests, but
deserves to wait until the next beta before we formalize in any way.

### DIFF
--- a/docker/test-images/zipkin-cassandra/Dockerfile
+++ b/docker/test-images/zipkin-cassandra/Dockerfile
@@ -12,7 +12,7 @@
 # the License.
 #
 
-# java_version is used for install and runtime layers of zipkin-elasticsearch6
+# java_version is used for install and runtime layers of zipkin-cassandra
 #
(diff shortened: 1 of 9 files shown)


docker/test-images/zipkin-cassandra/Dockerfile
```diff
@@ -12,10 +12,12 @@
 # the License.
 #
 
-# java_version is used for install and runtime layers of zipkin-cassandra
+# java_version is used for install and runtime layers of zipkin-elasticsearch6
```
codefromthecrypt:
```suggestion
# java_version is used for install and runtime layers of zipkin-cassandra
```

PR #3322, opened 2020-12-05
Ensures docker pull happens before push and polishes release files

codefromthecrypt:


codefromthecrypt:
going to cut a patch after that, so that latest and greatest is available (ex armeria) prior to announcing the already released 2.23

PR #3324, opened 2020-12-06
Waits longer and tries more often on connection reset

codefromthecrypt:
Elasticsearch wget download has been flaking causing the deploy job to fail. If this stuff keeps happening we may need to make a matrix job for deploy


anuraaga (approved):

codefromthecrypt:
ok no flakes now (jinxing myself)

thx for the look @anuraaga!

PR #3325, opened 2020-12-06
Stops using buildx completely

codefromthecrypt:
It is frequently the case that buildx uses the wrong image for cassandra
(the jre not the JDK) when arm64. This moves to the older but more
direct manifest approach to assembling multi-arch images until buildx
matures.

This follows from similar work in openzipkin/docker-alpine and -java

### DIFF
--- a/.github/workflows/docker_push.yml
+++ b/.github/workflows/docker_push.yml
@@ -17,6 +17,12 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 1  # only needed to get the sha label
+      - name: Cache Docker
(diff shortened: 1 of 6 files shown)


.github/workflows/docker_push.yml
```diff
@@ -17,6 +17,12 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 1  # only needed to get the sha label
+      - name: Cache Docker
```
codefromthecrypt, line 20: re-instated build cache as even with this removed buildx still picked wrong layers, might as well cache as mostly we'd want to.

PR #3329, opened 2020-12-13
WIP Try to diagnose test failure on ES 7.10

codefromthecrypt:
@xeraa The note about deprecation was valid, but actually the test fails, which is something I didn't raise a PR on before as was out of time..

Here's an http trace of the failing test which is related to composable templates. If you run this with environment ES_DEBUG=true you would see what I do ideally, but anyway maybe you can look and see what is wrong or if another http command in the test would make what is wrong easier to know.
[httptrace.txt](https://github.com/openzipkin/zipkin/files/5683989/httptrace.txt)

Here's the actual /_template output just prior to writing the first document

```json

  "eatezipkinindextemplate_gettraces_returnssuccess-dependency_template": {
    "order": 0,
    "index_patterns": [
      "eatezipkinindextemplate_gettraces_returnssuccess-dependency-*"
    ],
    "settings": {
      "index": {
        "number_of_shards": "5",
        "number_of_replicas": "1",
        "requests": {
          "cache": {
            "enable": "true"
          }
        }
      }
    },
    "mappings": {
      "enabled": false
    },
    "aliases": {}
  },
  "eatezipkinindextemplate_gettraces_returnssuccess-span_template": {
    "order": 0,
    "index_patterns": [
      "eatezipkinindextemplate_gettraces_returnssuccess-span-*"
    ],
    "settings": {
      "index": {
        "number_of_shards": "5",
        "number_of_replicas": "1",
        "requests": {
          "cache": {
            "enable": "true"
          }
        }
      }
    },
    "mappings": {
      "_source": {
        "excludes": [
          "_q"
        ]
      },
      "dynamic_templates": [
        {
          "strings": {
            "mapping": {
              "norms": false,
              "ignore_above": 256,
              "type": "keyword"
            },
            "match_mapping_type": "string",
            "match": "*"
          }
        }
      ],
      "properties": {
        "traceId": {
          "norms": false,
          "type": "keyword"
        },
        "duration": {
          "type": "long"
        },
        "remoteEndpoint": {
          "dynamic": false,
          "type": "object",
          "properties": {
            "serviceName": {
              "norms": false,
              "type": "keyword"
            }
          }
        },
        "localEndpoint": {
          "dynamic": false,
          "type": "object",
          "properties": {
            "serviceName": {
              "norms": false,
              "type": "keyword"
            }
          }
        },
        "_q": {
          "norms": false,
          "type": "keyword"
        },
        "timestamp_millis": {
          "format": "epoch_millis",
          "type": "date"
        },
[...]

PR #3333, opened 2020-12-25
Adds support for Elasticsearch 7.10

codefromthecrypt:
This fixes a comparison bug which installed the incorrect template.

What happened was v7.10 was mistaken for v7.1

Fixes #3329

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchVersion.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchVersion.java
@@ -17,35 +17,93 @@ import com.fasterxml.jackson.core.JsonParser;
 import com.linecorp.armeria.common.AggregatedHttpRequest;
 import com.linecorp.armeria.common.HttpMethod;
 import java.io.IOException;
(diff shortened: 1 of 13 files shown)


codefromthecrypt:
@basvanbeek FYI!

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchVersion.java
```diff
@@ -17,35 +17,93 @@
...
+
+      try {
+        int major = Integer.parseInt(matcher.group(1));
+        int minor = Integer.parseInt(matcher.group(2));
+        return new ElasticsearchVersion(major, minor);
+      } catch (NumberFormatException e) {
+        throw new IllegalArgumentException("Invalid .version.number: " + version);
+      }
     }
-    return Float.valueOf(version.substring(0, 3));
```
codefromthecrypt, line 48: this was the bug cc @xeraa!

xeraa: Ah, nice find. Thanks for fixing that!

PR #3350, opened 2021-04-03
[WIP] Enhance TraceTimeline component

### LINKED ISSUE #2509: visualize annotations in lens
Spans may have annotations in them. For example "brave.flush" is present when a span is abandoned (usually by accident). We should display them. The old ui used to use bubbles for this, but there are problems with that approach as noted in #1923

### LINKED ISSUE #2906: NaNs in LensUI 2.19 instead of timings

### LINKED ISSUE #2948: Show ellipsis when the service name is too long in trace tree

### LINKED ISSUE #2997: Use html tags for rendering trace graph instead of svg
Currently, svg is used for the trace graph.
This was implemented for rendering speed.

However, normal css cannot be used with svg, e.g. `text-overflow`.
This creates some rendering issues like https://github.com/openzipkin/zipkin/issues/2948 .

The problem of rendering speed can be solved by virtualization like using react-virtualized (https://github.com/bvaughn/react-virtualized).
So I think we should give priority to ease of drawing rather than drawing speed.

At first glance, svg is very good for drawing complex graphs flexibly, but it can be difficult to do finer things.
I think we should draw the chart with normal html tags & css, even if it's a bit painful at first.
Trace graphs are not as complex as they cannot be drawn with html & css.

### LINKED ISSUE #3118: text-anchor center in trace view makes it hard to know service name

tacigar:
# Summary

* Use TypeScript for TraceTimeline components
* Simplify tree-generation algorithm.
* Use normal HTML tags instead of using SVG for trace tree graph.
* Resolve some known issues.

# Screenshots

### Before

スクリーンショット 2021-04-04 23 16 10

### After

スクリーンショット 2021-04-04 23 16 02

# Description

## Simplify tree-generation algorithm

The algorithm for generating trees has been complex.
So I have greatly simplified it and made it more readable and testable.

## Use normal HTML tags instead of using SVG for trace tree graph

Trace trees were drawn using SVG. This was for rendering efficiency.
But compared to the regular DOM, styling with SVG is a bit more difficult, and some styling problem like #2948 occurs.
So, I used regular HTML tags instead of SVG in this PR.
This was an issue that had already been registered in #2997 .

* fix #2948 
* fix #2997 
* fix #3118 

However, the rendering of regular HTML tags is slower than SVG, and if you have to draw a lot of spans, the browser will get stuck. So I used a technique called "virtualization" to avoid the browser getting stuck for drawing.

### fix #2509 

Now annotations are marked (white line).

スクリーンショット 2021-04-04 23 41 59

and... error spans will be marked with an exclamation mark.

スクリーンショット 2021-04-04 23 49 12

### fix #2906 

Probably the shared trace can be drawn normally.

スクリーンショット 2021-04-04 23 54 01


tacigar:
Combining #3349 and #3348 with this PR change, trace page will look like this:

スクリーンショット 2021-04-04 22 43 47

PR #3367, opened 2021-08-14
Fixes int overflow issue for large TTL values

### LINKED ISSUE #3357: timestampRange calculation error due to int variable instead of long
timestampRange calculation error due to int variable instead of long

CassandraSpanStore.java

Change needed from:
final int indexTtl;

To:
final long indexTtl

It will calculate wrong if TTL from table is set for 7890000.

The result will be a bucket from the future.

vy8:
Fixes #3357.

The line `indexTtl * 1000` will produce an integer overflow for large indexTtl values. For example 7890000 * 1000 will produce an integer overflow (some negative number).

The simple fix is to use long in the multiplication.

### DIFF
--- a/zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/CassandraSpanStore.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/CassandraSpanStore.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2020 The OpenZipkin Authors
+ * Copyright 2015-2021 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


jcchavezs:
Ping @anuraaga

jcchavezs:
@vy8 could you add a unit test for this?

vy8:
Hey @jcchavezs just did. Even thought it's a simple test that checks if startMillis is smaller than endMillis it will fail with the old code that doesn't have the explicit 'L' for long.

zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/CassandraSpanStore.java
```diff
@@ -305,8 +305,8 @@
     UUID endUUID;
   }
 
-  TimestampRange timestampRange(QueryRequest request) {
-    long oldestData = Math.max(System.currentTimeMillis() - indexTtl * 1000, 0); // >= 1970
+  TimestampRange timestampRange(QueryRequest request, int indexTtl) {
```
jcchavezs, line 308: Do we truly need to pass indexTtl as param? I see this is a class attribute: https://github.com/openzipkin/zipkin/pull/3367/files#diff-d85e858924df5caed00b20b67deac8389b768227fc7fdf8f7575884ae941a663L54

vy8: @jcchavezs the indexTtl value comes from a static method and then gets set to a final class attribute. To create a test I need to either:
a) Set the property using reflection for the test.
b) Mock a static method.
c) Pass the indexTtl as a param.

I chose c) since it seemed as the least amount of hassle. Tell me which option you would prefer and I'll implement it.
Cheers!

jcchavezs:
Somehow the test is failing. Would you please look into it?

vy8:
@jcchavezs bumped the year in the license from 2020 to 2021 and all should be good.

codefromthecrypt:
thanks. will be in next release

PR #3368, opened 2021-08-23
upgarde to support armeria thrift0.14

zhfeng:


jcchavezs:
Fixed by https://github.com/openzipkin/zipkin/pull/3394

PR #3375, opened 2021-09-08
Bump java base container 15.0.4_p5

llinder:
- use latest java base container
- use buildx for multi-arch containers
- remove ryuk properties override


llinder:
Lots of dependency version bumps and a python2 issue with cqlsh. After a lot of tweaking versions the build looks good now.

codefromthecrypt (approved):
coolio. I think people will appreciate the CVE less result!

PR #3377, opened 2021-09-09
Limit usage of ES template parameter "include_type_name".

llinder:
Using parameter "include_type_name" with versions of ES that didn't yet support the parameter failes with `unrecognized parameter: [include_type_name]`

This change addresses that by limiting the usage to ES >= 6.7 < 7.0

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2020 The OpenZipkin Authors
+ * Copyright 2015-2021 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -46,8 +46,7 @@
 import zipkin2.storage.Traces;
 
 import static com.linecorp.armeria.common.HttpMethod.GET;
-import static zipkin2.elasticsearch.ElasticsearchVersion.V7_0;
-import static zipkin2.elasticsearch.ElasticsearchVersion.V7_8;
+import static zipkin2.elasticsearch.ElasticsearchVersion.*;
```
codefromthecrypt: nit say no to wildcards :D

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -361,12 +360,15 @@ String buildUrl(IndexTemplates templates, String type) {
     if (version().compareTo(V7_8) >= 0 && templatePriority() != null) {
       return "/_index_template/" + indexPrefix + type + "_template";
     }
-    if (version().compareTo(V7_0) < 0) {
-      // because deprecation warning on 6 to prepare for 7 :
+    if (version().compareTo(V6_7) >= 0 && version().compareTo(V7_0) < 0) {
```
codefromthecrypt, line 365: interesting anyone can manage running anything besides latest 6.x given ES support policy, but SGTM

codefromthecrypt (approved):
there's probably a test in ElasticsearchStorageTest that conditionally adds this based on a version, and there could be a new test to prove it doesn't for 6.6.6 the devil's ES

Issue #3374, opened 2021-09-07
Helm chart for zipkin


jonkerj:
## Feature
A helm chart to deploy zipkin on Kubernetes, maintained by the openzipkin community. Helm charts are a (very) common means to install software on Kubernetes. There are a couple of charts for zipkin floating around, but (warning, opions ahead) none of them is official + up to date + of good quality. Most mature projects around Kubernetes offer charts supported by the project nowadays.

## Rationale
This makes it easy for Kubernetes users to install Zipkin.

## Example Scenario
n/a

## Prior Art
* [zipkin results on artifacthub](https://artifacthub.io/packages/search?ts_query_web=zipkin&sort=relevance&page=1)

We (Equinix Managed Service NL) are willing to submit our own (currently closed source) chart if it's going to be included "upstream".

basvanbeek:
Hi @jonkerj,

Great to see you and your company are willing to contribute. As community we see the value of providing good generic helm charts for Zipkin so please submit a PR and we'll review.

jcchavezs:
I think this is also something @jbahire could be curious about as he has
done great work with helm charts in @hypertrace.

jcchavezs:
Thanks @jonkerj!

jcchavezs:
Follow up for this https://github.com/openzipkin/zipkin/issues/3408

codefromthecrypt:
a combination of factors including abandoned and/or unmerged PRs, as well version and tag clutter in this repo led to moving helm to its own repo, owned by its contributors. Looking forward to https://github.com/openzipkin/zipkin-helm/issues/1 completing and things getting online again.

PR #3378, opened 2021-09-27
Add Helm chart

jonkerj:
As discussed in #3374, we'd like to submit our Helm chart for inclusion at Zipkin. It may need some CD or other stuff, all feedback is welcome!

### DIFF
--- a/chart/zipkin-server/templates/deployment.yaml
+++ b/chart/zipkin-server/templates/deployment.yaml
@@ -1,3 +1,16 @@
+{{- /*
+Copyright 2015-2021 The OpenZipkin Authors
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+in compliance with the License. You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License
+is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+or implied. See the License for the specific language governing permissions and limitations under
+the License.
+*/}}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
--- a/chart/zipkin-server/templates/deployment.yaml
+++ b/chart/zipkin-server/templates/deployment.yaml
@@ -1,0 +1,58 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "zipkin.fullname" . }}
+  labels:
+    {{- include "zipkin.labels" . | nindent 4 }}
+spec:
+  {{- if not .Values.autoscaling.enabled }}
+  replicas: {{ .Values.replicaCount }}
+  {{- end }}
+  selector:
+    matchLabels:
+      {{- include "zipkin.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "zipkin.selectorLabels" . | nindent 8 }}
+    spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "zipkin.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          env:
+            - name: STORAGE_METHOD
+              value: mem
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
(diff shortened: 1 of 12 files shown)


jcchavezs:
Test is failing because of lack of license in YAML files. Not sure this is truly needed:

image

Shall we @basvanbeek @abesto?

basvanbeek:
yes we should have the license headers in.

jonkerj:
No problem, I've added them. Can someone trigger a re-test?

jonkerj:
Hm, the headers are there, albeit probabky in a different format your test expects. I've used Helm block comments (`{{- /* ... */ }}`) where possible, as the license applies to the templates and not to the generated manifests. I suspect the test does not like that. Although `Chart.yaml` and `values.yaml` contain simple YAML-comments (as they are not templated) and also make the test angry. Does the test work at all for YAML? Anything I can do about this?

jonkerj:
Did some digging in [license-maven-plugin](https://github.com/mojohaus/license-maven-plugin), and [it seems to me](https://github.com/mojohaus/license-maven-plugin/tree/master/src/main/java/org/codehaus/mojo/license/header/transformer) that it is limited in types of files / comment styles supported. I cannot seem to find YAML or Helm-style comments in there. 

One could (in theory) submit YAML/Helm support into that plugin, and bump the license-mavin-plugin version in this repo afterwards, but I am not eager to jump on that boat.

jcchavezs:
I am inclined to say we should ignore that folder for now but I'd like to
hear from @basvanbeek and @llinder. Also would be cool to see @abesto
around here.

abesto:
Who me? Hi! :D
TBH I don't think I should have a word on setting the requirements for this - that is, whether it's OK to not have license headers in these files. I know this was extremely important under ASF, I'm not sure how much we care about it now. Personally I wouldn't worry about ignoring all YAML files.

On the technical front, AFAICT YAML supports line comments starting with `#`, which the Maven plugin also supports with the `properties` comment style. So possibly it's a matter of submitting a PR to the plugin that adds `.yaml` to the default extension list? Though note, that transformer also understands hashbangs, which are invalid in YAML, but whatever. Better yet, https://www.mojohaus.org/license-maven-plugin/check-file-header-mojo.html#extraExtensions looks like it's possible to add an extra extension in configuration. So maybe that'd fly?

jcchavezs:
@jonkerj could we avoid that folder to move on? cc @basvanbeek

jonkerj:
> @jonkerj could we avoid that folder to move on? cc @basvanbeek

Sure, but I wouldn't know how. I am not familiar with the build system (it's maven, right?) and got lost trying to find this out. If someone can point me to other stuff getting ignored by this check, I'll see that I replicate that to my folder.

jonkerj:
Bump

mrfelek:
@jonkerj
ingress.yaml:
`{{- $svcPort := .Values.service.port -}}`   should be `{{- $svcPort := .Values.service.zipkin.port -}}`
values.yaml:
`service:
  tracing:
    type: ClusterIP
    port: 80
  zipkin:
    type: ClusterIP
    port: 9411
`
Anyway... good work.. Thank you!

jonkerj:
@mrfelek : thanks for the sharp eye, don't know how we could have missed that. Still looking for hints on how to exclude the chart subdir from the license header check. Can anyone point me towards the right direction?

jcchavezs:
@jonkerj are you on twitter? I'd like to have a word with you.

jonkerj:
I'm sorry, I'm not on Twitter. I'm on several slacks (Talos, CNCF, Cilium, Kubernetes), though

jonkerj:
Don't want to sound impatient, but I've got the feeling this PR and the conversation around it is stalling a bit. Don't get me wrong: we really love to help in getting this chart included, and believe it would boost the adaption of Zipkin in a Kubernetes environment. 

If this has no priority from the main devs: totally fine and we'd respect that, and we will find a different approach, but please be clear about this.

jcchavezs:
Let's merge this I will address the headers in a next PR.

chart/zipkin-server/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+securityContext:
+  # capabilities:
+  #   drop:
+  #   - ALL
+  readOnlyRootFilesystem: true
+  runAsNonRoot: true
+  runAsUser: 1000
+
+service:
+  tracing:
```
jcchavezs, line 51: Curious, why do we need this? What is it exposed under 80?

chart/zipkin-server/templates/deployment.yaml
```diff
@@ -0,0 +1,72 @@
...
+      serviceAccountName: {{ include "zipkin.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          env:
+            - name: STORAGE_METHOD
```
jcchavezs, line 50: Just curious, what is this exactly?

jonkerj: Not really sure what you mean, but let me take you through it line by line:
47: inserts the security context settings
48: sets the image to be used, defaulting to AppVersion from Chart.yaml
49/50: half of the code setting the storage backend to `mem`

PR #3381, opened 2021-10-11
Bump Armeria, Cassandra, Micrometer and Jackson deps

llinder:
There are a few CVEs with Netty. This change bumps a few Netty related dependencies to align on the latest Netty version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,15 +51,15 @@
 
     <!-- This allows you to test feature branches with jitpack -->
     <armeria.groupId>com.linecorp.armeria</armeria.groupId>
-    <armeria.version>1.3.0</armeria.version>
+    <armeria.version>1.12.0</armeria.version>
     <!-- Match Armeria version to avoid conflicts including running tests in the IDE -->
-    <netty.version>4.1.54.Final</netty.version>
+    <netty.version>4.1.69.Final</netty.version>
 
     <!-- It's easy for Jackson dependencies to get misaligned, so we manage it ourselves. -->
-    <jackson.version>2.12.0</jackson.version>
+    <jackson.version>2.12.4</jackson.version>
 
-    <java-driver.version>4.9.0</java-driver.version>
-    <micrometer.version>1.6.2</micrometer.version>
+    <java-driver.version>4.11.3</java-driver.version>
+    <micrometer.version>1.6.10</micrometer.version>
 
     <!-- Used for Generated annotations -->
     <javax-annotation-api.version>1.3.1</javax-annotation-api.version>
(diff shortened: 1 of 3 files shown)


zipkin-server/src/test/java/zipkin2/server/internal/ITZipkinServerTimeout.java
```diff
@@ -73,7 +73,7 @@
     Response response = get("/api/v2/trace/" + TRACE.get(0).traceId());
     assertThat(response.isSuccessful()).isFalse();
 
-    assertThat(response.code()).isEqualTo(503);
+    assertThat(response.code()).isEqualTo(500);
```
llinder, line 76: Checked and this is resulting in a RequestTimeoutException but end result is a 500. Switching the test seems OK to me. Alternative is probably some more work with exception mapping to keep a 503.

pom.xml
```diff
@@ -51,15 +51,15 @@
...
     <armeria.groupId>com.linecorp.armeria</armeria.groupId>
-    <armeria.version>1.3.0</armeria.version>
+    <armeria.version>1.12.0</armeria.version>
     <!-- Match Armeria version to avoid conflicts including running tests in the IDE -->
-    <netty.version>4.1.54.Final</netty.version>
+    <netty.version>4.1.69.Final</netty.version>
 
     <!-- It's easy for Jackson dependencies to get misaligned, so we manage it ourselves. -->
-    <jackson.version>2.12.0</jackson.version>
+    <jackson.version>2.12.4</jackson.version>
```
jeqo:
```suggestion
    <jackson.version>2.12.5</jackson.version>
```

pom.xml
```diff
@@ -51,15 +51,15 @@
...
 
-    <java-driver.version>4.9.0</java-driver.version>
-    <micrometer.version>1.6.2</micrometer.version>
+    <java-driver.version>4.11.3</java-driver.version>
+    <micrometer.version>1.6.10</micrometer.version>
```
jeqo:
```suggestion
    <micrometer.version>1.6.12</micrometer.version>
```

zipkin-server/src/test/java/zipkin2/server/internal/ITZipkinServerTimeout.java
jeqo: Good catch. 503 seems appropriate but as long as is not in the [spec](https://github.com/openzipkin/zipkin-api/blob/master/zipkin2-api.yaml) looks good to me to stick to `500`.

jeqo (approved):
LGTM. Did some small bumps to align to latest patch versions.

Thanks @llinder

PR #3397, opened 2021-12-02
chore: adds support for yaml license headers.

jcchavezs:
We chose to skip the check in chart repository because although it is a yaml file, the comment structure for the helm templates requires a different format than yaml (see #3378) and even with custom header type we can't match both.

Ping @jonkerj

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,10 @@
     the License.
 
 -->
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -610,6 +613,7 @@
             <exclude>**/generated/**</exclude>
             <exclude>.dockerignore</exclude>
             <exclude>build-bin/*</exclude>
+            <exclude>chart/**</exclude>
```
jcchavezs, line 616: Here it is the important change.

PR #3399, opened 2021-12-06
fix: chart: remove unneeded `tracing` service port

jonkerj:
In #3378 we've accidentally added two services to expose Zipkin, instead of only one. This PR fixes that.


jeqo (approved):
LGTM, though we don't have a test suite to validate this chart. 

@jonkerj could you confirm if this works on your side before merging?

mrfelek:
@jeqo  Already deployed zipkin with this helm commit - it works.

PR #3406, opened 2021-12-14
Upgrade log4j to 2.16 with further CVE fixes

llinder:


llinder:
Not as severe as the last log4j issue but bumping anyway to ensure exposure is eliminated https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-45046

jcchavezs:
Awesome. Would you do the same for brave?

llinder:
> Awesome. Would you do the same for brave?

Actually interesting that log4j is a dependency there and it would almost be better to just remove the dependency. I will check out the project later this afternoon after I'm done fighting a few fires.

PR #3413, opened 2021-12-20
log4j fix for zipkin-elasticsearch7

TheAschr:
Elasticsearch 7.10.1 is vulnerable to CVE-2021-44228


jcchavezs:
These are test images. Not sure if they represent and actual risk.

Issue #3401, opened 2021-12-10
chart/zipkin-server - elasticsearch support


mrfelek:
Rightnow chart/zipkin-server supports only "mem" storage type. As docker image allows to use elasticsearch I've prepared deployment.yaml file that supports elasticsearch. 
It was tested and it works. It doesn't require to make any changes in values.yaml as default STORAGE_TYPE is still set to mem.

Don't be hesitate to modify code. It can always be better :)

## Feature
zipkin-server helm chart that supports elasticsearch.

@jonkerj Would you be so kind and verify it and add to code? Deployment file in attachment. 
[deployment.zip](https://github.com/openzipkin/zipkin/files/7694609/deployment.zip)

jonkerj:
Sure, I'll look into it this week

bursztyn-pl:
Would be nice to have that feature out of the box. I've also verified it in my helm and it works :)

ywilkof:
+1 
Would be great to have this implemented!

jonkerj:
Sorry, it took a bit longer than expected: I've been busy and/or on holidays. Please see #3415, I'm looking for someone to confirm this actually works with ES (I don't run ES with Zipkin personally).

PR #3415, opened 2022-01-10
feature: support elasticsearch as output in Helm chart

jonkerj:
It took a while for me to tackle #3401, but this seems to do the trick. I tried to avoid complex logic in the template (like @mrfelek suggested), but opted for a JSON schema that includes logic to do the same. In the process, I bumped the image from 2.21.0 to 2.23.11 (someone still needs to look at a CI for this, will try that later on).

I tested it by feeding it these values and looking at the output of `helm template -f <test values files>`:

```
# this should only render memory env var
zipkin:
  storage:
    type: memory
```

```
# this should error gracefully
zipkin:
  storage:
    type: elasticsearch
```

```
# this should render correct ES env vars
zipkin:
  storage:
    type: elasticsearch
    elasticsearch:
      hosts: hostA hostB
      index: fooIndex
```

If anyone actually running Zipkin with ES can confirm this (I dont), I would be thankful.

### DIFF
--- a/chart/zipkin-server/Chart.yaml
+++ b/chart/zipkin-server/Chart.yaml
@@ -13,5 +13,5 @@ apiVersion: v2
 name: zipkin-server
 description: a distributed tracing system
 type: application
-version: 0.1.5
-appVersion: "2.21.0"
+version: 0.2.0
+appVersion: "2.23.16"
--- a/chart/zipkin-server/templates/deployment.yaml
+++ b/chart/zipkin-server/templates/deployment.yaml
@@ -48,7 +48,15 @@ spec:
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           env:
             - name: STORAGE_METHOD
-              value: mem
+              value: "{{ .Values.zipkin.storage.type }}"
+            {{- if eq .Values.zipkin.storage.type "elasticsearch" }}
+            {{- with .Values.zipkin.storage.elasticsearch }}
+            - name: ES_HOSTS
+              value: "{{ .hosts }}"
+            - name: ES_INDEX
+              value: "{{ .index }}"
+            {{- end }}
+            {{- end }}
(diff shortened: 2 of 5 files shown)


jonkerj:
Hmm, the missing header warning seems to come from `pom.xml`, which is not mine. Anything I can do about this?

chart/zipkin-server/Chart.yaml
```diff
@@ -13,5 +13,5 @@ apiVersion: v2
 name: zipkin-server
 description: a distributed tracing system
 type: application
-version: 0.1.5
-appVersion: "2.21.0"
+version: 0.2.0
+appVersion: "2.23.11"
```
mrfelek: Cannot find 2.23.11  maybe latest... for now is 2.23.16

mrfelek:
Tested on elasticsearch and works. I was not able to use old data from elasticsearch. New fresh data works.
Great job! Thank you

jcchavezs:
Merging as lint failure is unrelated. Fixing later.

chart/zipkin-server/templates/deployment.yaml
```diff
@@ -48,7 +48,15 @@ spec:
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           env:
             - name: STORAGE_METHOD
```
mrfelek, line 50: Regarding to https://github.com/openzipkin/zipkin/tree/master/zipkin-server  
it should be STORAGE_TYPE

jonkerj: well, it's merged now. Feel free to open a PR to fix this, though (I'm not an official zipkin dev)

jcchavezs:
PR here: https://github.com/openzipkin/zipkin/pull/3417

José Carlos Chávez


tir. 25. jan. 2022 kl. 10:38 skrev Jorik Jonker ***@***.***>:

Issue #3440, opened 2022-04-01
CVE-2022-22965 Spring Framework security issue


zhanw1983:
Security issue, raise by spring official organization.
For details, please refer to the spring official website，https://spring.io/blog/2022/03/31/spring-framework-rce-early-announcement

jcchavezs:
Do you mind opening a PR?

fre. 1. apr. 2022, 10:44 skrev zhanw996 ***@***.***>:

shakuzen:
The blog post mentions the conditions for being impacted by the vulnerability. I don't think Zipkin Server is vulnerable because it is packaged as a jar and does not even use Tomcat. So, this CVE isn't applicable to Zipkin unless I'm missing something.

ccbuildpro:
> The blog post mentions the conditions for being impacted by the vulnerability. I don't think Zipkin Server is vulnerable because it is packaged as a jar and does not even use Tomcat. So, this CVE isn't applicable to Zipkin unless I'm missing something.

We use zipkin by offical docker <https://zipkin.io/pages/quickstart.html>.
And this image has this problem。

shakuzen:
When running Zipkin via the docker image, it is also not packaged as a war file and it is not run on a standalone Tomcat, so I don't see how it is vulnerable. The Zipkin Server exec jar is extracted and run with `java -cp`: https://github.com/openzipkin/zipkin/blob/8a4f4b9c9a5a3204d9663ecef39d687785369c9a/docker/start-zipkin#L41-L43
What makes you think the Zipkin docker image is vulnerable?

zukudo:
CVE-2022-22965 is found in docker image.

Automated scans fail on: 
/zipkin/BOOT-INF/lib/spring-beans-5.3.13.jar
/zipkin/BOOT-INF/lib/spring-core-5.3.13.jar

shakuzen:
From the CVE you linked:

> These are the prerequisites for the exploit:
> -  JDK 9 or higher
> -  Apache Tomcat as the Servlet container
> -  Packaged as WAR
> -  spring-webmvc or spring-webflux dependency

As I mentioned in my previous comment, the Zipkin server is neither packaged as a WAR nor does it use Tomcat as the Servlet container, therefore it is not vulnerable to that CVE.

zukudo:
So it does not look like the problem related to zipkin jar file itself. The scope is container image. It looks like official docker image (openzipkin/zipkin) contains other vulnerable jars:
[...]

PR #3442, opened 2022-04-04
Update SpringBoot to 2.6.6 and Spring to 5.3.18 to address CVE-2022-2…

chenboy3:
PR to address issue https://github.com/openzipkin/zipkin/issues/3440 and CVE-2022-22965, as documented here: https://spring.io/blog/2022/03/31/spring-framework-rce-early-announcement. I'm not too familiar with zipkin, let me know if anything else is needed here!

Also would like to request an upgrade to zipkin-gcp as well for the same vulnerability: https://github.com/openzipkin/zipkin-gcp

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,8 +68,8 @@
     <javax-annotation-api.version>1.3.1</javax-annotation-api.version>
 
     <!-- update together -->
-    <spring-boot.version>2.4.13</spring-boot.version>
-    <spring.version>5.3.7</spring.version>
+    <spring-boot.version>2.6.6</spring-boot.version>
+    <spring.version>5.3.18</spring.version>
 
     <!-- MySQL connector is GPL, even if it has an OSS exception.
          https://www.mysql.com/about/legal/licensing/foss-exception/


shakuzen:
Zipkin Server shouldn't be vulnerable since it is packaged as a jar file and uses Armeria's embedded server. Likewise, zipkin-gcp would not be affected.
Upgrading the Spring Boot version two minors is not going to be as simple as changing the version in the POM. Looks like some tests are failing. I don't have time to look into this now, unfortunately. I'll try to come back to this if I do have time, but someone else feel free to work on it.

rswst8:
The jump to 2.6.6 is actually pretty big, but 2.5.13 also has the vulnerability fixes.  I put in a pull req to update to 2.5.13 and spring to 5.3.20.  All tests pass in my local IntelliJ

darkmastermindz:
@chenboy3 @jcchavezs @llinder This should also be resolved now that Spring Boot version is updated to 2.7.10 and Spring Framework is 5.3.27 as of #3538.

shakuzen:
Superseded by https://github.com/openzipkin/zipkin/pull/3538

Issue #3427, opened 2022-02-03
Provide option in OpenZipkin/Cassandra StorageType to turn On/off Hostname verification 


priyavivek2307:
Please first, look at [existing issues](https://github.com/openzipkin/zipkin/issues) to see if the feature has been requested before. If you don't find anything tell us what problem you’re trying to solve. Often a solution already exists! Don’t send pull requests to implement new features without first getting our support. Sometimes we leave features out on purpose to keep the project small.

## Feature
Currently, There is no option in Zipkin-Casandra Storage type to enable/disable Hostname Verification. By default , Host name verification defaults to True.

## Rationale
Sometimes, in test levels the jks certifciate doesn't contain all the SAN entries and hence having default cassandra behavior of having host name verification true , Hence, we can't integrate Cassandra to Zipkin.

## Example Scenario
Having this feature, We can work with different certificate (with/without) SAN entries in Test and production environment. Since Cassandra provides the support to turn on/off this behavior, Zipkin should not block any feature of underlying technology.

## Prior Art
datastax driver defaults to hostname verification, true:

[https://github.com/datastax/java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/internal/core/ssl/DefaultSslEngineFactory.java#L83-L85](https://nam11.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fdatastax%2Fjava-driver%2Fblob%2F4.x%2Fcore%2Fsrc%2Fmain%2Fjava%2Fcom%2Fdatastax%2Foss%2Fdriver%2Finternal%2Fcore%2Fssl%2FDefaultSslEngineFactory.java%23L83-L85&data=04%7C01%7Cpriyasharma.osv%40fedex.com%7C06226c54b75c47ca879808d9dc517a56%7Cb945c813dce641f884575a12c2fe15bf%7C0%7C0%7C637783064763923010%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=%2BwLt%2BBFnokrAuuXVZJ02fd2K30xroY0AJi8ClOt5ocU%3D&reserved=0)

2. Configuration for the java driver is here:
[...]

PR #3457, opened 2022-06-17
OpenZipkin/Cassandra StorageType to turn On/off Hostname verification #3427

priyavivek2307:


zipkin-server/src/main/resources/zipkin-server-shared.yml
```diff
@@ -144,6 +146,8 @@ zipkin:
       index-fetch-multiplier: ${CASSANDRA_INDEX_FETCH_MULTIPLIER:3}
       # Using ssl for connection, rely on Keystore
       use-ssl: ${CASSANDRA_USE_SSL:false}
+      # Override hostname verification in Cassandra setting
+      override-hostname-verification: ${CASSANDRA_OVERRIDE_HOSTNAME:false}
```
jcchavezs: I'd rather call the var `CASSANDRA_OVERRIDE_HOSTNAME_VERIFICATION` to be consistent.

priyavivek2307: updated the variable name as suggested. @jcchavezs

zipkin-storage/cassandra/src/main/java/zipkin2/storage/cassandra/internal/SessionBuilder.java
```diff
@@ -77,7 +79,12 @@ public static CqlSession buildSession(
     config = config.withBoolean(REQUEST_DEFAULT_IDEMPOTENCE, true);
 
     if (useSsl) config = config.withClass(SSL_ENGINE_FACTORY_CLASS, DefaultSslEngineFactory.class);
-
+    
+    if (overrideHostnameVerification)
```
ankit-gautam23: I think we can write this in one line, excluding the need of if-else.

Replacing this

if (overrideHostnameVerification) 
    	config = config.withBoolean(SSL_HOSTNAME_VALIDATION, true);
    else
    	config = config.withBoolean(SSL_HOSTNAME_VALIDATION, false);

With this

config = config.withBoolean(SSL_HOSTNAME_VALIDATION, overrideHostnameVerification);

ankit-gautam23 (changes requested):

codefromthecrypt (commented):
if you can use the square formatter, the change will be simpler as it won't change so many lines. I think you are using tabs where the project uses spaces

codefromthecrypt:
I think to get this across the finish line, it might be faster to redo it. Reason is that the cassandra property is `SSL_HOSTNAME_VALIDATION` and so that means our property should be `CASSANDRA_SSL_HOSTNAME_VALIDATION` for env and sslHostnameValidation for the field. using different names can cause confusion. Most of the work not done here is adding the spring yaml plumbing to propagate this setting, as well update the zipkin-server README section for cassandra.

zipkin-server/src/main/resources/zipkin-server-shared.yml
codefromthecrypt: for future, any time we are going to add a new variable, follow existing conventions first. The easiest way and regardless of cassandra or not, is to do a web or github search for the property and use that. If there are multiple styles used, prefer the most used, for example other tools that configure cassandra.

Regardless, we don't want any incoherence in our codebase, so we needn't invent anything and just reuse the norm. My follow-up PR does that, just explaining here as I noticed sometimes folks accidentally don't look and then ask others to use non-conventional names. https://github.com/apache/cassandra-java-driver/blob/8d5849cb38995b312f29314d18256c0c3e94cf07/core/src/main/java/com/datastax/oss/driver/api/core/config/TypedDriverOption.java#L230

codefromthecrypt:
reopened in https://github.com/openzipkin/zipkin/pull/3701 as I was unable to alter the branch under this PR.

codefromthecrypt:
https://github.com/openzipkin/zipkin/releases/tag/3.0.5 includes this

PR #3460, opened 2022-06-21
Initial commit for Postgres backend support

dwelch2344:
Opening this up for review in response to https://github.com/openzipkin/zipkin/issues/1232

Saw the rejected PR https://github.com/openzipkin/zipkin/pull/3228

Open to suggestions on reworking the schema for better storage. IMO the combination of index + structure should hold decent performance, and outside of actually using referential integrity (fk's and the like) I'm keen to try it as is. If performance became a problem, setting a minimum version of pg 10-12 + leveraging table partitioning would probably do a pretty good job to break up the data enough. 

Thoughts? @codefromthecrypt know you've been big on keeping this maintainable for future so happy to collaborate :)


codefromthecrypt:
I'm swamped in my new world of wasm, so won't be able to chime in or help support this. Folks currently on the show might decide differently than my previous opinions, so I'll leave it up to them!

metalhead44:
Awaiting this feature to get merged !!

bvo42:
+1

wilvdb:
Any news regarding this PR ?
I would be interested by the functionality.

jcchavezs:
Hi people. Thanks for showing interest on this feature. At the moment I don't think we have enough capacity to support a new storage in the core repo. How about moving this code to a repository where people can consume and try out and then when prooved to work we move to the contrib section?

In my experience working out a storage implementation isn't an easy task and hence we better do it carefuly.

PR #3471, opened 2022-08-18
Remove exception on traceIds of length 15

rogierslag:
References https://github.com/openzipkin/zipkin/pull/3249#discussion_r553047076

From the linked PR it seems this exception was incorrectly added as the code would prepend a trace of length 15 with a single leading zero (just like it would add two leading zeros for a trace of length 14).


codefromthecrypt (approved):
/me resists pun

depending on call-site this may need a version alignment/bump release in zipkin-reporter-java to make effective.

codefromthecrypt:
thanks for the fix @rogierslag!

jcchavezs:
Shall we release a patch?

codefromthecrypt:
yeah and then have zipkin-reporter-java consume it. I'm not sure if something else should go into the release here or not (ex if the docker images which seem to have been updated recently, should be consumed here)

PR #3483, opened 2022-11-02
Render ingress host

tobihein:
Support rendering hostnames like `my-app-{{ .Release.Name}}.example.com` to be more flexible with providing default-values with a common pattern

### DIFF
--- a/charts/zipkin/templates/ingress.yaml
+++ b/charts/zipkin/templates/ingress.yaml
@@ -39,13 +39,13 @@ spec:
     {{- range .Values.ingress.tls }}
     - hosts:
         {{- range .hosts }}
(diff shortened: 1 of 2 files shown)


charts/zipkin/templates/ingress.yaml
```diff
@@ -39,13 +39,13 @@ spec:
     {{- range .Values.ingress.tls }}
     - hosts:
         {{- range .hosts }}
-        - {{ . | quote }}
+        - {{ tpl . $ | quote }}
```
jcchavezs, line 42: As far as I understand this allows to evaluate host value as a tpl.

jcchavezs (approved):

PR #3493, opened 2022-12-23
Adjust styling of Dependencies page

tacigar:
* Remove shadows
    * In the previous Dependencies page, there were a lot of shadows and felt a bit cluttered.
* Remove AutoSizer.
    * Even though sizing can be done with just CSS, but AutoSizer was used
* Nothing has changed in the logic.

### Before

localhost_3000_zipkin_dependency_startTime=1671707527436 endTime=1671793927436 (1)

### After

localhost_3000_zipkin_dependency_startTime=1671707527436 endTime=1671793927436

https://user-images.githubusercontent.com/19551419/209411631-be113600-21cf-49d3-9600-602283740619.mov

### DIFF
--- a/zipkin-lens/src/components/DependenciesPage/DependenciesGraph.tsx
+++ b/zipkin-lens/src/components/DependenciesPage/DependenciesGraph.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2020 The OpenZipkin Authors
+ * Copyright 2015-2022 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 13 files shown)


zipkin-lens/src/components/DependenciesPage/DependenciesGraph.tsx
```diff
@@ -238,82 +223,70 @@ const DependenciesGraph: React.FC<DependenciesGraphProps> = ({
...
 
   return (
-    <Box width="100%" height="100%" data-testid="dependencies-graph">
-      <Grid container className={classes.containerGrid}>
-        <Grid
-          item
-          xs={focusedNodeName ? 7 : 12}
-          className={classes.graphItemGrid}
-        >
-          <AutoSizer>
```
tacigar, line 248: We don't need to use AutoSizer in this case.
Sizing can be done only with CSS.

PR #3494, opened 2022-12-23
Adjust styling of Search page

tacigar:
This PR adjusts styling of Search page. The changes are as follows

* Wider page width.
    * Previously, due to the narrow width, the number of search criteria that can be arranged in a row is small.
* Reduce unnecessary vertical empty space.
    * Previously, since there is a lot of unnecessary vertical empty space, fewer search results are displayed on the screen.
* Remove  trace ID from search results.
    * If the Trace ID is already known, you can search for traces directly from "Search by trace ID" of the Navbar.
    * So there is no use case for using trace IDs in a search page in a useful way.
* Nothing has changed in the logic.
 
### Before

Screenshot 2022-12-24 1 06 14

### After

Screenshot 2022-12-24 1 40 32

https://user-images.githubusercontent.com/19551419/209363433-f8bbe66c-2eaf-4935-ae59-7a120ba7134b.mov

### DIFF
--- a/zipkin-lens/src/components/DiscoverPage/DiscoverPageContent.tsx
+++ b/zipkin-lens/src/components/DiscoverPage/DiscoverPageContent.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2020 The OpenZipkin Authors
+ * Copyright 2015-2022 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -23,9 +23,8 @@ import {
   ButtonGroup,
   ButtonProps,
   Collapse,
-  Container,
   Divider,
-  Paper,
+  makeStyles,
   TextField,
   Typography,
 } from '@material-ui/core';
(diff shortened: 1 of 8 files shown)


llinder (approved):
Looks great!

tacigar:
Thanks @llinder !!

zipkin-lens/src/components/DiscoverPage/DiscoverPageContent.tsx
```diff
@@ -576,125 +594,97 @@ const DiscoverPageContent: React.FC<DiscoverPageContentProps> = ({
...
+              <Box width={300} ml={2}>
+                <Autocomplete
+                  multiple
+                  value={filters}
+                  options={filterOptions}
+                  renderInput={(params) => (
+                    <TextField
+                      {...params}
+                      variant="outlined"
+                      label="Service filters"
```
jcchavezs, line 673: No translation :(

zipkin-lens/src/components/DiscoverPage/DiscoverPageContent.tsx
```diff
@@ -576,125 +594,97 @@ const DiscoverPageContent: React.FC<DiscoverPageContentProps> = ({
...
+            <Typography variant="body1">
+              {filteredTraceSummaries.length === 1 ? (
+                <Trans>1 Result</Trans>
+              ) : (
+                <Trans>{filteredTraceSummaries.length} Results</Trans>
+              )}
+            </Typography>
+            <Box display="flex">
+              <ButtonGroup>
+                <Button onClick={expandAll}>Expand All</Button>
```
jcchavezs, line 661: I think we should spend some time adding translation wrappers.

PR #3498, opened 2022-12-25
Update packages of Lens

tacigar:
* Replace fortawesome(fontawesome) packages with Material-ui's icons.
    * Since material-ui/icons had similar icons, there was no need to add more dependency packages.
* Remove enzyme packages.
    * It was used to test only one small component. Otherwise, testing-library was used instead, so we should follow that..
* Update lingui packages to 3.15.0.
* Update classnames to 2.3.2.
* Update history to 5.3.0.
* Update lodash to 4.17.21.
* Update prop-types to 15.8.1.
* Update recharts to 2.2.0.
* Update shortid to 2.2.16.
* Update styled-components to 5.3.6.
* Update vizceral-react to 4.8.0.


codefromthecrypt:
Hi, @tacigar. Do you mind giving this another shot? It would really be nice for the next release to be up to date. I'm working on the java side of things.

codefromthecrypt:
If this is obviated by #3617, please close it out. Thanks!

PR #3499, opened 2022-12-25
Address the case where the root span parentId is not empty

tacigar:
There may be cases where the parentId of the root span is not empty but traceId..
This PR addresses that case.


tacigar:
This has been resolved by #3507 instead

PR #3446, opened 2022-05-10
Es options #3445

Filippo125:

PR #3434, opened 2022-03-11
feat: add extraEnv to chart

jonkerj:
This allows users to specify additional environment variables, such as JAVA_OPTS. This fixes #3429.

PR #3501, opened 2023-01-04
feat: adding es auth environment vars

erbrjr:
Adding Elasticsearch Authentication environment variables.

### DIFF
--- a/charts/zipkin/templates/deployment.yaml
+++ b/charts/zipkin/templates/deployment.yaml
@@ -55,6 +55,12 @@ spec:
               value: "{{ .hosts }}"
             - name: ES_INDEX
               value: "{{ .index }}"
+            - name: ES_USERNAME
+              value: "{{ .username }}"
+            - name: ES_PASSWORD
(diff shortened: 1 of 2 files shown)


jonkerj:
I've submitted #3434 a while ago, which could cover your use case too. It seems kind of hard to get someone to look at it, so let's make some noise 😄 

EDIT: there is also #3446, which does more or less the same. But it is getting the same amount of love as the other helm PRs

jcchavezs:
Please rebase @erbrjr

jcchavezs:
@jonkerj I think having specific fields for env in ES looks more appealing rather than ingesting whatever env var. Could you maybe clarify the env vars you are expecting to send?

jonkerj:
That's OK with me. Looking at this PR, it would involve `ES_USERNAME`, `ES_PASSWORD` and `ES_SSL_NO_VERIFY`.

The PR looks good to me, btw (after a rebase, probably).

jcchavezs (approved):

PR #3437, opened 2022-03-25
fix NullPointerException when parser.currentName() is null

weishaofei:
when parser.currentName() is null, switch will throw NullPointerException

PR #3503, opened 2023-01-18
feat: elasticsearch storage contains unknown fields

snower:
Optimize data parsing exception when elasticsearch storage contains unknown fields. For example, we record user-agent through tags, and then use elasticsearch's Ingest Pipeline to parse user-agent for easy query.

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonSerializers.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonSerializers.java
@@ -117,7 +117,7 @@ public final class JsonSerializers {
           result.shared(parser.getBooleanValue());
           break;
         default:
-          // Skip
+          skipUnknownValue(parser);
       }
     }
 
@@ -153,7 +153,7 @@ public final class JsonSerializers {
           port = parser.getIntValue();
           break;
         default:
-          // Skip
+          skipUnknownValue(parser);
       }
     }
 
@@ -184,7 +184,7 @@ public final class JsonSerializers {
           value = parser.getValueAsString();
           break;
         default:
-          // Skip
+          skipUnknownValue(parser);
       }
     }
 
@@ -192,6 +192,19 @@ public final class JsonSerializers {
       throw new IllegalArgumentException("Incomplete annotation at " + parser.currentToken());
     }
     return Annotation.create(timestamp, value);
+  }
+
+  private static void skipUnknownValue(JsonParser parser) throws IOException {
+    JsonToken value = parser.getCurrentToken();
+    if (value == JsonToken.START_OBJECT) {
+      while (parser.nextValue() != JsonToken.END_OBJECT) {
+        skipUnknownValue(parser);
+      }
+    } else if (value == JsonToken.START_ARRAY) {
+      while (parser.nextValue() != JsonToken.END_ARRAY) {
+        skipUnknownValue(parser);
+      }
+    }
   }
 
   public static final ObjectParser<DependencyLink> DEPENDENCY_LINK_PARSER = parser -> {
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonReaders.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonReaders.java
@@ -66,7 +66,6 @@ public final class JsonReaders {
   }
 
   static void visitObject(JsonParser parser, String name, Set<String> result) throws IOException {
-    checkStartObject(parser, true);
     JsonToken value;
     while ((value = parser.nextValue()) != JsonToken.END_OBJECT) {
       if (value == null) {
@@ -91,7 +90,9 @@ public final class JsonReaders {
             // End of input so ignore.
             return;
           }
-          visitObject(parser, name, result);
+          if (checkStartObject(parser, false)) {
+            visitObject(parser, name, result);
+          }
         }
         break;
       case START_OBJECT:


jcchavezs:
Ping @xeraa

xeraa:
"unknown fields" here is for reading a specific format from Elasticsearch, right? So not related to how Elasticsearch stores the data / the mapping?

Then I think this is more like an application usage question? I won't have as much input for that as I would for storage.

snower:
But I can't think that this is a usage problem. Elasticsearch's Ingest Pipeline will parse the input data to add new fields, such as parsing "user-agent" into separate device information and system information fields. And for the database support of the variable data structure, I think that the data should not be considered to be accurate and not deal with possible other situations. I think this is not rigorous, and this is just adding more and will not be used by itself. to the fields. At the same time, I think that a well-designed program should fully handle various situations of data without errors, which should include both the data submitted from the interface and the data read from the database.

codefromthecrypt:
depending on lack of feedback to the contrary, [this issue](https://github.com/openzipkin/zipkin/issues/3605) may change things on this PR, as the idea is changing to the default ES client. If change happens like this, worth re-visiting here.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/JsonSerializers.java
```diff
@@ -194,6 +194,19 @@ static Annotation parseAnnotation(JsonParser parser) throws IOException {
     return Annotation.create(timestamp, value);
   }
 
+  private static void skipUnknownValue(JsonParser parser) throws IOException {
```
ankit-gautam23, line 197: @snower @codefromthecrypt 
Can we rewrite this implementation for less line of code,

if(value == JsonToken.START_OBJECT || value == JsonToken.START_ARRAY){
    while(parser.nextValue() != (value == Json.START_OBJECT ? JsonToken.END_OBJECT : JsonToken.END_ARRAY)){
   skipUnknownValue(parser);
}
}

Let me know, if I do something wrong in this.

codefromthecrypt: @ankit-gautam23 as this PR is not updated recently, are you up to raising a new one. including a test to prove that the skipping works?

ankit-gautam23: @codefromthecrypt  Sure will include that too, Thanks for considering and Sorry for late response

codefromthecrypt:
@ankit-gautam23 @snower so what I understand here is that the type of tags is incorrect, for example, the value isn't a string, rather something else. So, I am guessing that you are talking about something like.. strange_tag here

```json
{
  "traceId": "4d1e00c0db9010db86154a4ba6e91385",
  "parentId": "86154a4ba6e91385",
  "id": "4d1e00c0db9010db",
  "kind": "CLIENT",
  "name": "get",
  "timestamp": 1472470996199000,
  "duration": 207000,
  "tags": {
    "http.path": "/api",
    "strange_tag": {"nested": "true"}
  }
}

```

So, I think if we did this what would end up happening is that because the indexing is custom, (`"tags": {"enabled": false }` in our index template, search might not be broken. However, since it is permitting broken values when our normal API doesn't allow it, other things might break. For example, our normal json codec should break on an invalid value.

If we do this for Elasticsearch, this does not mean we will do it for our default parser, because it is definitely not conformant data, and people should know that their data is invalid for zipkin. So, I want to set this expectation before we proceed.

If we do have this, not only does it need to be unit tested, but also added as an integration test to make sure it actually works. So, in `ITElasticsearchStorage` add a tests like..

* ITTraces.getTrace_skipsInvalidTagValue
  * lookup of trace by ID that includes a non-string value doesn't break returning it.
* ITSpanStore.getTraces_skipsInvalidTagValue
  * the search key in the query has one trace that matches the correct value and other has a non-string value doesn't break it.
* ITDependencies.getDependencies_linksSkipsInvalidTagValue
  * a trace that is linked based on span kind has a non-string value, and it doesn't break linking
  * This may also break zipkin-dependencies the separate project
* ITAutocompleteTags.skipsInvalidTagValue
  * autocomplete key is set to a trace which has a non-string value, and is ignored.

If any of the above don't work, we shouldn't land this, as it is already very difficult to support ES with template defined for our correct data format. Even if we do land this, it won't be a precedent for changing our core parser which correctly breaks if a tag has a non-string value.

This means we need in the PR a zipkin-storage/elasticsearch/RATIONALE.md for why we are doing this, knowing the index is for zipkin data and this is invalid. This rational should be strong but will not change our default parser, as permitting schema change in something we define as stable is a step backwards not forwards.

codefromthecrypt:
ps I will help with the test backfill on current code, as I think it will lead to resolving this one way or another more easy. In some ways this isn't a backfill, because for example, the collector enforces a JSON compliance. I'm pretty certain the only way data like this can end up in zipkin's indexes is if something that isn't zipkin is putting it there. Tests I'll look for and/or backfill will prove this out cc @reta I'll ping you when done maybe a couple days as busy week for me.

codefromthecrypt:
my suspicions were correct. Zipkin correctly rejects malformed spans, so if any are in zipkin's indices, something else wrote them. We only support our schema for use in our collectors, so beyond this is unsupported. It would need to be a compelling reason to attempt to support malformed data considering how few people help with Elasticsearch and how many problems we routinely have even with the supported config https://github.com/openzipkin/zipkin/pull/3695

codefromthecrypt:
See [this comment](https://github.com/openzipkin/zipkin/pull/3437#issuecomment-2011186880) as it is the same as for here. We cannot currently afford our own schema, so using data our collector intentionally rejects is outside the scope of our support ability. A change to make the data error easier to see is welcome.

PR #3507, opened 2023-02-09
Span with missing parent should be root

### LINKED ISSUE #3505: Trace with missing root span is not rendered correctly in zipkin-lens
## Describe the Bug
In zipkin-lens 2.24, viewing trace details where the root span is missing (a parentid is referring to a missing span), a blank page is rendered and an error is show in developer tools console. 

An example trace (abbreviated) looks like. Span with id "ebec1841bdb48ba1" is missing.

```
[
    {
        "traceId": "b523777284bbb42aa6304c7a5cd35fd7",
        "parentId": "ebec1841bdb48ba1",
        "id": "3eca62ab99ac8e84",
        "timestamp": 1675053275397814,
    },
    {
        "traceId": "b523777284bbb42aa6304c7a5cd35fd7",
        "parentId": "3eca62ab99ac8e84",
        "id": "7021b3404bd7b9d6",
        "timestamp": 1675053275401192,
    }
]
```
Here is the full trace: [gist](https://gist.github.com/drewby/1782583a17c5930905de11e2c97db320)

In 2.24, this renders as follows: 
error-2 24

## Steps to Reproduce
1. Post a trace to Zipkin with one or more spans with the root span missing. 
2. Click "show" to view details in zipkin-lens

## Expected Behaviour
Traces should render correctly even if root span is missing. Or an error message should be shown instead of a blank page. 

In 2.23, this example renders correctly as:
correct-2 23

drewby:
When all spans in a trace have a parentId, there is no root and an error occurs. Spans that have a parentId specified, but the parent does not exist should be treated as root. 

An example trace (abbreviated) looks like. Span with id "ebec1841bdb48ba1" is missing.

```
[
    {
        "traceId": "b523777284bbb42aa6304c7a5cd35fd7",
        "parentId": "ebec1841bdb48ba1",
        "id": "3eca62ab99ac8e84",
        "timestamp": 1675053275397814,
    },
    {
        "traceId": "b523777284bbb42aa6304c7a5cd35fd7",
        "parentId": "3eca62ab99ac8e84",
        "id": "7021b3404bd7b9d6",
        "timestamp": 1675053275401192,
    }
]
```
Here is the full trace: [gist](https://gist.github.com/drewby/1782583a17c5930905de11e2c97db320)

close #3505

### DIFF
--- a/zipkin-lens/src/components/TracePage/helpers.test.jsx
+++ b/zipkin-lens/src/components/TracePage/helpers.test.jsx
@@ -21,72 +21,76 @@
 import { convertSpansToSpanTree } from './helpers';
 
 describe('convertSpansToSpanTree', () => {
+  it('should return an empty array when there are no spans', () => {
+    const spans = [];
+    const result = convertSpansToSpanTree(spans);
+    expect(result).toEqual([]);
+  });
 
-    it('should return an empty array when there are no spans', () => {
-        const spans = [];
-        const result = convertSpansToSpanTree(spans);
-        expect(result).toEqual([]);
-    });
+  it('should return an array with correct root span', () => {
+    const spans = [
+      {
+        spanId: 'a',
+        spanName: 'root',
+        parentId: null,
+        childIds: ['b'],
+      },
+      {
+        spanId: 'b',
 
-    it('should return an array with correct root span', () => {
-        const spans = [{
-            spanId: 'a',
-            spanName: 'root',
-            parentId: null,
-            childIds: ['b'],
-        },
-        {
-            spanId: 'b',
-            
-            spanName: 'child',
-            parentId: 'a',
-            childIds: [],
-        }];
+        spanName: 'child',
+        parentId: 'a',
+        childIds: [],
+      },
+    ];
 
-        const result = convertSpansToSpanTree(spans);
-        expect(result.length).toEqual(1);
-        expect(result[0].spanId).toEqual('a');
-    });
+    const result = convertSpansToSpanTree(spans);
+    expect(result.length).toEqual(1);
+    expect(result[0].spanId).toEqual('a');
+  });
 
-    // This test is consistent with 'should work when missing root span' in
-    // span-cleaner.test.js
-    it('should return an array with the root element which is missing parent', () => {
-        const spans = [{
-            spanId: 'a',
-            spanName: 'missing parent',
-            parentId: 'm',
-            childIds: ['b'],
-        },
-        {
-            spanId: 'b',
-            spanName: 'child',
-            parentId: 'a',
-            childIds: [],
-        }];
+  // This test is consistent with 'should work when missing root span' in
+  // span-cleaner.test.js
+  it('should return an array with the root element which is missing parent', () => {
+    const spans = [
+      {
+        spanId: 'a',
+        spanName: 'missing parent',
+        parentId: 'm',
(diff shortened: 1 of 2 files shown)


jcchavezs:
Sounds very reasonable to me. I wonder if we should add a warning saying
there isn't clear root instead of pretend nothing happens. Shall we also
include a unit test for this?

Ping @basvanbeek, @tacigar and @shakuzen

basvanbeek:
I don't like the solution for a couple of reasons:
- A missing root span (or even an entire upper segment of a trace) should be made visible in the UI. This solution simply reroots without showing there is missing data
- this rerooting only works if the span to reroot to is the only child of this parent and this span is the only one missing its parent. In case of siblings this would not work or even worse, give a wrong presentation, same with multiple spans missing their parents.

Instead what should happen is that the UI should show every trace segment from the trace and show that a parent is missing  by adding a placeholder span identifying as missing. In this scenario it is possible to show something like:

```
- <missing>
  - child 1
  - child 2


- <missing>
  - child
    - child
``` 

where the order of the <missing span> segments is identified by time (as that's all we have to order in any reasonable way)

drewby:
@jcchavezs, sure I can look at adding a unit test. There weren't any tests for the rest of the Trace details view so was reluctant to be the first. :) 

@basvanbeek Thanks for comments.  

> * A missing root span (or even an entire upper segment of a trace) should be made visible in the UI. This solution simply reroots without showing there is missing data

I'm trying to replicate what happened in 2.23 when a parent span is missing. Here's a screenshot of the same trace data in 2.23.
correct-2 23

The 2.23 handling of a missing parent is consistent with other tracing visualizations I've seen. There are many reasons for the root to be missing. In this case the client application framework (ASP.NET) creates a trace context which is being picked up an SDK (Dapr SDK) and forwarded as part of the request to the service. The client application does not emit any spans to ZipKin and there is no expectation from the developer that trace spans would show up. 

> * this rerooting only works if the span to reroot to is the only child of this parent and this span is the only one missing its parent. In case of siblings this would not work or even worse, give a wrong presentation, same with multiple spans missing their parents.

I'm trying to figure out how the parentId lookup would have different results for two siblings. If the parent span exists, neither sibling would be considered a root, they'd both be childs of the parent span. If the parent span is missing, then both siblings would end up in the roots list. 

To add "missing" spans and mark them in the UI feels like a feature enhancement vs a bug fix. I don't think there is a UI element for representing a missing span today. What should the behavior be when one clicks the missing span for more details. 

I'd propose to have a quick fix for the current issue, which is a blank screen with no error message. And then put a feature enhancement in the backlog for identifying and investigating missing spans.

drewby:
@jcchavezs I've added the unit test.

I also found a similar handling of the missing root issue here: https://github.com/openzipkin/zipkin/blob/c6ad4a2bb67104ce72ed138a98291e4c4339c61c/zipkin-lens/src/zipkin/span-cleaner.test.js#LL786C6-L786C6

If we want to add a warning, should we add a property to AdjustedSpan called "missingParent" and then have a warning icon next to the span?

tacigar (approved):
Thank you very much @drewby 
Looks good to me!

tacigar:
> To add "missing" spans and mark them in the UI feels like a feature enhancement vs a bug fix.

Yeap, we should prioritize quick fixes. We can discuss about displaying "missing" spans later.

PR #3531, opened 2023-05-08
Zipkin charts adding namespaces

mfordjody:
zipkin chart lack namespaces

### DIFF
--- a/charts/zipkin/templates/_helpers.tpl
+++ b/charts/zipkin/templates/_helpers.tpl
@@ -45,6 +45,13 @@ Create chart name and version as used by the chart label.
 {{- end }}
 
 {{/*
+Create chart namespaces.
+*/}}
+{{- define "zipkin.namespace" -}}
+{{- default .Release.Namespace .Values.namespaceOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
 Common labels
 */}}
 {{- define "zipkin.labels" -}}
--- a/charts/zipkin/templates/deployment.yaml
+++ b/charts/zipkin/templates/deployment.yaml
@@ -15,6 +15,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ include "zipkin.fullname" . }}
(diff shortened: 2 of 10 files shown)


jcchavezs (approved):

jcchavezs:
Thanks @mfordjody. Could you check the error?

mfordjody:
The reason for the version conflict, but I currently cannot see the ci status

mfordjody:
Because the same version can cause conflicts, but I currently see the modification to version 0.7.0. I don't need to modify it again

jcchavezs:
I think the issue is we should add a `version: x.x.x-snapshot` or disable version checking by using ` --check-version-increment=false`. Any input @jonkerj ?

jonkerj:
It is very uncommon to template namespace from values, I'd recommend using `{{ .Release.Namespace }}` instead (which is very common practise). Any reason why you specifically want to sort of hardcode this from the values, @mfordjody?

As for the version/snapshot: if the chart version is CI'd during release, I'd agree with using snapshot/git refs for non-release, @jcchavezs.

mfordjody:
> It is very uncommon to template namespace from values, I'd recommend using `{{ .Release.Namespace }}` instead (which is very common practise). Any reason why you specifically want to sort of hardcode this from the values, @mfordjody?
> 
> As for the version/snapshot: if the chart version is CI'd during release, I'd agree with using snapshot/git refs for non-release, @jcchavezs.

good idea

jonkerj:
Much better. I would still personally prefer not using the define, and template `{{ .Release.Namespace }}` everywhere (it's what all major charts do), unless you really want to support the (IMO unlikely) scenario that the helm release lives in namespace A whereas the templated resources go into B. It may be a matter of taste, though. I'm a big fan of reducing complexity 😄

codefromthecrypt:
this hasn't moved in half a year, so I added you to https://github.com/openzipkin/zipkin-helm so you can work with the others and make changes as needed directly. Later, we should remove the helm assets from here.

PR #3545, opened 2023-05-31
Remove s390x from docker builds until cqlsh installation is fixed

llinder:
Installing cqlsh for s390x is requiring new things like installing rust, cargo among other things. After attempting to resolve all build dependenices the build still fails. Since s390x is really a small subset of users, we need to find help fixing this to continue supporting s390x with docker.

### DIFF
--- a/build-bin/docker/docker_push
+++ b/build-bin/docker/docker_push
@@ -64,7 +64,7 @@ for repo in ${docker_repos}; do
 done
 
 docker_args=$($(dirname "$0")/docker_args ${version})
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
this affected all images even though cassandra is only a test image. I can see how it isn't easy to make a change for one vs another.. will think about how to handle this

codefromthecrypt:
sorted in a more specific way here https://github.com/openzipkin/zipkin/pull/3613

PR #3546, opened 2023-06-06
Update hpa syntax

lz000:
The HPA syntax is old and already deprecated, here is the latest syntax: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/#autoscaling-on-multiple-metrics-and-custom-metrics

### DIFF
--- a/charts/zipkin/templates/hpa.yaml
+++ b/charts/zipkin/templates/hpa.yaml
@@ -30,12 +30,16 @@ spec:
     - type: Resource
       resource:
         name: cpu
-        targetAverageUtilization: {{ .Values.autoscaling.targetCPUUtilizationPercentage }}
+        target:
+          type: Utilization
+          averageUtilization: {{ .Values.autoscaling.targetCPUUtilizationPercentage }}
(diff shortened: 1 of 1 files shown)


jcchavezs:
Could you please rebase to include this change? https://github.com/openzipkin/zipkin/blob/1acbc8651df50aa8200cbff725285a1cb1c759d9/.github/ct.yml

lz000:
> Could you please rebase to include this change? https://github.com/openzipkin/zipkin/blob/1acbc8651df50aa8200cbff725285a1cb1c759d9/.github/ct.yml

Looks like my branch already has this change https://github.com/lz000/zipkin/blob/patch-2/.github/ct.yml

codefromthecrypt:
This hasn't moved in half a year, so I added you and @jonkerj as maintainers of https://github.com/openzipkin/zipkin-helm, so that you can make adjustments directly. If you need to publish to zipkin.io, you can also from there. For example, we do this both here and in brave (hint look at build-bin/deploy in either)

Issue #3468, opened 2022-08-03
Elasticsearch versions 5-7.x are supported, was: 8.0 


developer-tahir-saleem:
## openstack/zipkin-slim not connect with elasticsearch 8.3.0
We are using the docker image form https://hub.docker.com/r/openzipkin/zipkin-slim with  elastic search 8.0.0 or 8.3.0 Zipkin is not able to connect showing me error :

java.lang.IllegalArgumentException: Elasticsearch versions 5-7.x are supported, was: 8.0
        at zipkin2.elasticsearch.VersionSpecificTemplates.get(VersionSpecificTemplates.java:227)
        at zipkin2.elasticsearch.ElasticsearchStorage.versionSpecificTemplates(ElasticsearchStorage.java:356)
        at zipkin2.elasticsearch.ElasticsearchStorage.doEnsureIndexTemplates(ElasticsearchStorage.java:338)
        at zipkin2.elasticsearch.ElasticsearchStorage.ensureIndexTemplates(ElasticsearchStorage.java:330)
        at zipkin2.elasticsearch.ElasticsearchStorage.spanConsumer(ElasticsearchStorage.java:249)
        at zipkin2.server.internal.brave.ZipkinSelfTracingConfiguration$LocalSender.sendSpans(ZipkinSelfTracingConfiguration.java:163)
        at zipkin2.reporter.AsyncReporter$BoundedAsyncReporter.flush(AsyncReporter.java:299)
        at zipkin2.reporter.AsyncReporter$Flusher.run(AsyncReporter.java:378)
        at zipkin2.server.internal.brave.ZipkinSelfTracingConfiguration$1.run(ZipkinSelfTracingConfiguration.java:55)
        at java.base/java.lang.Thread.run(Unknown Source)

Can you please confirm if is there any version to support ES 8.3.0 .

lmarkusl:
same here.

https://github.com/openzipkin/zipkin/blob/c6ad4a2bb67104ce72ed138a98291e4c4339c61c/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/VersionSpecificTemplates.java#L225

```
  IndexTemplates get(ElasticsearchVersion version) {
    if (version.compareTo(V5_0) < 0 || version.compareTo(V8_0) >= 0) {
      throw new IllegalArgumentException(
        "Elasticsearch versions 5-7.x are supported, was: " + version);
    }
```
    
@llinder @jcchavezs 

[...]

PR #3552, opened 2023-06-23
Add support for elasticsearch 8

fazilali:
**Issue** - https://github.com/openzipkin/zipkin/issues/3468

- Not able to connect zipkin with elasticsearch 8.* version

**Solution**
- Add version 9 to avoid the check with version 8

### DIFF
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
@@ -167,8 +167,8 @@ public abstract class ElasticsearchStorage extends zipkin2.storage.StorageCompon
 
     /**
      * Only valid when the destination is Elasticsearch >= 7.8. Indicates the index template
-     * priority in case of multiple matching templates. The template with highest priority is used.
-     * Default to 0.
+     * priority in case of multiple matching templates. The template with the highest priority is
+     * used. Defaults to 0.
      *
      * <p>See https://www.elastic.co/guide/en/elasticsearch/reference/7.8/_index_template_and_settings_priority.html
      */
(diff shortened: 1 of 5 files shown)


shakuzen:
I'm not sure why CI isn't running on this pull request. Could you please try pushing a commit again to trigger it?

rcomanne:
Can we get this merged?

shakuzen:
@jcchavezs I've added the source for a zipkin-elasticsearch8 test image which is needed to run the integration test, but the CI here can't pull the test image because it hasn't been pushed yet. It's a bit of a chicken and egg situation.

fazilali:
@shakuzen How can we get this PR merged?

wu-sheng:
Elasticsearch8 would be supported on 3.x. 3.0.0-alpha1 will be cut soon.

codefromthecrypt:
thanks for the maintenance attempt @fazilali. I opened [this issue](https://github.com/openzipkin/zipkin/issues/3605) about an alternative future for zipkin v3. That said, I personally believe that v2 should be fixed, to know tests work before change. Monday, I'll try your change.

codefromthecrypt:
to start this, we need an ES 8 image. I'll work on this.

codefromthecrypt:
FYI ES changed their license after version 7.10, so it is no longer "OSS", but I suspect it is ok for us package a docker image like we did before, as it is a test image and not used for anything otherwise. cc @xeraa

codefromthecrypt:
Zipkin 2.25 will target ES 8 per https://github.com/openzipkin/zipkin/releases/tag/2.24.4 We have some work to do next week including sourcing an image

codefromthecrypt:
first step is making our ES 7 image use the Elastic licensed distribution. I am working on this.

codefromthecrypt:
please see notes at the bottom of #3468 about the base docker image, which is needed before this can proceed. If someone has time, please work on it in the next couple weeks, as the base project is totally ready for maintenance and change again.

codefromthecrypt (approved):
I had to do only minimal changes to get tests passing. Thanks tons @fazilali!

codefromthecrypt:
let's merge this after https://github.com/openzipkin/zipkin/pull/3666 as it should be in a new minor, followed by a change to remove testing for ES 6 (which is no longer supported by Elastic)

PR #3560, opened 2023-08-29
Upgrade testcontainers to 1.19.0

shakuzen:
This allows integration tests to be run on Apple silicon (which helps me and anyone else running IT on their recent macbook). Ports need to be explicitly exposed now.


shakuzen:
I'll merge since CI is green and this will help on some other work. If there's any issue, leave feedback post-merge and we can address it.

PR #3561, opened 2023-08-29
OTLP endpoint

marcingrzejszczak:
One option - GRPC support

- added generation of OTLP proto and grpc services (I think the latter is unnecessary)
- added a converter from OTLP to Zipkin Spans
- added an GRPC endpoint accepting OTLP proto format

TODO:

- tests
- docs
- optimizations
- classpath considerations

Another option - Proto over HTTP (https://github.com/openzipkin/zipkin/compare/master...marcingrzejszczak:zipkin:otlp_over_http?expand=1)

### DIFF
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -35,6 +35,18 @@
     <brave.version>5.13.2</brave.version>
     <log4j2.version>2.17.1</log4j2.version>
     <proto.generatedSourceDirectory>${project.build.directory}/generated-test-sources/wire</proto.generatedSourceDirectory>
+
+    <!-- OTLP -->
+    <build-helper-maven-plugin.version>3.4.0</build-helper-maven-plugin.version>
+    <exec-maven-plugin.version>3.1.0</exec-maven-plugin.version>
+    <maven-antrun-plugin.version>3.1.0</maven-antrun-plugin.version>
+    <protoc.version>24.1</protoc.version>
+    <otlp-proto.version>1.0.0</otlp-proto.version>
+    <os-maven-plugin.version>1.7.1</os-maven-plugin.version>
+    <grpc.version>1.57.2</grpc.version>
+    <protoc-gen-grpc-java.version>${grpc.version}</protoc-gen-grpc-java.version>
+    <javax.annotation-api.version>1.3.2</javax.annotation-api.version>
+    <ant-contrib.version>1.0b3</ant-contrib.version>
   </properties>
 
   <dependencyManagement>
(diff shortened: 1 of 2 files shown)


marcingrzejszczak:
@openzipkin/armeria @openzipkin/core any thoughts? Do you think that this makes any sense? Do you have any doubts about this?

trajano:
Any plans of integrating this soon so Zipkin server can accept OTLP traces?

wu-sheng:
A little update. We just added OTLP support on v3 branch. https://github.com/openzipkin/zipkin/pull/3580

We are planning to cut an RC release before the end of Nov.

codefromthecrypt:
TL;DR; We can't add maintenance of format drift and dependency management of an external data format. If we did, it would need to be proven out as a third party add on, and this project healthy, way before we do. At this point, none are true. No offense, but this needs to be moved to a separate repo, at least to begin with. The skywalking approach could also be moved to a separate repo. In any case, this won't land here in the near future.


---

Hi, marcin. long time no see. A long time ago, we spent a lot of effort to modularize non-core functionality, due to maintenance burden and deps. We evaluated a lot of options and went with spring for modularity and used that in a few areas, like [zipkin-gcp](https://github.com/openzipkin/zipkin-gcp). I think with the current state of maintenance, adding something with a lot of dependencies and an externally defined data format isn't a good idea now, even more so than before. We need to get more site-owners involved, and we have been neglecting their PRs, even to fix ES or armeria. We have to fix things like this before adding features.

Here is the rationale about modules, which was made after evaluating a lot of things several years ago

https://github.com/openzipkin/zipkin/blob/f75bf0fdad10db2616b10e237c9fbe827bf408a0/zipkin-server/RATIONALE.md?plain=1#L7

Here are several optional modules, which also can be added to the default server without causing maintenance here. Ack they haven't been maintained in a while, but that's more to the point on maintenance and also contributor status quo.

* https://github.com/openzipkin/zipkin-gcp/blob/1ad022a320665a5654a81082a2b509b91350cce5/module/src/main/resources/zipkin-server-gcp.yml
* https://github.com/openzipkin-contrib/zipkin-storage-kafka/blob/1b18b2369fd0cd54dba097799239a8b764be5fb4/module/src/main/resources/zipkin-server-storage-kafka.yml
* https://github.com/openzipkin/zipkin-aws/blob/8da16ce9b7e6098ccd5015d82d3c330d8b28c9bb/module/src/main/resources/zipkin-server-aws.yml
* https://github.com/openzipkin-contrib/zipkin-storage-forwarder/blob/e69b7982739e9d61ec100b3d9072daadaf0436a7/module/src/main/resources/zipkin-server-forwarder.yml
* https://github.com/jkschneider/zipkin-lucene-mapdb/blob/71968b4f896c1a38ac96db9441f99b9491867d81/src/main/resources/zipkin-server-berkeley.yml
* https://github.com/o1i0/zipkin-scouter/blob/53a017bf9e1b141c3d5b40781b748885c69c3688/autoconfigure-storage-scouter-udp/src/main/resources/zipkin-server-scouter.yml

codefromthecrypt:
opened https://github.com/openzipkin-contrib/zipkin-otel to carry any of this forward, if people decide to commit to review and maintenance there.

codefromthecrypt:
I will create the scaffolding this morning like zipkin-gcp for https://github.com/openzipkin-contrib/zipkin-otel this would also allow the [handler](https://github.com/openzipkin/zipkin-reporter-java/pull/213) or some variant there as well. Folks can also choose whether or not to, for convenience, migrate in the [propagation module for tracestate](https://github.com/openzipkin-contrib/brave-propagation-w3c). One reason to do that could be easy versioning.

in general please follow pattern of zipkin-gcp in decisions for how to config, where to add modules, etc. It will make it easy for people to swap into zipkin-otel and also remove some bikeshedding from naming conventions

cc @marcingrzejszczak @shakuzen

codefromthecrypt:
https://groups.google.com/g/zipkin-dev/c/YwKHhoR-hZQ/m/Fs7UBWMmBAAJ is ready for you to take over and run with

PR #3563, opened 2023-09-06
Allow use of jdk17

hezhangjian:


wu-sheng:
JDK17 would be supported on 3.x. 3.0.0-alpha1 would be cut soon.

PR #3567, opened 2023-09-19
Support multiple receivers in v3 server

mrproliu:
Currently, the Zipkin receiver in Skywalking only supports two modes: HTTP and Kafka. In this PR, I added support for the Zipkin raw data receiving protocol and configured it in the configuration file. At the same time, I modified the original GHA test files to support IT tests for each receiver.

In addition, I have observed that the Header time of each file is currently between 2015-2019. When all the features are completed, I can provide a separate PR to modify it to 2015-2023 uniformly.


wu-sheng:
> In addition, I have observed that the Header time of each file is currently between 2015-2019. When all the features are completed, I can provide a separate PR to modify it to 2015-2023 uniformly.

@jcchavezs What is the policy of us?

.github/workflows/test.yml
```diff
@@ -1,72 +1,73 @@
...
+  # We run tests on non-tagged pushes to master that aren't a commit made by the release plugin
+  push:
+    tags: ""
+    branches: master, zipkin-v3
+    paths-ignore:
+      - "**/*.md"
+      - "charts/**"
+  # We also run tests on pull requests targeted at the master branch.
+  pull_request:
+    branches: master, zipkin-v3
```
wu-sheng: I am feeling this doesn't take effects? Need merging?

mrproliu: Yes, I think it should be worked after merging.

wu-sheng: I feel you should submit this to the main branch to make it works.

mrproliu: But it has different project names. If I added to the main branch, all new PRs to the main branch should fail.

wu-sheng: I mean only branch name for the GHA on condition, rather than other parts of changes.

jcchavezs:
> In addition, I have observed that the Header time of each file is currently between 2015-2019. When all the features are completed, I can provide a separate PR to modify it to 2015-2023 uniformly.

We updated the license headers when the filed changed only. We don't align all files with the same year.

wu-sheng:
Could you try to open another PR to test CI?

PR #3569, opened 2023-09-19
Fix the v3 server CI tests

mrproliu:


wu-sheng (approved):
Great to see all these receivers are working now.

PR #3570, opened 2023-10-07
Support Cassandra Storage in v3

mrproliu:
Reimplementing Cassandra Storage in Zipkin Using SkyWalking OAP Core.

## Data Tables
1. **Table Creation**: In the previous version, tables were created using predefined CQL files. Now, they would be dynamically created based on specific Java files (automatically checked and created when Zipkin startup).
2. **Table Names and Fields**: The tables have changed due to differences in implementation principles with the original Zipkin. The corresponding relationships are as follows:
    * span: [ZipkinSpanRecord.java](https://github.com/apache/skywalking/blob/master/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/zipkin/ZipkinSpanRecord.java)
    * trace_by_service_span: Utilizes the Span table.
    * trace_by_service_remote_service: Utilizes the Span table.
    * span_by_service: [ZipkinServiceSpanTraffic.java](https://github.com/apache/skywalking/blob/master/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/zipkin/ZipkinServiceSpanTraffic.java)
    * remote_service_by_service: Utilizes the span_by_service table.
    * autocomplete_tags: Utilizes the additional table in **Span**, as shown [here](https://github.com/apache/skywalking/blob/master/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/zipkin/ZipkinSpanRecord.java#L169).
3. **Field Types**: Previously, custom data types were used to store `endpoint` and `annotation` information. Now, these would be converted into `text` and stored across multiple fields.
4. **Table Storage**: Previously, all data were stored in their respective standalone tables. Now, tables are created on a daily basis for data insertion and querying.

## Configuration

1. **TTL (Time To Live)**: In the previous version of Zipkin, the `default_time_to_live` field was predefined in the tables to allow the database to delete data. In the new version, data would be periodically cleared from database tables in Zipkin according to the DATA_TTL configuration.
2. **Auto Complete Cardinality**: This configuration has been removed in the new version, and no longer has any restrictions.


wu-sheng:
Let's put some summary about the feature here before the review.

zipkin-server/server-core/src/main/java/zipkin/server/core/CoreModuleProvider.java
```diff
@@ -133,8 +152,8 @@ public void prepare() throws ServiceNotProvidedException, ModuleStartException {
     final WorkerInstancesService instancesService = new WorkerInstancesService();
     this.registerServiceImplementation(IWorkerInstanceGetter.class, instancesService);
     this.registerServiceImplementation(IWorkerInstanceSetter.class, instancesService);
-    this.registerServiceImplementation(RemoteSenderService.class, new RemoteSenderService(getManager()));
-    this.registerServiceImplementation(RemoteSenderService.class, new RemoteSenderService(getManager()));
+    // no cluster mode for zipkin, for sending the streaming data to the local
+    this.registerServiceImplementation(RemoteSenderService.class, new SelfSenderService(getManager()));
```
wu-sheng: I think we still need cluster mode due to TTL task.

anuraaga:
I haven't been following that well but skimmed the description and wanted to ask, if the tables change significantly as I think they did, it seems like a big migration is needed by users. If a migration would be needed, is there a reason to develop Zipkin v3 vs asking users to switch to skywalking oap? I reread JC's email and Bas's issue comment and don't think I saw that addressed.

wu-sheng:
> I haven't been following that well but skimmed the description and wanted to ask, if the tables change significantly as I think they did, it seems like a big migration is needed by users. If a migration would be needed, is there a reason to develop Zipkin v3 vs asking users to switch to skywalking oap? I reread JC's email and Bas's issue comment and don't think I saw that addressed.

I think there was a discussion on the Zipkin admin mail list(I am not in that list).

About immigration, I think the general idea is not to keep the DB schema unchanged, we have new implementation in the OAP for MySQL/PostgreSQL/Elasticsearch/OpenSearch. And Cassandra is following the abstract.

anuraaga:
> About immigration, I think the general idea is not to keep the DB schema unchanged,

Thanks - indeed if this is the general idea, I wonder if all of these schema changes will be accompanied by some auto migration logic. If so, it's probably better to start by writing the migration logic before the server logic since it's always possible to run into impossible migrations.

If not, I still wonder, what will Zipkin v3 offer over switching to Skywalker for Zipkin users? I think it could be helpful to have a bullet list for that.

zipkin-server/server-core/src/main/java/zipkin/server/core/CoreModuleProvider.java
mrproliu: Sure, I will create other PR to add the cluster mode.

jcchavezs:
Although we aim to provide a drop in replacement for zipkin server we
discussed data retrocompatibility and decided to not to. Given the high
throughput and ingestion volume of such data, schema reconciliation can be
a pain.

What we agreed was that old data can't be consumed. If there are popular
demand on this (which I believe will be) maybe we can write tooling to
migrate data from one source to another so old data is also available in
new schema. Another option is that before one moves to the new server you
can have intermediate time where you leverage a OTel collector and write to
the two servers and progressively start moving to use new server.

wu-sheng:
Personally, I would say this kind of immigration is not worth so much effect. Because, the TTL of spans is short, such as days. In some cases, the data would lost the value before the immigration is done.

The reason for different schema designs in SkyWalking, is from what we learned from large-scale deployment. As SkyWalking agent + OAP usually work for 100% sampling. We created more concepts to help query(including SkyWalking UI and Lens UI) to have better responses.
As JC said, this is not unchangeable. We haven't began to gather feedback as currently, we are only working on this branch rather than main branch, and finish the first preview version, which makes everything work.

> If not, I still wonder, what will Zipkin v3 offer over switching to Skywalker for Zipkin users? I think it could be helpful to have a bullet list for that.

This PR is the value. SkyWalking as an APM is impossible to work with Cassandra successfully, but Zipkin server can.
Another thing like GraalVM support. SkyWalking OAP would have a long way(one year, even longer) to support GraalVM. But here as Zipkin v3, this should be able to support soon. OAP has many analysis relative things, which is hard to immigrate to run on GraalVM.

One of the key point is Zipkin VS SkyWalking OAP is that, OAP is too complex for Zipkin users. It support all cases for an observability team, but Zipkin v3 could be much lighter, more stable and focus on tracing relative things. Meanwhile, we could share the experiences on both sides.

anuraaga:
> write to the two servers and progressively start moving to use new server.

Yeah I assume that new server for new data, old server for old until deletion will be the practical approach vs trying any migration scripts.

So I'm still wondering why is this Zipkin? From what I follow in this thread, skywalking being more than tracing can actually impose limitations such as not working with Cassandra. That also seems weird to me though because if skywalking can't support Cassandra, there should be no need to change the Zipkin data model on Cassandra. If it's intended to be tracing-only Skywalking, isn't it better to still be part of Skywalking?

wu-sheng:
> So I'm still wondering why is this Zipkin? From what I follow in this thread, skywalking being more than tracing can actually impose limitations such as not working with Cassandra. That also seems weird to me though because if skywalking can't support Cassandra, there should be no need to change the Zipkin data model on Cassandra. If it's intended to be tracing-only Skywalking, isn't it better to still be part of Skywalking?

Whether SkyWalking community likes this, it would take a long time to make another decision from a totally another perspective. Zipkin server has this storage option, let's keep it supported.

And, I think more discussion could be made in another place. 

Let's keep this PR context clear to the Cassandra itself.

wu-sheng:
@mrproliu Please sync and avoid the conflicts as the bugs have been fixed through another PR.

wu-sheng:
Let's wait for @michaelsembwever back from vacation and do the review about Cassandra codes and table structure. He knows Cassandra much better than I could ever be.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/EmptyDAO.java
```diff
@@ -0,0 +1,489 @@
...
+  }
+
+  @Override
+  public TemplateChangeStatus disableTemplate(String id) throws IOException {
+    return null;
+  }
+
+  @Override
+  public void savePolicy(ContinuousProfilingPolicy policy) throws IOException {
+
```
wu-sheng: Could you create a NotImplementedException? And throw from here?
It is better to see a clear exception rather than NPE or other strange behavior as there is nothing executed.
After all, these methods should not be called.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/CassandraClient.java
```diff
@@ -0,0 +1,153 @@
...
+        authProvider,
+        config.getUseSsl());
+
+    // create keyspace if needs
+    final String keyspace = config.getKeyspace();
+    KeyspaceMetadata keyspaceMetadata = this.cqlSession.getMetadata().getKeyspace(keyspace).orElse(null);
+    if (keyspaceMetadata == null) {
+      String createKeyspaceCql = String.format(
+          "CREATE KEYSPACE IF NOT EXISTS %s " +
+              "WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'} " +
```
michaelsembwever, line 139: neither SimpleStrategy nor rf=1 is appropriate for any production system. NetworkTopologyStrategy and  rf=3 is the minimum for production.

michaelsembwever:
> In the previous version, tables were created using predefined CQL files.

Not having these makes it much harder to review the patch, and to benchmark it.
This needs the benchmark profiles so to be able to validate the scalability and superior performance characteristics of a Cassandra backend.  For example in master this is traces-stress.yaml and and trace_by_service_span-stress.yaml and span_by_service-stress.yaml files.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/CassandraTableInstaller.java
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    // record don't need to add the ID Column as partition key(for query performance)
+    if (model.isRecord()) {
+      columnDefinitions.add("PRIMARY KEY (" + (CollectionUtils.isEmpty(shardKeys) ? ID_COLUMN : Joiner.on(", ").join(shardKeys)) + ")");
+    } else {
+      columnDefinitions.add("PRIMARY KEY (" + ID_COLUMN + (CollectionUtils.isEmpty(shardKeys) ? "" : "," + Joiner.on(", ").join(shardKeys)) + ")");
+    }
+
+    final SQLBuilder sql = new SQLBuilder("CREATE TABLE IF NOT EXISTS " + table)
```
michaelsembwever: why are we using jdbc.SQLBuilder here? the java-driver used natively will be safer and cleaner.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/BatchCQLExecutor.java
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.skywalking.oap.server.library.client.request.UpdateRequest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import zipkin.server.storage.cassandra.CQLExecutor;
+import zipkin.server.storage.cassandra.CassandraClient;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class BatchCQLExecutor implements InsertRequest, UpdateRequest {
```
michaelsembwever: don't do batch statements.  and don't store then flush writes.  this is misunderstanding C*'s design and strengths.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraHistoryDeleteDAO.java
```diff
@@ -0,0 +1,113 @@
...
+  static final Logger LOG = LoggerFactory.getLogger(CassandraHistoryDeleteDAO.class);
+
+  private final CassandraClient client;
+  private final TableHelper tableHelper;
+  private final CassandraTableInstaller modelInstaller;
+  private final Clock clock;
+
+  private final Map<String, Long> lastDeletedTimeBucket = new ConcurrentHashMap<>();
+
+  public CassandraHistoryDeleteDAO(CassandraClient client, TableHelper tableHelper, CassandraTableInstaller modelInstaller, Clock clock) {
```
michaelsembwever: if we are bucketing by table do we still need TimeWindowCompactionStrategy? 

what is the performance impact of daily creating and dropping tables ? 

The correct (best performance and availability) design here is to define the time bucket inside the one table in its primary key, define the TTL at the table/write level, and let TimeWindowCompactionStrategy do what it's best at.  Here it feels like we are re-inventing the wheel at the application layer.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraZipkinQueryDAO.java
```diff
@@ -0,0 +1,297 @@
...
+    )) {
+      List<CompletionStage<List<String>>> completionTraceIds = new ArrayList<>();
+      if (CollectionUtils.isNotEmpty(request.annotationQuery())) {
+        final long timeBucket = TableHelper.getTimeBucket(table);
+        final String tagTable = TableHelper.getTable(ZipkinSpanRecord.ADDITIONAL_QUERY_TABLE, timeBucket);
+        for (Map.Entry<String, String> entry : request.annotationQuery().entrySet()) {
+          completionTraceIds.add(client.executeAsyncQuery("select " + ZipkinSpanRecord.TRACE_ID + " from " + tagTable +
+                  " where " + ZipkinSpanRecord.QUERY + " = ?" +
+                  " and " + ZipkinSpanRecord.TIME_BUCKET + " >= ?" +
+                  " and " + ZipkinSpanRecord.TIME_BUCKET + " <= ? ALLOW FILTERING",
```
michaelsembwever: is the primary key specified in the where clause here? 
allow filtering without the primary key specified is a full-table scan and will not work.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraZipkinQueryDAO.java
```diff
@@ -0,0 +1,297 @@
...
+              row -> row.getString(ZipkinSpanRecord.TRACE_ID),
+              entry.getValue().isEmpty() ? entry.getKey() : entry.getKey() + "=" + entry.getValue(),
+              duration.getStartTimeBucket(), duration.getEndTimeBucket()));
+        }
+      }
+      if (request.minDuration() != null) {
+        completionTraceIds.add(client.executeAsyncQuery("select " + ZipkinSpanRecord.TRACE_ID + " from " + table +
+                " where " + ZipkinSpanRecord.DURATION + " >= ?" +
+                " and " + ZipkinSpanRecord.TIME_BUCKET + " >= ?" +
+                " and " + ZipkinSpanRecord.TIME_BUCKET + " <= ? ALLOW FILTERING",
```
michaelsembwever: is the primary key specified in the where clause here? 
allow filtering without the primary key specified is a full-table scan and will not work.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraZipkinQueryDAO.java
```diff
@@ -0,0 +1,297 @@
...
+            row -> row.getString(ZipkinSpanRecord.TRACE_ID),
+            request.maxDuration(), duration.getStartTimeBucket(), duration.getEndTimeBucket()
+        ));
+      }
+      if (StringUtil.isNotEmpty(request.serviceName())) {
+        completionTraceIds.add(client.executeAsyncQuery("select " + ZipkinSpanRecord.TRACE_ID + " from " + table +
+                " where " + ZipkinSpanRecord.LOCAL_ENDPOINT_SERVICE_NAME + " = ?" +
+                " and " + ZipkinSpanRecord.TIME_BUCKET + " >= ?" +
+                " and " + ZipkinSpanRecord.TIME_BUCKET + " <= ? ALLOW FILTERING",
```
michaelsembwever: is the primary key specified in the where clause here? 
allow filtering without the primary key specified is a full-table scan and will not work.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraZipkinQueryDAO.java
```diff
@@ -0,0 +1,297 @@
...
+            row -> row.getString(ZipkinSpanRecord.TRACE_ID),
+            request.serviceName(), duration.getStartTimeBucket(), duration.getEndTimeBucket()
+        ));
+      }
+      if (StringUtil.isNotEmpty(request.remoteServiceName())) {
+        completionTraceIds.add(client.executeAsyncQuery("select " + ZipkinSpanRecord.TRACE_ID + " from " + table +
+                " where " + ZipkinSpanRecord.REMOTE_ENDPOINT_SERVICE_NAME + " = ?" +
+                " and " + ZipkinSpanRecord.TIME_BUCKET + " >= ?" +
+                " and " + ZipkinSpanRecord.TIME_BUCKET + " <= ? ALLOW FILTERING",
```
michaelsembwever: is the primary key specified in the where clause here? 
allow filtering without the primary key specified is a full-table scan and will not work.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraZipkinQueryDAO.java
```diff
@@ -0,0 +1,297 @@
...
+  @Override
+  public List<List<Span>> getTraces(Set<String> traceIds) throws IOException {
+    if (CollectionUtils.isEmpty(traceIds)) {
+      return Collections.emptyList();
+    }
+
+    final List<List<Span>> result = new ArrayList<>();
+    for (String table : tableHelper.getTablesWithinTTL(ZipkinSpanRecord.INDEX_NAME)) {
+      final PreparedStatement stmt = client.getSession().prepare("select * from " + table + " where " +
+          ZipkinSpanRecord.TRACE_ID + " in ?");
```
michaelsembwever: don't use `IN ` against. the primary key. async concurrent requests instead.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/CassandraTableInstaller.java
```diff
@@ -0,0 +1,261 @@
...
+            .map(ColumnName::getStorageName)
+            .collect(toList());
+
+    // adding the time_bucket as an index column when querying zipkin_query
+    columnsMissingIndex.add("time_bucket");
+    for (String column : columnsMissingIndex) {
+      final String index = "idx_" + table + "_" + column;
+      if (!indexExists(cassandraClient, table, index)) {
+        executeSQL(
+            new SQLBuilder("CREATE INDEX ")
```
michaelsembwever: 1. Avoid indexes, especially legacy 2i.  (v2 uses SASI, if you must better is to wait for 5.0 and use SAI, at minimum create a design that can still run without the index – true scalability as well as availability when an index needs to be rebuilt)
2. `indexExists(…)` does not work in a distributed environment.  it can be used as an optimisation but cannot be relied upon. use `IF NOT EXISTS`.  use the optimisation everywhere though (CREATE TABLE too).

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraCqlExecutor.java
```diff
@@ -0,0 +1,251 @@
...
+
+  protected <T extends StorageData> List<StorageData> getByIDs(CassandraClient client,
+                                                               String modelName,
+                                                               List<String> ids,
+                                                               StorageBuilder<T> storageBuilder) {
+    final List<String> modelTables = getModelTables(client, modelName);
+    List<StorageData> storageDataList = new ArrayList<>();
+
+    for (String table : modelTables) {
+      final SQLBuilder sql = new SQLBuilder("SELECT * FROM " + table + " WHERE id in ")
```
michaelsembwever: use the java-driver natively.  and don't use `IN` against the primary key.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraTagAutocompleteDAO.java
```diff
@@ -0,0 +1,70 @@
...
+  public Set<String> queryTagAutocompleteValues(TagType tagType, String tagKey, int limit, Duration duration) {
+    final Set<String> results = new HashSet<>();
+
+    for (String table : tableHelper.getTablesForRead(
+        TagAutocompleteData.INDEX_NAME,
+        duration.getStartTimeBucket(),
+        duration.getEndTimeBucket()
+    )) {
+      final SQLAndParameters sqlAndParameters = buildSQLForQueryValues(tagType, tagKey, limit, duration, table);
+      results.addAll(client.executeQuery(sqlAndParameters.sql() + " ALLOW FILTERING",
```
michaelsembwever: is the primary key specified in the where clause here?
allow filtering without the primary key specified is a full-table scan and will not work.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraTagAutocompleteDAO.java
```diff
@@ -0,0 +1,70 @@
...
+    final Set<String> results = new HashSet<>();
+
+    for (String table : tableHelper.getTablesForRead(
+        TagAutocompleteData.INDEX_NAME,
+        duration.getStartTimeBucket(),
+        duration.getEndTimeBucket()
+    )) {
+      final SQLAndParameters sqlAndParameters = buildSQLForQueryKeys(tagType, Integer.MAX_VALUE, duration, table);
+      results.addAll(client.executeQuery(sqlAndParameters.sql().replaceAll("(1=1\\s+and)|(distinct)", "") + " ALLOW FILTERING",
+          row -> row.getString(TagAutocompleteData.TAG_KEY), sqlAndParameters.parameters()));
```
michaelsembwever: is the primary key specified in the where clause here?
allow filtering without the primary key specified is a full-table scan and will not work.

michaelsembwever (changes requested):

michaelsembwever:
To summarise
 - Cassandra storage is for deployments that scale.  It can and should be designed to
  -- be as performant as, if not more than, other storage choices at no/low scale (bc C* is the preferred/only available db)
  --  work where the other storage won't
  -- scale linearly, so operators know it will continue to work as load increases, and costs over scale are predictable
  -- to validate any of this we need the benchmark profiles
- batch statements, and storing and flushing of writes, should not be done
- time bucketing at the primary key, and TTL in the table/write, is superior (and what TWCS is designed for)
- allow filtering used without specifying the primary key must be avoided
- using the IN clause against the primary key must be avoided
- legacy secondary indexes must be avoided
- SASI indexes can be optionally used, but the application needs to work (degraded mode) without them
- C* 5.0 offers SAI indexes, superior to SASI, ok to use (but still try to design without them)
- keyspace replication config must be configuration (and not be Simple nor rf<3 in any production)
- you cannot rely on read+write coordination at the application layer: e.g. `indexExists(…)`


If you have integration tests, you can also rely upon C* server-side guardrails to check you're designing and using correctly.  (Such a guardrails definition would also be valuable advice to production operators.)

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraTagAutocompleteDAO.java
mrproliu: Sure, the `ALLOW FILTERING` has been deleted.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraTagAutocompleteDAO.java
mrproliu: Sure, the `ALLOW FILTERING` has been deleted.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraCqlExecutor.java
mrproliu: Understood. I have divided each CQL query for execution.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/CassandraTableInstaller.java
mrproliu: > Avoid indexes, especially legacy 2i. (v2 uses SASI, if you must better is to wait for 5.0 and use SAI, at minimum create a design that can still run without the index – true scalability as well as availability when an index needs to be rebuilt)

Updated, still using the SASI index for now. I could update the SAI index when Cassandra v5.0 is ready. 

> indexExists(…) does not work in a distributed environment. it can be used as an optimisation but cannot be relied upon. use IF NOT EXISTS. use the optimisation everywhere though (CREATE TABLE too).

Update to using the pre-defined table CQL file.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraZipkinQueryDAO.java
mrproliu: Updated to using multiple equals query.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraZipkinQueryDAO.java
mrproliu: I have updated to use of the latest secondary table in the v2 for searching.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraZipkinQueryDAO.java
mrproliu: I have updated to use of the latest secondary table in the v2 for searching.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraZipkinQueryDAO.java
mrproliu: I have updated to use of the latest secondary table in the v2 for searching.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraZipkinQueryDAO.java
mrproliu: I have updated to use of the latest secondary table in the v2 for searching.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraHistoryDeleteDAO.java
mrproliu: Sure, I have updated to use the original Cassandra table settings, nothing more.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/BatchCQLExecutor.java
mrproliu: Understood. I have divided each CQL query for execution.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/CassandraTableInstaller.java
mrproliu: Deleted.

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/CassandraClient.java
mrproliu: These are the default settings in the [Zipkin CQL file](https://github.com/openzipkin/zipkin/blob/master/zipkin-storage/cassandra/src/main/resources/zipkin2-schema.cql#L1-L3), should I updated?

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/dao/CassandraHistoryDeleteDAO.java
mrproliu: Sure, I have updated to use the original Cassandra table settings, nothing more.

wu-sheng:
@michaelsembwever Could you review for another round?

zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/CassandraClient.java
michaelsembwever: they are appropriate for first-time / dev usage. but not appropriate for production usage. 

it needs to be configurable.
and ideally the application should log a warning when rf<3 saying it's not appropriate for production.

mrproliu: Yes, if rf<3 would be got a warning log. What do you mean configurable?

wu-sheng (approved):
@michaelsembwever I am going to merge this first, to unblock the Cassandra relative changes at the dependency repo side.

When you have time, feel free to review, and we will follow your recommendations.

PR #3573, opened 2023-10-10
Adaptation enhancements for various features in v3

mrproliu:
In this Pull Request, I have made several adaptation enhancements for the following features in the v3:

1. Adapt the data format consistency for the Collector Sample Rate ratio configuration.
2. Introduced a configuration for Strict Trace ID during queries.
3. Adapted the log format (although some colors could not be matched) and the Banner.
4. Exposed Health-related interfaces in HTTP Server queries.

### DIFF
--- a/zipkin-server/health-query-plugin/src/main/java/zipkin/server/query/health/ZipkinHealthController.java
+++ b/zipkin-server/health-query-plugin/src/main/java/zipkin/server/query/health/ZipkinHealthController.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2015-2023 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin.server.query.health;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.MediaType;
+import com.linecorp.armeria.server.ServiceRequestContext;
+import com.linecorp.armeria.server.annotation.Get;
+import io.vavr.collection.Stream;
+import org.apache.skywalking.oap.server.library.module.ModuleManager;
+import org.apache.skywalking.oap.server.library.module.ModuleServiceHolder;
+import org.apache.skywalking.oap.server.telemetry.TelemetryModule;
+import org.apache.skywalking.oap.server.telemetry.api.MetricFamily;
+import org.apache.skywalking.oap.server.telemetry.api.MetricsCollector;
+import org.apache.skywalking.oap.server.telemetry.api.MetricsCreator;
+
+import java.io.IOException;
+import java.io.StringWriter;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class ZipkinHealthController {
+  static final String STATUS_UP = "UP", STATUS_DOWN = "DOWN";
+
+  final MetricsCollector collector;
+  final private MetricsCreator metricsCreator;
+
+  public static final MediaType MEDIA_TYPE_ACTUATOR =
+      MediaType.parse("application/vnd.spring-boot.actuator.v2+json;charset=UTF-8");
+
+  public ZipkinHealthController(ModuleManager moduleManager) {
+    ModuleServiceHolder telemetry = moduleManager.find(TelemetryModule.NAME).provider();
+    metricsCreator = telemetry.getService(MetricsCreator.class);
+    collector = telemetry.getService(MetricsCollector.class);
+  }
+
+  @Get("/actuator/health")
+  public HttpResponse getActuatorHealth(ServiceRequestContext ctx) {
+    return newHealthResponse(MEDIA_TYPE_ACTUATOR);
+  }
+
+  @Get("/health")
+  public HttpResponse getHealth(ServiceRequestContext ctx) {
+    return newHealthResponse(MediaType.JSON_UTF_8);
+  }
+
+  HttpResponse newHealthResponse(MediaType mediaType) {
+    final Map<MetricFamily.Sample, String> componentsHealth = Stream.ofAll(collector.collect())
+        .flatMap(metricFamily -> metricFamily.samples)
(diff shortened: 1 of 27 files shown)


zipkin-server/server-starter/src/main/java/zipkin/server/config/ApplicationConfigLoader.java
```diff
@@ -119,7 +119,12 @@ private void replacePropertyAndLog(final String propertyName, final Object prope
     private Object convertValueString(String valueString) {
         try {
             Object replaceValue = yaml.load(valueString);
-            if (replaceValue instanceof String || replaceValue instanceof Integer || replaceValue instanceof Long || replaceValue instanceof Boolean || replaceValue instanceof ArrayList) {
+            if (replaceValue instanceof String ||
```
wu-sheng, line 122: I am wondering, why do we need to copy `ApplicationConfigLoader`? Is it possible to use `ZipkinServerBootstrap` calling the existing codes from SkyWalking starter codes?

zipkin-server/health-query-plugin/src/main/java/zipkin/server/query/health/ZipkinHealthController.java
```diff
@@ -0,0 +1,110 @@
...
+import org.apache.skywalking.oap.server.telemetry.api.MetricFamily;
+import org.apache.skywalking.oap.server.telemetry.api.MetricsCollector;
+import org.apache.skywalking.oap.server.telemetry.api.MetricsCreator;
+
+import java.io.IOException;
+import java.io.StringWriter;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class ZipkinHealthController {
```
wu-sheng: I think from a naming style perspective, this should be `ZipkinHealthHandler`

zipkin-server/server-starter/src/main/java/zipkin/server/config/ApplicationConfigLoader.java
mrproliu: It follows these two reasons:
1. If further customization of configuration files is needed later on, this file allows for easy re-customization.
2. If based on the `server-starter` project, which includes all module projects, we need to remove dependencies sequentially. When new modules are introduced in Skywalking OAP and synchronized to Zipkin later, we still need to remove unrelated dependencies.

wu-sheng: > If further customization of configuration files is needed later on, this file allows for easy re-customization.

The biggest concern is duplicate `ApplicationConfigLoader`. It has complex logic and very general. Should we consider moving this class from OAP starter to core and make it available easily here?

mrproliu: This means that we need to put all the application startup configuration readers in the core. I feel like this might cause some confusion? Because the core doesn't need to read configuration files.

wu-sheng: OK. Let's keep the duplicate codes for now. Personally, I don't like we copied the same codes.

wu-sheng (approved):

zipkin-server/receiver-zipkin-http/src/main/java/zipkin/server/receiver/zipkin/http/ZipkinHTTPReceiverConfig.java
```diff
@@ -17,58 +17,5 @@
 import org.apache.skywalking.oap.server.library.module.ModuleConfig;
 
 public class ZipkinHTTPReceiverConfig extends ModuleConfig {
-  private String restHost;
-  private int restPort;
-  private String restContextPath;
-  private int restMaxThreads = 200;
-  private long restIdleTimeOut = 30000;
```
wu-sheng: What does this removing mean? Could you explain the diff between `receiver-zipkin-http` and `http-query-plugin`. They seem to be consufed.

mrproliu: First, in the v2 version of Zipkin, they share the same HTTP port when receiver and query. So I moved the HTTP configuration in the HTTP Receiver and Querier into the core. 
The receiver is used to collect spans from agents, and the Query is used to query the span for UI.

wu-sheng: I think that is how sharing server works. The receiver is still able to work on the separate port, especially query port is exposed for the internet, I assume a separate port is better.
Could you check the OAP sharing server and take a reference to optimize the receiver?

mrproliu: I have separated the export ports for the receiver and querier modules (query span and health). Now, they can utilize the default HTTP server in the core module if the port is defined as `-1`. Otherwise, the module will start a new HTTP server and use it.

wu-sheng (approved):

zipkin-server/http-query-plugin/src/main/resources/META-INF/services/org.apache.skywalking.oap.server.library.module.ModuleProvider
```diff
@@ -0,0 +1,19 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
wu-sheng, line 2: The header is somehow wrong, please fix it. And consider skywalking-eyes to verify headers. Notice, I am not sure year range is supported or not.

mrproliu: > The header is somehow wrong, please fix it. And consider skywalking-eyes to verify headers.

I will create another PR to fix all the files that I created. 

>  Notice, I am not sure year range is supported or not.

I remember It's not supported, as [JC says](https://github.com/openzipkin/zipkin/pull/3567#issuecomment-1725182252), we don't need to align the year, so, we don't need to check the year. Alternatively, maybe we can only check the files involved in the Pull Request. However, I'm not sure if this is possible.

wu-sheng: > I remember It's not supported, as https://github.com/openzipkin/zipkin/pull/3567#issuecomment-1725182252, we don't need to align the year,

Not that. I mean I don't know whether skywalking-eyes support to check various start-end year range in all file headers.

mrproliu: Got it, The eyes could [verify the header by regex](https://github.com/apache/skywalking-eyes/blob/main/.licenserc.yaml#L40-L57), Therefore, I believe it is capable of verifying the year range.

PR #3584, opened 2023-11-07
Enhance the query performance and storage size of the Cassandra storage in v3 server

mrproliu:
1. Improve collect throughput: Change the synchronous writing of batch data synchronization to asynchronous writing.
2. Reducing Unnecessary Fields: Reduce unnecessary field storage to reduce storage space usage.
3. Query Optimization: Split the original query method into multiple files, so that each query can be optimized better.

### DIFF
--- a/zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/CassandraTableInstaller.java
+++ b/zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/CassandraTableInstaller.java
@@ -14,10 +14,13 @@
 
 package zipkin.server.storage.cassandra;
 
+import org.apache.skywalking.oap.server.core.storage.StorageData;
 import org.apache.skywalking.oap.server.core.storage.StorageException;
 import org.apache.skywalking.oap.server.core.storage.model.Model;
 import org.apache.skywalking.oap.server.core.storage.model.ModelCreator;
+import org.apache.skywalking.oap.server.core.zipkin.ZipkinSpanRecord;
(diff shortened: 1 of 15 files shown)


zipkin-server/storage-cassandra/src/main/java/zipkin/server/storage/cassandra/CassandraTableInstaller.java
```diff
@@ -43,6 +46,17 @@ public void start() {
 
   @Override
   public void whenCreating(Model model) throws StorageException {
+    if (ZipkinSpanRecord.INDEX_NAME.equals(model.getName())) {
+      // remove unnecessary columns
+      for (int i = model.getColumns().size() - 1; i >= 0; i--) {
+        final String columnName = model.getColumns().get(i).getColumnName().getStorageName();
+        if (StorageData.TIME_BUCKET.equals(columnName) ||
+            ZipkinSpanRecord.TIMESTAMP_MILLIS.equals(columnName) ||
+            JDBCTableInstaller.TABLE_COLUMN.equals(columnName)) {
```
wu-sheng, line 55: CassandraTableInstaller is referring to the JDBC implementation? Where does this TABLE COLUMN come from?

mrproliu: > CassandraTableInstaller is referring to the JDBC implementation?

Nope, For now, the table installer only caches the table metadata. 

> Where does this TABLE COLUMN come from?

There are definitions in skywalking core that are not necessary for the query.

PR #3594, opened 2023-11-15
Merge Zipkin v3 to the main branch.

wu-sheng:
We are planning to run a 3.0.0-rc0 release as we have finished all features covered in v3.\

The key features are as follows

* Newly support for all versions of Elasticsearch 8 and OpenSearch 1.x. And keep supports for Elasticsearch 6 and 7.
* Rapid support for the latest JDK releases, and fix for CVEs of dependencies.
* Refactored storage structures and query logic for Elasticsearch to improve performance. Elasticsearch. Reduce 20% disk volume and 40% query latency on the 2 million spans case.
* SQL Databases(MySQL and PostgreSQL) storage options are ready and stable for the product ready in the medium-scale deployment, or larger-scale when picking SQL Database mutations, such as Amazon Aurora.
* Introduce cluster mode deployment in Zipkin v3. It reduces the memory requirements for the caches of service names, tags, and local/remote endpoints. It doesn’t require an extra deployment of Elastic Curator for Zipkin data time-to-live(TTL) in the Elasticsearch storage. SQL Databases(MySQL and PostgreSQL with their mutations) storage options support TTL for the first time.
* Support for Apache BanyanDB (Alpha stage) as a new storage option, which is expected to offer better compression rates, lesser memory usage, and less CPU consumption compared to Elasticsearch data capacity(large-scale deployment) scenarios.
* Automatic conversion of OpenTelemetry trace format into Zipkin via gRPC and storing it as Zipkin native spans, eliminating the need for an OpenTelemetry Collector relay.
* A more powerful standalone memory mode, replacing the previous simple memory structure with H2 database, and supporting H2 DB Server for lightweight deployment.
* Enable self-observability of the Zipkin server, trace querying, and span reporting by binding them to different HTTP IP addresses and ports. It provides more flexibility in a complex and security-sensitive environment.


docker/Dockerfile
```diff
@@ -79,7 +79,7 @@ LABEL org.opencontainers.image.description="Zipkin slim distribution on OpenJDK
 
 COPY --from=install --chown=${USER} /install/zipkin-slim/ /zipkin/
 
-EXPOSE 9411
+EXPOSE 9411 9412
```
jcchavezs, line 82: Curious about this.

build-bin/maven/maven_release
```diff
@@ -40,4 +41,4 @@ if [ "$commit_local_release_branch" != "$commit_remote_release_branch" ]; then
 fi
 
 # Prepare and push release commits and the version tag (N.N.N), which triggers deployment.
-./mvnw --batch-mode -nsu -DreleaseVersion=${release_version} -Denforcer.fail=false -Darguments="-DskipTests -Denforcer.fail=false" release:prepare
+./mvnw --batch-mode -nsu -DreleaseVersion=${release_version} -Denforcer.fail=false -Darguments="-DskipTests -Denforcer.fail=false -Dcheckstyle.skip=true" release:prepare
```
jcchavezs, line 44: Curious about this, are we skipping the check style?

docker/README.md
```diff
@@ -38,8 +38,8 @@ We also provide [example compose files](examples/README.md) that integrate colle
 such as Kafka or Elasticsearch.
 
 ## Configuration
-Configuration is via environment variables, defined by [zipkin-server](https://github.com/openzipkin/zipkin/blob/master/zipkin-server/README.md). Notably, you'll want to look at the `STORAGE_TYPE` environment variables, which
-include "cassandra", "mysql" and "elasticsearch".
+Configuration is via environment variables, defined by [zipkin-server](https://github.com/openzipkin/zipkin/blob/master/zipkin-server/README.md). Notably, you'll want to look at the `ZIPKIN_STORAGE` environment variables, which
```
jcchavezs, line 41: While I don't question the naming change I want to make sure we support BOTH to ease migration.

docker/README.md
```diff
@@ -69,7 +69,7 @@ For example, to increase heap size, set `JAVA_OPTS` as shown in our [docker-comp
 
 For example, to add debug logging, set `command` as shown in our [docker-compose](examples/docker-compose.yml) file:
 ```yaml
-    command: --logging.level.zipkin2=DEBUG
+    command: -Dlog.level=DEBUG
```
jcchavezs, line 72: Can we support previous option too?

wu-sheng: This seems a little hard. log4j V.S. log4js. Unless we hard coded some hacking codes.

docker/README.md
wu-sheng: @mrproliu We could change this back. I think.

jcchavezs:
Changes look OK to me but I will wait for others to review. 

The only thing that catched my eye strongly was the env vars renaming. While this makes sense in some cases that I could identify, we must make sure migration isn't that much of a pain to users (that will have to drop their existing database anyways) and make sure old zipkin config remains working. One of the most important principles in entire zipkin ecosystem was to not to break sites even on major version changes, instead deprecate them in a major and remove them in the next major.

basvanbeek:
So let's not rush to land this in main branch. I'd rather do release candidates with some real effort of reaching out to community and site owners (including a serious mention about this on our website). 

When we've had sufficient positive feedback cycles, concerns dealt with, enough broadcasting to reach our users we can talk about planning an official release and moving to main branch.

basvanbeek (changes requested):
Let's give this more time and more communication around this effort. We need actual users to try and respond before an official release. 

This also means providing notice and broadcasting about this effort on website, gitter, google groups, etc. As well as providing RC builds to be downloadable from site, docker hub, etc.

wu-sheng:
> So let's not rush to land this in main branch. I'd rather do release candidates with some real effort of reaching out to community and site owners (including a serious mention about this on our website).

So, is it fine to release one from a branch? If this is OK, we can work on that way.

basvanbeek:
> > So let's not rush to land this in main branch. I'd rather do release candidates with some real effort of reaching out to community and site owners (including a serious mention about this on our website).
> 
> So, is it fine to release one from a branch? If this is OK, we can work on that way.

IMHO it is fine to do release candidates / tech previews from a branch. Not sure how much work it will be to make it available. I just don't feel it's right to merge to main branch prior to the broader community approval.

I want this effort to be correct and supported by the broader community. The impact is too large to swiftly go through without giving more time and consideration.

wu-sheng:
> IMHO it is fine to do release candidates / tech previews from a branch. Not sure how much work it will be to make it available.

OK, let me do this. (Release RCs from the branch)

jcchavezs:
Once you can release an image we could test the automation in
https://github.com/openzipkin/zipkin-php-example as smoke test first.

José Carlos Chávez


ons. 15. nov. 2023 kl. 10:45 skrev 吴晟 Wu Sheng ***@***.***>:

docker/Dockerfile
mrproliu: I will delete it.

build-bin/maven/maven_release
mrproliu: At present, zipkin does not include checkstyle. However, the newly introduced skywalking will search for and utilize the checkstyle configuration from the current project. So, I am currently disabling skywalking's checkstyle execution by deactivating it.

docker/README.md
mrproliu: Got it, I will make these environments both work.

PR #3595, opened 2023-11-16
Adapt v2 environment keys

mrproliu:
This PR is used to adapt in the environment variables of v2 and v3 versions.
Currently, the following two types of environment variables are not supported:
1. XXX_ENABLED: Previously only "true" or "false", now it is "default" or "-".
2. Environment variables that do not exist in the new version, such as `mysql.host`.


wu-sheng:
Are docs required to change?

mrproliu:
I will write the previously compatible environment keys in the document together.

PR #3597, opened 2023-11-16
Allowing Release from Specific Branches

mrproliu:
Currently, the release process is initiated from the master branch. However, when attempting to create a new release for the `zipkin-v3` branch, the process still defaults to using the master branch for publication. 
In this pull request, I've modified the workflow to checkout the branch intended for release, ensuring releases are initiated from the correct branch.

### DIFF
--- a/build-bin/maven/maven_release
+++ b/build-bin/maven/maven_release
@@ -25,19 +25,10 @@ export MAVEN_OPTS="$($(dirname "$0")/maven_opts)"
 
 trigger_tag=${1?trigger_tag is required. Ex release-1.2.3}
 release_version=$(build-bin/git/version_from_trigger_tag release- ${trigger_tag})
-release_branch=${2:-master}
 
-# Checkout master, as we release from master, not a tag ref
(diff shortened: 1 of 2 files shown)


wu-sheng (approved):
LGTM. As @basvanbeek suggested, we could release RC(x) from other branch, such as zipkin-ve. The release script should be changed.

build-bin/maven/maven_release
```diff
@@ -25,19 +25,10 @@ export MAVEN_OPTS="$($(dirname "$0")/maven_opts)"
 
 trigger_tag=${1?trigger_tag is required. Ex release-1.2.3}
 release_version=$(build-bin/git/version_from_trigger_tag release- ${trigger_tag})
-release_branch=${2:-master}
```
wu-sheng, line 28: Once we lost the branch information, we can't bump up the latest of codes. The script may exit with error status.

mrproliu: I have modified the branch selection policy here. It will now only extract the branch where the last commit within the current tag.

PR #3606, opened 2023-12-02
Updates to latest maven and plugins

codefromthecrypt:
This updates to the latest maven 3.9.6 and removes the takari wrapper plugin, as that's been obviated upstream for three years. This also bumps all the maven plugins used, which is good as many of them were pre-release versions formerly.

This intentionally only does maven stuff in one PR. There is a lot of out-of-date deps that really need to be addressed before a patch of the current codebase before doing anything else. This is just one step. Next will include test and main deps as well docker and github actions stuff, in any order.


pom.xml
```diff
@@ -522,7 +512,7 @@
             <configuration>
               <rules>
                 <requireJavaVersion>
-                  <version>[1.8,16)</version>
+                  <version>[1.8,22)</version>
```
anuraaga: Just curious if you found this in a release note or somewhere. A comment on how to update this could help the next person updating maven but not required

anuraaga (approved):

pom.xml
codefromthecrypt: no problem. I mainly intended to make sure we don't fail on the latest LTS.

codefromthecrypt: actually I will revert this part and do 21 support in a separate PR

PR #3607, opened 2023-12-02
ci: sets supported JDK range explicitly to 11,15

codefromthecrypt:
Before, we were using an implicit version of the JDK when releasing, had
the pom.xml limited to a non-LTS version (15).

This sets the same up explicitly in CI, and corrects CI to use latest
action versions and ubuntu. A later change should update the ceiling
build version to an LTS, at least 17 if not 21. Currently, we can't as
there is a compilation failure. I'll look into it personally in the next
PR.

### DIFF
--- a/.github/workflows/create_release.yml
+++ b/.github/workflows/create_release.yml
@@ -11,17 +11,22 @@ on:
 
 jobs:
   create_release:
-    runs-on: ubuntu-20.04  # newest available distribution, aka focal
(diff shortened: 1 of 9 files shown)


codefromthecrypt:
draft until green

anuraaga (approved):

codefromthecrypt:
thanks for the look yet again @anuraaga!

PR #3610, opened 2023-12-04
updates to the latest test dependencies

codefromthecrypt:


zipkin-storage/mysql-v1/src/test/java/zipkin2/storage/mysql/v1/MySQLExtension.java
```diff
@@ -113,7 +113,7 @@ int port() {
   // mostly waiting for https://github.com/testcontainers/testcontainers-java/issues/3537
   static final class MySQLContainer extends GenericContainer<MySQLContainer> {
     MySQLContainer() {
-      super(parse("ghcr.io/openzipkin/zipkin-mysql:2.23.2"));
+      super(parse("ghcr.io/openzipkin/zipkin-mysql:2.24.3"));
```
codefromthecrypt, line 116: this moves to the last published docker tag (well ignoring the accidental v3 one..). Doing this gets the recent cut of test images, even though they are a bit behind still. I'll update those later.

codefromthecrypt:
depends on #3612 and while at it I will look at all the images before bumping this one out of draft

benchmarks/src/test/java/zipkin2/server/ServerIntegratedBenchmark.java
```diff
@@ -91,7 +91,7 @@ class ServerIntegratedBenchmark {
```
codefromthecrypt, line 91: Note: this bench doesn't complete yet, but I can work on that in another PR. Good news is it at least runs on Apple Silicon now!

zipkin-storage/mysql-v1/src/test/java/zipkin2/storage/mysql/v1/MySQLExtension.java
codefromthecrypt: I refreshed these

anuraaga (approved):

PR #3612, opened 2023-12-04
cassandra: updates client driver and test binary

codefromthecrypt:
This updates to the latest Datastax driver and updates our test image to latest cassandra 4.x version. This also cleans up where our test image had leftovers from previous versions of cassandra and JREs.

I'm intentionally not updating the jre to a newer version, yet, because we should make sure things work on 17.

### DIFF
--- a/docker/test-images/zipkin-cassandra/install.sh
+++ b/docker/test-images/zipkin-cassandra/install.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 #
-# Copyright 2015-2021 The OpenZipkin Authors
+# Copyright 2015-2023 The OpenZipkin Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
 # in compliance with the License. You may obtain a copy of the License at
@@ -39,15 +39,7 @@ cat > pom.xml <<-'EOF'
       <version>${cassandra.version}</version>
       <exclusions>
         <exclusion>
-          <groupId>com.github.jbellis</groupId>
-          <artifactId>*</artifactId>
-        </exclusion>
-        <exclusion>
           <groupId>net.java.dev.jna</groupId>
-          <artifactId>*</artifactId>
-        </exclusion>
-        <exclusion>
-          <groupId>com.googlecode.concurrent-trees</groupId>
           <artifactId>*</artifactId>
         </exclusion>
         <exclusion>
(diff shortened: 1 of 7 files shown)


docker/test-images/zipkin-cassandra/install.sh
```diff
@@ -133,45 +113,35 @@ log4j.appender.stdout.Target=System.out
...
+# Run cassandra on a different port temporarily in order to setup the schema.
+# We also add exports and opens from Cassandra 4, except RMI, which isn't in our JRE image.
+# See https://github.com/apache/cassandra/blob/cassandra-4.0.11/conf/jvm11-server.options
+java -cp 'classes:lib/*' -Xms64m -Xmx64m -XX:+ExitOnOutOfMemoryError -verbose:gc \
+  -Djdk.attach.allowAttachSelf=true \
+  --add-exports java.base/jdk.internal.misc=ALL-UNNAMED \
   --add-exports java.base/jdk.internal.ref=ALL-UNNAMED \
   --add-exports java.base/sun.nio.ch=ALL-UNNAMED \
   --add-exports java.sql/java.sql=ALL-UNNAMED \
+  --add-opens java.base/java.lang=ALL-UNNAMED \
```
anuraaga: This one seems new though util/concurrent/concurrent.atomic aren't so maybe it's ok, but I was surprised since they seem to open packages that are already public.

I don't see them in the linked jvm11-server.options, if they're needed, is there a way to document how to find out what to write here?

anuraaga (commented):
Basically LGTM but want to confirm a point for future maintainers

docker/test-images/zipkin-cassandra/install.sh
codefromthecrypt: I'll try without it.. tests were crashing at runtime, but possibly something dirty on my end.

codefromthecrypt: thanks for being jiminy cricket! found it. basically cassandra 4 doesn't formally support JRE 17, and this workaround is in 5 (which we've not added yet). I will add the same comment

anuraaga (approved):

docker/test-images/zipkin-cassandra/install.sh
codefromthecrypt: Added a commit that may make sense. Main idea is to keep hunting down since this is a test image. Merge in anything cassandra 5 needs for JRE 17+, so that we can use the same base for cassandra 4 without a lot of sleuthing.

codefromthecrypt:
fyi I just noticed cassandra 3.x (which we don't have a test image for anymore) dropped out of support. We may want to reap that at some point (config etc that only discusses it) perhaps after adding v5.

See https://cassandra.apache.org/_/download.html

PR #3613, opened 2023-12-05
docker: moves everything to consistent docker base layer

codefromthecrypt:
This moves to consistent Alpine 3.18.5 and Java 17.0.9_p8. This updates test images to the latest minor or patch version of the corresponding software (e.g. Elasticsearch to latest 6.x). This does not change any major version, including the JRE. We can do these as a separate task.

### DIFF
--- a/build-bin/docker_push
+++ b/build-bin/docker_push
@@ -10,11 +10,18 @@ case ${version} in
     ;;
 esac
 
+# Don't attempt ppc64le until there's a package for recent LTS versions.
+# See https://github.com/openzipkin/zipkin/issues/3443
+export DOCKER_ARCHS="amd64 arm64 s390x"
+
 build-bin/docker/docker_push openzipkin/zipkin ${version}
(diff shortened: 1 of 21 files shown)


build-bin/docker_push
```diff
@@ -10,11 +10,18 @@ case ${version} in
...
+# See https://github.com/openzipkin/zipkin/issues/3443
+export DOCKER_ARCHS="amd64 arm64 s390x"
+
 build-bin/docker/docker_push openzipkin/zipkin ${version}
 DOCKER_TARGET=zipkin-slim build-bin/docker/docker_push openzipkin/zipkin-slim ${version}
 
 # testing images only push to ghcro.io
 export DOCKER_RELEASE_REPOS=ghcr.io
+# Don't attempt unfamiliar archs on test images
+export DOCKER_ARCHS="amd64 arm64"
```
codefromthecrypt, line 23: @llinder fyi, this is the maybe non-obvious trick to reduce arch for some images

PR #3614, opened 2023-12-05
deps: updates to armeria 1.26.3 and netty 4.1.100.Final

### LINKED ISSUE #3556: CVE-2023-38493 on direct dependency, com.linecorp.armeria:armeria
## Describe the Bug
In our vulnerability scans zipkin is getting flagged for [CVE-2023-38493](https://nvd.nist.gov/vuln/detail/CVE-2023-38493) due to the direct dependency on `com.linecorp.armeria:armeria`

Can `com.linecorp.armeria:armeria` be upgraded to 1.24.3 at-least?

codefromthecrypt:
Note: netty.version aligns with armeria

Fixes #3556

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,9 +52,9 @@
 
     <!-- This allows you to test feature branches with jitpack -->
     <armeria.groupId>com.linecorp.armeria</armeria.groupId>
-    <armeria.version>1.17.2</armeria.version>
+    <armeria.version>1.26.3</armeria.version>
     <!-- Match Armeria version to avoid conflicts including running tests in the IDE -->
-    <netty.version>4.1.95.Final</netty.version>
+    <netty.version>4.1.100.Final</netty.version>
 
     <!-- It's easy for Jackson dependencies to get misaligned, so we manage it ourselves. -->
     <jackson.version>2.16.0</jackson.version>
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
I'll bump the other main deps in another PR, as we're better off having everything updated vs whack-a-mole.

That said, I wanted to salvage @darkmastermindz's work here first.

pom.xml
```diff
@@ -52,15 +52,15 @@
 
     <!-- This allows you to test feature branches with jitpack -->
     <armeria.groupId>com.linecorp.armeria</armeria.groupId>
-    <armeria.version>1.17.2</armeria.version>
+    <armeria.version>1.26.3</armeria.version>
```
codefromthecrypt, line 55: ITElasticsearchClientInitialization seems to not behave nicely. I know a lot has changed between these two here. @openzipkin/armeria does anyone have time to diagnose?

https://github.com/openzipkin/zipkin/actions/runs/7099979222/job/19325100225?pr=3614

codefromthecrypt: cc also @ikhoon who isn't in that list but maybe should be as has helped in the past!

codefromthecrypt:
tomorrow I will raise a PR to bump all deps except armeria/netty, so that we can close the gaps on differences to just this, without worry that some other incidental dep is impacting the topic.

codefromthecrypt:
https://github.com/openzipkin/zipkin/pull/3615 updates the other things and I'll rebase this afterwards

codefromthecrypt:
pared down this PR to only include armeria+netty

codefromthecrypt:
so I checked and 1.17.2 was the highest version to not fail these health-check tests, so something changed after that

codefromthecrypt:
I think I maybe found it

pom.xml
codefromthecrypt: nevermind I think we accidentally nuked use of a timeout and perhaps it was accidentally the correct value by default

codefromthecrypt:
I noticed that the logs have more junk than usual, not related to this, more about spring. like ignoring our banner override in tests. I'll look into this stuff afterwards.

PR #3615, opened 2023-12-05
deps: updates main deps except armeria and amqp

codefromthecrypt:
This updates all main dependencies except armeria (handled #3614) and amqp (driver is incompatible so needs a new PR).


zipkin-collector/rabbitmq/pom.xml
```diff
@@ -29,6 +29,7 @@
 
   <properties>
     <main.basedir>${project.basedir}/../..</main.basedir>
+    <!-- TODO: this version is abandoned: we have to rewrite for 5.x to avoid CVEs -->
```
codefromthecrypt, line 32: needs a big job, but not a huge one

zipkin-storage/mysql-v1/pom.xml
```diff
@@ -29,7 +29,9 @@
   <properties>
     <main.basedir>${project.basedir}/../..</main.basedir>
 
-    <jooq.version>3.14.4</jooq.version>
+    <!-- TODO: upgrade when we make floor JDK 11 or 17
```
codefromthecrypt, line 32: interesting that jooq requires at least JDK 17 now ;)

codefromthecrypt (commented):
notes

PR #3617, opened 2023-12-05
[wip] Update Node version to 20.10.0 LTS

tacigar:


zipkin-lens/pom.xml
```diff
@@ -34,7 +34,7 @@
     <!-- Update occasionally based on LTS, but don't overrun what's in Alpine:
          https://nodejs.org/en/download/
          https://pkgs.alpinelinux.org/packages?name=nodejs&branch=edge -->
-    <node.version>14.21.3</node.version>
+    <node.version>v20.10.0</node.version>
```
codefromthecrypt: maybe..

```suggestion
    <node.version>20.10.0</node.version>
```

tacigar:
@codefromthecrypt In newer NPM versions, dependency resolution seems to be stricter. So it seems that many libraries need to be updated in order to pass the CI.
I'll update library versions first in other PRs since I don't want to include them in this PR.

tacigar:
This was take over by https://github.com/openzipkin/zipkin/pull/3658

PR #3618, opened 2023-12-06
Moves to only support LTS JDKs

codefromthecrypt:
This fixes errorprone configuration blocking use of JDK 17. At this point, we no longer use a non-LTS JDK 15 to compile.

This also introduces CI to ensure JDK 11 still works. This is needed because some libraries such as Zulu aggressively age out JDKs.

As a reminder, we currently target 1.6 for the core jar on release, so that zipkin-reporter can still work in old clients. We may need to vendor those classes in zipkin-reporter-java to unlock things in the future.

To reduce scope this also removes CI from the no-longer active v3 (skywalking) branch.


codefromthecrypt:
draft until green

anuraaga (approved):

PR #3624, opened 2023-12-11
deps: uses bom to manage log4j and removes explicit snappy override

codefromthecrypt:
snappy was documented as being brought in by spring, but it was kafka and is now current. This also corrects how versions were declared to override spring defaults, using dependencyManagement. Only one dep is oddly overridden by us, snakeyaml, so is now documented as such.


anuraaga (approved):

codefromthecrypt:
thanks for keeping an eye out, @anuraaga!

PR #3628, opened 2023-12-12
Update libraries without breaking changes

tacigar:


codefromthecrypt:
@timtebeek if you'all end up being able to do something like this with openrewrite, joy will bellow from the heavens ;)

timtebeek:
On [the roadmap](https://docs.moderne.io/administrator-documentation/references/language-roadmap)! :) Not there yet, but promising work being done internally. Hope to be able to share more in 2024.

codefromthecrypt (approved):

codefromthecrypt:
thanks for the progress @tacigar!

PR #3632, opened 2023-12-13
docker: works around zipkin-rabbitmq arm64 build fail in GHA deploy

codefromthecrypt:
Currently, the docker deployment is failing on zipkin-rabbitmq on arm64. This isn't reproducable locally on either amd64 of arm64, but basically the rabbitmq-server crashes Dockerfile in an invalid way when building arm64:

```
BOOT FAILED
===========
Error during startup: {error,
                          {incompatible_feature_flags,
                              {disabled_required_feature_flag,
                                  classic_mirrored_queue_version}}}
```

What's invalid is that the same build which includes those features works outside the GHA runner. After spending a good bit of time on this, I gave up. Maybe a later version of qemu will sort it out.

Meanwhile, I switched the parent layer to the official docker image, and also switched to a configuration approach to define the default queue. I also noticed we have a bug in our collector logic where rabbit must be running prior to zipkin starting. This is a bug and I'll fix it later, but meanwhile it can be worked around with a dependency order in the example docker-compose file.


codefromthecrypt:
opened an issue on the unrelated ci flake https://github.com/openzipkin/zipkin/issues/3633

anuraaga (approved):

codefromthecrypt:
Thanks for the offline advice @anuraaga!

PR #3642, opened 2023-12-13
ci: re-enables license plugin

codefromthecrypt:
The license plugin was disabled under the rationale that it was unclear why it needed to touch untouched files. There are two scenarios:

* first release of the year requires pom license updates per RELEASE.md
* if you merge an old pull request, which was made the year prior, you can get late failures on master. Best way out is to merge quickly, as this is good behaviour anyway.

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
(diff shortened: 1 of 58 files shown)


codefromthecrypt:
first run resulted in commits that wrote files with 2022, based on history, but since it is 2023 now I think a second run was needed as the commit itself touched the files. 🤞 this works now

anuraaga (approved):

PR #3643, opened 2023-12-13
tests: removes eager upgrade to okhttp alpha 5

codefromthecrypt:
the junit5 extension was written to depend on okhttp5, which hasn't been updated in a year and introduces classpath conflicts. This reverts that, so that the extension is useable in projects that use okhttp and would otherwise conflict on the okio dep okhttp5 brings in.

We need to release a patch after this to unlock zipkin-reporter

### DIFF
--- a/zipkin-junit5/pom.xml
+++ b/zipkin-junit5/pom.xml
@@ -61,16 +61,13 @@
 
     <dependency>
       <groupId>com.squareup.okhttp3</groupId>
-      <artifactId>mockwebserver3-junit5</artifactId>
-      <version>${okhttp5.version}</version>
-      <exclusions>
-        <!-- manage our own jupiter version -->
-        <exclusion>
-          <groupId>org.junit.jupiter</groupId>
-          <artifactId>*</artifactId>
-        </exclusion>
-      </exclusions>
+      <!-- Intentionally not the alpha mockwebserver3 until it can coexist with
+           okhttp 3, avoiding:
+           java.lang.NoSuchMethodError: 'void okhttp3.internal.concurrent.TaskRunner.<init>(okhttp3.internal.concurrent.TaskRunner$Backend, java.util.logging.Logger, int, kotlin.jvm.internal.DefaultConstructorMarker)' -->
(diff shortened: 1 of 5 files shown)


zipkin-junit5/pom.xml
```diff
@@ -46,16 +46,10 @@
 
     <dependency>
       <groupId>com.squareup.okhttp3</groupId>
-      <artifactId>mockwebserver3-junit5</artifactId>
```
codefromthecrypt, line 64: not sure why this dep was used because we weren't using any jupiter types defined in this package..

codefromthecrypt:
I checked in zipkin-reporter that this works even with okhttp 3.14.9 in the classpath.

codefromthecrypt:
closing this out as the root cause was an unnecessary dep. The benefit of this is that it excludes the junit:junit dep. After a close look while okhttp isn't releasing as often as before, it is being worked on in the main branch, so not a big deal to keep as is.

codefromthecrypt:
https://github.com/openzipkin/zipkin-reporter-java/pull/221 shows it works as is in the reporter, and I don't expect any issues with zipkin-finagle that also uses it in our org.

helidon which is the other main consumer doesn't appear to have a conflict and we can re-open this if that's unresolvable. I can see in their code they are already working around okhttp kotlin stuff, so they may be ok resolving it, and if not we can re-open this.

codefromthecrypt:
Brave doesn't use this, but in brave, using mockwebserver3 creates an irresolvable classpath conflict on below:

```
java.lang.NoSuchMethodError: 'void okhttp3.internal.concurrent.TaskRunner.<init>(okhttp3.internal.concurrent.TaskRunner$Backend, java.util.logging.Logger, int, kotlin.jvm.internal.DefaultConstructorMarker)'
```

This should be reverted because the cure is worse than the disease (junit:junit dep which is already present in most apps). We can revisit when okhttp releases again

codefromthecrypt:
rebased and added comment

anuraaga (approved):

PR #3644, opened 2023-12-14
refactor: Drop prefixless expressions in POM

codefromthecrypt:
Use this link to re-run the recipe: https://app.moderne.io/recipes/org.openrewrite.maven.cleanup.PrefixlessExpressions?organizationId=T3BlbiBaaXBraW4%3D

### DIFF
--- a/zipkin-lens/pom.xml
+++ b/zipkin-lens/pom.xml
@@ -331,7 +331,7 @@
                 </goals>
                 <configuration>
                   <classifier>javadoc</classifier>
-                  <classesDirectory>${basedir}/javadoc</classesDirectory>
+                  <classesDirectory>${project.basedir}/javadoc</classesDirectory>
                 </configuration>
               </execution>
             </executions>


codefromthecrypt:
@timtebeek is there a way to have the default template suggest the maven command to achieve the same? I think the login is nice, but it isn't good for having history of how this was done in the projects issues and/or commit log.

anuraaga (approved):

timtebeek:
> @timtebeek is there a way to have the default template suggest the maven command to achieve the same? I think the login is nice, but it isn't good for having history of how this was done in the projects issues and/or commit log.

Hmm I agree with you there, but I don't know if this is easy to change in the platform; we support Maven, Gradle, Bazel and other tools, and the instructions would have to differ based on what was used in the project. I've thought of linking to docs.openrewrite.org, but that then just shows an URL quite similar to the app.moderne.io link for this one: https://docs.openrewrite.org/recipes/maven/cleanup/prefixlessexpressions
And while that page does contain instructions for a rerun, I don't think such a link would add very much more details in this particular case that are not available from the Moderne link. Perhaps it would help if we allow access to recipe pages without a login such that the details are immediately visible; that's come up before, but gets tricky with authentication. No perfect solution I guess. :)

timtebeek:
I did notice you ran this one recipe, but we also have a few more Apache Maven best practices documented here: https://docs.openrewrite.org/recipes/maven/bestpractices

Here's a quick run that reorders pom elements and adds missing groupIds and plugin versions: https://app.moderne.io/results/ssUD0Xzgq

PR #3645, opened 2023-12-14
deps: reduces CVEs around jackson and okio

codefromthecrypt:
trivy found these. last thing after this is lens and upgrading the amqp driver.


pom.xml
```diff
@@ -363,9 +364,10 @@
           </executions>
         </plugin>
         <plugin>
-          <groupId>com.squareup.wire</groupId>
+          <!-- com.squareup.wire version was abandoned -->
```
codefromthecrypt, line 367: this was the trick. I don't think cash care much about maven anymore ;)

anuraaga (approved):

codefromthecrypt:
as yoda once said.. At this thanks for looking, rag.

PR #3648, opened 2023-12-14
deps: updates mariadb and armeria to latest

codefromthecrypt:
technically, this version of the mariadb driver is the only one officially supported for JRE 21


zipkin-server/src/main/java/zipkin2/server/internal/mysql/ZipkinMySQLStorageProperties.java
```diff
@@ -112,7 +112,7 @@ private String determineJdbcUrl() {
     }
 
     StringBuilder url = new StringBuilder();
-    url.append("jdbc:mysql://");
+    url.append("jdbc:mariadb://");
```
codefromthecrypt, line 115: mariadb recently enforces the jdbc url name has to be branded as such. That said, our images still install mysql so we test based on that.

anuraaga (approved):

codefromthecrypt:
ok fire in the hole `git push origin release-2.25.1`

PR #3652, opened 2023-12-15
docker: uses health check in examples and documents dependencies job

### LINKED ISSUE #3529: zipkin dependencies graph not work with elasticsearch storage
I am starting to use Zipkin and I have an issue when I run `docker-compose -f docker-compose-elasticsearch.yml up -d ` at docker/examples and the dependencies graph is not displayed when using Elasticsearch but it does work when I use MySQL or memory storage.

could anyone help me, please?
Thank you!

codefromthecrypt:
This uses health check in the docker-compose examples to reduce the amount of warnings end users see.

It also fixes a minor issue on the dependencies job, as well adds instructions for how to run it manually. This is needed for folks who don't want to wait an hour to see the graph.

Fixes #3529

### DIFF
--- a/docker/examples/README.md
+++ b/docker/examples/README.md
@@ -41,6 +41,14 @@ To start the Cassandra-backed configuration, run:
 $ docker-compose -f docker-compose-cassandra.yml up
 ```
 
+The `zipkin-dependencies` container is a scheduled task that runs every hour.
+If you want to see the dependency graph before then, you can run it manually
+in another terminal like so:
+
+```bash
+$ docker-compose -f docker-compose-cassandra.yml run --rm --no-deps --entrypoint start-zipkin-dependencies dependencies
+```
+
 ## Elasticsearch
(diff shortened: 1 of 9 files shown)


docker/examples/README.md
```diff
@@ -41,6 +41,14 @@ To start the Cassandra-backed configuration, run:
 $ docker-compose -f docker-compose-cassandra.yml up
 ```
 
+The `zipkin-dependencies` container is a scheduled task that runs every hour.
+If you want to see the dependency graph before then, you can run it manually
+in another terminal like so:
+
+```bash
+$ docker-compose -f docker-compose-cassandra.yml run --rm --no-deps --entrypoint start-zipkin-dependencies dependencies
```
codefromthecrypt, line 49: Thanks @anuraaga for the tip that led to this!

anuraaga (approved):

PR #3664, opened 2023-12-29
docker: adds Elastic licensed 8.11.3 test image

codefromthecrypt:
This adds an image tagged as elasticsearch8 which is based on our Java base layer, similar to elasticsearch7 except aspects specific to Elasticsearch 8.x.

Notably, Elasticsearch 8.x requires a launcher process, and defines Java options as ES_JAVA_OPTS. Similar to 7.x Elastic requires modules not in our JRE image, so we use the full one instead.

Note: As this is a test image, not a production image, usage should not violate the Elastic Basic license developers typically use, and is the default.


codefromthecrypt:
after this is in, I'll cut a docker tag and take a look at #3552, for example if it works with the minimal modifications in that PR or not

PR #3665, opened 2023-12-29
docker: allows elasticsearch8 tests to delete with wildcards

codefromthecrypt:
I checked and this is the only image change needed for tests to complete. There are a couple small things I need to add to #3552, but they are all client-side.


zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchStorage.java
```diff
@@ -268,6 +268,8 @@ public void clear() throws IOException {
     Set<String> toClear = new LinkedHashSet<>();
     toClear.add(indexNameFormatter().formatType(TYPE_SPAN));
     toClear.add(indexNameFormatter().formatType(TYPE_DEPENDENCY));
+    // Note: Elasticsearch 8.x requires this config to clear with wildcards:
```
codefromthecrypt, line 271: the `clear()` function is only used in tests, and documented as such, if you scroll up.

PR #3666, opened 2023-12-29
Makes all except core jar Java 11+

codefromthecrypt:
This moves all code to minimum Java 11, except the core jar, which remains (mostly) Java 1.6 for use in zipkin-reporter.

Notably, this pushes animal-sniffer to the core module, as the others can rely on the release property of modern compilers.

Java 11 instead of 11 or 21 because that's the last SDK that can also compile 1.6.

Note: this bumps to the next minor as it is worth communicating.


zipkin-storage/mysql-v1/src/main/java/zipkin2/storage/mysql/v1/MySQLSpanConsumer.java
```diff
@@ -86,30 +86,30 @@ public Void apply(DSLContext create) {
         Map<TableField<Record, ?>, Object> updateFields = new LinkedHashMap<>();
         if (timestamp != 0L) {
           // tentatively we can use even a shared timestamp
-          insertSpan.set(ZIPKIN_SPANS.START_TS, timestamp);
+          insertSpan = insertSpan.set(ZIPKIN_SPANS.START_TS, timestamp);
```
codefromthecrypt, line 89: note: changes like this weren't bugs, just the new error-prone plugin is strict about builders and we can't easily affect the jooq code to add the annotation to it. doing this is harmless, except the cruft

PR #3671, opened 2024-01-04
Adds Eureka discovery integration

### LINKED ISSUE #1870: Options for Service Discovery

codefromthecrypt:
Zipkin can register itself in [Eureka](https://github.com/Netflix/eureka), allowing traced services to discover its listen address and health state. This is enabled when `EUREKA_SERVICE_URL` is set to a valid v2 endpoint of the [Eureka REST API](https://github.com/Netflix/eureka/wiki/Eureka-REST-operations).

| Variable                   | Instance field | Description                                                                                    |
|----------------------------|----------------|------------------------------------------------------------------------------------------------|
| `DISCOVERY_EUREKA_ENABLED` | N/A            | `false` disables Eureka registration. Defaults to `true`.                                      |
| `EUREKA_SERVICE_URL`       | N/A            | v2 endpoint of Eureka, e.g. https://eureka-prod/eureka/v2. No default                          |
| `EUREKA_APP_NAME`          | .app           | The application this instance registers to. Defaults to `zipkin`                               |
| `EUREKA_HOSTNAME`          | .hostName      | The hostname used with `${QUERY_PORT}` to build the instance `vipAddress`. Defaults to detect. |
| `EUREKA_INSTANCE_ID`       | .instanceId    | Defaults to `${EUREKA_HOSTNAME}:${EUREKA_APP_NAME}:${QUERY_PORT}`.                             |

Example usage:

```bash
$ EUREKA_SERVICE_URL=http://localhost:8761/eureka/v2 java -jar zipkin.jar
```

To aid in testing, this adds a docker image
ghcr.io/openzipkin/zipkin-eureka, and an example docker-compose configuration for it.

Fixes #1870

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -119,5 +119,5 @@ jobs:
         run: |
           build-bin/docker/configure_docker &&
           build-bin/maven/maven_go_offline &&
-          build-bin/maven/maven_build -pl :${{ matrix.name }} --am &&
+          build-bin/maven/maven_build install -pl :${{ matrix.name }} --am &&
           build-bin/test -Dgroups=docker -pl :${{ matrix.name }} -Dlicense.skip=true
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -110,8 +110,14 @@ jobs:
       # via forks, and login session ends up in ~/.docker. This is ok because
       # we publish DOCKER_PARENT_IMAGE to ghcr.io, hence local to the runner.
       - name: Test with Docker
-        run:
-          | # configure_test seeds NPM cache, which isn't needed for these tests
-          build-bin/maven/maven_go_offline &&
+        # configure_test seeds NPM cache, which isn't needed for these tests.
+        #
+        # What we are doing here is configuring docker, then building the
+        # module prior to testing it with docker. We do that in two passes
+        # because we want to build the module's dependencies, but not run their
+        # tests.
+        run: |
           build-bin/docker/configure_docker &&
-          build-bin/test -Dgroups=docker -pl :${{ matrix.name }} --am -Dlicense.skip=true
+          build-bin/maven/maven_go_offline &&
+          build-bin/maven/maven_build -pl :${{ matrix.name }} --am &&
+          build-bin/test -Dgroups=docker -pl :${{ matrix.name }} -Dlicense.skip=true
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -89,6 +89,7 @@ jobs:
           - name: zipkin-storage-cassandra
           - name: zipkin-storage-elasticsearch
           - name: zipkin-storage-mysql-v1
+          - name: zipkin-server
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -75,7 +75,7 @@ jobs:
           path: ~/.npm
           key: ${{ runner.os }}-npm-packages-${{ hashFiles('zipkin-lens/package-lock.json') }}
       - name: Test without Docker
-        run: build-bin/maven_go_offline && build-bin/test -Ddocker.skip=true ${{ matrix.maven_args }}
+        run: build-bin/maven_go_offline && build-bin/test -DexcludedGroups=docker ${{ matrix.maven_args }}
 
   test_docker:
     runs-on: ubuntu-22.04 # newest available distribution, aka jellyfish
@@ -114,4 +114,4 @@ jobs:
           | # configure_test seeds NPM cache, which isn't needed for these tests
           build-bin/maven/maven_go_offline &&
           build-bin/docker/configure_docker &&
-          build-bin/test -pl :${{ matrix.name }} --am -Dlicense.skip=true
+          build-bin/test -Dgroups=docker -pl :${{ matrix.name }} --am -Dlicense.skip=true
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -112,12 +112,11 @@ jobs:
       - name: Test with Docker
         # configure_test seeds NPM cache, which isn't needed for these tests.
         #
-        # What we are doing here is configuring docker, then building the
-        # module prior to testing it with docker. We do that in two passes
-        # because we want to build the module's dependencies, but not run their
-        # tests.
+        # What we are doing here is configuring docker, then installing the
+        # module's dependencies prior to testing it with docker. This allows
+        # us to avoid running tests except the leaf module.
         run: |
           build-bin/docker/configure_docker &&
           build-bin/maven/maven_go_offline &&
-          build-bin/maven/maven_build install -pl :${{ matrix.name }} --am &&
+          MAVEN_GOAL=install build-bin/maven/maven_build -pl :${{ matrix.name }} --am &&
           build-bin/test -Dgroups=docker -pl :${{ matrix.name }} -Dlicense.skip=true
--- a/docker/examples/docker-compose.yml
+++ b/docker/examples/docker-compose.yml
@@ -1,5 +1,5 @@
 #
-# Copyright 2015-2024 The OpenZipkin Authors
+# Copyright 2015-2020 The OpenZipkin Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
 # in compliance with the License. You may obtain a copy of the License at
--- a/docker/examples/docker-compose.yml
+++ b/docker/examples/docker-compose.yml
@@ -1,5 +1,5 @@
 #
-# Copyright 2015-2020 The OpenZipkin Authors
+# Copyright 2015-2024 The OpenZipkin Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
 # in compliance with the License. You may obtain a copy of the License at
--- a/docker/examples/docker-compose.yml
+++ b/docker/examples/docker-compose.yml
@@ -1,5 +1,5 @@
 #
-# Copyright 2015-2020 The OpenZipkin Authors
+# Copyright 2015-2024 The OpenZipkin Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
 # in compliance with the License. You may obtain a copy of the License at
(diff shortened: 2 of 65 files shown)


zipkin-server/src/test/java/zipkin2/server/internal/eureka/ITZipkinEureka.java
```diff
@@ -0,0 +1,138 @@
...
+class ITZipkinEureka {
+
+  @Container static EurekaContainer eureka = new EurekaContainer();
+
+  @Autowired Server zipkin;
+  OkHttpClient client = new OkHttpClient.Builder().followRedirects(false).build();
+
+  /** Get the serviceUrl of the Eureka container prior to booting Zipkin. */
+  @DynamicPropertySource static void propertyOverride(DynamicPropertyRegistry registry) {
+    registry.add("zipkin.discovery.eureka.serviceUrl", eureka::serviceUrl);
```
codefromthecrypt, line 75: @openzipkin/armeria fyi the last attempt I made for armeria testing didn't work because I didn't know about this trick.. to defer some property. Everything is cool now as you can see! cc particularly @minwoox as he tried to help me a few years ago on this.

zipkin-server/src/test/java/zipkin2/server/internal/eureka/ITZipkinEureka.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  private Response getEureka(String path) throws IOException {
+    return client.newCall(new Request.Builder()
+      .url(eureka.serviceUrl() + path)
+      .header("Accept", "application/json") // XML is default
+      .build()).execute();
+  }
+
+  static final class EurekaContainer extends GenericContainer<EurekaContainer> {
```
codefromthecrypt: @anuraaga suggested to do this container approach, and yeah it works!

zipkin-server/src/main/resources/zipkin-server-shared.yml
```diff
@@ -209,6 +222,7 @@ server:
     min-response-size: 2048
 
 armeria:
+  healthCheckPath: /health
```
codefromthecrypt, line 225: I needed this to make the eureka integration in armeria use our composite endpoint we control. Otherwise, it wants to use /internal/health which won't include the full status. armeria could introduce a eureka health check path option, if it wanted, which would allow users to set this specific only to that part. Doing so would avoid a double-registration warning in the logs, which is distracting but has no other bad effect.

zipkin-server/pom.xml
```diff
@@ -199,6 +199,13 @@
       <scope>runtime</scope>
     </dependency>
 
+    <!-- Eureka service discovery -->
+    <dependency>
+      <groupId>${armeria.groupId}</groupId>
+      <artifactId>armeria-eureka</artifactId>
```
codefromthecrypt, line 205: because armeria already includes this, there's no real bloat in supporting this. It is just a small http client integration without any other deps. This is because eureka is basically a rest api.

zipkin-server/src/test/java/zipkin2/server/internal/eureka/ITZipkinEureka.java
```diff
@@ -0,0 +1,138 @@
...
+    assertThat(readString(json, "$.application.instance[0].vipAddress"))
+      .isEqualTo(zipkinHostname + ":" + zipkinPort);
+  }
+
+  @Test @Order(2) void deregistersOnClose() {
+    zipkin.close();
+    await().untilAsserted( // wait for deregistration
+      () -> assertThat(getEurekaAsString("/apps"))
+        .isEqualTo(
+          "{\"applications\":{\"versions__delta\":\"1\",\"apps__hashcode\":\"\",\"application\":[]}}"));
```
anuraaga: Not sure if it could be worth doing `assertThat(readList(getEurekaAsString("/apps"), "$.applications.application")).isEmpty()`. A couple of these look like implementation details and maybe change with eureka versions. Or maybe `getEureka("/apps/ZIPKIN")` would 404.

pom.xml
```diff
@@ -493,6 +493,9 @@
             <bnd>SCRIPT_STYLE</bnd>
             <ejs>XML_STYLE</ejs>
             <css>SLASHSTAR_STYLE</css>
+            <!-- build-bin non-trivial scripts -->
```
codefromthecrypt, line 496: fixed license plugin warnings I noticed...

docker/test-images/zipkin-eureka/README.md
```diff
@@ -0,0 +1,12 @@
+## zipkin-eureka Docker image
```
codefromthecrypt, line 1: there is no docker image for eureka, and in fact as of 2.x they don't release a working war file anymore. At any rate, this server was easy to build, and now we can test without thinking too much.

docker/examples/docker-compose.yml
```diff
@@ -31,8 +31,6 @@ services:
     # Environment settings are defined here https://github.com/openzipkin/zipkin/blob/master/zipkin-server/README.md#environment-variables
     environment:
       - STORAGE_TYPE=mem
-      # Point the zipkin at the storage backend
```
codefromthecrypt, line 34: fuzz from long ago

anuraaga (approved):

codefromthecrypt (commented):
notes

zipkin-server/src/test/java/zipkin2/server/internal/eureka/ITZipkinEureka.java
codefromthecrypt: sounds better, thanks!

zipkin-server/src/test/java/zipkin2/server/internal/eureka/ITZipkinEureka.java
anuraaga: Another JFYI :) Now that using Java 11, there would be no cost at all for accessing private members of nested classes (that was improved in Java 10).

Not something to refactor the whole codebase for, but figured it's good to know about.

codefromthecrypt: oh cool I didn't know that. will change to help future copy/paste

zipkin-server/src/test/java/zipkin2/server/internal/eureka/ITZipkinEureka.java
```diff
@@ -0,0 +1,144 @@
...
+  properties = {
+    "server.port=0",
+    "spring.config.name=zipkin-server",
+    "spring.main.banner-mode=off",
+    "zipkin.storage.type=", // cheat and test empty storage type
+    "zipkin.collector.http.enabled=false",
+    "zipkin.query.enabled=false",
+    "zipkin.ui.enabled=false"
+  })
+@Testcontainers
```
anuraaga: Sorry if I provided an incomplete code snippet, this probably needs `disabledWithoutDocker=true`

https://javadoc.io/doc/org.testcontainers/junit-jupiter/latest/org/testcontainers/junit/jupiter/Testcontainers.html#disabledWithoutDocker--

codefromthecrypt: thanks!

codefromthecrypt: switched tests to use groups (`@Tag("docker")`) because the property-based skipping thing was ending up as a nested exception and not skipping tests. I still added the `disabledWithoutDocker=true` which is helpful, but not the same. We want a way to disable docker tests even when docker is available.

.github/workflows/test.yml
```diff
@@ -75,7 +75,7 @@ jobs:
           path: ~/.npm
           key: ${{ runner.os }}-npm-packages-${{ hashFiles('zipkin-lens/package-lock.json') }}
       - name: Test without Docker
-        run: build-bin/maven_go_offline && build-bin/test -Ddocker.skip=true ${{ matrix.maven_args }}
+        run: build-bin/maven_go_offline && build-bin/test -DexcludedGroups=docker ${{ matrix.maven_args }}
```
codefromthecrypt, line 78: so first pass will not use docker due to this

.github/workflows/test.yml
```diff
@@ -113,4 +114,4 @@ jobs:
           | # configure_test seeds NPM cache, which isn't needed for these tests
           build-bin/maven/maven_go_offline &&
           build-bin/docker/configure_docker &&
-          build-bin/test -pl :${{ matrix.name }} --am -Dlicense.skip=true
+          build-bin/test -Dgroups=docker -pl :${{ matrix.name }} --am -Dlicense.skip=true
```
codefromthecrypt: matrix will only run docker tests due to this (quicker than before I think)

zipkin-server/src/test/java/zipkin2/server/internal/eureka/ITZipkinEureka.java
```diff
@@ -0,0 +1,141 @@
...
+  properties = {
+    "server.port=0",
+    "spring.config.name=zipkin-server",
+    "spring.main.banner-mode=off",
+    "zipkin.storage.type=", // cheat and test empty storage type
+    "zipkin.collector.http.enabled=false",
+    "zipkin.query.enabled=false",
+    "zipkin.ui.enabled=false"
+  })
+@Tag("docker")
```
codefromthecrypt, line 58: tests who use docker should tag like this, so we can skip it in the first PR test stage.

codefromthecrypt (commented):
notes on the docker change related to CI and the new jupiter infra. a later PR will refactor the other tests more.

codefromthecrypt:
Almost there.. green, but zipkin-server is redundantly testing the other docker stuff due to --am. I think we should do a build pass first, *then* run tests, and not use --am. I'll do it

docker/test-images/zipkin-eureka/src/main/resources/application.yaml
```diff
@@ -0,0 +1,42 @@
...
+# is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+# or implied. See the License for the specific language governing permissions and limitations under
+# the License.
+#
+
+# Add configuration to disable as much server caching as possible. Note that
+# not all configuration here are defined by netflix/eureka, rather some are
+# specific to spring-cloud/spring-cloud-netflix.
+eureka:
+  client:
```
codefromthecrypt, line 19: last couple hours died due to finding these settings. basically bad luck in CI where the first test passed into a stale data interval which outlasted awaitility's threshold!

codefromthecrypt:
Thanks for the review and suggestions @anuraaga and thanks for the review on the prior PR @minwoox. more work can happen post merge as necessary.

codefromthecrypt:
p.s. for folks at home, remember zipkin publishes a new docker image on commits to master, so you can try this via `ghcr.io/openzipkin/zipkin:master` (not latest which is last released)

PR #3674, opened 2024-01-05
codec: adds offset parameter to SpanBytesEncoder.write

codefromthecrypt:
This adds an offset parameter to `SpanByteEncoder.write` which allows re-use of a buffer, similar to `writeList`. This is in preparation of a change to zipkin-reporter-java which needs to update dependencies anyway. Doing this allows the async reporter to allocate less memory while accumulating spans.

```java
pubiic int write(Span span, byte[] out, int pos)
```

Note: this doesn't add a ByteBuffer or similar overload for write because the code is optimized and only implemented for byte arrays. Doing more than this would be a bigger job.


codefromthecrypt:
actually closing this out as I looked carefully and there's no way the current sender could take advantage of this. sorry about the distraction.

Issue #3670, opened 2024-01-03
Docker container ends up with defunkt wget processes


rogierslag:
## Describe the Bug
When running the Docker container for longer periods of time, zombie proceses for wget show up, which can be traced back to the `openzipkin/zipkin` container.

Based on the code, the wget seems to originate from the healthcheck as done here https://github.com/openzipkin/zipkin/blob/master/build-bin/docker/docker-healthcheck#L48.

## Steps to Reproduce
Keep the Docker container running for a while (at least on our infrastructure)

## Expected Behaviour
Whenever wget fails for whatever reason, it should be terminated gracefully and not become a zombie process

codefromthecrypt:
hmm could it be that the interval is 1s and for some reason the wgets are taking longer?

I raised this which should be sensible in context https://github.com/openzipkin/zipkin/pull/3677

what is the ad-hoc latency for your /health endpoint?

rogierslag:
Ran a quick test with `ab` which yields well below 1s response times.

```
ab -c 5 -n 500 https://zipkin.internal.magnet.me/health
This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking zipkin.internal.magnet.me (be patient)
Completed 100 requests
Completed 200 requests
Completed 300 requests
Completed 400 requests
Completed 500 requests
Finished 500 requests


Server Software:        nginx/1.24.0
Server Hostname:        zipkin.internal.magnet.me
Server Port:            443
SSL/TLS Protocol:       TLSv1.2,ECDHE-RSA-AES128-GCM-SHA256,2048,128
Server Temp Key:        ECDH P-384 384 bits
TLS Server Name:        zipkin.internal.magnet.me

Document Path:          /health
Document Length:        410 bytes

Concurrency Level:      5
Time taken for tests:   13.170 seconds
Complete requests:      500
Failed requests:        0
Total transferred:      324500 bytes
HTML transferred:       205000 bytes
Requests per second:    37.96 [#/sec] (mean)
Time per request:       131.704 [ms] (mean)
Time per request:       26.341 [ms] (mean, across all concurrent requests)
Transfer rate:          24.06 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:       63   91  15.0     88     161
[...]

PR #3677, opened 2024-01-05
docker: set HEALTHCHECK timeouts to 1s

codefromthecrypt:
I set the only available timeout constraints on the BusyBox v1.31.1 `wget` and `nc` commands to 1s to hopefully not accumulate hung HEALTHCHECK stuff.

See #3670


codefromthecrypt:
test/test_docker wait on HEALTHCHECK so will hit this script

reta (approved):

PR #3681, opened 2024-01-09
Migrates self-tracing to Brave 6

codefromthecrypt:


zipkin-server/src/test/java/zipkin2/server/internal/throttle/ThrottledCallTest.java
```diff
@@ -34,6 +34,7 @@
 import zipkin2.Call;
 import zipkin2.Callback;
 import zipkin2.reporter.AwaitableCallback;
```
codefromthecrypt, line 36: this test snuck in this, but exposing the adapter was less work than recreating AwaitableCallback only for this test

PR #3682, opened 2024-01-09
Fix typo in IllegalArgumentException

c13:
It seems that exceptions are a bit outdated after Elasticsearch 8 support was added in [#3552](https://github.com/openzipkin/zipkin/pull/3552)


zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/VersionSpecificTemplates.java
```diff
@@ -226,7 +226,7 @@ String autocompleteTemplate(ElasticsearchVersion version) {
   IndexTemplates get(ElasticsearchVersion version) {
     if (version.compareTo(V5_0) < 0 || version.compareTo(V9_0) >= 0) {
       throw new IllegalArgumentException(
-        "Elasticsearch versions 5-7.x are supported, was: " + version);
+        "Elasticsearch versions 5-8.x are supported, was: " + version);
```
codefromthecrypt, line 229: technically, we don't support 5 anymore, but aren't enforcing it.

codefromthecrypt (approved):
thanks!

PR #3684, opened 2024-01-10
Prepares for 3.0 by updating to Spring Boot 3 and floor Java 8

codefromthecrypt:
We currently still have a problem as the latest version shows up as 3.0.0-rc0 due to an accidental release. Moreover, a request to delete that version hasn't happened. This switches to 3.x by dropping support for Java 6 and moving to Spring Boot 3.

As noted, recent Zipkin Reporter 3 and Brave 6 decouple instrumentation from this core jar. While it is still in use for other formats, the call sites using them are all minimum Java 8 anyway. By updating to floor Java 8, we can use later JDKs to release than 11, though this change doesn't include doing that, yet.

As Spring Boot requires JRE 17, we update the floor JDK to 17. However, zipkin-dependencies is currently stuck at 11, so until that's not the case, this leaves collector and storage modules at Java 11, even if the server can only run in 17+.

This will be mergable once I manually verify Spring Boot 3 doesn't break modularity with zipkin-aws and -gcp.


codefromthecrypt:
used openrewrite to migrate tests and server code to 17, by doing this and reverting the opposite

https://docs.openrewrite.org/recipes/java/migrate/upgradetojava17
```
mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-migrate-java:RELEASE -Drewrite.activeRecipes=org.openrewrite.java.migrate.UpgradeToJava17
```

codefromthecrypt:
@timtebeek fyi, the last commit fixes some things that were mostly, but not completely done with the 17 update recipe. It seemed to dislike some things about how our linefeeds work as well unicode specifically 💩

codefromthecrypt:
@timtebeek just a hunch, but I think it is about multi-line literal conversion where the <17 has no trailing newline or unicode. Was easy to fix these, so no biggie!

codefromthecrypt:
oh wow it is green

codefromthecrypt:
surprised this worked. here are the minimal changes for zipkin-gcp. I'll add more to the branch after using openrewrite to migrate some stuff https://github.com/openzipkin/zipkin-gcp/compare/zipkin3?expand=1

.github/workflows/create_release.yml
```diff
@@ -24,13 +24,13 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'zulu'  # zulu as it supports a wide version range
-          java-version: '11'  # earliest LTS and last that can compile the 1.6 release profile.
+          java-version: '17'  # earliest LTS and last that can compile the 1.6 release profile.
```
anuraaga, line 27: I guess this can't compile 1.6 profile

zipkin-collector/scribe/src/test/java/zipkin2/collector/scribe/ScribeSpanConsumerTest.java
```diff
@@ -228,9 +228,10 @@ private void expectSuccess(ScribeSpanConsumer scribe, LogEntry entry) throws Exc
...
+    entry.message = """
```
anuraaga, line 231: Cool

anuraaga (approved):

.github/workflows/create_release.yml
codefromthecrypt: yeah 11 was the last that could

codefromthecrypt: for posterity! https://bugs.openjdk.org/browse/JDK-8206440

codefromthecrypt:
merging and will reference zipkin-aws and zipkin-gcp via the snapshot repository and link back PRs. Thanks for the review @anuraaga!

codefromthecrypt:
and here's zipkin-aws https://github.com/openzipkin/zipkin-aws/pull/209

codefromthecrypt:
here's zipkin-gcp https://github.com/openzipkin/zipkin-gcp/pull/209

codefromthecrypt:
will cut 3.0 tomorrow morning unless screams!

shakuzen:
I'm late but great work, @codefromthecrypt.

zipkin-server/src/test/java/zipkin2/server/internal/elasticsearch/TestResponses.java
```diff
@@ -36,7 +36,7 @@ final class TestResponses {
         "minimum_index_compatibility_version" : "5.0.0"
       },
       "tagline" : "You Know, for Search"
-    }\
+    }
```
timtebeek, line 39: I'm trying to understand these reverts better; I'd thought text block without trailing new line would be correctly converted [with a test like this](https://github.com/openrewrite/rewrite-migrate-java/blob/07443b1ece4744963bc4159cdd3edcef5d4b38bb/src/test/java/org/openrewrite/java/migrate/lang/UseTextBlocksTest.java#L48-L70).

JShell seems to agree at first glance:
```
jshell> var old = "foo\n" + "bar";  
old ==> "foo\nbar"

jshell> old.equals("""
   ...> foo
   ...> bar\
   ...> """)
$2 ==> true
```
How did this then end up with something that needed to be reverted?

codefromthecrypt: Not 100pct sure as thinking about other things, but we had a lot of stuff in this format style... which may be an uncommon pattern

```
      String json = "{\n"
        + "  \"traceId\": \"6b221d5bc9e6496c\",\n"
        + "  \"name\": \"get-traces\",\n"
        + "  \"id\": \"6b221d5bc9e6496c\",\n"
        + "  \"annotations\": [\n"
        + "    { \"timestamp\": 1472470996199000}\n"
        + "  ]\n"
        + "}";
```

PR #3686, opened 2024-01-11
fixes default release version from 17 to 11

codefromthecrypt:
This was a typo and doesn't affect anything except zipkin-dependencies, as that's the only project that uses non-overridden targets, and requires version 11 (due to spark conflicts that hopefully will one day go away).


anuraaga (approved):

shakuzen (approved):

codefromthecrypt:
thanks for the look @anuraaga and @shakuzen! I'll release 3.0.1 so I can align zipkin-dependencies before going offlline mostly 10days or so.

PR #3687, opened 2024-01-12
removes log4j warning message

codefromthecrypt:
I noticed this testing homebrew
```bash
$ java -jar zipkin.jar
ERROR StatusLogger Log4j2 could not find a logging implementation. Please add log4j-core to the classpath. Using SimpleLogger to log to the console...

                  oo
                 oooo

```

Also, fixed some README fuzz and updated docker images to latest

### DIFF
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -266,6 +266,13 @@
       <scope>provided</scope>
     </dependency>
 
+    <!-- the "slim" jar will exclude this -->
+    <dependency>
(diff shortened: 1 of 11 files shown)


zipkin-server/pom.xml
```diff
@@ -266,6 +266,13 @@
       <scope>provided</scope>
     </dependency>
 
+    <!-- the "slim" jar will exclude this -->
```
codefromthecrypt, line 269: here is the main change

Issue #3649, opened 2023-12-14
Add a way to install zipkin through homebrew on macos


remyleone:
## Feature

I would like to install zipkin using homebrew using `brew install zipkin`

## Rationale

It would be helpful for other macos user to have a way to experiment quickly and easily with zipkin while keeping it up to date with brew

## Example Scenario

I got a mac, I don't have docker, I want to try zipkin out, I would like to have a way to test it without compiling.

## Prior Art
* Links to prior art
* More links

codefromthecrypt:
sounds very sensible

codefromthecrypt:
for now you can try the quick start https://zipkin.io/pages/quickstart.html

```bash
$ curl -sSL https://zipkin.io/quickstart.sh | bash -s
$ java -jar zipkin.jar
```

One question for the home-brew thing, is if this should be the full zipkin build (includes kafka etc) or slim (smaller and a bit quicker to startup, but no messaging support)

remyleone:
Kafka is already available in [homebrew](https://github.com/Homebrew/homebrew-core/blob/f49bf3578641d7bd93021553c7ecf99255b137e7/Formula/k/kafka.rb), maybe the other components in the full zipkin build are also present.

I guess you could have a jar that contains only zipkin code and either include a dependency on kafka explicitly (and the other dependencies) or having something without kafka support included but that could find it opportunistically if present.

codefromthecrypt:
oh zipkin's support of kafka is optional. I'm leaning towards the full build

codefromthecrypt:
someone who wants to help with this can copy something like this https://github.com/Homebrew/homebrew-core/blob/4ef4ac3f2708daa8be276fcea3fdaade8ddded3a/Formula/d/denominator.rb into a new pull request

except use https://repo1.maven.org/maven2/io/zipkin/zipkin-server/2.26.0/zipkin-server-2.26.0-exec.jar as the jar

@reta wanna give this a try? I've done several homebrew recipes and the community is nice.

reta:
> @reta wanna give this a try? I've done several homebrew recipes and the community is nice.

@codefromthecrypt really want to help here, but I don't have MacOS box available to me (I have Windows and Linux boxes) :(

codefromthecrypt:
ah yeah indeed macos is important for testing. I'll help with this!

codefromthecrypt:
[...]

PR #3689, opened 2024-01-14
ci: adds lint check and fixes bad links and yaml formatting

codefromthecrypt:
This adds new scripts `configure_lint` and `lint` which work like our other build-bin scripts. These lint the workflow yaml and our markdown links.

I also removed the separate step for javadoc checking as it is easy to do that in the release test target, as well uses less resources vs doing it twice.

See https://github.com/openzipkin/zipkin-gcp/pull/212 for the first impl of this. I will propagate the same style elsewhere after.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -9,7 +9,7 @@ When submitting code, please apply [Square Code Style](https://github.com/square
 ## License
 
 By contributing your code, you agree to license your contribution under
-the terms of the [APLv2](LICENSE).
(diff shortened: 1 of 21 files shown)


codefromthecrypt:
besides normal motivation, one is to make sure things pass before adding new links for #3649

reta (approved):

codefromthecrypt:
I refactored this to be coherent with the other steps we do like `test` and `deploy`. This makes it easy to verify on your laptop the same as CI would do.

PR #3695, opened 2024-01-17
backfills tests for behaviour on malformed due to non-string tags

codefromthecrypt:
This adds tests to prove that all zipkin collectors return errors when json is malformed due to non-string tags.

See https://github.com/openzipkin/zipkin/pull/3503#issuecomment-1894350504

### DIFF
--- a/zipkin/src/main/java/zipkin2/internal/JsonCodec.java
+++ b/zipkin/src/main/java/zipkin2/internal/JsonCodec.java
@@ -15,7 +15,6 @@ package zipkin2.internal;
 
 import java.io.IOException;
 import java.io.InputStreamReader;
-import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


zipkin/src/main/java/zipkin2/internal/JsonCodec.java
```diff
@@ -222,7 +221,9 @@ public static <T> void writeList(WriteBuffer.Writer<T> writer, List<T> value, Wr
 
   static IllegalArgumentException exceptionReading(String type, Exception e) {
     String cause = e.getMessage() == null ? "Error" : e.getMessage();
-    if (cause.contains("Expected BEGIN_OBJECT") || cause.contains("malformed")) {
+    if (cause.contains("Expected BEGIN_OBJECT")
```
codefromthecrypt, line 224: while this normalizes the error as intended, this PR already shows any data that has malformed tags in zipkin were created out-of-band, like not using zipkin, as all collectors regardless of HTTP or Kafka use this path

PR #3699, opened 2024-01-23
docker: fixes spring config in zipkin-eureka test image

codefromthecrypt:
Formerly, we didn't test the authenticated side, so some review drift passed even though authenticating would crash on startup:

```
Parameter 0 of method filterChain in zipkin.test.EurekaSecurity required a bean of type 'org.springframework.security.config.annotation.web.builders.HttpSecurity' that could not be found.
```

### DIFF
--- a/build-bin/docker-compose-zipkin-eureka.yml
+++ b/build-bin/docker-compose-zipkin-eureka.yml
@@ -1,0 +1,32 @@
+#
+# Copyright 2015-2024 The OpenZipkin Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+# in compliance with the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License
+# is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+# or implied. See the License for the specific language governing permissions and limitations under
+# the License.
+#
+
+# uses 2.4 so we can use condition: service_healthy
+version: "2.4"
+
+# Test both authenticated and unauthenticated, as if there is a Spring problem,
+# the latter will crash. We only need to use HEALTHCHECK for this.
+services:
+  eureka:
+    image: openzipkin/zipkin-eureka:test
+    container_name: eureka
+  sut:
+    image: openzipkin/zipkin-eureka:test
+    container_name: sut
+    environment:
+      EUREKA_USERNAME: testuser
+      EUREKA_PASSWORD: testpassword
+    depends_on:
+      eureka:
+        condition: service_healthy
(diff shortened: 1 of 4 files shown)


build-bin/docker-compose-zipkin-eureka.yml
```diff
@@ -0,0 +1,32 @@
...
+# the License.
+#
+
+# uses 2.4 so we can use condition: service_healthy
+version: "2.4"
+
+# Test both authenticated and unauthenticated, as if there is a Spring problem,
+# the latter will crash. We only need to use HEALTHCHECK for this.
+services:
+  eureka:
```
codefromthecrypt, line 21: without a compose, it effectively only runs this part. I daisy-chained the two paths since we are using spring security for username/password stuff. Maybe someone can convert this to armeria and use normal simple basic auth filter one day, but I went with the spring-security route initially.

docker/test-images/zipkin-eureka/src/main/java/zipkin/test/EurekaSecurity.java
```diff
@@ -33,7 +33,7 @@
 @Configuration
 @ConditionalOnProperty("eureka.username")
 @EnableConfigurationProperties(EurekaProperties.class)
-@ImportAutoConfiguration(SecurityAutoConfiguration.class)
+@Import(SecurityAutoConfiguration.class)
```
reta: So this one didn't work out? (just curious since we're importing auto configuration)

codefromthecrypt:
I removed spring-security as it was far more effort than value to just verify basic auth.

docker/test-images/zipkin-eureka/src/main/java/zipkin/test/EurekaSecurity.java
codefromthecrypt: yeah it threw the error I mentioned in the description. Then, I fought for an hour over CSRF. Finally, I axed this thing for a basic filter (pun intended)

build-bin/docker-compose-zipkin-eureka.yml
reta: Sorry missed that

reta (approved):

build-bin/docker-compose-zipkin-eureka.yml
codefromthecrypt: no worries I should have checked. Anyway the prior one failed on health check for frustrating reasons.

codefromthecrypt:
cool finally passing ;) now I will forget all the config lambda-foo and try to remember what I was doing

docker/test-images/zipkin-eureka/pom.xml
```diff
@@ -50,15 +50,6 @@
       <artifactId>spring-cloud-starter-netflix-eureka-server</artifactId>
     </dependency>
 
-    <dependency>
-      <groupId>org.springframework.security</groupId>
```
codefromthecrypt, line 54: note this didn't change the image size as even if the eureka layer is 55mb the current classpath requires modules not in our JRE, so we use the JDK image as noted in the Dockerfile. The image sum with the JDK layer ends up as 430MB. Maybe a trimmer eureka will happen one day, but it isn't terribly important.

PR #3704, opened 2024-01-24
deps: bumps to latest library versions, notably slf4j

codefromthecrypt:
SLF4J maintains than an upgrade from 1.x to 2.x is compatible, with some caveats about implementations matching. By upgrading, we can move off an abandoned test dependency.

This also updates other dependencies that were slightly out of date.

### DIFF
--- a/zipkin/pom.xml
+++ b/zipkin/pom.xml
@@ -64,22 +64,6 @@
 
   <build>
     <plugins>
-      <plugin>
-        <groupId>de.qaware.maven</groupId>
-        <artifactId>go-offline-maven-plugin</artifactId>
-        <configuration>
-          <dynamicDependencies>
-            <DynamicDependency>
-              <groupId>org.codehaus.mojo.signature</groupId>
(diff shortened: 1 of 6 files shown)


zipkin/pom.xml
```diff
@@ -64,22 +64,6 @@
...
   <build>
     <plugins>
-      <plugin>
-        <groupId>de.qaware.maven</groupId>
-        <artifactId>go-offline-maven-plugin</artifactId>
-        <configuration>
-          <dynamicDependencies>
-            <DynamicDependency>
-              <groupId>org.codehaus.mojo.signature</groupId>
-              <artifactId>java16</artifactId>
```
codefromthecrypt, line 74: unused as we no longer compile 1.6

codefromthecrypt:
made a comment so next version of armeria aligns micrometer with spring boot drift: surprising for us at least considering this is a patch, not a minor bump! https://github.com/line/armeria/pull/5372#pullrequestreview-1840229270

PR #3707, opened 2024-01-24
docker: adds more defaults to elasticsearch images

codefromthecrypt:


codefromthecrypt:
going to add these settings and summarize status on https://github.com/openzipkin/zipkin-dependencies/pull/220 as I can't really spend any more time on this at all.

PR #3708, opened 2024-01-24
docker: uses portable config between ES 7 and 8

codefromthecrypt:
this doesn't fix ES 8 on https://github.com/openzipkin/zipkin-dependencies/pull/220, but it makes some things more coherent, so why not.


docker/test-images/zipkin-elasticsearch8/Dockerfile
```diff
@@ -26,13 +26,13 @@ FROM scratch as scratch
```
reta, line 26: @codefromthecrypt could we try with:

```
ARG java_version= 21.0.1_p12
```

There is a regression in 21.0.2 (https://github.com/elastic/elasticsearch/pull/104347) which should be fixed in `8.12.1`, I think this is why you see the weird container issues.

codefromthecrypt: cool trying now!

codefromthecrypt: totally works. not sure how you figured this out, but so glad you are on the team!

codefromthecrypt: ES 7 already sorted it, so bumping the version https://github.com/elastic/elasticsearch/commits/v7.17.17

codefromthecrypt:
ps the prior build flake would be fixed when we consume these images, as it is whining about over memory. crazy how much memory ES needs these days!

codefromthecrypt:
recut images via pushing the tag docker-3.0.5

docker/test-images/zipkin-elasticsearch8/Dockerfile
reta: > totally works. not sure how you figured this out, but so glad you are on the team!

I just knew that from OpenSearch (since we largely have a same codebase and I was working on mitigation this week)

codefromthecrypt: guess I hadn't properly doxxed you ;) I had no idea me randomly asking you to help with elasticsearch was not so random.

PR #3709, opened 2024-01-24
cassandra: cleans up code and tests

codefromthecrypt:
This cleans up deprecation and style issues found in IntelliJ analysis. It also migrates docker tests to junit-jupiter and fixes the logging configuration which broke when we updated to SLF4J 2.x.

This also updates all use of docker test images to the latest, in attempts to share base image layers.

PR #3710, opened 2024-01-25
elasticsearch: adjust for recent deprecations

codefromthecrypt:
one more adjustment to make all things clean

### DIFF
--- a/zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ITElasticsearchStorage.java
+++ b/zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ITElasticsearchStorage.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2023 The OpenZipkin Authors
+ * Copyright 2015-2024 The OpenZipkin Authors
  *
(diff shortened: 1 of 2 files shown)


zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ITElasticsearchStorage.java
```diff
@@ -163,12 +163,14 @@ static void aggregateDependencies(ElasticsearchStorage storage, List<Span> spans
     final AggregatedHttpResponse response =
       webClient.execute(HttpRequest.of(RequestHeaders.of(HttpMethod.GET,
         "/_migration/deprecations"))).aggregate().join();
-    if (!response.contentAscii().isEmpty()) {
```
codefromthecrypt, line 166: as of recent times, the body would never be empty. This compares against the literal empty instead of using json parsing.

codefromthecrypt: we didn't look at the log warning, and recently we couldn't have because logging was broken! #3709 fixes logging

PR #3715, opened 2024-02-07
Updates to Armeria 1.27.1 and fixes Eureka VIP registration bug

codefromthecrypt:
This updates to latest Armeria 1.27.0, which removes some tech debt. This also fixes a bug in the default behavior of Armeria, which is that it sets a `host:port` value for the VIP address in Eureka. While the `instanceId` needs to be qualified with port, the VIP address must not, as there's already a port field.

cc @openzipkin/armeria


codefromthecrypt:
tomorrow I'll raise a PR on the armeria side about the bad default on vipAddress, which shouldn't be host like http host header, rather a host like dns hostname.

zipkin-server/src/test/java/zipkin2/server/internal/eureka/BaseITZipkinEureka.java
```diff
@@ -60,7 +59,9 @@
     "zipkin.storage.type=", // cheat and test empty storage type
     "zipkin.collector.http.enabled=false",
     "zipkin.query.enabled=false",
-    "zipkin.ui.enabled=false"
+    "zipkin.ui.enabled=false",
+    // TODO: Remove when armeria doesn't set the vip address to include a port.
```
codefromthecrypt: actually, this is probably fine to leave here. I'll remove the TODO once I raise the PR in armeria to fix the default. It isn't really our job to test the default.

reta (approved):

zipkin-server/src/test/java/zipkin2/server/internal/eureka/BaseITZipkinEureka.java
codefromthecrypt: raised the PR and removed the comment https://github.com/line/armeria/pull/5451

codefromthecrypt:
https://github.com/line/armeria/pull/5451 should fix upstream

zipkin-server/src/test/java/zipkin2/server/internal/eureka/BaseITZipkinEureka.java
```diff
@@ -60,7 +59,8 @@
     "zipkin.storage.type=", // cheat and test empty storage type
     "zipkin.collector.http.enabled=false",
     "zipkin.query.enabled=false",
-    "zipkin.ui.enabled=false"
+    "zipkin.ui.enabled=false",
+    "zipkin.discovery.eureka.hostname=localhost"
```
codefromthecrypt, line 63: I ran `./gradlew install` from https://github.com/line/armeria/pull/5451 and set the armeria version to what it produced (1.27.2-SNAPSHOT). Then, I removed this property and the results of the tests were expected, with my machine's hostname in the instanceId (along with the port), and my machine's hostname without the port as the vipAddress

codefromthecrypt:
Thanks for the review @reta!

PR #3717, opened 2024-02-08
Fixes logging glitches and updates docker images

codefromthecrypt:
This does a few things that polish docker test images and logging:

* updates base layers to Alpine 3.19.1
* uses Elasticsearch 8.12.1 which allows re-use of a consistent JRE
* Fixes logging glitch in zipkin-eureka
* Fixes ES_HTTP_LOGGING in zipkin-slim (which uses simplelogger)

### DIFF
--- a/zipkin-server/src/main/resources/simplelogger.properties
+++ b/zipkin-server/src/main/resources/simplelogger.properties
@@ -13,3 +13,7 @@ org.slf4j.simpleLogger.showShortLogName=true
 org.slf4j.simpleLogger.log.com.linecorp.armeria=WARN
 # But allow to say it's ready to serve requests
 org.slf4j.simpleLogger.log.com.linecorp.armeria.server.Server=INFO
+# # and when registered in Eureka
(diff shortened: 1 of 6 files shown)


zipkin-server/src/main/resources/simplelogger.properties
```diff
@@ -13,3 +13,7 @@ org.slf4j.simpleLogger.showShortLogName=true
 org.slf4j.simpleLogger.log.com.linecorp.armeria=WARN
 # But allow to say it's ready to serve requests
 org.slf4j.simpleLogger.log.com.linecorp.armeria.server.Server=INFO
+# # and when registered in Eureka
+org.slf4j.simpleLogger.log.com.linecorp.armeria.server.eureka.EurekaUpdatingListener=INFO
```
codefromthecrypt, line 17: copied from the same properties set in zipkin-shared yaml

anuraaga (approved):

PR #3720, opened 2024-02-09
docker: tests with latest images and fixes mysql pin

codefromthecrypt:


docker/test-images/zipkin-mysql/Dockerfile
```diff
@@ -36,8 +36,8 @@ COPY --from=scratch /docker-bin/* /usr/local/bin/
 HEALTHCHECK --interval=1s --start-period=30s --timeout=5s CMD ["docker-healthcheck"]
 ENTRYPOINT ["start-mysql"]
 
-# Use latest from https://pkgs.alpinelinux.org/packages?name=mysql
-ARG mysql_version=10.11.6-r0
```
codefromthecrypt, line 40: I accidentally added the -r0 when updating this earlier. It is a problem because when -r1 comes out, the build would break.

reta (approved):

anuraaga (approved):

PR #3721, opened 2024-02-09
eureka: comments why we aren't currently solving bad UX

codefromthecrypt:
Our test image shows the server twice, once as a failed replica. This is despite adding several properties. However, alternatives in the current codebase are worse, at least alternatives I'm aware of. I opened https://github.com/spring-cloud/spring-cloud-netflix/issues/4251 to capture notes, though not really expecting a lot of follow-up. I'm commenting this, just to save another person a bunch of debugging to several poor ends.

poop


anuraaga (approved):

codefromthecrypt:
thanks for the peek @anuraaga

PR #3725, opened 2024-02-12
Fixed language selector.

SamTV12345:
This pull request fixes the language selector. Unfortunately the po files don't work with react-i18next so I had to convert them into json files. 

Closes #3723

PR #3727, opened 2024-02-16
lens: fixes language selector.

codefromthecrypt:
re-raised because #3725 was on a branch named master, so couldn't polish it. Thanks for the help, @SamTV12345!

i18n fix

recorded with Giphy Capture


zipkin-lens/src/translations/i18n.ts
```diff
@@ -1,30 +1,36 @@
...
 i18n
-  .use(LanguageDetector)
   .use(initReactI18next)
+  .use(LanguageDetector)
   .init(
     {
-      resources
+      resources,
+      fallbackLng: FALLBACK_LOCALE,
+      debug: true,
```
codefromthecrypt:
```suggestion
```

zipkin-lens/src/components/App/LanguageSelector.tsx
```diff
@@ -63,15 +64,24 @@ const LanguageSelector = () => {
...
+
       i18n.changeLanguage(locale);
-      setLocale(locale);
     },
-    [currentLocale, i18n],
+    [i18n],
   );
 
+  useEffect(() => {
+    console.log('Browser locale', i18n.language);
```
codefromthecrypt:
```suggestion
```

codefromthecrypt:
added a screen recording that shows it works from an integrated zipkin server

zipkin-lens/src/components/App/LanguageSelector.tsx
```diff
@@ -63,15 +64,23 @@ const LanguageSelector = () => {
...
+    [i18n],
   );
 
+  useEffect(() => {
+    if (LANGUAGES.find((lang) => lang.locale === i18n.language)) {
+      i18n.changeLanguage(i18n.language);
+    } else {
+      i18n.changeLanguage(FALLBACK_LOCALE); // fallback to default language if the selected language is not supported
+    }
+  }, []);
```
anuraaga:
```suggestion
  }, [i18n]);
```

Not sure but believe this is supposed to be here

anuraaga (approved):

zipkin-lens/src/components/App/LanguageSelector.tsx
codefromthecrypt: thanks, I'll give it a go locally

codefromthecrypt:
works still (also cookies still work to persist a selection)

codefromthecrypt:
Thanks for the help @SamTV12345 and the review and polishing @anuraaga !

zipkin-lens/src/components/App/LanguageSelector.tsx
SamTV12345: I think the problem is that this will update the locale twice. So if I click on the button it will trigger i18next to change the locale. But because the locale is changed the useEffect is called and the language is again changed. The idea was to have this useEffect only be executed on page rendering.

PR #3729, opened 2024-02-16
Fix UiConfig test

### LINKED ISSUE #3724: lens: UiConfigTest "provides config when resolved" no longer runs
UiConfigTest "provides config when resolved" no longer runs after refactoring to vitest. While the underlying code may still work, we need this to run again to be sure.

cc @SamTV12345 @anuraaga if either of you can help with this.

anuraaga:
Fixes #3724

The key point was to replace fetch-mock, which works with jest, with vitest mocking. But I also found combining the two tests into one to be easier to reason about (and debug until the finish).

I ran `npm i` to update lockfile after removing fetch-mock but it seems to have had a very dramatic effect, hope it didn't bring back any CVEs but not sure how to do better...


zipkin-lens/src/components/UiConfig/UiConfig.test.jsx
```diff
@@ -35,39 +37,32 @@ const renderUiConfig = async () => {
...
     expect(screen.getAllByText('Suspended')).length(1);
 
+    expect(fetchSpy).toHaveBeenCalledWith(UI_CONFIG);
+
+    resolve(new Response(configJson));
     // We need to get off the processing loop to allow the promise to complete and resolve the
     // config.
     await new Promise((resolve) => setTimeout(resolve, 1));
 
-    rerender(); // was rerender(<UiConfig />);
```
anuraaga, line 68: Presumably I was calling rerender before for some reason, but whether a dep update or what, it seems to not be needed anymore and actually cause the test to fail

codefromthecrypt:
pushed a commit as a couple other tests were still fetch'n

codefromthecrypt (approved):
cool on green I agree it is easier to for passer-by's like me reason with!

codefromthecrypt:
thanks again @anuraaga !

PR #3731, opened 2024-02-19
Adds UI_ENABLED mapping to make more obvious how to disable it

### LINKED ISSUE #3456: Disable Zipkin UI without disabling read endpoints
## Feature
Add a runtime property that only disables zipkin UI without disabling read endpoints.

## Rationale
Currently, if you need to disable Zipkin UI, the only option is to set QUERY_ENABLED to false. However, this disables the HTTP read endpoint. If you want to get trace data using the Zipkin read endpoint, but you need to disable the UI, there is no solution.

## Example Scenario
Example scenario can be explained as follows:-

1. Zipkin does not have OAuth 2.0 natively integrated. Therefore, OAuth 2.0 compliant integration like Okta is not possible with Zipkin.
2. Instead of providing users with Zipkin UI without OAuth 2.0 auth, use Grafana as Grafana has built-in Okta integration support. Trace data is sent from line-of-business applications to Zipkin. Use Grafana's built-in Zipkin support to use Zipkin as your data source.
3. View trace data from the Grafana UI.
4. Disable Zipkin UI to avoid unauthenticated access.

As you can see from the steps above,  Grafana will not be able to retrieve trace data from Zipkin if the Zipkin read endpoint is disabled.

codefromthecrypt:
fixes #3456

### DIFF
--- a/zipkin-server/README.md
+++ b/zipkin-server/README.md
@@ -136,7 +136,7 @@ If your only goal is to disable the UI, use `UI_ENABLED` instead. Defaults to tr
 in the collector to support search. This does not disable the entire UI, as trace by ID and
 dependency queries still operate. Disable this when you use another service (such as logs) to find
 trace IDs. Defaults to true
-* `UI_ENABLED`: `false` disables UI, mounted at '/zipkin'. Defaults to true
+* `UI_ENABLED`: `false` disables the web UI, mounted at '/zipkin'. Defaults to true
 * `QUERY_TIMEOUT`: Sets the hard timeout for query requests. Accepts any duration string (e.g., 100ms).
 A value of 0 will disable the timeout completely. Defaults to 11s.
 * `QUERY_LOG_LEVEL`: Log level written to the console; Defaults to INFO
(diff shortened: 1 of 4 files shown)


zipkin-server/README.md
```diff
@@ -131,11 +131,12 @@ Here are the top-level configuration of Zipkin:
 * `QUERY_PORT`: Listen port for the HTTP API and web UI; Defaults to 9411
 * `QUERY_ENABLED`: `false` disables the HTTP read endpoints under '/api/v2'. This also disables the
 UI, as it relies on the API. If your only goal is to restrict search, use `SEARCH_ENABLED` instead.
-Defaults to true
+If your only goal is to disable the UI, use `UI_ENABLED` instead. Defaults to true
 * `SEARCH_ENABLED`: `false` disables searching in the query API and any indexing or post-processing
 in the collector to support search. This does not disable the entire UI, as trace by ID and
 dependency queries still operate. Disable this when you use another service (such as logs) to find
 trace IDs. Defaults to true
+* `UI_ENABLED`: `false` disables UI, mounted at '/zipkin'. Defaults to true
```
```suggestion
* `UI_ENABLED`: `false` disables the web UI, mounted at '/zipkin'. Defaults to true
```

anuraaga (approved):

codefromthecrypt:
maybe SO doesn't like robots ... https://stackoverflow.com/questions/tagged/spring-boot

PR #3732, opened 2024-02-20
grpc: enables collector by default

codefromthecrypt:
The grpc service hasn't had any requests for change since introduction, and the code in armeria is stable. This enables the gRPC service by default (on both slim and normal builds as it only needs armeria).

This moves it out of experimental (something I neglected to do in 3.0, but better late than never).

cc @openzipkin/armeria

### DIFF
--- a/zipkin-server/src/test/java/zipkin2/server/internal/ITZipkinGrpcCollector.java
+++ b/zipkin-server/src/test/java/zipkin2/server/internal/ITZipkinGrpcCollector.java
@@ -104,7 +104,7 @@ class ITZipkinGrpcCollector {
         assertThat(responseBody.exhausted()).isTrue(); // We expect a single response
 
         // Now, verify the Length-Prefixed-Message
-        assertThat(compressedFlag).isEqualTo(0); // server didn't compress
+        assertThat(compressedFlag).isZero(); // server didn't compress
         assertThat(messageLength).isZero(); // there are no fields in ReportResponse
       }
     }
--- a/zipkin-server/src/test/java/zipkin2/server/internal/ITZipkinGrpcCollector.java
+++ b/zipkin-server/src/test/java/zipkin2/server/internal/ITZipkinGrpcCollector.java
@@ -32,7 +32,6 @@ import zipkin2.TestObjects;
 import zipkin2.codec.SpanBytesDecoder;
 import zipkin2.codec.SpanBytesEncoder;
 import zipkin2.proto3.ListOfSpans;
-import zipkin2.proto3.ReportResponse;
 import zipkin2.storage.InMemoryStorage;
 
 import static okhttp3.Protocol.H2_PRIOR_KNOWLEDGE;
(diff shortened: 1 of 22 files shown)


zipkin-server/src/test/java/zipkin2/server/internal/ITZipkinGrpcCollector.java
```diff
@@ -80,30 +79,33 @@ class ITZipkinGrpcCollector {
...
-      .post(RequestBody.create(requestBody.snapshot(), MediaType.get("application/grpc")))
-      .build())
-      .execute();
-
-    BufferedSource responseBody = response.body().source();
-    assertThat((int) responseBody.readByte()).isEqualTo(0); // uncompressed
-    long encodedLength = responseBody.readInt() & 0xffffffffL;
-
-    return ReportResponse.ADAPTER.decode(responseBody);
+  void callReport(ListOfSpans spans) throws IOException {
```
codefromthecrypt, line 84: I refactored this to close resources.. only other change is verifying the encodedLength is the same as the input. Since the result has no fields, I changed to void return

reta (approved):

zipkin-server/src/test/java/zipkin2/server/internal/ITZipkinGrpcCollector.java
```diff
@@ -80,30 +79,33 @@ class ITZipkinGrpcCollector {
...
+        .url(url(server, "/zipkin.proto3.SpanService/Report"))
+        .addHeader("te", "trailers")
+        .post(RequestBody.create(requestBody.snapshot(), MediaType.get("application/grpc")))
+        .build();
+      try (Response response = client.newCall(request).execute();
+           BufferedSource responseBody = response.body().source()) {
+
+        assertThat((int) responseBody.readByte()).isEqualTo(0); // uncompressed
+        long encodedLength = responseBody.readInt() & 0xffffffffL;
+        assertThat(encodedLength).isEqualTo(encodedMessage.size());
```
anuraaga: I must be missing something but I don't quite understand why the response length is equal to the request message length. Shouldn't it be 0 since the response proto is empty?

codefromthecrypt: good catch haha.. let me look at this

codefromthecrypt: yeah really bad of me to match the two words without thinking. we exhausted the encoded message so it is zero.

codefromthecrypt: thanks, PTAL!

minwoox (approved):
Looks good. 👍 
Could you add @ikhoon and @jrhee17 to @openzipkin/armeria?

codefromthecrypt:
> Could you add @ikhoon and @jrhee17 to @openzipkin/armeria?

gladly and done!

Issue #3412, opened 2021-12-19
table doesn't exist in 2.23.15


withinboredom:
## Describe the Bug
A clear and concise description of what the bug is. If you have a solution in mind, skip raising an issue and open a pull request instead.

Started zipkin with an already existing cassandra instance, with `CASSANDRA_ENSURE_SCHEMA`. The logs indicate the migrations are run:

```

                  oo
                 oooo
                oooooo
               oooooooo
              oooooooooo
             oooooooooooo
           ooooooo  ooooooo
          oooooo     ooooooo
         oooooo       ooooooo
        oooooo   o  o   oooooo
       oooooo   oo  oo   oooooo
     ooooooo  oooo  oooo  ooooooo
    oooooo   ooooo  ooooo  ooooooo
   oooooo   oooooo  oooooo  ooooooo
  oooooooo      oo  oo      oooooooo
  ooooooooooooo oo  oo ooooooooooooo
      oooooooooooo  oooooooooooo
          oooooooo  oooooooo
              oooo  oooo

     ________ ____  _  _____ _   _
    |__  /_ _|  _ \| |/ /_ _| \ | |
      / / | || |_) | ' / | ||  \| |
     / /_ | ||  __/| . \ | || |\  |
    |____|___|_|   |_|\_\___|_| \_|

:: version 2.23.15 :: commit 63365bb ::

2021-12-19 20:15:59.014  INFO [/] 1 --- [oss-http-*:9411] c.l.a.s.Server                           : Serving HTTP at /[0:0:0:0:0:0:0:0%0]:9411 - http://127.0.0.1:9411/
2021-12-19 20:16:39.527  INFO [/] 1 --- [cking-tasks-1-1] z.s.c.Schema                             : Installing schema /zipkin2-schema.cql for keyspace zipkin2
2021-12-19 20:16:45.237  INFO [/] 1 --- [cking-tasks-1-1] z.s.c.Schema                             : Installing indexes /zipkin2-schema-indexes.cql for keyspace zipkin2
2021-12-19 20:16:48.649  INFO [/] 1 --- [ing-tasks-1-177] z.s.c.Schema                             : Upgrading schema /zipkin2-schema-upgrade-1.cql
2021-12-19 20:16:49.819  INFO [/] 1 --- [ing-tasks-1-177] z.s.c.Schema                             : Upgrading schema /zipkin2-schema-upgrade-2.cql
[...]

PR #3740, opened 2024-02-28
cassandra: adds log statement to help debug schema install failure

codefromthecrypt:
I'm not on a good enough network at the moment to pull cassandra's image, but after this is merged, troubleshooting #3412 should be easier with normal docker and logging config


reta (approved):

codefromthecrypt:
FWIW it wasn't about this, it was that sasi is disabled by default (added in 3.x enabled by default, and then disabled in 4.x with no property override!)

PR #3748, opened 2024-03-05
Fix ServerIntegratedBenchmark

### LINKED ISSUE #3646: ServerIntegratedBenchmark no longer runs
## Describe the Bug

ServerIntegratedBenchmark doesn't finish anymore. While this is not a critical tool, it is helpful and should work. Particularly, the zipkin container doesn't come up (at least on my M1 mac) even though it works in normal 

## Steps to Reproduce

Try running ServerIntegratedBenchmark

## Expected Behaviour

All the storage options should pass

anuraaga:
Fixes #3646 

I found some issues

- log backend is log4j, not simple logger so lack of file prevented any logging to debug. I figured replacing the config is better than tracking out the log4j transitive
- The netty version override didn't seem to be kicking in. Sorry, I'm not generally experienced with Maven and also rusty in Java so just redefining is the only approach I could find
- Something about the healthcheck change doesn't seem to interact well with testcontainers. Doing a direct ping without involving curl instead seems fine to me

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -179,6 +179,44 @@
       <version>${testcontainers.version}</version>
       <scope>test</scope>
     </dependency>
+
+    <!-- overrides from zipkin-server project are not picked up here so redefine -->
+    <!-- temporary until netty-4.1.108.Final -->
+    <dependency>
+      <groupId>io.netty</groupId>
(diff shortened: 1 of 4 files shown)


codefromthecrypt (approved):
great. thanks for taking the time on this!

PR #3749, opened 2024-03-05
license: removes copyright year and uses SPDX ID

codefromthecrypt:
As a small project, we have to conserve resources and not sign up for work that isn't required. I've recently realized many [commercial](https://github.com/hashicorp/consul/blob/main/command/cli/cli.go#L1-L2) and/or [CNCF projects](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/trace/Tracer.java#L1-L4) both use SPDX IDs and also don't bother with copyright year. If their legal team is ok with this, surely a volunteer team without access to one should be, too!

Doing so accomplishes the following, all of which help volunteers:
* significantly increases readability of files, particularly small ones
* removes beginning of year maintenance, which cause a lot of FUD last year. IIRC some deployment failed and hours were spent in spite of docs.
* eliminates the need to do a full source check out, just to satisfy the license plugin. This means we can use actions/checkout defaults.

(same change as https://github.com/openzipkin/zipkin-reporter-java/pull/257)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -23,7 +23,8 @@ jobs:
           # Prevent use of implicit GitHub Actions read-only GITHUB_TOKEN
           # because javadoc_to_gh_pages pushes commits to the gh-pages branch.
           token: ${{ secrets.GH_TOKEN }}
-          # allow build-bin/javadoc_to_gh_pages to get the full history
+          # Allow build-bin/javadoc_to_gh_pages to fetch and push gh-pages
+          # See https://github.com/actions/checkout/issues/578
           fetch-depth: 0
       - name: Setup java
         uses: actions/setup-java@v4
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -17,17 +17,8 @@ If you are adding a new file it should have a header like below. This
 can be automatically added by running `./mvnw com.mycila:license-maven-plugin:format`.
 
 ```
-/**
- * Copyright 2019 The OpenZipkin Authors
- *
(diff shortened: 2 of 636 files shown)


.github/workflows/deploy.yml
```diff
@@ -23,7 +23,8 @@ jobs:
           # Prevent use of implicit GitHub Actions read-only GITHUB_TOKEN
           # because javadoc_to_gh_pages pushes commits to the gh-pages branch.
           token: ${{ secrets.GH_TOKEN }}
-          # allow build-bin/javadoc_to_gh_pages to get the full history
+          # Allow build-bin/javadoc_to_gh_pages to fetch and push gh-pages
+          # See https://github.com/actions/checkout/issues/578
```
codefromthecrypt, line 27: notable thing is the comment was imprecise

anuraaga (approved):

codefromthecrypt:
one nice side effect is the raw HTML isn't dominated by license text anymore ;)

```bash
$ curl -s localhost:9411/zipkin/index.html
<!--

    Copyright The OpenZipkin Authors
    SPDX-License-Identifier: Apache-2.0

-->
<!DOCTYPE html>
<html>
  <head>
    <base href="/zipkin" >
    <meta charset="utf-8" />
    <link rel="icon" href="/zipkin/favicon.ico" />
    <title>Zipkin</title>
    <script type="module" crossorigin src="/zipkin/static/js/index.51c4d5b0.js"></script>
    <link rel="stylesheet" href="/zipkin/static/css/index.a7aa86b4.css">
  </head>
  <body>
    <noscript>You need to enable JavaScript to run this app.</noscript>
    <div id="root"></div>
    
  </body>
</html>
```

PR #3757, opened 2024-04-13
ci: pins markdown-link-check due to regression

codefromthecrypt:
see https://github.com/tcort/markdown-link-check/issues/304


codefromthecrypt:
docker is also broken (bad build day sheesh) fixed by #3758, which is broken due to lint until this is merged ;)

PR #3758, opened 2024-04-13
docker: fix cassandra pip install and bump all images

codefromthecrypt:
cqlsh 4.x is not compatible with Python 3.12 by default.

See https://issues.apache.org/jira/browse/CASSANDRA-19206


docker/test-images/zipkin-cassandra/install.sh
```diff
@@ -170,7 +170,7 @@ is_cassandra_alive() {
 is_cassandra_alive || exit 1
 
 echo "*** Installing cqlsh"
-apk add --update --no-cache python3 py3-pip
```
codefromthecrypt, line 173: these fought over python 3.11 vs 3.12, so installing via python instead

PR #3763, opened 2024-04-16
Raises floor JRE version from 11 to 17 except core

codefromthecrypt:
This raises the floor JRE version of libraries except core from 11 to 17. The only reason we had 17 in the past was due to Spark limitations that affected zipkin-dependencies. After this, folks can also consider getting rid of AutoValue for the record type (sadly no OpenRewrite for that ;).

This change was manual except the parts about string formatting, exception coercing and multi-line literals. Those parts were done with OpenRewrite via below (reverting changes to core):

```bash
$ mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-migrate-java:RELEASE -Drewrite.activeRecipes=org.openrewrite.java.migrate.UpgradeToJava17
```

The other non-manual step was about JOOQ, which was copy/pasting commands from the README, which worked without any issue.


zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/VersionSpecificTemplates.java
```diff
@@ -111,23 +111,25 @@ String spanIndexTemplate(ElasticsearchVersion version) {
...
-        + "      },\n"
-        + "      \"filter\": {\n"
-        + "        \"traceId_filter\": {\n"
-        + "          \"type\": \"pattern_capture\",\n"
-        + "          \"patterns\": [\"([0-9a-f]{1,16})$\"],\n"
-        + "          \"preserve_original\": true\n"
-        + "        }\n"
-        + "      }\n"
-        + "    }\n");
+      result += ("""
```
codefromthecrypt, line 114: this one is notably easier to read/edit!

zipkin-storage/mysql-v1/pom.xml
```diff
@@ -20,8 +20,7 @@
   <properties>
     <main.basedir>${project.basedir}/../..</main.basedir>
 
-    <!-- 3.16 is the last to compile with JDK 11 -->
```
codefromthecrypt, line 23: no more revlock

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/IndexNameFormatter.java
```diff
@@ -156,8 +156,8 @@ public List<String> formatTypeAndRange(@Nullable String type, long beginMillis,
   }
 
   String formatIndexPattern(String format, GregorianCalendar current, String prefix) {
-    return String.format(
-      format,
+    return 
+      format.formatted(
```
anuraaga: I think this indent isn't intentional (move format.formatted back to previous line?)

anuraaga (approved):

shakuzen (approved):

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/internal/IndexNameFormatter.java
codefromthecrypt: openrewrite had its own ideas on this one ;) will move it back, as yeah looks better

codefromthecrypt:
thanks for the eyes, folks

Issue #3634, opened 2023-12-13
flakey test: ZipkinExtensionTest.postSpans_disconnectDuringBody


codefromthecrypt:
#3631 Tried to solve this by closing resources, but that wasn't enough

```
 Error:  zipkin2.junit5.ZipkinExtensionTest.postSpans_disconnectDuringBody -- Time elapsed: 0.034 s <<< ERROR!
java.io.IOException: unexpected end of stream on http://localhost:43457/...
	at okhttp3.internal.http1.Http1ExchangeCodec.readResponseHeaders(Http1ExchangeCodec.kt:209)
	at okhttp3.internal.connection.Exchange.readResponseHeaders(Exchange.kt:111)
	at okhttp3.internal.http.CallServerInterceptor.intercept(CallServerInterceptor.kt:95)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.kt:34)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:95)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.kt:84)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.kt:65)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
	at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:205)
	at okhttp3.internal.connection.RealCall.execute(RealCall.kt:158)
	at zipkin2.junit5.ZipkinExtensionTest.postSpansV1(ZipkinExtensionTest.java:176)
	at zipkin2.junit5.ZipkinExtensionTest.assertPostSpansV1Success(ZipkinExtensionTest.java:184)
	at zipkin2.junit5.ZipkinExtensionTest.postSpans_disconnectDuringBody(ZipkinExtensionTest.java:131)
[...]

PR #3765, opened 2024-04-17
Supports OpenSearch V2 through ES_* environment variables

reta:
## Add Zipkin support of OpenSearch storage

The Elasticsearch storage now supports OpenSearch as a backed as well. The implementation relies on `distribution` property that is returned as part of Elasticsearch / OpenSearch HTTP `GET /` API. Please note that the storage version is now abstracted as `BaseVersion` with two implementation: `ElasticsearchVersion` and `OpensearchVersion`. 

Although OpenSearch is a fork of Elasticsearch as of 7.10.2, the projects diverged sufficiently far from each other. Luckily, `Zipkin` relies on the features that have not been impacted (so far) and work the same way across both projects.

- [x] Add javadocs
- [x] Add documentation

### DIFF
--- a/zipkin-storage/elasticsearch/README.md
+++ b/zipkin-storage/elasticsearch/README.md
@@ -3,7 +3,7 @@
 This is a plugin to the Elasticsearch storage component, which uses
 HTTP by way of [Armeria](https://github.com/line/armeria) and
 [Jackson](https://github.com/FasterXML/jackson). This uses Elasticsearch 5+
-features, but is tested against Elasticsearch 7-8.x.
+features, but is tested against Elasticsearch 7-8.x and OpenSearch 2.x.
 
 ## Multiple hosts
 Most users will supply a DNS name that's mapped to multiple A or AAAA
@@ -26,9 +26,9 @@ with one difference described below. We add a "timestamp_millis" field
 to aid in integration with other tools.
 
 ### Timestamps
-Zipkin's timestamps are in epoch microseconds, which is not a supported date type in Elasticsearch.
+Zipkin's timestamps are in epoch microseconds, which is not a supported date type in Elasticsearch / OpenSearch.
 In consideration of tools like like Kibana, this component adds "timestamp_millis" when writing
-spans. This is mapped to the Elasticsearch date type, so can be used to any date-based queries.
+spans. This is mapped to the Elasticsearch / OpenSearch date type, so can be used to any date-based queries.
 
 ## Indexes
 Spans are stored into daily indices, for example spans with a timestamp
@@ -68,8 +68,9 @@ $ curl -s 'localhost:9200/zipkin*span-2017-08-11/_search?q=_q:error=500'
 
 The reason for special casing is around dotted name constraints. Tags
 are stored as a dictionary. Some keys include inconsistent number of dots
-(ex "error" and "error.message"). Elasticsearch cannot index these as it
-inteprets them as fields, and dots in fields imply an object path.
+(ex "error" and "error.message"). Elasticsearch / OpenSearch cannot index
+these as it inteprets them as fields, and dots in fields imply an object 
+path.
 
 ### Trace Identifiers
 Unless `ElasticsearchStorage.Builder.strictTraceId` is set to false,
@@ -122,9 +123,9 @@ be written, nor analyzed.
 
 ### Composable Index Template
 Elasticsearch 7.8 introduces [composable templates](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-templates.html) and
-deprecates [legacy/v1 templates](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates-v1.html) used in version prior.
-Merging of multiple templates with matching index patterns is no longer allowed, and Elasticsearch will return error on PUT of the second template
-with matching index pattern and priority. Templates with matching index patterns are required to have different priorities, and Elasticsearch will
+deprecates [legacy/v1 templates](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates-v1.html) used in version prior (fully supported by OpenSearch).
+Merging of multiple templates with matching index patterns is no longer allowed, and Elasticsearch / OpenSearch will return error on PUT of the second template
+with matching index pattern and priority. Templates with matching index patterns are required to have different priorities, and Elasticsearch / OpenSearch will
 only use the template with the highest priority. This also means that [secondary template](https://gist.github.com/codefromthecrypt/1af1259102e7a2da1b3c9103565165d7)
 is no longer achievable.
 
@@ -133,7 +134,15 @@ providing `ES_TEMPLATE_PRIORITY` environment variable.
 
 ## Customizing the ingest pipeline
 
+### Elasticsearch
+
 You can setup an [ingest pipeline](https://www.elastic.co/guide/en/elasticsearch/reference/master/pipeline.html) to perform custom processing.
+
+### OpenSearch
+
+You can setup an [ingest pipeline](https://opensearch.org/docs/latest/ingest-pipelines/) to perform custom processing.
+
+### Setting up ingest pipeline
 
 Here's an example, which you'd setup prior to configuring Zipkin to use
 it via `ElasticsearchStorage.Builder.pipeline`
@@ -162,7 +171,12 @@ to reduce load. This is implemented by
 [DelayLimiter](../../zipkin/src/main/java/zipkin2/internal/DelayLimiter.java)
 
 ## Data retention
-Zipkin-server does not handle retention management of the trace data. Use the tools recommended by ElasticSearch to manage data retention, or your cluster
-will grow indefinitely!
+Zipkin-server does not handle retention management of the trace data. Use the tools recommended by to manage data retention, or your cluster will grow indefinitely!
+
+### Elasticsearch
  * [Curator](https://www.elastic.co/guide/en/elasticsearch/client/curator/current/index.html)
  * [Index Lifecycle Management](https://www.elastic.co/guide/en/elasticsearch/reference/7.3/index-lifecycle-management.html)
+
+### OpenSearch
+ * [Curator](https://github.com/flant/curator-opensearch)
+ * [Index Lifecycle Management](https://opensearch.org/docs/latest/im-plugin/ism/index/)
--- a/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/BaseVersion.java
+++ b/zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/BaseVersion.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright The OpenZipkin Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package zipkin2.elasticsearch;
+
+import static zipkin2.elasticsearch.internal.JsonReaders.enterPath;
+
+import java.io.IOException;
+import java.util.function.Supplier;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.linecorp.armeria.common.AggregatedHttpRequest;
+import com.linecorp.armeria.common.HttpMethod;
+
+import zipkin2.elasticsearch.internal.client.HttpCall;
+
+/**
+ * Base version for both Elasticsearch and OpenSearch distributions.
+ */
+public abstract class BaseVersion {
+  final int major, minor;
+
+  BaseVersion(int major, int minor) {
+    this.major = major;
+    this.minor = minor;
+  }
+
+  /**
+   * Gets the version for particular distribution, returns either {@link ElasticsearchVersion}
+   * or {@link OpensearchVersion} instance.
+   * @param http the HTTP client
+   * @return either {@link ElasticsearchVersion} or {@link OpensearchVersion} instance
+   * @throws IOException in case of I/O errors
+   */
+  static BaseVersion get(HttpCall.Factory http) throws IOException {
+    return Parser.INSTANCE.get(http);
+  }
+
+  /**
+   * Does this version of Elasticsearch / OpenSearch still support mapping types?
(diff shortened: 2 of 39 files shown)


zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/OpenSearchExtension.java
```diff
@@ -0,0 +1,31 @@
...
+  static final Logger LOGGER = LoggerFactory.getLogger(OpenSearchExtension.class);
+
+  OpenSearchExtension(int majorVersion) {
+    super(new OpenSearchContainer(majorVersion));
+  }
+
+  // mostly waiting for https://github.com/testcontainers/testcontainers-java/issues/3537
+  static final class OpenSearchContainer extends GenericContainer<OpenSearchContainer> {
+      OpenSearchContainer(int majorVersion) {
+      super(parse("ghcr.io/openzipkin/zipkin-opensearch" + majorVersion + ":3.3.1"));
```
reta: @codefromthecrypt could you clarify a thing for me here please: I prepared this image for testing in this pull request but I need it published before, how it usually works? thanks!

reta: (PS: I've been using the image locally for testing)

codefromthecrypt:
I think it is easiest to make the image first in a separate PR. Then, merge and we can use it.

I don't recall the history of the feature ask, but if someone hasn't asked for 1.x maybe better to only do 2.x because there's a linear time increase per images.  Usually we don't add old versions as a first storage option.

At any rate, you would do a PR for the image, and make sure it is in the docker readme test workflows etc. When that merges, after master builds clean, you can re-cut the last version docker images as docker-x.x.x. Finally, as it is a new image, it will be private by default, so you need to click to make it public https://github.com/orgs/openzipkin/packages

codefromthecrypt:
@reta for the error around `ghcr.io/openzipkin/zipkin-opensearch2:3.3.1` you want to push a tag docker-3.3.1 off latest master. this will re-publish and also include the new one, unless we missed a spot somewhere.

codefromthecrypt:
also don't forget to set this public https://github.com/orgs/openzipkin/packages/container/zipkin-opensearch2/settings

reta:
> @reta for the error around `ghcr.io/openzipkin/zipkin-opensearch2:3.3.1` you want to push a tag docker-3.3.1 off latest master. this will re-publish and also include the new one, unless we missed a spot somewhere.

Thanks @codefromthecrypt , it seems like we need to make a 3.3.1 release first since Docker images are looking for Zipkin 3.3.1 bits ...

codefromthecrypt:
@reta oh sorry.. I didn't remember what release was current. change the test code to use 3.3.0 and re-cut docker-3.3.0. that's the simplest way and will get the new test image

reta:
> also don't forget to set this public https://github.com/orgs/openzipkin/packages/container/zipkin-opensearch2/settings

@codefromthecrypt apologies, I see the image is 100% pushed but I cannot see it in the `Packages`, is it my permissions? ~(because it seems like I cannot change visibility level for existing packages as well)~ I see that I could change the visibility for existing packages

codefromthecrypt:
switched visibility to public ;) https://github.com/orgs/openzipkin/packages/container/zipkin-opensearch2/settings

codefromthecrypt:
when ready, re-prep the PR description, after making sure any markdown that need a usage example update have been so. Also, anything in general READMEs about things we test mention opensearch since after this, it is true.

codefromthecrypt:
also bump poms 3.3.1-SNAPSHOT -> 3.4.0-SNAPSHOT as def not a patch ;)

codefromthecrypt:
also I think a question originally was about hosted aws opensearch (zipkin-aws). want to try integrating with that via snapshot locally prior to mark this for review?

reta:
> also I think a question originally was about hosted aws opensearch (zipkin-aws). want to try integrating with that via snapshot locally prior to mark this for review?

Absolutely, thanks @codefromthecrypt !

reta:
> also I think a question originally was about hosted aws opensearch (zipkin-aws). want to try integrating with that via snapshot locally prior to mark this for review?

@codefromthecrypt I have to admit failure to verify the flow, I was not able to get AWS managed OpenSearch instance(s) for testing purposes. On an optimistic note, AFAIK the managed OpenSearch uses the same open source distribution and should work seamlessly.

codefromthecrypt:
okie dokie. I'm a bit swamped, but if I've not reviewed this by saturday, nag me as I should be able to get time to look closely by then.

zipkin-storage/elasticsearch/README.md
```diff
@@ -3,7 +3,7 @@
 This is a plugin to the Elasticsearch storage component, which uses
```
codefromthecrypt, line 3: thanks for the detailed updates here

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ITOpenSearchStorageV2.java
```diff
@@ -0,0 +1,32 @@
...
+class ITOpenSearchStorageV2 extends ITElasticsearchStorage {
+
+  @RegisterExtension static OpenSearchExtension opensearch = new OpenSearchExtension(2);
+
+  @Override OpenSearchExtension elasticsearch() {
+    return opensearch;
+  }
+
+  @Nested
+  class ITEnsureIndexTemplate extends zipkin2.elasticsearch.integration.ITEnsureIndexTemplate {
```
codefromthecrypt, line 27: I forget.. this appears duplicated and is using the method `elasticsearch()` not the field. Can this be pulled up?

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/OpensearchVersion.java
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright The OpenZipkin Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package zipkin2.elasticsearch;
+
+import java.util.Objects;
+
+/** Helps avoid problems comparing versions by number. Ex 2.10 should be > 2.9 */
+public final class OpensearchVersion extends BaseVersion implements Comparable<OpensearchVersion> {
```
codefromthecrypt, line 10: please unit test this class or the base one?

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchSpecificTemplates.java
```diff
@@ -0,0 +1,107 @@
...
+import static zipkin2.elasticsearch.ElasticsearchVersion.V5_0;
+import static zipkin2.elasticsearch.ElasticsearchVersion.V6_0;
+import static zipkin2.elasticsearch.ElasticsearchVersion.V6_7;
+import static zipkin2.elasticsearch.ElasticsearchVersion.V7_0;
+import static zipkin2.elasticsearch.ElasticsearchVersion.V7_8;
+import static zipkin2.elasticsearch.ElasticsearchVersion.V9_0;
+
+import zipkin2.internal.Nullable;
+
+final class ElasticsearchSpecificTemplates extends VersionSpecificTemplates<ElasticsearchVersion> {
```
codefromthecrypt, line 17: if there is new interesting logic here, can you add to the existing unit test?

codefromthecrypt (approved):
Thanks for doing this! As far as I understand, this makes a heuristic internally to detect OS vs ES, so that the server config pretty much remains the same. Do you know if zipkin-dependencies works with OS using the current driver? Because merging this without dependencies working will immediately lead to another problem.

p.s. maybe cc anyone who has asked for OS on this PR so that they can know about it directly

reta:
> Do you know if zipkin-dependencies works with OS using the current driver?

Ah, thanks a lot for pointing it out @codefromthecrypt, it won't work with Elasticsearch, we will need to add OpenSearch module alongside the Elasticsearch one.

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/ElasticsearchSpecificTemplates.java
reta: Sure, `ElasticsearchSpecificTemplates` seems to be covered but `OpensearchSpecificTemplates`, I will add tests shortly

zipkin-storage/elasticsearch/src/test/java/zipkin2/elasticsearch/integration/ITOpenSearchStorageV2.java
reta: Sorry, you mean replace `elasticsearch()` -> `elasticsearch`? (or better to say `opensearch` in this case)

reta:
Hm ... `lint` check failure is new:

```
 build-bin/configure_lint: 5: markdown-link-check: not found

added 66 packages in 3s

16 packages are looking for funding
  run `npm fund` for details
usage: yamllint [-h] [-] [-c CONFIG_FILE | -d CONFIG_DATA] [--list-files]
                [-f {parsable,standard,colored,github,auto}] [-s]
                [--no-warnings] [-v]
                [FILE_OR_DIR ...]
yamllint: error: one of the arguments FILE_OR_DIR - is required
```

codefromthecrypt:
let's hold merging this until that's sorted out as I don't want to have a new storage option that doesn't work due to zipkin-dependencies (bomb)

reta:
> let's hold merging this until that's sorted out as I don't want to have a new storage option that doesn't work due to zipkin-dependencies (bomb)

I could pick it up later this week, very likely over the weekend (the good thing, AFAIK OpenSearch has Spark / Hadoop support on par with Elasticsearch)

codefromthecrypt:
thanks @reta!

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/OpensearchVersion.java
reta: Did both, thanks!

zipkin-storage/elasticsearch/src/main/java/zipkin2/elasticsearch/BaseVersion.java
```diff
@@ -0,0 +1,108 @@
...
+import com.fasterxml.jackson.core.JsonToken;
+import com.linecorp.armeria.common.AggregatedHttpRequest;
+import com.linecorp.armeria.common.HttpMethod;
+
+import zipkin2.elasticsearch.internal.client.HttpCall;
+
+/**
+ * Base version for both Elasticsearch and OpenSearch distributions.
+ */
+public abstract class BaseVersion {
```
codefromthecrypt, line 25: @reta so for the dependencies job, one way is to flatten this to just Version and not use a type hierarchy to retrieve the distribution property.

Then, ElasticsearchStorage.version() can return this and the depedencies main class can handle if it is "opensearch" or not (assume ES if not literally opensearch).

codefromthecrypt: only issue would be if armeria harms the dep tree of zipkin-dependencies, because right now, we only use zipkin-storage-elasticsearch as a test dep

codefromthecrypt: anyway a simple http client that gets the distribution property from json shouldn't be too bad worst case.

reta: > @reta so for the dependencies job, one way is to flatten this to just Version and not use a type hierarchy to retrieve the distribution property.

I was thinking about it, here is the deal: the delendencies job does not depend on zipkin-elasticsearch nor armenia, so reusing this exact approach is a bit problematic. However, we could use simple one time HTTP call instead (using HttpClient), basically just checking the distribution, not even version. Wdyt?

reta: > anyway a simple http client that gets the distribution property from json shouldn't be too bad worst case.

Haha, while I was typing my reply you basically came to the same conclusions 😄

reta:
And https://github.com/openzipkin/zipkin/issues/3634 strikes again

reta:
@codefromthecrypt this is super funny:

```
 ERROR: 1 dead links found!
  [✖] https://docs.datastax.com/en/developer/java-driver/4.17/manual/core/request_tracker/#request-logger → Status: 404
```

The link opens ... but:

```
$ curl -iv https://docs.datastax.com/en/developer/java-driver/4.17/manual/core/request_tracker/#request-logger

HTTP/1.1 404 Not Found
Content-Type: text/html
Content-Length: 36054
Connection: keep-alive
Last-Modified: Wed, 22 May 2024 18:02:05 GMT
x-amz-server-side-encryption: AES256
Accept-Ranges: bytes
Server: AmazonS3
Date: Fri, 24 May 2024 18:08:32 GMT
ETag: "1e69f79408df5da533fcc320698669a9"
Vary: Accept-Encoding
X-Cache: Error from cloudfront
Via: 1.1 10f978be0e76903f65b35bee6581c310.cloudfront.net (CloudFront)
X-Amz-Cf-Pop: YUL62-P2
X-Amz-Cf-Id: w5XQ_h8cv8834R5pBln1oIEpVltBAX6SnR56UMhk6f3F-cjtj0diVw==
Age: 73337
X-XSS-Protection: 1; mode=block
X-Frame-Options: SAMEORIGIN
Content-Security-Policy: child-src 'self'; frame-src youtube.com https://www.youtube.com; worker-src 'self' blob:
X-Content-Type-Options: nosniff
Strict-Transport-Security: max-age=31536000

<!DOCTYPE html>
<html lang="en">
  <head>
 ....
```

codefromthecrypt:
> this is super funny:

yes, no PR is complete without unrelated bruising from lint checkers ;)

codefromthecrypt:
ok! finally time to ship this 👍

PR #3766, opened 2024-04-19
Add Zipkin support of OpenSearch storage (add Docker images)

reta:
Add Zipkin support of OpenSearch storage (add Docker images) as dicussed https://github.com/openzipkin/zipkin/pull/3765#issuecomment-2067097535

### DIFF
--- a/docker/test-images/zipkin-opensearch2/Dockerfile
+++ b/docker/test-images/zipkin-opensearch2/Dockerfile
@@ -1,0 +1,70 @@
+#
+# Copyright The OpenZipkin Authors
+# SPDX-License-Identifier: Apache-2.0
+#
+
+# java_version is used for install and runtime layers of zipkin-opensearch2
+#
+# Use latest version here: https://github.com/orgs/openzipkin/packages/container/package/java
+# This is defined in many places because Docker has no "env" script functionality unless you use
+# docker-compose: When updating, update everywhere.
+ARG java_version=21.0.2_p13
+
+# We copy files from the context into a scratch container first to avoid a problem where docker and
+# docker-compose don't share layer hashes https://github.com/docker/compose/issues/883 normally.
+# COPY --from= works around the issue.
+FROM scratch as scratch
+
+COPY build-bin/docker/docker-healthcheck /docker-bin/
+COPY docker/test-images/zipkin-opensearch2/start-opensearch /docker-bin/
+COPY docker/test-images/zipkin-opensearch2/config/ /config/
+
+FROM ghcr.io/openzipkin/java:${java_version} as install
+
+WORKDIR /install
+
+# Use latest 2.x version from https://opensearch.org/downloads.html
+# This is defined in many places because Docker has no "env" script functionality unless you use
+# docker-compose: When updating, update everywhere.
+ARG opensearch2_version=2.13.0
+
+# Download only the OSS distribution (lacks X-Pack)
+RUN \
+# Connection resets are frequent in GitHub Actions workflows \
+wget --random-wait --tries=5 -qO- \
+# We don't download bin scripts as we customize for reasons including BusyBox problems
+https://artifacts.opensearch.org/releases/bundle/opensearch/${opensearch2_version}/opensearch-${opensearch2_version}-linux-x64.tar.gz| tar xz \
+    --wildcards --strip=1 --exclude=jdk
+COPY --from=scratch /config/ ./config/
+
+# Use a full Java distribution rather than adding test modules to the
+# production -jre base layer used by zipkin and zipkin-slim.
+FROM ghcr.io/openzipkin/java:${java_version} as zipkin-opensearch2
+LABEL org.opencontainers.image.description="OpenSearch distribution on OpenJDK and Alpine Linux"
+ARG opensearch2_version=2.13.0
+LABEL opensearch-version=$opensearch2_version
+
+# The full license is also included in the image at /opensearch/LICENSE.txt.
+LABEL org.opencontainers.image.licenses="Apache-2.0"
+
+# Add HEALTHCHECK and ENTRYPOINT scripts into the default search path
+COPY --from=scratch /docker-bin/* /usr/local/bin/
+# We use start period of 30s to avoid marking the container unhealthy on slow or contended CI hosts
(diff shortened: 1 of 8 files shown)


codefromthecrypt:
the cve is eureka. I'll help raise a PR https://github.com/openzipkin/zipkin/actions/runs/8761712611?pr=3766

docker/test-images/zipkin-opensearch2/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+COPY --from=scratch /config/ ./config/
+
+# Use a full Java distribution rather than adding test modules to the
+# production -jre base layer used by zipkin and zipkin-slim.
+FROM ghcr.io/openzipkin/java:${java_version} as zipkin-opensearch2
+LABEL org.opencontainers.image.description="OpenSearch distribution on OpenJDK and Alpine Linux"
+ARG opensearch2_version=2.13.0
+LABEL opensearch-version=$opensearch2_version
+
+# Install bash
```
codefromthecrypt: why?

docker/test-images/zipkin-opensearch2/start-opensearch
```diff
@@ -0,0 +1,17 @@
...
+#
+# This intentionally locates config using the current working directory, in order to consolidate
+# Dockerfile instructions to WORKDIR
+set -eu
+
+# Configure the Docker HEALTHCHECK
+export HEALTHCHECK_PORT=9200
+export HEALTHCHECK_PATH=/_cluster/health
+
+exec ./bin/opensearch
```
codefromthecrypt: in all our other java images, including elasticsearch, we use the java commands directly. This prevents us to have to install things like bash, or accidentally have zombie or otherwise issues when the entrypoint has several processes below it. Can you try to do the same here?

docker/test-images/zipkin-opensearch2/Dockerfile
reta: OpenSearch startup script uses bash, I have tried to trick it with sh - did not work out

docker/test-images/zipkin-opensearch2/start-opensearch
codefromthecrypt: most notably this gives us JRE flexibility and transparency. As this is a test image, we want to be able to update the JRE even past what's formally specified sometimes. When the java args are here we get an advantage to modify to apply things like a missing open or whatever. One way to figure out what's going on is to run the normal docker image then `docker exec` into it and use PS or whatever if pid 1 isn't java. Hope this helps!

reta: Sure, it will take some time though (I've tried that inititially but got dragged into tons of startup issues)

codefromthecrypt: ack, now you know the labor I've done for many years ;)

codefromthecrypt: good news is typically the hardest is the first one. For example, ES was mechanical until 8.x when a lot of things changed. anyway thanks for having a look, as it saves me from doing the same.

docker/test-images/zipkin-opensearch2/Dockerfile
reta: (this is gone now)

docker/test-images/zipkin-opensearch2/start-opensearch
reta: should be all set here, @codefromthecrypt !

docker/test-images/zipkin-opensearch2/start-opensearch
```diff
@@ -0,0 +1,39 @@
...
+# This intentionally locates config using the current working directory, in order to consolidate
+# Dockerfile instructions to WORKDIR
+set -eu
+
+# Configure the Docker HEALTHCHECK
+export HEALTHCHECK_PORT=9200
+export HEALTHCHECK_PATH=/_cluster/health
+
+# Set tmp folder
+OPENSEARCH_TMPDIR=`java -cp 'classes:lib/*' org.opensearch.tools.launchers.TempDirectory`
```
codefromthecrypt: we are setting this to /tmp as an env variable in dockerfile to /tmp. do you want to always overwrite it like this?

codefromthecrypt: or does TempDirectory consider an existing OPENSEARCH_TMPDIR.. if so, maybe mention in the comment like below on OPENSEARCH_JAVA_OPTS

codefromthecrypt:
```suggestion
export OPENSEARCH_TMPDIR=`java -cp 'classes:lib/*' org.opensearch.tools.launchers.TempDirectory`
```

docker/test-images/zipkin-opensearch2/start-opensearch
```diff
@@ -0,0 +1,39 @@
...
+
+# The JVM options parser produces the final JVM options to start OpenSearch.
+# It does this by incorporating JVM options in the following way:
+#   - first, system JVM options are applied (these are hardcoded options in the
+#     parser)
+#   - second, JVM options are read from jvm.options and jvm.options.d/*.options
+#   - third, JVM options from OPENSEARCH_JAVA_OPTS are applied
+#   - fourth, ergonomic JVM options are applied
+OPENSEARCH_JAVA_OPTS=`export OPENSEARCH_TMPDIR; java -cp 'classes:lib/*' org.opensearch.tools.launchers.JvmOptionsParser "$PWD/config"`
```
codefromthecrypt:
```suggestion
OPENSEARCH_JAVA_OPTS=`java -cp 'classes:lib/*' org.opensearch.tools.launchers.JvmOptionsParser "$PWD/config"`
```

codefromthecrypt (approved):
looks good!

docker/test-images/zipkin-opensearch2/start-opensearch
reta: Good question, so `TempDirectory` uses `java.io.tmpdir`  and pre-creates the `opensearch` folder there

codefromthecrypt:
thanks for taking this forward, @reta!

PR #3769, opened 2024-05-23
Update to Spring Boot 3.3.0

reta:
Update to Spring Boot 3.3.0

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -11,7 +11,7 @@
   <parent>
     <groupId>io.zipkin</groupId>
     <artifactId>zipkin-parent</artifactId>
-    <version>3.3.1-SNAPSHOT</version>
+    <version>3.4.0-SNAPSHOT</version>
   </parent>
 
   <artifactId>benchmarks</artifactId>
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -11,7 +11,7 @@
   <parent>
     <groupId>io.zipkin</groupId>
     <artifactId>zipkin-parent</artifactId>
-    <version>3.4.0-SNAPSHOT</version>
+    <version>3.3.1-SNAPSHOT</version>
   </parent>
 
   <artifactId>benchmarks</artifactId>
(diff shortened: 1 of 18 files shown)


benchmarks/pom.xml
```diff
@@ -11,7 +11,7 @@
   <parent>
     <groupId>io.zipkin</groupId>
     <artifactId>zipkin-parent</artifactId>
-    <version>3.3.1-SNAPSHOT</version>
+    <version>3.4.0-SNAPSHOT</version>
```
codefromthecrypt: I don't think we should bump minor on this, rather bump minor afterwards when you add the elastic stuff.

docker/test-images/zipkin-mysql/Dockerfile
```diff
@@ -28,7 +28,7 @@ HEALTHCHECK --interval=1s --start-period=30s --timeout=5s CMD ["docker-healthche
 ENTRYPOINT ["start-mysql"]
 
 # Use latest from https://pkgs.alpinelinux.org/packages?name=mysql (without the -r[0-9])
-ARG mysql_version=10.11.7
+ARG mysql_version=10.11.8-r0
```
codefromthecrypt:
```suggestion
ARG mysql_version=10.11.8
```

codefromthecrypt (approved):
fine to make one for spring-boot only, but I tend to bump all deps before a patch release. If you are up to it, can you audit? Otherwise, I'll do post merge

benchmarks/pom.xml
codefromthecrypt: so after we bump all the deps, we cut 3.3.1, and after that's out we bump your PR for 3.4.0

reta: Got it, sure, I will revert that

reta:
> fine to make one for spring-boot only, but I tend to bump all deps before a patch release. If you are up to it, can you audit? Otherwise, I'll do post merge

Sure, will do!

reta:
@codefromthecrypt we should be all set, thanks for looking!

codefromthecrypt:
thanks a lot for the help, I'll cut a patch now!

codefromthecrypt:
actually I will bump alpine first

PR #3770, opened 2024-05-24
Fix potential github action smells

ceddy4395:
Hey! 🙂
I've made the following changes to your workflow:

- Use cache parameter instead of cache option
  -  Using the cache option from the setup/java is easier to configure and thus less prone to mistakes. Furthermore, it makes the workflow smaller and easier to read and understand. 
- Avoid jobs without timeouts
  - Jobs without a timeout can cause runners to be occupied when some process falls into an infinite loop or has multiple retries to perform a certain task, which will inevitably fail. Furthermore, it is also useful to be notified when a test-suite all of a sudden takes a lot longer than it used to, considering keeping tests fast is a good practice.

(These changes are part of a research Study at TU Delft looking at GitHub Action Smells. [Find out more](https://ceddy4395.github.io/research/gha-smells.html))

### DIFF
--- a/.github/workflows/test_readme.yml
+++ b/.github/workflows/test_readme.yml
@@ -28,6 +28,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     # skip commits made by the release plugin
     if: "!contains(github.event.head_commit.message, 'maven-release-plugin')"
+    timeout-minutes: 5
     strategy:
       matrix:
         include:  # Not ubuntu as already tested as a part of the docker job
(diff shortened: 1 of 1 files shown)


codefromthecrypt (approved):
looks good to me on green. thanks!

(you'll notice we have many repos with the same actions, the most production are zipkin-reporter-java, brave, zipkin-dependencies, zipkin-aws and zipkin-gcp if bored ;))

ceddy4395:
Thank you for the approval! 
I'll have a look if I have some time to spare! :) 

I noticed the `lint` job is failing because of some markdown links giving a 404? Would you like to have that fixed in this PR or in a different one?

codefromthecrypt:
maybe you can rebase this and possibly it will sort out those links? I think a recent change fixed it.

ceddy4395:
@codefromthecrypt Ah yes that did indeed fix the problems! Thank you

codefromthecrypt:
thanks for the help and teaching us new tricks @ceddy4395!

PR #3774, opened 2024-08-03
deps: updates all main and test java library versions

codefromthecrypt:


codefromthecrypt:
depends on https://github.com/openzipkin/zipkin/pull/3775 as the docker images reference package versions that no longer exist

PR #3775, opened 2024-08-03
Updates to latest docker images, aligning maven version

codefromthecrypt:
This uses latest java images, and aligns with the following tools provided by them:
* maven:3.9.8
* maven-help-plugin:3.4.1
* maven-dependency-plugin:3.7.1

To align, we run "mvn wrapper:wrapper -Dtype=bin" to update maven and manually change any pom or script references to share the above plugin versions.

Besides text replacement for current java versions, this also had to switch `docker-compose` to `docker compose` as the former is no longer in the base image.

All test images are updated to latest versions, as well.

This intentionally doesn't touch anything else, except bumping testcontainers. For example, this doesn't update maven plugins that aren't cached in the java image.


codefromthecrypt:
security check will fail until #3774 because this doesn't update main deps. so we can ignore it until then

codefromthecrypt:
actually security fail is npm but anyway we need this in so that the build can run.

PR #3778, opened 2024-08-25
Update Spring Boot (address CVEs) and few other dependencies

reta:
Update Spring Boot (address CVEs) and few other dependencies (see please https://spring.io/blog/2024/08/23/spring-boot-releases-fix-for-cve-2024-38807)

### DIFF
--- a/zipkin-collector/scribe/pom.xml
+++ b/zipkin-collector/scribe/pom.xml
@@ -58,5 +58,12 @@
       <version>${awaitility.version}</version>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <version>${junit-jupiter.version}</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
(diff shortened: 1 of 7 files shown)


zipkin-collector/scribe/pom.xml
```diff
@@ -58,5 +58,12 @@
       <version>${awaitility.version}</version>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
```
codefromthecrypt, line 64: should we do this in the parent pom, or is it because we have something that isn't jupiter...

codefromthecrypt (approved):
thanks!

I double-checked and alpine 3.20.2 and jdk 21.0.4_p7 are both still latest (for docker)

zipkin-collector/scribe/pom.xml
reta: > should we do this in the parent pom, or is it because we have something that isn't jupiter...

right, there are few (benchmarks, zipkin-lens) that do not use JUnit

PR #3780, opened 2024-10-07
Routine dependency updates

reta:
Routine dependency updates


codefromthecrypt (approved):
did a quick look on docker and while alpine has a patch release, we can wait until there's a new jre patch and roll forward both. right now JRE is same as we have https://pkgs.alpinelinux.org/packages?name=openjdk21

PR #3787, opened 2025-01-05
Update Spring Boot to 3.4.1 and other dependency updates

reta:
Update Spring Boot to 3.4.1 and other dependency updates

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -11,7 +11,7 @@
   <parent>
     <groupId>io.zipkin</groupId>
     <artifactId>zipkin-parent</artifactId>
-    <version>3.4.4-SNAPSHOT</version>
+    <version>3.5.0-SNAPSHOT</version>
   </parent>
 
   <artifactId>benchmarks</artifactId>
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -11,7 +11,7 @@
   <parent>
     <groupId>io.zipkin</groupId>
     <artifactId>zipkin-parent</artifactId>
-    <version>3.5.0-SNAPSHOT</version>
+    <version>3.4.4-SNAPSHOT</version>
   </parent>
 
   <artifactId>benchmarks</artifactId>
--- a/docker/test-images/zipkin-eureka/pom.xml
+++ b/docker/test-images/zipkin-eureka/pom.xml
@@ -29,7 +29,7 @@
       <dependency>
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-dependencies</artifactId>
-        <version>3.3.6</version>
+        <version>3.4.1</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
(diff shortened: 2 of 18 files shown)


benchmarks/pom.xml
```diff
@@ -11,7 +11,7 @@
   <parent>
     <groupId>io.zipkin</groupId>
     <artifactId>zipkin-parent</artifactId>
-    <version>3.4.4-SNAPSHOT</version>
+    <version>3.5.0-SNAPSHOT</version>
```
codefromthecrypt: if the boot jump isn't incompatible downstream I would leave it as a patch.

way to test this is to see if stackdriver still builds without code/config changes (only bumping deps)

codefromthecrypt (commented):
only thing I would suggest is not using a minor bump as it isn't a minor-worthy change. I would expect a change in yaml or code here or downstream if it were. Basically only do a minor when some useful feature is new (bumps aren't that) or we are forced into it (like JDK forces us into a JRE floor change)

benchmarks/pom.xml
reta: > if the boot jump isn't incompatible downstream I would leave it as a patch.

I haven't seen any incompatibility issues (other projects), but I will double-check on stackdriver , thanks!

codefromthecrypt (approved):
gracias!

benchmarks/pom.xml
reta: Build against 3.4.4-SNAPSHOT: 

```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Zipkin Google Cloud Platform (Parent) 2.2.7-SNAPSHOT:
[INFO]
[INFO] Zipkin Google Cloud Platform (Parent) .............. SUCCESS [  2.678 s]
[INFO] Zipkin to Stackdriver Trace Translation ............ SUCCESS [  8.927 s]
[INFO] Zipkin Storage: Google Stackdriver Trace ........... SUCCESS [ 17.712 s]
[INFO] Zipkin Server Module: GCP .......................... SUCCESS [ 18.256 s]
[INFO] Brave Encoder: Google Stackdriver Trace ............ SUCCESS [  6.159 s]
[INFO] Zipkin Encoder: Google Stackdriver Trace ........... SUCCESS [  5.138 s]
[INFO] Zipkin Sender: Google Stackdriver Trace ............ SUCCESS [ 10.173 s]
[INFO] brave-propagation-stackdriver ...................... SUCCESS [  1.091 s]
[INFO] Zipkin Sender: Google PubSub ....................... SUCCESS [ 13.385 s]
[INFO] collector-pubsub ................................... SUCCESS [  9.603 s]
[INFO] Zipkin GCP: Benchmarks ............................. SUCCESS [  8.506 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:41 min
[INFO] Finished at: 2025-01-06T11:10:41-05:00
[INFO] ------------------------------------------------------------------------
```

codefromthecrypt: thanks for being thorough!

reta:
@codefromthecrypt the release attempt (3.4.4) failed mid-though while uploading Docker images [1], do you think I should trigger the 3.4.5? (since Apache Maven artifacts seems to be published) Thank you.

[1] https://github.com/openzipkin/zipkin/actions/runs/12641661725/job/35224515103

PR #3788, opened 2025-01-24
Add Pulsar collector

CodePrometheus:
resolves #2297 

1. Add pulsar collector
2. Add tests for all changes
3. Add zipkin pulsar docker related files
4. Add corresponding usage documents

PR #3789, opened 2025-01-30
docker: add pulsar example

CodePrometheus:
Since there is no collector-pulsar module at present, compiling docker will result in an error, so when this PR is merged, I will add `zipkin-collector-pulsar` to the github actions in https://github.com/openzipkin/zipkin/pull/3788.


CodePrometheus:
@codefromthecrypt Could you kindly review this when you have some time? Thank you!

docker/examples/docker-compose-pulsar.yml
```diff
@@ -0,0 +1,32 @@
...
+
+version: '2.4'
+
+services:
+  pulsar:
+    image: ghcr.io/openzipkin/zipkin-pulsar:${TAG:-latest}
+    container_name: pulsar
+    ports: # expose the pulsar port so apps can publish spans.
+      - "6650:6650"
+      - "8080:8080"
```
codefromthecrypt: when ready for this change (docker-compose), note this is a common port, so maybe comment it out and say "uncomment to expose the UI port"

docker/test-images/zipkin-pulsar/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+# Add HEALTHCHECK and ENTRYPOINT scripts into the default search path
+COPY --from=scratch /docker-bin/* /usr/local/bin/
+# We use start period of 30s to avoid marking the container unhealthy on slow or contended CI hosts
+HEALTHCHECK --interval=1s --start-period=30s --timeout=5s CMD ["docker-healthcheck"]
+
+# Usually, we read env set from pid 1 to get docker-healthcheck parameters.
+# However, pulsar-server has to start as root even if permissions are dropped
+# later. So, we expose it in the Dockerfile instead.
+ENV HEALTHCHECK_PORT=8080
```
codefromthecrypt, line 31: shouldn't we check the service port 6650 (esp if we are using tcp)

codefromthecrypt (commented):
I think for this change, in order to be a proper dependency, should remove the docker-compose examples until there's server code to use them. Otherwise, I think it looks good to me as this should be enough to publish the image!

docker/test-images/zipkin-pulsar/Dockerfile
CodePrometheus: I modified it to HTTP on port 8080, which is enough for the healthcheck.

docker/examples/docker-compose-pulsar.yml
CodePrometheus: Ok, I'll modify it when it's ready.

docker/test-images/zipkin-pulsar/Dockerfile
codefromthecrypt: ok mainly I was concerned about race condition where 8080 is ready, but service port not. If this becomes a problem we can revisit

codefromthecrypt:
looks like the healthcheck approach is failing on 404. you will need a valid http path

https://github.com/openzipkin/zipkin/actions/runs/13084307860/job/36514939047?pr=3789

also, once done please lower the log level to warn as it will be unreadable to have info due to the health checks (examples shouldn't scroll too long due to chatty or health check in docker, we usually set to WARN)

CodePrometheus:
Well, I have tested it locally and it should be OK now.

docker/examples/README.md
```diff
@@ -118,6 +118,18 @@ $ docker compose -f docker-compose-rabbitmq.yml up
 Then configure the [RabbitMQ sender](https://github.com/openzipkin/zipkin-reporter-java/blob/master/amqp-client/src/main/java/zipkin2/reporter/amqp/RabbitMQSender.java)
 using a `host` value of `localhost` or a non-local hostname if in docker.
 
+
+## Pulsar
```
codefromthecrypt: Maybe wait for the next pr with server code to include README in docker compose example that will use it

codefromthecrypt (approved):
Small charge then merge. Thanks!

docker/examples/README.md
CodePrometheus: OK, removed, let me add it in the next pr.

CodePrometheus:
It seems that the error is unrelated to the changes and may be due to a network issue, could you please run it again.

reta (approved):

PR #3792, opened 2025-02-16
Updates docker images to latest base layer and package versions

codefromthecrypt:


codefromthecrypt:
security red will be fixed in #3793

PR #3799, opened 2025-04-06
Routine dependency updates

reta:
Routine dependency updates

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,12 +47,12 @@
     <zipkin-proto3.version>1.0.0</zipkin-proto3.version>
 
     <armeria.groupId>com.linecorp.armeria</armeria.groupId>
-    <armeria.version>1.31.3</armeria.version>
+    <armeria.version>1.32.3</armeria.version>
     <!-- Match Armeria version to avoid conflicts including running tests in the IDE -->
-    <netty.version>4.1.118.Final</netty.version>
+    <netty.version>4.1.119.Final</netty.version>
 
     <!-- It's easy for Jackson dependencies to get misaligned, so we manage it ourselves. -->
-    <jackson.version>2.18.2</jackson.version>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -170,24 +170,35 @@
          For example, if you imported Netty's BOM here, using Brave would also download that BOM as
          it depends indirectly on io.zipkin.zipkin2:zipkin. As Brave itself is indirectly used, this
          can be extremely confusing when people are troubleshooting library version assignments. -->
+    <dependencies>
+      <dependency>
+        <groupId>org.junit</groupId>
```
reta: Bringing JUnit BOM here since the runner requires `junit-platform-launcher` presence, which is versioned individually (f.e Jupiter 5.12.1, Launcher 1.12.1)

reta:
@codefromthecrypt I wondering if you have any insights why `trivy` workflow fails? (without starting to scan even) , thank you!

pom.xml
codefromthecrypt: I know it is annoying, but let's not import any boms in the top-level parent. there's a comment directly above explaining. it is annoying that there are paired versions, but it isn't the only pair we have, and this impacts anyone using the pom..

codefromthecrypt:
The failure goes to the summary screen. looks like zipkin-eureka needs a bump? https://github.com/openzipkin/zipkin/actions/runs/14298903065?pr=3799

reta:
> The failure goes to the summary screen. looks like zipkin-eureka needs a bump? https://github.com/openzipkin/zipkin/actions/runs/14298903065?pr=3799

Thanks a lot, sorry, had troubles founding the report :(

codefromthecrypt (approved):

codefromthecrypt:
thanks!

PR #3801, opened 2025-04-27
Update Docker images (and a few dependecy updates)

reta:
Update Docker images (and a few dependency updates)


codefromthecrypt (approved):
thanks for the bumps!

PR #3803, opened 2025-06-01
Update to Spring Boot 3.5.0

reta:
Update to Spring Boot 3.5.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
 
   <groupId>io.zipkin</groupId>
   <artifactId>zipkin-parent</artifactId>
-  <version>3.5.2-SNAPSHOT</version>
+  <version>3.6.0-SNAPSHOT</version>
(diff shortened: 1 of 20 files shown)


pom.xml
```diff
@@ -189,7 +190,12 @@
       <version>${junit-jupiter.version}</version>
       <scope>test</scope>
     </dependency>
-
+    <dependency>
+      <groupId>org.junit.platform</groupId>
```
reta, line 194: For some reason, surefire was not picking up launcher :(

codefromthecrypt (approved):
maybe we should wait to release it until armeria updates to spring boot 3.5 just in case there is a change there @openzipkin/armeria

reta:
> maybe we should wait to release it until armeria updates to spring boot 3.5 just in case there is a change there @openzipkin/armeria

Will hold on with release, thanks @codefromthecrypt !

Issue #3796, opened 2025-02-17
Zipkin not working with Opensearch - appears to be double encoding UTF-8


mikebars:
## Describe the Bug
Zipkin fails to start when using Opensearch (but succeeds when using Elasticsearch)

## Steps to Reproduce

1. run `docker compose up` and wait for containers
2. run `curl --verbose http://0.0.0.0:9200` to see Elasticsearch / Opensearch information
3. run `curl --verbose http://0.0.0.0:9411/health` to see Zipkin health

---

### Elasticsearch (working)

`docker-compose.yml`:

```yaml
services:
  elasticsearch:
    container_name: elasticsearch
    environment:
      - _JAVA_OPTIONS=-Xms512m -Xmx512m -XX:UseSVE=0
      - action.destructive_requires_name=false
      - discovery.type=single-node
      - http.host=0.0.0.0
      - transport.host=127.0.0.1
      - xpack.monitoring.collection.enabled=false
      - xpack.security.enabled=false
      - xpack.security.http.ssl.enabled=false
    healthcheck:
      interval: 5s
      retries: 10
      start_period: 10s
      test: curl --silent http://localhost:9200/_cluster/health | grep --extended-regexp '"status":"(green|yellow)"'
      timeout: 10s
    image: elastic/elasticsearch:8.17.2
    restart: on-failure
    ports:
      - "9200:9200"
      - "9300:9300"

  zipkin:
    container_name: zipkin
    depends_on:
      elasticsearch:
        condition: service_healthy
    environment:
      - ES_HOSTS=http://elasticsearch:9200
      - JAVA_OPTS=-XX:UseSVE=0
      - STORAGE_TYPE=elasticsearch
    image: openzipkin/zipkin:latest
    ports:
      - "9411:9411"
    restart: on-failure
```

---

output of `curl --verbose http://0.0.0.0:9200`:

```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying 0.0.0.0:9200...
* Connected to 0.0.0.0 (127.0.0.1) port 9200 (#0)
> GET / HTTP/1.1
> Host: 0.0.0.0:9200
> User-Agent: curl/7.88.1
> Accept: */*
> 
< HTTP/1.1 200 OK
< X-elastic-product: Elasticsearch
< content-type: application/json
[...]

PR #3809, opened 2025-08-28
Fix invalid comparison of Elasticsearch version string

tmurakam:
Must use equals() or switch-case to compare Strings.

See https://github.com/openzipkin/zipkin/issues/3796#issuecomment-3227538965


codefromthecrypt (approved):
good catch

PR #3816, opened 2026-03-22
Updates docker images to latest, and matches maven version to it

codefromthecrypt:


codefromthecrypt:
rebased on the sonatype fixed master

PR #3820, opened 2026-03-23
Remove executable jar which failed sonatype validation on release

codefromthecrypt:
also revert the bump as the release never deployed. finally get rid of non LTS warnings for node.

### DIFF
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -11,7 +11,7 @@
   <parent>
     <groupId>io.zipkin</groupId>
     <artifactId>zipkin-parent</artifactId>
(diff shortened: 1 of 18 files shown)


zipkin-server/pom.xml
```diff
@@ -577,7 +577,6 @@
         <version>${spring-boot.version}</version>
         <configuration>
           <mainClass>zipkin.server.ZipkinServer</mainClass>
-          <executable>true</executable>
```
codefromthecrypt, line 580: this is no longer allowed in sonatype, but we have no docs suggesting to ./zipkin.jar anyway

PR #3821, opened 2026-03-23
Updates Zipkin to run on JRE 25, floor JDK 21

codefromthecrypt:
JDK 25 can still compile down to Java 8 so we don't need anything special here.

This also updates all test images we can to run on 25, except those who use SecurityManager:
* Elasticsearch (all versions) and OpenSearch
* Cassandra 4.x (we don't use 5.x and figuring that out is out of scope)

** This repo floor is bytecode 8: brave and zipkin-reporter-java are the only ones IIRC that must compile bytecode 1.6

### DIFF
--- a/.github/workflows/create_release.yml
+++ b/.github/workflows/create_release.yml
@@ -22,13 +22,13 @@ jobs:
         uses: actions/setup-java@v5
         with:
           distribution: 'zulu'  # zulu as it supports a wide version range
-          java-version: '17'  # earliest LTS supported by Spring Boot 3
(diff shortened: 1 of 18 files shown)


codefromthecrypt:
ok not so lucky, might need to bump some other things, which is fine as long as compile targets are the same

reta (approved):

codefromthecrypt:
opensearch 2.x also uses security manager so we have to leave that pinned to Java 21

codefromthecrypt:
thank god

